PR #3, opened 2019-03-13
membership module

mnaamani:
Adding membership module.


src/membership.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    fn validate_text(text: &Vec<u8>) -> Vec<u8> {
+        let mut text = text.clone();
+        text.truncate(Self::max_about_text_length() as usize);
+        text
+    }
+
+    fn validate_avatar(uri: &Vec<u8>) -> Vec<u8> {
+        let mut uri = uri.clone();
+        uri.truncate(Self::max_avatar_uri_length() as usize);
```
siman: Not sure that truncate is a good approach for avatar: if you truncate an URL then it became invalid.

mnaamani: We have to limit the length somehow. We can accommodate a longer url  upto 2000 ? https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

src/membership.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    type SubscriptionId: Parameter + Member + SimpleArithmetic + Codec + Default + Copy + As<usize> + As<u32> + MaybeSerializeDebug;
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
+#[derive(Encode, Decode)]
+pub struct Profile<T: Trait> {
+    id: T::MemberId, // is it necessary to have the id in the struct?
+    handle: u32,
+    avatarUri: Vec<u8>,
```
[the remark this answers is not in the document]
mnaamani: decided on default empty vector == empty string to represent same as not setting avatar or about text. Also makes the batch updating logic simpler.. how to differentiate between user wanting to pass None to 'unset' avatar, or None as in they don't want to update the avatar

src/membership.rs
siman: The same as with a handle:
`ensure!(uri.len() <= Self::max_avatar_uri_length() as usize, "avatar uri is too long");`

src/membership.rs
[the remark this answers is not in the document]
siman: Good question. In such case, it should be `Option<Option<Vec<u8>>>` and then if you want to set `avatar_uri` to None you will pass it with `UserInfo` as `Some(None)`. And if you don't want to update it, then just `None`.

mnaamani:
Yeah interesting. Does parity-coded know how to encode Option<Option<T>>
and could it be implemented in polkadot-js-apps as custom type I wonder.

src/membership.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    fn _change_member_handle(who: &T::AccountId, handle: Vec<u8>) -> dispatch::Result {
+        let mut profile = Self::ensure_has_profile(who)?;
+        Self::validate_handle(&handle)?;
+        Self::ensure_unique_handle(&handle)?;
+        <Handles<T>>::remove(&profile.handle);
+        <Handles<T>>::insert(handle.clone(), profile.id);
+        profile.handle = handle;
+        <MemberProfile<T>>::insert(profile.id, profile);
```
siman: Missing an event `MemberUpdatedHandle`

src/membership.rs
```diff
@@ -0,0 +1,381 @@
...
+            Self::_change_member_avatar(&who, &uri)?;
+        }
+
+        /// Change member's handle.
+        fn change_member_handle(origin, handle: Vec<u8>) {
+            let who = ensure_signed(origin)?;
+            Self::_change_member_handle(&who, handle)?;
+        }
+
+        fn batch_change_member_profile(origin, user_info: UserInfo) {
```
siman: maybe just `update_profile`?
```suggestion
        fn update_profile(origin, user_info: UserInfo) {
```

PR #4, opened 2019-03-20
Integrate Travis CI into master branch

KillaBoi:
Created .travis.yml with relavant content required for successful testing to see whether if the proposed PR is successful and whether if it fails.

Caching has not yet been added however this is WIP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ deploy:
   - provider: script
     script: bash setup.sh
     on:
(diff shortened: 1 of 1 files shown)


mnaamani:
Please direct the PR to the development branch.

I think we will only need travis to do a single job with rust stable, for testing native build and running cargo test.

When building the wasm runtime, the setup.sh can pull down the nightly toolchain.

PR #5, opened 2019-03-21
Add CI integration to Substrate Runtime Updates/Upgrades

KillaBoi:
Creation of Travis file into Development branch.
•Adjusted the guard on master branch only checks
•Removed other versions of Rust that are not required.

---------------------------------------------------------------
TODO:
•Need to add caching to optimise travis build
•Need to check out Docker build to see if it is faster and can be optimised
•Hashing checks for travis builds so that they can be compared to the build users make


KillaBoi:
it'll only do the CI check once for this PR for some weird reason

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+          command: |
+            rustup toolchain install nightly
+            rustup run nightly rustc --version --verbose
+            rustup run nightly cargo --version --verbose
+            rustup run nightly cargo build
+            rustup target add wasm32-unknown-unknown --toolchain nightly
+
+      - run:
+          name: Test
+          command: ./setup.sh
```
mnaamani: This step does the same installation as the "Nightly Build" so its not really needed.
But you called it test, so maybe you meant it to run `cargo test` instead?

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+      - run:
+          name: Setup
```
mnaamani: Perhaps rename the step to Build, since it is running the build script

KillaBoi: Have done, by mistake I accidentally did it the other way round and was failing for me

.travis.yml.example
```diff
@@ -0,0 +1,15 @@
...
+rust:
+  - 1.32.0
+cache: 
+  - rust
+  - cargo
+jobs:
+  include:
+    - stage: setup
+    - provider: script
+      script: bash ci.sh
```
mnaamani: why introduce a new ci.sh script, can't we just run `setup.sh` and `build.sh` sequentially?

KillaBoi: Some of it was experimentation, but I've got it working now, this was because of me having issues with Travis and it taking so long to compile...

PR #6, opened 2019-03-23
Staked Roles

mnaamani:
Builds on membership module, #3 

Some refactoring of council elections, and runtime upgrade proposals module to only allow members to participate.

Adding a simple staking mechanism to enter the new roles introduced in athens. It follows a similar mechanism to validator staking, but allows for using a separate account than will be associated with a members account. A member can enter a role (become an actor for a role) with these steps:

1. create a new account/keypair (the actor account)
2. tx: with actor account, register a request to enter a specific role and use a particular member identity
3. tx: (any account) transfer some tokens into actor account
4. tx: using member account, approve request generated in step 2. If there is 'slot' to fill for the role
    and there is enough balance in the actor account it will be bonded and role is entered into.

Still todo:
- [x] paying out rewards
- [x] governance functions to
  - add/udpdate roles and role parameters
  - eject under performing actors
- [x] tests


src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub reward: BalanceOf<T>,
+
+    // payouts are made at this block interval
+    pub reward_period: T::BlockNumber,
+
+    // how long tokens remain locked for after unstaking
+    pub unbonding_period: T::BlockNumber,
+
+    // minimum amount of time before being able to unstake
+    pub bonding_time: T::BlockNumber,
```
siman: The name looks inconsistent with `reward_period ` and `unbonding_period `
```suggestion
    pub bonding_period: T::BlockNumber,
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+
+
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Encode, Decode, Copy, Clone, Eq, PartialEq)]
+pub struct RoleParameters<T: Trait> {
+    // minium balance required to stake to enter a role
+    pub min_stake: BalanceOf<T>,
+
+    // the maximum number of spots available to fill for a role
+    pub max_actors: u32,
```
siman: Better to swap a position: first `min_actors`, then `max_actors`.
It's easier for perception like in counting: from smaller to bigger.

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+decl_storage! {
+    trait Store for Module<T: Trait> as Actors {
+        /// requirements to enter and maintain status in roles
+        Parameters get(parameters) : map Role => Option<RoleParameters<T>>;
+
+        /// the roles members can enter into
+        AvailableRoles get(available_roles) : Vec<Role>; // = vec![Role::Storage];
+
+        /// Actors list
+        Actors get(actors) : Vec<T::AccountId>;
```
siman: Actually, this is a list of actors' account ids, not Actors
```suggestion
        ActorIds get(actor_ids) : Vec<T::AccountId>;
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        /// actor accounts mapped to their actor
+        ActorsByAccountId get(actors_by_account_id) : map T::AccountId => Option<Actor<T>>;
```
siman: Actor, not Actors, because you return option of an actor, not a list of actors.
```suggestion
        ActorByAccountId get(actor_by_account_id) : map T::AccountId => Option<Actor<T>>;
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        /// actor accounts associated with a role
+        AccountsByRole get(accounts_by_role) : map Role => Vec<T::AccountId>;
```
siman:
```suggestion
        AccountIdsByRole get(account_ids_by_role) : map Role => Vec<T::AccountId>;
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        /// actor accounts associated with a member id
+        AccountsByMember get(accounts_by_member) : map MemberId<T> => Vec<T::AccountId>;
```
siman:
```suggestion
        AccountIdsByMemberId get(account_ids_by_member_id) : map MemberId<T> => Vec<T::AccountId>;
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+
+decl_event! {
+    pub enum Event<T> where
+      <T as system::Trait>::AccountId {
+        Staked(AccountId, Role),
+    }
+}
+
+impl<T: Trait> Module<T> {
+    fn role_is_available(role: Role) -> bool {
```
siman:
```suggestion
    fn is_role_available(role: Role) -> bool {
```

src/roles/actors.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+decl_module! {
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        fn deposit_event<T>() = default;
+
+        fn on_initialise(now: T::BlockNumber) {
+            // clear expired requests
+            if now % T::BlockNumber::sa(DEFAULT_REQUEST_CLEARING_INTERVAL) == T::BlockNumber::zero() {
```
siman: Consider creating an extra variable just to make code easier to read, then you can get rid of a comment:
```suggestion
            let is_request_expired = now % T::BlockNumber::sa(DEFAULT_REQUEST_CLEARING_INTERVAL) == T::BlockNumber::zero()
            if is_request_expired {
```

src/roles/actors.rs
mnaamani: yep caught that last night and fixed.

src/roles/actors.rs
mnaamani: I see you point in general, but in this exact case the that check is just a check to see if we reached the interval where we actually check each request.

siman (approved):

Issue #7, opened 2019-03-25
Implement Data Object Type Registry


jfinkhaeuser:
- [x] Data Object Type field
- [x] Registry as map of ID to DOT
- [x] Modify only as root
- [x] Tests

jfinkhaeuser:
Resolved with that merge.

PR #13, opened 2019-03-25
For #7, implement simple Data Object Type registry

jfinkhaeuser:
Should be the majority of the DOT registry. Tests are missing at this point; I want to add them as well.


jfinkhaeuser:
So, with the tests added, it'd be great to get some feedback from @mnaamani and @siman :)

siman:
A pull requested should go into the `development` branch, not `master`. Please check this PR: https://github.com/Joystream/substrate-runtime-joystream/pull/3

src/storage/types.rs
```diff
@@ -0,0 +1,138 @@
...
+
+            dot.description = data_object_type.description.clone();
+            dot.active = data_object_type.active;
+
+            <DataObjectTypeMap<T>>::insert(id, dot);
+
+            Self::deposit_event(RawEvent::DataObjectTypeUpdated(id));
+        }
+
+        pub fn activate_data_object_type(origin, id: T::DataObjectTypeID, active: bool)
```
mnaamani: The function name implies parameter `active` shouldn't be passed. Perhaps remove parameter and add counterpart `fn deactive_data_object_type(origin, id: T::DataObjectTypeID)` each function setting the `active` field correspondigly.

...or changing name to `change_active_status_of_data_object_type(..)`

src/storage/types.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+impl <T: Trait> Module<T>
+{
+    fn ensure_data_object_type(id: T::DataObjectTypeID) -> Result<DataObjectType<T>, &'static str>
+    {
+        let dot = Self::data_object_type(&id).ok_or(MSG_DOT_NOT_FOUND)?;
+        Ok(dot)
```
mnaamani: It can be one liner expression, since ok_or() returns a `Result`
```
fn ensure_data_object_type(id: T::DataObjectTypeID) -> Result<DataObjectType<T>, &'static str> {
    Self::data_object_type(&id).ok_or(MSG_DOT_NOT_FOUND)
}
```

src/storage/types.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait> as DataObjectTypeRegistry
+    {
+        // Start at this value
+        pub FirstDataObjectTypeID get(first_data_object_type_id) config(first_data_object_type_id): T::DataObjectTypeID = T::DataObjectTypeID::sa(DEFAULT_FIRST_DATA_OBJECT_TYPE_ID);
+
+        // Increment
+        pub NextDataObjectTypeID get(next_data_object_type_id) build(|config: &GenesisConfig<T>| config.first_data_object_type_id): T::DataObjectTypeID = T::DataObjectTypeID::sa(DEFAULT_FIRST_DATA_OBJECT_TYPE_ID);
```
mnaamani: Just to point out that the build() method is only executed (native) when initializing the genesis block, but not for a runtime that is an update to a live chain. However this is not an issue in this case because you are setting a default value.

src/lib.rs
```diff
@@ -244,6 +252,7 @@ construct_runtime!(
 		CouncilElection: election::{Module, Call, Storage, Event<T>, Config<T>},
 		Council: council::{Module, Call, Storage, Event<T>, Config<T>},
 		Memo: memo::{Module, Call, Storage, Event<T>},
+		DataObjectType: types::{Module, Call, Storage, Event<T>, Config<T>},
```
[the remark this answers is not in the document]
jfinkhaeuser: I sort of agree. I mean, I agree with the aesthetics, I just don't know what the best thing to do here is. I think it should be called `DataObjectTypeRegistry`, and the module should be called `registry` to keep things simpler, because that's what the whitepaper said.

WDYT?

src/storage/types.rs
jfinkhaeuser: Yeah... I'm not a big fan of multiple functions for the same purpose, nor am I a fan of ultra long function names. I don't much like either suggestion, but I also agree that it could be better.

The correct name IMHO would include "toggle", but to be grammatically correct-ish, the name would then have to be something like `toggle_data_object_type_active_state` and that's super long again.

I think I'll go with a second function. My main objection to that is that if you do that for several flags, you have hundreds of single purpose functions. You can use the `update_...` function for the same thing, though, so this should only be a shortcut for toggling the active state.

src/storage/types.rs
jfinkhaeuser: Thanks! I missed that.

src/lib.rs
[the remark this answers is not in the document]
jfinkhaeuser: Pushed a new commit that should address this.

src/storage/types.rs
jfinkhaeuser: Fixed in another commit.

src/storage/types.rs
jfinkhaeuser: Also updated.

jfinkhaeuser:
Ok, resolved conflicts and changed everything according to feedback. Looks good, no?

src/storage/data_object_type_registry.rs
```diff
@@ -0,0 +1,154 @@
...
+use srml_support::{StorageMap, StorageValue, decl_module, decl_storage, decl_event, ensure, Parameter};
+use runtime_primitives::traits::{SimpleArithmetic, As, Member, MaybeSerializeDebug, MaybeDebug};
+use system::{self, ensure_root};
+use crate::traits;
+
+pub trait Trait: system::Trait + MaybeDebug
+{
+    type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+
+    type DataObjectTypeID: Parameter + Member + SimpleArithmetic + Codec + Default + Copy
```
siman: By following Substrate and our naming convention for IDs, it should be: `ID` -> `Id`
```suggestion
    type DataObjectTypeId: Parameter + Member + SimpleArithmetic + Codec + Default + Copy
```

src/storage/data_object_type_registry.rs
```diff
@@ -0,0 +1,154 @@
...
+static MSG_REQUIRE_DO_TYPE_ID: &str = "Can only update Data Object Types that are already registered (with an ID)!";
+
+const DEFAULT_FIRST_DATA_OBJECT_TYPE_ID: u64 = 1;
+
+#[derive(Clone, Encode, Decode, PartialEq)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct DataObjectType<T: Trait>
+{
+    // If the OT is registered, an ID must exist, otherwise it's a new OT.
+    pub id: Option<T::DataObjectTypeID>,
```
siman: I would suggest changing `id` type to just `T::DataObjectTypeId` and creating an additional struct just for updates:
```
pub struct DataObjectTypeUpdate<T: Trait> {
  description: Option<Vec<u8>>,
  active: Option<bool>,
}
```

And in `update_*` function, update only those fields of stored value that are `Some` in `DataObjectTypeUpdate`

src/storage/data_object_type_registry.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use rstd::prelude::*;
+use parity_codec::Codec;
+use parity_codec_derive::{Encode, Decode};
+use srml_support::{StorageMap, StorageValue, decl_module, decl_storage, decl_event, ensure, Parameter};
+use runtime_primitives::traits::{SimpleArithmetic, As, Member, MaybeSerializeDebug, MaybeDebug};
+use system::{self, ensure_root};
+use crate::traits;
+
+pub trait Trait: system::Trait + MaybeDebug
```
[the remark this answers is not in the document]
jfinkhaeuser: I'm fine with changing that, but you're likely going to see this a lot from me :)

src/storage/data_object_type_registry.rs
```diff
@@ -0,0 +1,154 @@
...
+        // Mapping of Data object types
+        pub DataObjectTypeMap get(data_object_type): map T::DataObjectTypeID => Option<DataObjectType<T>>;
+    }
+}
+
+decl_event! {
+    pub enum Event<T> where
+        <T as Trait>::DataObjectTypeID
+    {
+        DataObjectTypeAdded(DataObjectTypeID),
```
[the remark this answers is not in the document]
jfinkhaeuser: Makes sense, will change.

src/storage/data_object_type_registry.rs
jfinkhaeuser: Yeah, that's a naming convention that I've seen a lot in Java, and it's pretty stupid. ID is not a word, it's an abbreviation, so CamelCasing it makes no sense at all. Still, I will change it.

src/storage/data_object_type_registry.rs
jfinkhaeuser: Ok, why? I suppose it may keep transaction sizes down. I suppose that could be worth it, but is that the rationale?

src/storage/data_object_type_registry.rs
siman: Here is a quote from [ID article](https://en.wikipedia.org/wiki/ID) on Wiki:

> ID, Id, id or I.D. may refer to:
> 
> * Identity document, any document used to verify a person's identity
> * Identifier, a symbol which uniquely identifies an object or record

I believe that in our cases (`AccounId`, `MemberId`, `ProposalId`) an `Id` stands for `Identifier` (i.e. **id**_entifier. ID could be considered as an abbreviation only in case if we reffer to **I**_dentity **D**_ocument – that is not our case at all.

src/storage/data_object_type_registry.rs
mnaamani: The tx size is certainly a benefit, but I think the rationale behind alex's suggestion is so that you wouldn't have to change the function signature in the future if you introduce new fields to theDataObjectType type. (I don't want to put words in his mouth but that is how I understood it when he recommended the same for the members module [update_profile()](https://github.com/Joystream/substrate-runtime-joystream/blob/b5926390b1bf9e94fb60cc1c17a615a14a8fb7e2/src/membership/members.rs#L241) method)

mnaamani: I actually didn't look at the function signature (i assumed there was one parameter per field).. so my point doesn't actually make sense.

siman: Because object's `id` is used quite often (at least on UI: list stored objects, view details of stored objects, etc.) and having it as an Option will require additional check `storedObj.id.isSome ? ...then : else...` everywhere. Comparing to update – this operation is rate and exists only on Update Form.

PR #19, opened 2019-03-27
Proposal for not having Option<T::DataObjectTypeId> in the DO type

jfinkhaeuser:
Don't merge immediately, @mnaamani - this is for discussion (and possibly merging later on).

So this PR gets the ID out of the struct, which I think is part of what @siman wanted.

I understand the utility of adding a separate update struct, but am not happy with having two structs with essentially the same members that could get out of sync if programmers don't pay attention. So this is my proposal for finishing #7, essentially.

If there's data structures that see updates to individual fields a lot, I think the separate update struct makes sense. Here, the only member that should get toggled a lot is the `active` flag, and that should happen rarely enough (tied to storage tranches coming post-Athens).

WDYT?

### DIFF
--- a/src/storage/data_object_type_registry.rs
+++ b/src/storage/data_object_type_registry.rs
@@ -15,17 +15,13 @@ pub trait Trait: system::Trait + MaybeDebug {
         + As<usize> + As<u64> + MaybeSerializeDebug + PartialEq;
 }
 
-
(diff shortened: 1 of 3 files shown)


mnaamani:
I think this is very sensible, and I like how much simpler the code is.

PR #23, opened 2019-03-29
disable and clear account indices

mnaamani:
Clear indices, and prevent new accounts from generating an AccountIndex.

https://github.com/Joystream/substrate-runtime-joystream/pull/2 for context

The base issue is in our testnet we allow accounts to go to zero balance and not get reaped, because we configured existential deposit to be zero. Without this change account index reuse will occur and if anyone is relying on the account index in place of full account id (mainly in transfers) they may be unpleasantly surprised.


mnaamani:
obsolete now, done in #28

PR #24, opened 2019-03-30
Implement data directory for #8

jfinkhaeuser:
Fixes #8 when merged.

Following the whitepaper is a little more complex than the issue bullet list describes.

Missing:
- [x] tests
- [ ] a README, describing the flows (from the whitepaper)

PR #25, opened 2019-03-30
Fix most warnings

jfinkhaeuser:
I'm constantly distracted by the warnings during development. This only makes sense after #24 is merged, so keep open for now.


mnaamani:
The line at the top of most of the files has been kept:

`#![cfg_attr(not(feature = "std"), no_std)]`

The compiler is giving us a warning about not using `no_std` attribute anywhere.

I understand that its important when compiling the runtime to WASM that we have to disable any crate from using `std` feature. This line is obviously related, but I don't understand it properly to know if it is safe to remove. Any ideas?

mnaamani:
Great that you took the time to get rid of all these warnings. ❤️ 
LGTM (didn't review the storage implementation)

jfinkhaeuser:
> The line at the top of most of the files has been kept:
> 
> `#![cfg_attr(not(feature = "std"), no_std)]`
> 
> The compiler is giving us a warning about not using `no_std` attribute anywhere.
> 
> I understand that its important when compiling the runtime to WASM that we have to disable any crate from using `std` feature. This line is obviously related, but I don't understand it properly to know if it is safe to remove. Any ideas?

I *think* it means we should have this only in `src/lib.rs`, but would have to try it out.

mnaamani:
oh right, that it only makes sense to place it in the crate root?

mnaamani:
This is just a nuance, but technically the title of the PR is fix warnings (we are not silencing them by setting a compiler flag) which is a good thing :)
I wonder if we should have a commit hook to prevent committing code with warnings? Perhaps for another time.

jfinkhaeuser:
> oh right, that it only makes sense to place it in the crate root?

I read it that way, but I'm not sure.

jfinkhaeuser:
There's also at least one warning I didn't do anything about, but you'll see that soonish, I think.

PR #28, opened 2019-04-02
Upgrade to substrate v1.0.0

mnaamani:
Working on updating runtime dependency on latest substrate v1.0.0rc1


mnaamani:
This now builds, and tests are passing.
Continuing work in https://github.com/Joystream/substrate-node-joystream/pull/51

Still need to add Grandpa and lock accounts staked in a role.

PR #26, opened 2019-04-01
Content directory

jfinkhaeuser:
Based off another branch, so makes sense only after #25 is merged.

However, this should be enough for a first stab at a content directory.

PR #29, opened 2019-04-03
Run rustfmt over everything

jfinkhaeuser:
Based off #26, so will only make sense after that. All this does is run `rustfmt` once over the code base so that subsequently edited files don't create huge diffs.


jfinkhaeuser:
Should sort of be a draft, until the previous PRs are merged. But it shouldn't need editing or review, so it's also sort of ready.

Issue #11, opened 2019-03-25
Storage provider list


jfinkhaeuser:
- [ ] Anyone can enter
- [ ] Kicked out by root

mnaamani:
Started on in https://github.com/Joystream/substrate-runtime-joystream/pull/6

jfinkhaeuser:
Perfect - sorry I didn't see that :)

mnaamani:
We should probably discuss on this, I think it might be worth having a separate specialized module for each role, and have the roles module just be responsible for entry and exist, and paying out rewards.

A specialized storage_role module could have some more logic and actually pick the 'active' providers from the staked roles modules based on its own set of rules, things like error reporting could be collected there

jfinkhaeuser:
Makes a lot of sense at this level or detail - let's do discuss this :)

For Athens, though, storage providers aren't particularly controlled - so the specialized module would be very slim.

jfinkhaeuser:
@siman this is done with the staked roles, no? Do you still have modifications in a PR, or should I close this?

jfinkhaeuser:
Apparently that's done.

PR #31, opened 2019-04-04
Add use of staked storage provider roles to storage subsystem.

jfinkhaeuser:
Implements #11 

@mnaamani There's a problem here in that some tests panic. I think I may not be initializing the actor storage with test data, but it's hard for me to figure out what I should do. Do you have an insight here?

### DIFF
--- a/src/roles/actors.rs
+++ b/src/roles/actors.rs
@@ -197,12 +197,7 @@ impl<T: Trait> Roles<T> for Module<T> {
     }
 
     fn account_has_role(account_id: &T::AccountId, role: Role) -> bool {
-        let res = Self::actor_by_account_id(account_id);
-        if res.is_none() {
-            return false;
-        }
-        let actor = res.unwrap();
-        actor.role == role
+        Self::actor_by_account_id(account_id).map_or(false, |actor| actor.role == role)
     }
 
     fn random_account_for_role(role: Role) -> Result<T::AccountId, &'static str> {
--- a/src/roles/actors.rs
+++ b/src/roles/actors.rs
@@ -12,6 +12,8 @@ use system::{self, ensure_signed};
 
 use crate::traits::{Members, Roles};
 
+static MSG_NO_ACTOR_FOR_ROLE: &str = "For the specified role, no actor is currently staked.";
+
 #[derive(Encode, Decode, Copy, Clone, Eq, PartialEq, Debug)]
 pub enum Role {
     Storage,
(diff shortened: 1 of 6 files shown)


src/roles/actors.rs
```diff
@@ -193,6 +195,29 @@ impl<T: Trait> Roles<T> for Module<T> {
...
+        if 0 == ids.len() {
+            return Err(MSG_NO_ACTOR_FOR_ROLE);
+        }
+        let seed = <system::Module<T>>::random_seed();
+        let mut rand: u64 = 0;
+        for offset in 0..8 {
+            rand += (seed.as_ref()[offset] as u64) << offset;
+        }
+        let idx = (rand as usize) % ids.len();
+        return Ok(ids[idx].clone());
```
mnaamani: I don't think `clone` is necessary, the value can be moved safely at this point since there is no use of ids after this line. The return statement can also be changed into an expression which will be the return value:

```Ok(ids[idx])```

File: src/storage/data_object_storage_registry.rs [hunk 2]
```diff
@@ -104,10 +108,9 @@ decl_module! {
         pub fn add_relationship(origin, cid: T::ContentId) {
             // Origin has to be a storage provider
             let who = ensure_signed(origin)?;
-            // TODO check for being staked as a storage provider
-            // if !T::Members::is_active_member(&who) {
-            //     return Err(MSG_CREATOR_MUST_BE_MEMBER);
-            // }
+
+            // Check that the origin is a storage provider
+            ensure!(<T as Trait>::Roles::account_has_role(&who, actors::Role::Storage), MSG_ONLY_STORAGE_PROVIDER_MAY_CREATE_DOSR);
```
mnaamani: I don't think the cast to Trait is necessary here.

src/roles/actors.rs
mnaamani: my bad actually, cloning is needed or you get `cannot move out of borrowed content`

src/storage/data_object_storage_registry.rs
mnaamani: I was wrong again.. :)

mnaamani: I think its because the Roles type name is on multiple module traits, so the cast is needed to 'pick' the correct one. (disambiguation)

mnaamani:
LGTM

src/roles/actors.rs
```diff
@@ -193,6 +195,29 @@ impl<T: Trait> Roles<T> for Module<T> {
...
     }
+
+    fn account_has_role(account_id: &T::AccountId, role: Role) -> bool {
+        let res = Self::actor_by_account_id(account_id);
+        if res.is_none() {
+            return false;
+        }
+        let actor = res.unwrap();
+        actor.role == role
+    }
```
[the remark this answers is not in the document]
jfinkhaeuser: Hmm. In all honesty, I find that far less readable.

Ah, well, let's do it.

PR #42, opened 2019-04-17
Improved role staking

mnaamani:
Here is my proposal for improving the staking mechanism for the Actors/Roles.

Locked amount. We were locking the entire balance in the actor account.

The amount locked is now only the minimum stake required to enter role. Unlike the validators set, currently there is no elaborate selection algorithm for who makes it into the set of actors for a specific role, only criteria is staking minimum amount and there being an empty slot. (In future a dedicated working group will have responsibility of managing this set) So there is no need to lock the entire account balance, or even an advantage for a member to add more than minimum stake to their actor account.

Actor account can still withdraw for following reasons: `TransactionPayment`, and `Fee`. This allows for covering transaction fees for any extrinsic sent by actor account. The account cannot transfer balance or reserve while it is active in its role or unbonding.

Rewards are paid out to actor account if its balance falls below the minimum stake requirement. (This will happen for two reasons, spending on tx fees and if the minimum stake is increased). Otherwise it goes to the member account.

### DIFF
--- a/src/roles/actors.rs
+++ b/src/roles/actors.rs
@@ -2,7 +2,7 @@ use crate::currency::{BalanceOf, GovernanceCurrency};
 use parity_codec_derive::{Decode, Encode};
 use rstd::prelude::*;
 use runtime_primitives::traits::{As, Bounded, MaybeDebug, Zero};
-use srml_support::traits::{Currency, LockIdentifier, LockableCurrency, WithdrawReason};
+use srml_support::traits::{Currency, LockIdentifier, LockableCurrency, WithdrawReasons, WithdrawReason};
 use srml_support::{decl_event, decl_module, decl_storage, ensure, StorageMap, StorageValue};
 use system::{self, ensure_signed};
 
@@ -224,15 +224,15 @@ impl<T: Trait> Module<T> {
         Self::remove_actor_from_service(actor_account, role, member_id);
     }
 
-    // Locks account and prevents transfers and reservation. Account still pay basic
-    // transaction fees and therefore send transactions that don't demand reserving balance
+    // Locks account and only allows paying for transaction fees. Account cannot
+    // transfer or reserve funds.
     fn update_lock(account: &T::AccountId, stake: BalanceOf<T>, until: T::BlockNumber) {
         T::Currency::set_lock(
             STAKING_ID,
             account,
             stake,
             until,
-            WithdrawReason::Transfer | WithdrawReason::Reserve,
+            WithdrawReasons::all() & !(WithdrawReason::TransactionPayment | WithdrawReason::Fee),
         );
     }
 }
--- a/src/roles/actors.rs
+++ b/src/roles/actors.rs
@@ -215,15 +215,25 @@ impl<T: Trait> Module<T> {
         stake: BalanceOf<T>,
     ) {
         // simple unstaking ...only applying unbonding period
-        T::Currency::set_lock(
-            STAKING_ID,
+        Self::update_lock(
             &actor_account,
             stake,
             <system::Module<T>>::block_number() + unbonding_period,
-            WithdrawReason::Transfer | WithdrawReason::Reserve,
         );
 
         Self::remove_actor_from_service(actor_account, role, member_id);
+    }
+
+    // Locks account and prevents transfers and reservation. Account still pay basic
+    // transaction fees and therefore send transactions that don't demand reserving balance
+    fn update_lock(account: &T::AccountId, stake: BalanceOf<T>, until: T::BlockNumber) {
+        T::Currency::set_lock(
+            STAKING_ID,
+            account,
+            stake,
+            until,
+            WithdrawReason::Transfer | WithdrawReason::Reserve,
+        );
     }
 }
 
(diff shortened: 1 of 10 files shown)


src/roles/actors.rs
```diff
@@ -212,19 +212,29 @@ impl<T: Trait> Module<T> {
...
+        stake: BalanceOf<T>,
     ) {
         // simple unstaking ...only applying unbonding period
-        let value = T::Currency::free_balance(&actor_account);
-        T::Currency::set_lock(
-            STAKING_ID,
+        Self::update_lock(
             &actor_account,
-            value,
+            stake,
```
jfinkhaeuser: This is IMHO a little tricky.

Previously, actors staked their entire balance, so updating the lock for their balance makes some sense. Now you've updated this to apply not to the entire balance, but a chosen amount. So far, so good.

However, when you look further below where this function is called, it always passes the minimum staking balance for the role here. So consider this scenario:

- Account stakes when minimum staking balance is A.
- Role parameters change minimum staking balance to B.
- Account unstakes when minimum staking balance is B.

If A > B, the actor wouldn't get their entire amount unstaked again, which is definitely an issue. If B > A, perhaps unstaking fails completely?

I think the approach is correct, but it seems you then also have to record the exact amount staked when it is staked, and then unstake exactly that amount, no more and no less.

src/roles/actors.rs
```diff
@@ -212,19 +212,29 @@ impl<T: Trait> Module<T> {
...
+
+    // Locks account and prevents transfers and reservation. Account still pay basic
+    // transaction fees and therefore send transactions that don't demand reserving balance
+    fn update_lock(account: &T::AccountId, stake: BalanceOf<T>, until: T::BlockNumber) {
+        T::Currency::set_lock(
+            STAKING_ID,
+            account,
+            stake,
+            until,
+            WithdrawReason::Transfer | WithdrawReason::Reserve,
```
jfinkhaeuser: I don't know all the possible `WithdrawReason` values, so trust you here.

This is a little brittle with regards to changes, if e.g. more `WithdrawReason` values are added. You could start with `::all()` and exclude transaction fees, perhaps?

src/roles/actors.rs
```diff
@@ -277,25 +287,23 @@ decl_module! {
                     for actor in accounts.into_iter().map(|account| Self::actor_by_account_id(account)) {
                         if let Some(actor) = actor {
                             if now > actor.joined_at + params.reward_period {
-                                // send reward to member account - not the actor account
-                                if let Ok(member_account) = T::Members::lookup_account_by_member_id(actor.member_id) {
-                                    let _ = T::Currency::deposit_into_existing(&member_account, params.reward);
+                                // reward can top up balance if it is below minimum stake requirement
```
jfinkhaeuser: This looks like the right change, but for two comments:

1. The reward must still exceed the balance used up during the lifetime of a block for anything to ever be paid out to the member *and* to guarantee that the actor can continue acting. I'm not sure this is enough, therefore.
1. The effect of this is that busy actors actually produce less reward for their member than idle actors. I think that's the opposite of what's intended.

Maybe it's enough for getting Athens up and running, because the storage provider incentives are really only properly to be added in future. But it really doesn't seem entirely complete at this point.

src/roles/actors.rs
```diff
@@ -376,13 +385,20 @@ decl_module! {
...
-            Self::apply_unstake(actor.account.clone(), actor.role, actor.member_id, role_parameters.unbonding_period);
+            Self::apply_unstake(actor.account.clone(), actor.role, actor.member_id, role_parameters.unbonding_period, role_parameters.min_stake);
 
             Self::deposit_event(RawEvent::Unstaked(actor.account, actor.role));
         }
 
         pub fn set_role_parameters(role: Role, params: RoleParameters<T>) {
+            let new_stake = params.min_stake.clone();
             <Parameters<T>>::insert(role, params);
+            // Update locks for all actors in the role. The lock for each account is already until max_value
```
jfinkhaeuser: Increasing the minimum staking amount while actors are already staked would presumably fail to update the lock for those who don't have excess balance. It's likely that you have to gracefully unstake these actors, or something?

ghost (changes requested):
I think this is the right direction to go, but I think I also identified a bunch of edge cases that need covering. Maybe not all need a solution right now, but some should IMHO be addressed.

src/roles/actors.rs
jfinkhaeuser: e.g.
```rust
WithdrawReason::all() & ~WithdrawReason:TransactionFee // or whatever the reason is called
```

src/roles/actors.rs
mnaamani: The locking mechanism should not fail. It is permitted to lock more than the current balance.

src/roles/actors.rs
mnaamani: good point, will fix

src/roles/actors.rs
mnaamani: As we discussed IRL, these edge cases are correct, but would have to be solved by the working group in selecting the best set of role parameters to ensure storage providers are earning more than than their tx fees, as well as introducing a reward model that takes into account how busy the actors are.

src/roles/actors.rs
mnaamani: As we discussed, unstaking will not because of how locking mechanism in the balances module is implemented. The lock is a limit/restriction set on how funds can be withdrawn or transferred (it doesn't actually involve moving any funds).

jfinkhaeuser: Yes, that makes sense.

src/roles/actors.rs
jfinkhaeuser: My main concern for Athens is really that storage providers don't run out of balance for transaction fees - I guess it's okay for now to require people to pay attention.

src/roles/actors.rs
jfinkhaeuser: That was new to me, thanks!

jfinkhaeuser:
Ok, after the discussion, I think it's ok to move forward. I'd change the flags setting code just to be a bit safer, but otherwise let's merge this PR :)

mnaamani:
Updated flag setting code and tested locally works as expected. This is good to merge

PR #46, opened 2019-04-18
add GPLv3 License

mnaamani:


mnaamani:
Are we going with GPLv3, if so we can merge this.

dagurval:
Why is this boilerplate needed?

PR #50, opened 2019-04-26
storage: use primary storage liason as fallback

### LINKED ISSUE #49: Fix liaison handling
See https://github.com/Joystream/storage-node-joystream/issues/57

Rather than having a special hardcoded liaison value that is queried from the app, instead the function that picks the liaison should always return the same value.

mnaamani:
Fixes #49


ghost (approved):
Works for me.

PR #55, opened 2019-04-29
pin patch version of syn package

mnaamani:
Pin the minor patch version of the syn package to avoid using version 0.15.32 which introduced a breaking internal change. (see https://github.com/paritytech/substrate/issues/2326 for background)

This fixes the compile error:

```rust
   Compiling srml-support-procedural-tools v1.0.0 (https://github.com/joystream/substrate.git?rev=6dfc3e8b057bb00322136251a0f10305fbb1ad8f#6dfc3e8b)
error[E0432]: unresolved import `syn::token::CustomKeyword`
  --> /Users/bedeho/.cargo/git/checkouts/substrate-21e9e9267949a89d/6dfc3e8/srml/support/procedural/tools/src/syn_ext.rs:26:5
   |
26 | use syn::token::CustomKeyword;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `CustomKeyword` in `token`

```

Although it is the srml module (more than one actually) that depends on this package, adding the dependency in our library Config.toml seems to override the dependency on the lower modules, so I didn't have to fork substrate branch which is a good thing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,11 @@
 authors = ['Joystream']
 edition = '2018'
 name = 'joystream-node-runtime'
-version = '5.3.0'
+version = '5.4.0'
+
+[dependencies]
+# restrict version of syn, see https://github.com/paritytech/substrate/issues/2326
+syn = { version = ">= 0.15.26, < 0.15.32", features = ["full"] }
 
 [features]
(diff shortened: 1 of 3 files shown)


bedeho:
Worked!

Issue #36, opened 2019-04-11
Fix test for storage module and add new


siman:

mnaamani:
Implemented in #37

PR #57, opened 2019-05-07
Example: Integrating example-module into runtime

mnaamani:
== THIS PR IS FOR DEMONSTRATION ONLY - DO NOT MERGE ==

Example of how to integrate an [external module](https://github.com/Joystream/runtime-example-module) into the runtime.

Demonstrates how to apply the [recommended convention](https://github.com/Joystream/joystream/issues/36) to make a module easily reusable for another runtime, by providing your own traits for your expectations on other modules, rather than relying on public traits, or traits in the module you are expecting to use. In this example we apply this to the "membership" system.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,14 @@ std = [
     'consensus-authorities/std',
     'grandpa/std',
     'finality-tracker/std',
+    'runtime-example-module/std',
 ]
+
+[dependencies.runtime-example-module]
+git = 'https://github.com/joystream/runtime-example-module.git'
+default_features = false
+rev = 'a14d1433ab9691170cd0b68fd6a82197ee7c61b9'
+
 [dependencies.aura]
 default_features = false
 git = 'https://github.com/joystream/substrate.git'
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -37,7 +37,14 @@ std = [
...
     'grandpa/std',
     'finality-tracker/std',
+    'runtime-example-module/std',
 ]
+
+[dependencies.runtime-example-module]
+git = 'https://github.com/joystream/runtime-example-module.git'
+default_features = false
+rev = 'a14d1433ab9691170cd0b68fd6a82197ee7c61b9'
+
```
mnaamani: During development before pushing a repo you might also want to use a path dependency:

```
[dependencies.runtime-example-module]
path = '../path-to-local-repo/example-module'
default_features = false
```

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

mnaamani:
closing as it is for demonstration purposes only

PR #70, opened 2019-06-22
roles: fix remove_actor implementation

mnaamani:
Was incorrectly interpreting `actor_account` as a member account


bwhm (approved):
Looks good to me!

PR #81, opened 2019-09-11
Add new roles functionality to membership module

mnaamani:
Changes to membership module proposed in https://github.com/Joystream/joystream/issues/112

Building on #80 (better to merge that first to make this PR easier to review)

### DIFF
--- a/src/membership/members.rs
+++ b/src/membership/members.rs
@@ -575,15 +575,20 @@ impl<T: Trait> Module<T> {
         role_id: T::RoleId,
         actor_id: T::ActorId,
     ) -> Result<(), &'static str> {
-        // limits - how many roles in total
-        //        - single instance of role
+        // For now default policy across all roles is:
+        // members can only have a single actor instance per role
+        // members can enter any roles
+        // no limit on total number of roles a member can enter
+        // Note: Role specific policies, for example "member can only enter council role once at a time"
+        // should be enforced by the council module (client modules)
+        ensure!(!Self::member_is_in_role(member_id, role_id), "member already in role");
 
         // ensure is active member
         let profile = Self::ensure_profile(member_id)?;
         ensure!(!profile.suspended, "suspended members can't enter a role");
 
+        // guard against duplicate ActorInRole
         let actor_in_role = ActorInRole { role_id, actor_id };
-        // ensure actor_id not already set for role
         ensure!(
             !<MembershipIdByActorInRole<T>>::exists(&actor_in_role),
             "role actor already exists"
--- a/src/membership/members.rs
+++ b/src/membership/members.rs
@@ -265,31 +265,31 @@ decl_module! {
         }
 
         /// Change member's about text
-        pub fn change_member_about_text(origin, member_id: T::MemberId, text: Vec<u8>) {
+        pub fn change_member_about_text(origin, text: Vec<u8>) {
             let controller = ensure_signed(origin)?;
-            Self::ensure_is_member_controller_account(controller.clone(), member_id)?;
+            let member_id = Self::ensure_is_member_controller_account(&controller)?;
             Self::_change_member_about_text(member_id, &text)?;
         }
 
         /// Change member's avatar
-        pub fn change_member_avatar(origin, member_id: T::MemberId, uri: Vec<u8>) {
+        pub fn change_member_avatar(origin, uri: Vec<u8>) {
             let controller = ensure_signed(origin)?;
-            Self::ensure_is_member_controller_account(controller.clone(), member_id)?;
+            let member_id = Self::ensure_is_member_controller_account(&controller)?;
             Self::_change_member_avatar(member_id, &uri)?;
         }
 
         /// Change member's handle. Will ensure new handle is unique and old one will be available
         /// for other members to use.
-        pub fn change_member_handle(origin, member_id: T::MemberId, handle: Vec<u8>) {
+        pub fn change_member_handle(origin, handle: Vec<u8>) {
             let controller = ensure_signed(origin)?;
-            Self::ensure_is_member_controller_account(controller.clone(), member_id)?;
+            let member_id = Self::ensure_is_member_controller_account(&controller)?;
             Self::_change_member_handle(member_id, handle)?;
         }
 
         /// Update member's all or some of handle, avatar and about text.
-        pub fn update_profile(origin, member_id: T::MemberId, user_info: UserInfo) {
+        pub fn update_profile(origin, user_info: UserInfo) {
             let controller = ensure_signed(origin)?;
-            Self::ensure_is_member_controller_account(controller.clone(), member_id)?;
+            let member_id = Self::ensure_is_member_controller_account(&controller)?;
 
             if let Some(uri) = user_info.avatar_uri {
                 Self::_change_member_avatar(member_id, &uri)?;
@@ -304,7 +304,7 @@ decl_module! {
 
         pub fn set_controller_key(origin, controller: T::AccountId) {
             let master = ensure_signed(origin)?;
-            let member_id = Self::ensure_is_member_primary_account(master)?;
+            let member_id = Self::ensure_is_member_primary_account(&master)?;
 
             let mut profile = Self::ensure_profile(member_id)?;
 
@@ -321,7 +321,7 @@ decl_module! {
 
         pub fn set_primary_key(origin, new_primary: T::AccountId) {
             let old_primary = ensure_signed(origin)?;
-            let member_id = Self::ensure_is_member_primary_account(old_primary.clone())?;
+            let member_id = Self::ensure_is_member_primary_account(&old_primary)?;
 
             // ensure new key not is used by someone else (master or controller)
             ensure!(!<MemberIdByControllerAccountId<T>>::exists(&new_primary), "account already paired with member");
@@ -373,30 +373,19 @@ decl_module! {
 
 impl<T: Trait> Module<T> {
     pub fn is_active_member(who: &T::AccountId) -> bool {
-        match Self::ensure_is_member(who).and_then(|member_id| Self::ensure_profile(member_id)) {
+        match Self::ensure_is_member_primary_account(who)
+            .and_then(|member_id| Self::ensure_profile(member_id))
+        {
             Ok(profile) => !profile.suspended,
             Err(_err) => false,
         }
     }
 
-    pub fn lookup_member_id(who: &T::AccountId) -> Result<T::MemberId, &'static str> {
-        Self::ensure_is_member(who)
-    }
-
-    pub fn lookup_account_by_member_id(id: T::MemberId) -> Result<T::AccountId, &'static str> {
+    pub fn primary_account_by_member_id(id: T::MemberId) -> Result<T::AccountId, &'static str> {
         if <AccountIdByMemberId<T>>::exists(&id) {
             Ok(Self::account_id_by_member_id(&id))
         } else {
             Err("member id doesn't exist")
-        }
-    }
-
-    pub fn get_profile(id: &T::AccountId) -> Option<Profile<T>> {
-        if let Some(member_id) = Self::ensure_is_member(id).ok() {
-            // This option _must_ be set
-            Self::member_profile(&member_id)
-        } else {
-            None
         }
     }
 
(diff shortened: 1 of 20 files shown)


bedeho:
Can you rebase this for review.

src/membership/members.rs
```diff
@@ -70,6 +98,8 @@ pub struct Profile<T: Trait> {
     pub entry: EntryMethod<T>,
     pub suspended: bool,
     pub subscription: Option<T::SubscriptionId>,
+    pub controller_account: T::AccountId,
```
bedeho: Can you add more comments here about what different fields are. This is particularly important in order to distinguish role of different accounts. Also the `roles` field deserves to be explained, given its complex structure.

src/membership/members.rs
```diff
@@ -70,6 +98,8 @@ pub struct Profile<T: Trait> {
...
+    pub roles: BTreeMap<RoleId, Vec<ActorId>>,
```
bedeho: Would it be more accurate to have `BTreeMap<RoleId, BTreeMap<ActordId,()> >` as the type here, to enforce that the same actor id cannot exist multiple times under the same role id? This presumes `BTreeMap` is iterable, which I suspect it is.

src/membership/members.rs
```diff
@@ -122,14 +150,16 @@ decl_storage! {
         /// MemberId to assign to next member that is added to the registry
         pub NextMemberId get(next_member_id) build(|config: &GenesisConfig<T>| config.first_member_id): T::MemberId = T::MemberId::from(DEFAULT_FIRST_MEMBER_ID);
 
-        /// Mapping of member ids to their corresponding primary accountid
+        /// Mapping of member ids to their corresponding primary account_id
         pub AccountIdByMemberId get(account_id_by_member_id) : map T::MemberId => T::AccountId;
 
-        /// Mapping of members' account ids to their member id.
+        /// Mapping of members' primary account ids to their member id.
```
bedeho: Can we also drop all the optional keys here in storage now? was just a bad workaround for the benefit of Pioneer. Probably better in separate PR though.

src/membership/members.rs
```diff
@@ -122,14 +150,16 @@ decl_storage! {
...
         pub MemberIdByAccountId get(member_id_by_account_id) : map T::AccountId => Option<T::MemberId>;
 
+        /// Mapping of members' controller account ids to their member id.
+        pub MemberIdByControllerAccountId get(member_id_by_controller_account_id) : map T::AccountId => Option<T::MemberId>;
```
bedeho: I don't think we need this, and we also should not _require_ that controller accounts are unique across members anyway.

src/membership/members.rs
```diff
@@ -165,53 +195,44 @@ decl_storage! {
...
-        Self::ensure_is_member(who)
-    }
-
-    fn lookup_account_by_member_id(id: Self::Id) -> Result<T::AccountId, &'static str> {
-        if <AccountIdByMemberId<T>>::exists(&id) {
-            Ok(Self::account_id_by_member_id(&id))
-        } else {
-            Err("member id doesn't exist")
-        }
+        MemberSetPrimaryKey(MemberId, AccountId),
```
bedeho: Key => Account

src/membership/members.rs
```diff
@@ -287,6 +308,38 @@ decl_module! {
...
         }
 
+        pub fn set_controller_key(origin, controller: T::AccountId) {
+            let master = ensure_signed(origin)?;
+            let member_id = Self::ensure_is_member_primary_account(&master)?;
+
+            let mut profile = Self::ensure_profile(member_id)?;
+
+            // ensure new key is not used by someone else (master or controller)
+            ensure!(!<MemberIdByControllerAccountId<T>>::exists(&controller), "account already paired with member");
```
bedeho: No need to check any kind of uniqueness.

src/membership/members.rs
```diff
@@ -287,6 +308,38 @@ decl_module! {
...
+            ensure!(!<MemberIdByAccountId<T>>::exists(&controller), "account already paired with member");
+
+            <MemberIdByControllerAccountId<T>>::remove(&profile.controller_account);
+            <MemberIdByControllerAccountId<T>>::insert(&controller, member_id);
+            profile.controller_account = controller.clone();
+            <MemberProfile<T>>::insert(member_id, profile);
```
bedeho: Mutate over insert.

src/membership/members.rs
```diff
@@ -287,6 +308,38 @@ decl_module! {
...
+            Self::deposit_event(RawEvent::MemberSetControllerKey(member_id, controller));
+        }
+
+        pub fn set_primary_key(origin, new_primary: T::AccountId) {
+            let old_primary = ensure_signed(origin)?;
+            let member_id = Self::ensure_is_member_primary_account(&old_primary)?;
+
+            // ensure new key not is used by someone else (master or controller)
+            ensure!(!<MemberIdByControllerAccountId<T>>::exists(&new_primary), "account already paired with member");
```
bedeho: Don't check against the controller, even if controllers were forced to be unique would this check not be justified.

src/membership/members.rs
```diff
@@ -122,14 +150,16 @@ decl_storage! {
         /// MemberId to assign to next member that is added to the registry
         pub NextMemberId get(next_member_id) build(|config: &GenesisConfig<T>| config.first_member_id): T::MemberId = T::MemberId::from(DEFAULT_FIRST_MEMBER_ID);
 
-        /// Mapping of member ids to their corresponding primary accountid
+        /// Mapping of member ids to their corresponding primary account_id
```
bedeho: Change the name of the storage field to be more direct, e.g. `PrimaryAccountByMemberId`.

Also, do we actually need two-way lookup MemberId <=> PrimaryAccount?

src/membership/members.rs
```diff
@@ -122,14 +150,16 @@ decl_storage! {
...
         pub AccountIdByMemberId get(account_id_by_member_id) : map T::MemberId => T::AccountId;
 
-        /// Mapping of members' account ids to their member id.
+        /// Mapping of members' primary account ids to their member id.
         pub MemberIdByAccountId get(member_id_by_account_id) : map T::AccountId => Option<T::MemberId>;
```
bedeho: Again, sharpen name of this field, and same question about the need for two-way lookup. If not required, then drop one of them.

src/membership/members.rs
```diff
@@ -485,4 +557,145 @@ impl<T: Trait> Module<T> {
         <MemberProfile<T>>::insert(id, profile);
         Ok(())
     }
+
+    // Checks if a member identified by their controller account occupies a role at least once
+    pub fn member_is_in_role(controller_account: &T::AccountId, role: Role) -> bool {
+        if let Ok(member_id) = Self::ensure_is_member_controller_account(controller_account) {
+            Self::ensure_profile(member_id).ok().map_or_else(
```
bedeho: This profile lookup cannot fail, but this code reads as it can. Write with an assertion that it's not possible.

src/membership/members.rs
```diff
@@ -485,4 +557,145 @@ impl<T: Trait> Module<T> {
...
+
+        let mut profile = Self::ensure_profile(member_id)?;
+        let mut new_ids = vec![actor_id];
+
+        if let Some(current_ids) = profile.roles.get_mut(&role.into()) {
+            current_ids.append(&mut new_ids);
+        } else {
+            profile.roles.insert(role.into(), new_ids);
+        }
+        <MemberProfile<T>>::insert(member_id, profile);
```
bedeho: Mutate over insert.

src/membership/members.rs
```diff
@@ -485,4 +557,145 @@ impl<T: Trait> Module<T> {
...
+    ) -> Result<(), &'static str> {
+        Self::can_unregister_role_on_member(controller_account, role, actor_id)?;
+
+        let member_id = Self::ensure_is_member_controller_account(controller_account)?;
+        let mut profile = Self::ensure_profile(member_id)?;
+
+        if let Some(current_ids) = profile.roles.get_mut(&role.into()) {
+            //current_ids.remove_item(&actor_id); // unstable nightly feature!
+            current_ids.retain(|id| *id != actor_id);
+            <MemberProfile<T>>::insert(member_id, profile);
```
bedeho: Mutate over insert.

src/membership/members.rs
```diff
@@ -485,4 +557,145 @@ impl<T: Trait> Module<T> {
...
+        }
+
+        <MembershipIdByActorInRole<T>>::remove(ActorInRole {
```
bedeho: Perhaps we do not remove? In particular, if we want to hold a history of prior roles held by a member? just mark as expired?

src/membership/members.rs
```diff
@@ -485,4 +557,145 @@ impl<T: Trait> Module<T> {
...
+        role: Role,
+        actor_id: ActorId,
+    ) -> Result<(), &'static str> {
+        Self::can_unregister_role_on_member(controller_account, role, actor_id)?;
+
+        let member_id = Self::ensure_is_member_controller_account(controller_account)?;
+        let mut profile = Self::ensure_profile(member_id)?;
+
+        if let Some(current_ids) = profile.roles.get_mut(&role.into()) {
+            //current_ids.remove_item(&actor_id); // unstable nightly feature!
```
bedeho: Perhaps add a length/size check to `current_ids` here for clarity, skip operations and writing to state if empty.

src/membership/members.rs
bedeho: The entire code path that follows hereafter the `if` check should not happen if the conditional is not satisfied, or even if it is but `current_ids.size() == 0`. Reorg to reflect this.

src/membership/members.rs
```diff
@@ -57,11 +60,36 @@ const DEFAULT_MAX_HANDLE_LENGTH: u32 = 40;
 const DEFAULT_MAX_AVATAR_URI_LENGTH: u32 = 1024;
 const DEFAULT_MAX_ABOUT_TEXT_LENGTH: u32 = 2048;
 
+pub type RoleId = u32;
```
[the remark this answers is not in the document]
mnaamani: Ideally we want the type `BTreeMap<enum Role, ActorId>` to represent the members roles.
Unfortunately there is no implementation of the codec Encode/Decode traits for BTreeMap when the key is a n `enum`. To avoid implementing it and having to subsequently implement corresponding encoders and decoders in polkadot-js api/types libraries I went with a simpler approach of using a `u32` as the key, but keeping the `enum Role` to have "named roles".

I did consider dropping BTreeMap and just go with a `Vec<ActorInRole>`.

Regarding the named roles, its possible to get rid of the `enum Role`. There are several approaches, that involves passing in a type on the trait that handles mapping.

src/membership/members.rs
mnaamani: Yes BTreeMap is Iterable.
I agree, effectively using it as a Set, because that's what we really want to represent and it gives us the type protection to enforce it. I'm just concerned with the complexity of this type and weather we can dervie Encode/Decode traits for it. Will have to check.

If its not possible we can always implement a `Set` backed by a `Vec`.

mnaamani:
> The coupling is paying off in terms of throwing out all that mock code <3.
> 
>     1. Great if you could add some comments to some of the structs added about what they represent.
Will do
> 
>     2. Can we replace the concept of "primary account" with "root account", the latter is more clear in its role.
> 
I tried not to change names of storage variables or methods in these modifications to limit scope of changes needed in pioneer, colossus and the stats server. But I agree is would be better.

>     3. I wonder now if using the root account as the identifier in the client modules trying to reference a membership was the right call. This was in order to avoid having client modules use a sort of actor identifier, but I think it may have been cleaner. Right now we are having to enforce uniqueness of the root key across members to generate the same result.
I agree its cleaner. It did occur to me as well. If we decide to go with this approach however then extrinsics sent in by actors will always have to include a MemberId parameter, similar to the Principal idea in the versioned store permissions module.

src/membership/members.rs
mnaamani: It will require changes in pioneer obviously.
Keeping it as separate PR is a good idea.

src/membership/members.rs
mnaamani: Obviously this is discussed more deeply in the PR comment, but its there to make lookup of a member by their controller account faster. 

If we don't want to enforce uniqueness of the key then it can be a `map T::AccountId => Vec<T::MemberId>`

src/membership/members.rs
mnaamani: Oh I wasn't aware `mutate()` was possible with maps. does the call have any affect if the value doesn't exist?
Will look into it.

src/membership/members.rs
mnaamani: That check is there to avoid needlessly updating the state if the new key is the same as the old. I will update the check to be clearer.

src/membership/members.rs
mnaamani: It is primarily needed to store the keyspace of the member accounts map.
Otherwise there wouldn't be a way to iterate over all members.

The Actors module uses it to lookup the member account id of a storage provider to send rewards to.

To get rid of it totally we need to switch to using the `enumerable` version of the StorageMap. I think its called  linked_map.

src/membership/members.rs
mnaamani: The prior check in `can_unregister_role_on_member()` will not allow code to reach this point in the first place.

mnaamani: so will include some invariant checks and reorg the code.

src/membership/members.rs
mnaamani: Since @siman pointed out that there is a `[BTreeSet](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html)` we can have:

`pub roles: BTreeMap<RoleId, BTreeSet<ActorId>>` 

Not sure about the `Codec` support in substrate and polkadot-js for such a type.

cc: @yourheropaul

src/membership/members.rs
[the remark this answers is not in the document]
bedeho: Now I can't recall what the premise is correct here? `BTreeMap<Role, ActorId>` would imply that a member can only ever occupy a single role of a given type at any time, which is a constraint we do not wish to capture?

src/membership/members.rs
bedeho: Agreed, the current solution `ActorInRoleSet<T::ActorId>` looks fine.

PR #83, opened 2019-09-26
build dummy wasm_binary.rs in IDE

mnaamani:
This updates the custom build script to only generate a dummy WASM binary in the IDE. It also bypasses the issue that in the IDE with RLS extension the CARGO env variable points to `rls` not `cargo`, and for the full build the wasm-builder-runner is trying use the `run` sub command for cargo which is not available in `rls`

Previously the IDE would continuously try and fail to run the build, and the inclusion of
`include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));` in `lib.rs` would fail, making it difficult to work on the codebase with RLS extension enabled without commenting out the line temporarily.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -14,30 +14,33 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
+use std::{env, process::Command, string::String};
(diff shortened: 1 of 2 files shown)


mnaamani:
implemented in https://github.com/Joystream/substrate-runtime-joystream/pull/81

PR #85, opened 2019-10-22
tracking polkadot-master branch of substrate

mnaamani:


bedeho (changes requested):
Very nice work!

Could we not depend on a branch, that just seems like an issue waiting to bite us at some point.

Annoying to have to write out actual commit multiple times though, they should have supported toml constants/variables, but chose not to apparently

https://github.com/rust-lang/cargo/issues/4221

PR #87, opened 2019-10-29
Rome version of content working group

bedeho:
# Specification

https://github.com/Joystream/joystream/issues/107

# Progress

- [x] Main types and storage variables defined.
- [x] Main extrinsics defined
- [x] Main events defined.
- [x] `create_channel`
- [x] `transfer_channel_ownership`
- [x] `add_curator_opening`
- [x] `accept_curator_applications`
- [x] `begin_curator_applicant_review`
- [x] `fill_curator_opening`
- [x] `terminate_curator_role`
- [x] `apply_on_curator_opening`
- [x] `set_lead`
- [x] `unset_lead`
- [x] `update_curator_creation_policies`
- [x] `leave_curator_role`
- [x] `terminate_curator_application`
- [x] `withdraw_curator_application`
- [x] `account_can_act_as_principal`
- [x] `update_curator_reward_account`
- [x] `update_curator_role_account`
- [x] `update_channel_curation_status`
- [x] `update_channel_as_owner`
- [x] `update_channel_as_curator`

## Future

- `update_curator_application_role_account`: switch out account using membership, in case key was lost?

## Dropped

These were all dropped when the principal model was simplified.

- `update_lead_credential`
- `update_any_member_credential`
- `update_any_curator_credential`
- `create_dynamic_credential`
- `update_dynamic_credential`

# Comments

- There is significant effort involved in always checking whether a given member can or cannot step into a given role at some point in time, and then later to actually move them in and out of that role. In some cases, there may be a double check, such as the initial check when someone applies for a position, and then a second one when they are hired by the lead. All of this takes effort to manage, and mostly it exists so that one can model very fine-grained rules about who can step into what role at what time. We should at a later time reconsider the value of this because currently, we don't even have any policy in mind, it's just a blanket yes for everything.

# Missing

 - https://github.com/Joystream/substrate-runtime-joystream/issues/96

- Constraining max number of applications and openings. Not only simultaneous, but also cumulative, because this is not cleaned up in hiring/staking: https://github.com/Joystream/joystream/issues/126#issuecomment-557792468.


mnaamani:
@bedeho the build is currently failing because we need to update the dependencies to use the same commit hash for substrate, as well as some minor tweaks to fix issues that only show up when building for WASM.

bedeho:
Great that you are on top of it, I have a temporary fix for the Vec issue here

https://github.com/Joystream/substrate-versioned-store-permissions-module/pull/3

so I am not blocked by that.

bedeho:
I know get this build error when doing `cargo test`

```
error[E0277]: `substrate_versioned_store::Property` doesn't implement `core::fmt::Debug`
   --> /Users/bedeho/.cargo/git/checkouts/substrate-versioned-store-permissions-module-9fdc387c352881a2/694a504/src/lib.rs:103:1
    |
103 | / decl_module! {
104 | |     pub struct Module<T: Trait> for enum Call where origin: T::Origin {
105 | |
106 | |         /// Sets the admins for a class
...   |
385 | |     }
386 | | }
    | |_^ `substrate_versioned_store::Property` cannot be formatted using `{:?}` because it doesn't implement `core::fmt::Debug`
    |
    = help: the trait `core::fmt::Debug` is not implemented for `substrate_versioned_store::Property`
    = note: required because of the requirements on the impl of `core::fmt::Debug` for `sr_api_hidden_includes_decl_storage::hidden_include::sr_primitives::sr_std::prelude::Vec<substrate_versioned_store::Property>`
    = note: required because of the requirements on the impl of `core::fmt::Debug` for `(core::option::Option<<T as Trait>::PrincipalId>, u64, sr_api_hidden_includes_decl_storage::hidden_include::sr_primitives::sr_std::prelude::Vec<u16>, sr_api_hidden_includes_decl_storage::hidden_include::sr_primitives::sr_std::prelude::Vec<substrate_versioned_store::Property>)`
    = note: required by `core::fmt::Debug::fmt`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

It is correct that `Property` does not have `Debug` trait, but its unclear to me why this issue is getting triggered only here, because building the version stored directly goes fine.

I guess the permissions module is getting used in some way which is new when its a dependency in the runtime.

I will try to fix and add those that trait, but I hope we can have a more safe and sound solution to resolve this issue of things breaking when they are used in ways that are not so visible when they are built and run on their own.

src/content_working_group/lib.rs
```diff
@@ -0,0 +1,2422 @@
...
+    }
+}
+
+// ======================================================================== //
+
+decl_storage! {
+    trait Store for Module<T: Trait> as ContentWorkingGroup {
+
+        /// The mint currently funding the rewards for this module.
+        pub Mint get(mint) config(): <T as minting::Trait>::MintId;
```
mnaamani: we don't have configurable mints at genesis.. so setting a mintid at genesis is problematic atm.
Either we can have this be an `Option<MintId>` with the mint getting created later with a sudo call,
or my preferred way to use a build() with some extra_genesis values to create a new mint.

src/content_working_group/lib.rs
```diff
@@ -0,0 +1,2422 @@
...
+                Self::ensure_channel_description_is_valid(description)?;
+            }
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            Self::update_channel(
+                &channel_id,
+                &None,
```
mnaamani: `new_channel_name` isn't used to update
```suggestion
                &new_channel_name,
```

src/content_working_group/lib.rs
```diff
@@ -0,0 +1,2422 @@
...
+
+            // Trigger event
+            Self::deposit_event(RawEvent::BeganCuratorApplicationReview(curator_opening_id));
+        }
+
+        /// Fill opening for curator
+        pub fn fill_curator_opening(
+            origin,
+            curator_opening_id: CuratorOpeningId<T>,
+            successful_curator_application_ids: BTreeSet<CuratorApplicationId<T>>
```
mnaamani: Does an empty set need to be handled in a special way?
Would have to look at behavior of hiring module when calling `fill_opening` with empty application set.

src/content_working_group/lib.rs
```diff
@@ -0,0 +1,2422 @@
...
+                None //new_payout_interval
+            )
+            .expect("Must be set, since curator has recurring reward");
+
+            // Trigger event
+            Self::deposit_event(RawEvent::CuratorRewardAccountUpdated(curator_id, new_reward_account));
+        }
+
+        /// An active curator leaves role
+
```
mnaamani: This can be called via sudo extrinsic. I don't think this is what you intended. Its probably the method that will be called from handler for StakingEventHandler::unstaked ?

src/content_working_group/lib.rs
mnaamani: It also just occurred to me, I don't see the mint id being used when creating reward relationships.

mnaamani: I see now the relationship is created through an extrinsic where the mintid is provided..

src/content_working_group/lib.rs
```diff
@@ -0,0 +1,2422 @@
...
+            Self::deposit_event(RawEvent::CuratorApplicationTerminated(curator_application_id));
+        }
+
+        /// Apply on a curator opening.
+        pub fn apply_on_curator_opening(
+            origin,
+            member_id: T::MemberId,
+            curator_opening_id: CuratorOpeningId<T>,
+            role_account: T::AccountId,
+            source_account: T::AccountId,
```
mnaamani: There are no guarantees that the origin controls the funds in the source_account provided, which means they can stake funds from arbitrary accounts.

In the storage role staking module we split this staking into a two step process but I found that a little too complex both on the runtime and ui side.

So maybe we can opt for something simpler here, for example that the source account must be the member's root_account or controller_account?

src/content_working_group/lib.rs
mnaamani: No values are ever inserted into `UnstakerByStakeId` map. Is that step missing in `apply_on_curator_opening()`

mnaamani: Correction.. (as per chat) step is missing in `deactivate_curator()`

bedeho: Yes this is a bug, we must add to this map whenever there is async unstaking initaited, that is staking with an unboding period.

src/content_working_group/lib.rs
mnaamani: Here is one approach to create mint with build() at genesis: 438c031bfd429cc1f01e77ba8bf35c6c319dc138

src/content_working_group/lib.rs
mnaamani: Here is a commit from #103 which makes unstaked a normal method

mnaamani:
I think we need to add some extrinsics to give sudo ability to update the mint capacity, and or adjustment policy.

src/content_working_group/lib.rs
bedeho: Well spotted, quite a severe oversight, I don't love the remedy, but may be best to do for Rome, and fix properly later.

mnaamani:
merging, but there are pending issues to be fixed. see #117

PR #89, opened 2019-11-07
Role checking fixed

bedeho:


mnaamani:
Looks good, it just needs a round of `cargo-fmt`, but I can't confirm tests are passing because build is failing. The compiler error is in libp2p-core-0.12.0 which is a dependency of substrate. This is out of our hands but should be fixed upstream eventually if v0.12 is still being maintained.

Hmm looks like v0.13 was [published](https://crates.io/crates/libp2p-core) which means we will probably have to bump our dependency on substrate to a newer commit.

```bash
error[E0004]: non-exhaustive patterns: `Err(UnsupportedInputLength)` not covered
   --> /Users/mokhtar/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-core-0.12.0/src/peer_id.rs:132:15
    |
132 |         match multihash::encode(alg, &enc) {
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `Err(UnsupportedInputLength)` not covered
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
```

mnaamani:
As discussed @bedeho will check-in a working Cargo.lock file to resolve this build issue.

PR #103, opened 2019-12-09
Rome

mnaamani:


src/content_working_group/lib.rs
```diff
@@ -1896,7 +1893,52 @@ decl_module! {
             // Ensure root is origin
             ensure_root(origin)?;
 
-            Self::do_set_lead(member, role_account)?;
+            // Ensure there is no current lead
```
bedeho: Why was this change made? I may be a good idea, but not clear what was the context for this.

src/lib.rs
```diff
@@ -441,6 +444,73 @@ impl versioned_store_permissions::CredentialChecker<Runtime> for SudoKeyHasAllCr
...
+
+parameter_types! {
+    pub const CurrentLeadCredential: Credential = 0;
+    pub const AnyActiveCuratorCredential: Credential = 1;
+    pub const AnyActiveChannelOwnerCredential: Credential = 2;
+    pub const PrincipalIdMappingStartsAtCredential: Credential = 1000;
+}
+
+pub struct ContentWorkingGroupCredentials {}
+impl versioned_store_permissions::CredentialChecker<Runtime> for ContentWorkingGroupCredentials {
```
bedeho: Why is this implemented here? There already is this here

https://github.com/mnaamani/substrate-runtime-joystream/blob/integrate-content-wg/src/content_working_group/lib.rs#L2017

I do not recall how to think about this mapping process, but hte fact that there is iteration in this implementaiton makes me think this is almost certainly wrong.

bedeho (changes requested):
This was very difficult to review, I think we need a better process for this.
Spotted one important issue, but largely by accident.

src/content_working_group/lib.rs
mnaamani: I originally had do_set_lead() because it needed to be called from genesis construction and from the extrinsic dispatch. When we decided against setting a lead at genesis I removed it.

src/lib.rs
mnaamani: That implementation (in the working group itself) only does one to one mapping. I needed to support a range of Credentials that represented a group, so I made it as a separate impl, which re-uuses the impl you linked for range of credentials that represent 1:1 mapping to a principal.

mnaamani: Yes the need for iteration is because there is a single map of principalId -> principal, so it was tge only way to identify different principal types (without adding more state)

PR #104, opened 2019-12-13
Default impl for PaidMembershipTerms

kuronyago:
for https://github.com/Joystream/substrate-runtime-joystream/issues/101
- remove impl `Default` trait
- removed unused constants

### DIFF
--- a/src/membership/members.rs
+++ b/src/membership/members.rs
@@ -140,17 +140,6 @@ pub struct PaidMembershipTerms<T: Trait> {
     pub text: Vec<u8>,
 }
 
-/// Must be default constructible because it indirectly is a value in a storage map.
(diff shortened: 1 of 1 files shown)


mnaamani:
A further improvement is to actually get rid of the default impl altogether. The storage map `PaidMembershipTermsById` uses `Option<PaidMembershipTerms>` as the value.
And we are explicitly calling `PaidMembershipTerms::default()` in the build() method of that map:

https://github.com/Joystream/substrate-runtime-joystream/blob/6d1867e41dc2df00482dafe3091b5f55ee36e2bb/src/membership/members.rs#L183

It can be replaced with:
```
let terms = PaidMembershipTerms {
       fee: config.default_paid_membership_fee,
       text: Vec::default(),
};
```

mnaamani:
LGTM

Thank you for the contribution @kuronyago

blrhc:
@kuronyago Thank you very much! If you have an XMR address, we'd love to send you a small bounty as a token of appreciation for contributing to our project.

Issue #117, opened 2020-01-08
Pending Fixes For Content Working Group


mnaamani:
Some pending issues in the working group implementation.

- Better initial configuration of a mint for the working group - fix:  #119

- ~`update_channel_as_owner` doesn't use parameter `new_channel_name` when calling `update_channel()` - fix #118
original [comment](https://github.com/Joystream/substrate-runtime-joystream/pull/87#discussion_r353064685)~

-  `apply_on_curator_opening()` - Origin doesn't prove that it controls source account from which funds will be staked. - fix #121 
original [comment](https://github.com/Joystream/substrate-runtime-joystream/pull/87#discussion_r354708914)

- `unstaked` is implemented as a privileged method (can be invoked by sudo key via extrinsic) which isn't correct. It should just be a normal method - fix #120
original [comment](https://github.com/Joystream/substrate-runtime-joystream/pull/87#discussion_r353077595)

- `fill_curator_opening` doesn't setup a reward relationship  - fix #123

mnaamani:
some of these issues are fixed in #103 but I will prepare separate PRs to address each for better review process.

PR #119, opened 2020-01-09
content working group: create mint and optionaly lead at genesis

mnaamani:
Addresses - https://github.com/Joystream/substrate-runtime-joystream/pull/87#discussion_r353063741

from list of issues: #117


bedeho (commented):
1. Looking at the original issue you pointed out here https://github.com/Joystream/substrate-runtime-joystream/pull/87#discussion_r353063741, it's not clear to me what exactly the problem is. The fact that the mint module cannot be configured with a mint at genesis currently is not really an obstacle to this id being configurable here, because we have often just used configurability as a way to setup tests, not per say to setup genesis states. Am I missing something here?
2. Even if we could not have it configurable for this reason, then I think it would be simpler to just make it optional, as you suggest, or drop it entirely. Right now, its not even used in the content working group as far as I can tell?

Do we need to set the lead at genesis, if not, then I would prefer to remove all of that `add_extra_genesis` code also.

mnaamani:
Regarding setting lead at genesis, no there is no necessity but my fix was really in response to the code comment that it was "Not configurable, because default set value breaks semantics". So we can drop it.

mnaamani:
Regarding the mint configuration will respond in and related issue [here](https://github.com/Joystream/substrate-runtime-joystream/pull/123#pullrequestreview-356153345)

bedeho (approved):

PR #123, opened 2020-01-13
Create recurring reward relationship for curator when filling role

mnaamani:


bedeho (changes requested):
- Now that I am seeing the usage of the mint, perhaps we want to have the flexibility to have different mints over time, or not set it from the get go, or not have to be forced to somehow ensure that its set before a call to fill an opening. In which case an optional mint_id could make sense. What do you think?

- Better to unpack the construction of `reward_relationship` into multiple steps, its a bit to clever right now.

mnaamani:
> Now that I am seeing the usage of the mint, perhaps we want to have the flexibility to have different mints over time, or not set it from the get go, or not have to be forced to somehow ensure that its set before a call to fill an opening. In which case an optional mint_id could make sense. What do you think?

I agree having an optional mint is more flexible. If we choose that approach we will need to add a few more dispatchable calls to support managing mints in general (remember the mint module has not dispatchable calls).  We still need some dispatchables in any case just to manage the mint created at genesis. 

However I would prefer having a simpler setup of one mint generated at genesis that is well managed by the Lead. At least for Rome this will be sufficient.

I will fix conflict and update implementation of constructing reward relationship.

bedeho (approved):
Ok, agreeing with your assessment, will merge.

PR #126, opened 2020-01-17
Allow optional channel fields to have an empty vector as a value

siman:
If such optional channel fields as title, description, avatar or banner have an empty `Vec<u8>` as a value, then it means that value is not set for this field.


siman:
Closing this PR. @bedeho @mnaamani and I decided to wrap optional fields into `Option`s to make it clear that the fields are optional. An empty `Vec` is quite ambiguous.

PR #164, opened 2020-02-26
prevent multiple active applications by a member to an opening

### LINKED ISSUE #163: Accepting multiple applications for same member causes assertion failure
**Problem**

The same member can apply more than once to an opening in the curator working group.
An assertion failure is occurring when the lead accepts more than one application for the same member because of a failed attempt to register the same role more than once on the member.

https://github.com/Joystream/substrate-runtime-joystream/blob/c63a02d922c46d11387b5e26df5dab00ec8d864a/src/content_working_group/lib.rs#L1594

This edge case was not accounted for.

**Solutions**
1. Prevent same member from applying more than once - if applican't wishes to perhaps increase their stake during the application process they must withdraw their application first then re-apply.

2. In fill_curator_opening() guard against successful_curator_application_ids set having applications from same member id.

3. Ensure there is always only one application per member per opening: When same member applies a second time to an opening, the new application replaces the old one.

I like Option 3 but this will require some changes in the UI/Hiring-Flow in pioneer. So I think I will prepare a fix with option 1.

mnaamani:
Fixes #163

### DIFF
--- a/src/content_working_group/lib.rs
+++ b/src/content_working_group/lib.rs
@@ -225,8 +225,10 @@ pub static MSG_APPLY_ON_CURATOR_OPENING_UNSIGNED_ORIGIN: &str = "Unsigned origin
 pub static MSG_APPLY_ON_CURATOR_OPENING_MEMBER_ID_INVALID: &str = "Member id is invalid";
 pub static MSG_APPLY_ON_CURATOR_OPENING_SIGNER_NOT_CONTROLLER_ACCOUNT: &str =
     "Signer does not match controller account";
-static MSG_ORIGIN_IS_NIETHER_MEMBER_CONTROLLER_OR_ROOT: &str =
+pub static MSG_ORIGIN_IS_NIETHER_MEMBER_CONTROLLER_OR_ROOT: &str =
     "Origin must be controller or root account of member";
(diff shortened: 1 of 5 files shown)


src/content_working_group/lib.rs
```diff
@@ -2055,6 +2060,29 @@ impl<T: Trait> versioned_store_permissions::CredentialChecker<T> for Module<T> {
 }
 
 impl<T: Trait> Module<T> {
+    fn ensure_member_has_no_active_application_on_opening(
+        opening_id: CuratorOpeningId<T>,
+        member_id: T::MemberId,
+    ) -> Result<(), &'static str> {
+        for (_id, curator_application) in <CuratorApplicationById<T>>::enumerate() {
```
bedeho: This iterates all applications on all openings, which is not great, or entirely safe.

At the very least there is `CuratorOpening::curator_applications` of type `BTreeSet<CuratorApplicationId>` which allows you to iterate only applications to the given curator opening. This is still not _great_, but we will do a refactor later where membership binding is done inside subsystem modules, not in membership module, which would make the lookup direct, so this may be the best we can do for now, short of introducing a temporary reverse index.

PR #171, opened 2020-02-27
remove metadata concept from storage module

mnaamani:
Intent is to require only a single extrinsic (by the member uploading content) to add a data object to the storage.

Changes:
- Add new content id to know content ids vector when storage node accepts the upload, instead of waiting for member to send a second transaction add_metadata.

- Cleanup: remove all "metadata" related state and methods as this is all now in the versioned store.

### DIFF
--- a/src/storage/data_directory.rs
+++ b/src/storage/data_directory.rs
@@ -79,24 +79,6 @@ impl Default for ContentVisibility {
     }
 }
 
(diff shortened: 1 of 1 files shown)


bedeho (approved):

siman (commented):
Looks good to me!

PR #173, opened 2020-02-28
move all credential checking logic inside working group module

mnaamani:
Cleanup Credential checking and keep it inside the content working group module.
Same mapping of Credential to PrincipalId for the "reserved" credentials.
No more shifting of credentials to map to principal id, there is now 1:1 mapping of the Credential in the permission set to the principal id in the working group module.

This change doesn't affect rome inital setup as we are only using the "reserved" credentials in the initial permissions sets.


mnaamani:
Do this refactoring as part of next testnet release

PR #178, opened 2020-03-06
Fix transfer_channel_ownership

mnaamani:
Content working group transfer_channel_ownership tries to:
Unregister and then to Re-register same ActorInRole (ChannelOwnerRole, ChannelId) to another member, which fails because membership module doesn't remove the original mapping of role -> member_id.
We do this to preserve history.

So for the particular case of transfering I added a new transfer_role() method.

Updated transfer_channel_ownership to use that method instead, and added test case.

Questions:
 - In case member looses access to their keys, channel transfer is not possible. Should we add a method for root origin to be able to transfer?


shamil-gadelshin:
This branch produces multiple build warnings.

src/membership/members.rs
```diff
@@ -709,10 +709,46 @@ impl<T: Trait> Module<T> {
...
+        Ok(())
+    }
+
+    pub fn transfer_role(
+        actor_in_role: ActorInRole<T::ActorId>,
+        new_member_id: T::MemberId,
+    ) -> Result<(), &'static str> {
+        Self::can_transfer_role_to_member(actor_in_role, new_member_id)?;
+        Self::unregister_role(actor_in_role)?;
+        let _ = <MembershipIdByActorInRole<T>>::remove(&actor_in_role);
```
shamil-gadelshin: `let _ = ...  `is redundant

shamil-gadelshin:
I suggest adding some comments on all public methods, but it's not in our guidelines so - its to you.

src/content_working_group/lib.rs
```diff
@@ -1210,13 +1210,21 @@ decl_module! {
             // Ensure channel owner has signed
             let channel = Self::ensure_channel_owner_signed(origin, &channel_id)?;
 
-            // Ensure prospective new owner can actually become a channel owner
-            let (new_owner_as_channel_owner, _next_channel_id) = Self::ensure_can_register_channel_owner_role_on_member(&new_owner, Some(channel_id))?;
+            // Ensure prospective new owner can actually become a channel owner (with a new channel id)
+            // We do not pass the existing channel id because it is already owned and the call would
+            // return with Err, since the membership system doesn't allow the same ActorInRole to be assigned
+            // to more than one member.
+            let _ = Self::ensure_can_register_channel_owner_role_on_member(&new_owner, None)?;
```
shamil-gadelshin: `let _ = ...` is redundant

src/membership/members.rs
```diff
@@ -725,14 +733,7 @@ impl<T: Trait> Module<T> {
...
     ) -> Result<(), &'static str> {
-        // Ensure member exists
-        let profile = Self::ensure_profile(member_id)?;
-
-        // ensure is active member
-        ensure!(!profile.suspended, "SuspendedMemberCannotEnterRole");
-
-        // TODO: if above policy checks are identical to ones in can_register_role_on_member() factor them out
-        // to a separate function
+        let _profile = Self::register_role_on_member_policy_check(&member_id, &actor_in_role)?;
```
shamil-gadelshin: `let _profile = ...` is redundant

mnaamani:
> This branch produces multiple build warnings.

Yes it does, i fixed a couple of ones.

The majority that are left are harmless, unused methods in the fixtures of the working group tests.
They should be fixed so that they don't hide from us more important warning, but will leave it to a separate PR.

There are two which should probably be addressed soon.
This one will become a compiler error eventually: #179 
This one is lower priority and the runtime is builtin fine for now: #180

shamil-gadelshin (approved):
Looks good.

bedeho (changes requested):
1) role->member_id was not meant to by for history, at least not in my mind.

2) even if it was, we should def. stop using state storage for archival purposes, its has costs, any little upside when the query layer arrives. If viewing history was a valuable feature now before we had such a query layer, that would be one thing, but it's not, we do not surface this in Pioneer.

3) the role of the membership module as regulating any of this is going to go away, at least this is what I had in mind. It's going away in two senses a) role bindings should be moved out b) there should be no global enforcement of who can step into what role as a member. See here: https://github.com/Joystream/substrate-runtime-joystream/issues/115

Hence I advocate some much simpler fix to sidestep this.

mnaamani:
I went for the simplest fix as advised, by removing the mapping when unregistering the role.

bedeho (approved):

PR #182, opened 2020-03-06
Monorepo. 7 modules with 1 common module.

shamil-gadelshin:
Converted modules from inner folders to inner crates:
- common (contains widely used GovernanceCurrency)
- membership
- roles
- storage
- service-discovery
- memo
- content-working-group
- governance

Comments:
- `content-working-group` module wasn't tested because of the forum `module` dependency issue
- the `runtime` itself builds successfully, but the `node` cannot be built because of the same issue with `forum` module


runtime-modules/content-working-group/Cargo.toml
```diff
@@ -0,0 +1,114 @@
...
+package = 'substrate-primitives'
+rev = 'c37bb08535c49a12320af7facfd555ce05cce2e8'
+
+[dependencies.codec]
+default-features = false
+features = ['derive']
+package = 'parity-scale-codec'
+version = '1.0.0'
+
+[dependencies.forum]
```
mnaamani: I think dependency on forum is only for use of the validation method, `InputValidationLengthConstraint` that should probably be moved to the common crate later.

shamil-gadelshin:
`cargo fmt` and `cargo test` commands work recursively when applied on the workspace Cargo.toml.

I tested it on my proposals module workspace with three modules.

runtime-modules/content-working-group/Cargo.toml
shamil-gadelshin: https://github.com/Joystream/substrate-runtime-joystream/issues/184

mnaamani:
So we need to turn our root Cargo.toml setup to a `[workspace]` because at the moment it is just a library package, which I think comes next in your plan.

mnaamani:
> Confirmed Runtime and Node builds successfully.
> Lets merge after #186 and #185

Oh I forgot this is towards monorepo branch. Will merge.

PR #195, opened 2020-03-18
Staking: add slash_immediate function

### LINKED ISSUE #161: Fix stake module and allow slashing and unstaking in the same block.
Stake module public API includes:
- `initiate_slashing` - with **slash_period** parameter
- `initiate_unstaking` - with **unstaking_period** parameter

Currently, it is possible to call `initiate_unstaking` with zero unstaking_period to perform instant unstaking. But `initiate_slashing` demands greater than zero slash_period. Combined, these two functions cause an error when used together on instant unstaking with some slashes.

mnaamani:
- [x] Add functionality to staking module to do immediate slashing
- [x] Unit tests

Fixes #161

### DIFF
--- a/runtime-modules/stake/src/lib.rs
+++ b/runtime-modules/stake/src/lib.rs
@@ -245,14 +245,14 @@ where
     /// Returns the actual slashed amount.
     fn apply_slash(
         &mut self,
-        slash: Slash<BlockNumber, Balance>,
+        slash_amount: Balance,
         minimum_balance: Balance,
     ) -> Balance {
         // calculate how much to slash
-        let mut slash_amount = if slash.slash_amount > self.staked_amount {
+        let mut slash_amount = if slash_amount > self.staked_amount {
             self.staked_amount
         } else {
-            slash.slash_amount
+            slash_amount
         };
 
         // apply the slashing
@@ -274,7 +274,7 @@ where
 
         for (slash_id, slash) in self.get_slashes_to_finalize().iter() {
             // apply the slashing and get back actual amount slashed
-            let slashed_amount = self.apply_slash(*slash, minimum_balance);
+            let slashed_amount = self.apply_slash(slash.slash_amount, minimum_balance);
 
             finalized_slashes.push((*slash_id, slashed_amount, self.staked_amount));
         }
@@ -423,13 +423,7 @@ where
                 // irrespective of wether we are unstaking or not, slash!
 
                 let actually_slashed = staked_state.apply_slash(
-                    Slash {
-                        slash_amount,
-                        // below values are irrelevant
-                        is_active: true,
-                        blocks_remaining_in_active_period_for_slashing: BlockNumber::zero(),
-                        started_at_block: BlockNumber::zero(),
-                    },
+                    slash_amount,
                     minimum_balance,
                 );
 
--- a/runtime-modules/stake/src/lib.rs
+++ b/runtime-modules/stake/src/lib.rs
@@ -243,11 +243,7 @@ where
 
     /// Executes a Slash. If remaining at stake drops below the minimum_balance, it will slash the entire staked amount.
     /// Returns the actual slashed amount.
-    fn apply_slash(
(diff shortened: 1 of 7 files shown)


runtime-modules/stake/src/macroes.rs
```diff
@@ -17,3 +17,15 @@ macro_rules! ensure_stake_exists {
...
     }};
 }
+
+#[macro_export]
+macro_rules! ensure_staked_amount {
+    ($stake:expr, $error:expr) => {{
+        match $stake.staking_status {
+            StakingStatus::Staked(ref staked_state) => staked_state.staked_amount,
+            _ => {
+                return Err(StakeActionError::Error(ImmediateSlashingError::NotStaked));
```
mnaamani: oops I'm not using the $error:expr here

shamil-gadelshin:
Do we need to call T::StakingEventsHandler::slashed() somewhere?

runtime-modules/stake/src/macroes.rs
shamil-gadelshin: Yes, you are not 😃

runtime-modules/stake/src/lib.rs
```diff
@@ -408,6 +408,38 @@ where
...
+            StakingStatus::Staked(ref mut staked_state) => {
+                // irrespective of wether we are unstaking or not, slash!
+
+                let actually_slashed = staked_state.apply_slash(
+                    Slash {
+                        slash_amount,
+                        // below values are irrelevant
+                        is_active: true,
+                        blocks_remaining_in_active_period_for_slashing: BlockNumber::zero(),
+                        started_at_block: BlockNumber::zero(),
```
shamil-gadelshin: started_at_block: now() looks better for immediate slash in my opinion. Even if these values are irrelevant.

mnaamani:
> Do we need to call T::StakingEventsHandler::slashed() somewhere?
> 
> `fn finalize_slashing_and_unstaking()` calls it after the
> 
> `let imbalance = Self::withdraw_funds_from_stake_pool(slashed_amount);`
> 
> Or it's not necessary because you return the SlashImmediateOutcome?

Thats a good point actually, I think we should indeed. I was looking for the event handler for slashing. I mistakenly thought we didn't have one.. been a long time since I looked at the code.

So I'll update to call the handler, passing the imbalance and change the value in SlashImmediateOutcome to be a Balance instead of an imbalance. What do you think?

runtime-modules/stake/src/lib.rs
mnaamani: Good observation here.
I think the better solution is to modify the signature of  `fn apply_slash`
from:
```rust
fn apply_slash(
        &mut self,
        slash: Slash<BlockNumber, Balance>,
        minimum_balance: Balance,
    ) -> Balance
```
to:
```
fn apply_slash(
        &mut self,
        slash_amount: Balance,
        minimum_balance: Balance,
    ) -> Balance
```
since it only really needs the `slash_amount` property of `Slash` struct, and we don't need to pass any of those args..

runtime-modules/stake/src/macroes.rs
mnaamani: will fix :)

shamil-gadelshin:
> So I'll update to call the handler, passing the imbalance and change the value in SlashImmediateOutcome to be a Balance instead of an imbalance. What do you think?

I believe that a consistent approach with the rest of the staking module is better. So, I vote for slashing handler 😄

runtime-modules/stake/src/lib.rs
shamil-gadelshin: This solution is even better.

mnaamani:
There is problem however, to call the `slashed()` handler, it expects a `SlashId` associated with a slashing that was initiated with initiate_slashing. So I would have to update the signature to allow an `Option<SlashId>` instead.

shamil-gadelshin (approved):
Looks good.

PR #239, opened 2020-03-25
Added set_mint_capacity dispatchable and tests

### LINKED ISSUE #237: Add set_mint_capacity root dispatchable to content wg
We need a root dispatchable method to be executed by the `Set Content Working Group Mint Capacity` Proposal #234 to set the capacity of the working group.

This gives the council ability to grant the working group Lead a budget from which to pay curators their recurring rewards. Its most likely that the lead would initiaite this proposal based on their hiring plan and budget for curator payouts.

The council can also take action to reduce the capacity possibly to zero if the content working group is not performing as expected or believes the lead is over spending.

mnaamani:
Fixes #237

### DIFF
--- a/runtime-modules/content-working-group/src/lib.rs
+++ b/runtime-modules/content-working-group/src/lib.rs
@@ -1080,7 +1080,9 @@ decl_event! {
         CuratorApplicationId = CuratorApplicationId<T>,
         CuratorId = CuratorId<T>,
         CuratorApplicationIdToCuratorIdMap = CuratorApplicationIdToCuratorIdMap<T>,
+        MintBalanceOf = minting::BalanceOf<T>,
         <T as system::Trait>::AccountId,
+        <T as minting::Trait>::MintId,
     {
         ChannelCreated(ChannelId),
         ChannelOwnershipTransferred(ChannelId),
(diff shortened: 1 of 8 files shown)


runtime-modules/content-working-group/src/lib.rs
```diff
@@ -2033,6 +2041,41 @@ decl_module! {
             let mint = <minting::Module<T>>::mints(mint_id); // must exist
             let new_capacity = mint.capacity() + additional_capacity;
             let _ = <minting::Module<T>>::set_mint_capacity(mint_id, new_capacity);
```
shamil-gadelshin: I recommend here to add "?":
```
<minting::Module<T>>::set_mint_capacity(mint_id, new_capacity)?;
```
Since set_mint_capacity() returns the `Result` and it's possible that another error would be added later inside it - it's safer to process this error rather than ignore it.

runtime-modules/content-working-group/src/lib.rs
```diff
@@ -2033,6 +2041,41 @@ decl_module! {
...
+            let mint_id = Self::mint();
+
+            // Mint must exist - it is set at genesis
+            let mint = <minting::Module<T>>::mints(mint_id);
+
+            let current_capacity = mint.capacity();
+
+            if new_capacity != current_capacity {
+                // Cannot fail if mint exists
+                let _ = <minting::Module<T>>::set_mint_capacity(mint_id, new_capacity);
```
shamil-gadelshin: Same as above

runtime-modules/content-working-group/src/lib.rs
mnaamani: That is a good idea.

mnaamani: done in 1d115a84a74548faa113f678570de987e90417b4

runtime-modules/content-working-group/src/lib.rs
mnaamani: done in 1d115a84a74548faa113f678570de987e90417b4

shamil-gadelshin (approved):
LGTM

Issue #243, opened 2020-03-26
Add a `replace_lead()` method to content working group


mnaamani:

mnaamani:
completed in #252

PR #252, opened 2020-03-26
Add `replace_lead()` to content working group

mnaamani:
Implements #243

### DIFF
--- a/runtime-modules/content-working-group/src/lib.rs
+++ b/runtime-modules/content-working-group/src/lib.rs
@@ -1906,103 +1906,23 @@ decl_module! {
             );
         }
 
-        /*
-         * Root origin routines for managing lead.
-         */
-
-
-        /// Introduce a lead when one is not currently set.
-        pub fn set_lead(origin, member: T::MemberId, role_account: T::AccountId) {
-
+        /// Replace the current lead. First unsets the active lead if there is one.
+        /// If a value is provided for new_lead it will then set that new lead.
+        /// It is responsibility of the caller to ensure the new lead can be set
+        /// to avoid the lead role being vacant at the end of the call.
+        pub fn replace_lead(origin, new_lead: Option<(T::MemberId, T::AccountId)>) {
             // Ensure root is origin
             ensure_root(origin)?;
 
-            // Ensure there is no current lead
-            ensure!(
-                <CurrentLeadId<T>>::get().is_none(),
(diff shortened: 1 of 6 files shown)


runtime-modules/content-working-group/src/lib.rs
```diff
@@ -2122,6 +2042,98 @@ impl<T: Trait> versioned_store_permissions::CredentialChecker<T> for Module<T> {
...
+
+        // Store lead
+        <LeadById<T>>::insert(new_lead_id, new_lead);
+
+        // Update current lead
+        <CurrentLeadId<T>>::put(new_lead_id); // Some(new_lead_id)
+
+        // Update next lead counter
+        <NextLeadId<T>>::mutate(|id| *id += <LeadId<T> as One>::one());
+
```
shamil-gadelshin: Probably, assert() can be replaced with "?" function invocation. Code reordering is required.

runtime-modules/content-working-group/src/lib.rs
```diff
@@ -2122,6 +2042,98 @@ impl<T: Trait> versioned_store_permissions::CredentialChecker<T> for Module<T> {
...
+        // Unregister from role in membership model
+        let current_lead_role = role_types::ActorInRole {
+            role: role_types::Role::CuratorLead,
+            actor_id: lead_id,
+        };
+
+        let unregistered_role = <members::Module<T>>::unregister_role(current_lead_role).is_ok();
+
+        assert!(unregistered_role);
```
shamil-gadelshin: Probably, assert() can be replaced with "?" function invocation. Code reordering is required.

runtime-modules/content-working-group/src/lib.rs
mnaamani: Thats a good point, I think this is the new convention we decided on. However here I avoided changing the implementation of set_lead and unset_lead I just relocated them and converted them from being dispatchables to be private methods on the `Module`.

I can update it not much extra work.

mnaamani: done in e4f57a28617fb9b66734164f9707ad5dd110e4be

runtime-modules/content-working-group/src/lib.rs
mnaamani: done in e4f57a28617fb9b66734164f9707ad5dd110e4be

shamil-gadelshin (approved):
LGTM

Issue #244, opened 2020-03-26
Add `spend_from_council_mint()` method to council module


mnaamani:

mnaamani:
Implemented in #253

Issue #242, opened 2020-03-26
Add a mint and `set_mint_capacity()` to council module


mnaamani:
ref: https://github.com/Joystream/substrate-runtime-joystream/issues/158
ref: https://github.com/Joystream/substrate-runtime-joystream/issues/234

mnaamani:
Implemented in #253

PR #253, opened 2020-03-27
Add a council mint

mnaamani:
Implements #242 and #244

- Adding a mint to the council, ~~initialized at genesis~~
- Migration step at runtime upgrade to create the council mint
- root dispatchable methods for proposal
  - spend_from_council_mint() - fails if no mint exists
  - set_council_mint_capacity() - creates mint if it doesn't exist

### DIFF
--- a/runtime-modules/governance/src/council.rs
+++ b/runtime-modules/governance/src/council.rs
@@ -67,7 +67,7 @@ impl<T: Trait> Module<T> {
         Self::active_council().iter().any(|c| c.member == *sender)
     }
 
-    // Initializes a new mint, discarding previous mint if it existed.
+    /// Initializes a new mint, discarding previous mint if it existed.
     pub fn create_new_council_mint(
         capacity: minting::BalanceOf<T>,
     ) -> Result<T::MintId, &'static str> {
--- a/runtime-modules/governance/src/council.rs
+++ b/runtime-modules/governance/src/council.rs
@@ -33,10 +33,10 @@ decl_storage! {
 
         pub TermEndsAt get(term_ends_at) config() : T::BlockNumber = T::BlockNumber::from(1);
 
-        /// The mint that funds council member rewards and spending proposals budget.
-        pub CouncilMint get(council_mint) build(|_config: &GenesisConfig<T>| {
-            <Module<T>>::create_new_council_mint()
-        }): <T as minting::Trait>::MintId;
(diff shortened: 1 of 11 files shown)


runtime-modules/governance/src/council.rs
```diff
@@ -60,6 +66,14 @@ impl<T: Trait> Module<T> {
     pub fn is_councilor(sender: &T::AccountId) -> bool {
         Self::active_council().iter().any(|c| c.member == *sender)
     }
+
+    // Initializes a new mint
+    pub fn create_new_council_mint() -> T::MintId {
+        let mint_id = <minting::Module<T>>::add_mint(minting::BalanceOf::<T>::zero(), None)
+            .expect("Failed to create a mint for the council");
```
shamil-gadelshin: I'm not happy with panic in the runtime. Can we put mint only on successful creation and during the setting mint parameters - introduce another "ensure_*"? check on whether mint actually exist?

mnaamani: That is a fair point. I originally had this `expect()` inside the `build()` method before refactoring, where it was okay to panic while constructing genesis. 

Given that add_mint can fail and we call it in the migration, I will have to make the mint an `Option<MintId>`

Let me revise.

mnaamani: Updated to make the mint an `Option` so it doesn't need to be created at genesis and we can avoid panic. The mint can be automatically created when setting capacity if it doesn't exist.

runtime/src/migration.rs
```diff
@@ -1,28 +1,24 @@
...
 
         // ...
-        // add initialization of other modules introduced in this runtime
+        // add initialization of modules introduced in new runtime release. This
+        // would be any new storage values that need an initial value which would not
+        // have been initialized with config() or build() mechanism.
         // ...
 
+        // Create the Council mint. If it fails, we can't do anything about it here.
+        let _ = governance::council::Module::<T>::create_new_council_mint(
```
shamil-gadelshin: We agreed to print an error message in such cases. Don't we?

shamil-gadelshin:
Single error message comment. Will be fixed later. 

LGTM

runtime/src/migration.rs
mnaamani: Done in https://github.com/Joystream/substrate-runtime-joystream/pull/275/commits/a6148b9a91b7c179ec0eb729ff1584e36b59bddb

PR #258, opened 2020-03-28
Docs for set_election_parameters

mnaamani:
Improve docs for set_election_parameters and use it when building genesis to ensure valid parameters at start of a chain.


shamil-gadelshin (approved):
LGTM

PR #175, opened 2020-03-02
Proposals v2. Iteration 4. Discussion system for the proposals.

shamil-gadelshin:
Old PR discussion can be found here https://github.com/shamil-gadelshin/substrate-runtime-joystream/pull/2.

Features and comments:
1. discussion system prototype supports API and extrinsics:
- create_thread()
- add_post()
- update_post()
2. **codex** integrates proposals from the **engine** module with discussion threads from the **discussion** module
3. **membership** module integration is postponed until post-Rome: https://github.com/Joystream/substrate-runtime-joystream/issues/174
4. `post` storage is implemented via `double_map`
5. constant parameters were redesigned as `Get<>` implementation instead of `config()`

PR #261, opened 2020-04-01
Content directory 2.0

iorveth:
https://github.com/Joystream/joystream/issues/175
- Drop all events.
- Rename PropertyType::Internal to PropertyType::Reference, and same for PropertyValue::Internal, as we are dropping the notion of external references.
- Remove PropertyType::None, was added to have some value for required Default trait, but has no semantic content. Instead just return some other value needed, we should not sacrifice our type safety due to a bad Substrate design decision.


siman:
> Drop all events.
Interesting. Why do we drop all events? UI relies on some of them: When creating a video, UI gets an id of newly created video entity to redirect a user to the video page.

Issue #179, opened 2020-03-06
Fix use of array into iter


mnaamani:
https://github.com/Joystream/substrate-runtime-joystream/blob/385a376ed64e565ad410be76a7e424f0de340d66/src/lib.rs#L331

```
this method call currently resolves to `<&[T; N] as IntoIterator>::into_iter` (due to autoref coercions), but that might change in the future when `IntoIterator` impls for arrays are added.

note: `#[warn(array_into_iter)]` on by default
warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
note: for more information, see issue #66145 <https://github.com/rust-lang/rust/issues/66145>rustc(array_into_iter)
```

iorveth:
Most probably this warning originates from srml > staking > reward-curve module. 
Figured out the actual place, where it was introduced. 
![Screenshot from 2020-04-21 19-17-43](https://user-images.githubusercontent.com/18070359/79888365-2f361d80-8405-11ea-8acf-6b9feb515638.png)
I see 2 possible solutions here:

iorveth:
https://github.com/rust-lang/rust/issues/66145
Most probably this warning originates from `srml > staking > reward-curve module`. 
Figured out the actual place, where it was introduced. 
![Screenshot from 2020-04-21 19-17-43](https://user-images.githubusercontent.com/18070359/79888365-2f361d80-8405-11ea-8acf-6b9feb515638.png)
I see 2 possible solutions here:
- Wait for the moment, when we`ll update to the most recent substrate version, where this error is already fixed.
- Fork `reward-curve module`, replacing `into_iter` with `iter`

bedeho:
Def. lets wait, nice work!

mnaamani:
This is most likely resolved in https://github.com/Joystream/joystream/pull/1075 when updating substrate

PR #266, opened 2020-04-07
Merge versioned store & versioned-permissions core logic

iorveth:
Part of https://github.com/Joystream/joystream/issues/179
- `Class` and `ClassPermission`, only have one instance per class, and only one map
- `Entity` and `EntityPermission`, only have one instance per entity, and only one map.

### DIFF
--- a/runtime-modules/versioned-store-permissions/src/errors.rs
+++ b/runtime-modules/versioned-store-permissions/src/errors.rs
@@ -1,0 +1,38 @@
+// Validation errors
+// --------------------------------------
+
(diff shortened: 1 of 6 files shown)


bedeho:
I am not sure I understand what is going on here, wasnt this supposed to merge the logic into a single new module?

iorveth:
> I am not sure I understand what is going on here, wasnt this supposed to merge the logic into a single new module?

It was, i still did not remove the module i am merging from, i`ll do this after tests logic merging.

iorveth:
Sorry, it supposed to be draft pull request.

PR #267, opened 2020-04-07
Content directory 2.0

iorveth:
# Long-running PR for v2.0 of the content directory

## VLOS
- [ ] **https://github.com/Joystream/joystream-landing/issues/175**
- [x] Drop all events in VLOS

- [ ] Enforce unique class names. <= Why? perhaps we should even drop them?

- [ ] Enforce unique property names within each class. <= Why? perhaps we should even drop them?

- [x] Rename `PropertyType::Internal` to `PropertyType::Reference`, and same for `PropertyValue::Internal`, as we are dropping the no'tion of external references.

- [x] Remove `PropertyType::None`, was added to have some value for required Default trait, but has no semantic content. Instead just return some other value needed, we should not sacrifice our type safety due to a bad Substrate design decision.

- [x] Remove `Class::id` and `Entity::id`, as per convention #36 (comment).

- [x] Make `ClassId` and `EntityId`, configurable through runtime.

- [x] Raw index types u16 are used for index types, replace with aliases for each use case, e.g. `ClassId`, `EntityId`, `PropertyId` etc. for clarity. Also make public

- [x] Add boolean field in `ClassSchema` which indicates whether a schema is active or not. It is only possible to add schema support to an entity for an active schema. Introduce method for updating this status also.

- [x] `ClassSchema` -> `Schema`

- [x] Change values field of Entity to `BTreeMap<u16, PropertyValue>` where key is index of property in class, and kill `ClassPropertyValue`.

- [x] Change `in_class_schema_indexes` in `Entity` to `BTreeSet`, and give better name to field, like `supported_schemas`.

- [x] Joystream/substrate-versioned-store-module#13

- [x] Joystream/substrate-versioned-store-module#22

- [x] Drop from storage these from storage, they need not be tuneable at runtime, just make them configurable through runtime trait. 

```
pub PropertyNameConstraint get(property_name_constraint)
    config(): InputValidationLengthConstraint;

pub PropertyDescriptionConstraint get(property_description_constraint)
    config(): InputValidationLengthConstraint;

pub ClassNameConstraint get(class_name_constraint)
    config(): InputValidationLengthConstraint;

pub ClassDescriptionConstraint get(class_description_constraint)
    config(): InputValidationLengthConstraint;
```

- [ ] Make transactional https://github.com/Joystream/joystream/blob/development/runtime-modules/versioned-store-permissions/src/lib.rs#L341 using new transactiona overlay feature in Substrate.
- [ ] Remove `InputValidationLengthConstraint`, as it is no longer needed.
- [ ] Introduce a new trait for Module, so that other modules depending on VLOS can consume this trait on runtime, which also allows for mocking.

- [x] https://github.com/Joystream/joystream-landing/issues/179
- [x]  Merge permission and version store
- [x] `Class` and `ClassPermission`, only have one instance per class, and only one map
- [x] `Entity` and `EntityPermission`, only have one instance per entity, and only one map.
- [x] Drop `ClassPermission::properties_locked_from_controller`, and instead enrich `PropertyType` to include a locking flag for controllers.
- [x] Drop `ClassPermission::properties_locked_from_maintainers`, and instead enrich `PropertyType` to include a locking flag for maintainers.
- [x] Drop `ClassPermission::referenced_entity_must_have_same_controller`, and instead enrich `PropertyType::Reference` to include a same owner boolean flag.
## VLOS Permissions
- [x] Split credentials into groups and actors entities 
- [x] Implement respective logic for groups and actors entities
- [x] Design and implement new explicit entity roles respective logic

- [ ] **Security/configuration constraints for:**
- [ ] Number of classes
- [ ] Number of schemas
- [ ] Number of properties in schemas
- [ ] Size of property types in schemas
- [ ] Overall EntityCreationLimit
- [ ] EntityCreationLimit per individual
- [x] Implement a new entity permissions design
- [x] Implement a new class permissions design
- [x] Made property values lockable in general and for different actors, implement related logic
- [ ] Entities ownership transfer research and implementation
https://github.com/Joystream/joystream-landing/issues/178
- [ ] Respective test coverage for newly introduced logic
- [ ] Existing test coverage update
- [x] Already existing logic adjustment
- [ ] Full documentation
- [ ] Refactoring due to last substrate conventions


bedeho:
I had a chance to finally grab the partial work so far and get an overview when reviewing https://github.com/Joystream/joystream/pull/278. 

I think its best you go ahead and try to fully implement the entire module in one new PR, rather than this step by step process, because it's too hard to review in the intermediate state where I am not sure what is missing due to you not seeing it vs. not having gotten that far. 

I suspect this is a mistake, so I am pointing it out now for your benefit: You have included entity permissions as part of class permissions, this cannot work.

bedeho:
This target branch has been abandoned in favor of `content_directory_second_try`.

Issue #140, opened 2020-02-09
Delegatation in voting


bedeho:
Voters can pass their voting power to another voter. It probably only makes sense to be able to do this to voters that have explicitly announced that they accept delegation, this may require a new voting stage.

But this first https://github.com/Joystream/substrate-runtime-joystream/issues/139.

traumschule:
Did you also think about **voting controllers** that allow to control one or multiple stashes with one account to vote with all in one go for the same or multiple candidates?
The increased security of validator stashes is not available for council elections where stashes are needed "hot" every election. Another idea was to re-announce a candidacy (automatically), or joystream/pioneer#3865.
There might be a multi-sig solution but it seems voting power can't be delegated this way.

bedeho:
This is a great idea, it was a bad design to not have this, or something equivalent.

bedeho:
Another benefit pointed to by Yasir today is that it allows for more secure ways of using tokens to vote, as you can do a one time delegation, and then taking keys offline.

mochet:
* I think what would make this feature more "complete" is for there to be a social profile for each delegate (that can be updated occasionally). I have seen this in other projects, for example ChaosDAO on Polkadot where they have a nuanced explanation of how they vote.
* I also think there should be some kind of semi-substantial deposit/bond or burn required to create a delegate

PR #271, opened 2020-04-08
CLI Initial progress: council:info, basic account management

Lezek123:
This is my initial progress on the Joystream CLI sideproject (see: https://github.com/Joystream/joystream/issues/140)

I started building it using Node.js Open CLI Framework: https://oclif.io/

Currently there are just 8 commands available (updated at 10.04 to include those mentioned here: https://github.com/Joystream/substrate-runtime-joystream/issues/272):

**Accout management:**
- **account:import _JSONFILEPATH_** - allows importing account from json backup file so it can then be used within the CLI
- **account:choose** - allows the user to choose the default account to use inside CLI (lists all accounts and their available balances)
- **account:create _NAME_** - allows creating a brand new account (key pair).
- **account:forget** - allows the removal of imported/created accounts (user chooses the account to forget from the list)
- **account:current** - displays some information about currently choosen (default) account
- **account:export _PATH_ _[-a / --all]_** - allows exporting account(s) as json files to given path
- **account:transferTokens _ADDRESS_ _AMOUNT_** - allows transfering tokens from choosen account to given address.

**Council:**
- **council:info** - displays information about current council and council elections (similar to those that can be seen on the Council dashboard inside Pioneer)

Those commands are described more extensively in the auto-generated _README.md_ file.

The framework also provides flags like: --help, listings of commands in given category (ie. `.bin/run account`) etc.

Using main testnet endpoint (wss://rome-rpc-endpoint.joystream.org:9944) can be forced with the env variable `MAIN_TESTNET` set to `true`, ie.: `MAIN_TESTNET=true .bin/run account:current`, otherwise it uses the staging node by default (wss://rome-staging-2.joystream.org/staging/rpc/)


cli/src/Api.ts
```diff
@@ -0,0 +1,135 @@
...
+import { ApiPromise, WsProvider } from '@polkadot/api';
+import { formatBalance } from '@polkadot/util';
+import { Balance, Hash } from '@polkadot/types/interfaces';
+import { KeyringPair } from '@polkadot/keyring/types';
+import { AccountBalances, CouncilInfoObj, CouncilInfoTuple, AccountSummary, createCouncilInfoObj } from './Types';
+import { DerivedFees, DerivedBalances } from '@polkadot/api-derive/types';
+import { CLIError } from '@oclif/errors';
+import ExitCodes from './ExitCodes';
+
+const API_URL = process.env.WS_URL || (
```
bedeho, line 14: I think we are getting to the stage where the tool should allow the user to provide their own URL.

It does not need to be addressed in this PR, you can open a new one for this.

cli/src/Api.ts
```diff
@@ -0,0 +1,135 @@
...
+    private async initApi(): Promise<ApiPromise> {
+        formatBalance.setDefaults({ unit: TOKEN_SYMBOL });
+        const wsProvider:WsProvider = new WsProvider(API_URL);
+        registerJoystreamTypes();
+
+        return await ApiPromise.create({ provider: wsProvider });
+    }
+
+    private async getApi(): Promise<ApiPromise> {
+        while (this.initializing) {
```
bedeho: What is going on here? This construction looks very suspicious? If this was lifted from Pioneer, then its probably just because Pioneer is badly constructed. Let me know if I am missing something.

cli/src/Api.ts
```diff
@@ -0,0 +1,135 @@
...
+    process.env.MAIN_TESTNET ?
+        'wss://rome-rpc-endpoint.joystream.org:9944/'
+        : 'wss://rome-staging-2.joystream.org/staging/rpc/'
+);
+const TOKEN_SYMBOL = 'JOY';
+
+// Api wrapper for handling most common api calls and allowing easy API implementation switch in the future
+
+export default class Api {
+    private _api: ApiPromise | null = null;
```
bedeho: Revisit the role of these two state variables after you have looked at the comment about in `getApi`. Their values seem to probably have some invariant over it that can be enforced statically (e.g. `_api` is null if and only if `initializing` is true, or something like that´).

bedeho (changes requested):
- The install process seems to have failed on my machine for some reason. When I run the command after doing `npm install -g` in the source folder, I get

```sh
Bedehos-iMac-Pro:cli bedeho$ joystream-cli
(node:1925) [MODULE_NOT_FOUND] Error Plugin: joystream-cli: Cannot find module 'globby'
Require stack:
- /Users/bedeho/Work/substrate-runtime-joystream/cli/node_modules/@oclif/config/lib/plugin.js
- /Users/bedeho/Work/substrate-runtime-joystream/cli/node_modules/@oclif/config/lib/config.js
- /Users/bedeho/Work/substrate-runtime-joystream/cli/node_modules/@oclif/config/lib/index.js
- /Users/bedeho/Work/substrate-runtime-joystream/cli/node_modules/@oclif/command/lib/command.js
- /Users/bedeho/Work/substrate-runtime-joystream/cli/node_modules/@oclif/command/lib/index.js
- /Users/bedeho/Work/substrate-runtime-joystream/cli/bin/run
module: @oclif/config@1.14.0
task: not loading commands, globby not found
plugin: joystream-cli
root: /Users/bedeho/Work/substrate-runtime-joystream/cli
See more details with DEBUG=*
```

- Given the failure above, could we possibly add something to the CI tests that attempt to actually install the tool, and confirms that that process works? Even if you fix the issue above, something else is bound to break in the future. That could also be the start of what could become a set of integration tests, where the entire script is tested end-to-end, rather than just unit tests like now. This is possibly also the only way to make sure that the use of persisted state on the disk actually works end-to-end also.

Let me know if you think this is a good idea, if so, let's do that work in a separate PR.

- The "See code" links in the README do not work.

- Can you add a descriptive command to all types you make.

- There appears to be a `state.json` file, but its not easy to understand what you put there without reading all of the code in detail. Could you find some way of summarizing the data model you use for state that is persisted to disk. This could be either/both in the README and at the source level. Whatever you think works best.

cli/src/Api.ts
Lezek123: I refactored this class a few times, so that's why it may now seem kind of weird.
The idea was that it should prevent accidentialy initializing the api twice. So if the api initialization was already started somewhere, it should then await 10 ms intervals until the the initialization process has finished. So in case we do something like:

```
const api: Api = new Api();
api.getCouncilInfo(); // Async call, initializes the api if not yet initialized
api.getAccountSummary(SOME_ACCOUNT); // Same as above
```

The second call should not re-initialize the api, but instead wait until the first call finishes the initialization.

This made a little more sense when this class was still a singleton, later I decided to change it into something more like the original ApiPromise, so:

```
const api = await Api.create();
// Now we can just use the api knowing that it has already been initialized.
// It may also throw an error if we try to use it before initialization
```

But I probably kind of confused myself during the process and ended up with some hybrid of two potential approches. I'll fix it, by going for the second one, as it seems more "clean".

Lezek123:
> * The install process seems to have failed on my machine for some reason. When I run the command after doing `npm install -g`

It seems like there is an `oclif.manifest.json` file required in order for the npm install to work correctly (source: https://github.com/oclif/oclif/issues/132).
If the file is present, it is assumed that the CLI should run in "production mode".
This file should be automatically generated during `npm publish` / `npm pack` (see: prepack and postpack scripts in `package.json`), so I'm not sure whether we need to include it in the repository. Once the app is published, the `npm install -g joystream-cli` should work correctly. In dev environment we can use `npm link`, as is recommended in the official documentation (ie. https://oclif.io/docs/multi).

bedeho (approved):
Great that you have a hypothesis for why the install does not work. Could you add something to the README about how to work with the cli in development mode then? e.g. if you are working on it while making local changes, or if you just cloned the repo.

Lezek123:
> Great that you have a hypothesis for why the install does not work. Could you add something to the README about how to work with the cli in development mode then? e.g. if you are working on it while making local changes, or if you just cloned the repo.

Added the "Development" section into the README

PR #273, opened 2020-04-08
Merge vlos & vlos permissions modules

iorveth:
Part of  **Proposal: Merge permission and version store** https://github.com/Joystream/joystream/issues/179

- [x] `Class` and `ClassPermission`, only have one instance per class, and only one map
- [x] `Entity` and `EntityPermission`, only have one instance per entity, and only one map.

**VLOS**
- [x] Drop all events.
- [x] Remove `Class::id` and `Entity::id`, as per convention #36 (comment).
- [x] Rename `PropertyType::Internal` to `PropertyType::Reference`, and same for PropertyValue::Internal, as we are dropping the notion of external references.
- [x]  Remove `PropertyType::None`, was added to have some value for required Default trait, but has no semantic content. Instead just return some other value needed, we should not sacrifice our type safety due to a bad Substrate design decision.
- [x]  Add boolean field in `ClassSchema` which indicates whether a schema is active or not. It is only possible to add schema support to an entity for an active schema. Introduce method for updating this status also.
- [x]  `ClassSchema -> Schema`
- [x]  Change values field of Entity to `BTreeMap<u16, PropertyValue>` where key is index of property in class, and kill `ClassPropertyValue`.
- [x] Change `in_class_schema_indexes` in `Entity` to `BTreeSet`, and give better name to field, like `supported_schemas`.
- [x] Drop from storage these from storage, they need not be tuneable at runtime, just make them configurable through runtime trait.
```
pub PropertyNameConstraint get(property_name_constraint)
    config(): InputValidationLengthConstraint;

pub PropertyDescriptionConstraint get(property_description_constraint)
    config(): InputValidationLengthConstraint;

pub ClassNameConstraint get(class_name_constraint)
    config(): InputValidationLengthConstraint;

pub ClassDescriptionConstraint get(class_description_constraint)
    config(): InputValidationLengthConstraint;
```


bedeho (approved):
This looks like a good start.

There is still work here having to do with governance and other stuff, not sure why, but we need to move forward.

bedeho:
Actually, this does not pass cargo fmt check, please add an extra commit that fixes this, then I will merge.

Issue #148, opened 2020-02-10
Councilors receive reward


bedeho:
Is needed, should use recurring reward module. For now, let root origin set conditions, and all councilors have same conditions.

mnaamani:
Implemented in #275

PR #275, opened 2020-04-11
Runtime: Council - recurring rewards

mnaamani:
Implement #148


runtime-modules/governance/src/council.rs
```diff
@@ -37,6 +37,20 @@ decl_storage! {
         /// because it was introduced in a runtime upgrade. It will be automatically created when
         /// a successful call to set_council_mint_capacity() is made.
         pub CouncilMint get(council_mint) : Option<<T as minting::Trait>::MintId>;
+
+        /// The reward relationships currently in place. There may not necessarily be a 1-1 correspondance with
+        /// the active council, since there are multiple ways of setting/adding/removing council members, some of which
+        /// do not involve creating a relationship.
+        pub RewardRelationships get(reward_relationships) : map T::AccountId => T::RewardRelationshipId;
```
bedeho, line 44: Why is an account id used as the identifier here? Shouldn't the relationship id just live in the representation of the council member?

runtime-modules/governance/src/council.rs
```diff
@@ -75,6 +85,73 @@ impl<T: Trait> Module<T> {
...
+
+            Ok(relationship_id)
+        } else {
+            Err(recurringrewards::RewardsError::RewardSourceNotFound)
+        }
+    }
+
+    fn remove_reward_relationships() {
+        for seat in Self::active_council().into_iter() {
+            if RewardRelationships::<T>::exists(&seat.member) {
```
bedeho, line 131: Why is this `exists` test here? This implies its possible for the council member to have no reward?

runtime-modules/governance/src/council.rs
```diff
@@ -75,6 +85,73 @@ impl<T: Trait> Module<T> {
...
+                let id = Self::reward_relationships(&seat.member);
+                <recurringrewards::Module<T>>::remove_reward_relationship(id);
+            }
+        }
+    }
+
+    fn apply_council_seats(seats: Seats<T::AccountId, BalanceOf<T>>, term: T::BlockNumber) {
```
bedeho: What does this method actually do? this name doesn't say muhc, and is hard to reconcile with the body.

runtime-modules/governance/src/council.rs
```diff
@@ -75,6 +85,73 @@ impl<T: Trait> Module<T> {
...
+        <ActiveCouncil<T>>::put(seats.clone());
+
+        let next_term_ends_at = <system::Module<T>>::block_number() + term;
+
+        <TermEndsAt<T>>::put(next_term_ends_at);
+
+        for seat in seats.iter() {
+            let reward_added_result = Self::add_reward_relationship(&seat.member);
+
+            if reward_added_result.is_err() {
```
bedeho: What kind of error is this? what can cause?

bedeho (changes requested):

runtime-modules/governance/src/council.rs
mnaamani: Ideally yes, the council member is represented in the `struct Seat`.
I purposefully avoided modifying the struct to avoid dealing with decoding issues after runtime upgrade. This is not the first I've had to go this sub-optimal route.
Should I refactor and introduce a custom Decode impl to deal with this migration issue?

runtime-modules/governance/src/council.rs
mnaamani: Yes. There are two reasons for such a case.

1. After a runtime upgrade the existing council doesn't have rewards (I didn't add any migration step or default values for rewards)

2. There are helper methods for adding and setting a council directly with a dispatchable,  insert_member() and set_council().  I think we should get rid of these methods and only allow council to be elected, so I didn't bother to add rewards for seats added with these methods.

runtime-modules/governance/src/council.rs
mnaamani: this just has the body of code from `council_elected()` I moved it into its own method because I was having issue calling it from tests but later realized it was just a scoping issue.
I can either rename it to `on_council_elected()` or just move the body back into it original location.

runtime-modules/governance/src/council.rs
mnaamani: `recurringrewards::RewardsError`

Since the council mint is an `Option<MintId>` it could fail with `recurringrewards::RewardsError::RewardSourceNotFound`

runtime-modules/governance/src/council.rs
bedeho: From the internal discussion, this is basically just due to needing to side step migration of `Seat`.

runtime-modules/governance/src/council.rs
bedeho: From internal discussion: the first council post migration has no reward relationships, so this must deal with this.

runtime-modules/governance/src/council.rs
bedeho: Was moved back.

runtime-modules/governance/src/council.rs
bedeho: So we ignore this issue in this release, made an issue for the future here
https://github.com/Joystream/joystream/issues/307

runtime-modules/governance/src/council.rs
bedeho: So we ignore this issue in this release, made an issue for the future here
https://github.com/Joystream/joystream/issues/307

runtime-modules/governance/src/council.rs
bedeho: - This now appears to have been rewritten so as to just log if there is no mint set, or if the call to `<recurringrewards::Module<T>>::add_reward_relationship` fails. Pre-sumably this is because either error means the runtime is broken. That choice is fine with me.

- I don't think having an optional mint makes sense, because it does not reflect what is actually going on. But even beyond this, having a non-staic relationship with the mint itself -  let alone having a mutable mint id in stroage - does not make sense, because we do indeed statically want the council to have a mint. I think this is a deeper issue that we need to explore at a later time, perhaps the way we are using mints & recurring rewards is actually not that great. Will open some issue(s) on this later.

mnaamani:
@bedeho merge conflicts fixed and ready for merge

PR #278, opened 2020-04-13
Content dir 2 issue 13

iorveth:
## Make vector properties updatable as vectors

Joystream/substrate-versioned-store-module#13

### DIFF
--- a/runtime-modules/content-directory/src/errors.rs
+++ b/runtime-modules/content-directory/src/errors.rs
@@ -40,5 +40,7 @@ pub const ERROR_TEXT_PROP_IS_TOO_LONG: &str = "Text propery is too long";
 pub const ERROR_VEC_PROP_IS_TOO_LONG: &str = "Vector propery is too long";
 pub const ERROR_ENTITY_PROP_VALUE_VECTOR_IS_TOO_LONG: &str =
     "Propery value vector can`t contain more values";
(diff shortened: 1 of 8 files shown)


bedeho (changes requested):
My understanding of this solution is that it enforces the rule that each vector typed property value can only be mutated, in any way, at most once per entity per block.

Is this correct? 

If so, then this is really overly restrictive solution to the problem. You should keep in mind that both Substrate and Ethereum use nonces to regulate race condition/replay attacks, and neither prevents a single account to make at most one transaction per block. Not only is it too restrictive, it actually does not even solve the full problem, because two competing transactions can still make a conflicting update, so long as it's across two blocks. What you want is to _force_ each transaction to explicitly commit to known state of the vector.

Instead what you want is to introduce an explicit counter per such entity & value combination, and increment it each time there is a mutation operation. Whenever someone makes an extrinsic to make a vector mutation on an entity, they must provide the current counter value as a parameter, and the extrinsic is only accepted if it matches.

This approach allows someone to make multiple extrinsics in a row, with an increasing nonce value in each, where all can be accepted in the same block.

bedeho (changes requested):
This was a good step in the right direction. I was, however, hoping you would see that the proper place for the nonce value is `PropertyValue` representation. So

```Rust
pub enum PropertyValue {
...
    BoolVec(Vec<bool>, T::Nonce),
...
}
```

bedeho:
Superceeded by https://github.com/Joystream/joystream/pull/267

PR #279, opened 2020-04-13
Proposals.Version 2.Iteration 8

shamil-gadelshin:
### Features

- veto for the approved pending execution proposal
- reset a proposal after council reelection
- change upgrade runtime proposal: introduce fixed allowed member list
- add proposals: 'evict storage provider', 'set validator count', 'set storage role parameters'
- set actual proposal type parameters
- set actual proposal runtime parameters
- change 'veto_proposal' to affect approved proposal during the grace period
- fix clippy linter issues
- migrate proposal votes calculations to Perbill 
- remove allowed member_id list for runtime upgrade proposal
- improve comments


bedeho (changes requested):
- We decided to not have proposal for setting council capacity, for reasons that are too nuanced to unpack here, so just drop that one.

- Nice work on the docs!

bedeho (approved):

PR #283, opened 2020-04-14
Joystream/Substrate Query Node

bedeho:
# Design

https://github.com/Joystream/joystream-landing/issues/208

# Roadmap & Progress

- [x] Define input format for specifying a query API.
   - Make sure documentation in the input format can be propagated to the service documentation.
- [x] Make CLI tool that takes input format and generates database+resolvers+service using Warthog tooling under the hood.
- [x] Update CLI tool to accept event processor typescript code, together with API specification, and generate full set of assets (graphql server, indexer, database), and have some easy way to run in concert.
- [x] Use existing tooling to make Joystream query infrastructure that covers membership module.
- Support all scalars available in `The Graph`
  - [x] Boolean
  - [x] String
  - [ ] Bytes
  - [x] List
  - [x] BigInter
  - [x] Decimal
- Support FTS
  - [x] Multi-field
  - [x] Multi entity
- [ ] API preview
- [ ] Reverse entity lookup
- [ ] Enumerated types


bedeho:
I think we achieved the basic goals here, so I will merge when CI is ready, then we can get started on a new long running PR with some new goals.

Issue #224, opened 2020-03-23
Implement Migration steps from Rome


mnaamani:
Any migration steps needed when upgrading to Constantinople:

1. Reset content working group mint capacity to zero.
1. Deactivate active curators.
1. Set Storage Role reward to zero.
1. Deactivate all storage providers
1. Initialize the Council mint.
1. Reset the Council

mnaamani:
At least for balances there will not be any migration step. Keeping it open in case we need run some other "on runtime upgrade" tasks

mnaamani:
We should also set initial values for storage values that are expected to be set at genesis in new modules introduced in constantinople,  the proposal codex, engine and discussion modules.

mnaamani:
All the steps were implemented but we only ended up doing two steps in the upgrade:

- Initialize the Council mint
- Set initial proposal parameters which would have been set to 0 since the build() step would not have had an effect given we were doing a runtime upgrade.

It was decided that all other steps could be done manually with sudo, and it was safer that way since no tests were written for the migration steps.

PR #284, opened 2020-04-15
Runtime: Migration module - migrations for constantinople

mnaamani:
Ensure `build()` is called for SpecVersion state storage value to be correctly initialized, otherwise the call to runtime_upgraded() was incorrectly being called at block 1.

Will also implement additional migration steps from: https://github.com/Joystream/joystream/issues/224


shamil-gadelshin (approved):
LGTM

bedeho (approved):

PR #286, opened 2020-04-15
CLI: "api:" commands family (inspecting api and specifying custom api url)

Lezek123:
This pull request contains changes described in this issue:
https://github.com/Joystream/substrate-runtime-joystream/issues/281

The only difference is that I decided that `TYPE` will actually not be an argument, but a flag of the `api:inspect` command (it's easier to manage dependencies between flags in OCLIF), so the command usage is currently like this:

`api:inspect --type=type --module=module --method=method --exec --callArgs=args`

I also created the shortcuts for all the flags, so the shorter version will be:

`api:inspect -t=type -M=module -m=method -e -a=args`

When no `--callArgs` / `-a` flag is specified, but the `--exec` / `-e` is - user is prompted to provide the required argument(s) (this can also happen if he did specify some, but not all arguments via `--callArgs`). During this "prompring process" some more complex input types are also taken into account (ie. `Tuple`, `Option` etc.). This can be tested by calling methods like:

`$  ./bin/run api:inspect -t=query -M=dataObjectStorageRegistry -m=storageProviderServesContent -e`
_(The Tuple example)_

`$  ./bin/run api:inspect -t=query -M=stake -m=stakes -e`
_(The Option example)_

All the introduced methods (`api:inspect`, `api:setUri` and `api:getUri`) are described in the updated README file and can also be analyzed by using the `--help` flag.

Those changes required some refactoring regarding the API construction and `state.json` file.
I created a few abstract base classes for commands, that help to handle this logic:

- `DefaultCommandBase` - this is a base class for pretty much all of the commands.
- `StateAwareCommandBase` - extends the above class, this is a base class for all commands that require access to the state that is preserved across sessions (ie. selected account, api uri). Currently that's true for pretty much all of the commands also.
- `ApiCommandBase` - extends the above class, this is a base class for all commands that require access to the API.
- `AccountsCommandBase` - extends the above class, this is a base class for all `account:` family commands (and probably lots of other commands in the future), containing some shared account-related logic (like fetching currently selected account etc.)


bedeho (approved):
Well done, nothing worth changing here.

PR #287, opened 2020-04-16
Network tests: council - fix reading council from state

mnaamani:
serializing then parsing JSON is a neat trick but not necessary. I forgot to comment on this in previous PR review. @gleb-urvanov


mnaamani:
This was for demo purpose only. Should be incorporated in next PR being worked on by @gleb-urvanov

Issue #282, opened 2020-04-14
CLI: Memberships management


Lezek123:
There's a need to manage not only the accounts (keys), but also the available memberships and roles inside the CLI in order to allow the user to specify the membership and role as a context for executing some of the commands that are going be implemented in the future (related to voting and governance).

I already started implementing this with the assumption that each account (pair of keys) always has either one or zero related memberships, so the base memership and available roles can be determined based on the currently selected account, but this may require some more inspection.

Some initial ideas I've had for commands that would need to be included:

- `member:register` - create a membership profile for the currently selected account (if it doesn't already exist)
- `member:current` - display information about currently used membership (and available roles)
- `member:edit` - edit the current membership profile

There are definitely going to be more, but I'm still trying to figure out the correct way to handle this. I will update the issue as I go along.

Some commands for discovering other members will probably also need to exists, like listing, reading other members profiles etc., but I think that can be handled in a separate issue (as this one is focused mostly on how to manage user's own memberships)

PR #292, opened 2020-04-16
CLI: Memberships management

Lezek123:
This PR contains some of my initial work related to memberships management.
Related issue: https://github.com/Joystream/joystream/issues/282

### Determining current membership

Currently in my implementation there is an assumption that each account (key pair) either has or hasn't got an associated membership, this logic is based more or less on how the account's memership is beeing determined in the Pioneer's _Memberships_ tab.

All commands that require information about user's current membership (based on the account currently selected via `account:choose`) can extend the abstract `AccountsCommandBase` class and then use:

`const membership = await this.getRequiredMembership();`

Or:

`const membership = await this.getRequiredMembership(false);`

In the first example the user will be prompted to select an account first (if no account was previously selected via `account:choose`). In the second example the command will exit with the error message if no account is already choosen. In both cases the command will exit with an error if the current account has no associated membership.

The `Api` class has been extended with a few methods related to fetching the account membership information and currently active membership terms. I hope this will allow to change the logic of retrieving user's available memberships easily in the future if there's a need for that (ie. if there is or would be a way for one account to have multiple associated memberships or if selecting a membership will have to become something prior to selecting an account).

### Commands implementation

**There are 2 new commands already implemented:**

- `membership:current` - displays information about membership related to currently selected account (via `account:choose`). **The way it currently displays associated roles is not the best, so this could be changed, ie. if I have two channels it displays: `Roles: ChannelOwner ChannelOwner`.**
- `membership:terms` - displays currently active paid membership terms. In Pioneer those are used ie. to display the current fee when user wants to register a new membership, so this logic can be used when implementing a command like `membership:register`.

**Commands I planned to implement:**

I also created a draft of `membership:edit` command that will allow the user to modify the profile information. I was planning to use the `api.tx.members.updateProfile` and update the values based on provided flags. Prompting for input if no flags are provided can also be implemented. There is some logic of prompting for optional arguments already in `ApInspect.promptForOption` (`commands/api/inspect.ts`).

I was also planning to implement the `membership:register`, as this should be pretty easy once we have `membership:edit` and `membership:terms`.

I'm not sure if this way of handling memberships will be sufficient in regards to activities like voting in elections etc., but from what I could gather, there is not too much distinction currently beeing made between accounts (keys) and memberships (besides the fact that some accounts may not have an associated membership). If there is a need to handle this differently in the future, I think this will not be too hard to change and some way of selecting a membership could be implemented in a similar manner to how the account selection is currently possible via `account:choose`.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -69,6 +69,9 @@
       "account": {
         "description": "Accounts management - create, import or switch currently used account"
       },
+      "membership": {
+        "description": "Memberships management - register membership, display and modify your profile"
+      },
       "api": {
         "description": "Inspect the substrate node api, perform lower-level api calls or change the current api provider uri"
       }
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -6,9 +6,12 @@ import { formatBalance } from '@polkadot/util';
 import { Hash } from '@polkadot/types/interfaces';
 import { KeyringPair } from '@polkadot/keyring/types';
 import { Codec } from '@polkadot/types/types';
-import { AccountSummary, CouncilInfoObj, CouncilInfoTuple, createCouncilInfoObj } from './Types';
+import { AccountSummary, CouncilInfoObj, CouncilInfoTuple, createCouncilInfoObj, AccountMemberships } from './Types';
 import { DerivedFees, DerivedBalances } from '@polkadot/api-derive/types';
 import { CLIError } from '@oclif/errors';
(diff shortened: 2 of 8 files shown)


bedeho (approved):
It looks like a good start.

Will just leave this PR as draft, then someone can carry on and complete all remainig commands in fresh PR later.

Lezek123:
Very outdated now, will be re-done for Olympia: https://github.com/Joystream/joystream/issues/3011

PR #297, opened 2020-04-19
Query node

metmirr:


bedeho (changes requested):
- I expect this CLI tool to grow increasingly complex over time, so you are probably better off just using a framework for this. We are using this in another CLI, seems to work well, and gives you a lot for free. https://oclif.io/

But do this in a separate PR

- The CLI tool does not accept the schema json as an input that can be specified on the command line. It just tries to require a hard coded path. Not only is this inflexible, but it also means that the CLI cannot gracefully handle if the input is malformed in some way, and tell the user about this, it will just crash? Won't it?

metmirr:
2. 

- Accept schema path from command line. The default location is `./schema.json`
- Generate errors for invalid input data

bedeho (approved):

Issue #305, opened 2020-04-24
Fix cargo install problem


shamil-gadelshin:
Fix an error during "cargo install joystream-node --path node/" command

error[E0412]: cannot find type `GenesisConfig` in crate `proposals_codex`
   --> runtime/src/lib.rs:872:1
    |
872 | / construct_runtime!(
873 | |     pub enum Runtime where
874 | |         Block = Block,
875 | |         NodeBlock = opaque::Block,
...   |
919 | |     }
920 | | );
    | |__^ not found in `proposals_codex`

shamil-gadelshin:
Released.

PR #306, opened 2020-04-24
Fix ‘cargo install issue’ with codex

shamil-gadelshin:
GenesisConfig couldn't be found only when trying to install the node using the
"cargo install" command.

The fix seems to be unrelated to the problem, but it fixes it:

The "proposals_*" crates were added to the 'std' feature list.

Related issue:
https://github.com/Joystream/joystream/issues/305

### DIFF
--- a/runtime-modules/proposals/codex/Cargo.toml
+++ b/runtime-modules/proposals/codex/Cargo.toml
@@ -25,6 +25,8 @@ std = [
     'governance/std',
     'mint/std',
     'roles/std',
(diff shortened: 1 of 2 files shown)


mnaamani:
Good catch. Although I'm surprised this has only showed up when doing `cargo install` and not when doing `cargo build --release`. Perhaps we should add `cargo install` to a travis job.

mnaamani (approved):

PR #308, opened 2020-04-24
Clippy linter support &  fixes

shamil-gadelshin:
## Notes
- this PR contains a lot of different [clippy](https://github.com/rust-lang/rust-clippy/) linter fixes.
- travis.yml file was modified to support mandatory lint checks.
- some lints were disabled because they are not compatible with Substrate macros design.
- clippy install command: `rustup component add clippy`
- to run clippy locally use `BUILD_DUMMY_WASM_BINARY=1 cargo clippy`

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -1,216 +1,10 @@
-// Copyright 2017-2019 Parity Technologies (UK) Ltd.
-
-// This is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-// You should have received a copy of the GNU General Public License
-// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
-
-// Copyright 2019 Joystream Contributors
-
-//! # Runtime Example Module
-//!
-//! <!-- Original author of paragraph: @gavofyork -->
-//! The Example: A simple example of a runtime module demonstrating
-//! concepts, APIs and structures common to most runtime modules.
-//!
-//! Run `cargo doc --package runtime-example-module --open` to view this module's documentation.
-//!
-//! ### Documentation Template:<br>
-//! Add heading with custom module name
-//!
-//! # <INSERT_CUSTOM_MODULE_NAME> Module
-//!
-//! Add simple description
-//!
-//! Include the following links that shows what trait needs to be implemented to use the module
-//! and the supported dispatchables that are documented in the Call enum.
-//!
-//! - [`<INSERT_CUSTOM_MODULE_NAME>::Trait`](./trait.Trait.html)
-//! - [`Call`](./enum.Call.html)
-//! - [`Module`](./struct.Module.html)
-//!
-//! ## Overview
-//!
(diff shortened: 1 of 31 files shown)


runtime-modules/governance/src/election.rs
```diff
@@ -351,17 +360,21 @@ impl<T: Trait> Module<T> {
...
-        } else if new_council.len() > Self::council_size_usize() {
-            // we have more than enough applicants to form the new council.
-            // select top staked
-            Self::filter_top_staked(&mut new_council, Self::council_size_usize());
-        } else {
-            // Not enough applicants with votes to form a council.
-            // This may happen if we didn't add applicants with zero votes to the tally,
-            // or in future if we allow applicants to withdraw candidacy during voting or revealing stages.
-            // or council size was increased during voting, revealing stages.
+        match new_council.len() {
```
mnaamani, line 363: I'm not sure what clippy error resulted in need to replace this code with a match statement. Is it okay to use a match statement like `switch` ?

runtime-modules/forum/src/lib.rs
```diff
@@ -1153,6 +952,9 @@ impl<T: Trait> Module<T> {
         Self::ensure_can_mutate_in_path_leaf(&category_tree_path)
     }
 
+    // Clippy linter warning
+    #[allow(clippy::ptr_arg)] // disable it because of possible frontend API break
```
mnaamani, line 956: `ensure_can_mutate_in_path_leaf()` is private so it shouldn't break public API.

runtime/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = '2018'
 name = 'joystream-node-runtime'
 # Follow convention: https://github.com/Joystream/substrate-runtime-joystream/issues/1
 # {Authoring}.{Spec}.{Impl} of the RuntimeVersion
-version = '6.12.0'
+version = '6.12.2'
```
mnaamani, line 8: When updating version here we should also remember to change the version numbers in [runtime/src/lib.rs](https://github.com/Joystream/joystream/blob/development/runtime/src/lib.rs#L124)

mnaamani (changes requested):
Great start with linting.

Just need to update the `impl_version`
It also needs a cargo-fmt, (I did it locally and found it needed some formatting) surprised that travis didn't catch it.

mnaamani:
I think the issue with getting different rustfmt results might be because I have newer version locally than what is used in travis CI

runtime-modules/governance/src/election.rs
shamil-gadelshin: I fixed this lint warning here:
    = note: `#[warn(clippy::comparison_chain)]` on by default
    = help: Consider rewriting the `if` chain to use `cmp` and `match`.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: The problem here is not a method but its parameter type:
` &CategoryTreePath<T::BlockNumber, T::Moment, T::AccountId>,`

The linter suggested to replace it with an array pointer because CategoryTreePath is a Vec alias, and this, in turn, can break the API.

runtime/Cargo.toml
shamil-gadelshin: Increased from 0 to 1.

runtime-modules/governance/src/election.rs
mnaamani: Interesting. Thanks for clarification.

mnaamani (approved):

PR #309, opened 2020-04-26
Query node

metmirr:
This PR contains all work that has been done query node CLI. CLI generates a GraphQL server and block indexer node. Block indexer is based on https://github.com/bedeho/Hydra/tree/basic_querying/src/index-builder


joystream-query-node/cli/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
bedeho: `.gitignore`

bedeho (changes requested):
Here is a superficial summary of our conversation on this PR so far.

- Cleanup directory organization:
  - It may be a good idea to put everything under `query-node` top-level folder, rather than have two separate folders floating around.
  - `joystream-query-node` should only hold source assets needed to make and use the query node. Nothing autogenerated should be committed here. Basically just: mappings.ts + api.json + config.json ( set full node URL, Postgres authentication, port information for API server)
  - `substrate-query-node` should only hold source assets needed to make a query node for any runtime based on Substrate.
  -  Update READMEs to accurately describe the new organization.
- Radically simplify workflows: Bring the need for manual editing, navigating folders and multiple commands.
  - CLI tool should create a new folder `generated` which the developer should never edit, which holds
    -  warthog autogenerated stuff   
    -  index.js can start indexer + API server, uses `joystream-query-node` as a dependency
    -  `package.json`: dependency file for index.js process
  - It's not obvious whether `mapping.ts` and `config.json` needs to be copied into `generated`, perhaps it's a good idea to just have `index.js` consume them using a relative path, that way, there is no need to run a new command when these inputs are changed.

# Notes

 - Obviously `index.ts` needs not be one single file like that, its just meant as a representation for the entry point for actually starting and running the indexer and api servers. They could be separate apps even. So you have `generated/indexer` and `generated/api_server`, with separate codebases and processes. The important point is just that they should use existing via `joystream-query-node` as a dependency, not copying.

metmirr:
Updated directory structure

metmirr:
Update:

- Cleanup directory organization: 
  - Everything is under top-level `query-node` directory
  - `query-node/joystream-query-node` only holds `mappings/`, `schema.json` and `.env`
  - `query-node/substrate-query-node` holds cli and indexer template. Indexer template is a package that can be installed. CLI installs the indexer template package when creating an indexer for the joystream query node.
  - README updated but not completed yet.
- Radically simplify workflows: Bring the need for manual editing, navigating folders, and multiple commands.
  - Everything is generated and stored under the `generated` folder.
  - `mappings` are imported by indexer any update on mappings only requires a restart.
- Graphql server is under `generated/graphql-server` and indexer is under `generated/indexer`. It's better to keep them separated because warthog has its own files etc. It will be complex if putting them into the same directory.

metmirr:
`query-node/substrate-query-node` holds source assets needed to make a query node for any runtime based on Substrate but it also requires type register function. On the initialization of the indexer you must pass type register function as a parameter. For example the indexer for joystream is:

```ts
import { registerJoystreamTypes } from "@joystream/types";
const node = new QueryNodeManager(process);
node.start(providerUri, getProcessingPack(), registerJoystreamTypes);
```

So the question is should we add register type function into config file so developers can edit for their runtime. If yes then we would have something like this in the config file:

```
TYPE_REGISTER_PACKAGE_NAME=@joystream/types
TYPE_REGISTER_FUNCTION=registerJoystreamTypes
```

bedeho:
Good question.

First off, recall that `index.ts` (as I call it), consumes `substrate-query-node` library, so it can pass the registration function to it, so there is no need to touch the source of `substrate-query-node` directly, it just needs to take it as an argument. The question is then: where does the `index.ts` get it from!? I guess the most direct solution is what you sketch: the `index.ts` which is generated by the CLI will include the right import statement and argument passing, and the CLI will do this by looking at the config file.

It feels a bit rough, but it's not too bad, and perhaps we can figure out a better way in the future. As long as the CLI has to output index.ts anyway, then this is not that much worse.

bedeho (changes requested):
Nice work, let us do some final extra cleanup, then this is sufficient I think.

- Rename the folder `substrate-query-node`, perhaps we can call it `substrate-query-framework`, since its not directly a node, but really a set of tools, libs and formats which are used to make a node.

- Drop subfolders of `substrate-query-node/src/`

bedeho (approved):

PR #315, opened 2020-04-27
Create migration of the default values for the proposals codex module

shamil-gadelshin:
### Features
- change runtime migration
- change proposals default values on genesis config creation
- add a test


mnaamani (approved):
Locally I'm seeing different result for rustfmt compared with travis but it might be because I'm using newer version.

PR #319, opened 2020-04-28
Feature/rome constatinople migration test

gleb-urvanov:


tests/network-tests/package.json
```diff
@@ -4,17 +4,20 @@
...
-    "test": "mocha -r ts-node/register src/tests/*",
+    "test": "mocha -r ts-node/register src/tests/constantinople/*",
+    "test-migration": "mocha -r ts-node/register src/tests/rome/* && mocha -r ts-node/register src/tests/constantinople/*",
     "lint": "tslint --project tsconfig.json",
     "prettier": "prettier --write ./src"
   },
   "dependencies": {
-    "@joystream/types": "^0.7.0",
+    "@joystream/types": "../joystream-apps/packages/joy-types",
+    "@rome/types@npm:@joystream/types": "^0.7.0",
```
mnaamani, line 14: I see the first "@joystream/types": "../joystream-apps/packages/joy-types", is just a relative path to the latest constantinople @joystream/types package.

But I wanted to confirm my understanding, "@rome/types@npm:@joystream/types" is setting some sort of alias from `@rome/types` to `@joystream/types` published on npm ?

mnaamani: Would it be possible to do something like:

```
    "@constantinople/types@npm:@joystream/types": "^0.8.0",
    "@rome/types@npm:@joystream/types": "^0.7.0",
```

This might be useful but what would be more helpful is to see if we can have two packages in the same workspace setup the same alias to two different versions of the same package name?

tests/network-tests/src/tests/constantinople/electingCouncilTest.ts
```diff
@@ -1,13 +1,13 @@
 import { membershipTest } from './membershipCreationTest';
 import { KeyringPair } from '@polkadot/keyring/types';
-import { ApiWrapper } from '../utils/apiWrapper';
+import { ApiWrapper } from '../../utils/apiWrapper';
```
mnaamani: It probably makes sense to move the top `utils` folder inside constantinople folder (like rome) so it will be clear which code is for which version of the network.

tests/network-tests/src/tests/rome/romeRuntimeUpgradeTest.ts
```diff
@@ -36,23 +39,23 @@ describe('Runtime upgrade integration tests', () => {
...
+    const runtimeVoteFee: BN = apiWrapper.estimateVoteForRomeRuntimeProposalFee();
 
     // Topping the balances
     await apiWrapper.transferBalance(sudo, m1KeyPairs[0].address, runtimeProposalFee.add(proposalStake));
     await apiWrapper.transferBalanceToAccounts(sudo, m2KeyPairs, runtimeVoteFee);
 
     // Proposal creation
     const proposalPromise = apiWrapper.expectProposalCreated();
-    await apiWrapper.proposeRuntime(
+    await apiWrapper.proposeRuntimeRome(
```
mnaamani: I'm not so sure about adding `Rome` suffix here in the names of the functions. It doesn't seem necessary if we are separating the code for different networks into separate folders.

tests/network-tests/package.json
mnaamani: We cannot merge with package.json having such a relative path to a dependency.
How can we tackle this.

tests/network-tests/src/tests/rome/utils/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  public static wait(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  public static getTotalStake(seat: Seat): BN {
+    return new BN(+seat.stake.toString() + seat.backers.reduce((a, baker) => a + +baker.stake.toString(), 0));
+  }
+
+  public static readRuntimeFromFile(path: string): string {
+    return '0x' + fs.readFileSync(path).toString('hex');
```
mnaamani: Why is this function necessary. In other words, why are we having to pass the wasm blob as a hex string in the call to create the runtime upgrade proposal. Isn't it possible to pass the `Buffer` type or at least convert it more efficiently to the required `Codec` type.

mnaamani (changes requested):
I was able to run the test-migration successfully. I left a few comments about re-organizing some of the folder structure, but I'm okay with these changes being made later as there are several other PRs depending on this.

The main issue that needs to be resolved here however is how to properly have dependencies mutltiple different versions of our "types" package.

tests/network-tests/package.json
gleb-urvanov: > I see the first "@joystream/types": "../joystream-apps/packages/joy-types", is just a relative path to the latest constantinople @joystream/types package.
> 
> But I wanted to confirm my understanding, "@rome/types@npm:@joystream/types" is setting some sort of alias from `@rome/types` to `@joystream/types` published on npm ?

Exactly, it's an alias.

> Would it be possible to do something like:
> 
>     "@constantinople/types@npm:@joystream/types": "^0.8.0",
>     "@rome/types@npm:@joystream/types": "^0.7.0",

Yes, after .8 will be published to nexus

> We cannot merge with package.json having such a relative path to a dependency.
> How can we tackle this.

We can either publish .8 types, or I can put a placeholder here for the relative path.

tests/network-tests/src/tests/rome/romeRuntimeUpgradeTest.ts
gleb-urvanov: Indeed.

tests/network-tests/src/tests/rome/utils/utils.ts
gleb-urvanov: I will give it a try and will push the changes.

gleb-urvanov: I found no way to pass the `Buffer`, and changed the conversion to
```
u8aToHex(...)
```
But the object passed to API is still a hex string.

PR #323, opened 2020-04-29
Bureaucracy module prototype

shamil-gadelshin:
### Features:
- new instantiable bureaucracy module
- set_lead() extrinsic
- ensure_lead() method
- forum integration (removed old code and extrinsic)
- removed forum sudo from the genesis config
- no tests test(forum tests are broken)


runtime-modules/bureaucracy/src/types.rs
```diff
@@ -0,0 +1,50 @@
...
+        LeadRoleState::Active
+    }
+}
+
+/// Working group lead: curator lead
+/// For now this role is not staked or inducted through an structured process, like the hiring module,
+/// hence information about this is missing. Recurring rewards is included, somewhat arbitrarily!
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Debug, Clone, PartialEq)]
+pub struct Lead<AccountId, RewardRelationshipId, BlockNumber> {
```
bedeho: Needs a membership identifier to bind lead to some underlying member, this is what allows the lead to switch `role_account` if the wish to.

runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+        let new_lead_id = <NextLeadId<T, I>>::get();
+
+        let new_lead_role = ActorInRole::new(Role::CuratorLead, new_lead_id);
+
+        //
+        // == MUTATION SAFE ==
+        //
+
+        // Register in role - will fail if member cannot become lead
+        membership::members::Module::<T>::register_role_on_member(member, &new_lead_role)?;
```
bedeho: All this registration stuff goes away, as membership module will not be relied on to hold member <-> role binding information.

bedeho (changes requested):
- The coupling between the forum and the bureaucracy module seems needless. All the forum needs is the ability to do `<T as Trait>::is_account_lead(origin) bool`, which is implemented as a check that `origin` is signed with the `role_account` of the currently present lead, if set. The forum, which when done will be a very robust general-purpose Substrate forum, should not be encumbered by our governance abstraction.

- I think we should have this be a clean start, as the query node will be available before we put this in production, so I think we should drop all storage state that is purely archival as we move forward. This means
  - no need to store anything other than possibly the current set lead
  - no need to represent active vs exited leads
  - no need to store when a lead was inducted.
  - no need for lead ids

shamil-gadelshin:
I'm going to remove a check for a current lead existence:

```
        // Ensure there is no current lead
        ensure!(
            <CurrentLeadId<T, I>>::get().is_none(),
            MSG_CURRENT_LEAD_ALREADY_SET
        );
```

Because we don't use unset_lead().

shamil-gadelshin:
Should I also remove 

```
/// Type of minting reward relationship identifiers
pub type RewardRelationshipId<T> = <T as recurringrewards::Trait>::RewardRelationshipId;
```

Because we don't use it yet?

bedeho:
I suppose you could remove both, but both seem to be required down the road when we 

> I'm going to remove a check for a current lead existence:
> 
> ```
>         // Ensure there is no current lead
>         ensure!(
>             <CurrentLeadId<T, I>>::get().is_none(),
>             MSG_CURRENT_LEAD_ALREADY_SET
>         );
> ```
> 
> Because we don't use unset_lead().

We will have it later of course, the council must be able to remove a lead. More importantly, there may be no lead initially, hence the check still is appropriate. The storage representation should just be `current_lead: Option<Lead>`, which may or may not be set at any time.



> Should I also remove
> 
> ```
> /// Type of minting reward relationship identifiers
> pub type RewardRelationshipId<T> = <T as recurringrewards::Trait>::RewardRelationshipId;
> ```
> 
> Because we don't use it yet?

Sure you can remove it for now, but the lead will have to have the possibility of a reward at some point in the future. I am open to not using the recurring rewards module for that if you want.

shamil-gadelshin:
All done with the exception of "The coupling between the forum and the bureaucracy module".

If I understand properly we need this definition of the forum trait:

```
pub trait Trait:
    system::Trait + timestamp::Trait + bureaucracy::Trait<bureaucracy::Instance1> + Sized
```

In order to be able to call

```
    fn ensure_is_forum_lead(account_id: &T::AccountId) -> dispatch::Result {
        bureaucracy::Module::<T, bureaucracy::Instance1>::ensure_is_lead_account(account_id.clone())
    }
```

bedeho:
> reaucracy::Trait<bureaucracy::I

No, I don't think we even need this, there shouldn't need to be a reference any type in bureaucracy from the forum. The forum defines its own checker trait, and then there would be an integration code that binds it to whatever public method on the bureaucracy that allows the lookup.

Ironically, this is literally an `ensureOrigin` type lookup, as there is no need for a separate identifier, as there is only ever a single lead defined at any time.

shamil-gadelshin:
Ok. I understand now - you mean `sr_primitives::traits::EnsureOrigin`

I will provide the forum-bureaucracy binding in the _runtime_ crate.

shamil-gadelshin:
I implemented generic _EnsureOrigin_ in the bureaucracy, so no need for special integration struct for every bureaucracy instance.

bedeho:
Great! Merging, CI is complaining since forum tests are now broken, but its not clear that its worth fixing.

PR #325, opened 2020-04-30
Introducing new bureaucracy module

bedeho:


bedeho:
As long as we are using recurring rewards, we are allowing the lead to control how many workers the runtime has to iterate over because the lead currently faces no constraint on the maximum number workers the group can have.

I think we need to have such a limit, which can be set by the council, but even it also needs to be less than some hardcoded `bureaucracy` instance module parameter. To be clear, I think its fine for different instances to have different council set limits, and even hardcoded module limits. But there must be a limit.

Beyond this, when a worker is deactivated, we also need to remove such any reward relationship that may have existed. Currently, `recurring_reward::try_to_deactivate_relationship` is called, however, it only _deactivates_ the relationship, it still leaves it dangling (evil archiving!). The iteration will therefore still keep iterating over `RewardRelationships`, which grows perpetually, eventually, this blows up.

These errors are hard to detect, since there are many layers of indirection, and so the burden of being a good steward is not only not static but dissolves.

I want to suggest the following idea: What if we did not use recurring rewards and mints in these external modules, but instead absorbed that logic into this module? Reward recipients and relationships fall away, they just collapse into a `Worker`. There is just the one loop in `on_finalize`, no mints - no problem with a mint no existing, that iterates over all workers, and this set has no archiving. The main rationale for having the mint module was that you could have one mint directly transfer minting capacity to another ( `transfer_capacity`), simulating a dynamic hierarchy of budget constraints in a complex organization. This is however unlikely to ever need on the path we are on, and we are paying a substantial cost in complexity.

I am sure similar observations can be made about staking and hiring, as `bureaucracy` is in many ways a culmination of all of those. I am not suggesting we touch any other modules that currently use the mint + recurring rewards, such as the council, just to try our hand at not using them here for now.

Give this some thought, also input from @mnaamani could be helpful on this.

bedeho:
Another trivial point: I think the name of this module made more sense when it seemed like maybe it would be a substrate used in each working group module, but now it literally _is_ a working group, so should we give it a more direct name: `working_group`

bedeho:
I am having to write an integrated explanation of how a working group works for the handbook, this is why I keep pointing out these points: I am seeing we have the same security issue in `Hiring` and `Staking` indeed:

- `Hiring`: First off, I don't believe we have any limit on how many openings there can be that are active, hence the lead can just create an unlimited number of them. Even more subtly, when openings are finished, they are left in `Deactivated` stage. Both of these independently contribute to the iteration of `OpeningById` in `on_finalized` to service
  - beginning to accept applications
  - review period expiry

- `Stake`: Here there is the same issue. The lead can generate a large, possibly unbounded, number of stakes which are active, either because there is no application rationing (which lead controls), or there is rationing, but the unbonding is in the future, hence lots of stakes are still present in unstaking stage. There may be other channels I am missing. In the end, all these independently contribute to the iteration in `on_finalize`, which service unstaking and slashing.

Again, the worst problem here is not that we have to do something to fix this, its the opaqueness of the problem to begin with, which lead to similar problems being re-introduced down the line. 

I do wonder if the best thing is for us to have a version of this module which is free of archiving, iteration and usage of any of the existing modules (`Hiring`, `RecurringRewards`, `TokenMint`, `Stake`), and instead pull in a plain version of the business logic we need into this module.

PR #387, opened 2020-05-04
Update proposals parameters and limits.

### LINKED ISSUE #270: Change proposal parameters
Exact parameters defined here:
https://github.com/Joystream/substrate-runtime-joystream/issues/236

shamil-gadelshin:
Features:
- update shared parameters (rejection fee, cancellation fee, etc)
- update proposal type parameters (stakes)
- update proposal type limits (min-max validator count, role parameters etc)

### DIFF
--- a/runtime-modules/proposals/codex/src/lib.rs
+++ b/runtime-modules/proposals/codex/src/lib.rs
@@ -879,7 +879,7 @@ impl<T: Trait> Module<T> {
         );
 
         ensure!(
-            election_parameters.revealing_period <= T::BlockNumber::from(43200),
+            election_parameters.revealing_period <= T::BlockNumber::from(28800),
             Error::InvalidCouncilElectionParameterRevealingPeriod
         );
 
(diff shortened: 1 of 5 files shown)


bedeho (changes requested):
I guess this is an issue from a while back, but seems like a good time to fix:

- Can we avoid having any constants at all in actual code execution flow, and instead create suitable constants defined and used the way `CONTENT_WORKING_GROUP_MINT_CAPACITY_MAX_VALUE` is. I guess the cases in `parameters.rs` are the exception, which is OK I guess, but I suspect we will have to revisit how this is organized as the number of proposals and parameter grows. 

- Why are many parameter names of the form `Set*` or `set_`, this is very confusing, and very naturally lends itself to the interpration that a function with sideffects are being called when read in `parameters.rs`.

shamil-gadelshin:
- Ok. I will introduce named constants for all proposal limits
- Set_ or set_  prefixes in names are reflections of the related proposal, eg:
'set lead' proposal and `set_lead_proposal_grace_period`. All such names contain '_proposal_'  in the middle. Compare with `runtime_upgrade_proposal_grace_period`. We could remove such prefixes, but the naming would become inconsistent: 'lead_proposal_grace_period'. We can also invert the order of the words: `grace_period_of_set_lead_proposal`. It would probably break the dependent code like Gleb's tests.

bedeho:
I see, given that I suspect this is sort of storage is going away due to [this](https://github.com/Joystream/joystream/issues/391), let's leave it be then, in particular, if it requires synching integration test codebase.

bedeho (approved):

Issue #383, opened 2020-05-01
Task list


metmirr:
#### Indexer
- [x] Add information about last processed event #310 
- [x] Basic database operations (CRUD)
- [x] Use database transactions to make changes on database #310 
- [x] Start from a specific block number
- [x] Safe entity types for `db` module
- [x] Make mappings transaction safe. All database operations for a mapping should be done in a single database transaction

#### Mappings
- [x] Add another related event. One database table gets updates from multiple events.
- [ ] Unit tests

#### GraphQL
- [x] Change schema defination format from JSON to something more readable and compatible with TheGraph #313 
- [ ] Propagate description of fields and types to GraphQL
- [ ] Relationships between types

#### CLI
- [x] Add option to set last processed event to genesis (purging chain will cause error: `Provided block is ahead of chain.`)
- [ ] Update readme for requirement installation

#### Deployment
- [ ] Deployment for Indexer and GraphQL server #311

bedeho:
Superseded by https://github.com/orgs/Joystream/projects/15

PR #389, opened 2020-05-04
Query node

metmirr:
This pull request includes:

- Use database transactions to insert data into database
- Update last processed event on database
- Update`joystream-query-node/README.md` for new features [Add new mapping](https://github.com/metmirr/joystream/tree/query_node/query-node/joystream-query-node#add-a-new-mapping-for-joystream-memberregistered-event)

### DIFF
--- a/query-node/joystream-query-node/README.md
+++ b/query-node/joystream-query-node/README.md
@@ -47,27 +47,23 @@ export async function handleMemberRegistered(event: QueryEvent) {}
 4. Inside the handler function create a new instance of the entity and fill properties with event data.
 
 ```ts
-const member = new MemberRegistereds();
 // Get event data
 const { AccountId, MemberId } = event.event_params;
-
-member.accountId = AccountId;
-member.memberId = +MemberId;
+const member = new MemberRegistereds({ accountId: AccountId, memberId: +MemberId });
 ```
 
 5. Call `db.save()` method to save data on database
 
 ```ts
 // Save to database.
-db.save(member, event);
+db.save<MemberRegistereds>(member, event);
 ```
 
 6. Query database
 
 ```ts
 // Query from database
-const m = await db.get("MemberRegistereds", { memberId: MemberId });
-console.log(m);
+const m = await db.get(MemberRegistereds, { where: { memberId: 123 } });
 ```
 
 **Notes**
@@ -82,20 +78,16 @@ import { MemberRegistereds } from "../generated/indexer/entities/MemberRegistere
 import { QueryEvent, db } from "../generated/indexer/index";
 
 export async function handleMemberRegistered(event: QueryEvent) {
-  const member = new MemberRegistereds();
-
   // Get event data
   const { AccountId, MemberId } = event.event_params;
 
-  member.accountId = AccountId;
-  member.memberId = +MemberId;
+  const member = new MemberRegistereds({ accountId: AccountId, memberId: +MemberId });
 
   // Save to database.
-  db.save(member, event);
+  db.save<MemberRegistereds>(member, event);
 
   // Query from database
-  const m = await db.get("MemberRegistereds", { memberId: MemberId });
-  console.log(m);
+  const m = await db.get(MemberRegistereds, { where: { memberId: 123 } });
 }
 ```
 
--- a/query-node/joystream-query-node/README.md
+++ b/query-node/joystream-query-node/README.md
@@ -24,7 +24,82 @@ $ cd generated/indexer
 $ yarn start
 ```
 
-## Query Node Constructs Explained
+#### Add a new mapping for Joystream MemberRegistered event
+
+1. Every mapping function get a parameter of `QueryEvent` type
+
+```ts
+import { QueryEvent, db } from "../generated/indexer/index";
+```
+
+2. `db` object for database operations: save/get/remove
+
+```ts
+import { db } from "../generated/indexer/index";
+```
+
+3. Define the event handler function with the following signature
+
+```ts
+export async function handleMemberRegistered(event: QueryEvent) {}
+```
+
+4. Inside the handler function create a new instance of the entity and fill properties with event data.
+
+```ts
+const member = new MemberRegistereds();
+// Get event data
+const { AccountId, MemberId } = event.event_params;
+
+member.accountId = AccountId;
+member.memberId = +MemberId;
+```
+
+5. Call `db.save()` method to save data on database
+
+```ts
+// Save to database.
+db.save(member, event);
+```
+
+6. Query database
+
+```ts
+// Query from database
+const m = await db.get("MemberRegistereds", { memberId: MemberId });
+console.log(m);
+```
+
+**Notes**
+`db.save()` method take two parameters entity instance and event. `db.save()` method use `event` to keep track of last processed event.
+
+`db.get()` method. The first parameter is the name of the entity class to query and the second parameter is the find options (to match record).
+
+**Complete code**
+
+```ts
+import { MemberRegistereds } from "../generated/indexer/entities/MemberRegistereds";
+import { QueryEvent, db } from "../generated/indexer/index";
+
+export async function handleMemberRegistered(event: QueryEvent) {
+  const member = new MemberRegistereds();
+
+  // Get event data
+  const { AccountId, MemberId } = event.event_params;
+
+  member.accountId = AccountId;
+  member.memberId = +MemberId;
+
+  // Save to database.
+  db.save(member, event);
+
+  // Query from database
+  const m = await db.get("MemberRegistereds", { memberId: MemberId });
+  console.log(m);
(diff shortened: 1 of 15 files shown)


query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  constructor(connection: Connection) {
+    this._connection = connection;
+  }
+
+  /**
+   * Return the last processed event. If does not exists create a new one
+   */
+  async getLastProcessedEvent(event: QueryEvent): Promise<EventHistory> {
+    let eventHistory = await this._connection.getRepository(EventHistory).findOne();
```
bedeho: Will `findOne` always grab the last record?

query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -0,0 +1,57 @@
...
+    eventHistory.eventName = event.event_method;
+    eventHistory.blockNumber = event.block_number;
+    return eventHistory;
+  }
+
+  /**
+   * Fixes compatibility between typeorm and warthog models
+   * @param entity
+   */
+  fillRequiredWarthogFields(entity: any) {
```
bedeho: What is going on here? I did not understand this.

Also, this is not really a proper object method, this seems like a static method at best, if not a standalone function, but its hard to say given that its not clear to me what is going on.

query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -0,0 +1,57 @@
...
+    const eventHistory = await this.getLastProcessedEvent(event);
+
+    await this._connection.manager.transaction(async (manager: EntityManager) => {
+      await manager.save(entity);
+      await manager.save(eventHistory);
+    });
+  }
+
+  // Find entity that match find options
+  async get(entityName: string, options: {}) {
```
bedeho: Is this the only way to access the database, its not type safe, which sort of defeats the purpose of all the generated type classes.

query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -0,0 +1,57 @@
...
+    entity['id'] = shortid.generate();
+    entity['createdById'] = shortid.generate();
+    entity['version'] = 1;
+  }
+
+  /**
+   * Save given entity instance, if entity is exists then just update
+   * @param entity
+   */
+  async save(entity: any, event: QueryEvent) {
```
bedeho: having `entity: any` seems too loose given that we already know what types can arrive here at compile time. Perhaps a generic or something should be used here and instantiated with an enum across all proper entity types when used in the generated code.

query-node/substrate-query-framework/index-builder/src/db/entities.ts
```diff
@@ -0,0 +1,18 @@
+import { Entity, Column, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class EventHistory {
```
bedeho: This name is not very good, really this entity represents that entity was saved to the database as a result of processing an event. So a better name could be `SavedEntityEvent`.

query-node/substrate-query-framework/index-builder/src/db/entities.ts
```diff
@@ -0,0 +1,18 @@
+import { Entity, Column, PrimaryColumn } from 'typeorm';
+
+@Entity()
```
bedeho, line 8: Please add comments to type and properties.

query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -0,0 +1,57 @@
+import * as shortid from 'shortid';
+import { Connection, EntityManager } from 'typeorm';
+import { EventHistory } from '.';
+import { QueryEvent } from '..';
+
+export default class DB {
```
bedeho, line 15: Please add comments to type.

query-node/joystream-query-node/README.md
```diff
@@ -24,6 +24,81 @@ $ cd generated/indexer
...
+
+**Notes**
+`db.save()` method take two parameters entity instance and event. `db.save()` method use `event` to keep track of last processed event.
+
+`db.get()` method. The first parameter is the name of the entity class to query and the second parameter is the find options (to match record).
+
+**Complete code**
+
+```ts
+import { MemberRegistereds } from "../generated/indexer/entities/MemberRegistereds";
```
bedeho, line 72: This example needs to be tweaked a bit.

First off, we really want the query to be for memberships, not registration events. We could also want to index registration events, but I would say that is secondary. So the name of the type should be `Membership`. We process the `MembershipRegistered` (not `MemberRegistereds` as you have written) runtime event to add new entities to the `Membership` table. For completeness, you may as well include one other related event, for example, one that has to do with editing the profile of an event, or something. That way we have two separate events that have to do with providing one query into the memberships on the platform.

query-node/joystream-query-node/README.md
```diff
@@ -24,6 +24,81 @@ $ cd generated/indexer
...
+import { QueryEvent, db } from "../generated/indexer/index";
+
+export async function handleMemberRegistered(event: QueryEvent) {
+  const member = new MemberRegistereds();
```
bedeho: Does the `MemberRegistereds` not have a proper constructor?

query-node/joystream-query-node/README.md
```diff
@@ -24,6 +24,81 @@ $ cd generated/indexer
...
+  const { AccountId, MemberId } = event.event_params;
+
+  member.accountId = AccountId;
+  member.memberId = +MemberId;
+
+  // Save to database.
+  db.save(member, event);
+
+  // Query from database
+  const m = await db.get("MemberRegistereds", { memberId: MemberId });
```
bedeho: Let's drop such logging here, it's not good practice.

bedeho (changes requested):
I think this was a nice step in the right direction.

1. My understanding is that you are storing every single processed event in the database, in order to always know which is the most recent one. I think that is fine, the disadvantage is of course that the storage burden grows monotonically, regardless of what is actually happening in the chain. This is not the case for normal processing, but I think we can move forward with this. Its also a bit awkward that in mappings which update multiple tables, the same`EventHistory` instance is added multiple times. I do not think there is an actual bug, but its a bit odd. Are you sure there is no possibility of failure in such cases?

2. Rather than having the database instance `db` in the global scope, let's pass a reference to the database into each mapping. This has at least two benefits:

  - It makes it possible to mock the behaviour of `db` during unit testing of each mapper.
  - It removes the need for each mapper to pass in the event, which could be a source of error, instead, the provided reference could already have the correct event in scope for use with the transaction.

3. How do you generalize this to work with other forms of database mutation, such as deleting an entity, or editing an entity already in the database? I guess editing could work by just reading and then overwriting? is that the idea?

4. Where is the code that actually reads the database to figure out where to start synching when starting the indexer node?

query-node/substrate-query-framework/index-builder/src/db/db.ts
metmirr: Yes, it will. There is only one record in the table.

query-node/substrate-query-framework/index-builder/src/db/db.ts
metmirr: Database models are generated with warthog CLI. Every generated model extends warthog's `BaseModel` which has default these. And these fields are required (btw I also don't like it).

I agree it could be a static or better a standalone utility function.

metmirr:
1. No, it is not storing all processed events. `getLastProcessedEvent` overwrite the existing event if exists otherwise create a new event. There is only one record on the database.
2. That would be better. Especially for the `db`, I resisted to not add the event as parameter to `db.save()` but couldn't any find a better approach.
3. Since our models are auto-generated and we use database transactions using Data Mapper pattern is good for us. As you said editing is reading the entity from the database and then overwriting. I am using Data Mapper pattern which is a standard (read/remove/save). 
> In TypeORM you can use both the Active Record and the Data Mapper patterns https://typeorm.io/#/active-record-data-mapper. 
4. I haven’t implemented it yet.

query-node/joystream-query-node/README.md
metmirr: No, it doesn't have. But I can add it. By adding an option to `codegen` command.

bedeho (approved):
Looks good!

I see that the underlying transaction issue involving multiple saves still is outtanding, where is this and other work in progress tracked?

metmirr:
> Looks good!
> 
> I see that the underlying transaction issue involving multiple saves still is outtanding, where is this and other work in progress tracked?

Tasklist is here https://github.com/Joystream/joystream/issues/383

PR #395, opened 2020-05-06
Content dir 2

bedeho:


bedeho:
Superceeded by https://github.com/Joystream/joystream/pull/396

Issue #260, opened 2020-04-01
Batch transaction sending fails


gleb-urvanov:
Batch sending transactions (balance.transfer and others) fails on joystream node,but works for substrate node template. 

The following code
```
    let nonce = new BN(0);
    const keyring = new Keyring({ type: 'sr25519' });
    const sudo = keyring.addFromUri('//Alice');
    const nKeyPairs: KeyringPair[] = new Array();
    for (let i = 0; i < 10; i++) {
      nKeyPairs.push(keyring.addFromUri('//Bob'));
    }
    await Promise.all(
      nKeyPairs.map(async keyPair => {
        this.api.tx.balances.transfer(keyPair.address, new BN(100)).signAndSend(sudo, { nonce: nonce });
        nonce = nonce.addn(1);
      })
    );
```
Will result in first 2 transactions to be processed and included to blocks, one per block. The rest 8 are removed from the tx pool with the following error message seen for each:
`Extrinsic invalid: 0xf1368a77daa999b89984e37bc45d9ecd95b93d42fce0f601247c6d94299f1ef2`

The same code, run against the substrate node (build from substrate `c37bb0853`) will result in all 10 transactions processed and included in a single block.

Similar transactions, sent consecutively (after previous one is included in block) will succeed both on joystream and substrate nodes.

gleb-urvanov:
The issue also happens if the destination and the amount to be sent are not the same for batch transactions.
The issue is not caused by tx processing time, the code was tested on more powerful hardware and with increased block construction time, the result is the same (two transactions processed, one per block).
The issue does not happen if the transactions are sent from different accounts. In this case, several transactions are included in single block, as expected.
During the testing several times I was able to receive tx status `status.isFuture` for the nonce value `2`, while sending batch of more than 2 transactions with nonces 0, 1, 2.. N. 
Testing with `RUST_BACKTRACE` and with the node built in debug mode gave me no additional information.

mnaamani:
I finally got around to testing this.

I can confirm that I'm able to replicate the issue as described. It is appearing only for the latest joystream-node release, and not the substrate node built from upstream repo from same commit release. The problem also doesn't exist for Acropolis release (prior version of joystream-node)

So there is a clearly an implementation issue in our node itself. Will investigate further.

mnaamani:
Also @gleb-urvanov, clearly my suspicion that there was a bug in the code snippet turned out to be false.

[...]

PR #397, opened 2020-05-07
Increase the MaximumBlockWeight

mnaamani:
The maximum block weight was too small. 
This explains the issue https://github.com/Joystream/joystream/issues/260 being identified.

The issue wasn't as serious since most of joystream's extrinsics are not weighted

This PR Increases the MaximumBlockWeight, to be 1000x times larger - this allows approximately  750 balances::transfer transactions per block.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -176,7 +176,7 @@ pub fn native_version() -> NativeVersion {
 
 parameter_types! {
     pub const BlockHashCount: BlockNumber = 250;
-    pub const MaximumBlockWeight: Weight = 1_000_000;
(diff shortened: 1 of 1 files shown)


shamil-gadelshin (approved):
The substrate runtime node (of the version we use) contains the same value.

shamil-gadelshin:
The current version of the substrate runtime contains even 1000x higher default value.

mnaamani:
> The substrate runtime node (of the version we use) contains the same value.

Yes but only in the node-template runtime. In the main node implementation it is set to the higher value. I suspect I may have initially copied the value from the node-template runtime.

bedeho (approved):

PR #386, opened 2020-05-04
build script for network tests added, readme updated

gleb-urvanov:

PR #322, opened 2020-04-29
Feature/new proposals tests

gleb-urvanov:

PR #399, opened 2020-05-07
Feature/proposals test coverage

gleb-urvanov:
Depends on:
https://github.com/Joystream/joystream/pull/284
https://github.com/Joystream/joystream/pull/322
https://github.com/Joystream/joystream/pull/386

Require Constantinople joystream types (https://github.com/Joystream/apps/tree/joystream-constantinople)


tests/network-tests/package.json
```diff
@@ -4,17 +4,20 @@
...
     "build": "tsc --build tsconfig.json",
-    "test": "mocha -r ts-node/register src/tests/*",
+    "test": "mocha -r ts-node/register src/tests/constantinople/proposals/*",
+    "test-migration": "mocha -r ts-node/register src/tests/rome/* && mocha -r ts-node/register src/tests/constantinople/*",
     "lint": "tslint --project tsconfig.json",
     "prettier": "prettier --write ./src"
   },
   "dependencies": {
-    "@joystream/types": "^0.7.0",
+    "@joystream/types": "../joystream-apps/packages/joy-types",
```
mnaamani:
```suggestion
    "@joystream/types": "^0.9.1",
```
Final constantinople release

mnaamani: We might as well also use alias `@constantinople/types` ?

gleb-urvanov: done

scripts/run-test-chain.sh
```diff
@@ -0,0 +1,13 @@
...
+sed -i 's/"setValidatorCountProposalGracePeriod":.*/"setValidatorCountProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"runtimeUpgradeProposalGracePeriod":.*/"runtimeUpgradeProposalGracePeriod": 0,/' ./chainspec.json
+sed -i 's/"setElectionParametersProposalGracePeriod":.*/"setElectionParametersProposalGracePeriod": 0,/' ./chainspec.json
+sed -i 's/"textProposalGracePeriod":.*/"textProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"setContentWorkingGroupMintCapacityProposalGracePeriod":.*/"setContentWorkingGroupMintCapacityProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"setLeadProposalGracePeriod":.*/"setLeadProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"spendingProposalGracePeriod":.*/"spendingProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"evictStorageProviderProposalGracePeriod":.*/"evictStorageProviderProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"setStorageRoleParametersProposalGracePeriod":.*/"setStorageRoleParametersProposalGracePeriod": 0/' chainspec.json
+yes | cargo run --release -p joystream-node -- purge-chain --dev
```
mnaamani, line 13:
```suggestion
yes | cargo run --release -p joystream-node -- purge-chain --chain=chainspec.json
```

scripts/run-test-chain.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+cargo run --release -p joystream-node build-spec --chain dev > chainspec.json
+sed -i 's/"setValidatorCountProposalGracePeriod":.*/"setValidatorCountProposalGracePeriod": 0,/' chainspec.json
+sed -i 's/"runtimeUpgradeProposalGracePeriod":.*/"runtimeUpgradeProposalGracePeriod": 0,/' ./chainspec.json
```
mnaamani:
```suggestion
sed -i 's/"runtimeUpgradeProposalGracePeriod":.*/"runtimeUpgradeProposalGracePeriod": 0,/' chainspec.json
```

scripts/run-test-chain.sh
```diff
@@ -0,0 +1,13 @@
...
+sed -i 's/"setElectionParametersProposalGracePeriod":.*/"setElectionParametersProposalGracePeriod": 0,/' ./chainspec.json
```
mnaamani:
```suggestion
sed -i 's/"setElectionParametersProposalGracePeriod":.*/"setElectionParametersProposalGracePeriod": 0,/' chainspec.json
```

scripts/run-test-chain.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+cargo run --release -p joystream-node build-spec --chain dev > chainspec.json
```
mnaamani: Although this approach is temporary and eventually we will have parameterized build of the runtime, and producing this chainspec file as an artifact of running tests will not be necessary, it should be generated in some temporary location folder, maybe the system temporary path? 

This goes for any other temporary artifacts of course.

tests/network-tests/src/constantinople/tests/impl/electingCouncil.ts
```diff
@@ -1,13 +1,13 @@
-import { membershipTest } from './membershipCreationTest';
 import { KeyringPair } from '@polkadot/keyring/types';
-import { ApiWrapper } from './utils/apiWrapper';
+import { ApiWrapper } from '../../utils/apiWrapper';
 import { WsProvider, Keyring } from '@polkadot/api';
-import { initConfig } from './utils/config';
-import BN = require('bn.js');
```
mnaamani, line 6: I'm surprised this syntax ever worked in the first place! :)

tests/network-tests/src/constantinople/tests/impl/electingCouncil.ts
```diff
@@ -1,13 +1,13 @@
...
-import { registerJoystreamTypes, Seat } from '@joystream/types';
+import { initConfig } from '../../utils/config';
+import BN from 'bn.js';
```
mnaamani, line 4: I've often had typescript linter complain that `bn.js` doesn't have a default export and so in this import `BN` would be undefined. Has this been fixed in newer versions of `bn.js`?

tests/network-tests/package.json
```diff
@@ -4,16 +4,16 @@
   "license": "GPL-3.0-only",
   "scripts": {
     "build": "tsc --build tsconfig.json",
-    "test": "mocha -r ts-node/register src/tests/constantinople/*",
-    "test-migration": "mocha -r ts-node/register src/tests/rome/* && mocha -r ts-node/register src/tests/constantinople/*",
+    "test": "tap --files ts-node/register src/constantinople/tests/proposals/*Test.ts",
```
mnaamani, line 7: We don't seem to have a way to control the order of tests, as far as I can see they are run in alphabetic order of the file names (not different than mocha).

Given that the test scenarios should be independent, (currently they are all executed one after the other with the same chain so they aren't exactly isolated) the order is not important, but it would still be valuable to control.

The plan of course is to have a better way to setup a network before running the test scenarios, so this is probably the task of the orchestrator/task runner/

tests/network-tests/src/constantinople/tests/impl/electingCouncil.ts
```diff
@@ -17,18 +17,19 @@ export function councilTest(m1KeyPairs: KeyringPair[], m2KeyPairs: KeyringPair[]
   const K: number = +process.env.COUNCIL_ELECTION_K!;
   const greaterStake: BN = new BN(+process.env.COUNCIL_STAKE_GREATER_AMOUNT!);
   const lesserStake: BN = new BN(+process.env.COUNCIL_STAKE_LESSER_AMOUNT!);
-  const defaultTimeout: number = 120000;
+  const defaultTimeout: number = 300000;
```
mnaamani: I don't understand why the timeout had to be increased here, and in some other tests also.

Wouldn't it be better if the timeout value for each scenario was derived based on how many finalized blocks the tests should be completed within, and the chain's block interval?

Its also not clear, is this timeout for the entire suite or per `.test()`? If it is per test then the timeout should be specific to each test, not the max value to accommodate all the tests.

tests/network-tests/src/constantinople/tests/impl/membershipCreation.ts
```diff
@@ -15,15 +16,16 @@ export function membershipTest(nKeyPairs: KeyringPair[]) {
   const paidTerms: number = +process.env.MEMBERSHIP_PAID_TERMS!;
```
mnaamani: For all the functions inside the `impl` folder which are now the reusable unit, it might be better to have these scenario parameters be passed as arguments and the caller would read them from `process.env`

`registerJoystreamTypes()` is also called in each test setup, and I think the caller of the test only needs to do that one time, and might as well also create the api instance and pass that as argument as well.

tests/network-tests/src/constantinople/tests/proposals/impl/electionParametersProposal.ts
```diff
@@ -0,0 +1,131 @@
...
+    const votingPeriod: BN = await apiWrapper.getVotingPeriod();
+    const revealingPeriod: BN = await apiWrapper.getRevealingPeriod();
+    const councilSize: BN = await apiWrapper.getCouncilSize();
+    const candidacyLimit: BN = await apiWrapper.getCandidacyLimit();
+    const newTermDuration: BN = await apiWrapper.getNewTermDuration();
+    const minCouncilStake: BN = await apiWrapper.getMinCouncilStake();
+    const minVotingStake: BN = await apiWrapper.getMinVotingStake();
+
+    // Proposal stake calculation
+    const proposalStake: BN = new BN(200000);
```
mnaamani, line 43: There are some runtime configured minimum values for stake, we should probably read the value instead of picking a magic number?

tests/network-tests/src/constantinople/tests/proposals/impl/electionParametersProposal.ts
```diff
@@ -0,0 +1,131 @@
...
+    const newAnnouncingPeriod: BN = await apiWrapper.getAnnouncingPeriod();
+    const newVotingPeriod: BN = await apiWrapper.getVotingPeriod();
+    const newRevealingPeriod: BN = await apiWrapper.getRevealingPeriod();
+    const newCouncilSize: BN = await apiWrapper.getCouncilSize();
+    const newCandidacyLimit: BN = await apiWrapper.getCandidacyLimit();
+    const newNewTermDuration: BN = await apiWrapper.getNewTermDuration();
+    const newMinCouncilStake: BN = await apiWrapper.getMinCouncilStake();
+    const newMinVotingStake: BN = await apiWrapper.getMinVotingStake();
+    assert(
+      announcingPeriod.subn(1).eq(newAnnouncingPeriod),
```
mnaamani: A better approach here that will also avoid some mistakes and doing the correct assertions:

`const proposedAnnouncingPeriod = announcingPeriod.subn(1);`

then propose that value, so assertion becomes

`assert(newAnnouncingPeriod.eq(proposedAnnouncingPeriod))`

tests/network-tests/src/constantinople/tests/proposals/impl/evictStoraveProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+    const description: string = 'Testing validator count proposal ' + uuid().substring(0, 8);
+    const runtimeVoteFee: BN = apiWrapper.estimateVoteForProposalFee();
+    await apiWrapper.transferBalanceToAccounts(sudo, m2KeyPairs, runtimeVoteFee);
+    if (!(await apiWrapper.isStorageProvider(sudo.address))) {
+      await apiWrapper.createStorageProvider(sudo);
+    }
+    assert(await apiWrapper.isStorageProvider(sudo.address), `Account ${sudo.address} is not storage provider`);
+
+    // Proposal stake calculation
+    const proposalStake: BN = new BN(25000);
```
mnaamani, line 40: Again it would be better to use the required stake from runtime.

mnaamani (changes requested):
So overall I think the setup seems better with tap, but I left a bunch of suggestions and questions too.

I was able to run the test but I had to manually edit the chainspec.json file because the run-test-chain.sh script didn't work.

```
 FAIL  src/constantinople/tests/proposals/updateRuntimeTest.ts
 ✖ timeout!

  test: "  Upgrading the runtime test"
  expired: TAP

 FAIL  src/constantinople/tests/proposals/updateRuntimeTest.ts 1 failed of 12 15m
 ✖ timeout!
```

I couldn't really figure out why it failed.

The test in total took quite a while to run, surely with control of the elections and setting grace periods to 0 this should have been much shorter?

```
Suites:   1 failed, 8 passed, 9 of 9 completed
Asserts:  1 failed, 107 passed, of 108
Time:     38m
```

I think the helper methods for awaiting for Event to be emitted by transactions could be improved maybe to wait for a specific amount of time/blocks before throwing so maybe the test can pickup on failure of what event expectation caused the failure.

mnaamani:
> ```
>  FAIL  src/constantinople/tests/proposals/updateRuntimeTest.ts
>  ✖ timeout!
> 
>   test: "  Upgrading the runtime test"
>   expired: TAP
> 
>  FAIL  src/constantinople/tests/proposals/updateRuntimeTest.ts 1 failed of 12 15m
>  ✖ timeout!
> ```
> 
> I couldn't really figure out why it failed.

I realize now, the reason it failed is because the `Migrated` event would not have been raised because I didn't prepare one with a higher spec version than runtime of the test chain before updating it.

scripts/run-test-chain.sh
gleb-urvanov: done

scripts/run-test-chain.sh
gleb-urvanov: done

scripts/run-test-chain.sh
gleb-urvanov: done, `.tmp` folder introduced

tests/network-tests/src/constantinople/tests/impl/electingCouncil.ts
gleb-urvanov: The timeout time now is defined using number of blocks required for the test to pass.

tests/network-tests/src/constantinople/tests/impl/membershipCreation.ts
gleb-urvanov: Scenario parameters now read from environment in scenarios and are passed to the implementation as parameters.

Type registration and api instance are still test-specific. I've run into issues while trying move them up to the scenario level. Let me know if we should discuss it.

tests/network-tests/src/constantinople/tests/proposals/impl/electionParametersProposal.ts
gleb-urvanov: There are no way to read the required stake values from API. The definition of the values is here: https://github.com/Joystream/joystream/blob/development/runtime-modules/proposals/codex/src/proposal_types/parameters.rs

tests/network-tests/src/constantinople/tests/proposals/impl/electionParametersProposal.ts
gleb-urvanov: done

PR #401, opened 2020-05-07
WIP2: discovery node

dzhelezov:
This is a follow-up PR for the discovery node. Features to be implemented

- [x]  Bootstrap the members from Genesis using api calls on a first run
- [x]  Create basic indices in ElasticSearch on a first run
- [x]  Restore state from ElasticSearch on subsequent runs
- [x]  Upload the produced `QueryEventBlock`s to an ElasticSearch index


discovery_node/substrate-ingest/src/joynode/QueryNode.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Query index building node.
+    //private _indexBuilder: IndexBuilder;
+
+    constructor() {
+        super();
+        this._state = QueryNodeState.NOT_STARTED;
+    }
+
+    async build(config: Config) {
+        let providerUrl = config.get()?.joystream?.ws_provider || process.env.JOYSTREAM_PROVIDER;
```
bedeho, line 49: I think you need some guard to prevent build from being invoked multiple times on the same object, more deeply, if this is meant to be used to construct a single node one time, it could make sense to make this a static factory method.

discovery_node/README.md
```diff
@@ -0,0 +1,3 @@
+# Discovery node
```
bedeho, line 1: - Can we get the top-level README in shape to cover at least
  - what this is.
  - how you get it installed/built.
  - how you run it.
  - how you stop it.

discovery_node/substrate-ingest/src/Config.ts
```diff
@@ -0,0 +1,42 @@
+import { config } from 'dotenv';
+import * as fs  from 'fs'
+import * as path from 'path'
+import * as yaml from 'js-yaml'
+import log4js = require('log4js');
+
+
+export default class Config {
```
bedeho, line 8: Can we get a typesafe version of this going, so every property that can configured exists as an explicit property of this class, rather than `getProp`, which may return `undefined` for expected value, or not tell you about property name typos at compile time, and also requires type casting from `string` to whatever you actually want every time you read.

discovery_node/substrate-ingest/src/joynode/QueryNode.ts
```diff
@@ -0,0 +1,95 @@
...
+            throw Error('Query block producer is not initialized');
+        }
+
+        //this._queryBlockProducer.on('QueryEventBlock', (query_event_block: QueryEventBlock): void => {
+        //    this._onQueryEventBlock(query_event_block);
+        //});
+
+        this._state = QueryNodeState.STARTED;
+        logger.debug(`Starting at state ${state}`);
+        await this._queryBlockProducer.start(state.lastProcessedBlock + 1);
```
bedeho: We have to assume that faults can occur where the processing can stop at any time. In that case you will most of the time not want to start in the next block, unless the last event you processed was the last event in `state.lastProcessedBlock`, which is the exception, you will want to process the next event in that block in most cases.

discovery_node/substrate-ingest/src/joynode/StateBootstrap.ts
```diff
@@ -0,0 +1,58 @@
...
+const DEFAULT_BOOTSTRAP_HEIGHT = 0;
+
+export interface JoyEntity {
+    entity: Object,
+    type: string,
+    blockNumber: number,
+    id: any
+}
+
+export default class StateBootstrap extends EventEmitter {
```
bedeho, line 21: This does not appear to be a genuine class to me, as there is no state that is continuously mutated and maintained. Hence if it's a pure function, let's make it a function, it's safer and easier on the reader.

discovery_node/substrate-ingest/src/state/StateKeeper.ts
```diff
@@ -0,0 +1,38 @@
+import Config from '../Config'
+
+const logger = require('log4js').getLogger('state-keeper');
+
+export interface State {
```
bedeho, line 5: - My understanding is that you are processing all events simultaneously, but its unclear to me whethr you are intending for this to be the case going forward. I would expect that this is not possible, but if it some how was, then I guess the `eventIndex` can be dropped.

- Can we find a narrower name for this to help the reader? Really what this represents is really the index of an event in the history of the chain, as opposed to `eventIndex`, which is the index within a given block.

discovery_node/substrate-ingest/node_modules/rxjs/src/internal/config.ts
```diff
@@ -0,0 +1,35 @@
+let _enable_super_gross_mode_that_will_cause_bad_things = false;
```
bedeho: I assume this should not be committed, given that it lives in `node_modules` ?

discovery_node/substrate-ingest/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+async function main() {
+    command
+        .version('0.0.1')
+        .description("Joystream ElasticSearch import tool")
+        .option('-c, --config <file>', 'path to config', 'config.yml')
+        .option('-d --debug', 'debug')
+        .parse(process.argv);
+
+    const config = new Config(command.config);
+    // TODO: move to a sep init class, catch init errors
```
bedeho, line 26: I agree with this TODO, the setup code here is non-trivial, and not easily testable.

bedeho (changes requested):
Overall, this looks like a really good start, very impressed at the speed you have been able to have without any experience with ES.

- If I could request one overarching adjustment in style, it would be to more strongly trade off sussinctness in favour of clarity through simplicity. There are many parts of this initial code base where it takes susbtantial work to gain conviction that it will always work. Our highest priority for all of our code is that a poorly organized and resource constrained open community can maintain and improve everything we are building now.

- I think you referred to introducing a dependency injection library in one of our conversations. I certainly think that there are many cases here where classes are constructing their depenencies, which will prevent robust unit testing. If you agree, then it would be good to add it ask a task on the long-running PR here: https://github.com/Joystream/joystream/pull/413

- Can we have a two-line description of every major type class, including properties and methods, it makes it much faster for the reviewer or reader to develop the right mental model when getting into the code-base.

- My understanding of the current approach is that there are basically two indexes, one for all events and one for all membership profiles. Only the latter has a schema for documents that go in the index. It seems that only the event index is updated during normal block processing, while the membership index is built up during the initial phase, called bootstrapping, and not maintained directly later.

  - Please correct my understaning if anything is wrong in that summary.

  - My impression is that you are conceiving of an approach where the ingester does not understand any of the semantics carried by an event and the corresponding set of parameters. I say this based on the generalized indexing of events, and interpreting the construction with `JoyUnit` and the mapping dictionary. I do not believe this is feasible. The presence of a particular event and its parameters has no direct releance to the index in principle, it only instructs us how we may have to mutate a set of documents, possibly across multiple indexes. For example, an event like `DeletedContent(id)` has no direct value to be indexed itself. Now, it may be the case that you are in perfect agreement with this, but in that case I think there are very substantail parts of this code base that does not warrant direct review at this stage, and I will wait until you have the basic structure for this.

- Why has the ingester become poll based? This should not be needed, nor is it desirable.

discovery_node/substrate-ingest/src/state/StateKeeper.ts
dzhelezov: Currently the events are "processed" in batch as they are simply bulk-uploaded to ElasticSearch (so no processing really happens). In the future, each event would require individual processing, and that's why the index within block looks like a reasonable representation of what is the current state of the process. I updated the names to better reflect that. `StateKeepers` is assumed to be the place where the all three ETL parts are coordinated. Currently there's no Transform part, and the Load part is wired directly to the Extract part (through async generators).

discovery_node/substrate-ingest/node_modules/rxjs/src/internal/config.ts
dzhelezov: Oops, corrected

discovery_node/substrate-ingest/src/joynode/StateBootstrap.ts
dzhelezov: Probably move it to `QueryNode` then? My thinking was to isolate the initial state bootstrapping (which requires separate and enitity-specific API calls) from the rest of the event ingestion. Since the block producer is one level beneath the query node class, it seemed natural to move this logic into a separate class as well. Perhaps an even cleaner approach is to further extract an interfaces from `QueryNode`.

discovery_node/substrate-ingest/src/joynode/QueryNode.ts
dzhelezov: Agree, `StateKeeper` is thought to be responsible for the fail-safe behaviour. In the recovery mode, it will restart the producer from the last fully processed blocks and simply skip the already processed events in the next block if necessary.

discovery_node/README.md
dzhelezov: Added

bedeho:
What is the status here of this PR? Should this not be merged while we explore the alternative approache(s)? If it is, then there are still some review comments that have not been adressed I believe, e.g. `Why has the ingester become poll based? This should not be needed, nor is it desirable.`.

PR #418, opened 2020-05-12
Query node

metmirr:
This PR contains work:

- New schema format which is actually a GraphQL schema
- Update CLI to use new schema format for GraphQL server code generation
- Add `Membership` as a new type to joystream schema
- Add a new mapping `MemberUpdatedAboutText`


bedeho (commented):
- This looks really nice for the end-user, glad we did this.

-  Did  you write the code in `ModelCodeGenerator.ts `, `SchemaParser.ts`, I ask because it looks like generic parsing code, and I don't see that you have added any new dependency for doing this parsing. If  I am correct, then why are you copying it in, as opposed to using a library?

- The `s` at the end is really rough on understanding this for new people, did you get in touch with the developer to turn this off?

metmirr:
>     * Did  you write the code in `ModelCodeGenerator.ts `, `SchemaParser.ts`, I ask because it looks like generic parsing code, and I don't see that you have added any new dependency for doing this parsing. If  I am correct, then why are you copying it in, as opposed to using a library?

Yes, I wrote those two modules and I am using [graphql-js](https://github.com/graphql/graphql-js) for parsing the schema. I use `graphql-js` as much as I can to avoid writing a lot of code. 

We are using warthog as a dependency and warthog depends on graphql-js. That is why I didn't add graphql-js as a dependency. But perhaps I should.
 
>     * The `s` at the end is really rough on understanding this for new people, did you get in touch with the developer to turn this off?

No, I did not. I will get in touch not only for this also for the relationship between types.

query-node/joystream-query-node/mappings/members.ts
```diff
@@ -1,3 +1,32 @@
...
+  // Get event data
+  const { AccountId, MemberId } = db.event.event_params;
+
+  let member = new Memberships({ accountId: AccountId.toString(), memberId: +MemberId });
+
+  // Save to database.
+  db.save<Memberships>(member);
+
+  // Query from database
+  member = await db.get(Memberships, { where: { memberId: MemberId } });
```
bedeho: Can we drop this from the example, as one would not want to do this in a normal mapping.

query-node/joystream-query-node/mappings/members.ts
```diff
@@ -1,3 +1,32 @@
-export function handleMemberRegistered() {
-	console.log(handleMemberRegistered.name);
+import { Memberships } from '../generated/indexer/entities/Memberships';
+import { DB } from '../generated/indexer';
+
+export async function handleMemberRegistered(db: DB) {
+  // Get event data
+  const { AccountId, MemberId } = db.event.event_params;
+
+  let member = new Memberships({ accountId: AccountId.toString(), memberId: +MemberId });
```
bedeho: - Why is `+MemberId` used?
- What type is `MemberId`, given that its `u64`, which can encode a larger integer than `Number` in typescript, which is a signed double-precision 64-bit, with `MAX_SAFE_INTEGER = 9007199254740991`.

query-node/joystream-query-node/mappings/members.ts
```diff
@@ -1,3 +1,32 @@
...
+
+export async function handleMemberUpdatedAboutText(db: DB) {
+  // Get event data
+  const { MemberId } = db.event.event_params;
+
+  // Query from database since it is an existsing user
+  const member = await db.get(Memberships, { where: { memberId: MemberId } });
+
+  // Make sure member exists
+  if (member) {
```
bedeho: Can you assert that `member` is defined, and not do an if branch that may ignore the failure, because if it isn't, then either the blockchain has a bug, or the query node processor has a bug, it is not a potentially valid event in this state.

query-node/joystream-query-node/schema.graphql
```diff
@@ -0,0 +1,16 @@
+type Membership {
+  # Member's root account id
+  accountId: String!
```
bedeho, line 3: How is the primary key in database defined from this schema?

query-node/substrate-query-framework/cli/src/helpers/ModelCodeGenerator.ts
```diff
@@ -0,0 +1,158 @@
+import { ObjectTypeDefinitionNode, FieldDefinitionNode, ListTypeNode, NamedTypeNode } from 'graphql';
+import { GraphQLSchemaParser } from './SchemaParser';
+
+// Available types for model code generation
+export const availableTypes: { [key: string]: string } = {
+  String: '',
```
bedeho, line 6: Is it here possible to introduce your own types, with new names that map to custom Typescript types that you have written yourself? I ask, because this could allow us to introduce support for encodings we get from the runtime, like `u128`, etc.

bedeho (commented):
Mostly just questions, some minor changes.

query-node/joystream-query-node/schema.graphql
metmirr: It's a good question. Here is the flow:

Schema parser creates _model definitions_ from the schema and then warthog cli use these model definitions to create models.
Warthog cli only understands `columnName:columnType` pattern, this means we can't decide to make a column unique or primary key at this point. And by default warthog adds `id` to each model.  

Here is an example of the warthog cli model generation:

`warthog generate Membership accountId! memberId:int!`

query-node/substrate-query-framework/cli/src/helpers/ModelCodeGenerator.ts
metmirr: Yes, it is possible. But here again, we face warthog's model generation limitation. Warthog's `codegen` command only support these types for model generation.

query-node/joystream-query-node/mappings/members.ts
metmirr: Done in https://github.com/Joystream/joystream/pull/418/commits/e35594ca09ca47d7a2eb93d8e7e4b6a1f8c21b91

query-node/joystream-query-node/mappings/members.ts
metmirr: Done in https://github.com/Joystream/joystream/pull/418/commits/e35594ca09ca47d7a2eb93d8e7e4b6a1f8c21b91

query-node/joystream-query-node/mappings/members.ts
metmirr: May be can use BigInt here? [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt). Typescript support it and `MAX_SAFE_INTEGER` is the same.

metmirr:
- Removed `s` from the entity names by using custom naming strategy

query-node/joystream-query-node/mappings/members.ts
metmirr: It is polkadot.js Codec type. From our last meeting: it would be better by using typeorm's `ColumnOptions.transformer` to convert the value automatically with `to/from` functions.

PR #420, opened 2020-05-13
Bootstrap members from the genesis block

dzhelezov:
This PR is the first attempt to introduce "bootstrapping" of the database by sourcing entities directly via Substrate API. For that purpose one can implement methods starting with `boot...` with substrate API and query runner as parameters.


bedeho (approved):
Very nice work.

PR #422, opened 2020-05-14
Use common thread post id types

mnaamani:
Makes ThreadId and PostId types configurable on trait of Forum module, and configure both to be `u64`.

Avoids future issues similar to https://github.com/Joystream/apps/issues/434

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -845,8 +845,8 @@ parameter_types! {
 impl proposals_discussion::Trait for Runtime {
     type Event = Event;
     type PostAuthorOriginValidator = MembershipOriginValidator<Self>;
-    type ThreadId = u32;
-    type PostId = u32;
+    type ThreadId = u64;
+    type PostId = u64;
     type MaxPostEditionNumber = ProposalMaxPostEditionNumber;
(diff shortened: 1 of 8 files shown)


mnaamani:
I could have gone with renaming ThreadId and PostId like so:

`ForumThreadId, ForumPostId, ProposalDiscussionThreadId, .. `

however this approach seemed cleaner...thoughts?

runtime/src/lib.rs
```diff
@@ -94,6 +94,12 @@ pub type Moment = u64;
 /// Credential type
 pub type Credential = u64;
 
+/// Represents a thread identifier for both Forum and Proposals Discussion
```
bedeho, line 97: Can you here add an extended comment explaining the nature of the problem, and thus why the values MUST indeed be the same currently. That context is lost in the source.

bedeho (changes requested):
I think this is fine for now, but really Polkadotjs needs to be fixed, rather than us having an opaque constraint in our runtime.

runtime/src/lib.rs
mnaamani: done in de4ce1ef37309d96f748358ca429e3667b7176d2

PR #423, opened 2020-05-14
Fix 'Cannot set property EntityManager of #<Object> which has only a getter'

dzhelezov:
This PR makes a workaround for https://github.com/goldcaddy77/warthog/issues/358 by forcing `tslib` version resolution to `1.11.2` for both `graphql` and `indexer` generated sources.


metmirr:
This will work perfectly for the indexer but the Graphql server is based on warthog and warthog generates `package.json` for itself.

We generate graphql server with warthog and at this point warthog cli is not working at all because of the tslib. So the same work needs to be done here as well [warthog](https://github.com/goldcaddy77/warthog)

dzhelezov:
> This will work perfectly for the indexer but the Graphql server is based on warthog and warthog generates `package.json` for itself.
> 
> We generate graphql server with warthog and at this point warthog cli is not working at all because of the tslib. So the same work needs to be done here as well [warthog](https://github.com/goldcaddy77/warthog)

I think it does work (at least for me), since it also fixes the `package.json' warthog is generating:
```
let pkgFile = JSON.parse(fs.readFileSync('package.json', 'utf8'));
pkgFile.resolutions["tslib"] = "1.11.2"
fs.writeFileSync('package.json', JSON.stringify(pkgFile, null, 2));
```

metmirr:
Ohh I can see that you are editing the generated `package.json`.

metmirr (approved):
It fixes `tslib` error.

bedeho (approved):
Nice workaround.

goldcaddy77:
If I pin this in Warthog for now, would it help?

bedeho:
You mean pin the issue? that would probably be helpful to others indeed.

goldcaddy77:
Sorry I meant pin the version of `tslib` in Warthog.

PR #448, opened 2020-05-20
Bureaucracy module. Iteration 3. Roles lifecycle.

shamil-gadelshin:
### Features 
- roles lifecycle extrinsics
- new tests
- string errors were migrated to the Substrate `decl_error` macro errors
- better code comments

### New roles lifecycle extrinsics
- update_worker_role_account
- update_worker_reward_account
- leave_worker_role
- terminate_worker_role
- unset_lead
- unstake


runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -226,10 +309,182 @@ decl_module! {
...
+                return Err(Error::CurrentLeadNotSet);
+            }
+
+            // mutation
+
+            // Update current lead
+            <CurrentLead<T, I>>::kill();
+
+            // Unwrap lead
+            if let Some(lead) = lead {
```
bedeho: We statically know there is a lead, so this test cannot fail, e.g. do on L319

```Rust
let lead = ensure_lead_set()?; //Returns Ok(Lead)
```

runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -226,10 +309,182 @@ decl_module! {
...
+                // Trigger event
+                Self::deposit_event(RawEvent::LeaderUnset(lead.member_id, lead.role_account_id));
+            }
+        }
+
+        /// Update the associated role account of the active worker.
+        pub fn update_worker_role_account(
```
bedeho, line 329: Why are there two actor identifiers here? The `worker_id` should resolve to a worker profile, which holds the underlying `member_id`, which then can be used to lookup whether account signing `origin` is same as membershp controller account.

bedeho: I guess perhaps I missed this in an earlier review, but there should be `Worker.member_id`, and there would then probably need to be a `ensure_member_signed_as_worker(origin, worker_id) => Result<Worker, >`.

bedeho (changes requested):
All in all, excellent work, and very nice job switching out the error types.

- Could we use the more salient mutation safety line from before?

```Rust
           //
           // == MUTATION SAFE ==
           //
```

- It struck me that, waiting for the worker to unstake actually introduces a great deal of complexity and risk for us, and.... we don't use it for anything. It is only used to update the state from unstaking to done. Can't we just simplify and drop it all? Its worth considering at least.

bedeho (approved):
**Perfect**

PR #450, opened 2020-05-20
Query node

metmirr:
This PR:

- Update indexer to use warthog models
- Fixes some issues with bootstrapping
- Make DB class into an interface for mocking


query-node/substrate-query-framework/index-builder/src/bootstrap/Bootstrapper.ts
```diff
@@ -89,13 +88,13 @@ export default class Bootstrapper {
...
-     * @param boot boothandler
+     * @param bootEvent SubstrateEvent
      */
-    private async shouldBootstrap(em: EntityManager, boot: BootstrapFunc):Promise<boolean> {
+    private async shouldBootstrap(em: EntityManager, bootEvent: SubstrateEvent):Promise<boolean> {
         const event = await em.findOne(SavedEntityEvent, { 
             where: { 
-                event_name: 'Bootstrap', 
-                event_method: boot.name
+                eventName: bootEvent.event_name,
```
dzhelezov: Bootstrap events are not "real" events. So `event_name` is always `Bootstrap` and `event_method` simply indicates the executed handler. It is enough to locate the event and decide is the bootstrap was successfull

query-node/joystream-query-node/bootstrap/members.ts
```diff
@@ -1,4 +1,4 @@
-import { Membership } from '../generated/indexer/entities/Membership';
+import { Membership } from '../generated/graphql-server/src/modules/membership/membership.model';
```
dzhelezov, line 1: shouldn't we assume that the graphql-server may be on a different machine and is decoupled from the indexer?

query-node/substrate-query-framework/index-builder/src/db/DatabaseManager.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @constructor(event: QueryEvent, manager: EntityManager)
+ *
+ */
+export default class DatabaseManager implements DB {
+  // Transactional entity manager
+  private readonly _manager: EntityManager;
+
+  readonly event: SubstrateEvent;
+
+  constructor(event: SubstrateEvent, manager: EntityManager) {
```
dzhelezov: Is it right that we create a `DatabaseManager` per event? Shouldn't a manager class be more like a singleton?

query-node/substrate-query-framework/index-builder/src/db/db.ts
```diff
@@ -1,54 +1,32 @@
...
 import { SubstrateEvent } from '..';
-import * as helper from './helper';
 
 /**
- * Database access object based on typeorm. Use typeorm transactional entity manager to perform get/save/remove
- * database operations.
- *
- * @constructor(event: QueryEvent, manager: EntityManager)
- *
+ * Database interaction interface.
```
dzhelezov: So here we  restrict the clients (e.g. the mapping functions) of this interface to only CRUD operations, right? It would be impossible to run custom queries, is it by design?

dzhelezov (changes requested):
Please, have a look at the comments

query-node/substrate-query-framework/index-builder/src/bootstrap/Bootstrapper.ts
metmirr: That makes sense, but we don't have `event_method` field on `SavedEntityEvent`.

query-node/joystream-query-node/bootstrap/members.ts
metmirr: Yes, that is right but we are using graphql server models now and that means we depend on. Maybe we can copy model files to indexer in `codegen` command but I am not sure if that is the right way. Do you have any suggestions here?

query-node/substrate-query-framework/index-builder/src/db/db.ts
metmirr: That's right, only CRUD operations. I think we don't want the users to run custom queries. For example if they use typeorm `queryRunner` directly we will not be able to track the last processed event.

query-node/joystream-query-node/bootstrap/members.ts
dzhelezov: Simply copying is the easiest and I think it's been working so far, but perhaps we can package it into a module and import? The main advantage would be versioning, otherwise I think copying is fine

query-node/substrate-query-framework/index-builder/src/bootstrap/Bootstrapper.ts
dzhelezov: Oh okay, so it's been refactored out at some point? So I suggest using eventName: `Bootstrap.${boot.name}` then and lookup by the name only.

query-node/substrate-query-framework/index-builder/src/db/db.ts
dzhelezov: OK, makes sense

query-node/substrate-query-framework/index-builder/src/db/DatabaseManager.ts
metmirr: I have been thinking about this actually. Since we need to access `event` object via `db` in the mappings and let users perform only `CRUD` operations. I couldn't find any other solution. Maybe the name is misleading?

dzhelezov (approved):

query-node/substrate-query-framework/index-builder/src/db/DatabaseManager.ts
bedeho: I agree with @dzhelezov here, and `event` is not even used for anything internally in `DatabaseManager`. In the mapping, it currently just receivs the `DB` instance, but you could easily just change this, so you go from

`handleMemberRegistered(db: DB)` to `handleMemberRegistered(db: DB, event: SubstrateEvent)`

Is there any reason this would not work?

query-node/substrate-query-framework/index-builder/src/db/db.ts
bedeho: I think we should not allow the user to make custom `write` operations, but arbitrary `read` queries would be fine.

Its not very important, but would it be possible to offer a way to only do the latter, not the former?

bedeho (changes requested):
Overall looks good, just two questions. If its possible to fix the first issues, lets do that in this PR right away, as its trivial.

query-node/substrate-query-framework/index-builder/src/db/db.ts
metmirr: `typeorm` provides [`SelectQueryBuilder`](https://typeorm.io/#/select-query-builder). I tried it but it is not `read-only` safe. It means one can call `insert` on `select`.

```ts
await getConnection()
    .createQueryBuilder()
    .select('membership')
    .from(Membership, 'membership')
    .insert() // It works
    .into(Membership)
    .values([{ memberId: 234, accountId: '5FdyqzHPZe71EzDU3pHKBWEPChbUitfoToUgrDtnWoojFtic' }])
    .execute();
```

Right now you can perform `join`, `relations`, `order` with `db.get()`. I think if we add `getMany()` that would be enough

bedeho: Ok great, that is quite good actually.

bedeho (approved):
Great job.

PR #475, opened 2020-05-26
Query node fts queries

dzhelezov:
This PR adds initial Fulltext Query (FTS) capabilities to the query node. Currently it supports only a single-entity single-field text queries via `@fullTextSearchable(query=<query_name>)` directive in the schema file. The current layout is thought to be extensible enough to support even cross-table queries.

The set of supported directives and query definitions will likely change as the scope is extended.
The implementation proceeds as follows. The input schema file is parsed and an intermediate model (`WarthogModel` class) is created. Then the model is combined with the Mustache templates to generate a migration (for text-search indexes and auxilliary columns), a resolver (to expose a GraphQL query through `type-graphql`) and a service (which builds and executes an SQL query against the corresponding full-text search index).


query-node/substrate-query-framework/cli/src/helpers/SchemaParser.ts
```diff
@@ -30,8 +31,10 @@ export class GraphQLSchemaParser {
     if (!fs.existsSync(schemaPath)) {
       throw new Error('Schema not found');
     }
-    const docNode = parse(fs.readFileSync(schemaPath, 'utf8'));
-    const schema = buildASTSchema(docNode);
+    const ast = parse(fs.readFileSync(schemaPath, 'utf8'));
+    // in order to build AST with undeclared directive, we need to 
+    // switch off SDL validation
+    const schema = buildASTSchema(ast, { assumeValidSDL : true });
```
metmirr: Is it good to assume the schema is valid? I mean one can do a mistake then we will not generate code correctly perhaps there will be errors but not sure...
For custom scalar types (e.g BigInt) I also need to add scalar type definition to the schema (e.g `scalar BigInt`) but setting this to `true` doesn't work for scalar types it is only for directives I think. So I have to do something like this:

```ts
const SCALAR_TYPE_DEFINATIONS = `
scalar BigInt
scalar BigDecimal

// Maybe you can add directive here
// directive @@fullTextSearchable(query: "handles") on FIELD_DEFINITION
`
const ast = parse(SCALAR_TYPE_DEFINATIONS + fs.readFileSync(schemaPath, 'utf8'));
```

dzhelezov: This is a good point and a possible way to make the SDL valid. There a few things to bear in mind: 1) we then probably need to add the required GraphQL `__Query` top-level entity (as right now we have a rather weird hack that the absence of this top-level entity is _the only_ allowed validation error) 2) we'd still need additional validation checks (like that the directive is applied only on the string fields and also some query-specific ones).

metmirr: That's right!

dzhelezov: I can add the required change to this PR.

metmirr: That would be really good and I will add my scalar definations then.

dzhelezov: See the commit below; I had to add another hacky clause to the type name filter so that that bogus `Query` type doesn't show up in the model.

query-node/joystream-query-node/scripts/reset-dev.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
bedeho, line 1: Is this script Joystream specific, or should it be part of the framework for any chain?

query-node/joystream-query-node/scripts/run-dev.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
bedeho, line 1: Is this script joystream specific, or should it be part of the framework for any chain?

bedeho (commented):
Had a look at this, and seeing that it's gotten quite involved - I am not able to really get a clear view of what is going on - in particular how the presence and requirements of the FTS is not disturbing the old Warthog stuff - within the time constraints I am facing, but I am really impressed the progress you were able to make here. I just left two minor questions. I will try to take another stab at getting familiar with the codebase once this refactoring has been merged

 https://github.com/Joystream/joystream/pull/513

query-node/joystream-query-node/scripts/run-dev.sh
dzhelezov: This script is for dev purposes only. It start-ups the indexer and the graphql server on the same machine. `reset-dev.sh` drops-creates and then bootstraps the database to simplify the development. Not sure what is the right place for these scripts and if we actually need them to keep in the repo, but thought it may be useful for devs to have them.

PR #490, opened 2020-05-27
Membership module refactoring

### LINKED ISSUE #93: Cleanup membershp module
Both module storage and interface need to be cleaned up.

The storage and forum client modules need the old interface, and thus storage model, to work, so we have postponed.

1. Drop storage fields

```Rust

pub MemberIdsByRootAccountId get(member_ids_by_root_account_id) : map T::AccountId => Vec<T::MemberId>;

// Mapping of a controller account id to vector of member ids it controls
pub MemberIdsByControllerAccountId get(member_ids_by_controller_account_id) : map T::AccountId => Vec<T::MemberId>;
```

and all modules that depend on them.

2. No method shall identify members through accounts, but we can have checker methods (return `bool`) that can verify the validity of a provided account.

3. Change `Profile` to `Membership`, and storage field `MemberProfile` to `MembershipById`. Its far less confusing.

4. `MembersCreated` => `NextMemberId`

5. `Handles` => `MemberIdByHandle`

6. Drop `UserInfo`, does nothing.

7. Drop `Option` wrapper for around value types of maps, e.g. `MemberProfile`.

### LINKED ISSUE #115: Simplify membership module
# Background

Currently, the membership module holds the map that binds a member to all the roles of the members at any time. By role, we here not only mean a type, like a _storage provider_ or a _counsilor_, but also a specific instance of that role, so e.g.  _storage provider 66_ or _counsilor 210_ .

Specifically for the role to member binding, there is the storage field.
```Rust
pub MembershipIdByActorInRole get(membership_id_by_actor_in_role): map ActorInRole<T::ActorId> => T::MemberId;
```
and also, each member profile holds the set of current actor roles

```Rust
pub struct Profile<T: Trait> {
    /// The set of registered roles the member has enrolled in.
    pub roles: ActorInRoleSet<T::ActorId>,
}
```

Whenever a member is to enter or leave a role, in some other module, then suitable calls have to be made to keep this state in the membership module up to date.

Quite separately from this, the membership module happens to have the ability to gate whether a member can, in fact, enter or leave a role at any given time, and also includes a static list of role types that allows it to possibly enforce role type-specific policies. The types are currently defined by

```Rust
pub enum Role {
    StorageProvider,
    Publisher,
    CuratorLead,
    Curator,
}
```

# Problem

## Needless complexity caused by role entry gating

Currently, we do not even have any proposal for what sort of actual gating policy we would want to implement that would be role-specific. This appears to be a totally made-up requirement, and it is the only reason the specific role types are encoded in `Role`. Moreover, the only uniform policy which appears semi-plausible, that a disabled member, cannot enter a given role, does not appear to be of much real use. If we give up this idea entirely, then all of these checks go away, which greatly simplifies all of the other modules. One particularly impractical consequence of the current gating is for example that when hiring, one has to recheck that someone who is chosen for a position is still eligible for entry into the role after the initial check that is made when they apply. This is because entry permission could have changed in the intermediate time.

## Move membership to role binding out

Keeping the binding information in the membership module requires more runtime state, which ultimately can get out of synch. Its safer and simpler to have each module keep the relevant membership id in the already existing representation of the role, e.g.

```Rust

struc Curator {
...
member_id: T::MemberId
}
```

shamil-gadelshin:
### Changes
- incapsulate storage fields MemberIdsByRootAccountId and MemberIdsByControllerAccountId. Remove public access to those fields.
- rename Profile to Membership, and storage field MemberProfile to MembershipById. 
- rename MembersCreated => NextMemberId.
- rename Handles => MemberIdByHandle.
- drop UserInfo.
- CheckedUserInfo was renamed to the ValidatedUserInfo and kept for simplicity as a private internal structure.
- drop Option wrapper for around value types of maps
- remove roles from the membership module
- remove roles from the content working group module
- move ActorId from membership to the content working group
- add MemberId to the content working group Lead
- move Role from membership to roles module


runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,647 @@
...
+decl_storage! {
+    trait Store for Module<T: Trait> as Membership {
+        /// MemberId to assign to next member that is added to the registry, and is also the
+        /// total number of members created. MemberIds start at Zero.
+        pub NextMemberId get(members_created) : T::MemberId;
+
+        /// Mapping of member's id to their membership profile
+        pub MembershipById get(membership) : map T::MemberId => MembershipOf<T>;
+
+        /// Mapping of a root account id to vector of member ids it controls.
```
bedeho, line 153: These two fields were supposed to be blocked

https://github.com/Joystream/joystream/issues/93#issue-522940475

It's actually a substantial issue because there is non-trivial business logic here and in other modules that depend on the implied (controller & role) account uniqueness, hence removing them will help us track down and update all of those.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,647 @@
...
+            } else {
+                Err(ControllerAccountForMemberCheckFailed::NotControllerAccount)
+            }
+        } else {
+            Err(ControllerAccountForMemberCheckFailed::NotMember)
+        }
+    }
+
+    /// Returns true if account is either a member's root or controller account
+    pub fn is_member_account(who: &T::AccountId) -> bool {
```
bedeho, line 455: This must go.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,647 @@
...
+    pub fn ensure_membership(id: T::MemberId) -> Result<MembershipOf<T>, &'static str> {
+        if <MembershipById<T>>::exists(&id) {
+            Ok(Self::membership(&id))
+        } else {
+            Err("member profile not found")
+        }
+    }
+
+    /// Ensure that given member has given account as the controller account
+    pub fn ensure_is_controller_account_for_member(
```
bedeho, line 437: I think this is exclusively used for authenticating claimed members, both in this module, and by other caller module. We may as well make it `ensure_member_signed_as_controller(origin, member_id)`.

bedeho (changes requested):

runtime-modules/membership/src/lib.rs
shamil-gadelshin: This is the 'allowed check' from the https://github.com/Joystream/joystream/issues/93#issue-522940475 (task item 2).

runtime-modules/membership/src/lib.rs
shamil-gadelshin: I incapsulate them to be crate-only: the 'allowed check' is_member() requires them as main components. So, they are blocked. Also, I got rid of their getters, so no one can use this collections from external modules.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Yes. And when I try to create it, I found a similar method with a slightly different name `ensure_member_controller_account_signed()` nearby.

runtime-modules/membership/src/lib.rs
bedeho: Seeing that point now, sorry for the confusion this caused, this was probably meant to say something like `ensure_*(origin, MemberId)` is ok, but clearly its not. But we do not want this. There should be no constraint on the account space, as this requires, so `is_member_account` should go, and everything that relies on it in the runtime must be updated.

runtime-modules/membership/src/lib.rs
bedeho: Ok, so we remove this then?

shamil-gadelshin:
Several issues were created in order to remove` is_member()`:
- https://github.com/Joystream/joystream/issues/519
- https://github.com/Joystream/joystream/issues/520
- https://github.com/Joystream/joystream/issues/521
- https://github.com/Joystream/joystream/issues/522

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Removed that entry, but found another one more complicated use-case in the content working group in the apply_on_curator_opening(). So we keep it for now.

bedeho (approved):

PR #491, opened 2020-05-27
Style reducers and Buttons.

fcescob:
This PR implents a new styling system and refactors the Button component with it, considerably reducing boilerplate. #487 

## The Buttons

The buttons should work as expected and should be a loyal copy of the Figma design. All of them are present in Storybook with hover and active states.

## The styling system

This PR changes how we define styles for a single component, instead of declaring how the component should look like based on some props, we now declare how a component should look like based on props _and_ on the previous styles. 

To give one concrete example, this:
```javascript
const styles = (props) => css`
  display: flex;
// Some condition that depends on the props
  color: type === 'primary' ? !disabled ? ...
`
```

In the example above, nesting can get unreadable quickly and it makes it hard to see which styles are applied at a glance.
The same example in our new system would be:
```javascript
const colorFromType = (styles, props) => ({
    ...styles,
    color: props.type === 'primary' ? 'blue' : 'hotpink';
})

const disabled = (styles, props) => ({
   ...unsetColorProperties(styles),
  color: 'grey'
})

const full = (styles, props) => ({
  ...styles,
  width: full ? '100%' : styles.width
})
const styles = makeStyles([colorFromType, full, disabled]) 
```

This may look like more boilerplate, but now `full` and `disabled` are reusable across components. 
We have got out of composing styles with the CSS cascade, and so we gained fine-grained control on which properties we keep at each styling function that we apply.
This is very helpful when dealing with corner states like "disabled" where some properties declared before, like box-shadow or color, might leak and give us some unexpected effect (disabled button colored like an error one on focus).

## How It works

The new styling system only has 2 core concepts. 
1. Style Reducers / Styling functions, which is just a fancy word for a function taking the current styles and some props as arguments and returning a style Object. 
```
f: (styles: styleObj, props: any) => styleObj
```
All functions with this signature are valid style functions. 

2. The `Reducer` function, which allows exposes an interface for combining style functions in a coherent way.

## Caveats of the new styling system and things to improve

The new styling system is far from perfect, and has small, but noticeable downsides.

- We now write styles in object notation instead of a template literal format, and given how Emotion works, it might prove hard to allow authoring styles in a template literal fashion (to my understanding, this would required re-parsing the styles with a CSS parser at each function composition).

- The Reducer function is not correctly typed, along with the `styleObj` type, which makes IDE support when writing styles inferior to the native Emotion experience (you get the possible properties, but not their allowed values in autocomplete)

## Feedback

Any feedback is much welcome 👀


bedeho (approved):
This looks nice, I like the benefits of these composable style functions, not least that they are easy to test. Given that we are dealing with such a meat-and-potatoes kind of task, I am a bit concerned that we are doing our own totally new thing that is not already captured in some framework or as an identified approach. I guess we will learn in due time, so I am very happy to move forward with this as is, and in any case its great that you laser focused on maintainbility.

Issue #402, opened 2020-05-07
Make Pioneer Modules Consumable 


fcescob:
# Problem

As of today, Pioneer modules are not consumable outside of Pioneer, making it hard to share functionalities and reuse code across projects.

For example, it's impossible (or at least very hard) to use `joy-media` to fetch media contents in an external project, even `joy-utils`, which should be reusable by definition, is not importable as a single module has cyclical dependencies with `joy-members`.

![Untitled Diagram](https://user-images.githubusercontent.com/29527327/81328747-b5a55d00-909d-11ea-96b5-0b89b5f46c3e.png)

The example above is true for multiple Pioneer modules like `joy-roles`, `joy-election` etc... which all depend on each other in some way.

# What I propose

I believe the underlying issue to be how runtime interaction are scattered around the codebase and tightly coupled with the UI.

When importing a Pioneer module, you're probably interested in some runtime interaction, and don't care much about the UI components of that specific module.
I propose to extract all runtime interaction functionalities into a unified api, consumable everywhere.

![Proposed](https://user-images.githubusercontent.com/29527327/81329259-93f8a580-909e-11ea-9cc5-ce341078528d.png)

Each Pioneer module would then rely on this api to define its business logic, making each module self contained, even better, external apps wouldn't need to import Pioneer modules since the api covers all runtime interactions.

## Transition
The transition from the current Pioneer could be made of three steps and be relatively painless. 

1. Refactor and expand `joy-utils` to be `joy-api` and cover runtime interactions. The only Pioneer module used here should be `joy-types`

2. Start refactoring each module and make it self contained, with no dependency on other Pioneer modules beside `joy-types` and `joy-api`

3.  Bring up to date external modules with Polkadot apps and prefer NPM modules over forked ones in the repo.

# Feedback

Very interested in hearing comments and discussing other solutions!

semeano:
Just a couple of questions:

- Pioneer would also be an "external app" or would consume it in a different way?
- How do you import class definitions with an API? (e.g. SimpleCache from `joy-utils`)
- Can you transpile your code without a local type/class definition?

I'm not saying it wouldn't work, I'm just asking because I don't know how that would work.

The only thing I would say is, when refactoring the modules (and Pioneer), be sure to follow the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

fcescob:
Good questions! Let me try to address those in order:

[...]

PR #494, opened 2020-05-28
Joy-proposals - extract potentially reusable code into joy-utils

Lezek123:
This PR contains the initial progress I've made on extracting potentially reusable code from Pioneer packages into `joy-utils` (which may then become part of `joystream-js`). In this particular one I was only extracting code from `joy-proposals` package.

**Rationale:**

The issue was origianlly raised by @Gamaranto here: https://github.com/Joystream/joystream/issues/402

Pioneer consists of many "uncomsumable" packages like `joy-media`, `joy-members` etc., which circulary depend on each other and are pretty much ususable beyond Pioneer, even though they contain a lot of code that could be potentially be reusable.

Among other problems, those packages also usually duplicate a lot of other packages' logic, ie. a few of them (`joy-media`, `joy-roles`, `joy-proposals`) define their own transport layer with methods like `blockTimestamp` which returns timestamp by a given block number (a method that could obviously be useful for other packages/projects). There are also a lot of hooks, components, functions, constants etc. like that, which we currently either import between packages causing circular dependency hell or rewrite in the scope of each new package, losing time and multiplying code and potential points of failure.

**The new structure of `joy-utils`**

In order to prepare `joy-utils` to be an independent library (or part of such) that could handle all this reusable Pioneer's code I ended up with 5 base categories of code that we could put there:
- `consts` - there is a noticable amount of hardcoded constants that cannot either yet or ever be fetched from the runtime. In `joy-proposals` those are for example: validation constraints (enforced by runtime), proposal types, approval/slashing quorum and threshold by type, required stakes by type etc. In `joy-media` we have schemas like `Video`, `Channel`, `Content` etc. that would also be more useful as part of a separate library. By extracting those out we could share them among other projects and not have to hardcode them multiple times.
- `functions` - this group will contain all the functions defined in the Pioneer packages that could be reused (either by other packages or projects), ie. some common calculations, converters etc.
- `transport` - the aim here is to create a unified interface to fetch data from the api. This is probably the hardest one to get right, because it would need to also handle mock transports, potential caching, subscriptions etc. Having a unified transport layer we would potentially also need to re-think the logic of HOCs like `withMyAccount` etc. (I don't think this needs to be done all at once though, besides it should be considered how query node and other WIP solutions could affect this). My current idea is that there will be one global `Transport` instance for the Pioneer (available through context) that will internally instantialize more specific transports like `ProposalsTransport` or `MembersTransport`. Those more specific transports could be injected to each other in the main `Transport` constructor (ie. `CouncilTransport` could have `MembersTransport` instance injected). This way we will have a very clear chain of dependency and avoid getting back to circular references, while maintaining some kind of modularity. Inside Pioneer, for example, we could use `transport.proposals.all()` to fetch all proposals (with additional information collected by calling various lower-level api methods) or `transport.council.activeMembers()` to fetch the profiles of all active council members. Having Pioneer rebuild in a way that it will rely on a unified transport layer should allow us to switch the actual transport implementation more easily when needed.
- `react` - here we could keep reusable code specific to react, like `hooks`, `components`, `contexts` etc. This could then be extracted as a separate library, having the actual `joystream-js` just as a dependency.
- `types` - here we could keep types that don't need to be part of `@joystream/types` but will be still valueable for some higher-level interactions with the data. For example, in case of `joy-proposals` we have `ParsedProposal` which contains a creation date (based on creation block), fetched and parsed creator profile etc. (although this type itself may need some improvements, it's way easier to use a construct like that in order to display the data on the frontend instead of operating just on the types available in `@joystream/types`).

**The scope of this PR**

In this specific PR I was tyring to re-build `joy-proposals`. having the above structure in mind. It was mostly about splitting the code and putting it in the right places inside the structure I created in `joy-utils`, I wasn't trying to change the actual implementation too much. Obviously this will not allow us to see the real benefits of refactoring the entire Pioneer this way, but it may server as a good starting point for a discussion and allow us to test some implications of such a rebuild on a smaller scale first.


pioneer/packages/joy-utils/src/functions/misc.ts
```diff
@@ -0,0 +1,32 @@
...
+      false
+    );
+  });
+}
+
+export function splitOnUpperCase(str: string) {
+  return str.split(/(?=[A-Z])/);
+}
+
+export function slugify(str: string) {
```
fcescob: If our future plan is to make `joy-utils` a consumable module, we should strive to include only the core necessities to both reduce bundle size and allow for easier documentation.
I would keep here only the functions used internally by `joy-utils` and leave the others implementation to the consumers of `joy-utils`.

pioneer/packages/joy-utils/src/functions/proposals.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+    default: {
+      throw new Error(`Proposal Type is invalid. Got ${type}.`);
+    }
+  }
+  return stake;
+}
+
+export function calculateMetaFromType(type: ProposalType): ProposalMeta {
+  const image = "";
```
fcescob: `image` here is about the UI, I would leave this to the consumers to deal with the it as well.

fcescob (commented):
Left two small comments, but looks very good to me, I really like the approach you took with the transports. Great job!

Lezek123:
Added a few minor adjustments that should fix the mentioned issues:
- Moved hardcoded proposals data from `functions/proposals` to `consts/proposals` (as that's where it really belongs) and changed the structure of it from functions to a single object with proposal types as keys. Figured out we don't need to separate stakes from other hardcoded data, so just added it as part of `ProposalMeta`.
- Removed functions like `snakeCaseToCamelCase`, `slugify` etc. and replaced them with `loadash` equivalents.
- Removed `image` as part of hardcoded `ProposalMeta` (we don't have this implemented yet anyway)

fcescob (approved):

PR #501, opened 2020-05-28
Travis: Remove github releases config

mnaamani:
Deployment builds will be done on a different platform so removing deployment related config.
Also force triggers build of WASM runtime.


shamil-gadelshin (approved):
LGTM

PR #508, opened 2020-05-29
Bureaucracy module. Add slashing to the stake operations.

shamil-gadelshin:
## New slashing extrinsics summary
1. slash_worker_stake(origin, worker_id, balance) - demands a leader origin, slashes the worker stake. No limits, no actions on zero stake.

2. decrease_worker_stake(origin, worker_id, balance) - demands a leader origin, decreases the worker stake and returns the remainder to the worker role_account. Can be decreased to zero, no actions on zero stake.

3. increase_worker_stake(origin, worker_id, balance) - demands a worker origin, increases the worker stake. Transfers tokens from the worker role_account to the stake. No limits on the stake.


runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -835,6 +856,93 @@ decl_module! {
...
+
+            ensure!(balance != <BalanceOf<T>>::zero(), Error::StakeBalanceCannotBeZero);
+
+            let stake_profile = worker.role_stake_profile.ok_or(Error::NoWorkerStakeProfile)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            ensure_on_wrapped_error!(
```
bedeho, line 878: I think this needs to be above the `MUTATION SAFE`

runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -835,6 +856,93 @@ decl_module! {
...
+            Self::ensure_origin_is_set_lead(origin)?;
+
+            let worker = Self::ensure_worker_exists(&worker_id)?;
+
+            ensure!(balance != <BalanceOf<T>>::zero(), Error::StakeBalanceCannotBeZero);
+
+            let stake_profile = worker.role_stake_profile.ok_or(Error::NoWorkerStakeProfile)?;
+
+            ensure_on_wrapped_error!(
+                <stake::Module<T>>::ensure_can_decrease_stake(&stake_profile.stake_id, balance)
```
bedeho: Is this call actually required, what does it do that is not already covered in `decrease_stake_to_account`?

runtime-modules/bureaucracy/src/lib.rs
```diff
@@ -835,6 +856,93 @@ decl_module! {
...
+        pub fn increase_worker_stake(origin, worker_id: WorkerId<T>, balance: BalanceOf<T>) {
+            // Checks worker origin, worker existence
+            let worker = Self::ensure_worker_signed(origin, &worker_id)?;
+
+            ensure!(balance != <BalanceOf<T>>::zero(), Error::StakeBalanceCannotBeZero);
+
+            let stake_profile = worker.role_stake_profile.ok_or(Error::NoWorkerStakeProfile)?;
+
+            ensure_on_wrapped_error!(
+                <stake::Module<T>>::ensure_can_increase_stake(&stake_profile.stake_id, balance)
```
bedeho: Same question here actually.

bedeho (changes requested):
Excellent job!

runtime-modules/bureaucracy/src/lib.rs
shamil-gadelshin: Why? slash_immediate() - will cause changes in the storage. And it should be after the 'mutation' sign.

runtime-modules/bureaucracy/src/lib.rs
shamil-gadelshin: Actually, decrease_stake_to_account() should call the ensure_can_decrease_stake() internally. But it doesn't. But even if it would do that we use 'Parity recommended' style to use 'ensure_*' methods before mutation methods.

runtime-modules/bureaucracy/src/lib.rs
shamil-gadelshin: Same answer

runtime-modules/bureaucracy/src/lib.rs
bedeho: Because mutation is not safe on line 876 at the moment, and this gives that impression, which could lead someone to add other business logic in that wedge. Its only safe _after_  `slash_immediate` has successfully executed, because it checks the very last condition internally.

runtime-modules/bureaucracy/src/lib.rs
bedeho: 1) Are you saying there is currently a bug in `decrease_stake` then? is it not making all required checks itself?

2) I don't think it makes sense to do if it does not add anything, just makes it harder to see what is going on. When we make calls to other methods in our own library stack we often end up doing this.

bedeho (approved):

PR #512, opened 2020-05-29
Make PP and ToS routes work again

Lezek123:
This PR fixes a bug with `Privacy Policy` and `Terms of Service` links not working on member register page, due to the fact that those routes has been mistakenly removed while we only wanted to remove the tabs from the sidebar.


mnaamani (approved):
Tested and works as expected.

PR #514, opened 2020-05-29
Pioneer: Fix for #511

### LINKED ISSUE #511: Pioneer Build Error
Strangely getting a build error locally, whereas github workflow and netlify checks are passing.

`yarn workspace pioneer build`

on development branch

```
Version 3.7.2
joy-proposals/src/Proposal/ChooseProposalType.tsx:42:55 - error TS2339: Property 'category' does not exist on type 'never'.

42             .filter(typeInfo => !category || typeInfo.category === category)
                                                         ~~~~~~~~

joy-utils/src/react/context/transport.tsx:5:8 - error TS2613: Module '"/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' has no default export. Did you mean to use 'import { Transport } from "/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' instead?

5 import Transport from "../../transport";
         ~~~~~~~~~

joy-utils/src/react/hooks/proposals/useProposalSubscription.tsx:34:13 - error TS7006: Parameter 'unsubscribe' implicitly has an 'any' type.

34       .then(unsubscribe => {
               ~~~~~~~~~~~

joy-utils/src/react/hooks/useTransport.tsx:2:8 - error TS2613: Module '"/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' has no default export. Did you mean to use 'import { Transport } from "/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' instead?

2 import Transport from "../../transport";
         ~~~~~~~~~


Found 4 errors.
```

```sh
MokoBook:joystream mokhtar$ yarn --version
1.16.0
MokoBook:joystream mokhtar$ node --version
v12.17.0
```

mnaamani:
Attempt to Fix #511

### DIFF
--- a/pioneer/packages/joy-utils/src/react/context/transport.tsx
+++ b/pioneer/packages/joy-utils/src/react/context/transport.tsx
@@ -2,7 +2,7 @@ import React, { createContext, useContext } from "react";
 import { ApiContext } from "@polkadot/react-api";
 import { ApiProps } from "@polkadot/react-api/types";
 
-import { Transport } from "../../transport";
+import Transport from "../../transport/index";
 
 export const TransportContext = createContext<Transport>((null as unknown) as Transport);
 
--- a/pioneer/packages/joy-utils/src/react/context/transport.tsx
+++ b/pioneer/packages/joy-utils/src/react/context/transport.tsx
@@ -2,7 +2,7 @@ import React, { createContext, useContext } from "react";
 import { ApiContext } from "@polkadot/react-api";
 import { ApiProps } from "@polkadot/react-api/types";
 
-import Transport from "../../transport";
+import { Transport } from "../../transport";
 
 export const TransportContext = createContext<Transport>((null as unknown) as Transport);
 
--- a/.github/workflows/pioneer-pr.yml
+++ b/.github/workflows/pioneer-pr.yml
@@ -19,7 +19,6 @@ jobs:
         yarn install --frozen-lockfile
         yarn workspace pioneer run build:code
 
-jobs:
   pioneer_build_code_osx:
     name: (MacOS) Build Code
     runs-on: macos-latest
--- a/.github/workflows/pioneer-pr.yml
+++ b/.github/workflows/pioneer-pr.yml
@@ -1,10 +1,28 @@
-name: Pioneer-PR
+name: Pioneer
 on: [pull_request, push]
 
 jobs:
-  pioneer_build_code:
-    name: Pioneer Build Code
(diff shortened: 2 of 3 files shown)


mnaamani:
I'm convinced now there is something uniquely wrong with my local dev environment. Clearly I have some cached typescript definitions files interfering with the build

pioneer/packages/joy-utils/src/react/context/transport.tsx
```diff
@@ -2,7 +2,7 @@ import React, { createContext, useContext } from "react";
 import { ApiContext } from "@polkadot/react-api";
 import { ApiProps } from "@polkadot/react-api/types";
 
-import Transport from "../../transport";
+import { Transport } from "../../transport";
```
mnaamani: I can clearly see by examining `../../transport/index.ts` that there is a default export class statement which means this fix is incorrect, but I can't figure out how my IDE is not seeing the same thing :)

Screen Shot 2020-05-29 at 9 19 48 PM

Lezek123: I think I know what's the issue here. There is `joy-utils/src/Transport` (file) and `joy-utils/src/transport` (directory). Perhaps in some environments where paths are considered case insensitive import from file takes priority over the import from directory and hence the error.

mnaamani:
I think you are right on the money! OSX has strange hybrid case sensitivity
and this is failing on my Mac.

Not the first time we’ve had issues with case sensitive issues with file
names :(

mnaamani:
So the quick and dirty fix 2f2df78 confirms it is the issue you pointed out. I'm not sure there is an environment var to control the precedence of how node handles import paths when a file or folder match, but in this particular case its not really the issue, its actually the fact that it can't tell the different between `transport.ts` and `Transport.ts` and `./transport/` due to OSX filesystem case sensitivity.

I added a job to github workflows to build for OSX so we can spot such intricate "bugs" in future, but it would be better I think to follow a convention where we try to avoid having a file and folder with similar names that would cause this case to be problematic.

Lezek123 (approved):
The CI works well, I was able to reproduce the error in my fork.
This simple fix seems to be enough for now, I will rename the file or change the structure in this directory in some PR later (`joy-utils` needs some cleanup anyway, as it is currently partially the old `joy-utils` and patrially a work-in-progress `joystream-js`)

PR #517, opened 2020-06-01
Epoch/era bug fix

Lezek123:
This PR fixes a bug with calculating `eraProgress` in the `Staking overview`.
It seems to be a bug in polkadot's `api-derive` package, which can be noticed in `v0.96.1` (the version we use), but appears to be fixed in `v0.97.1`.

See: https://github.com/polkadot-js/api/blob/v0.96.1/packages/api-derive/src/session/info.ts
And: https://github.com/polkadot-js/api/blob/v0.97.1/packages/api-derive/src/session/info.ts
(inside _createDerivedLatest_ definition)

Since changing the api version may be a bit more risky I fixed it "manually" for now and added an explainatory comment.

### DIFF
--- a/pioneer/packages/app-explorer/src/SummarySession.tsx
+++ b/pioneer/packages/app-explorer/src/SummarySession.tsx
@@ -12,10 +12,15 @@ import { withCalls } from '@polkadot/react-api';
 import translate from './translate';
 import { formatNumber } from '@polkadot/util';
 
+import { SessionIndex } from '@polkadot/types/interfaces';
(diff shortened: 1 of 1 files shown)


mnaamani:
Nice workaround. The only choice really, because bumping the api version breaks the app totally, the connection will just fail. Because the metadata version the api expects from the runtime will not match.

I see the `SummarySession` component is used both in explorer and staking apps so the fix covers both views :+1:

Tested and LGTM

mnaamani (approved):

PR #518, opened 2020-06-01
Unbonding "blocks left" fix

Lezek123:
This PR is an attempt to fix the "Unbonding period" bug mentioned in https://github.com/Joystream/apps/issues/464

**The issue**

Incorrect value of "blocks left" appears inside the tooltip in "unbonding" section of the address info (ie. `Validators => Account actions` or `My keys`) when unbonding funds.

**The cause**

I tracked this issue down all the way to the `@polkadot/api` repository, since the data displayed in this part of the UI comes directly from `api.derive.staking.info`.

The formula used to calculate the `remainingBlocks` in `api.derive.staking.info` is as follows (see: https://github.com/polkadot-js/api/blob/v0.96.1/packages/api-derive/src/staking/info.ts):
```
const remaining = eraLength.mul(era).sub(bestNumber);
```
Where `era` equals to the era the funds get unbonded at.

In case of our current testnet (https://testnet.joystream.org/) the values at the moment of my test were:
```
Current era: 2058
Current best block: 1040200
```
This doesn't really add up perfectly if we assume that eras start from `0` and always increment by `1` after each `600` blocks starting from genesis. In that case we would have best block of ~`1234800`, so using the polkadot's formula we're getting a result higher by `194600` blocks than expected.

As @bwhm mentioned: "It is certainly true that some eras have been shorter than 600, as it seems to happens when a high percentage of the current validators drop off in the same era."

**The solution**

I decided to implement a different way of calculating `remainingBlocks` value for unbonding funds, relying on `stakingLedger` and some of the session and era information (kind of related to: https://github.com/Joystream/joystream/pull/517).

The current formula for calculating `blocksLeft` is:
```
(era - currentEra) * eraLength - eraProgress
```
Where `era` is the era at which the funds are unbonded and `eraProgress` is calculated the same way as in https://github.com/Joystream/joystream/pull/517 after the fix.

**_Some tests are still in progress..._**

### DIFF
--- a/pioneer/packages/react-components/src/AddressInfo.tsx
+++ b/pioneer/packages/react-components/src/AddressInfo.tsx
@@ -2,8 +2,10 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { DerivedBalances, DerivedStaking } from '@polkadot/api-derive/types';
-import { ValidatorPrefs0to145 } from '@polkadot/types/interfaces';
+import { createType } from '@polkadot/types';
+import { DerivedBalances, DerivedStaking, DerivedUnlocking, DerivedSessionInfo } from '@polkadot/api-derive/types';
+import { ValidatorPrefs0to145, BlockNumber, EraIndex, SessionIndex } from '@polkadot/types/interfaces';
+import { u64 } from '@polkadot/types/primitive';
 import { BareProps, I18nProps } from './types';
 
 import BN from 'bn.js';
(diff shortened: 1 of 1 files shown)


Lezek123:
This seems to be a more complex issue. After the unbonding period is finished (in terms of eras) the funds still seem to be untransferrable and figure as "locked". Perhaps the fact that some eras have been shorter than 600 blocks is affecting the runtime too in a way (it is also possible that some other issue is responsible for that).

Further investigation required...

Lezek123:
Was solved after migration to a newer version of Polkadotjs/apps

PR #524, opened 2020-06-01
Rename discussion module types.

shamil-gadelshin:
Rename Post and Thread types from the discussion module to prevent name interference with forum types. The current Substrate doesn't support namespaces on the frontend part.


bedeho (approved):
Thanks, really apprecite the fast turnaround.

Issue #523, opened 2020-06-01
Proposal disussion system


Lezek123:
**Related issues:**
- https://github.com/Joystream/apps/issues/322
- https://github.com/Joystream/apps/issues/329
- https://github.com/Joystream/apps/issues/338
- https://github.com/Joystream/apps/issues/339
- https://github.com/Joystream/apps/issues/340

**Core:**
- Add missing types: `DiscussionThread` and `DiscussionPost`
- Implement new methods in `joy-utils/transport/proposals` that'll allow fetching Proposal's thread and posts (and maybe add some more friendly type like `ThreadWithPosts`)
- Create components like `ProposalDicussion`, `ProposalDiscussionPost`, `ProposalDiscussionPostForm` etc., integrate them with the transport and add the discussion section on `ProposalDetails` page

**Additional ideas:**
- Implement posts pagination
- Display number of discussion posts in proposal preview box

Lezek123:
Implemented.
Related PRs: https://github.com/Joystream/joystream/pull/530 and https://github.com/Joystream/joystream/pull/621

Issue #365, opened 2020-01-20
Runtime Proposal System Requirements


bedeho:
# Runtime Proposal System Requirements

### Proposal types

There is a finite set of proposal _types_ which can be submitted. Different proposal types will almost always also differ in the associated type of side-effect, and each type is characterised by values for the following properties.

- **Stake:** Whether stake is required to submit proposal, if so how much, and whether its slashable. If a stake exists, then it is also specified what the unbonding period will be (which is the same in all proposal life cycle outcomes), and a possibly non-zero slashing fee associated with withdrawing a proposal, or having it rejected.
- **Approval Quorum:** Percentage of voters that must vote for approval to get approval outcome.
- **Slashing Quorum:** Percentage of voters that must vote for slashing to get slashing outcome.
- **Constitutionality:** How many voting periods in a row which must approve proposal for it to be accepted.
- **Voting period:** The length of a time in a single period where votes may be submitted.
- **Grace period:** The length of of time after a proposal has been approved, before it goes into effect.

The following two lists of proposal types make sense to include after Rome.

**Must have:**
- Pure text/signal
- Mint + transfer
- Runtime upgrade
- Evict storage provider
- Set content working group lead
- Set forum lead
- Set content working group budget
- Set content working lead salary

**More speculative and complex:**
- Change number of validators
- Change number of storage providers
- Set validator reward
- Storage provider reward

### Proposals

Anyone can attempt to submit an instance of any kind of proposal at any time, and the submission is associated with a membership. For each proposal instance actually opened, there will be the following

- **Identifier:** A unique identifier value which identifies this instance across all proposals ever created.
- **Proposer:** Identifier of member proposing.
- **Stake:** Any stake associated with the proposal.
- **Created::** When it was created.
- **Stage**: One among the following.
  - Active: A new proposal that is available for voting, possibly already having passed multiple past councils if the constitutionality of the proposal type is >1. There is a limit to how many proposals can be in this stage at any given time, and if limit is reached, new proposals are not accepted into the system. If the a council period ends, then all votes cast are reset for all active proposals, and the expired period is not counted towards the constitutionality. However, this failure to reach the quorum in this period should not be counted as a failure to satisfy constitutionality constraint.
  - Withdrawn: If withdrawn by a proposer during the active period.
  - Approved: If approval quorum is satisfied in the required number of times in a row. The following substages exist:
    - Grace period: Time between approval and execution.
[...]

PR #530, opened 2020-06-02
Proposals discussion - initial implementation

Lezek123:
This PR introduces the core of proposals discussion module in Pioneer (related issue: https://github.com/Joystream/joystream/issues/523)

**In particular:**
- Adds missing types in `@joystream/types`: `DiscussionThread` and `DiscussionPost`.
_I used the "safe approach" of handling strings which are stored as `Vec<u8>` by the runtime, which is: I defined them as `Bytes` and I handle the conversion to string later using `bytesToText` helper method I created in `joy-utils`. This should prevent bugs like https://github.com/Joystream/apps/issues/419. We could later use this approach for all other similar fields, but it would probably be easier to implement some normalization on the runtime side._
- Adds transport method to fetch proposal discussion (thread and posts) by `proposalId`.
_Because `api.query.proposalsDiscussion.postThreadIdByPostId` is a double map and with `@polkadot/api v0.96.1` we don't have access to helpers like `.keys()` or `.entries()` yet, I implemented something that works similar to the `entries()` method called `doubleMapEntries` inside `joy-utils/transport/base.ts`. It allows us to fetch all posts in a given thread using just the first key of this double map - the `ThreadId` (otherwise we'd have to check all the post ids between `1` and `total_posts_count` one by one in order to retrieve the list of posts in a given thread/proposal which is not optimal to say the least). It's a little hacky solution, because we rely on some particularity with regards to how values inside double maps are currently stored, so I added a visible comment describing this solution and its possible drawbacks, marked as both "TODO" and "FIXME" just in case we'll run into some trouble during the runtime / `@polkadot/api` upgrade._

- Adds a list of discussion posts on proposal details page
- Adds a form allowing to both add a new post and edit an existing one (`DiscussionPostForm`)
Currently editing a post is prevented by runtime though, due to `maxPostEditionNumber` beeing set to `0`.
- Adds markdown styling for discussion posts, proposal rationale and text proposal content.
- Adds some minor visual improvements (ie. to allow users to more easily distinct between the actual proposal parameters and the proposal rationale that could be styled with markdown in such a way that it would resemble proposal parameters).


Lezek123:
Added proposal details desktop layout improvements based on @bedeho's suggestions.
Moving the votes to the right side allows the discussion to be more visible without having to scroll a long way down to reach it.
I also changed the base branch for this PR to be the new `proposal_discussion_release` one.

Lezek123:
Resolved the conflicts and fixed linter issues.

pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -174,4 +180,68 @@ export default class ProposalsTransport extends BaseTransport {
   async subscribeProposal (id: number|ProposalId, callback: () => void) {
     return this.proposalsEngine.proposals(id, callback);
   }
+
+  // Find postId having only the object and storage key
+  // FIXME: TODO: This is necessary because of the "hacky" workaround described in ./base.ts
+  // (in order to avoid fetching all posts ever created)
+  async findPostId (post: DiscussionPost, storageKey: string): Promise<PostId | null> {
+    const blockHash = await this.api.rpc.chain.getBlockHash(post.created_at);
+    const events = await this.api.query.system.events.at(blockHash) as Vec<EventRecord>;
```
mnaamani: You have clearly highlighted that this is a hacky solution, but even if encoding of the double map is unchanged in future, reading old Event data and state in general from a node is not always guaranteed. 

It all depends on how much history it is keeping, ie. the level of state pruning specified with `--pruning` argument. By default it is 256 block. Our public rpc nodes are configured to be archival so it will not be an issue if our endpoints are used. But anyone using a non-archival node as the rpc endpoint results are going to differ.

Of course in future (near future hopefully) once we have a working query node this will be come a moot point.

pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
```diff
@@ -106,32 +135,41 @@ function ProposalDetails ({
...
-        title={ proposal.title }
-        description={ proposal.description }
-        params={ proposal.details }
-        iAmProposer={ iAmProposer }
-        proposalId={ proposalId }
-        proposerId={ proposal.proposerId }
-        isCancellable={ isVotingPeriod }
-        cancellationFee={ proposal.cancellationFee }
-      />
-      { iAmCouncilMember && (
```
mnaamani, line 122: I was surprised to see that any member is able to participate in the discussion, perhaps I mistakenly assumed it was only supposed to be for council members?

mnaamani:
Testing posting works as expected. I modified the `maxPostEditionNumber` and confirmed editing post works.

Only suggestion I have is the layout of the discussion posts.


This is what it looks like when the browser window is wide enough:


Screen Shot 2020-06-04 at 5 03 40 PM

And when on a mobile/tablet with narrower screen it looks like:


Screen Shot 2020-06-04 at 5 03 27 PM

I find the second (narrow screen) view cleaner so maybe we can have the same layout for all screen sizes. 

And perhaps have the date and time to the right of the avatar instead of below it.

These are small details, otherwise its good to go.

pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
Lezek123: Based on https://github.com/Joystream/joystream/issues/365 there were two models of access that were considered:

> (...) Write access as one among two modes, either open, in which case any member can write, or closed, in which case only proposer and council can initially write, but any third party can be invited to write as well.

I think only the _open_ model was actually implemented, so now any member can post in a discussion. Such posts (from members outside of council) are accepted by the runtime, so I wasn't trying to prevent them from the UI either.

pioneer/packages/joy-utils/src/transport/proposals.ts
Lezek123: This is actually the part I didn't predict when started implementing this _hacky solution_, which is: how do I get the `PostId` having only the `Post` struct and the associated storage key, without falling back to the bad practice of making api call for each exiting post id?

I think I wrongly assumed that `api.query.module.method.key()` is performing an api call, when it's actually not, so iterating through it in order to find post id by storage key should not be a problem and the use of events here is actually not needed at all.

I'll fix that.

Lezek123: Fixed in https://github.com/Joystream/joystream/pull/530/commits/4c9224616c6725b8881fc7c7e99392cd304273a7

mnaamani: Excellent!
Checked.. working correctly.

pioneer/packages/joy-utils/src/transport/base.ts
```diff
@@ -38,4 +43,50 @@ export default abstract class BaseTransport {
...
+    for (let secondKey = secondKeyStart; secondKey <= maxSecondKey; ++secondKey) {
+      const storageKey = this.queryMethodByName(methodName).key(firstKey, secondKey);
+      storageKeyToSecondKey[storageKey] = secondKey;
+    }
+
+    // Create the resulting entries array
+    const entries: { secondKey: number; value: T }[] = [];
+    for (const key of allEntriesStorageKeys) {
+      const value: any = await this.api.rpc.state.getStorage(key);
+      if (typeof value === 'object' && value !== null && value.raw) {
```
mnaamani, line 82: In the past when playing with reading directly from storage I found that basically if you look at the raw bytes, if the first byte is 0 then the key/value pair is not in storage. If it is 1 then it exists and the remaining bytes are the raw data. Is there no "shortcut" property on the return value to do this check in a type safe way? No need to change here just wondering for future.

mnaamani (approved):

PR #533, opened 2020-06-02
Content directory: Extrinsics success cases coverage

iorveth:
1. `mock` setup
1. Basic extrinsics success cases coverage 
1. Make separate files for each extrinsic test coverage
1. Minor logic fixes


iorveth:
Here you also can find, that `entity_created_in_operation` representation in `transaction` extrinsic was reverted back to `BtreeMap`. During test coverage, i figured out, that we do need to know an exact `operation_id`, under which given `Entity` was previously created

bedeho (approved):
Nice work, LGTM.

PR #552, opened 2020-06-04
Simplify add_schema_support_to_entity extrinsic

### LINKED ISSUE #545: Simplify add_schema_support_to_entity extrinsic
Make a version, that has no mutable variables, following #531.

### LINKED ISSUE #560: Complete reference counting logic simplification

iorveth:
Closes #545
Closes #560

### DIFF
--- a/runtime-modules/content-directory/src/helpers.rs
+++ b/runtime-modules/content-directory/src/helpers.rs
@@ -43,10 +43,10 @@ impl InputValidationLengthConstraint {
 }
 
 /// Structure, respresenting inbound entities rc mappings to their respective count for each `entity_id`
-pub struct InboundEntitiesRc<T: Trait>(BTreeMap<T::EntityId, ReferenceCounter>);
+pub struct InboundEntitiesRc<T: Trait>(BTreeMap<T::EntityId, InboundReferenceCounter>);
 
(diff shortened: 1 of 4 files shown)


bedeho (changes requested):
- This looks better but you are still not following the guideline from last extrinsics, avoid mutables when possible, e.g. in `get_involved_entity_rcs`, probably other places also.

- I think you also need to declare variables as close as possible to where they are needed
e.g. entity_ids_to_increase_rcs declared early, not used for many lines

- `EntitiesRc` has no clear semantic representation, what is this? The name and scope of responsibilities need to be clarified, as well as making sure the representation is as safe as possible if you are going to keep using the maps. Some of the methods interact with similarly named methods on `Module`, for reasons that seem hard to understand quickly.

iorveth:
- we only have 1 mutable variable for helper structure in the local scope of function, that returns it changed performing single clear operation after all security checks performed.)
I do not understand why this considered bad approach. I can rewrite without mutables at all, but seems, that we sacrifice clarity here.
- Done
- I am going to rework `EntitiesRc` as discussed in the chat, though it involved in other extrinsic s  as well. IMHO, It will be appropriate to rework related logic in separate PR.

bedeho:
Superceeded by https://github.com/Joystream/joystream/pull/655.

PR #557, opened 2020-06-05
Cross-enitity support for text searches

dzhelezov:
The text search queries in this PR are defined as follows:
```gql
type A {
   field1: String @fulltext(query: "myquery"),
   field2: String @fulltext(query: "myquery")
}

type B {
   field3: String @fulltext(query: "myquery"),
   field4: String @fulltext(query: "myquery")
}
```
This will generate a query with return type `SearchResult[]` roughly defined as 
```gql
type SearchResult {
   item: A | B,
   rank: number, 
   highligt: String
}

type Query {
   myquery(text: String, limit?: Int): SearchResult[]
}
```

This PR introduces:
- [x] Refactoring for the `SchemaParser` and `WarthogModel` to minimize the dependency on the directive semantics
- [x] Cross-entity and cross-field support for text searches
- [x] Support for highlights: the API search queries support an additional `highlight` field returning the matched text fragment
- [x] (WIP) Support for field weights: the weights are fully integrated into the model. However, the query definition parser does not support weights yet, as it will be refactored in the future to match The Graph-style query definitions.


bedeho (approved):
Nice work!

I assume our current approach precludes a given entity+field combination from being used in more than one free-text search query? I don't think its a big deal, but I guess that would be an additional argument for having standalone FTS search declaration.

dzhelezov:
> Nice work!
> 
> I assume our current approach precludes a given entity+field combination from being used in more than one free-text search query? I don't think its a big deal, but I guess that would be an additional argument for having standalone FTS search declaration.

Technically, multiple directives on a single field will work as long as the SDL validation flag is switched off (there's even a unit test for that, but I commented it out). Standalone FTS declarations definitely look like a more natural approach though.

metmirr (approved):

bedeho:
Will merge when CI passes.

PR #563, opened 2020-06-05
Merge proposal discussion release work into development

mnaamani:
Merge back any work done on release into development


Lezek123 (approved):
Looks good, I think we'll also need to update the version of `@joystream/types` because of the changes in `types/src/proposals.ts` and `types/src/versioned-store/EntityCodec.ts` (should it be done in another PR to `development` or perhaps to `proposal_discussion_release` before the merge?)

mnaamani:
> Looks good, I think we'll also need to update the version of `@joystream/types` because of the changes in `types/src/proposals.ts` and `types/src/versioned-store/EntityCodec.ts` (should it be done in another PR to `development` or perhaps to `proposal_discussion_release` before the merge?)

Great catch. Yes lets update it on the release branch before merge. I will also publish to npm. We will have to update all dependents (storage node, status server, any other production servers?)

mnaamani:
Updated types in #619

PR #619, opened 2020-06-08
@joystream/types - publish v0.10.0

mnaamani:
Updated and published `@joystream/types` to npm.

yarn package alias using `"@constantinople/types@npm:@joystream/types": "^0.10.0"` used in network integration tests  forces us to publish package which is not ideal while developing the tests. @gleb-urvanov is looking into a workaround, worst case we just drop the alias (but keep it for the older network tests)

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -8,7 +8,7 @@
   },
   "bugs": "https://github.com/Joystream/substrate-runtime-joystream/issues",
   "dependencies": {
-    "@joystream/types": "^0.9.1",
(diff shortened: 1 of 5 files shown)


Lezek123 (approved):
Tested after reinstalling packages with yarn.
Everything seems to work correctly (Pioneer, CLI, network tests).
All changes in the types workspace are still reflected when using `yarn workspace pioneer start` and in the CLI after running `yarn workspace @joystream/types build`.

Issue #458, opened 2020-05-23
Pioneer v1 UX: Stop using tabs for actions


bedeho:
As part of #400:

In many sections, tabs are used to get to screens for initiating actions, this makes no sense, e.g. forum here

<img width="369" alt="Screenshot 2020-05-23 at 12 25 05" src="https://user-images.githubusercontent.com/437292/82728470-8d9a3880-9cf0-11ea-86c4-b01c6f8bfee3.png">

Instead, lift top-level actions out as menu buttons over any possible tabs, if there are still tabs remaining. In many cases I am sure you really would end up with just a single screen, hence there would be no need for tabs.

Make a list of all such places, and fix them.

kdembler:
List of affected tabs:
- Media -> New channel
- Forum -> New category
- My Keys -> Vanity address (potentially rename to "Generate vanity address")
- Address Book -> View memo
- Council -> Reveal a vote
- Proposals -> New proposal

I've excluded tabs that are not exclusively an action, for example Membership -> Edit my profile as this view can also serve as a way to just view current user's profile

I've also skipped views from the "Fully featured" mode, my guess is we don't really care about UX of these, my understanding is they're mostly for internal use

PR #627, opened 2020-06-08
Proposals view tweaks

### LINKED ISSUE #461: Pioneer v1 UX: Proposal state tab selector
As part of #400:

This is a very bad way to show different states

Screenshot 2020-05-22 at 15 22 38

Not only because there are so many, but also the way the number is shown. 

Last, but perhaps not least, two levels of tabs is not that great.

kdembler:
This PR tries to tackle some of the UX issues with proposals view. Firstly, it changes proposals to be sorted starting from the newest. Secondly, filter tabs are visually tweaked to better present count of proposals. 

#461 also mentions two levels of tabs as a potential issue. However, this can be tackled as part of #458 - after that task, the top-level tabs should go away. I believe tabs for filters will look good after that.

View after changes:
Screen Shot 2020-06-08 at 17 20 31

Resolves #461

### DIFF
--- a/pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
+++ b/pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
@@ -55,7 +55,7 @@ type ProposalPreviewListProps = {
 const FilterContainer = styled.div`
   display: flex;
   justify-content: flex-end;
-`
+`;
 const FilterOption = styled.span`
   display: inline-flex;
   align-items: center;
--- a/pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
+++ b/pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
@@ -57,6 +57,7 @@ function ProposalPreviewList ({ bestNumber }: ProposalPreviewListProps) {
 
   const proposalsMap = mapFromProposals(proposals);
   const filteredProposals = proposalsMap.get(activeFilter) as ParsedProposal[];
+  const sortedProposals = filteredProposals.sort((p1, p2) => p2.id.cmp(p1.id));
 
   return (
     <Container className="Proposal">
@@ -72,9 +73,9 @@ function ProposalPreviewList ({ bestNumber }: ProposalPreviewListProps) {
           ))}
         </Menu>
         {
-          filteredProposals.length ? (
+          sortedProposals.length ? (
             <Card.Group>
-              {filteredProposals.map((prop: ParsedProposal, idx: number) => (
+              {sortedProposals.map((prop: ParsedProposal, idx: number) => (
                 <ProposalPreview key={`${prop.title}-${idx}`} proposal={prop} bestNumber={bestNumber} />
               ))}
             </Card.Group>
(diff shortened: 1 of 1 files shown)


pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
```diff
@@ -50,13 +51,34 @@ type ProposalPreviewListProps = {
...
+`;
+
 function ProposalPreviewList ({ bestNumber }: ProposalPreviewListProps) {
   const transport = useTransport();
   const [proposals, error, loading] = usePromise<ParsedProposal[]>(() => transport.proposals.proposals(), []);
   const [activeFilter, setActiveFilter] = useState<ProposalFilter>('All');
 
   const proposalsMap = mapFromProposals(proposals);
   const filteredProposals = proposalsMap.get(activeFilter) as ParsedProposal[];
+  const sortedProposals = filteredProposals.sort((p1, p2) => p2.id.cmp(p1.id));
```
fcescob, line 96: Might be worth to move this to `mapFromProposals` so we do it only once and have them sorted by newest automatically.

kdembler: It won't change the number of sorts though, `mapFromProposals` is called on every render, same as `sort` in the code as it is. With the current approach, we may actually do less sorting because we only sort the proposals for the currently selected filter, in `mapFromProposals` we would sort all the proposals. That's a micro-optimization though 😄

kdembler: If you think that's better for readability though, let me know and I will change it.

bedeho (changes requested):
The counter bubble + top level tab removal helps, I agree, but I dont think having 7 tabs with abstract labels is a good UX. Its immedieatly not clear to the user that these are proposal states, and we are using a lot of visual space on something which people rarely will engage with (all the different state variations). Moreover, we may want to actually have tabs for some use scene selection later, which would reintroduce the double tab issue, and lastly.

So I recommend trying another variation for selecting the proposal state that better addresses these issues.

pioneer/packages/joy-proposals/src/Proposal/ProposalPreviewList.tsx
fcescob: My bad! In terms of readability, your version is more explicit, so let's keep it that way

kdembler: :+1:

kdembler:
Got rid of the tabs in favor of a dropdown with all the states. It takes roughly the same amount of space as the tabs, but it's mostly empty space so it shouldn't be as prominent visually. It should also be more clear what the intention is. @bedeho please let me know if you think that's better

Screen Shot 2020-06-10 at 12 21 27

Screen Shot 2020-06-10 at 12 33 10

kdembler:
After all, moved the dropdown to the right side of the screen so it takes less space:

Screen Shot 2020-06-10 at 13 40 17

This way it's also consistent with the category dropdown in the new proposal view.

fcescob (approved):

fcescob (approved):

PR #635, opened 2020-06-09
Network tests: ansible playbook to run node and tests

gleb-urvanov:
Contains ansible playbooks to run tests against network made of one or two nodes:
run-tests-single-node-playbook.yml
run-tests-two-nodes-playbook.yml

Also includes playbook to build testing node and run tests against single node:
build-and-run-tests-single-node-playbook.yml


devops/ansible/install-dependencies-playbook.yml
```diff
@@ -0,0 +1,15 @@
+- hosts: 127.0.0.1
+  user: root
+  become: yes
+  become_method: sudo
+  tasks:
+    - name: install pip
+      apt: name=python-pip state=present
```
mnaamani, line 7: `apt` is not available on mac os or windows, so we need a platform neutral way to install python and pip.

I get this error when running single node tests playbook:
```
MokoBook:joystream mokhtar$ ansible-playbook devops/ansible/build-and-run-tests-single-node-playbook.yml --ask-become-pass
BECOME password: 
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [127.0.0.1] **************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************
ok: [127.0.0.1]

TASK [install pip] ************************************************************************************************************************************
[WARNING]: Updating cache and auto-installing missing dependency: python3-apt
fatal: [127.0.0.1]: FAILED! => {"changed": false, "cmd": "apt-get update", "msg": "[Errno 2] No such file or directory: b'apt-get'", "rc": 2}

PLAY RECAP ********************************************************************************************************************************************
127.0.0.1                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

devops/ansible/install-dependencies-playbook.yml
```diff
@@ -0,0 +1,15 @@
+- hosts: 127.0.0.1
+  user: root
+  become: yes
+  become_method: sudo
```
mnaamani, line 4: I haven't tested on windows, but would this be a problem running ansible-playbook on a windows machine? Don't bother too much if it needs workaround, running on linux and macos is sufficient I think.

tests/network-tests/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "GPL-3.0-only",
   "scripts": {
     "build": "tsc --build tsconfig.json",
-    "test": "tap --files ts-node/register src/constantinople/tests/proposals/*Test.ts",
+    "test": "tap --files ts-node/register src/constantinople/tests/proposals/*Test.ts -T",
```
mnaamani, line 7: what does the `-T` argument do?

mnaamani (changes requested):

mnaamani:
Lets try to remove the need to build a separate image for the purpose of testing. One docker image is sufficient. The only difference between the 'production' image and the testing image is the need to provide a separate chainspec file. We can run the `build-spec` command from the original image and the perl script can transform it. Using docker volumes the chainspec file for testing produced on the host machine can be used inside the running container. The perl script can be refactored to make it more reusable by removing the explicit call to build-spec file and it can read the original spec file on stdin and output on stdout.

I'm building manually the images now anyway to test running the ansible playbook, and will report my experience :)

mnaamani:
Additional point regarding the single image for the node. I'm primarily talking about having a single `Dockerfile` that works for both. We can still ofc tag the build separately. In future when we a have configurable runtime (using env variables) to customize the build, the same `Dockerfile` can still be used after updating it to pass env variables to the `cargo build` command.

mnaamani:
I get an error when running:

```sh
sh scripts/build-joystream-testing-node-docker-image.sh
```

```
   Compiling substrate-service v2.0.0 (https://github.com/paritytech/substrate.git?rev=c37bb08535c49a12320af7facfd555ce05cce2e8#c37bb085)
   Compiling substrate-cli v2.0.0 (https://github.com/paritytech/substrate.git?rev=c37bb08535c49a12320af7facfd555ce05cce2e8#c37bb085)
    Finished release [optimized] target(s) in 37m 23s
Removing intermediate container 802162a3258f
 ---> e65f57e2d8b2
Step 6/18 : RUN /joystream/scripts/create-test-chainspec.sh
 ---> Running in 679785d1a714
/bin/sh: 1: /joystream/scripts/create-test-chainspec.sh: Permission denied
The command '/bin/sh -c /joystream/scripts/create-test-chainspec.sh' returned a non-zero code: 126
```

maybe check-in the files with execute permisions? `chmod +x scripts/*.sh`

devops/ansible/install-dependencies-playbook.yml
gleb-urvanov: Ansible does not have support for windows. As a workaround, VM solutions could be used for windows.

tests/network-tests/package.json
gleb-urvanov: `-T` removes timeout. Default timeout may run out during application startup. The actual timeout is set up in the beginning of each testing scenario.

gleb-urvanov:
As discussed, I will rework the code to use a single Dockerfile to create both production and testing nodes.

Issue #639, opened 2020-06-09
Circular dependencies in @joystream/types


Lezek123:
Recentlly when solving an issue in [joystream-api-examples](https://github.com/Joystream/joystream-api-examples) that arised after updating the `@joystream/types` to a new version there, I noticed that there are a lot of problematic circular dependencies in `@joystream/types`.

This basically means that files are importing from each other in a very chaotic way, making it almost impossible to predict whether ie. the classes they are trying to import will already be defined when they need to be.

This sometimes results in errors like:
```
Class extends value undefined is not a constructor or null
```

In places where some module imports a class from other module and tries to extend it, ie.:
```
// ApplicationId is imported from "../hiring",
// but not initialized yet because of some circular dependencies

export class WorkerApplicationId extends ApplicationId { }; // Error :(
```

I found an article describing an interesting approach to solve the circular dependency problem:
https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

The solution described here is to create a file called `internal.js` which both imports and exports everything we would like to import between modules. The modules then import __only__ from this file, therefore specifying the correct order in `internal.js` should prevent us from running into this kind of errors once and for all.

I haven't tested this solution yet, but it seems very reasonable and could potentially help us avoid a lot of problems when working with / developing the `@joystream/types` library.

mnaamani:
Sounds like it works, and I think I have seen this pattern used by the polkadot-js/api types package itself.

Lezek123:
Fixed in https://github.com/Joystream/joystream/pull/687
CI checks added in https://github.com/Joystream/joystream/pull/752 and updated in https://github.com/Joystream/joystream/pull/963

PR #567, opened 2020-06-05
Storage working group runtime changes

shamil-gadelshin:
## Implemented changes
- [Data directory module](https://github.com/Joystream/joystream/issues/566)
- [Data object storage registry module](https://github.com/Joystream/joystream/issues/565)
- [Data object type registry module](https://github.com/Joystream/joystream/issues/564)
- [Service discovery module](https://github.com/Joystream/joystream/issues/569)
- [Bureaucracy module](https://github.com/Joystream/joystream/issues/571)
    - implemented only: add method to enumerate workers - it is required to generate a random storage provider.
- [Runtime](https://github.com/Joystream/joystream/issues/573)
    - remove LookUpRoles
    - implement StorageProviderHelper - to allow picking a random provider

[Root issue is here](https://github.com/Joystream/joystream/issues/544).

PR #640, opened 2020-06-09
Update @joystream/types for Nicaea

Lezek123:
This PR introduces the initial type definitions for the Nicaea release.
After those changes `@joystream/types` library should be compatible with the runtime based on the current `bureaucracy` branch (see: https://github.com/Joystream/joystream/pull/325).

**Circular dependency issues**
To avoid the circular dependency problem described in https://github.com/Joystream/joystream/issues/639 I decided to temporarly use the `require` statement to include the new `bureaucracy/index.ts` file at the very end of the root `/index.ts` file (which assures me that everything else is already loaded at this point). It's not a perfect solution, but until the `@joystream/types` library is restructured to avoid circular dependency issues in general, there's probably not too much that can be done here other than this or randomly changing the imports order until it starts to work.

**Error in joy-roles' `Admin.controller.tsx`** 
I also noticed an error in `joy-roles/src/tabs/Admin.controller.tsx` regarding the usage of incorrect ids in a mapping (`openingId` instead of `curatorOpeningId`). It went unnoticed before, because we've only had one module that was taking advantage of the `hiring` module, therefore those ids had always been identical, but after adding some openings via a module other than `contentWorkingGroup` (`forumBureaucracy` in that case) `workig-groups/admin` tool was not working correctly anymore until this fix was introduced.

**`types/.editorconfig`**
I also added the `types/.editorconfig` file (which was missing after moving the `types` folder away from the `pioneer` folder), because otherwise this workspace was catching the root folder `/.editorconfig` which caused VSC to incorrectly auto-indent the files inside `/types`.

**Flat namespace issues**
Due to flat namespaces in `@polkadot/api`, which have already caused us many problems, the current structure of registered types is very fragile (although it works as is). Ie. I created a type called `BureaucracyOpeningPolicyCommitment` which represents the `OpeningPolicyCommitment` from `/runtime-modules/bureaucracy/src/types.rs`, that is slightly different than the one in `/runtime-modules/content-working-group/src/lib.rs`.

Even though I declared the `policy_commitment` field of `WorkerOpening` to be of `BureaucracyOpeningPolicyCommitment` type, the `content-working-group`'s `OpeningPolicyCommitment` type is still used when sending the `forumBureaucracy.addWorkerOpening` extrinsic (which can be observed via the "Extrinsics" tab). Therefore when sending this extrinsics we have field names like `terminate_curator_role_stake_unstaking_period` instead of `terminate_worker_role_stake_unstaking_period`.

This doesn't seem to cause any bigger issues currently, probably because the fields in both those structs have exactly the same types, but it may become an issue if they start to differ in a more meaningful way.

### DIFF
--- a/pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
@@ -689,7 +689,7 @@ export class AdminController extends Controller<State, ITransport> {
         title = (hrt).job.title;
       }
 
-      this.state.openings.set(openingId.toNumber(), {
+      this.state.openings.set(i, {
         openingId: openingId.toNumber(),
         curatorId: i,
         applications: new Array<application>(),
--- a/types/.editorconfig
+++ b/types/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+[*]
+indent_style=space
+indent_size=2
+tab_width=2
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+max_line_length=120
+insert_final_newline=true
--- a/pioneer/packages/joy-forum/src/Context.tsx
+++ b/pioneer/packages/joy-forum/src/Context.tsx
@@ -2,7 +2,8 @@
 // NOTE: The purpose of this context is to immitate a Substrate storage for the forum until it's implemented as a substrate runtime module.
 
 import React, { useReducer, createContext, useContext } from 'react';
-import { Category, Thread, Reply, ModerationAction, BlockchainTimestamp } from '@joystream/types/forum';
+import { Category, Thread, Reply, ModerationAction } from '@joystream/types/forum';
+import { BlockAndTime } from '@joystream/types/media';
 import { Option, Text, GenericAccountId } from '@polkadot/types';
 
 type CategoryId = number;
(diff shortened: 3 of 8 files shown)


pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
```diff
@@ -689,7 +689,7 @@ export class AdminController extends Controller<State, ITransport> {
         title = (hrt).job.title;
       }
 
-      this.state.openings.set(openingId.toNumber(), {
+      this.state.openings.set(i, {
```
mnaamani, line 692: Nice catch. I think we caught an identical bug at one point in the runtime itself for the same exact reason.

types/.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+[*]
+indent_style=space
+indent_size=2
+tab_width=2
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+max_line_length=120
+insert_final_newline=true
```
mnaamani, line 10: I think we should have a common style across all the projects in the repo. @shamil-gadelshin and I briefly discussed this today. If we can pick a standard that is also matching with `cargo-fmt` for the rust code we can update just the `.editorconfig` at the root

mnaamani (changes requested):
The issue with the name conflict for type `OpeningPolicyCommitment` is unfortunate but it doesn't warrant changing names in the runtime I suppose. At some point we will replace the content working group module with an instance of a Buraucracy module so this issue will go away. 

But it might be prudent to move the identical types that are currently common to both these modules into a separate file highlighting this issue. This applies to `SlashableTerms`, `SlashingTerms` as well. Or at least add comments about above each type declaration.

Lezek123:
I added a few changes to make the types compatible with the runtime after this PR: https://github.com/Joystream/joystream/pull/567 

- Replaced `BlockchainTimestamp` with `BlockAndTime`
- Added missing `StorageProviderId` type
- Added a few helper-getters to some structs (which I'm using in my later work, but since those changes are inside the `@joystream/types` workspace I dedecided to add them to this PR)

PR #642, opened 2020-06-09
Reorder sidemenu.

bedeho:
Let's try this for the first time then!

Here are the original PRs: 
 - https://github.com/Joystream/joystream/pull/620
 - https://github.com/Joystream/joystream/pull/637

Basically a trivial reordering of tabs in order to show the most salient stuff first, as you see, its also been reviewed by @Gamaranto. 

This change would later be absorbed into Nicaea, and also into development, correct?


mnaamani:
Looks good.

> This change would later be absorbed into Nicaea, and also into development, correct?
Yes correct.

PR #657, opened 2020-06-10
WIP: Query node graph plugin

dzhelezov:
This is a WIP on replacing the direct warthog calls with a plugin exposed through a gluegun CLI interface (+ additional linter fixes)


dzhelezov:
This PR should be closed as we've abandoned the plugin approach

Issue #668, opened 2020-06-11
update_entity_permissions extrinsic simplification


iorveth:
Refactor `update_entity_permissions` extrinsic following #531

PR #655, opened 2020-06-10
update_entity_property_values extrinsic simplification, fixes to previous PR comments

iorveth:
The scope of this PR includes:
1. `update_entity_property_values` extrinsic simplification, following the spirit of #531   

1. Additional `Schema` `ensure_property_value_is_valid_reference` method simplification    
    
1. Fixes to PR comments mentioned by @bedeho:    
- [x] Ensure checks in `add_schema_support_to_entity` & `update_entity_property_values` extrinsics divided into much smaller steps to improve perception.

- [x] Additional checks for `add_schema_support_to_entity` extrinsic to force it fail loudly, when any of provided `property_values` is invalid. (Already added to the current `Entity` `property_values` mapping or does not a member of specified `Schema`).
This change also simplified further workflow a lot, as we do not keep in mind, that some of provided `property_values` can possibly be invalid.

- [x] Reference counting logic simplification:
Provided implementation introduces `InboundReferenceCounterDelta` and `EntitiesInboundRcsDelta` helper structures, which implement much safer approach to work with `Entities` `InboundReferenceCounter` logic.
```
/// Enum, used to specify, which mode of operation should be chosen,
/// when calling `fill_in_inbound_entities_rcs` method on the instance of `EntitiesInboundRcsDelta`
#[derive(Clone, PartialEq, Eq, Copy, Debug)]
pub enum DeltaMode {
    Increment,
    Decrement,
}

/// Representing delta on which respective `InboundReferenceCounter` should be changed.
/// Direction of performed operation depends on `DeltaMode`.
#[derive(Default, Clone, PartialEq, Eq, Copy, Debug)]                      
pub struct InboundReferenceCounterDelta {
    pub reference_counter: InboundReferenceCounter,
    pub delta_mode: DeltaMode,
}

/// Structure, respresenting `entity_id` mappings to their respective `InboundReferenceCounterDelta`
pub struct EntitiesInboundRcsDelta<T: Trait>(BTreeMap<T::EntityId, InboundReferenceCounterDelta>);
```
Closes https://github.com/Joystream/joystream/issues/616
Depends on: #562, #552, #615

PR #615, opened 2020-06-08
Clear/Remove at entity property vector extrinsics simplification

iorveth:
This PR aims to refactor `remove_at_entity_property_vector`, `clear_entity_property_vector` extrinsics following the spirit of https://github.com/Joystream/joystream/issues/531

Depends on: https://github.com/Joystream/joystream/pull/562,       https://github.com/Joystream/joystream/pull/552

Closes: https://github.com/Joystream/joystream/issues/614
Closes: https://github.com/Joystream/joystream/issues/613

PR #562, opened 2020-06-05
Simplify insert_at_entity_property_vector extrinsic  

iorveth:
Closes https://github.com/Joystream/joystream/issues/561
Depends on https://github.com/Joystream/joystream/pull/552

PR #669, opened 2020-06-11
Megafix

iorveth:
# New description

This PR will contain review for all unmerged work of the following PRs

-  #552
- #562
- #615
- #655

They will be closed.

# Old description

1. `update_entity_permissions` extrinsic refactoring,  following #531 https://github.com/Joystream/joystream/pull/669/commits/fada3bef0b42ea598652c35caf8fe78dfc073cc7
1. `make_updated_entity_property_values` functionality fix https://github.com/Joystream/joystream/pull/669/commits/cb809efde3671a2d2e672a6061ca58fc120cfe1b

Closes https://github.com/Joystream/joystream/issues/668
Depends on: #562, #552, #615, #655


runtime-modules/content-directory/src/schema.rs
```diff
@@ -541,6 +543,21 @@ impl Schema {
         &self.properties
     }
 
+    /// Ensure keys of provided `property_values` are valid indices of current `Schema`
+    pub fn ensure_property_value_indices_are_valid<T: Trait>(
```
bedeho: `ensure_has_properties`

runtime-modules/content-directory/src/lib.rs
```diff
@@ -426,6 +326,61 @@ impl<T: Trait> Class<T> {
     }
 }
 
+/// Structure, respresenting inbound entity rcs for each `Entity`
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Copy, Debug)]
+pub struct InboundReferenceCounter {
```
bedeho, line 336: Lets just drop this separate counter struct and just embed the two fields directly in `Entity`, it does not do anything for us.

runtime-modules/content-directory/src/helpers.rs
```diff
@@ -0,0 +1,174 @@
...
+            Err(too_short_msg)
+        } else if length > self.max() {
+            Err(too_long_msg)
+        } else {
+            Ok(())
+        }
+    }
+}
+
+/// Enum, used to specify, which mode of operation should be chosen,
```
bedeho: As mentioned in the call, this construction does not clearly work, because we cannot guarantee that all operations have side effects that leave both counters with the same sign, nor do we want to limit ourselves to this in the near future. Lets instead do the simplest approach

```Rust

/// Net side effect on a given entity from some operation
struct EntityReferenceCounterSideEffect {
	total: i128
	sameOwner: i128
}

/// The net side effect on a set of entities from some operations.
type ReferenceCounterSideEffects = BTreeMap<EntityId, EntityReferenceCounterSideEffect>
```

`ReferenceCounterSideEffects` and `EntityReferenceCounterSideEffect` can be used as event playloads, as required in https://github.com/Joystream/joystream/issues/656

bedeho (changes requested):
As mentioned, I will not do a full review now, as some big changes will be needed, and the workflow failure made the PR very big, after fixing the issues in this review iteration, we will also absorb changes in a branch which fixes broken `transfer_entity_ownership` semantics.

1.  Add a full human readable description to every check and mutation, having method comments is no substitute because the reader cannot keep looking those up when building their mental model of the execution, so instead of 

```
schema.ensure_property_value_indices_are_valid(&property_values)?;
```
do

```
// Ensure property values are for properties in the given schema
schema.ensure_property_value_indices_are_valid(&property_values)?;
```

2. Side effects can be computed  more simply by the following sort of approach

```Rust

struct EntityReferenceCounterSideEffect {
	total: i128
	sameOwner: i128
}

type ReferenceCounterSideEffects = BTreeMap<EntityId, EntityReferenceCounterSideEffect>

fn compute_add_schema_support_to_entity_reference_counter_sideffects(property_values: BTreeMap<PropertyId, PropertyValue<T>>, class_properties: Vec<Property<T>>) -> ReferenceCounterSideEffects {
	
	// iterate over all property values
	propety_values
	.iter()
	.cloned()
	// if property creates sideffect, we map it, otherwise we filter
	.filter_map(|property_id, property_value| -> (EntityId, EntityReferenceCounterSideEffect) {

		match poperty_value {
			Single(singleValue) => {

				match singleValue.value {
					Reference(entity_id) => Some(make_sideffect(entity_id, class_properties[property_id])) // make_sideffect is just introduced to keep example small
					_ => None

				}

			}
			Vector => { /* ... same idea  ... */}
		}


	})
	// now aggeregate all sideffects on a single entity together into one side effect map
	.fold()

}
```

3. We keep having to pass around a map of property values and a vector of properties on a class, and lookup in the latter based on the former, even after doing integrity checks on the property ids being valid. This means we do not get strong compile time gaurantees about integrity of provdied property ids through execution lifetime of most extrinsics, and its also hard on the reader to see the vector and map travel together, but separately, many places. Instead we can compute something like this after the initial check in an extrinsic

```Rust

struct ValueForExistingProperty {
 property: Property<T>,
 value:  PropertyValue<T>,
}

BTreeMap<PropertyId, ValueForExistingProperty>
```

runtime-modules/content-directory/src/helpers.rs
iorveth: Good point. Done here https://github.com/Joystream/joystream/pull/669/commits/0121eb25dfcdf979b47aea28225606d132d29469

runtime-modules/content-directory/src/lib.rs
iorveth: That only can be achieved, if we'll use  `EntityReferenceCounterSideEffect` structure instead of  `Entity`  s `ReferenceCounter` representation. Though, `Entity` `ReferenceCounter` values should always be unsigned.

runtime-modules/content-directory/src/schema.rs
iorveth: Done https://github.com/Joystream/joystream/pull/669/commits/3e4e20a8a685c970988b8a177aabc13747b82662

iorveth:
1. Add a full human readable description to every check and mutation, having method comments is no substitute because the reader cannot keep looking those up when building their mental model of the execution

Here is an attempt to do as required https://github.com/Joystream/joystream/pull/669/commits/a091668723fe72feed441f39cd3ce71e3c8fa17f. 

2. Side effects can be computed more simply by the following sort of approach

Done here https://github.com/Joystream/joystream/pull/669/commits/0121eb25dfcdf979b47aea28225606d132d29469

3. We keep having to pass around a map of property values and a vector of properties on a class, and lookup in the latter based on the former, even after doing integrity checks on the property ids being valid. This means we do not get strong compile time gaurantees about integrity of provdied property ids through execution lifetime of most extrinsics, and its also hard on the reader to see the vector and map travel together, but separately, many places. Instead we can compute something like this after the initial check in an extrinsic

Made this kind of wrapper here, but on references. Did not sure, if cloning such a big values in some places is considered a good approach.
https://github.com/Joystream/joystream/pull/669/commits/9fb1dff020b43cc3aac3c0d1e392c72ff9e0c215
https://github.com/Joystream/joystream/pull/669/commits/701b4feafaf6050ce34a1e17704571f2e0915a9d

runtime-modules/content-directory/src/helpers.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// Retrieve `Property` and `PropertyValue` references
+    pub fn unzip(&self) -> (&Property<T>, &PropertyValue<T>) {
+        (self.0, self.1)
+    }
+}
+
+/// Mapping, used to represent `PropertyId` relation to its respective `ValueForExistingProperty` structure
+pub struct ValuesForExistingProperties<'a, T: Trait>(
```
bedeho, line 25: Why is this a `struct` rather than just a type definition?

`type ValuesForExistingProperties = BTreeMap ..`

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1198,37 +1268,62 @@ decl_module! {
...
+
+            // Ensure Class Schema under given index exist, return corresponding Schema
+            let schema = class.ensure_schema_exists(schema_id)?.to_owned();
 
+            // Ensure provided schema can be added to the Entity
             schema.ensure_is_active()?;
 
-            let entity_values = entity.get_values_ref();
+            // Ensure all provided property values are for properties in the given schema
+            schema.ensure_has_properties(&property_values)?;
```
bedeho: A `ValuesForExistingProperties` shuold be computed in this check already, because we are here making sure that such an object an exist.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1198,37 +1268,62 @@ decl_module! {
...
 
-            for prop_id in schema.get_properties().iter() {
-                if entity_values.contains_key(prop_id) {
-                    // A property is already added to the entity and cannot be updated
-                    // while adding a schema support to this entity.
-                    continue;
-                }
+            // Perform all required checks to ensure property_values under provided schema property_ids are valid
+            Self::ensure_property_values_are_valid(
+                &class_properties, schema.get_properties(), entity_controller, &property_values,
```
bedeho: The whole point of `ValuesForExistingProperties` was to avoid what is still being done here in this routine.

Please make sure to use it everywhere it may be of use, which is not only for this method in many other extrinsics, but probably also other unrelated methods.

bedeho: Why do you need to pass `schema.get_properties()`, even without the `ValuesForExistingProperties` change?

At this point, we have ensured that `property_values` only has values for properties in the schema, why do we need to refer back to it as a separate input here?

bedeho: Why does this method do lots of combined checks for validity of new values, but this is its own standalone check:
`ensure_property_values_unique_option_satisfied`? It seems that if there is one validity call should cover them all, however the call should of course do them in separate steps for clarity, as you already do in `ensure_property_value_to_update_is_valid`.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1032,38 +1031,99 @@ decl_module! {
...
+                    Self::make_updated_property_value_references_with_same_owner_flag_set(
+                        entity_property_id_references_with_same_owner_flag_set, &entity_property_values,
+                        &new_property_value_references_with_same_owner_flag_set,
+                    ) {
+
+                        // Create wrapper structure from provided entity_property_values_updated and their corresponding Class properties
+                        let values_for_updated_properties = ValuesForExistingProperties::from(&class_properties, &entity_property_values_updated);
+
+                        // Traverse all values_for_updated_properties to ensure unique property satisfied (if required)
+                        Self::ensure_property_values_unique_option_satisfied(values_for_updated_properties)?;
```
bedeho: As per comment above, why is this on its own outside of `ensure_property_values_are_valid` here?

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1689,88 +1858,311 @@ impl<T: Trait> Module<T> {
...
+
+            // Ensure all PropertyValue's with unique option set are unique, except of null non required ones
+            property
+                .ensure_unique_option_satisfied(value, &updated_values_for_existing_properties)?;
+        }
+        Ok(())
+    }
+
+    /// Perform all necessary checks to ensure `new_values_for_existing_properties` are valid
+    pub fn ensure_new_property_values_are_valid(
```
bedeho: Is this needed, what is the meaning of it vs. `ensure_property_values_are_valid`.
Please reply only after having done refactor on that method first.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1689,88 +1858,311 @@ impl<T: Trait> Module<T> {
...
+                    Some(entity_property_value) if *entity_property_value != *new_property_value
+                )
             })
+            .collect()
+    }
+
+    /// Update `entity_property_values` with `new_property_values`.
+    /// Returns `Some(BTreeMap<PropertyId, PropertyValue<T>>)`,
+    /// if `entity_property_values` have been updated, and `None` otherwise.
+    pub fn make_updated_values(
```
bedeho: Why does this exist, what is it doing?

bedeho (changes requested):
- The comments are a good start, they help - in particular the inline comments, however, your comments on methods are revealing the fact that often the semantics of chosen methods is not as crisp as it should be, example

```Rust
    /// Perform all necessary checks to ensure `new_values_for_existing_properties` are valid
    pub fn ensure_new_property_values_are_valid
```

This does not actually explain what the method does, it references a variable name used in some arbitrary scope as a core part of the explanation, that does not work. The explanation should state, in terms of the input values in the function, what it returns, and what other side effects it may have on storage or anything else. Another example 

```Rust
`Returns Some(BTreeMap<PropertyId, PropertyValue<T>>)`
```

It's redundant to explain the type, which is here wrong (`Option<..>`), but more importantly, it should explain the functional relationship between the output value and input values, which is not covered.

- Please throw out all existing reference counting logic as it currently stands. You have only superficially updated it in order to avoid the mutable variables and representation forcing same sign on both counters. At the heart, you are still computing the side effects as if each extrinsic will only ever increase or decrease both, you just have a representation which does not force it. You are also still overloading the same side effect apparatus for all extrinsics, which resulting in a serious lack of semantic clarity, and thus confidence in the correctness. Instead do as the last review suggested, _exactly_. So please kill all of this, and whatever else I am missing
   - `perform_entities_inbound_rcs_delta_calculation`
   - `calculate_entities_inbound_rcs_delta`
   - `get_updated_inbound_rcs_delta`
   - `DeltaMode`
   - `ReferenceCounterSideEffects::update`
   - `ReferenceCounterSideEffects::update_entities_rcs` (having a type method that results in global state side-effects reduces clarity, type methods make sense when they only mutate local `self` state at most)
 
-  Method `update_entity_property_values` is doing checks after the mutation safe marker, so either it has as serious bug, or the marker is in the wrong place

# Details

- `create_entity` does not need to mutate `entity_creation_voucher`, just make it with the right count in one step.

-  `ClassSchemaAdded` should be submitted outside storage mutation, same with any other event. Storage mutation closures should really do the minimal possible beyond updating the variable. Ideally everything is precomputed before the call.

- Move `// == MUTATION SAFE ==` up in `create_class`, it should be right after check `ensure_class_maintainers_are_valid`.

- Move `// == MUTATION SAFE ==` up in `update_class_permissions`, it should be right after check `ensure_class_maintainers_are_valid`.

- Move `// == MUTATION SAFE ==` up in `transfer_entity_ownership`, it should be right after check `ensure_property_values_are_valid`.

# Comment

Again I have not really reviewed all method in detail for correctness, because there are so many higher level issues. Only submit fixes to the issue pointed out here, all new stuff must go in another Pr.

runtime-modules/content-directory/src/lib.rs
iorveth: Thanks for pointing this issue. Made a reusable version of `ensure_property_values_are_valid` here https://github.com/Joystream/joystream/pull/669/commits/affc8b4b9b482df85445f3c3ba924fd586f42dee

runtime-modules/content-directory/src/lib.rs
iorveth: Good point, already decoupled `required` check as well here https://github.com/Joystream/joystream/commit/affc8b4b9b482df85445f3c3ba924fd586f42dee

runtime-modules/content-directory/src/lib.rs
iorveth: That's not a case, because `ValuesForExistingProperties` are not computed from `Schema` `property_ids`. They are computed from `PropertyValue`'s  and their corresponding `Class` `Property`'s  https://github.com/iorveth/substrate-runtime-joystream/blob/update_entity_permissions_extrinsic/runtime-modules/content-directory/src/helpers.rs#L51, which ids can easily be totally different from `Schema` `property ids` (non required properties)

runtime-modules/content-directory/src/lib.rs
iorveth: Good point, already decoupled required check as well here affc8b4

runtime-modules/content-directory/src/helpers.rs
iorveth: Because i wanted to encapsulate its https://github.com/iorveth/substrate-runtime-joystream/blob/update_entity_permissions_extrinsic/runtime-modules/content-directory/src/helpers.rs#L51 associated method. It's not possible to implement methods for type aliases. We also can create a specific trait for `ValuesForExistingProperties`, though i did not see any advantage. 
There's no runtime disadvantage to this - both versions will take the exact same amount of space.

iorveth:
`Move // == MUTATION SAFE == up in transfer_entity_ownership, it should be right after check ensure_property_values_are_valid.`

It shouldn`t, as we have following ensure prior the mutation https://github.com/iorveth/substrate-runtime-joystream/blob/update_entity_permissions_extrinsic/runtime-modules/content-directory/src/lib.rs#L1102

iorveth:
```Method update_entity_property_values is doing checks after the mutation safe marker, so either it has as serious bug, or the marker is in the wrong place```

Seems i missed to move it after refactoring performed. 

`Move // == MUTATION SAFE == up in create_class, it should be right after check ensure_class_maintainers_are_valid.`

`Move // == MUTATION SAFE == up in update_class_permissions, it should be right after check ensure_class_maintainers_are_valid.`

Done here https://github.com/Joystream/joystream/pull/669/commits/7549389cf4aca2d3cda6b5c7b6768356a80a912f

iorveth:
`ClassSchemaAdded should be submitted outside storage mutation, same with any other event. Storage mutation closures should really do the minimal possible beyond updating the variable. Ideally everything is precomputed before the call.`

Fixed this kind of issues here
https://github.com/Joystream/joystream/pull/669/commits/bfa804fdb764db31299fc3d26ade229402d8dd86 
https://github.com/Joystream/joystream/pull/669/commits/8245aa88f9c68deb2a8661601fae1bd66596ef1a

runtime-modules/content-directory/src/lib.rs
iorveth: It used to update existing `entity_property_values` with `new_property_values`.
Made an attemt to fix its documentation here https://github.com/Joystream/joystream/pull/669/commits/e300b5b05b783ee112fc394e2bfdb940938c6156#diff-0a165b042aa89a04d896ad36b8a33af5R2150

iorveth:
> The comments are a good start, they help - in particular the inline comments, however, your comments on methods are revealing the fact that often the semantics of chosen methods is not as crisp as it should be, example
    ```/// Perform all necessary checks to ensure `new_values_for_existing_properties` are valid```
    `pub fn ensure_new_property_values_are_valid`

> This does not actually explain what the method does, it references a variable name used in some arbitrary scope as a core part of the explanation, that does not work. The explanation should state, in terms of the input values in the function, what it returns, and what other side effects it may have on storage or anything else. 
Another example
`Returns Some(BTreeMap<PropertyId, PropertyValue<T>>)`
It's redundant to explain the type, which is here wrong (Option<..>), but more importantly, it should explain the functional relationship between the output value and input values, which is not covered.

Made an attempt to improve documentation due to requirements here https://github.com/Joystream/joystream/pull/669/commits/e300b5b05b783ee112fc394e2bfdb940938c6156

iorveth:
> Please throw out all existing reference counting logic as it currently stands. You have only superficially updated it in order to avoid the mutable variables and representation forcing same sign on both counters. At the heart, you are still computing the side effects as if each extrinsic will only ever increase or decrease both, you just have a representation which does not force it. You are also still overloading the same side effect apparatus for all extrinsics, which resulting in a serious lack of semantic clarity, and thus confidence in the correctness. Instead do as the last review suggested, exactly. So please kill all of this, and whatever else I am missing
perform_entities_inbound_rcs_delta_calculation
calculate_entities_inbound_rcs_delta
get_updated_inbound_rcs_delta
DeltaMode
ReferenceCounterSideEffects::update
ReferenceCounterSideEffects::update_entities_rcs (having a type method that results in global state side-effects reduces clarity, type methods make sense when they only mutate local self state at most)

Made a particular refactoring of `get_updated_inbound_rcs_delta`, where we have those 3 steps to compute side-effects and added more descriptive comments. Though it `s implementation should be probably revised later, as discussed.

Hope it`ll help a little. 

https://github.com/Joystream/joystream/pull/669/commits/e300b5b05b783ee112fc394e2bfdb940938c6156#diff-0a165b042aa89a04d896ad36b8a33af5R1830
https://github.com/Joystream/joystream/pull/669/commits/96b8d42ce9f9b05bd5b21f97d1f1790453de5a57

runtime-modules/content-directory/src/lib.rs
bedeho: As discussed internally, while `Schema::ensure_has_properties` is not the right check itself, `ValuesForExistingProperties` computation can be made safe, and this is done by lifting it up earlier and returning `Err` on bad input.

PR #672, opened 2020-06-11
Storage working group runtime changes. Iteration 2.

### LINKED ISSUE #570: Remove obsolete roles::actors module
Remove obsolete `roles::actors` module because it loses any meaning after the integration with the `bureaucracy` module.

[Root issue is here](https://github.com/Joystream/joystream/issues/544).

### LINKED ISSUE #571: Change the bureaucracy module
### Change the **bureaucracy** module to support the **storage** module integration.

1. Add mint initialization in the genesis.
2. Add method to enumerate workers - it is required to generate a random storage provider.
3. Remove the 'Set storage role parameters' proposal.
4. Remove the 'Evict storage provider' proposal.

[Root issue is here](https://github.com/Joystream/joystream/issues/544).

### LINKED ISSUE #573: Change the runtime crate
### Change the **runtime** crate to support the storage working group integration.

1. Remove or replace `LookupRoles` to query the working group.
2. Data migration:
    - **bureaucracy** mint migration.
    - delete the old data in the data directory module (DataObject).
    - delete the old data in the data object storage registry module 
(DataObjectStorageRelationship).
    - InputContraints values.
3. Add tests.

[Root issue is here](https://github.com/Joystream/joystream/issues/544).

### LINKED ISSUE #630: Restore the old forum code and remove any bureaucracy dependencies.
We should rollback the old forum code and erase all traces of the bureaucracy in the forum crate. We won't use this functionality in the nearest releases.

Subtask:
- Delete `SlashingTerms` artifact.

shamil-gadelshin:
## Changes
- rename "bureaucracy" to the "working group"
- rollback the 'forum-to-working group' integration
- init mint for the working group
- create runtime migrations
    - mint initialization for the storage working group
    - text constraints: `OpeningHumanReadableText`, `WorkerApplicationHumanReadableText`, `WorkerExitRationaleText`
    - remove old data: `data_directory`, `data_object_storage_registry`


node/src/chain_spec.rs
```diff
@@ -263,10 +267,10 @@ pub fn testnet_genesis(
         }),
         data_object_storage_registry: Some(DataObjectStorageRegistryConfig {
             first_relationship_id: 1,
-        }),
-        actors: Some(ActorsConfig {
-            enable_storage_role: true,
-            request_life_time: 300,
+            storage_working_group_mint_capacity: 0,
```
mnaamani, line 273: I don't think this is the best place for setting these configuration values.

The way it is done in substrate for the collective module which is instantiated multiple times in the runtime. You can give each instance its own configuration: 
See here https://github.com/paritytech/substrate/blob/1f536e98c509566245056369b752e9a4871de85c/bin/node/cli/src/chain_spec.rs#L284

And the collective module the logic in it's own add_extra_genesis block:
https://github.com/paritytech/substrate/blob/1f536e98c509566245056369b752e9a4871de85c/frame/collective/src/lib.rs#L146

runtime-modules/forum/src/lib.rs
```diff
@@ -9,20 +9,19 @@
...
 use srml_support::{decl_event, decl_module, decl_storage, dispatch, ensure, Parameter};
-use system::{ensure_signed, RawOrigin};
-
-pub use common::constraints::InputValidationLengthConstraint;
-use common::BlockAndTime;
 
 mod mock;
 mod tests;
 
+pub use common::constraints::InputValidationLengthConstraint;
```
mnaamani: No need to re export, it is used in `node/src/forum_config/mod.rs` . So It would need be changed there to import from `common` instead.

runtime-modules/forum/src/lib.rs
```diff
@@ -379,14 +382,39 @@ decl_module! {
...
         fn deposit_event() = default;
 
+        /// Set forum sudo.
+        fn set_forum_sudo(origin, new_forum_sudo: Option<T::AccountId>) -> dispatch::Result {
+            ensure_root(origin)?;
+
+            /*
+             * Question: when this routine is called by non sudo or with bad signature, what error is raised?
+             * Update ERror set in spec
+             */
```
mnaamani, line 392:
```
2020-06-16 16:40:30.099 tokio-runtime-worker-6 DEBUG runtime  DispatchError
2020-06-16 16:40:30.099 tokio-runtime-worker-6 DEBUG runtime  20
2020-06-16 16:40:30.099 tokio-runtime-worker-6 DEBUG runtime  0
2020-06-16 16:40:30.099 tokio-runtime-worker-6 DEBUG runtime  RequireRootOrigin
```

runtime-modules/forum/src/lib.rs
```diff
@@ -379,14 +382,39 @@ decl_module! {
...
+
+            // Hold on to old value
```
mnaamani, line 394: This could be simplified a bit by `take`ing the value so you don't need to kill it later. 
```rust
// Take old value
let old_forum_sudo = <ForumSudo<T>>::take();

// Update forum sudo
if let Some(account_id) = new_forum_sudo.clone() {
    <ForumSudo<T>>::put(account_id);
}
```

runtime-modules/service-discovery/src/mock.rs
```diff
@@ -11,10 +11,10 @@ pub use sr_primitives::{
 
 use srml_support::{impl_outer_event, impl_outer_origin, parameter_types};
 
-mod bureaucracy_mod {
-    pub use bureaucracy::Event;
-    pub use bureaucracy::Instance2;
-    pub use bureaucracy::Trait;
+mod working_group_mod {
+    pub use working_group::Event;
+    pub use working_group::Instance2;
```
mnaamani: might make sense to alias `Instance2` to `StorageWorkingGroup` with `as` to make it easier to understand which instance we are working with? In other places as well if it makes sense, this would be important in future as we add more instances. Also why did we skip Instance1 ?

runtime-modules/storage/Cargo.toml
```diff
@@ -73,15 +74,20 @@ default_features = false
 package = 'substrate-membership-module'
 path = '../membership'
 
+[dependencies.minting]
```
mnaamani: when removing the add extra genesis block out of the `data_object_storage_registry` module we can remove some unnecessary dependencies from here? Or do we still need all the dependencies of `substrate-working-group-module` as well, ie. minting, stake etc..

runtime-modules/storage/src/data_object_storage_registry.rs
```diff
@@ -121,6 +125,34 @@ decl_storage! {
         /// Keeps a list of storage relationships per content id.
         pub RelationshipsByContentId get(relationships_by_content_id): map T::ContentId => Vec<T::DataObjectStorageRelationshipId>;
     }
+    add_extra_genesis {
```
mnaamani: This can be moved out into the working-group module itself, and the config settings can be set on each instance in the runtime configuration. see my previous comment.

runtime-modules/storage/src/tests/data_object_storage_registry.rs
```diff
@@ -155,3 +157,33 @@ fn test_toggle_ready() {
         );
     });
 }
+
+#[test]
```
mnaamani: I think these tests will also move into the working-group module

runtime/src/lib.rs
```diff
@@ -755,26 +756,11 @@ impl migration::Trait for Runtime {
     type Event = Event;
 }
 
-// Forum bureaucracy
-impl bureaucracy::Trait<bureaucracy::Instance1> for Runtime {
+// Storage working group
+impl working_group::Trait<working_group::Instance2> for Runtime {
```
mnaamani: Why the change to Instance2 ? Are you reserving Instance1 for the content working group? Or was instance1 for the forum.. I kind of got confused here. Maybe we can import working_group::Instance2 and working_group::Instance1 with an alias at the top to avoid this confusion and possibly making a mistake references the wrong instance.

runtime/src/lib.rs
```diff
@@ -867,7 +853,6 @@ construct_runtime!(
         Memo: memo::{Module, Call, Storage, Event<T>},
         Members: members::{Module, Call, Storage, Event<T>, Config<T>},
         Forum: forum::{Module, Call, Storage, Event<T>, Config<T>},
-        Actors: actors::{Module, Call, Storage, Event<T>, Config},
```
mnaamani, line 870: Good bye 👍

runtime/src/lib.rs
```diff
@@ -884,9 +869,9 @@ construct_runtime!(
         ProposalsEngine: proposals_engine::{Module, Call, Storage, Event<T>},
         ProposalsDiscussion: proposals_discussion::{Module, Call, Storage, Event<T>},
         ProposalsCodex: proposals_codex::{Module, Call, Storage, Error, Config<T>},
-        // --- Bureaucracy
-        ForumBureaucracy: bureaucracy::<Instance1>::{Module, Call, Storage, Event<T>},
-        StorageBureaucracy: bureaucracy::<Instance2>::{Module, Call, Storage, Event<T>},
+        // --- Working groups
+        // reserved for the future use: ForumWorkingGroup: working_group::<Instance1>::{Module, Call, Storage, Event<T>},
```
mnaamani, line 874: oh hello :) that explains things.

runtime-modules/storage/src/data_object_storage_registry.rs
```diff
@@ -121,6 +125,34 @@ decl_storage! {
...
+        config(worker_exit_rationale_text_constraint): InputValidationLengthConstraint;
+        build(|config: &GenesisConfig<T>| {
+            // Create a mint.
+            let mint_id_result =
+                <minting::Module<T>>::add_mint(config.storage_working_group_mint_capacity, None);
+
+            if let Ok(mint_id) = mint_id_result {
+                <working_group::Mint::<T, working_group::Instance2>>::put(mint_id);
+            } else {
+                print("Failed to create a mint for the storage working group");
```
mnaamani: I would add a panic! here there is no point initializing a chain spec with bad state. And this isn't runtime code so its okay to fail loudly.

runtime/src/migration.rs
```diff
@@ -18,10 +21,24 @@ impl<T: Trait> Module<T> {
         // would be any new storage values that need an initial value which would not
         // have been initialized with config() or build() chainspec construction mechanism.
         // Other tasks like resetting values, migrating values etc.
+
+        Self::initialize_storage_working_group_mint();
+        Self::initialize_storage_working_group_text_constraints();
+
+        // ** Order is important!
+        Self::clear_storage_data_object_registry_data();
+        Self::clear_storage_data_directory_data();
```
mnaamani: maybe both steps in a single `fn clear_storage_data()` and avoid need for the "order is important" comment ?

How about also resetting the `next_*_id` state variables to their initial values as well ?

runtime-modules/storage/src/data_object_storage_registry.rs
```diff
@@ -25,10 +25,14 @@
 use codec::{Codec, Decode, Encode};
 use rstd::prelude::*;
 use sr_primitives::traits::{MaybeSerialize, Member, SimpleArithmetic};
-use srml_support::{decl_error, decl_event, decl_module, decl_storage, ensure, Parameter};
+use srml_support::{
+    decl_error, decl_event, decl_module, decl_storage, ensure, print, Parameter, StorageValue,
```
mnaamani: adding `StorageValue` seems un-necessary. Did clippy complain?

mnaamani (changes requested):
Fantastic work.
left a few suggestions.

runtime/src/migration.rs
shamil-gadelshin: Both fixed.

runtime-modules/storage/src/data_object_storage_registry.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/data_object_storage_registry.rs
shamil-gadelshin: Rust thing: a trait declaration should be in scope.

runtime/src/lib.rs
shamil-gadelshin: Introduced the alias for instances. Instance1 is reserved for the Forum.

runtime-modules/service-discovery/src/mock.rs
shamil-gadelshin: Introduced the alias for instances. Instance1 is reserved for the Forum.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: Fixed.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: I didn't change the forum code during this PR - only restored the old version.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: I didn't change the forum code during this PR - only restored the old version. Probably that was old Bedeho's question.

mnaamani: ok

runtime-modules/storage/src/data_object_storage_registry.rs
mnaamani: correct, but if I'm not mistaken the `decl_storage!` macro now brings the necessary ones into scope.. like StorageMap, StorageValue etc.. anyway its not a big deal if it stays.

shamil-gadelshin: Hmm. You are right. It compiles without it. There were errors previously.

runtime-modules/storage/Cargo.toml
shamil-gadelshin: Moved mint module to the dev-dependencies. All of these modules mint, stake ... etc are required for mocking.

runtime-modules/storage/src/data_object_storage_registry.rs
shamil-gadelshin: Done.

runtime-modules/storage/src/tests/data_object_storage_registry.rs
shamil-gadelshin: Agree.

node/src/chain_spec.rs
shamil-gadelshin: Actually, you're right.  I moved the code to the working group and it is much better now. Thanks for the good review.

runtime-modules/storage/Cargo.toml
mnaamani: oh right of course :+1:

runtime-modules/working-group/src/lib.rs
```diff
@@ -1247,6 +1261,30 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
             )
             .map_err(|e| e.into())
     }
+
+    fn initialize_working_group(
+        opening_human_readable_text_constraint: InputValidationLengthConstraint,
+        worker_application_human_readable_text_constraint: InputValidationLengthConstraint,
+        worker_exit_rationale_text_constraint: InputValidationLengthConstraint,
+        storage_working_group_mint_capacity: minting::BalanceOf<T>,
```
mnaamani:
```suggestion
        working_group_mint_capacity: minting::BalanceOf<T>,
```

runtime-modules/working-group/src/lib.rs
```diff
@@ -1247,6 +1261,30 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+    ) {
+        // Create a mint.
+        let mint_id_result =
+            <minting::Module<T>>::add_mint(storage_working_group_mint_capacity, None);
```
mnaamani:
```suggestion
            <minting::Module<T>>::add_mint(working_group_mint_capacity, None);
```

runtime-modules/working-group/src/lib.rs
```diff
@@ -1247,6 +1261,30 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+
+        if let Ok(mint_id) = mint_id_result {
+            <Mint<T, I>>::put(mint_id);
+        } else {
+            panic!("Failed to create a mint for the storage working group");
```
mnaamani:
```suggestion
            panic!("Failed to create a mint for the working group");
```

mnaamani: I'm not sure if we can identify which instance also in the error message?

runtime-modules/working-group/src/tests/mock.rs
```diff
@@ -124,17 +124,42 @@ impl recurringrewards::Trait for Test {
...
     type Event = TestEvent;
 }
 
 pub type Membership = membership::members::Module<Test>;
-pub type WorkingGroup1 = Module<Test, Instance1>;
+
+pub type TestWorkingGroupInstance = crate::Instance1;
+pub type TestWorkingGroup = Module<Test, TestWorkingGroupInstance>;
+
+pub(crate) const STORAGE_WORKING_GROUP_MINT_CAPACITY: u64 = 40000;
```
mnaamani: `STORAGE_` prefix for these constants here could be removed.. I'm just being picky :)

runtime-modules/working-group/src/tests/mod.rs
```diff
@@ -1592,8 +1617,37 @@ fn set_working_group_mint_capacity_fails_with_mint_not_found() {
...
         let capacity = 15000;
-        let result = WorkingGroup1::set_mint_capacity(RawOrigin::None.into(), capacity);
+        let result = TestWorkingGroup::set_mint_capacity(RawOrigin::None.into(), capacity);
 
         assert_eq!(result, Err(Error::RequireRootOrigin));
     });
 }
+
+#[test]
+fn ensure_setting_genesis_storage_working_group_mint_succeeds() {
```
mnaamani: `_storage` in the names of the test for the working group can be dropped.

runtime/src/migration.rs
```diff
@@ -117,9 +115,10 @@ impl<T: Trait> Module<T> {
 
             potential_id += <T as storage::data_object_storage_registry::Trait>::DataObjectStorageRelationshipId::one();
         }
-    }
 
-    fn clear_storage_data_directory_data() {
+        storage::data_object_storage_registry::NextRelationshipId::<T>::put(<T as storage::data_object_storage_registry::Trait>::DataObjectStorageRelationshipId::one());
```
mnaamani: `one()` works because the constant `DEFAULT_FIRST_RELATIONSHIP_ID` defined in the module is equal to 1.
But maybe better to just`kill()` the value, next time it is read it will use the default storage value in this case configured as the constant:

```rust
pub FirstRelationshipId get(first_relationship_id) config(first_relationship_id):
            T::DataObjectStorageRelationshipId = T::DataObjectStorageRelationshipId::from(DEFAULT_FIRST_RELATIONSHIP_ID);
```

mnaamani (changes requested):

runtime-modules/proposals/codex/src/proposal_types/mod.rs
```diff
@@ -47,14 +46,8 @@ pub enum ProposalDetails<MintedBalance, CurrencyBalance, BlockNumber, AccountId,
     /// Balance for the `set content working group mint capacity` proposal
     SetContentWorkingGroupMintCapacity(MintedBalance),
 
-    /// AccountId for the `evict storage provider` proposal
-    EvictStorageProvider(AccountId),
```
mnaamani, line 51: I just realized we actually need to keep a "dummy" value here to preserve the enumeration index correct when reading state of proposals of this type. 

can we change it to:
`_EvicStorageProvider(AccountId)`

runtime-modules/proposals/codex/src/proposal_types/mod.rs
```diff
@@ -47,14 +46,8 @@ pub enum ProposalDetails<MintedBalance, CurrencyBalance, BlockNumber, AccountId,
...
-
     /// Validator count for the `set validator count` proposal
     SetValidatorCount(u32),
-
-    /// Role parameters for the `set storage role parameters` proposal
-    SetStorageRoleParameters(RoleParameters<CurrencyBalance, BlockNumber>),
```
mnaamani, line 57: `_SetStorageRoleParameters(RoleParameters<CurrencyBalance, BlockNumber>)`
This means we also need to preserve the `struct RoleParameters` type.

I think we can remove the encapsulated value if we can ensure we don't try to read in the runtime any proposal detail of this variant.

runtime-modules/working-group/src/tests/mock.rs
shamil-gadelshin: Fixed.

runtime-modules/working-group/src/tests/mod.rs
shamil-gadelshin: Fixed.

runtime-modules/proposals/codex/src/proposal_types/mod.rs
shamil-gadelshin: Fixed.

runtime-modules/proposals/codex/src/proposal_types/mod.rs
shamil-gadelshin: Fixed.

runtime/src/migration.rs
shamil-gadelshin: Fixed.

runtime-modules/working-group/src/lib.rs
shamil-gadelshin: Fixed.

mnaamani (approved):

Issue #558, opened 2020-06-05
Forum improvements


bedeho:
# Background

One of the biggest pains with the current forum is that it's really hard to keep of track whether there is any new activity in the forum to check in on, you just get this top-level view This is equally the case when you go into a specific category. In addition, there are lots of smaller loose ends in the user experience.

https://testnet.joystream.org/#/forum

# Proposal

- A standard solution that is found in most other forums is to list, in each category (including also the root level), what the N most recent posts are, with a link to jump to it. It probably should be something slightly more sophisticated, like the last N posts in the last N threads which had activity, in order to avoid getting a bunch of posts all from the same thread. This could be shown ab
- The actual top-level display of each category makes little sense, we should
  - drop the "New thread" button
  - drop the "Creator column"
- Drop "New category" and "Forum sudo" buttons.
- Each category should also display its description.
- All posts and threads need a date and time, as requested by the council https://testnet.joystream.org/#/proposals/4.
- Manual line breaks inputted during the composition of a new post is not displayed properly when viewed later.
- Each post should have a post number next to it which is linkable.
- When clicking reply, the user is taken to a new page and cannot see posts on the same page. The reply section should be at the bottom of each page of the thread so that a user can reference current posts without needing two windows open.
- Uppercase title is put to all lowercase on thread view (it shows correctly on subforum view) (shown incorrectly here: https://testnet.joystream.org/#/forum/threads/)
- Posts don't support indented markdown lists
- The default font size of the post content is too small.
- Need the option to reply to or quote previous posts.

Overall, feel free to augment the visual style in any way which supports usability.

mochet:
For Markdown I created a post which shows how it currently renders using **https://markdown-it.github.io/**
https://testnet.joystream.org/#/forum/threads/37

It seems to render many things fine.

bedeho:
There is a way of changing the scope here slightly, which will make it radically simpler to address the underlying discovery problem to a significant extent, but with much less time and effort.

If we instead are happy to just show the last `N` posts globally across the entire forum on only the front-page, then we can do a single read operation that reads exactly the last `N` forum posts from the chain, as they will have identifiers `k, k-1, ... , k - (N-1)`, where `k` is the id of the most recent post.

mochet:
Two features that are missing that would be great to see.
The ability to edit a thread title. You can edit post contents, but not thread titles.
The ability to lock threads.

Didn't want to make a new issue for just these items, so if that needs doing let me know and I'll create one.

[...]

PR #678, opened 2020-06-12
Change action tabs to buttons

kdembler:
This PR tackles #458 and tries to improve the UX by getting rid of tabs that are strictly actions. Some tabs were offering an action but also serve other, informative purposes, so they weren't changed into buttons (for example, Membership -> Edit my profile, tab title was changed to "My profile").

I've skipped the Forum view, as it will be soon changed by #558 so it didn't seem reasonable to invest time into it.

List of affected tabs:

- Media -> New channel
I think the new channel button isn't a top-level action so it was removed from the navigation. "My channels" tab already offers the same button anyway. A header was added to the edit channel form so that it's clear what view it is after the tab was removed.

- My Keys -> Vanity address
The tab was changed to a button and the vanity address form got its headers as well:
Screen Shot 2020-06-12 at 11 59 48
Screen Shot 2020-06-12 at 11 59 56

- Address book -> View memo
Tabs were removed altogether in this view because they weren't required anymore. Instead a header with breadcrumbs was added so that user knows what view they're looking at and has a way to go back without using browser navigation:
Screen Shot 2020-06-12 at 12 02 08
Screen Shot 2020-06-12 at 12 02 20

- Council -> Reveal a vote
I find this action to not be top-level as well so it was removed from the tabs and instead placed in the "Votes" tab:
Screen Shot 2020-06-12 at 12 03 37

- Proposals -> New proposal
Same as with "View memo", tabs were removed and breadcrumbs were added:
Screenshot 2020-06-17 at 10 35 47 AM
Screen Shot 2020-06-12 at 12 08 22

Also, as I've mentioned in the issue, I omitted updating the "Fully featured" views, as I believe those are for the internal use and the effort is not worth it.

### DIFF
--- a/pioneer/packages/app-accounts/src/Overview.tsx
+++ b/pioneer/packages/app-accounts/src/Overview.tsx
@@ -7,6 +7,8 @@ import { SubjectInfo } from '@polkadot/ui-keyring/observable/types';
 import { ComponentProps } from './types';
 
 import React, { useState } from 'react';
+import { Link, useLocation } from 'react-router-dom';
+import { Button as SUIButton } from 'semantic-ui-react';
 import keyring from '@polkadot/ui-keyring';
 import accountObservable from '@polkadot/ui-keyring/observable/accounts';
 import { getLedger, isLedger, withMulti, withObservable } from '@polkadot/react-api';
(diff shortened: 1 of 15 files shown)


bedeho:
Fix merge conflict.

kdembler:
Fixed

bedeho:
Fix merge conflict again.

kdembler:
Fixed the merge conflict and updated the screenshot of proposals view

bedeho:
LGTM, well done.

Issue #663, opened 2020-06-10
Update Pioneer: Media App


mnaamani:
Following changes in https://github.com/Joystream/joystream/pull/567

Update:
- types - building on https://github.com/Joystream/joystream/pull/640
- DiscoveryProvider
  - to work with changes in storage runtime
  - improve caching code

mnaamani:
Partly Implemented in #687 will create separate issues for improving caching and updating dependent apps (colossus, status server, api-examples)

mnaamani:
Done in https://github.com/Joystream/joystream/pull/722

Issue #664, opened 2020-06-10
CI - Additional Checks for Nicaea


mnaamani:
- Add github actions to run tests for 
  - cli
  - colossus
- Add check when any file changes in `runtime/` folder, ensure runtime version is incremented and Cargo.toml version matches convention.
- Build checks: https://github.com/Joystream/joystream/pull/687#issuecomment-644097609
- ~after yarn install check that yarn.lock wasn't modified, this indicates it wasn't checked in correctly following dependency changes~ - this is already handled by using the `--frozen-lockfile` argument

mnaamani:
Partial work in https://github.com/Joystream/joystream/pull/752
Covers build checks for cli, types.

fcescob:
Adding this to the `before_install` scripts makes Travis skip the build if no Rust or Cargo files were changed:

```yaml
before_install:
  - |
    if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qE "(.rs|cargo.(lock|toml))$"
      then
      echo "No changes to Rust or Cargo Files, CI not running."
      travis_terminate 0
    fi

```

It is currently working on the `init_atlas` branch (as soon as #693 is merged). 
Full file here: https://github.com/Gamaranto/joystream/blob/atlas-first-screen/.travis.yml

mnaamani:
@Gamaranto wonderful thank you for adding this. I will try to add the same for nicaea branch, and perhaps also update our git hook scripts to include this logic.

I wonder does the git diff command behave the same for pushes after the PR is actually merged travis runs again for the target branch that was updated.

mnaamani:
update shared eslint config with suggestion from https://github.com/Joystream/joystream/pull/943#issuecomment-658652904

mnaamani:
checks for the types library being added by @Lezek123 in https://github.com/Joystream/joystream/pull/963

mnaamani:
completed in:
https://github.com/Joystream/joystream/pull/1023
https://github.com/Joystream/joystream/pull/1020
https://github.com/Joystream/joystream/pull/973
https://github.com/Joystream/joystream/pull/963

PR #687, opened 2020-06-12
Pioneer media app fixes and types updates

mnaamani:
Part of https://github.com/Joystream/joystream/issues/663 - excluding resolver cache improvements
- [x] Update DataObject and DataObjectStorageRelationship types
- [x] Change version to 0.11.0
- [x] Use relative path to @joystream/types package by dependents in workspace 
- [x] Update dependents in repo to use path instead of version
- [x] Drop SchemaId type
- [x] Drop ContentVisibility type
- [x] Removed old types definitions from [unused old module](https://github.com/Joystream/joystream/blob/8c2c1928b8f5cb88a58eeb9a59a37d2e339a98b7/src/storage/downloads.rs)
- [x] Refactor `types/index.ts` - move council and election types into separate file, common types into common.ts
- [x] Updated pioneer after refactor of types

Tried to identify as many cases of circular references as possible.


mnaamani:
I'm currently getting error in web console when running pioneer:

`TypeError: class heritage _members.ActorId is not an object or null`

Will investiage

mnaamani:
Still work in progress, after all the refactoring in types will need to update pioneer.

cli/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "bugs": "https://github.com/Joystream/substrate-runtime-joystream/issues",
   "dependencies": {
-    "@joystream/types": "^0.10.0",
+    "@joystream/types": "./types",
```
Lezek123, line 11: Wouldn't this cause a problem if we decide to publish this package? (Similarly to `"@joystream/types": "*"`?)

Lezek123 (commented):
The circular dependency problem is solved. I went through the changes and tested a few things:
- [X] Checked almost all CLI commands (including the new `working-groups` one I merged into this branch) - found no issues
- [X] Tested Pioneer's forum, proposals, council election process etc. (functionalities where file changes were included) - found no issues
- [X] Tested whether merging my new cli and Pioneer related changes doesn't cause any issues - it doesn't
- [X] Whether this fixes the issue I encountered in `joystream-api-examples` - it does (but an update in that repository is required)
- [X] Checked whether circular dependency detection tool doesn't detect any circular dependencies anymore - I used `madge` tool here and it didn't detect any circular dependencies. I then purposefully created a circular dependency and re-build the types and it was succesfully detected, so the tool seems to be working fine.
- [X] Ran network tests (had to run `yarn install` first though) - there were a few issues with transaction priority and timeouts, but didn't seem like anything related to the changes in this PR
- [ ] Didn't fully test the media app yet as it would probably require some more setup (and I'm not sure if that's even possible already?)

I was thinking that it would be nice to have some additional CI checks included (not necessarily in this PR), ie.:
- `yarn workspace @joystream/types build` (now that they are not part of Pioneer workspace anymore)
- A script like `yarn workspace @joystream/types build && madge --circular types` to avoid circular dependencies in the future
- `yarn workspace joystream-cli build` (just to make sure CLI code compiles) _Edit: that's actually only possible after https://github.com/Joystream/joystream/pull/692 gets merged_

cli/package.json
mnaamani: Yes it would indeed. I think we need to investigate what is the best practice inside a workspace. I did make you change a similar import once with '*' for the same reason :) 
We really need to look into `lerna` and if it perhaps helps to resolve the correct version at the point when we actually publish.

@Gamaranto is exploring using `lerna` now while working on atlas.

It is worth mentioning also that he used it to build packages in the correct order based on the dependency tree. Maybe it could also help to detect circular dependencies in general?

mnaamani:
> I was thinking that it would be nice to have some additional CI checks included (not necessarily in this PR), ie.:
> 
>     * `yarn workspace @joystream/types build` (now that they are not part of Pioneer workspace anymore)
> 
>     * A script like `yarn workspace @joystream/types build && madge --circular types` to avoid circular dependencies in the future
> 
>     * `yarn workspace joystream-cli build` (just to make sure CLI code compiles) _Edit: that's actually only possible after #692 gets merged_

Yes, I had an issue for nicea to add some new CI checks here https://github.com/Joystream/joystream/issues/664 will add your suggestions to it.

Lezek123 (approved):

Issue #577, opened 2020-06-05
Pioneer: List storage working group opportunities among workig-groups/opportunities


Lezek123:
Listing the new storage working group opportunities will require interacting with different api endpoints, ie.:

**Get the list of CWG opporunities:**
- `contentWorkingGroup.nextCuratorOpeningId`
- `contentWorkingGroup.curatorOpeningById`
- ...

**Get the list of Storage group opportunities:**
- `storageBureaucracy.nextWorkerOpeningId`
- `storageBureaucracy.workerOpeningById`
- ...

Besides that, since currently `contentWorkingGroup` will not yet be refactored to use the new `bureaucracy` module, we will have to interact with different types for both (ie. `WorkerOpening` vs `CuratorOpening` etc.)

**The UI:**
On the `workig-groups/opportunities` overview we should have the ability to filter the list by working group, which could be implemented in a similar manner to how proposal types can be filtered by category on the `proposals/new` page. There should also be a separate route for each group's opportunities (ie.: `/working-groups/opportunities/curators`, `/working-groups/opportunities/storage`).

Lezek123:
Closed with: https://github.com/Joystream/joystream/pull/691

PR #691, opened 2020-06-12
Pioneer: Listing Storage Working Group opportunities

Lezek123:
Implements https://github.com/Joystream/joystream/issues/577

**How to test?**
Because current version of the runtime will throw `OpeningTextTooLong` error when trying to create an opening with any value inside `human_readable_text`, some change in the runtime code is required in order to properly test this functionality (currently `storageBureaucracy.openingHumanReadableText`, which shows validation constraints for this field, returns `{"min":0,"max_min_diff":0}`). During my tests I just used a modified version of the runtime where I commented out everything inside `ensure_opening_human_readable_text_is_valid` and replaced it with `Ok(())`.

When this issue is dealt with, `storageBureaucracy.addWorkerOpening` extrinsic can be used to create example openings. In order for the opening to display on the `/working-groups/opportunities` page, a valid json inside `human_readable_text` is required. It can be copied from the new curator opening form at `working-groups/admin` (the raw json is under `Role staking policy` field) and converted into hex, ie. via https://www.rapidtables.com/convert/number/ascii-to-hex.html

**What to expect**
- Working group opportunities for both "Storage providers" and "Curators" should be listed on `/working-groups/opportunities`, opportunities should have labels in the top right corner based on the group they regard.
- Opportunities for Storage providers should be also acessible from `/working-groups/opportunities/storageProviders` and for Curators from `working-groups/opportunities/curators`.
- A select field in the top right corner on `/working-groups/opportunities` should allow changing the group of intrest, which should also affect the current url.
- Everything related to Curator openings and roles (`working-groups/admin` tool, applying, `working-groups/my-roles` page) should remain fully functionable.
- Applying for Storage provider openings should not yet be possible (a message about this not yet beeing implemented should be displayed) and no functionality other than listing opportunities should be expected for Storage Working Group.

### DIFF
--- a/pioneer/packages/joy-roles/src/tabs/Opportunities.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/Opportunities.tsx
@@ -4,7 +4,7 @@ import NumberFormat from 'react-number-format';
 import marked from 'marked';
 import CopyToClipboard from 'react-copy-to-clipboard';
 
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
 import {
   Button,
   Card,
@@ -14,7 +14,9 @@ import {
   Label,
   List,
   Message,
-  Statistic
+  Statistic,
+  Dropdown,
+  DropdownProps
 } from 'semantic-ui-react';
 
 import { formatBalance } from '@polkadot/util';
@@ -37,6 +39,9 @@ import {
 } from '../openingStateMarkup';
 
 import { Loadable } from '@polkadot/joy-utils/index';
+import styled from 'styled-components';
+import _ from 'lodash';
+import { WorkingGroups, AvailableGroups } from '../working_groups';
 
 type OpeningStage = OpeningMetadataProps & {
   stage: OpeningStageClassification;
@@ -458,6 +463,14 @@ export type WorkingGroupOpening = OpeningStage & DefactoMinimumStake & OpeningMe
   applications: OpeningStakeAndApplicationStatus;
 }
 
+const OpeningTitle = styled.h2`
+  display: flex;
+  align-items: flex-end;
+`;
+const OpeningLabel = styled(Label)`
+  margin-left: auto !important;
+`;
+
 type OpeningViewProps = WorkingGroupOpening & BlockTimeProps & MemberIdProps
 
 export const OpeningView = Loadable<OpeningViewProps>(
@@ -473,7 +486,10 @@ export const OpeningView = Loadable<OpeningViewProps>(
 
     return (
       <Container className={'opening ' + openingClass(props.stage.state)}>
-        <h2>{text.job.title}</h2>
+        <OpeningTitle>
+          {text.job.title}
+          <OpeningLabel>{ _.startCase(props.meta.group) }</OpeningLabel>
+        </OpeningTitle>
         <Card fluid className="container">
           <Card.Content className="header">
             <OpeningHeader stage={props.stage} meta={props.meta} />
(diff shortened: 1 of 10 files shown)


pioneer/packages/joy-roles/src/transport.substrate.ts
```diff
@@ -288,17 +339,17 @@ export class Transport extends TransportBase implements ITransport {
...
-      const cApplication = new SingleLinkedMapEntry<CuratorApplication>(
-        CuratorApplication,
-        await this.cachedApi.query.contentWorkingGroup.curatorApplicationById(i)
+      const cApplication = new SingleLinkedMapEntry<GroupApplication>(
+        group === WorkingGroups.ContentCurators ? CuratorApplication : WorkerApplication,
+        await this.cachedApiMethodByGroup(group, 'applicationById')(i)
       );
 
-      if (cApplication.value.curator_opening_id.toNumber() !== curatorOpeningId) {
+      if (cApplication.value.worker_opening_id.toNumber() !== groupOpeningId) {
```
mnaamani, line 352: Do we not have to do a similar conditional check on the type? wether it is a curator or worker application type to determine if we should access `curator_opening_id` or `worker_opening_id` property of `cApplication.value` ?

mnaamani: Oh I missed this helper function which works around this..

```
// Helper for bureaucracy compatibility
  get worker_opening_id(): CuratorOpeningId {
    return this.curator_opening_id;
  }
```

pioneer/packages/joy-roles/src/tabs/Opportunities.tsx
```diff
@@ -497,17 +513,47 @@ export const OpeningView = Loadable<OpeningViewProps>(
...
-        {props.openings && props.openings.map((opening, key) => (
+        <FilterOpportunities>
+          <FilterOpportunitiesDropdown
+            placeholder="All opportunities"
+            options={
+              [{ value: '', text: 'All opportunities' }]
+                .concat(AvailableGroups.map(g => ({ value: g, text: _.startCase(g) })))
+            }
+            defaultValue={''}
+            value={group}
```
mnaamani, line 550: Both `defaultValue` and `value` props are defined here, when loading the opportunities page I noticed the following error in the console:

Screen Shot 2020-06-15 at 5 28 39 PM

mnaamani:
> Because current version of the runtime will throw `OpeningTextTooLong` error when trying to create an opening with any value inside `human_readable_text`, some change in the runtime code is required in order to properly test this functionality (currently `storageBureaucracy.openingHumanReadableText`, which shows validation constraints for this field, returns `{"min":0,"max_min_diff":0}`). During my tests I just used a modified version of the runtime where I commented out everything inside `ensure_opening_human_readable_text_is_valid` and replaced it with `Ok(())`.

Ah yes these were not configurable at genesis so all the constraint values had default of zero.

Just leaving a note here for @shamil-gadelshin to include initializing good default values for the input constraints in the runtime migrations steps, as well as maybe making the values configurable at genesis.

mnaamani:
If the width of the browser is very narrow, the dropdown to select the group seems to disappear.

mnaamani:
Tested with creating opportunities as described. Worked fine. So there is only the issue with the dropdown error in console, although it doesn't seem to affect the expected behavior in firefox.

Lezek123:
Fixed the console error issue.

> If the width of the browser is very narrow, the dropdown to select the group seems to disappear.

Couldn't reproduce this one. There is some reloading happening when the browser window is resized (which probably shouldn't happen) and it may cause the dropdown to dissapear for a moment (the loading animation should be displayed instead, but it's easy to miss). This isn't something that was introduced in this PR and perhaps I'll get to some of those minor UI issues once I'm done with all the core Nicaea tasks.

mnaamani (approved):

Issue #598, opened 2020-06-08
CLI: Create working-groups command family, add group overview and "lead/worker only" gates


Lezek123:
In order to prepare the CLI for the introduction of lead/worker actions like managing openings, updating role account etc., it may be a good start to implement some core, reusable functionality first.

I think a good start could include:
- Creating a new commands family (ie. working-groups)
- Adding a group overview that'll display current lead, list of workers etc. It will be useful because the commands that we're going to add later will need to be aware of this information anyway.
- Adding a reusable "gate" that will determine whether currently selected account is associated with the lead position and prevent entering a command if it's not (displaying a clear error/warning message)
- Adding a similar gate for the worker, except that in this case we will also need something like `promptForWorker` (similar to `AccountCommandBase.promptForAccount`) in case there is more than 1 worker associated with the currently selected account.
- Because of the above it may be a good idea to add a separate command to list all the workers/roles associated with the currently selected account, or perhaps - membership (some slightly related work has already been started in https://github.com/Joystream/joystream/pull/292, so it may be a good idea to finish and merge it first)

Lezek123:
Closed with: https://github.com/Joystream/joystream/pull/692

PR #692, opened 2020-06-12
CLI: Init working-groups commands family

Lezek123:
Implements https://github.com/Joystream/joystream/issues/598

Core functionality implemented:
- `./cli/bin/run working-groups` commands family
- `./cli/bin/run working-groups:overview --group=GROUP` command that displays working group leader and members/workers information (similar to working group overview in Pioneer)
- added reusable `getRequiredLead` and `getRequiredWorker` methods inside `WorkingGroupsCommandBase` class, which serve as both a gate that will prevent users without access (worker/lead) from trying to execute a command that requires it AND a way of retrieving lead/worker entity and member profile information associated with the selected CLI account (since there could be more than 1 worker associated with a given key, there is also a support for choosing one of the available roles in that case)

I also introduced a few general improvements:
- Default currency symbol and decimals for `formatBalance` are now set based on the values provided by the runtime when initializing the api (same way it's done within Pioneer)
- Added dev `Alice` and `Bob` accounts as "special" hidden accounts that can be selected using `./cli/bin/run accounts:choose --showSpecial` (for easier testing)
- Added `build` script for `joystream-cli` workspace and fixed the `tsconfig` issue (jest vs mocha types conflict)

### DIFF
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -160,12 +160,8 @@ export default class Api {
         return this._api.query[module];
     }
 
-    protected async memberProfileById(memberId: MemberId, throwError = true): Promise<Profile | null> {
+    protected async memberProfileById(memberId: MemberId): Promise<Profile | null> {
         const profile = await this._api.query.members.memberProfile(memberId) as Option<Profile>;
-
-        if (throwError && profile.isNone) {
-          throw new Error(`Profile not found for member id: ${memberId.toNumber()}!`);
-        }
 
         return profile.unwrapOr(null);
     }
@@ -178,7 +174,11 @@ export default class Api {
         }
 
         const lead = optLead.unwrap();
-        const profile = (await this.memberProfileById(lead.member_id))!;
+        const profile = await this.memberProfileById(lead.member_id);
+
+        if (!profile) {
+            throw new Error(`Group lead profile not found! (member id: ${lead.member_id.toNumber()})`);
+        }
 
         return { lead, profile };
     }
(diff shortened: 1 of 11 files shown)


cli/src/Api.ts
```diff
@@ -111,4 +142,110 @@ export default class Api {
...
+
+    protected workingGroupApiQuery(group: WorkingGroups) {
+        const module = apiModuleByGroup[group];
+        return this._api.query[module];
+    }
+
+    protected async memberProfileById(memberId: MemberId, throwError = true): Promise<Profile | null> {
+        const profile = await this._api.query.members.memberProfile(memberId) as Option<Profile>;
+
+        if (throwError && profile.isNone) {
```
mnaamani: Why is the function throwing if a member id doesn't exist. Shouldn't the caller just test if the return value is null?
In general I think it makes more sense for the function to throw if we add a check for the existence of a key in a map (by reading directly from storage with api.rpc.getStorage)

mnaamani (approved):
Looks good, I tested on linux, still doesn't work on my mac. I only left a comment but it should be good to merge.

cli/src/Api.ts
Lezek123: I think you're right, I copied some code from Pioneer's `joy-roles` transport here and probably got too influenced by the fact that a similar method was throwing an error there.

It seems to be not too good design though, because then we have to use TypeScript assertions to "convince" the compiler that the value is never `null` when `throwError === true`.

Fixed in a new commit.

PR #693, opened 2020-06-12
Atlas first screen

fcescob:
This PR aims to implement Atlas First Screen as described in Figma. 
The development experience in the monorepo has also greatly been improved (along with the build process) and all the changes to configuration files are a reflection of that.

## What To Expect
The design for the desktop view should be functional and resemble the one on Figma. 

## What To Not Expect

Unfortunately, this screen is missing features, mainly animations, which are what gives life to the page. There is a basic carousel animation, but I've been unsuccessful to reproduce any working animation so far. 

To be precise, the core ones that I'm missing:

- [ ] Swipe Animation for the mobile view. Without this one a "mobile view" is useless (it's just a long column of Images)
- [ ] Hover Animation for the Video Preview. This one is a really nice one to have, and is also tied to functionality.
- [x] Proper Carousel Animation. (done in b684f0b )

## Feedback

Any type of feedback is welcome.


bedeho (changes requested):
# DevOps

1. The commit history has lots of noise, why does your source branch have lots of unrelated work?

2. For some reason, Storybook still is not rendering fonts properly, so its hard to make sense out of that preview. What is required to address this?

3. I know we discussed prior where to draw the line in terms of what components should be put in the component library, but independent of that, it would be very useful to be able to review variations on full scenes in a Storybook, rather than looking at an app. For example, 
  - I would like to review all the different modes of a video preview, for example, one with hovering, one with press, one which has been partially viewed etc. I can't do that now anywhere without firing the whole app.
 - I would like to see how a component looks with different types of images used. Right now I am looking at the app, which will have a fixed set of hardcoded images, but often the feasibility of a design also depends on how well it works with a range of realistic imagery.
 -  I would like to review a still version of a major screen that just shows the loading state, or perhaps different substages of loading - where different parts have come in.

# Presentation

1. The font also is wrong here.

2. The height between lines here seems to small compared to design

Screenshot 2020-06-17 at 16 17 59

3. Can we replace the background image from the head to a big photo which scales and has a suitable black blend. For example the new york photo on the other screen. He introduced the head at random, but I dont think its a good example of what that space would be used for.

4. Horizontal scaling down should resize the big main title to be smaller, rather than wrap

Screenshot 2020-06-17 at 16 21 27

5. This alignment is way off

Screenshot 2020-06-17 at 16 21 03

6. The content needs to behave differently when the viewport scales. Right now, they have a fixed size, but they should flow with the size of the viewport. You should also take the freedom to then adjust how many elements to show horizontally as function of horizontal space, so as to avoid having extremely large or extremely small content previews. The fixed size also creates an unfortunate interaction between the < and > carousel wheels. They should always be aligned with the end of some content, but now they end up clipping.

Screenshot 2020-06-17 at 16 26 15

Screenshot 2020-06-17 at 16 31 39


7. The content preview should have a black blend in the edges, and the wheel should be centered on the image, not the total vertical height of the preview, because almost all of the visual weight is in the image.

Screenshot 2020-06-17 at 16 26 48

8. Spacing seems very off here compared to design

Screenshot 2020-06-17 at 16 28 06

9. This is not in all of the designs, but would you mind aligning the main title on the page with the correct left position of the content and section titles?

10. The category bar does not have a carousel button, but I think we will not have it in the end anyway, so its fine.

11. This seems to be missing quite a few places

Screenshot 2020-06-17 at 16 29 50

fcescob:
I have a couple of followup questions:

1. I remember there were some small issues with the fonts licensing, should I go ahead and replace the fonts with the ones in the design?

2. Regarding the variety of images used, I can either add more mocked ones (all the ones available on figma) or I can actually take them from random image API, the former is trivial to implement while the latter takes more time but gives way more image variety.

Also, the head svg has been replaced with a background image in 5c2b5b1

bedeho:
1. @bwhm can help with the license.
2. Lets just use the ones on Figma for now.

bedeho:
I forgot to add one thing in the review: I think the carousel really has to navigate the full width of the currently visible row, not just one-by-one element, its not an effective way to get through all the content.

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -1,47 +1,86 @@
...
-  poster?: string
-  onClick?: any
-  onChannelClick?: any
-} & VideoPreviewStyleProps
+	title: string;
+	channel?: string;
+	channelImg?: string;
+	showChannel?: boolean;
+	showMeta?: boolean;
+	time?: string;
```
kdembler: Do we want this to be a string and give the responsibility of time formatting to the consumer? I would expect this to be a timestamp/`Date`/`luxon`/something else that holds the date.

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -1,47 +1,86 @@
...
+	views?: string;
```
kdembler: Shouldn't that be a `number` instead?

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -1,47 +1,86 @@
...
+	poster?: string;
+	onClick?: any;
```
kdembler: Wouldn't it be better to have at least very generic Function interface here? Because that's typed as `any` the TS compiler will skip checking for whether the value is actually there. There's a bug currently in the code, because `onClick` is possibly `undefined` but we call it unconditionally. If we change the prop type from `any` then the compiler will give us an error (as it should)

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -1,47 +1,86 @@
...
-          )}
-        </div>
-      </div>
-    </div>
-  )
+	let styles = makeStyles({ showChannel, poster, ...styleProps });
+	return (
+		<div css={styles.container} onClick={onClick}>
+			<div css={styles.coverContainer}>
+				<div
```
kdembler: What's the reason behind this being a `div` instead of an `img`? I believe if an image is part of the content (as it is here) then it should be an `img` tag (for SEO, for screen readers, etc.)

kdembler:
My small input on carousel and animations: maybe instead of doing a classical carousel that's fully controlled by the code, we could do a scroll-based carousel, something like https://nickpiscitelli.github.io/Glider.js/? Then we don't have to implement some (or any) of the custom animations and swiping works out of the box (for both mobile and desktop). That's an area I haven't explored that much so just something to think about

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Yes absolutely. This is just placeholder content for now, so I wanted to use the most basic things possible to get a feel of how things look like.

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Same as above, label this as a string to skip all number processing

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Good point, will change this to a regular `img` tag

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
kdembler: Okay, thanks for clearing that up, we may change the type later once we decide on what to use to format dates.

One more suggestion I have would be changing the field name to something more specific. `time` can be ambiguous, something like `createdAt` could be a standard prop name for every field like this. If you feel that this should be done outside of this PR then I'm fine with that as well.

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Done in 4177239 .

When we start adding stateful logic, we should add a `srcset` and serve different image sizes based on screensize.

fcescob:
The last commit ( 4177239 ) fixes the Carousel, it correctly handles scrolling when the Carousel container is resized and overscrolling is prevented. 
The Carousel wheels have also been centered, with the special case of being centered in the middle of the thumbnail in the case of a video gallery.
Some images have stopped working, but I guess it is because they were the external URLs taken directly from Figma.

fcescob:
DevOps 2&3 have been moved to https://github.com/Joystream/joystream/issues/770

fcescob:
Point 7 of presentational issues moved to https://github.com/Joystream/joystream/issues/771

fcescob:
Point 9 of presentational issues moved to #772

fcescob:
Moved Fluid Typography(header size on small screens) to #773

fcescob:
Point 6 of presentational issues moved to #774

fcescob:
Header is currently too short and has additional markup, opened issues #778 .

atlas/packages/app/src/components/VideoGallery.tsx
```diff
@@ -0,0 +1,127 @@
...
+	margin: auto ${theme.spacing.m};
+`;
+
+export default function VideoGallery({ title = "" }: VideoGalleryProps) {
+	const videos = videoPlaceholders.concat(videoPlaceholders).concat(videoPlaceholders);
+	const [controlsTop, setControlsTop] = useState<SerializedStyles>(css``);
+	const imgRef = useCallback((node: HTMLImageElement) => {
+		if (node != null) {
+			setControlsTop(css`
+				top: calc(${Math.round(node.clientHeight) / 2}px - 24px);
```
kdembler: Could you explain what's the purpose of this? (passing ref, etc.)

atlas/packages/app/src/components/VideoGallery.tsx
```diff
@@ -0,0 +1,127 @@
...
+const articleStyles = css`
+	max-width: 320px;
+	margin: auto ${theme.spacing.m};
+`;
+
+export default function VideoGallery({ title = "" }: VideoGalleryProps) {
+	const videos = videoPlaceholders.concat(videoPlaceholders).concat(videoPlaceholders);
+	const [controlsTop, setControlsTop] = useState<SerializedStyles>(css``);
+	const imgRef = useCallback((node: HTMLImageElement) => {
+		if (node != null) {
```
kdembler, line 106: We should probably never do loose equality comparisons and we should have an ESLint rule to enforce this.

We can also do a broader check than just for `null` here with `if (node)`

atlas/packages/app/src/components/VideoGallery.tsx
kdembler: It seems like we want to get the height of the image for some reason, but that height is directly controlled by the prop of `VideoPreview`. That tells me there's some issue here. It seems to me we should either:
a) set the height via prop and not check the ref
b) have `VideoPreview` choose the height automatically and then check it via ref
Currently, in this component we're checking a value that's ours to set.

Another question is whether we should use `height` and `width` props at all. By using them we're leaking styling inside the code. Maybe `VideoPreview` should just have some default value that can be overridden via styling, not props?

fcescob: I needed to align the wheels of the carousel to the middle of the `VideoPreview` thumbnail / image but, in general, I had no way of knowing the height of a thumbnail at priori (things could be resized etc...). 
So my solution was to attach a callback ref to the image inside `VideoPreview` to get the height and set the correct position of carousel wheels. 

It feels a bit hacky, so I'm _very_ open to suggestions or different implementations.

kdembler: Does my 2nd answer give you some ideas? I think we may have posted at the same time

atlas/packages/app/src/components/VideoGallery.tsx
fcescob: I agree on adding an ESLint rule to check for loose equality check, the `null` comparison here is very safe though, since `x == null` will return `true` iff `x === undefined || x === null` and there is no type coercion involved

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
kdembler: @Gamaranto `onChannelClick` was changed but `onClick` is still `any`. I'm also wondering whether we need to pass the event down the tree or maybe an empty callback will be enough

atlas/packages/app/src/components/VideoGallery.tsx
kdembler: I agree, here it shouldn't do anything weird, but I think it's a good rule to use strict equality wherever we can.

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -1,47 +1,68 @@
...
+	poster,
+	onClick = () => {},
+	onChannelClick = () => {},
+	...styleProps
+}) => {
+	let styles = makeStyles({ showChannel, poster, ...styleProps });
+	return (
+		<div css={styles.container} onClick={onClick}>
+			<div css={styles.coverContainer}>
+				<img src={poster} ref={imgRef} css={styles.cover} alt={`${title} by ${title} thumbnail`} />
```
kdembler: Why `${title} by ${title}`?

atlas/packages/app/src/components/VideoGallery.tsx
kdembler: Also, if the line is left as it is, we shouldn't use a magic value in the code that's not described in any way. `24px` here is just an arbitrary value and I have no idea what it's supposed to mean. Does it correlate to some other style? Does it have to be changed along with some other value in another file?

If possible in a case like this, this should be some imported variable, if not, then we should add a comment.

kdembler: Btw, why do we even need to pass the height to carousel controls? Shouldn't they just reside inside the carousel container and center themselves vertically automatically?

atlas/packages/app/src/App.tsx
```diff
@@ -1,33 +1,29 @@
...
-              videos={videos}
-            />
-            <VideoView path="/videos/:idx" channels={channels} videos={videos} />
-          </ScrollToTop>
-        </Router>
-      </Provider>
-    </main>
-  )
+	return (
+		<main className="main-section">
```
kdembler: What do we need this `className` for?

atlas/packages/app/src/App.tsx
```diff
@@ -1,33 +1,29 @@
...
-
-let { channels, videos } = data
+const backgroundImg =
+	"https://s3-alpha-sig.figma.com/img/351c/9556/7bd65e79c56d66e273279ce04f8574aa?Expires=1593388800&Signature=PAvJw8izjlFlZRMv9V9zMNkyZPimYsj4PzJGXk0Q6mDPlV6Z87PLAT9xyImbi~p263Mu7bMcT8jdqjayXQq3iEgn1JxdYVCrHSyXgCrfmwxmWMZktEY-4kyO9M1RSWBlF9fsd19a45m0L9NtesyDU0IYz51tfYwE2OUE21TJmi4vdjiilPg6kRhPA3Z2N-Vj9ir9of7PtidoQwF-8b9fPejs7nDAHs3AX6YpIYTA36v2TGtS1emtxOCrO6JaRkr61yEdERo1RAVIeK36TeWXcCNE~mIIVoy~A9Abics4O2yQc0oxspu0Ac987NC-kGAHQS8FNeiAxWy7Ux49oNEMfQ__&Key-Pair-Id=APKAINTVSUGEWH5XD5UA";
 
 export default function App() {
-  return (
-    <main className="main-section">
-      <Provider store={store}>
-        <Router primary={false}>
```
kdembler, line 19: Shouldn't we keep the router?

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
+		const childrensLength = elementsRefs.current.reduce(
+			(accWidth, el) => (el != null ? accWidth + el.clientWidth : accWidth),
+			0
+		);
+		setChildrenLength(childrensLength);
+	}, [children.length]);
 
-	let styles = useCSS(styleProps);
-
+	const styles = useMemo(() => useCSS({}), []);
```
kdembler: `useMemo` isn't necessary here since it has empty dependency array anyway

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
-		<div css={styles.wrapper}>
-			<div ref={container} css={styles.container}>
-				{children.map((item, index) => (
-					<div key={`carousel-${index}`} css={styles.item}>
-						{item}
-					</div>
+		<div css={[styles.container, containerCss]}>
+			<div css={styles.itemsContainer} ref={containerRef}>
+				{children.map((element, idx) => (
+					<animated.div
```
kdembler: Is there a specific reason why we animate each child element separately? This seems like a lot of unnecessary work. Couldn't we have all the elements inside some "track" container and translate that track instead? Then we have only one element to animate.

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
+				direction="right"
+				onClick={() => {
+					handleScroll("right");
+				}}
+			/>
 		</div>
 	);
-}
+
+	function handleScroll(direction: "left" | "right") {
```
kdembler, line 42: This should be defined before the `return`. It's a good practice to always have the `return` statement be the last one in a function component.

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
+	if (!Array.isArray(children)) {
+		return <>{children}</>;
 	}
+	let [props, set] = useSpring(() => ({
+		transform: `translateX(0px)`,
+	}));
+	const [x, setX] = useState(0);
+	const { width: containerWidth, ref: containerRef } = useResizeObserver<HTMLDivElement>();
+	const elementsRefs = useRef<(HTMLDivElement | null)[]>([]);
+	const [childrenLength, setChildrenLength] = useState(0);
```
kdembler: What is this variable? Looking at the `useEffect` it looks like total width of all the items but the name suggests it's a length (of an array my first thought was)

atlas/packages/app/src/App.tsx
kdembler: This is carried from the previous code, but it seems we don't need it

atlas/packages/components/scripts/build-index.js
```diff
@@ -1,17 +0,0 @@
-const fs = require("fs");
```
kdembler, line 1: This script got deleted but is still referenced in `README.md` and `package.json`

atlas/packages/app/src/App.tsx
fcescob: nice catch, will clean it up.

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: Length is indeed confusing, will change it to width

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Typo, fixed it :)

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
+	children,
+	containerCss,
+	leftControlCss,
+	rightControlCss,
+	onScroll = () => {},
+}) => {
+	if (!Array.isArray(children)) {
+		return <>{children}</>;
 	}
+	let [props, set] = useSpring(() => ({
```
kdembler: This naming is confusing as well. `props` in React is a word with a specific meaning, as props being passed to a component. `set` also doesn't describe what the intent is

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,105 @@
...
-					<div key={`carousel-${index}`} css={styles.item}>
-						{item}
-					</div>
+		<div css={[styles.container, containerCss]}>
+			<div css={styles.itemsContainer} ref={containerRef}>
+				{children.map((element, idx) => (
+					<animated.div
+						style={props}
+						key={`Carousel-${idx}`}
+						ref={(el) => {
```
kdembler: I believe this can be just `ref={elementsRefs.current[idx]}` (not 100% sure)

fcescob:
After adding some more issues, the following list should include all UI incosistencies, DevOps enhancements , and code organization issues mentioned above. 
Each bullet point has a brief description of the problem together with a link to an issue which should be closed only when a PR solves said problem. 

I almost certainly missed something, so please double check the list to make sure it covers everything.

## Presentational 
- Channel Avatar should have the initials of the channel name in uppercase when no picture is available #800 
- Carousel wheel should not show when scrolling is not possible: #801
- Tag buttons in "Top Category" are missing proper hover states #802
- Channel Card UI is way off: #803
- Playlist Preview is just a regular VideoPreview for now #804
- Video Preview is missing some hover states #749 
- Menu is missing #483
- Navbar is missing #747 
- Footer is missing #805 

## DevOps 
- Storybook lacks theming and doesn't render the correct fonts #770 
- Some components are missing from the stories #806
- Chromia preview doesn't display #808
- Tests are not working #748
- Improve PR checks #813 
- Linter tweaks #811 

## Code Quality

- Refactor all React components to an arrow function syntax. #807 
- Clean up "dead" code #809

kdembler:
The list nicely summarizes what's left to do but it misses a lot of points I've made above regarding the code itself (not organization). I'm not sure if it was your intention to include those as well.

Most of the comments I've made aren't a lot of work, but I have slightly bigger concerns regarding the Carousel. More specifically, the biggest things from my comments are animating each child element separately (I think we should animate their container instead) and the need to manually set the control button styles by the Carousel consumer (I believe they should be centered out of the box)

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: fixed

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
fcescob: Fixed

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: Fixed in the last commit

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: Renamed `props` to `scroll` and `set` to `setScroll`

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: I don't think changing that would work, it gives the following TS error which is quite expressive:
```typescript
Type 'HTMLDivElement | null' is not assignable to type 'string | RefObject<HTMLDivElement> | ((instance: HTMLDivElement | null) => void) | null | undefined'.
  Type 'HTMLDivElement' is not assignable to type 'string | RefObject<HTMLDivElement> | ((instance: HTMLDivElement | null) => void) | null | undefined'.
    Type 'HTMLDivElement' is not assignable to type 'string'.ts(2322)
```

`elementsRefs.current` holds an array of `HTMLDivElements` and an `HTMLDivElement` can't be passed as a ref since react will only accept a ref object, a callback ref or null. 
I could make `elementsRefs` hold `RefObjects` instead of `HTMLDivElements`, but that has the price of making other parts of the code less readable. 
Not worth it imho.

Another possible way of shortening this is by changing it to:  
```javascript
<animated.div
     // ... Other props
     ref={el => (elementsRef.current[idx]=el)}
</animated.div>
```
Which is shorter, but harder to read / understand at a glance.

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: Moved it up. I like to hide functions under return to hide the implementation details away, but this one is quite short, so I moved it up.

atlas/packages/app/src/components/VideoGallery.tsx
fcescob: Let me try to dissect what's going on here:

- `VideoGallery` is the common ancestor of both `Gallery` and each `VideoPreview` children.

- `Gallery` centers the carousel wheels in the middle of the container by default, there are however some cases, like when we want to center the wheels in the middle of a video thumbnail, where the center of the container is not the desired position of the wheels.
To account for those cases, `Gallery` exposes a `leftControlCss` and a `rightControlCss` prop, those are escape hatches to reposition the wheels from the center of the `Gallery` container. 

- We want to center the wheels in the middle of the video thumbnail, to do so we need to measure the height of the thumbnail (the purpose of the `imgRef`) and then reposition the Carousel wheels accordingly, this has to happen in `VideoGallery` since it is the common ancestor.

Now coming to your concerns:

- I don't understand what you mean by `width` and `height` props, if you refer to the ones of the thumbnail, so just hardcode them and get rid of the ref to clean code, I don't think that would work since I would expect `VideoPreview` to adjust it's own `width` and `height` to preserve a fixed aspect ratio on resizing, so an "external" component would have no way of knowing those values.

- The 24px is just the height of the Carousel wheel. I could either add a ref to measure it or label it with a value, I would just go for the labeling since it's the wheels don't resize on smaller screensizes.

atlas/packages/components/src/components/Carousel/Carousel.tsx
kdembler: You're right, we would need to hold `RefObject[]` instead, it may not be worth it.

Regarding the second part of your response, I think we should go that way. I actually believe `ref={el => (elementsRef.current[idx]=el)}` is easier to understand, since it does just one thing - assign the ref, we don't need to return it as well in that function, right?

atlas/packages/app/src/components/Hero.tsx
```diff
@@ -0,0 +1,89 @@
...
+					display: flex;
+					margin-top: 40px;
+					& > * {
+						margin-right: 1rem;
+					}
+				`}
+			>
+				<Button
+					containerCss={css`
+						width: 116px;
```
kdembler, line 35: Why do we manually set size of those buttons? These should be based on some common sizing. (for example, constant height, width based on content). If we really must set the width, then at least the height shouldn't be here

atlas/packages/app/src/components/Hero.tsx
```diff
@@ -0,0 +1,89 @@
...
+					`}
+				>
+					Share
+				</Button>
+			</div>
+		</Header>
+	);
+};
+export default Hero;
+// export default function Hero({ backgroundImg }: Partial<HeroProps>) {
```
kdembler: Is this still needed?

atlas/packages/app/src/components/VideoGallery.tsx
kdembler: Okay, now I see what's going on with the manual styling of the wheels and that looks okay.

Regarding my concerns:
- `width` and `height` props: what I mean here is that `VideoPreview` exposes props to set these values. So `VideoGallery` as `VideoPreview` consumer is responsible for setting these values if it wants them other than the default. But even though `VideoGallery` is the de factor controller of `VideoPreview`'s cover `height`, it still checks it via a ref. So we have a communication issue here - we check a value that's ours to set. The easiest approach would probably be to remove `height` and `width` props from `VideoPreview`

- the `24px` value. Sure, if this will be a constant then there's no point of measuring it. My point was that we shouldn't use a value like this outside of its context without any description. If this is the height of the wheel, let's declare it alongside wheel styles and import here.

atlas/packages/app/src/components/Hero.tsx
fcescob: Forgot it, will clean it up.

atlas/packages/components/src/components/Button/Button.style.ts
```diff
@@ -1,10 +1,10 @@
 import { typography, colors } from "../../theme"
 import { makeStyles, StyleFn } from "../../utils"
-import { disabled, dimensionsFromProps, log } from "../../theme/fragments"
+import { disabled, dimensionsFromProps } from "../../theme/fragments"
 
 export type ButtonStyleProps = {
 	text?: string
-	type?: "primary" | "secondary"
+	type?: "primary" | "secondary" | "tertiary"
```
kdembler, line 7: Is `tertiary` a term used when describing styles? I haven't encountered it before. `primary` and `secondary` are pretty standard, maybe here we could use a name describing the button better? Like `transparent`, `info`, etc?

atlas/packages/components/src/components/Carousel/Carousel.tsx
```diff
@@ -1,39 +1,107 @@
...
-	const container = useRef(null);
+const Carousel: React.FC<Partial<CarouselProps>> = ({
+	children,
+	containerCss,
+	leftControlCss,
+	rightControlCss,
+	onScroll = () => {},
+}) => {
+	if (!Array.isArray(children)) {
+		return <>{children}</>;
```
kdembler: This will be visible better once we enforce eslint, but we shouldn't return here because we use hooks later on. Per hooks rules they shouldn't be conditional so we can return only after we've called all the hooks

atlas/packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -15,23 +13,23 @@ type DropdownProps = {
...
+	options = [],
 	disabled = false,
 	onChange = () => {},
 	...styleProps
 }: DropdownProps) {
 	const inputRef = useRef(null);
 	const [isActive, setIsActive] = useState(!!value);
 	const [inputTextValue, setInputTextValue] = useState(
-		!!value ? options.find(option => option.value === value).text : ""
+		!!value ? options.find(({ value: optionValue = NaN }) => optionValue === value)?.text : ""
```
kdembler: Why do we use `NaN` as a default value here?

atlas/packages/components/src/components/Header/Header.style.ts
```diff
@@ -1,30 +1,67 @@
...
+	marginTop: spacing.m
+})
+
+export const useCSS = (props: HeaderStyleProps) => ({
+	container: makeStyles([container])(props),
+	content: makeStyles([content])(props),
+	title: makeStyles([title])(props),
+	subtitle: makeStyles([subtitle])(props)
+})
+// export let makeStyles = ({
```
kdembler: Do we need this?

atlas/packages/components/src/components/RadioButton/RadioButton.tsx
```diff
@@ -10,13 +10,13 @@ export default function RadioButton({
...
 	disabled = false,
-	onClick,
+	onClick = () => {},
 	...styleProps
 }: RadioButtonProps) {
 	const styles = useCSS({ disabled, position, ...styleProps });
 
 	return (
-		<div css={styles.container} onClick={disabled ? null : onClick}>
+		<div css={styles.container} onClick={onClick}>
```
kdembler: Why this change? We probably shouldn't call the click handler when the button is disabled (so unclickable)

kdembler: Is it a case that handler won't be called anyway because the button is disabled in the DOM?

atlas/packages/components/src/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -2,32 +2,32 @@ import { css } from "@emotion/core";
 import { breakpoints } from "../../theme";
 
 export type VideoPlayerStyleProps = {
-  width?: string | number;
-  height?: string | number;
-  responsive?: boolean;
-  ratio?: string;
+	width?: string | number;
```
kdembler, line 5: Why these indent changes? I thought we are using 2 spaces? There's a couple more files affected by this

atlas/packages/components/src/components/Button/Button.style.ts
fcescob: It's not super common, but you can find it paired with `primary` and  `secondary`, like here https://uxdesign.cc/ui-cheat-sheets-buttons-7329ed9d6112

atlas/packages/components/.storybook/theme.js
```diff
@@ -0,0 +1,44 @@
+import { create } from "@storybook/theming/create";
+/* Can't import the theme directly here because it is written in TS.
```
kdembler: Are you sure that's the case? I've managed to import the theme no problem.

One more thing I ended up doing was wrapping the story in `Layout` component from the `app` package. Not sure if that's an anti-pattern with how our code is structured but that enabled no code duplication and one-to-one styling with the app. Potentially we could move the `globalStyles` to `components` package and alleviate this problem. But then again, I don't think this is in the scope of this PR.

fcescob: Just checked, and looks like I can import fine too, not sure what was the issue I had. Nonetheless, I though about it and I think it's better to keep it self contained, so we can reuse it as is for the website Storybook and for future projects.

atlas/packages/app/src/components/Hero.tsx
fcescob: Removed `height`

atlas/packages/app/src/components/Hero.tsx
fcescob: Cleaned it up

atlas/packages/components/src/components/Header/Header.style.ts
fcescob: Cleaned it up.

atlas/packages/components/src/components/Carousel/Carousel.tsx
fcescob: Moved it down and added a check to `useEffect` to prevent errors.

atlas/packages/components/src/components/RadioButton/RadioButton.tsx
fcescob: You're right, that was necessary, I think I had the intention to rewrite the Radio Button component but ended up skipping it in this PR. I re-added the tertiary operator and moved the rewrite to issue #836.

atlas/packages/components/src/components/Dropdown/Dropdown.tsx
fcescob: Was nonsense, changed it to an empty string instead.

atlas/packages/app/src/components/VideoGallery.tsx
kdembler: @Gamaranto I believe this is still unresolved

Edit: Seems Github only elevated the reply and not the thread, this is what I'm referring to: https://github.com/Joystream/joystream/pull/693#discussion_r445491276

atlas/packages/app/src/App.tsx
kdembler: @Gamaranto I don't think this has been addressed

kdembler (changes requested):

atlas/packages/app/src/App.tsx
fcescob: Let's keep it out of this PR and re-add it later when we actually use it, opened an issue about it #861

kdembler: We can do that, but why remove it in the first place since we'll be adding it back for sure?

fcescob: Well, I've added it back, it doesn't hurt I guess :D

atlas/packages/app/src/components/VideoGallery.tsx
fcescob: I see what you mean now, and I do agree with the points you raised. 
I can't see a trivial way to fix the first point, so I've just removed `width` and `height` from the props as you suggested. Regarding point 2, I've added it as a constant in `VideoGallery` since I believe height of the control whell might be contextual and adding it there naturally evolves towards measuring it with a ref if we ever need to.

kdembler: Awesome :)

kdembler (approved):
LGTM!

Issue #695, opened 2020-06-14
UI Issue: Spending Proposal submit button not working


mochet:
The UI will not allow for a spending proposal to be sent through, "submit proposal" remains inactive regardless of input. Submitting via extrinsic does work.
![image](https://user-images.githubusercontent.com/50830161/84595818-16793f80-ae6b-11ea-9f02-e395c54b7af6.png)

I think this was working before, so maybe something broke it with a recent Pioneer update

bedeho:
Thanks for reporting, I guess we should fix this ASAP.

Lezek123:
Should be fixed with: https://github.com/Joystream/joystream/pull/696

Issue #694, opened 2020-06-13
UI issue: Proposal not showing all votes


mochet:
Proposal:
https://testnet.joystream.org/#/proposals/12

Chain state data:
"votingResults":{"abstensions":0,"approvals":6,"rejections":0,"slashes":0

Issue:
Only 3/10 votes are showing on the UI. The proposal passed with 6 votes, but failed actual execution.

Lezek123:
As far as I remember only the votes of the current council members are beeing fetched, which is probably what causes the votes of no longer active council members not to be included on this proposal page (and can also be observed for other old proposals)

The reason it was implemented this way was because we had no good way of dealing with the `proposalsEngine.voteExistsByProposalByVoter` double map (we needed both proposal id and member id to retrieve the vote value, so to avoid checking all 450 existing member ids, we just decided to execute a query for each active council member, but didn't predict it would cause an issue like this one).

Now that there exists a workaround soultion I implemented while working on the proposal discussion system, that allows us to fetch all values in the double map using only the first key (proposal id in this case) - this issue shouldn't be hard to fix.

bedeho:
1. This seems to be a mistake on the runtime complicating things, as the memberid should not have been used to identify the voter, but instead the council member id, which is within a very small range.
2. All of this goes away with the query node, so do not overinvest in finding a perfect solution, just pick a workaround that fixes the obvious UI issue.

Lezek123:
Should be fixed with: https://github.com/Joystream/joystream/pull/696

PR #696, opened 2020-06-14
Pioneer proposals: spending address validation and council votes fix

Lezek123:
This PR covers possible fixes for issues https://github.com/Joystream/joystream/issues/694 and https://github.com/Joystream/joystream/issues/695

- Removes address validation with `checkAddress` as it turns out it doesn't work the way it was expected (requires different form of prefix) and in fact isn't necessary anyway, since address is already validated through the `InputAddress` component beeing used in `SpendingProposalForm`.
- Uses the new `doubleMapEntries` method to fetch proposal votes, instead of only checking the votes of active council members. The problem with the second approach is that the votes are "lost" on old proposal details pages when the council changes (and some of the old members are not part of the current council anymore). This also forces us to replace the votes section header from _**Votes: (X/Y)**_ (ie. **_Votes: (5/10)_**) to _**Votes (X)**_, since we're not aware of how many council members were active at the time when voting took place (which could possibly have been way back in the past). Alternatively we could add a query to fetch the number of currently active council members and display the **_Votes: (X/Y)_** header only when the proposal is still in the voting period (otherwise just display **_Votes: (X)_**) or use the query with `.at()` to get the number of active council member at the block when proposal was created (which would be probably the best solution).

### DIFF
--- a/pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
+++ b/pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
@@ -5,7 +5,7 @@ import Body from './Body';
 import VotingSection from './VotingSection';
 import Votes from './Votes';
 import { MyAccountProps, withMyAccount } from '@polkadot/joy-utils/MyAccount';
-import { ParsedProposal, ProposalVote } from '@polkadot/joy-utils/types/proposals';
+import { ParsedProposal, ProposalVotes } from '@polkadot/joy-utils/types/proposals';
 import { withCalls } from '@polkadot/react-api';
 import { withMulti } from '@polkadot/react-api/with';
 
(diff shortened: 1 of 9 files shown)


bedeho:
Awesome work, but doesn't this go in `master` ? I say this because we want these fixes to go in production now, right?

Lezek123:
I'm not sure if `master` branch is already configured to deploy Pioneer?
Last time I was introducing some fixes I was creating a PR both here and in the `Joystream/apps` repository (it went to `development` here and to `master` of `Joystream/apps`).
I can chage the base to `master` here, will just add one more commit to display the **_Votes: (X/Y)_** correctly.

Lezek123:
Implemented a solution described in the first comment to display the number of active council members at the time the proposal was created over the **_Votes_** section:
> Alternatively we could add a query to fetch the number of currently active council members and display the Votes: (X/Y) header only when the proposal is still in the voting period (otherwise just display Votes: (X)) **or use the query with .at() to get the number of active council member at the block when proposal was created (which would be probably the best solution).**

Lezek123:
Rebased and changed the PR base branch to `master`.

mnaamani:
> I'm not sure if `master` branch is already configured to deploy Pioneer?
> Last time I was introducing some fixes I was creating a PR both here and in the `Joystream/apps` repository (it went to `development` here and to `master` of `Joystream/apps`).
> I can chage the base to `master` here, will just add one more commit to display the **_Votes: (X/Y)_** correctly.

Yes netlify will now deploy from master branch to production testnet.joystream.org

mnaamani (approved):

mnaamani:
Tested :+1: merging

PR #697, opened 2020-06-15
Membership schema

metmirr:
This PR:
- Adds mappings for membership module
- Fixes `forEach` async/await problem 
- Use `bn.js` for large numbers. e.g block number, event index

### DIFF
--- a/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
+++ b/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
@@ -122,7 +122,9 @@ async function main() {
       "WS_PROVIDER_ENDPOINT_URI environment variable is not set! Make sure you set it in your .env file or system wide."
     );
   }
-
+  
+  // Required for warthog models.
+  dotenv.config({ path: '../graphql-server/.env' });
   await createConnection();
 
   const node = new QueryNodeManager(process);
--- a/query-node/substrate-query-framework/index-builder/src/db/helper.ts
+++ b/query-node/substrate-query-framework/index-builder/src/db/helper.ts
@@ -10,6 +10,9 @@ import { DeepPartial } from 'typeorm';
  * @param entity: DeepPartial<T>
  */
 export function fillRequiredWarthogFields<T>(entity: DeepPartial<T>): DeepPartial<T> {
+  // Modifying an existing entity so do not add warthog fields
+  if (entity.hasOwnProperty('id')) return entity;
+
   const requiredFields = {
     id: shortid.generate(),
     createdById: shortid.generate(),
--- a/query-node/joystream-query-node/bootstrap/members.ts
+++ b/query-node/joystream-query-node/bootstrap/members.ts
@@ -1,4 +1,4 @@
-import { Membership } from '../generated/graphql-server/src/modules/membership/membership.model';
+import { Member } from '../generated/graphql-server/src/modules/member/member.model';
 import { DB, getLogger } from '../generated/indexer';
 
 import { ApiPromise } from '@polkadot/api';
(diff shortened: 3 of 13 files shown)


bedeho (approved):
LGTM!

query-node/substrate-query-framework/index-builder/src/db/helper.ts
```diff
@@ -10,6 +10,9 @@ import { DeepPartial } from 'typeorm';
  * @param entity: DeepPartial<T>
  */
 export function fillRequiredWarthogFields<T>(entity: DeepPartial<T>): DeepPartial<T> {
+  // Modifying an existing entity so do not add warthog fields
+  if (entity.hasOwnProperty('id')) return entity;
```
dzhelezov, line 14: What about the extra fields like `createdAt` ? Do we still need them in this case?

query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
```diff
@@ -122,7 +122,9 @@ async function main() {
       "WS_PROVIDER_ENDPOINT_URI environment variable is not set! Make sure you set it in your .env file or system wide."
     );
   }
-
+  
+  // Required for warthog models.
+  dotenv.config({ path: '../graphql-server/.env' });
```
dzhelezov, line 127: This makes the indexer dependent on the graph server, isn't it? What env properties do we need here apart from the DB?

metmirr: Yes. Just `WARTHOG_APP_HOST` and `WARTHOG_APP_PORT` actually, DB env is the same with warthog. Should I add them to indexer's `.env` or?

query-node/substrate-query-framework/index-builder/src/db/helper.ts
metmirr: This guard is for updating existing db records. We don't need those extra fields but warthog `BaseModel` adds them :(

query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
dzhelezov: Yeah, I think we already have `templates/dotenv.mst` for that don't we?

query-node/substrate-query-framework/index-builder/src/db/helper.ts
dzhelezov: Ah ok. I thought that this check is like 'if the user has provided an ID field in the schema, use it'

query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
dzhelezov: Probably a more serious problem here is that we somehow need to import the model files from graphql-server. I remember we already had this issue, how do we solve it?

metmirr: Yeah, we have discussed importing models but we don't have a solution yet.

query-node/substrate-query-framework/index-builder/src/db/helper.ts
metmirr: 👍

query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
dzhelezov: Ok, it's out of scope for this PR, but here's an issue for the future reference: https://github.com/Joystream/joystream/issues/723

metmirr: 👍

PR #699, opened 2020-06-15
Pioneer forum improvements

kdembler:
This PR tackles all the forum improvements requested in #558.

### Notable changes

- Categories list columns were updated and recent activity (in unique threads) was added to the top-level categories.
Screenshot 2020-06-19 at 12 30 44 PM

- Thread view got refreshed so that:
  - Thread creator and timestamp are presented
  - Each post has a timestamp and post number
  - Post actions are in a separate row
Screenshot 2020-06-19 at 12 33 58 PM

- Thread reply form was moved inside the thread view and it supports both new reply and editing.
Screenshot 2020-06-19 at 12 37 37 PM

- Each post can be quoted using the button in the action row. At the moment, quoting supports only post text, adding author name, for example, would require some async fetching and additional work that may not be worth at the time.
Screenshot 2020-06-19 at 12 38 44 PM

- Each post has a linkable number that once clicked, selects the post, and inserts its ID into the URL. This URL can be shared with others so that they will be taken directly to this specific reply.

- Threads list now lists the thread creation date.
Screenshot 2020-06-19 at 12 31 47 PM

- Forum tabs were removed and replaced with breadcrumbs.
Screenshot 2020-06-19 at 12 29 41 PM

- Some visual tweaks were performed:
  - Increasing font size for forum texts
  - Displaying properly-cased thread title
  - Styling of markdown blockquotes

### Items that weren't resolved

- Incorrect display of manual linebreaks was reported in the issue but that's expected behavior of Markdown (double new line is required). To keep the syntax standard, no changes were made.

- Lack of support for intended lists was reported as well, but that's something I couldn't reproduce:
Screenshot 2020-06-19 at 12 48 50 PM
@bedeho do you have any more details on that specific item?

### Rationale behind some changes
So that some specific actions/items can be linked, query params in the URL were used. However, for some reason, current forum paging (threads and replies) wasn't using query params but page number inside the path itself (e.g. `forum/threads/1/page/2`). That is a pretty non-standard approach.

To make the paging more predictable and to enable easier integration with other features, paging was switched to query param based one (e.g. `forum/thread/1?page=2`). So that old links (in proposals for example) still work properly, a redirect component was added that will translate previous paging schema to a new one.

### DIFF
--- a/pioneer/packages/joy-forum/src/CategoryList.tsx
+++ b/pioneer/packages/joy-forum/src/CategoryList.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
 import ReactMarkdown from 'react-markdown';
 import { Table, Dropdown, Button, Segment, Label } from 'semantic-ui-react';
+import styled from 'styled-components';
 import { History } from 'history';
 import orderBy from 'lodash/orderBy';
 import BN from 'bn.js';
@@ -108,6 +109,10 @@ type ViewCategoryProps = InnerViewCategoryProps & {
   id: CategoryId;
 };
 
+const CategoryPreviewRow = styled(Table.Row)`
+  height: 55px;
+`;
+
 function InnerViewCategory (props: InnerViewCategoryProps) {
   const { history, category, page = 1, preview = false } = props;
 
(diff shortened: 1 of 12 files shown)


bedeho (commented):
Fantastic work tackling such a diverse range of problems in a messy code base, and solving many UX problems at the same time. I love the recent activity view, its a game-changer. I will leave it to @Gamaranto to review implementation, the only thing I wonder if it would be even better would be if the  `Recent Activity` was lifted above the categories list. What do you think? For most recurring users, this is the point of primary interest, not the category view.

fcescob (approved):

kdembler:
@bedeho sure thing, moved the activity above the categories.
Screenshot 2020-06-22 at 11 22 33 AM

fcescob (approved):

Issue #578, opened 2020-06-05
Pioneer: Storage working group overview on "Working Group" tab


Lezek123:
There should be an overview similar to the one currently displayed for _Content curators_ on the `/working-groups` page.

This includes:
- Short description of the Storage Working Group role.
- List of members.
- Information whether some opportunities to join are currently available. If so - a link to the relevant page (`/working-groups/opportunities/storage`)
- Current lead

Probably some layout changes should be considered, because if we decide to add a few more groups this page will quickly become "oversaturated" with information.

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/711

PR #711, opened 2020-06-16
Storage group overview

Lezek123:
Implements https://github.com/Joystream/joystream/issues/578 - storage working overview similar to content working group overview on the `/working-groups` index page.

All content working group functionality should continue to work correctly.

Some storybook previews have been "broken", but since a lot of them didn't work even before those changes I assumed it's not that important (just fixed the TS issues for now).


mnaamani (approved):
Nice, tested against latest runtime #672 with slightly modified #722 
I'll include a minor fix in 722 for renamed runtime module

Issue #679, opened 2020-06-12
Network testing: worker application happy case


gleb-urvanov:
Implement the following scenario:
- run "buy membership" testing scenario, resulting in N member accounts;
- set lead using sudo;
- add N worker openings;
- begin accepting worker application;
- apply on worker openings using N member accounts;
- begin worker applications review;
- fill working openings.
The scenario will result in creation of N workers

gleb-urvanov:
https://github.com/Joystream/joystream/pull/713

PR #713, opened 2020-06-16
Network testing: worker application happy case

gleb-urvanov:
https://github.com/Joystream/joystream/issues/679


gleb-urvanov:
included into https://github.com/Joystream/joystream/pull/743

Issue #656, opened 2020-06-10
All events include reference counting delta sideffects


bedeho:
The query node needs to have such information added to all applicable content directory events in order to maintain this information in the index.

PR #715, opened 2020-06-17
All events include reference counting delta sideffects

iorveth:
An attempt to implement https://github.com/Joystream/joystream/issues/656

Depends on: https://github.com/Joystream/joystream/pull/669
Closes: https://github.com/Joystream/joystream/issues/656

### DIFF
--- a/runtime-modules/content-directory/src/lib.rs
+++ b/runtime-modules/content-directory/src/lib.rs
@@ -1917,30 +1917,42 @@ impl<T: Trait> Module<T> {
             DeltaMode::Increment,
         );
 
-        let reference_counter_side_effects = match (
+        // Add up both net decremental_reference_counter_side_effects and incremental_reference_counter_side_effects
+        // to get one net sideffect per entity.
+        let reference_counter_side_effects = Self::calculate_updated_inbound_rcs_delta(
             decremental_reference_counter_side_effects,
             incremental_reference_counter_side_effects,
-        ) {
-            (
-                Some(decremental_reference_counter_side_effects),
-                Some(incremental_reference_counter_side_effects),
-            ) => {
-                // Add up both net decremental_reference_counter_side_effects and incremental_reference_counter_side_effects
-                // to get one net sideffect per entity.
-                let reference_counter_side_effects = decremental_reference_counter_side_effects
-                    .update(incremental_reference_counter_side_effects);
-                Some(reference_counter_side_effects)
-            }
-            (Some(decremental_reference_counter_side_effects), _) => {
-                Some(decremental_reference_counter_side_effects)
-            }
-            (_, Some(incremental_reference_counter_side_effects)) => {
-                Some(incremental_reference_counter_side_effects)
-            }
-            _ => None,
-        };
+        );
 
         Ok(reference_counter_side_effects)
+    }
+
+    // Add up both net first_reference_counter_side_effects and second_reference_counter_side_effects (if some)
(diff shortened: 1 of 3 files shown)


bedeho (changes requested):
This looks correct, except that `map` is not meant for iteration + side effects (`::update_entities_rcs`), it is meant to purely compute a new value based on an old one. It's important to preserve such semantics for the reader. Also, by combining the call, and doing a map that has no effect on the final value, you are decreasing clarity without adding any upside.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1478,18 +1490,27 @@ decl_module! {
             //
 
             // Decrease reference counters of involved entities (if some)
-            if let Some(entity_ids_to_decrease_rcs) = property_value_vector.get_vec_value().get_involved_entities() {
+            let entities_inbound_rcs_delta = if let Some(entity_ids_to_decrease_rcs) =
```
bedeho: This set of steps is quite involved, could you perhaps 
a) move it out into a standalone function, and just have that function compute the final `entities_inbound_rcs_delta`, so that reading the main extrisnics body is a bit easier, and the reader can go and read only this function if they want to make sure its correct.
c) in that function, can you break things up a bit. For example you are doing a two function calls, a `Option` match, and an `if` body computation, all in one line (1493). This can be broken up in steps for greater readability. It also is a bit rough with different variabls that have the same name, but different type, in different near by scopes.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1867,10 +1904,30 @@ impl<T: Trait> Module<T> {
             DeltaMode::Increment,
         );
 
-        // Add up both net decremental_reference_counter_side_effects and incremental_reference_counter_side_effects
-        // to get one net sideffect per entity.
-        Ok(decremental_reference_counter_side_effects
-            .update(incremental_reference_counter_side_effects))
+        let reference_counter_side_effects = match (
```
bedeho: Can we break this into its own routine, its quite complex. I know it only has one caller right now, but its easier to understand `get_updated_inbound_rcs_delta` if there is an isolated call with a descriptive name for this step.

bedeho (changes requested):
Looks good to me. Nice work on picking a different type for the single entity vs. multi-entity side effects, this is exactly the sort of thing I have been asking for, well done.

I have only asked for minor cosmetic fixes.

runtime-modules/content-directory/src/lib.rs
iorveth: a) Done here https://github.com/Joystream/joystream/pull/715/commits/4272f54e4a93a23e0bf001a1744ecbbf09863a04
c) https://github.com/Joystream/joystream/pull/715/commits/103f0e61e0523918801db2d9f7253ed981d3db73

runtime-modules/content-directory/src/lib.rs
iorveth: Made an attempt to achieve this here https://github.com/Joystream/joystream/pull/715/commits/26ac2b2ef091c388a222c410e5f3a5575c652811
Though, do not sure about namings in `calculate_updated_inbound_rcs_delta`. Do you have a better idea?

bedeho: I think its fine for now, lets get this wrapped up.

bedeho (approved):
LGTM.

PR #722, opened 2020-06-17
Fix Pioneer and types - following buraucracy renaming and update to working-group module

mnaamani:
Following changes in https://github.com/Joystream/joystream/pull/672 
Fix types and pioneer accordingly.
Followed your lead @Lezek123 re: stories, I just dumped the elements.stories.tsx we can add stotries again in future if needed.


mnaamani:
Linter and pioneer build errors will be fixed after merging https://github.com/Joystream/joystream/pull/711

pioneer/packages/joy-media/src/Upload.tsx
```diff
@@ -248,8 +249,8 @@ class Component extends React.PureComponent<Props, State> {
 
     // TODO get corresponding data type id based on file content
     const dataObjectTypeId = new BN(1);
-
-    return [newContentId, dataObjectTypeId, new BN(file.size), ipfs_cid];
+    const { myMemberId } = useMyMembership();
```
Lezek123: `useMyMembership()` is a react hook, which won't work inside class component (only function components can use hooks). This causes an error that breaks the app when trying to upload. To fix this we'd have to change `withMembershipRequired` at the bottom of this file to `withOnlyMembers` (which passes additional props) and add `MyAccountProps` to the `Props` type. It would make `this.props.myMemberId` available inside the component.

pioneer/packages/joy-roles/src/elements.tsx
```diff
@@ -58,10 +54,10 @@ export function HandleView (props: ProfileProps) {
   );
 }
 
-type MemberProps = ActorProps & BalanceProps & ProfileProps
+type MemberProps = BalanceProps & ProfileProps
```
Lezek123: Those `MemoView`, `MemberView`, `WorkerDetailsView` and `BalanceView` componentes appear to be unused now that the related story has been deleted, so it's probably safe to just remove them here too (they seem outdated and not very useful for Nicaea anyway).

Lezek123 (changes requested):

pioneer/packages/joy-roles/src/elements.tsx
mnaamani: I agree I will remove them.

pioneer/packages/joy-media/src/Upload.tsx
mnaamani: awesome thanks for double checking this. I didn't actually get a chance to test it, but I had an inner voice telling me this looked wrong but wasn't sure why.

Will update as advised.

pioneer/packages/joy-roles/src/elements.tsx
mnaamani: done in 3e604de951d047289c0fe7c4d25aa5eda9c1e0f6

pioneer/packages/joy-media/src/Upload.tsx
mnaamani: done in bb3321b35842045b02e1fb631538e3c88e4f87f4

pioneer/packages/joy-media/src/Upload.tsx
```diff
@@ -248,8 +248,8 @@ class Component extends React.PureComponent<Props, State> {
 
     // TODO get corresponding data type id based on file content
     const dataObjectTypeId = new BN(1);
-
-    return [newContentId, dataObjectTypeId, new BN(file.size), ipfs_cid];
+    const { myMemberId } = this.props;
+    return [newContentId, myMemberId, dataObjectTypeId, new BN(file.size), ipfs_cid];
```
Lezek123: Actually, now I noticed the order of the parameters is incorrect too.
It should be: `return [myMemberId, newContentId (...)`

Lezek123 (changes requested):

pioneer/packages/joy-media/src/Upload.tsx
mnaamani: fixed in d5641f18260248352719db6b980fcc3dbcc44ede

Lezek123 (approved):
Should be good to go now

PR #728, opened 2020-06-18
Changes to the working group: make leader hireable.

### LINKED ISSUE #497: Improve working group module
Improvement list:
- a leader could be terminated as a regular worker
- a leader can change its parameters (reward account, increase stake)
- a leader can leave his/her position 
- remove leader worker id from the get_all_worker_ids() 
- add stake refunding to the working group

### LINKED ISSUE #500: Lead hireable in bureaucracy
EnsureOrigin gating, will be implemented as root origin council proposal logic in our runtime.

### LINKED ISSUE #611: Lead can update reward of worker
SSIA.

shamil-gadelshin:
The working group leader should be hirable like a regular worker. All working group operations should be applicable to the leader. Only the council is allowed to invoke operations with the leader.

### Changes
- hiring workflow
- roles operations
- stake management
- improve comments
- stake refunding
- move out content working group structs to the separate file in the runtime
- optional slashing on the role termination
- changed get_worker_ids() (excluded the leader and renamed)
- new extrinsic added (update_reward_amount)

### DIFF
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -244,6 +244,9 @@ decl_error! {
 
         /// Require signed origin in extrinsics.
         RequireSignedOrigin,
+
+        /// Working group size limit exceeded.
+        MaxActiveWorkerNumberExceeded,
     }
 }
 
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -17,6 +17,9 @@ decl_error! {
         /// There is leader already, cannot hire another one.
         CannotHireLeaderWhenLeaderExists,
 
+        /// Cannot fill opening with several applications.
+        CannotHireSeveralLeader,
+
         /// Not a lead account.
         IsNotLeadAccount,
 
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -14,6 +14,9 @@ decl_error! {
         /// Current lead is not set.
         CurrentLeadNotSet,
 
+        /// There is leader already, cannot hire another one.
+        CannotHireLeaderWhenLeaderExists,
+
         /// Not a lead account.
         IsNotLeadAccount,
 
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -241,6 +241,9 @@ decl_error! {
 
         /// Require root origin in extrinsics.
         RequireRootOrigin,
+
+        /// Require signed origin in extrinsics.
+        RequireSignedOrigin,
     }
 }
 
@@ -249,6 +252,7 @@ impl From<system::Error> for Error {
         match error {
             system::Error::Other(msg) => Error::Other(msg),
             system::Error::RequireRootOrigin => Error::RequireRootOrigin,
+            system::Error::RequireSignedOrigin => Error::RequireSignedOrigin,
             _ => Error::Other(error.into()),
         }
     }
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -17,8 +17,8 @@ decl_error! {
         /// There is leader already, cannot hire another one.
         CannotHireLeaderWhenLeaderExists,
 
-        /// Cannot fill opening with several applications.
-        CannotHireSeveralLeader,
+        /// Cannot fill opening with multiple applications.
+        CannotHireMultipleLeaders,
 
         /// Not a lead account.
         IsNotLeadAccount,
--- a/runtime-modules/working-group/src/lib.rs
+++ b/runtime-modules/working-group/src/lib.rs
@@ -56,8 +56,8 @@ use rstd::collections::btree_set::BTreeSet;
 use rstd::prelude::*;
 use rstd::vec::Vec;
 use sr_primitives::traits::{Bounded, One, Zero};
-use srml_support::traits::{Currency, ExistenceRequirement, WithdrawReasons};
-use srml_support::{decl_event, decl_module, decl_storage, ensure};
+use srml_support::traits::{Currency, ExistenceRequirement, Imbalance, WithdrawReasons};
+use srml_support::{decl_event, decl_module, decl_storage, ensure, print};
 use system::{ensure_root, ensure_signed};
 
 use crate::types::ExitInitiationOrigin;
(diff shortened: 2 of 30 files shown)


shamil-gadelshin:
I made a refactoring in the `runtime/lib.rs`. Moved out `content working group` structs to the separate file and plan to move all structs of the `versioned store` module.

`runtime/lib.rs` is 1K lines of code. It is a very complex and concept rich file.
Could we introduce a rule - "Don't place the integration code in the runtime/lib.rs"? @mnaamani @bedeho

runtime-modules/working-group/src/errors.rs
```diff
@@ -14,6 +14,12 @@ decl_error! {
         /// Current lead is not set.
         CurrentLeadNotSet,
 
+        /// There is leader already, cannot hire another one.
+        CannotHireLeaderWhenLeaderExists,
+
+        /// Cannot fill opening with several applications.
+        CannotHireSeveralLeader,
```
bedeho: CannotHireMultipleLeaders

bedeho (changes requested):
- A heroic effort on all these updates, not easy to review though. I think it would have been better to separate out the name changes in a separate PR from other substntial conceptual changes, any way you could separate the two now?

- I don't think the representation of the lead vs. worker is quite right, fragments such as this in `update_role_account` are a telltale sign

```Rust
            // Update role account
            WorkerById::<T, I>::mutate(worker_id, |worker| {
                worker.role_account = new_role_account_id.clone()
            });

            // Update lead data if it is necessary.
            let lead = <CurrentLead::<T, I>>::get();
            if let Some(lead) = lead {
                if lead.worker_id == worker_id {
                    let new_lead = Lead{
                        role_account_id: new_role_account_id.clone(),
                        ..lead
                    };
                    <CurrentLead::<T, I>>::put(new_lead);
                }
            }
```

We are basically having to go and keep two separate but duplicated state fields in synch.
Or the fact that when you check that a worker is the lead, you do not even need to consult whether `CurrentLead` is actually set.

Initially, I was going to say that we should drop `Lead` type and make `CurrentLead` just hold a member id, but now that I think of it, why do we even need to hold on to any internal representation of who the current lead is in the module at all? We could just drop both? Let me know if I am lost at sea here, if not, lets do this change before I review any further.

runtime-modules/working-group/src/errors.rs
shamil-gadelshin: Ok

shamil-gadelshin:
I tried to have ID cache for the leader, but I agree with you that it creates more problems than solves. I change it to the worker_id.

runtime-modules/working-group/src/lib.rs
```diff
@@ -309,6 +311,11 @@ decl_storage! {
 
         /// Worker exit rationale text length limits.
         pub WorkerExitRationaleText get(worker_exit_rationale_text) : InputValidationLengthConstraint;
+
+        /// Map member id by hiring application id.
+        /// Required by StakingEventsHandler callback call to refund the balance on unstaking.
+        pub MemberIdByHiringApplicationId get(fn member_id_by_hiring_application_id):
```
bedeho, line 323: Why is this needed? It was not needed in the content working group?

The only place I see it used is in `refund_working_group_stake`, in order to do this

```
let member_id = Module::<T, I>::member_id_by_hiring_application_id(hiring_application_id);
```

But you could already resolve the member id from an application by just looking at `Application::member_id`?

runtime-modules/working-group/src/lib.rs
```diff
@@ -1259,6 +1187,101 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+        }
+
+        <NegativeImbalance<T>>::zero()
+    }
+
+    /// Returns all existing worker id list excluding the current leader worker id.
+    pub fn get_regular_worker_ids() -> Vec<WorkerId<T>> {
+        let lead_worker_id = Self::current_lead();
+
+        <WorkerById<T, I>>::enumerate()
```
bedeho, line 1267: Do we have some limit somewhere which limits how large this map?

runtime-modules/working-group/src/lib.rs
```diff
@@ -1259,6 +1187,101 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+        } else {
+            print("Working group broken invariant: no member profile.");
+            return imbalance;
+        }
+
+        <NegativeImbalance<T>>::zero()
+    }
+
+    /// Returns all existing worker id list excluding the current leader worker id.
+    pub fn get_regular_worker_ids() -> Vec<WorkerId<T>> {
```
bedeho, line 1264: You could make this slightly more flexible, and have it be `get_worker_ids(include_lead: bool)`.

bedeho (changes requested):
Looks good!

I would say that this still was too much for one PR, quite a few changes did seem orthogonal and could have been delivered separately.

runtime-modules/working-group/src/lib.rs
shamil-gadelshin: I prefer to leave it as it is because the only usage now if for assigning random storage provider. This random provider id is picked using only worker ids. Within the context of the picking storage provider to add 'include_lead' flag looks confusing for me. If you think this is a good idea to have such a feature - I'd rather add separate method - `gel_all_worker_ids()`.

runtime-modules/working-group/src/lib.rs
shamil-gadelshin: Not yet, but we remove old workers on leaving or termination, so this collection should not be too long.

runtime-modules/working-group/src/lib.rs
shamil-gadelshin: The similar functionality in the content working group is here:

`pub UnstakerByStakeId get(unstaker_by_stake_id) config(): linked_map StakeId<T> => WorkingGroupUnstaker<LeadId<T>, CuratorId<T>>;`

But we have no `stake_id` in the working group. It is hidden inside the hiring module.

The method like this can be found in the runtime crate implemented within `ContentWorkingGroupStakingEventHandler`.

runtime-modules/working-group/src/lib.rs
bedeho: ok

runtime-modules/working-group/src/lib.rs
bedeho: It's ultimately under the control of the lead, so they may choose to make it arbitrary long, which would be a problem.

So long as we are doing some enumeration, I think we will then need to have a hardcoded trait parameter which limits how many workers can be added to the map. I suggest such a check should be added to `fill_opening`.

runtime-modules/working-group/src/lib.rs
bedeho: I think I understand.

bedeho (approved):
LGTM, nice job.

Issue #741, opened 2020-06-19
Pioneer: List of the "Working groups" tab issues


Lezek123:
There are a few old UI issues related to the "Working groups" tab that may be worth tacking before Nicaea release: 

- [x] A neverending loading screen when no opportunities are available. - **_FIXED_**
- [x] There seems to be a wrong assumption inside `/joy-roles/src/flows/apply.tsx` that the concept of _Minimum combined stake_, which has to do with `max_active_applicants`, only regards openings that have both role and application stake set to `AtLeast` (which is not true, because the situation is very similar ie. when only one stake set to `AtLeast`) -  **_FIXED_**
- Kind of related to the issue above: If there is a maximum number of considered applicants (set via `application_rationing_policy.max_active_applicants`), user should be prevented from trying to apply right in the first step (where he specifies the stakes). Currently the flow allows the user to go all the way to step 3, even though it's already "known" that the extrinsic will fail. The failing extrinsic doesn't give any information about the reason unless the user is using his/her own node with logging enabled. _**EDIT: As preventing it would be more complicated with the current flow I settled on displaying a very clear error message in case user will (most likely) not qualify to be considered for an application**_ 
- [X] _Minimum required stake_ in the first step of applying process is displayed as `0` for the first candidate that is applying to an opening that has both stakes set to `AtLeast`. - **_FIXED_**
- Sometimes if I get back a few steps when applying for an opening and then try to go forward and submit the application with `[Make transaction and submit application]` button - the app breaks. The error says something about not beeing able to find the key. I wasn't able to reproduce it purposefully though. - _**Couldn't reproduce for quite so time** (probably not an issue anymore)_
- [X] The way role keys are named during the application can result in creating multiple keys/account with the same name, which causes duplicate-key React console error and may possibly have some unexpected consequences (possibly related to the issue above) - **_FIXED_**

Lower priority:
- [X] There is a `Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. You can only call Hooks at the top level of your React function.` error when entering `/working-groups/admin` (probably related to the invalid use of `useMyAccount` under `AdminView` at `/joy-roles/src/tabs/Admin.controller.tsx`)
- Applying to an opening right after it is created via `/working-groups/admin` tool breaks the app (most likely it's only a problem in case user is a lead and tries to apply to his own openings. Refreshing the page is enough to fix the issue in that case anyway) - _**Edit: probably not an issue anymore**_
[...]

Issue #618, opened 2020-06-08
Pioneer: Manage storage working group applications (applicant)


Lezek123:
Since all steps of the application process itself will remain within Pioneer, there will be a need to upgrade the list of user's applications at  `working-groups/my-roles` to include those from the `storageBureaucracy`/`storageWorkingGroup` module and add support for withdrawing those too. Optionally: filtering applications by role.

Lezek123:
Done in https://github.com/Joystream/joystream/pull/742 (along with the general `My roles` adjustment)

Issue #597, opened 2020-06-08
Pioneer: Storage working group applying process


Lezek123:
Update the working group applying process in order for it to work with both content working group and storage working group (using correct api endpoints, types etc. for given group).

Lezek123:
Done in https://github.com/Joystream/joystream/pull/742

Issue #726, opened 2020-06-17
Lead openings and applications - adjusting CLI and Pioneer


Lezek123:
Some adjustments are probably going to be needed in Pioneer and the CLI w.r.t. the introduction of _lead openings_, which will most likely differ from standard _worker openings_ only by a flag value (like `opening_type: Worker / Lead`).

In Pioneer the proper distinction may include:
- Displaying separate message regarding available openings for _lead_ and _workers_ in the overview at `/working-groups` index page
- Separate label, filter (in select dropdown) and routes for _lead openings_ (ie. `/working-groups/opportunities/storageProviders` and `/working-groups/opportunities/storageProvidersLead`)
- Perhaps some minor adjustments in the applying process and `/working-groups/my-roles`, but probably not too many of those will be needed.

As it is not yet obvious what mechanisms will be implemented on the runtime side to allow this distinction and whether there will be a need to make it more strict or not, I'm adding an initial estimation which may be subject to change.

Lezek123:
Implemented in the compund PRs that were open at the time:
- Pioneer: https://github.com/Joystream/joystream/pull/742
- CLI - https://github.com/Joystream/joystream/pull/908

PR #742, opened 2020-06-19
Pioneer: Applying for storage working group opening and final "Working groups" tab issues

Lezek123:
Implements: https://github.com/Joystream/joystream/issues/597

Note that the listing of user's active applications for storage working group etc. is not implemented yet, as there is another issue for that (https://github.com/Joystream/joystream/issues/618)

**ts-ignore issue**
While working on this issue I removed the `ts-ignore` inside `/joy-roles/src/flows/apply.controller.tsx` as it is a very bad practice to use it, because it silences a lot of potential TypeScript errors during the build. It should have actually been caught by eslinter's `ban-ts-ignore` rule, but for some reason it wasn't (perhaps because it wasn't added as a valid-syntax comment, but actually a string value inside jsx).

This forced me to fix a few typing errors inside `/joy-roles/src/flows/apply.tsx` that were revealed once the `ts-ignore` was removed.

**Other pre-existing UI issues**
I also decided to list all the pre-existing `Working Group` tab issues I was able to find in https://github.com/Joystream/joystream/issues/741 to allow better understanding of the problems that may occur during testing, but were not actually introduced in this PR (I'll try to tackle those once the `Working Groups` tab is fully adjusted to support Storage Working Group)


Lezek123:
Updated to include https://github.com/Joystream/joystream/issues/618 and some of the fixes from https://github.com/Joystream/joystream/issues/741:

`working-groups/my-roles` page should now be fully adjusted to work with Storage Working Group. Although in issue https://github.com/Joystream/joystream/issues/618 it wasn't planned to make it fully compatible (active roles were not considered), I decided to go for it, since it was very easy to implement at this point and makes the Pioneer's support for Storage Working Group _complete_.

Issues described in https://github.com/Joystream/joystream/issues/741:
- Fixed the issue with neverending loading screen in case there are no openings available.
- Fixed the applying flow for openings with variable stakes and max. number of applicants specified. User should now be able to correcty determine the minimum stake he needs to specify to be considered when applying for such openings, as well as his position in the "applicants hierarchy" during the first step of the application. Note that if both stakes are set to `AtLeast`, a different component handles this step than when only one of them is set to `AtLeast`.
- Fixed role key name generation issue (keys with duplicate names should not be generated anymore)

pioneer/packages/joy-roles/src/tabs/MyRoles.tsx
```diff
@@ -102,9 +105,10 @@ function RoleName (props: NameAndURL) {
 }
 
 export interface ActiveRole extends NameAndURL {
-  curatorId: CuratorId;
+  workerId: CuratorId;
```
mnaamani: Maybe start to transition to the equivalent types from working-group ?

pioneer/packages/joy-roles/src/transport.substrate.ts
```diff
@@ -727,13 +760,13 @@ export class Transport extends TransportBase implements ITransport {
...
-          const tx = this.api.tx.contentWorkingGroup.applyOnCuratorOpening(
-            membershipIds[0],
-            new u32(id),
-            new GenericAccountId(roleAccount as string),
-            roleStake.eq(Zero) ? null : roleStake,
-            appStake.eq(Zero) ? null : appStake,
-            applicationText
+          const tx = this.apiExtrinsicByGroup(group, 'applyOnOpening')(
+            membershipIds[0], // Member id
+            new u32(id), // Worker/Curator opening id
```
mnaamani: opening id in the runtime is actually a u64. I think just passing a number is fine, the type will be converted to the correct type expected by the transaction. I realize this code is from prior iterations but I just spotted it now. Using the wrong type will affect the remaining values in the encoded transaction. Its probably best to use `WorkerId` here.

mnaamani: maybe we should check other places where this mistake might have been made.

mnaamani (changes requested):
I left a comment about using correct type in one of the tx calls.

Tested the UI. Only thing I noticed that is incorrect (perhaps its existing behavior) but when appliccations are withdrawn, the total number of applications on an opening still counts the withdrawn applications.

Otherwise all works as expected.

pioneer/packages/joy-roles/src/transport.substrate.ts
Lezek123: Fixed. There were some occurences like that in `Admin.controller.tsx` too. Those weren't causing any trouble porobably because `@polkadot/api` knows how to convert any input according to the api metadata (unless the id is > 4294967295 in which case the`new u32()` would be a problem), but it's good to keep those consistent.

pioneer/packages/joy-roles/src/tabs/MyRoles.tsx
Lezek123: I overlooked that one, fixed.

Lezek123:
Added fixes according to the comments.

I noticed that there was a `ts-nocheck` clause inside `Admin.controller.tsx` which silenced a lot of useful TypeScript errors. I fixed those and it seems to have simultaneously fixed a few UI issues with the opening creation form at `working-groups/admin` (ie. there were some problems with changing the stake type between `AtLeast` and `Exact`, changing stake values etc. and they seem to be gone now).

I added `ban-ts-comment` rule to eslint to possibly avoid running into this problem again. There is one more occurrence of `ts-ignore` in `/packages/react-components/src/AddressCard.tsx` (which was purposefully silenced, so I just updated the `eslint-disable-next-line` clause there) and 2 `ts-nocheck` occurrences in storybook-related files inside `joy-roles` (which I silenced for now with `eslint-disable-next-line`, but marked with `TODO: FIXME:`)

Lezek123:
> Only thing I noticed that is incorrect (perhaps its existing behavior) but when appliccations are withdrawn, the total number of applications on an opening still counts the withdrawn applications.

I don't think I was chaning anything there, it looks like this number just represents total number of applications regardless of their current status (which can be `Active`, `Unstaking` or `Inactive`). It is also displayed for already filled openings etc., so it makes some sense, but I understand that pairing it with `max_active_applications` can be confusing and indicate that this number means _active_ applications.

Lezek123:
I actually found another issue related to those non-active applications.

Because the status of them wasn't checked inside `myApplicationRank` and `openingApplicationRanks` methods, it caused unreliable rank perdiction and incorrect `minStake` value when applying in case one of the previous applicants has withdrawn its application.

Should be fixed now.

mnaamani:
I created an opening with an empty `{}` json for the human readable text so when I visit the `my-roles` page it crashes with the following error (I presume its due to not finding the expected properties on the parsed object):

Screen Shot 2020-07-07 at 10 25 42 AM

Maybe we should handle badly formatted human readable text.

Lezek123:
Updated with some UI adjustments following "Make lead hireable" changes (https://github.com/Joystream/joystream/issues/726) that I already had on my branch locally.

Also merged `nicaea` to fix conflicts and added a method to parse `human_readable_text` with fallback to valid `GenericJoyStreamRoleSchema` object that has all required values set to `Unknown` in case the actual json is invalid.

pioneer/.eslintrc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     '@typescript-eslint/camelcase': 'off',
     'react/prop-types': 'off',
     'new-cap': 'off',
-    '@typescript-eslint/interface-name-prefix': 'off'
+    '@typescript-eslint/interface-name-prefix': 'off',
+    '@typescript-eslint/ban-ts-comment': 'error'
```
mnaamani, line 19: This is a good rule to have:

` '@typescript-eslint/ban-ts-comment': 'error'`

might add it to #891 cc: @Gamaranto @kdembler

pioneer/packages/joy-roles/src/flows/apply.elements.stories.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO: FIXME: Remove the ts-nocheck and fix errors!
```
mnaamani, line 1: It better to create github issues rather than TODO's in the code. We can label them with `good first issue` for contributors to find something simple and easy to work on.

mnaamani (approved):

pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
```diff
@@ -553,7 +399,9 @@ export class AdminController extends Controller<State, ITransport> {
   }
 
   startAcceptingApplications (creatorAddress: string, id = 0) {
-    const tx = this.api.tx.contentWorkingGroup.acceptCuratorApplications(new u32(id));
+    const tx = this.api.tx.contentWorkingGroup.acceptCuratorApplications(id);
+    // FIXME: That's a bad way to send extrinsic in Pioneer (without "queueExtrinsic" etc.)
+    // and probably the reason why it always appears as succesful
```
mnaamani, line 404: I'm sure here you mean that you are not handling case when the transaction returns an error, but is still finalized?

mnaamani:
Works as expected.
One enhancement that I would suggest is in the my-roles, to show in the current roles, the worker id.
This is needed by the storage providers when running their node.

pioneer/.eslintrc.js
kdembler: That's a good one in general but I would use the option `allow-with-description` so we can use `ts-ignore` in specific cases. Recently I've worked with a library with stale typings and had to ignore typescript error to use some newer option.

PR #688, opened 2020-06-12
WIP: generate Warthog model files from templates

dzhelezov:
This is a WIP PR (built on top of https://github.com/Joystream/joystream/pull/641) for https://github.com/Joystream/joystream/issues/673
All the key features are already available.

TODOs:
- [x]  Replace hardcoded paths and namings with calls to warthog's Config
- [x]  Refactor: unify FTS and Model files rendering, extract interfaces, separate rendering and path resolution
- [x] Test: Cover with unittests key entity relations and types

Issue #609, opened 2020-06-08
Add support for enumerated types


bedeho:
SSIA.

dzhelezov:
This is the issue to be fixed: https://github.com/goldcaddy77/warthog/issues/372

bedeho:
Done.

PR #744, opened 2020-06-19
Enum type support for query node schemas

dzhelezov:
This PR includes
- Initial schema support for enum types (templates, model support) (https://github.com/Joystream/joystream/issues/609). 
- Basic tests for enums
- Major refactoring of https://github.com/Joystream/joystream/pull/688

Due to the upcoming major changes in the way we handle non-primitive types (enum included) the current implementation will likely be amended in the future, but the core structure will probably remain the same.


metmirr (approved):
LGTM!

PR #746, opened 2020-06-20
Update storage node

mnaamani:
## Scope of changes

- Initial fixes and refactoring to work with nicaea.
- removed all code/commands related to registering/staking in deprecated roles runtime module
- Leaving all code in vanilla javascript, doing some js-standard linter fixes.
- Added some helper commands in the storage-cli to setup develoment environment.
- Got rid of need for a configuration file, all options are passed as command arguments
- Added `--provider-id [number]` argument to identify 
- Added `--dev` option when running colossus server to run with development settings
- Fixed some bugs in discovery api when provider not found
- Updated `helios`

## Not Done
- Fix broken tests (some need to be moved into network-integration tests - not in scope for nicaea but if we have time
  would be nice to have
- Fix upload, download, head commands in storage-cli
- applying auto fix eslinter to codebase

## Howto test

```sh
# Build `joystream/node:latest` docker image locally (drink some coffee/tea)
./scripts/build-joystream-node-docker-image.sh

# build dependencies
yarn

# runs a development chain and ipfs node
# then execute scripts to setup the chain with developer settings
# and finally runs Colossus in developer mode
./storage-node/scripts/run-dev-instance.sh
```

Wait a couple of minutes for the ipns identity of the node to be published:
```sh
  discovery::publish publishing... +42ms
  ipfs-http-client:request POST localhost:5001/api/v0/name/publish?key=self&resolve=false&arg=QmcSjtVMfDSSNYCxNAb9PxNpEigCw7h1UZ77gip3ghfbnA&stream-channels=true 200 OK +3m
  discovery::publish {
  discovery::publish   name: 'QmV2cNp9uTVGabAhuip1yZt697DPy1kivvgBLZvNUaX4KT',
  discovery::publish   value: '/ipfs/QmcSjtVMfDSSNYCxNAb9PxNpEigCw7h1UZ77gip3ghfbnA'
  discovery::publish } +3m
  joystream:runtime:base Got nonce for 5FNPVC9sBT4ns61t6WTCvcsHm438YkZ2o96NixNZ2P1J4vhj from chain: 0 +3m
  joystream:runtime:base Signing and sending tx +0ms
  joystream:runtime:base TX status: Ready +48ms
  joystream:runtime:base TX Ready. +0ms
  joystream:runtime:base TX status: Finalized +5s
  joystream:runtime:base TX Finalized. +1ms
  joystream:colossus publishing complete, scheduling next update +3m
```
...now you should be able to upload  videos. 

```sh
# in another terminal, run pioneer
yarn workspace pioneer start
```

Browse pioneer on http://localhost:3000/
You should find Alice account is the storage working group lead and is a storage provider

Create a media channel. And upload a file.

Only the upload works.. I haven't initialized the content-directory schemas so after the upload it will give an error that `MediaObject entity class is undefined`

Screen Shot 2020-06-25 at 5 41 28 PM

You can verify the asset was uploaded by inspecting chain state or running helios:

`yarn helios`

to stop the server just `Ctrl-C` (this just kills colossus)
then script below to kill the docker containers.

```sh
./storage-node/scripts/stop-dev-instance.sh
```

If you don't have docker you can just compile the joystream-node as usual, and run an ipfs daemon

```sh
ipfs daemon
```

```sh
cargo build --release
cargo run --release -- --dev
```

```sh
# after starting the ipfs daemon and chain run scripts below to init and start the dev storage-node
yarn storage-cli dev-init
yarn colossus server --dev
```


storage-node/packages/cli/bin/dev.js
```diff
@@ -0,0 +1,165 @@
...
+  const filledMap = await api.signAndSendThenGetEventResult(alice, fillTx, {
+    eventModule: 'storageWorkingGroup',
+    eventName: 'WorkerOpeningFilled',
+    eventProperty: 'WorkerApplicationIdToWorkerIdMap'
+  })
+
+  if (filledMap.size !== 1) {
+    throw new Error('openening was not filled as expected')
+  }
+
```
shamil-gadelshin, line 116: You were going to fix it, weren't you?

shamil-gadelshin:
Running: yarn helios

> ....
> Data Directory has 1 assets
> 5GEFD7ThJAaquzfXzzSpwRHF85j5972KfcKpd3qo6WkprwET replication 0/1 - Accepted

Why replication 0/1? Is it OK?

mnaamani:
Yes I spotted that when demoing it to you yesterday ;) but didn’t get
chance to investigate yet.

storage-node/packages/cli/bin/dev.js
```diff
@@ -0,0 +1,165 @@
+/* eslint-disable no-console */
+
+'use strict'
+
+const debug = require('debug')('joystream:storage-cli:dev')
+const assert = require('assert')
```
shamil-gadelshin, line 6: Do we need this (assert)? Couldn't find the usage.

storage-node/packages/cli/bin/dev.js
```diff
@@ -0,0 +1,165 @@
+/* eslint-disable no-console */
```
shamil-gadelshin, line 1: Why do we mix console.log and debug? It's unclear when to use which, especially if include cli.js in the analysis.
Probably won't be necessary this 'disable lint' statement.

storage-node/packages/cli/bin/dev.js
```diff
@@ -0,0 +1,165 @@
...
+  console.log('Running setup')
+
+  // set localhost colossus as discovery provider
+  // assuming pioneer dev server is running on port 3000 we should run
+  // the storage dev server on a different port than the default for colossus which is also
+  // 3000
+  debug('Setting Local development node as bootstrap endpoint')
+  await api.discovery.setBootstrapEndpoints(alice, [`http://localhost:${developmentPort()}/`])
+
+  debug('Transfering tokens to storage role account')
```
shamil-gadelshin: Typo 'Transfering' -> 'Transferring'

storage-node/packages/runtime-api/workers.js
```diff
@@ -0,0 +1,119 @@
...
+    for (let i = 0; i < ids.length; i++) {
+      const id = ids[i]
+      if (providers[id].role_account.eq(roleAccount)) {
+        return id
+      }
+    }
+
+    return null
+  }
+
```
shamil-gadelshin, line 92: We have a working_group function (get_all_worker_ids) for that.
I can introduce [the custom rpc](https://substrate.dev/recipes/3-entrees/custom-rpc.html) for this call.

storage-node/packages/cli/bin/dev.js
```diff
@@ -0,0 +1,165 @@
...
+      handle: 'alice'
+    })
+  } else {
+    debug('Alice is already a member')
+  }
+
+  // Make alice the storage lead
+  debug('Setting Alice as Lead')
+  // prepare set storage lead tx
+  const setLeadTx = api.api.tx.storageWorkingGroup.setLead(aliceMemberId, alice)
```
shamil-gadelshin: You are using WorkerApi in this file. Could we wrap such calls also to be consistent?

storage-node/packages/colossus/bin/cli.js
```diff
@@ -303,95 +228,75 @@ async function announce_public_url(api, config) {
...
+  } else {
+    throw new Error(`Command '${command}' not recognized, aborting!`)
   }
 }
 
 main()
   .then(() => {
-    console.log('Process exiting gracefully.');
-    process.exit(0);
+    console.log('Process exiting gracefully.')
```
shamil-gadelshin: self-congratulation? 😃

shamil-gadelshin:
Do we include the '.eslintrc.js' rules alteration in the scope of the PR?

shamil-gadelshin (commented):
We should create an issue about fixing integration tests.

storage-node/packages/cli/bin/dev.js
mnaamani: I was using it to assert later that filledMap.size was equal to one, then replaced it with a check and throwing, so forgotr to remove that import. Honestly I had my linter turned off while coding because of a huge number of warnings I could hardly see the code :)

Perhaps its a good idea now that I have working code to turn the linter back on and fix all the issues?

storage-node/packages/cli/bin/dev.js
mnaamani: Well for a cli, its generally useful to give user some feedback on progress but not overwhelm them. My strategy was to give minimum information in console.log and verbose output using debug.

Enabling this es-lint rule is really suitable more for a web app where its discouraged to log to the console and it was on by default. For a cli its very hard to get away from outputting to the console :)

storage-node/packages/cli/bin/dev.js
mnaamani: I see your point. I didn't consider that because I felt that these calls aren't really needed by the storage node itself, but were mainly for this 'integration/development' setup. Will move it so when we consolidate the disparate api layers from the different apps we have it will in the right place.

storage-node/packages/colossus/bin/cli.js
mnaamani: lol, I don't think I added that, I just removed the process.exit call (while I had the linter active) But I agree its silly to have.

storage-node/packages/runtime-api/workers.js
mnaamani: Thats an interesting strategy. We have casually considered in the past to add custom rpc calls to query state, especially complex queries. However since you could have an older version of a node working fine with later runtimes, one doesn't always need to run the latest version of a node, so the availability of the rpc call isn't guaranteed to the user. Does that make it less reliable in a way. But its definitely worth exploring.

> Do we include the '.eslintrc.js' rules alteration in the scope of the PR?

I was originally thinking of making both formatting and esliting fixes in a separate PR, so would include modifying the rules also. Maybe worth doing it in this PR now that you mention it.

shamil-gadelshin: I can see your point of preserving older nodes in the working state. Probably we could release such 'custom rpc optimizations' in a bunch during the node upgrade, but it would also make the system more fragile because we would need to do more frequent node upgrades than necessary. Ok, then we leave it until we really need it in order to avoid some time & resources consuming iteration through the storage node.

storage-node/packages/helios/bin/cli.js
```diff
@@ -1,166 +1,185 @@
...
       let info = await axios.head(assetUrl)
       content[encodeAddress(contentIds[i])] = {
         type: info.headers['content-type'],
         bytes: info.headers['content-length']
       }
+      // TODO: cross check against dataobject size
       found++
-    } catch(err) { console.log(`${assetUrl} ${err.message}`); continue; }
+    } catch (err) {
+      // console.log(`${assetUrl} ${err.message}`)
```
shamil-gadelshin: Leftover debug comment.

storage-node/packages/helios/bin/cli.js
```diff
@@ -1,166 +1,185 @@
...
+      missing++
+      continue
```
shamil-gadelshin: continue is the last step in the loop anyway

shamil-gadelshin (changes requested):
.

shamil-gadelshin (changes requested):
Nice work with a lot of effort. Also wonderful dev-setup! 

As a general idea: I noticed the inconsistent comment style, sometimes we have comments on the methods sometimes we don't. Could we use the same style (have comments preferably 😄) at least on the new methods or heavily altered?

mnaamani:
> Running: yarn helios
> 
> > ....
> > Data Directory has 1 assets
> > 5GEFD7ThJAaquzfXzzSpwRHF85j5972KfcKpd3qo6WkprwET replication 0/1 - Accepted
> 
> Why replication 0/1? Is it OK?

I double checked the code in helios. I think you probably ran helios right after you uploaded the file. Perhaps the node hadn't yet sent the transactions to update the relationship to ready state.

storage-node/packages/cli/bin/dev.js
mnaamani: fixed in d6905d6735b4822a844b612dded2cf55d793acca

storage-node/packages/cli/bin/dev.js
mnaamani: it is actually used in `assert_file()`

shamil-gadelshin (approved):
LGTM.

Good job with adding comments and updating tests!

PR #750, opened 2020-06-22
Remove import of members and old forum posts

mnaamani:
Its cleaner to start a new chain without members or forum posts, at least for development its not required. And the member imports were initially done when we launched a new testnet.

I'm not sure if there is an assumption that Alice is already a member in the network integration tests @gleb-urvanov (there is a member on testnet who chose to use this well known key which is odd)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1569,7 +1569,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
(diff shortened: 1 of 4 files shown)


shamil-gadelshin (approved):
LGTM: Yes, let's remove unnecessary data from the new chains.

PR #764, opened 2020-06-23
Query node relations

metmirr:
1. The field name is not overwritten
2. A derived field can not reference another derived field
3. For a given type `type A { rel1: B }`
    1. If `B` has a field of type `A`
        1. If the field is a list then `@derivedFrom` is required and the relationship is OneToMany
        2. If the field is not a list and no `@derivedFrom` directive then it is OneToOne relationship
    2. If field was not found on `B` then OneToMany relationship
4. For a given type `type A { rel1: [B] }`
    1. If `B` has a field of type `A`
        1. If the field is not a list then error incorrect relationship detected
        2. If the field is a list then must have `@derivedFrom` directive otherwise an error
5. For a given type `type A { rel1: [B] @derivedFrom(field: “relA”)}`
    1. Look for `relA` field on `B` if not found an error
    2. If `relA` is a list then the relationship is Many2Many otherwise OneToMany
-------
# GraphQL Entity Relationships

### One-To-One (1:1) Relationships

In One-To-One relation, one entity instance is related to only one instance of another entity. One side of the relationship should always derive.

```graphql
type User @entity {
  name: String!
  profile: Profile! @derivedFrom(field: "user")
}

type Profile @entity {
  avatar: String!
  user: User!
}
```

Database tables:

```
          user
| Column  | Type
----------|-------
| id      | character varying
| name    | character varying
```

```
          profile
| Column  | Type
----------|-------
| id      | character varying
| avatar  | character varying
| userId  | character varying FOREIGN KEY UNIQUE CONSTRAINT
```

### One-To-Many (1:n) Relationships

In One-To-Many relation, one entity instance is related to multiple instances of the other entity.

```graphql
type User @entity {
  name: String
}

type Post @entity {
  title: String
  author: User!
}
```

Database table for the `Post` entity:

```
          post
| Column  | Type
----------|-------
| id      | character varying
| avatar  | character varying
| authorId  | character varying FOREIGN KEY
```

The only difference between `1:1` and `1:n` is the unique constraint that `1:1` has.

### Many-To-Many (n:n) Relationships

Many-To-Many is a relationship where one entity instance is related to many instances of other entity and vice-versa. In this relationship, one side of the relation must derive.

```graphql
type User @entity {
  name: String
  books: [Book!] @derivedFrom(field: "authors")
}

type Book @entity {
  title: String
  authors: [User!]
}
```

A junction table is created for n:n relationship.

Database tables:

```
          book
| Column  | Type
----------|-------
| id      | character varying
| title   | character varying
```

```
          book_user
| Column  | Type
----------|-------
| book_id | character varying
| user_id | character varying
```

### Reverse Lookups

Defining reverse lookups on an entity allows you to query other side of the relation. Use `@derivedFrom` directive to add reverse lookup to an entity.

**Example**
If we want to access a user's `posts` from the user entity we should add a derived field to `User` entity:

```graphql
type User @entity {
  posts: [Post!] @derivedField(field: "author")
}

type Post @entity {
  author: User!
}
```

## Relationships In Mappings

Each GraphQL entity has a corresponding typeorm entity and we use these entities to perform CRUD operations.

**Example**

We will create a new post for an existing user:

```ts
export async function handleNewPost(db: DB, event: SubstrateEvent) {
  const { userId, title } = event.params;
  const user = await db.get(User, { where: { id: userId } });

  const newPost = new Post();
  newPost.title = title;
  newPost.author = user;

  db.save<Post>(newPost);
}
```

### DIFF
--- a/query-node/substrate-query-framework/cli/src/generate/RelationshipGenerator.ts
+++ b/query-node/substrate-query-framework/cli/src/generate/RelationshipGenerator.ts
@@ -1,4 +1,5 @@
 import { WarthogModel, Field, ObjectType } from '../model';
+import { generateJoinColumnName, generateJoinTableName } from './utils';
 
 export class RelationshipGenerator {
   visited: string[];
@@ -10,8 +11,21 @@ export class RelationshipGenerator {
   }
 
   addMany2Many(field: Field, relatedField: Field, objType: ObjectType, relatedObject: ObjectType): void {
-    field.relation = { type: 'mtm', columnType: field.type, joinTable: true, relatedTsProp: relatedField.name };
-    relatedField.relation = { type: 'mtm', columnType: relatedField.type, relatedTsProp: field.name };
+    field.relation = {
+      type: 'mtm',
+      columnType: field.type,
+      joinTable: {
+        tableName: generateJoinTableName(objType.name, relatedObject.name),
+        joinColumn: generateJoinColumnName(objType.name),
+        inverseJoinColumn: generateJoinColumnName(relatedObject.name),
+      },
+      relatedTsProp: relatedField.name,
+    };
+    relatedField.relation = {
+      type: 'mtm',
+      columnType: relatedField.type,
+      relatedTsProp: field.name,
+    };
 
     objType.relatedEntityImports.add(relatedObject.name);
     relatedObject.relatedEntityImports.add(objType.name);
@@ -20,8 +34,16 @@ export class RelationshipGenerator {
   }
 
   addOne2Many(field: Field, relatedField: Field, objType: ObjectType, relatedObject: ObjectType): void {
-    field.relation = { type: 'otm', columnType: field.type, relatedTsProp: relatedField.name };
-    relatedField.relation = { type: 'mto', columnType: relatedField.type, relatedTsProp: field.name };
+    field.relation = {
+      type: 'otm',
+      columnType: field.type,
+      relatedTsProp: relatedField.name,
+    };
+    relatedField.relation = {
+      type: 'mto',
+      columnType: relatedField.type,
+      relatedTsProp: field.name,
+    };
 
     objType.relatedEntityImports.add(field.type);
     relatedObject.relatedEntityImports.add(objType.name);
@@ -36,8 +58,17 @@ export class RelationshipGenerator {
   }
 
   addOne2One(field: Field, relatedField: Field, objType: ObjectType, relatedObject: ObjectType): void {
-    field.relation = { type: 'oto', columnType: field.type, joinColumn: true, relatedTsProp: relatedField.name };
-    relatedField.relation = { type: 'oto', columnType: relatedField.type, relatedTsProp: field.name };
+    field.relation = {
+      type: 'oto',
+      columnType: field.type,
+      joinColumn: true,
+      relatedTsProp: relatedField.name,
+    };
+    relatedField.relation = {
+      type: 'oto',
+      columnType: relatedField.type,
+      relatedTsProp: field.name,
+    };
 
     objType.relatedEntityImports.add(relatedObject.name);
     relatedObject.relatedEntityImports.add(objType.name);
@@ -85,7 +116,8 @@ export class RelationshipGenerator {
             const derivedFields = relatedFields.filter(f => f.derivedFrom?.argument === field.name);
             if (derivedFields.length === 0) {
               throw new Error(
-                `Incorrect one to one relationship. '${relatedObject.name}' should have a derived field with @derivedFrom(field: "${field.name}") directive`
+                `Incorrect one to one relationship. '${relatedObject.name}' should have a derived field 
+                with @derivedFrom(field: "${field.name}") directive`
               );
             } else if (derivedFields.length === 1) {
               if (!derivedFields[0].isList) {
--- a/query-node/substrate-query-framework/cli/src/generate/RelationshipGenerator.ts
+++ b/query-node/substrate-query-framework/cli/src/generate/RelationshipGenerator.ts
@@ -1,5 +1,6 @@
 import { WarthogModel, Field, ObjectType } from '../model';
 import { generateJoinColumnName, generateJoinTableName } from './utils';
+import { camelCase } from 'lodash';
 
 export class RelationshipGenerator {
   visited: string[];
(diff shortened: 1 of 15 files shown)


dzhelezov:
> 3. For a given type `type A { rel1: B }`
>    
>    1. If `B` has a field of type `A`
>       
>       2. If the field is not a list and no `@derivedFrom` directive then it is OneToOne relationship

What if `B` has two or more fields of type `A` without `@derivedFrom` ?

metmirr:
> > 1. For a given type `type A { rel1: B }`
> >    
> >    1. If `B` has a field of type `A`
> >       
> >       1. If the field is not a list and no `@derivedFrom` directive then it is OneToOne relationship
> 
> What if `B` has two or more fields of type `A` without `@derivedFrom` ?

If multiple fields found then between those fields I search for a match that could be `OneToOne` `const relatedField = relatedFields.find(f => !f.isList)` if no result then I throw an error `Incorrect relationship detected!`

dzhelezov:
> If multiple fields found then between those fields I search for a match that could be `OneToOne` `const relatedField = relatedFields.find(f => !f.isList)` if no result then I throw an error.

Hm, I think `find(...)` simply picks up the first element for which the predicate is true. So the way `relatedField` is chosen is not easily predictable. Can we enforce a more explicit mapping here?

metmirr:
UPDATE
@dzhelezov suggested enforcing one-to-one relation with a derived field instead of picking a random suitable field would be a better approach.

Case 3: if multiple fields found then it will look for a field that has `@derivedFrom(field: "rel1")`:

1. a field with `@derivedFrom(field: "rel1")` directive is found, it is `OneToOne` relationship
2. not found an error will be thrown
3. there are multiple fields with the same directive (`@derivedFrom(field: "rel1")`) error

dzhelezov:
What is the naming convention for the database columns? In one example it's `authorId`, another is `author_id, post_id`. Probably we should document all the naming conventions we use, which is I believe (correct me if I'm wrong): snake case for the database, kebab case for file and folder names, and camel case for class field names (and class names are PascalCase: like camel case, but with the first letter capitalized).

query-node/substrate-query-framework/cli/src/generate/RelationshipGenerator.ts
```diff
@@ -0,0 +1,143 @@
...
+    o1.relatedEntityImports.add(o2.name);
+    o2.relatedEntityImports.add(o1.name);
+  }
+
+  generate(): void {
+    this.model.types.forEach(currentObject => {
+      for (const field of currentObject.fields) {
+        if (this.isVisited(field, currentObject)) continue;
+
+        // ============= Case 1 =============
```
dzhelezov: Perhaps we can refactor Case1, Case2, Case3 into separate methods with self-describing names? Otherwise it's kinda hard to follow what these cases really differentiate.

query-node/substrate-query-framework/cli/src/parse/WarthogModelBuilder.ts
```diff
@@ -95,6 +97,7 @@ export class WarthogModelBuilder {
       description: o.description?.value,
       isInterface: o.kind === 'InterfaceTypeDefinition',
       interfaces: o.kind === 'ObjectTypeDefinition' ? this.getInterfaces(o) : [],
+      relatedEntityImports: new Set<string>(),
```
dzhelezov: It feels like this field is not a genuine property of `ObjectType` but is calculated based on the fields. This is a potentially dangerous place, as we have methods for adding new fields to the existing ObjectType, and this would likely lead to an inconsistent state. Maybe it can be replaced by a method on `WarhogModel`, smth like `WarthogModel.getReferencedTypes(o: ObjectType)` ?

dzhelezov (changes requested):
Nice one! have a few small suggestions in the comments

metmirr:
> What is the naming convention for the database columns? In one example it's `authorId`, another is `author_id, post_id`. Probably we should document all the naming conventions we use, which is I believe (correct me if I'm wrong): snake case for the database, kebab case for file and folder names, and camel case for class field names (and class names are PascalCase: like camel case, but with the first letter capitalized).

Snake case for done in here: https://github.com/Joystream/joystream/pull/876

metmirr:
I have changed the logic for `isRelationtipType` method from `['otm', 'mto', 'oto'].some(s => s === this.type);` to `return this.relation ? true : false;` since I don't use `['otm', 'mto', 'oto']`

query-node/substrate-query-framework/cli/src/parse/WarthogModelBuilder.ts
metmirr: Here I am initiating the set but what you mean is to get `relatedEntityImports`. Do you mean to move it `Set` initialization to `addObjectType` method?

dzhelezov (approved):

bedeho (commented):
I am writing this reply under the assumption that the initial description at the top about how the input schemas imply what relationships are deduced for the database and API.

My main goal in this review was to just establish what the relationship semantics are for input schemas, as I am just looking to make sure we can do what we need, and also understand what is not possible.

To that end, one thing I really like about this description is that it clearly shows what things are modelled at in the datbase. However, it's a big problem is that it does not actually define what the _rules_ are, it just provides examples of what you _can do_. This is very similar to The Graph, and that is problematic.

Is there a better way to describe what is allowed, and for each such case.

I have specific questions that currently are not unambigiously addressed, like

1) in 1-1 case, do I have to use `derivedFrom` on one side, or is it an option? if its an option, then what is the difference? if its not an option, then why is it needed at all? why cant you just have non-list type reference on each end?

2) in 1-1 case, what happens if I do these cases

```graphql
type User @entity {
  name: String!
  profile: Profile! @derivedFrom(field: "user")
  another_profile: Profile! @derivedFrom(field: "user")
}
```

or

```graphql
type User @entity {
  name: String!
  profile: Profile! @derivedFrom(field: "user")
  another_profile: Profile!
}
```

or

```graphql
type User @entity {
  name: String!
  profile: Profile!
  another_profile: Profile!
}
```

3) What happens if I do something similar as in 2 but with 1-2-m, that is multiple relationships involving the same set of entities, and how does use of derivedFrom interact?

4) Same as 4, but with m-2-m.

metmirr:
### Rules
1. At the moment we support only one relationship between two types.
2. `@derivedField` is required for `1:1` in order to refer to the correct field. If there are multiple fields of the same type then it will be uncertain which field is related to which field. (_since we support only one relationship it is unnecessary to enforce @derivedFrom for `1:1`_). For example 

```graphql
type User @entity {
  name: String!
  profile: Profile!
  another_profile: Profile!
}

type Profile @entity {
  avatar: String!
  user: User!
  anotherUser: User!
}
```

3. Having more than one field with the same parameter is not allowed for the derived field (`@derivedField(field: "<field_name>")`). For example 
```graphql
profile: Profile! @derivedFrom(field: "user")
another_profile: Profile! @derivedFrom(field: "user")
```
4. `@derivedField` is optional for `1:n` relationship.

### Answers to questions
1. Rule 2
2. Rule 1

Issue #766, opened 2020-06-23
Forum tech debt + cleanup


bedeho:
# Debt

- Remove internal actor representations: https://github.com/Joystream/joystream/issues/762
- Remove commented out `add_extra_genesis` fragment.
- `CategoryByModerator` models a set, as such it should just map to an empty type, as the `bool` value type has no semantic meaning, and can be misleading.
- Remove all label label semantics in the runtime.
- Drop the concept of `ChildPositionInParentCategory` which is used for `Category.position_in_parent_category`, just use an optional identifier to a parent.

## Removing archival state

Before we had proper query infrastructure (now being built), all the business logic state powering the UIs had to come directly from the blockchain state. This meant we shoved lots of write only data in there for archival purposes, this should be avoided. Here we will attempt to remove all such data.

- We no longer store category names, post/thread titles or texts, or any other text, just the hash.
- We no longer enforce length constraints on any such texts either, hence all such parameters and checks can be removed.
- We no longer store old versions of an edited post.
- We no longer store poll alternative texts: `PollAlternative.alternative_text`, just hash.
- We no longer store write only dates (e.g. encoded by `BlockchainTimestamp`), such as when something was created (e.g. `Poll.start_time`).
- We no longer store `Post.nr_in_thread` and `Thread.nr_in_category`.
- We no longer store `num_unmoderated_posts` and `num_moderated_posts` in `Thread`.
- Drop all deletion indicators (post, threads, categories), see next section for genuine deletion.
- We no longer store title, description, creation date, deletions status in categories. They should however just have a single counter for the number of threads and the number of subcategories in as these are not archival, since they must be read when evaluating whether category deletion is safe. But there is no need to distinguish between number of moderated vs. unmoderated threads.

# Enhancements

- Make thread title editable by author.
- Allow moderators to move threads between categories, if they have power in both source and destination.
- Make maximum category depth parameter getable through runtime trait parameters, remove from storage: `MaxCategoryDepth`
- Efficient thread deletion: https://github.com/Joystream/joystream/issues/760
- Moderation semantics for posts changed to mean that it is deleted by the moderator, and that some accompanying rationale is submitted, but is not itself stored.
- Moderation semantics for threads changed in the same way as posts. This can only be done _after_ efficient thread deletion has been implemented.
- Forum category deletion: https://github.com/Joystream/joystream/issues/761
- Allow both the author and forum moderators to update archival status.
- Generalize PostReaction: https://github.com/Joystream/joystream/issues/763
[...]

Issue #759, opened 2020-06-23
Sideproject: Completing and Incorporating Forum 2.0


bedeho:
# Background

We have a completed version of a new set of forum functionalities in the `development` branch of this repo

https://github.com/Joystream/substrate-forum-module

There is also migration code there which can be used in a runtime upgrade to migrate over the old forum state to the new one. Lastly, there is some benchmarking code that attempts to measure how much time a given machine would take to run a migration on a forum of a given size. It has been used to make sure that the migration portion of the runtime upgrade could safely be executed.

# Goals

Here are the goals for this project, in order of execution. Notice that all the work after point 5. is about the non-runtime work of fully putting the forum in production. It's not yet clear how far we should go down that path, as it effectively makes the project into a full network release that probably requires the input of other team members for timely completion.

1. Introduce the new forum code into the monorepo, while preserving its commit history.
2. Augment the forum to get rid of technical debt and make it compatible with the new working group system. This will require updating unit tests and the migration code. Read here https://github.com/Joystream/joystream/issues/766.
3. Introduce proper module documentation.
4. Introduce the forum module into the runtime.
5. Introduce a working group for the forum in the runtime, and integrate it with the forum.
<-- non-runtime work below -->
6. Update the Joystream/types typescript library to allow off-chain typescript apps to talk to the new full node.
7. Update the CLI to allow it to work with the forum working group.
8. Update the CLI to allow it to be used by moderators
   - lead can manipulate moderator assignment to categories
   - lead can create categories (or could moderators also do this?)
   - moderators can delete posts
   - moderators can archive threads
   - moderators can delete threads
9. Update the network integration testing to also cover basic user and moderator actions on the forum.
10. Write a runtime upgrade scenario, with forum migration, into the network integration testing.
11. Write query node schemas for the 2.0 forum: most important improvement is representing full history on all objects: who edited/removed/created/.. what+when.
12. Update the Pioneer experience to expose new forum functionality and use the query node.

# Workflow

We will follow the current Joystream branching model

https://github.com/Joystream/joystream/issues/638

[...]

PR #768, opened 2020-06-23
Forum 2

ondratra:
Draft for new forum; solving issues #759 & #766.


mnaamani:
The runtime build CI check is failing because of bad formatting. We have git hooks configured which don't seem to be running on your side.

Try running `yarn` in the root folder it should install all dependencies including husky which will setup and configure the githooks.

If that fails you can try to do `yarn cargo-checks` to see what's wrong.

To fix formatting issues, you can just run `cargo fmt --all`

ondratra:
@mnaamani Thanks for the info about `yarn`. Maybe we should introduce info about the necessity of running `yarn` command in root to the documentation as there doesn't seem to be a mention about it right now? I've also noticed before that Travis is using `cargo clippy` in addition to `cargo fmt --all`. 

Btw: Travis builds fail on compile errors, not the formatting errors (except the last commit where I introduced one formatting error you've noticed). I'm still solving those errors.

mnaamani:
Thanks for the update @ondratra I will make a note of updating the instructions in README

mnaamani:
> Travis is using `cargo clippy` in addition to `cargo fmt --all`

Correct. If you run `yarn cargo-checks` it will also do both formatting and clippy checks

PR #785, opened 2020-06-24
@joystream/types: remove need for lib from import path

mnaamani:
- Finally getting rid of need to have `/lib/` in the import path.
- Enabled `forceConsistentCasingInFileNames` tsconfig option which 
   "Disallow inconsistently-cased references to the same file"
  This option would have detected the issue we once had on OSX https://github.com/Joystream/joystream/pull/514
- Updated cli and tests in nicaea branch to use new path


mnaamani:
Doing a test by packing the package with `npm pack` when I try to install it, it actually runs `build` script, and I'm getting some errors:

```
$ npm run build
npm WARN lifecycle The node binary used for scripts is /var/folders/x1/jzcg_82x2yndlklx7ystth580000gn/T/yarn--1593013889109-0.3527586406031573/node but npm is using /usr/local/Cellar/node@12/12.18.0/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> @joystream/types@0.11.0 build /Users/mokhtar/tmp/package
> tsc --build tsconfig.json

node_modules/@polkadot/types/types.d.ts:1:29 - error TS2307: Cannot find module '@polkadot/keyring/types' or its corresponding type declarations.

1 import { SignOptions } from '@polkadot/keyring/types';
                              ~~~~~~~~~~~~~~~~~~~~~~~~~

src/versioned-store/index.ts:8:39 - error TS7016: Could not find a declaration file for module 'lodash'. '/Users/mokhtar/tmp/package/node_modules/lodash/lodash.js' implicitly has an 'any' type.
  Try `npm install @types/lodash` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash';`

8 import { camelCase, upperFirst } from 'lodash'
                                        ~~~~~~~~

src/media.ts:8:46 - error TS2307: Cannot find module '@polkadot/keyring' or its corresponding type declarations.

8 import { encodeAddress, decodeAddress } from '@polkadot/keyring';
                                               ~~~~~~~~~~~~~~~~~~~


Found 3 errors.
```

Will try to resolve

mnaamani:
Fixed by adding missing dependencies  in https://github.com/Joystream/joystream/pull/785/commits/26c008f7c5c31cb02c230a46cb7387ee7014bb86

types/.gitignore
```diff
@@ -1 +1,10 @@
-/lib/
+# Don't track build artifacts
+**/*.js
+**/*.d.ts
+hiring/schemas
```
Lezek123: I'm not sure why this is inlcuded as part of `.gitignore`

Lezek123 (commented):
Seems to work fine after removing `node_modules` inside the root and project folders and running `yarn install`.
Tested CLI, Pioneer and Network Tests.
Also tried using the tarball created with `npm pack` within the CLI.
Didn't run into any issues that I could track to changes introduced in this PR.

Merging this will require updates in: https://github.com/Joystream/joystream/pull/707 (where some additional imports using the old `/lib` path were included)

types/.gitignore
mnaamani: Added it to ignore the `role.schema.json` which is moved into there because its imported by `src/hiring/index.ts`

```
import * as role_schema_json from './schemas/role.schema.json'
```

I will be more explicit and make it ignore that exact file.

Lezek123 (approved):
I confused `hiring/schemas` with `src/hiring/schemas`, now it makes sense, I think it's good to go then.

gleb-urvanov (approved):

PR #817, opened 2020-06-25
Content dir 2.0 tests: curator group related extrinsics full coverage

iorveth:
This PR aims to introduce full test coverage for curator group related extrinsics


bedeho (approved):
Very nice job.

PR #822, opened 2020-06-26
Content dir 2 tests: class related extrinsics full coverage

iorveth:
The scope of this PR includes failure cases coverage for all `Class` related (root) extrinsics.
Depends on: https://github.com/Joystream/joystream/pull/817


bedeho (approved):
Awesome job here, really impressed by the breadth of cases ;)

One thing which does suck a bit is that many of the cases will probably go away when we stop storing texts buffers in storage, as we then will not need to care about length constraints any longer.

Issue #762, opened 2020-02-24
Remove internal actor representations


bedeho:
We have internal actor representations `ForumSudo`, `ForumUser` and `Moderator`, they should not live here.

# `ForumSudo`

Rather than storing a hardcoded account, we do not store anything, and instead expect the runtime trait to have a function `is_lead(AccountId) -> bool`, which will provide the right answer.
All forum sudo reserved extrinsic will expect a signed origin, and then ensure that this function returns `true`.

In the runtime, the function will be implemented as a check that the given account si the account of the currently set lead in the corresponding working group. If no such lead is set, then it obviously also returns `false`.

# `ForumUser`

We just drop the type and storage fields for the user entirely. Instead, all attempts to use the forum functionality as a user will expect the caller to provide a forum user id, and signed origin, and then the forum module will simply ask a function on the runtime trait whether the provided id can authenticate with the account signing the origin. This function will provide a `bool` indicating the answer.

In the runtime, the function will be implemented as a check which interprets the id as a membership id, and compares the account with the role account of the member. There other examples of this approach, for example in the new working group module.

# `Moderator`

Here the idea is quite similar. We store nothing in the forum itself, and just expect a lookup function `is_moderator(ModeratorId, AccountId) -> bool` on the runtime.

In the runtime, the function will be implemented as a check that interprets the id as a worker id in the corresponding working group, and that the account matches the role account of that worker.

bedeho:
Done.

PR #827, opened 2020-06-26
Forum internal actors

ondratra:
related to #762

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -255,9 +255,7 @@ pub fn testnet_genesis(
             default_paid_membership_fee: 100u128,
             members: crate::members_config::initial_members(),
         }),
-        forum: Some(crate::forum_config::from_serialized::create(
-            endowed_accounts[0].clone(),
-        )),
+        forum: Some(crate::forum_config::from_serialized::create()),
         data_object_type_registry: Some(DataObjectTypeRegistryConfig {
             first_data_object_type_id: 1,
         }),
--- a/node/src/forum_config/from_serialized.rs
+++ b/node/src/forum_config/from_serialized.rs
@@ -3,8 +3,7 @@
 use super::new_validation;
 use node_runtime::{
     forum::{Category, Post, Thread},
-    AccountId, BlockNumber, CategoryId, ForumConfig, ForumUserId, ModeratorId, Moment, PostId,
-    ThreadId,
+    BlockNumber, CategoryId, ForumConfig, ForumUserId, ModeratorId, Moment, PostId, ThreadId,
 };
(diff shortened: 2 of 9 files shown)


bedeho (changes requested):
Well done, however there is room for improvement in crafting a better logical unit of work.

You do not give any real scope in the PR description, so I can only go by the title, however, you are inconsistently treating the removal of different roles, any there is no obvious reason why. The lead seems to be removed entirely, the user type has quite a lot removed - but there is still for example the event `ForumUserCreated`, and there is no user authentication introduced, and the moderator type is entirely untouched. 

You could have sliced this many different ways, the simplest would just be to only fix the concept of sudo and lead, however you went beyond this in an inconsistent way, so its hard to evaluate what is going on.

Please try to refract the PR scope text to match some better unit of work, and update the work outstanding.

ondratra:
This PR was marked probably sooner than it should, sorry for that. 

Changes focus on module sections of #762. The internal representation of forum user, moderator, and sudo were removed. Most of the code is reusing old concepts of sudo, as you mentioned. Important functions [ensure_is_forum_sudo](https://github.com/Joystream/joystream/pull/827/files#diff-5c3c034195f94c91f269a928d62e410cL1967), `ensure_is_forum_member_with_correct_account`, `ensure_is_moderator_with_correct_account` were reimplemented while connections to the rest of the code were (mostly) retained.

I am not sure what the authentication mechanism should be. I thought the point of [is_lead](https://github.com/Joystream/joystream/blob/375896fc6941e70a323c1993ce031d13daf03483/runtime-modules/forum/src/lib.rs#L287), `is_moderator`, and `is_forum_member` is to delegate checks for user privileges and authentication itself to external code. Right now module excepts that runtime has its own representation of user ids, etc. and passes it to module's actions like [add_post(origin, formUserId, ...)](https://github.com/Joystream/joystream/blob/375896fc6941e70a323c1993ce031d13daf03483/runtime-modules/forum/src/lib.rs#L1250)

`ForumUserCreated` is removed now as well as moderator (see 2 latest commits).

bedeho (changes requested):
Looks goods, only a few minor things now.

- We are repeating ourselves a lot in the authentication. Its always to steps
1. check origin is signed
2. check that signer account and possible id is who they claim they are

- Can we rename `ensure_is_forum_member_with_correct_account` to `ensure_is_forum_user`.

- Can we rename `ensure_is_moderator_with_correct_account` to `ensure_is_moderator`.



can we just combine this into a single check, for each role of course, which just takes the origin and returns both the account (which is almost always ignored), and the underlying

bedeho (approved):
Perfect!

PR #838, opened 2020-06-30
Forum category position removal

ondratra:
Solves subtask of #766 - *Drop the concept of ChildPositionInParentCategory which is used for Category.position_in_parent_category, just use an optional identifier to a parent.*

To minimize merge conflict potential, this PR assumes #827 will be merged first.


bedeho (approved):
Perfect.

PR #839, opened 2020-06-30
Forum category to moderator type

ondratra:
Solves subtask of #766 - *CategoryByModerator models a set, as such it should just map to an empty type, as the bool value type has no semantic meaning, and can be misleading.*

To minimize merge conflict potential, this PR assumes #827 will be merged first.


bedeho (approved):
Perfect.

PR #837, opened 2020-06-30
Forum label removal

ondratra:
Solves subtask of #766 - *Remove all label label semantics in the runtime.*

To minimize merge conflict potential, this PR assumes #827 will be merged first.

PR #843, opened 2020-06-30
Forum archival state removal 1

ondratra:
Solves subtasks of #766:
- Remove all label label semantics in the runtime
- We no longer store category names, post/thread titles or texts, or any other text, just the hash.
- We no longer enforce length constraints on any such texts either, hence all such parameters and checks can be removed.
- We no longer store poll alternative texts: `PollAlternative.alternative_text`, just hash.

To minimize merge conflict potential, this PR assumes #837 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/mock.rs
+++ b/runtime-modules/forum/src/mock.rs
@@ -112,16 +112,17 @@ pub const INVLAID_CATEGORY_ID: <Runtime as Trait>::CategoryId = 333;
 
 pub const NOT_REGISTER_MODERATOR_ID: <Runtime as Trait>::ModeratorId = 666;
 
-pub fn generate_poll() -> Poll<<Runtime as timestamp::Trait>::Moment> {
+pub fn generate_poll(
+) -> Poll<<Runtime as timestamp::Trait>::Moment, <Runtime as system::Trait>::Hash> {
     Poll {
-        poll_description: b"poll description".to_vec(),
+        hash: generate_hash(),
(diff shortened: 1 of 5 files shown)


bedeho (changes requested):
- The runtime has to compute the hash from the raw payload, the user does not get to provide the hash, otherwise, there is no way to assure its authenticity.

- You have named all the fields `hash`, that's not very helpful. So instead make `Poll::hash` into `Poll::description_hash`.

- You have combined fields, such as category title and description, they should be separate hashes.

runtime-modules/forum/src/mock.rs
```diff
@@ -514,6 +467,12 @@ pub fn migration_not_done_config() -> GenesisConfig<Runtime> {
     create_genesis_config(false)
 }
 
+pub fn generate_hash() -> <Runtime as system::Trait>::Hash {
+    let hash = <Runtime as system::Trait>::Hash::random();
```
bedeho: I presume this generates a random hash during the tests? If so then I think we should avoid this, any form of non-determinism in testing invariably leads to hard to detect problems later.

bedeho (changes requested):
Looks good, just one more fix.

runtime-modules/forum/src/mock.rs
ondratra: Random hash generation was replaced by a better (deterministic) mock.

bedeho (approved):
Perfect.

PR #844, opened 2020-06-30
Forum archival state removal 2

ondratra:
Solves subtask of #766 - *We no longer store old versions of an edited post.*

To minimize merge conflict potential, this PR assumes #843 will be merged first.

PR #845, opened 2020-06-30
Forum archival state removal 3

ondratra:
Solves subtask of #766 - *We no longer store write only dates (e.g. encoded by BlockchainTimestamp), such as when something was created (e.g. Poll.start_time).*

To minimize merge conflict potential, this PR assumes #844 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -383,24 +383,10 @@ pub struct Moderator<AccountId> {
     pub self_introduction: Vec<u8>,
 }
 
-/// Convenient composite time stamp
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
-#[derive(Encode, Decode, Default, Clone, PartialEq, Eq)]
-pub struct BlockchainTimestamp<BlockNumber, Moment> {
-    /// Current block number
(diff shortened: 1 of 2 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -461,18 +447,15 @@ pub struct Poll<Timestamp, Hash> {
...
 
     /// Unique identifier
     pub hash: Hash,
 
     /// Possible moderation of this post
-    pub moderation: Option<ModerationAction<ModeratorId, BlockNumber, Moment>>,
-
-    /// When post was submitted.
-    pub created_at: BlockchainTimestamp<BlockNumber, Moment>,
+    pub moderation: Option<ModerationAction<ModeratorId>>,
```
bedeho, line 447: Why is this still here?

bedeho (changes requested):

runtime-modules/forum/src/lib.rs
ondratra: We discussed complete removal of moderation after I created PRs *forum archival state* 3 and 4. To prevent merge conflicts in 4, `moderationAction` will be removed in additional PR with a couple of other changes.

bedeho (approved):

PR #846, opened 2020-06-30
Forum archival state removal 4

ondratra:
Solves subtasks of #766:
- We no longer store Post.nr_in_thread and Thread.nr_in_category.
- We no longer store num_unmoderated_posts and num_moderated_posts in Thread.
- Drop all deletion indicators (post, threads, categories), see next section for genuine deletion.
- We no longer store title, description, creation date, deletions status in categories. They should however just have a single counter for the number of threads and the number of subcategories in as these are not archival, since they must be read when evaluating whether category deletion is safe. But there is no need to distinguish between number of moderated vs. unmoderated threads.

To minimize merge conflict potential, this PR assumes #845 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -340,8 +340,6 @@ const ERROR_POST_MODERATED: &str = "Post is moderated.";
 
 // Errors about category.
 const ERROR_CATEGORY_NOT_BEING_UPDATED: &str = "Category not being updated.";
-const ERROR_CATEGORY_CANNOT_BE_UNARCHIVED_WHEN_DELETED: &str =
-    "Category cannot be unarchived when deleted.";
 const ERROR_MODERATOR_MODERATE_CATEGORY: &str = "Moderator can not moderate category.";
 const ERROR_EXCEED_MAX_CATEGORY_DEPTH: &str = "Category exceed max depth.";
 const ERROR_ANCESTOR_CATEGORY_IMMUTABLE: &str =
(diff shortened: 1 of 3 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -520,29 +481,14 @@ pub struct Category<CategoryId, ThreadId, Hash> {
...
-    /// of a thread, the variables are incremented and decremented, respectively.
-    ///
-    /// These values are vital for light clients, in order to validate that they are
-    /// not being censored from subcategories or threads in a category.
+    /// Number of subcategories
     pub num_direct_subcategories: u32,
-    pub num_direct_unmoderated_threads: u32,
-    pub num_direct_moderated_threads: u32,
+
+    // Number of threads in category
```
bedeho: Can we extend this comment to reflect why this is stored, namely that we need to ensure that a category is empty to allow deletion of categories. Same for `num_direct_subcategories`.

runtime-modules/forum/src/lib.rs
```diff
@@ -770,7 +705,7 @@ decl_module! {
         }
 
         /// Update category
-        fn update_category(origin, category_id: T::CategoryId, new_archival_status: Option<bool>, new_deletion_status: Option<bool>) -> dispatch::Result {
+        fn update_category(origin, category_id: T::CategoryId, new_archival_status: Option<bool>) -> dispatch::Result {
```
bedeho: 1. It does not make sense to have `Option<bool>` parameter now, it only made sense when this single extrinsic was used to update more than one state variable, currently, there is no good reason to pass `None`.
2. You may as well just rename this to `update_category_archival_status`.

bedeho (changes requested):
Only minor fixes required, otherwise it looks good!

ondratra:
Both issues were addressed.

bedeho (approved):
Perfect.

PR #842, opened 2020-06-30
Content directory 2.0 tests: update_entity_permissions full coverage

iorveth:
Closes https://github.com/Joystream/joystream/issues/790
Depends on: #841, #822, #817

PR #841, opened 2020-06-30
Content_dir 2.0 tests: update_entity_creation_voucher & transaction extrinsics full test coverage

iorveth:
Closes https://github.com/Joystream/joystream/issues/789
Depends on: https://github.com/Joystream/joystream/pull/822, https://github.com/Joystream/joystream/pull/817

PR #848, opened 2020-07-01
update_entity_permissions full test coverage

### LINKED ISSUE #790: update_entity_permissions full test coverage
2 failure case

### LINKED ISSUE #791: create_entity extrinsic full test coverage
6 failure cases

iorveth:
Closes #791
Depends on: #842, #841, #822, #817

### DIFF
--- a/runtime-modules/content-directory/src/permissions.rs
+++ b/runtime-modules/content-directory/src/permissions.rs
@@ -107,8 +107,8 @@ pub fn perform_curator_in_group_auth<T: Trait>(
     // Ensure curator authorization performed succesfully
     ensure_curator_auth_success::<T>(curator_id, account_id)?;
 
-    // Ensure CuratorGroup under given curator_group_id exists, retrieve corresponding one
-    let curator_group = Module::<T>::ensure_curator_group_exists(curator_group_id)?;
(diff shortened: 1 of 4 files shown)


runtime-modules/content-directory/src/permissions.rs
```diff
@@ -107,8 +107,8 @@ pub fn perform_curator_in_group_auth<T: Trait>(
     // Ensure curator authorization performed succesfully
     ensure_curator_auth_success::<T>(curator_id, account_id)?;
 
-    // Ensure CuratorGroup under given curator_group_id exists, retrieve corresponding one
-    let curator_group = Module::<T>::ensure_curator_group_exists(curator_group_id)?;
+    // Retrieve corresponding curator group
+    let curator_group = Module::<T>::curator_group_by_id(curator_group_id);
```
bedeho, line 111: Why was this check removed? 

This would return a default constructed group if the `curator_group_id` was not valid, and I assume that the next ensure check would trigger `ERROR_CURATOR_GROUP_IS_NOT_ACTIVE`, which would stop the call. This is however not safe. It will depend on behaviour of `is_active` in default mode in a way which is not clear going forward, and it also returns a misleading dispatch error.

PR #851, opened 2020-07-01
Fix zero stake bug in the hiring module.

### LINKED ISSUE #702: Working group: panic/exception when trying to apply on opening with "at least 0" staking policy
Worker opening could be created with staking policy stating at least 0 stake, eg the following one:
```
application_staking_policy: {
            amount: 0,
            amount_mode: 'AtLeast',
            crowded_out_unstaking_period_length: 0,
            review_period_expired_unstaking_period_length: 0,
          },
          role_staking_policy: {
            amount: 0,
            amount_mode: 'AtLeast',
            crowded_out_unstaking_period_length: 0,
            review_period_expired_unstaking_period_length: 0,
          },
```

While trying to create a worker application with empty (None) stake, the following error will appear:
```
DEBUG runtime  DispatchError
DEBUG runtime  36
DEBUG runtime  56
DEBUG runtime  AddWorkerOpeningStakeMissingWhenRequired
```

While trying to create a worker with 0 stake, the panic will happen:
```
DEBUG runtime  panicked at 'assertion failed: `(left == right)`
  left: `Err(Error(CannotStakeZero))`,
 right: `Ok(())`', /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8/src/libcore/macros/mod.rs:16:9
```

shamil-gadelshin:
Bugfix description:
- We had a check in the hiring module to not allow the stake less than `Currency::minimum_balance()` 
- Currency::minimum_balance() could be traced to the `ExistentialDeposit` constant in the runtime which is set to zero. 
- It's unclear why we had bound the minimum_balance with the minimum allowed stake value, but with our settings (minimum balance = 0) we have a contradiction with the minimum allowed stake value in the stake module witch is greater than zero.

Changes:
- I introduced the `MinimumStakeBalance` const in the hiring module and set it to one in the runtime and change the code to use this constant.
- I change the tests in the hiring module and introduce another one in the working group module.

https://github.com/Joystream/joystream/issues/702

**UPDATE (02.07.2020):**
1. Restored `Currency::minimum_balance()` as one of the valid checks
2. Deleted unnecessary constant.
3. Add `AddOpeningError::StakeAmountCannotBeZero` error in the hiring module
4. Add support for the new error message in the working group and content directory working group modules.


bedeho (approved):
LGTM.

PR #867, opened 2020-07-02
Forum archival state removal 5

ondratra:
Solves the rest of #766's *Removing archival state* subtask.
- removes poll `start_time`
- fixes some tests around poll time constraints
- removes `moderationAction`

To minimize merge conflict potential, this PR assumes #846 will be merged first.

### DIFF
--- a/node/src/forum_config/from_serialized.rs
+++ b/node/src/forum_config/from_serialized.rs
@@ -3,7 +3,7 @@
 use super::new_validation;
 use node_runtime::{
     forum::{Category, Post, Thread},
-    CategoryId, ForumConfig, ForumUserId, Hash, ModeratorId, Moment, PostId, ThreadId,
+    CategoryId, ForumConfig, ForumUserId, Hash, Moment, PostId, ThreadId,
 };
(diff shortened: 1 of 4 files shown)


bedeho (changes requested):
Happy to merge this, but just be aware that `moderate_thread` i now in the awkward state of not actually doing any updates to storage. This is fine, so long as it's fixed later.

May I just ask that you rename `ensure_moderate_category` to `ensure_can_moderate_category`.

ondratra:
`ensure_can_moderate_category` is renamed. 

There are no other changes to `moderate_thread` in other PRs. I am not exactly sure what moderation represents, that's why I don't know if it can be removed entirely or there should be some storing added. Can you elaborate on that, please?

bedeho:
The initial issue addresses this

> Moderation semantics for posts changed to mean that it is deleted by the moderator, and that some accompanying rationale is submitted, but is not itself stored.
Moderation semantics for threads changed in the same way as posts. This can only be done after efficient thread deletion has been implemented.

PR #868, opened 2020-07-02
fix router

kdembler:
Last PR merge broke the Router which requires specific children components and prevents Atlas from rendering, this PR fixes that.


fcescob (approved):

fcescob:
Both the preview and the code changes looks good to me, approved!

bedeho (approved):

PR #870, opened 2020-07-02
improve video preview component, add to storybook

### LINKED ISSUE #749: Enhance VideoPreview [Atlas]
Implement additional states for the Video Preview Component, like the ones seen in https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=730%3A5283

kdembler:
This PR improves the `VideoPreview` component by adding a couple of missing things:

- video duration overlay
- video watch progress bar
- hover state

Some small details were refined as well, like spacing and click areas.

Screenshot 2020-07-02 at 3 49 32 PM

The component was added to the Storybook as well. Knobs were used to provide options allowing to check how it looks like in all possible variants:
Screenshot 2020-07-02 at 3 50 52 PM

@Gamaranto I've had some Storybook changes on the branch that overlapped with yours. I've used the Layout from the `app` package that should give us more consistent styling. Let me know if you think we should extract it into `components` as part of this PR

Resolves #749


atlas/packages/components/src/theme/typography.ts
```diff
@@ -1,14 +1,14 @@
 export default {
 	fonts: {
 		base: "Inter, Lato, 'Helvetica Neue', Arial, Helvetica, sans-serif",
-		headers: "PxGroteskRegular, Lato, 'Helvetica Neue', Arial, Helvetica, sans-serif"
+		headers: "PxGroteskRegular, Lato, 'Helvetica Neue', Arial, Helvetica, sans-serif",
```
fcescob, line 4: I believe those trailing commas are just Prettier changes, correct?

atlas/packages/components/.storybook/preview.jsx
```diff
@@ -0,0 +1,27 @@
...
+import { Layout } from "app/src/components";
+import theme from "./theme";
+
+const wrapperStyle = css`
+	padding: 10px;
+`;
+
+const stylesWrapperDecorator = (styleFn) => (
+	<div css={wrapperStyle}>
+		<Layout>{styleFn()}</Layout>
```
fcescob: I think we should stick with a self contained component here instead of importing `Layout` from the app, it makes Storybook build process not self-contained and harder to re-use in other projects

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
```diff
@@ -9,58 +23,90 @@ type VideoPreviewProps = {
...
+
+	const handleChannelClick = (e: React.MouseEvent<HTMLElement>) => {
+		if (!onChannelClick) {
+			return;
+		}
+		e.stopPropagation();
+		onChannelClick(e);
+	};
+
+	const handleClick = (e: React.MouseEvent<HTMLElement>) => {
```
fcescob, line 62: `handleClick` and `handleChannelClick` look almost identical, and the use case of triggering a callback only if some predicate is met looks very common. 
It improves the UX  since if cursor changes, then there is _for sure_ something to click. We could use this in other places where we have `onClick` handlers like buttons and so on.
I suggest to: 
- Define an HOF that takes a Callback, and returns a function which fires the callback iff a condition is met. In our example:

```javascript

const when = predicate => fn => e => predicate ? fn() : undefined
const handleClick = myFn(clickable)(onClick)
const handleChannelClick = myFn(channelClickable)(onChannelClick)
```

- Move this to `utils`.

I've left out calling `e.stopPropagation( )` from above, to make the example smaller, but you get the idea. This might be out of scope, so we could open an issue and move it to a future PR.

atlas/packages/components/src/theme/typography.ts
kdembler: Yes, I have auto formatting on save and it added them. Once we cleanup and format the entire project we shouldn't get this kind of changes

atlas/packages/components/.storybook/preview.jsx
kdembler: All right, would you be fine with moving the declaration of `globalStyles` to `components` then so we can use it for both Storybook and `Layout` in `app`?

atlas/packages/components/src/components/VideoPreview/VideoPreview.tsx
kdembler: I'm not sure if this is an abstraction worth having at this time:
- Checking a predicate inside a callback isn't a lot of work, especially once you have an event handler anyway. It's a matter of adding one `if` line
- The specific conditions and order in which you want to check them may be pretty custom to the callback. Handlers may start off pretty simple, allowing the use of the helper, but grow in time, requiring moving the condition inside the handler, making refactor more work
- I believe it would be easier to read when the condition is inside the handler itself, you don't need to check any helper function living outside of the component for its behaviour, you just look at the handler

atlas/packages/components/.storybook/preview.jsx
fcescob: Sounds like a good idea, let's go for it.

kdembler: Done, created `GlobalStyle` component and used it both inside `Layout` and for Storybook

bedeho (changes requested):
Looking at the design animation, this is not quite correct yet, please consult

https://www.dropbox.com/s/mx2wminhpwxx7by/Screen%20Recording%202020-06-04%20at%2011.14.42.mov?dl=0

kdembler:
@bedeho Pushed changes adding animation details:
* Play icon slide up
* Faded background image for better contrast

bedeho (approved):

PR #871, opened 2020-07-02
Forum thread editable by author

ondratra:
Solves subtask of #766:
- Make thread title editable by author.

To minimize merge conflict potential, this PR assumes #867 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -367,6 +367,14 @@ const ERROR_DATA_MIGRATION_NOT_DONE: &str = "data migration not done yet.";
 
 use system::ensure_signed;
 
+/// Represents a revision of the text of a Post
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq)]
+pub struct PostTextChange<Hash> {
+    /// Text that expired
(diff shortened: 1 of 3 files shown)


ondratra:
There is an inconsistency between editing the post's text and thread's title. Post saves a history of changes to the text while thread does not. How will we unite this behavior?

bedeho:
Neither should store history of any kind, just compute and store hash commitment of most recent text.

ondratra:
Ok, history was removed.

runtime-modules/forum/src/lib.rs
```diff
@@ -929,25 +926,16 @@ decl_module! {
             // Signer does not match creator of post with identifier postId
             ensure!(post.author_id == forum_user_id, ERROR_ACCOUNT_DOES_NOT_MATCH_POST_AUTHOR);
 
-            // Update post text and record update history
+            // Update post text
             <PostById<T>>::mutate(post_id, |p| {
                 let text_hash = T::calculate_hash(&new_text);
```
bedeho: Seeing that this was already done before but in general, computation should not be done inside the `mutation` call if possible, only updating the value should be done in the closure.

bedeho (changes requested):
Looks like post edit history was dropped properly, just a minor tweak.

bedeho (approved):

PR #876, opened 2020-07-03
Qnode naming strategy

metmirr:
This PR enables snake case naming for the database.
It is built upon https://github.com/Joystream/joystream/pull/764


dzhelezov (approved):
Looks good!

PR #877, opened 2020-07-03
Runtime DataObject migration

mnaamani:
- Re-enabled the clearing of data_directory at runtime upgrade
- Added a new `inject_data_objects(origin, BTreeMap<T::ContentId, DataObject<T>>)` root dispatchable method the directory
- Unit tests for 
- Had to refactor `struct DataObject` because after introducing `BTreeMap<T::ContentId, DataObject<T>>` type into the runtime public API, compiler complained that type `T::Trait` didn't implement the `Debug` trait.
- Updates @joystream/types with necessary new type `DataObjectsMap`, (removed some old stale registrations)

You can do a simple test of injection on a fresh develoment storage setup using the pioneer js toolbox http://localhost:3000/#/js

Paste this code snippet and run it:

```javascript
  const id = types.createType('ContentId', 'TEST')
  console.log(id)

  const object = types.createType('DataObject', {
    'owner': 200,
    'added_at': {'block': 100, 'time': 5000},
    'type_id': 1,
    'size': 40,
    'liaison': 0,
    'liaison_judgement': 'Accepted',
    // https://testnet.joystream.org/#/media/videos/925
    'ipfs_content_id': 'QmNw2pRA32PGTxVeq7ktaM9ttUHkj5P3is3QLqamLCJ7Bk'
  })

  const objects = types.createType('DataObjectsMap')

  objects.set(id, object)

  const injectTx = api.tx.dataDirectory.injectDataObjects(objects)
  // injectDataObjects requires root origin, so send as proposal on sudo module
  // fetch the required sudo account.
  const sudo = await api.query.sudo.key()

  // if the key is not in the keyring it will not work
  api.tx.sudo.sudo(injectTx).signAndSend(sudo)
```

You can then inspect the chain state to verify the object id `0x5445535400000000000000000000000000000000000000000000000000000000` 

exists.
If you check the log of the storage node, you should see it on the next sync run pinning the item:

```
  ipfs-http-client:request POST localhost:5001/api/v0/files/stat?with-local=true&arg=%2Fipfs%2FQmNw2pRA32PGTxVeq7ktaM9ttUHkj5P3is3QLqamLCJ7Bk&stream-channels=true 200 OK +3s
  joystream:storage:storage Pinning QmNw2pRA32PGTxVeq7ktaM9ttUHkj5P3is3QLqamLCJ7Bk +3s
  joystream:sync sync run complete +3s
```

The file is large so it may take some time to download :)

I will prepare the export (from constantinople testnet) / import scripts in a separate PR

### DIFF
--- a/runtime-modules/storage/src/data_directory.rs
+++ b/runtime-modules/storage/src/data_directory.rs
@@ -300,7 +300,7 @@ decl_module! {
         /// The number of objects that can be added per call is limited to prevent the dispatch
         /// from causing the block production to fail if it takes too much time to process.
         /// Existing data objects will be overwritten.
-        fn inject_data_objects(origin, objects: DataObjectsMap<T>) {
+        pub(crate) fn inject_data_objects(origin, objects: DataObjectsMap<T>) {
             ensure_root(origin)?;
 
             // limit size - do some benchmarking to test how much we can allow per call.
--- a/runtime-modules/storage/src/data_directory.rs
+++ b/runtime-modules/storage/src/data_directory.rs
@@ -25,6 +25,7 @@ use codec::{Decode, Encode};
 use rstd::collections::btree_map::BTreeMap;
 use rstd::prelude::*;
 use sr_primitives::traits::{MaybeSerialize, Member};
+use srml_support::traits::Get;
 use srml_support::{decl_error, decl_event, decl_module, decl_storage, ensure, Parameter};
 use system::{self, ensure_root};
 
@@ -57,6 +58,8 @@ pub trait Trait:
 
     /// Validates member id and origin combination.
     type MemberOriginValidator: ActorOriginValidator<Self::Origin, MemberId<Self>, Self::AccountId>;
+
+    type MaxObjectsPerInjection: Get<u32>;
 }
 
 decl_error! {
@@ -77,8 +80,8 @@ decl_error! {
         /// Require root origin in extrinsics.
         RequireRootOrigin,
 
-        /// DataObject Injection Failed. Invalid Input
-        DataObjectsInjectionFailed
+        /// DataObject Injection Failed. Too Many DataObjects.
+        DataObjectsInjectionExceededLimit
     }
 }
 
@@ -204,6 +207,8 @@ decl_module! {
         /// Predefined errors.
         type Error = Error;
 
+        /// Maximum objects allowed per inject_data_objects() transaction
+        const MaxObjectsPerInjection: u32 = T::MaxObjectsPerInjection::get();
 
         /// Adds the content to the system. Member id should match its origin. The created DataObject
         /// awaits liaison to accept or reject it.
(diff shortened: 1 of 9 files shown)


mnaamani:
I accidentally included some pioneer changes I was working on the side, will remove 15d889b93c2b10277f7207c1fbb42760e3aa6533

mnaamani:
To export data objects I used the following code in the js-toolbox on current testnet https://testnet.joystream.org/#/js

```js
  const ownerAccountToMemberId = async (accountId) => {
    const memberIds = await api.query.members.memberIdsByRootAccountId(accountId)
    return memberIds[0] || null
  }

  const ids = await api.query.dataDirectory.knownContentIds()
  // When a BTreeMap is constructed for injection the node will fail to decode
  // it if its not sorted.
  ids.sort()

  let transformed = await Promise.all(ids.map(async (id) => {
    let obj = await api.query.dataDirectory.dataObjectByContentId(id)
    if (obj.isNone) { return null }
    obj = obj.unwrap()

    return [id, {
      owner: await ownerAccountToMemberId(obj.owner),
      added_at: obj.added_at,
      type_id: obj.type_id,
      size: obj.size_in_bytes,
      liaison: new types.u64(0),
      liaison_judgement: obj.liaison_judgement,
      ipfs_content_id: obj.ipfs_content_id }]
  }))

  console.log(`Exporting ${transformed.length} objects`)
  console.log(JSON.stringify(transformed))
```

copying the resulting json string into this snippet on the development node to import:

```js
  const exported = `_THE_JSON_FROM_EXPORT_SCRIPT_`

  const objects = types.createType('DataObjectsMap')

  JSON.parse(exported).forEach((pair) => {
    objects.set(
      types.createType('ContentId', pair[0]),
      types.createType('DataObject', pair[1])
    )
  })

  const injectTx = api.tx.dataDirectory.injectDataObjects(objects)

  const sudo = await api.query.sudo.key()
  api.tx.sudo.sudo(injectTx).signAndSend(sudo)
```

At time of writing there were 129 objects, and all were injected successfully in a single transaction. So We can bump the max limit to 150.

Or it might be safer to just improve the code to make multiple transactions.

@bwhm could you share your code for filtering only objects that are referenced in the versioned store, so I can do that filtering on export.

runtime-modules/storage/src/data_directory.rs
```diff
@@ -279,13 +295,32 @@ decl_module! {
...
+        /// The operation is "silent" - no events will be emitted as objects are added.
+        /// The number of objects that can be added per call is limited to prevent the dispatch
+        /// from causing the block production to fail if it takes too much time to process.
+        /// Existing data objects will be overwritten.
+        pub(crate) fn inject_data_objects(origin, objects: DataObjectsMap<T>) {
             ensure_root(origin)?;
 
-            // == MUTATION SAFE ==
+            // limit size - do some benchmarking to test how much we can allow per call.
+            const MAX_OBJECTS: usize = 20;
```
shamil-gadelshin: This value looks like part of the interface. Probably we should make it configurable and export it using Substrate.

shamil-gadelshin (changes requested):

runtime-modules/storage/src/data_directory.rs
mnaamani: I was really trying to limit the change to the runtime api as much as possible to avoid adding an technical debt, since this method is a one time helper method, but I see your point. I'll export it as a constant.

mnaamani: fixed in 8d88c3823285ff4889109385424e36156c059812

shamil-gadelshin (approved):
LGTM

Issue #813, opened 2020-06-25
Set up proper CI [Atlas]


kdembler:
Right now, Atlas PRs run a bunch of unrelated checks (Pioneer, Rust CI, etc.). We should investigate our options on running these conditionally and disable them on Atlas PRs if feasible.

Additionally, we should add Atlas tests as a new check. We could run lint, prettier and tests to easily see the state of PR

fcescob:
Covered in #878

Issue #811, opened 2020-06-25
Tweak up ESLint [Atlas]


kdembler:
We should consider enabling linting with `eslint` as part of the git hooks (it's worth investigating how long will the linting take, but it should be fine).

There are also some rule tweaks we should consider doing:
- make sure `eqeqeq` is enabled
- potentially disable `react/prop-types` since we are using arrow functions that aren't currently supported by that rule: https://github.com/yannickcr/eslint-plugin-react/issues/2353

kdembler:
Also, I think we probably should disable or tweak `a11y` rules (for now at least). Since we're in the mocked components phase, we will get a lot of errors for non-problematic code.

fcescob:
Covered in #878

Issue #748, opened 2020-06-22
Fix Tests [Atlas]


fcescob:
Currently, Tests are out of date and don't run. 
Fix this by working tests for all components.

fcescob:
Covered in #878

kdembler:
Closing this as we have the initial tests coverage, if we come up with anything else to test, let's create a new issue with more details

PR #878, opened 2020-07-03
Atlas improve CI

fcescob:
This PR tackles issues #813 #811  and partially #748 

## What To Expect

All CI checks should be working, ESlint and tests should run correctly.

## What To Not Expect

The test suite completeness is not in the scope of this PR, but should be addressed in a separate PR.
This PR won't add changes to code formatting nor fix linting/formatting issues, doing so would cause a lot of changes across the codebase and would make it hard to review this same PR. I suggest we open a separate PR with the only scope of addressing linting and formatting issues.


.github/workflows/atlas-pr.yml
```diff
@@ -8,22 +8,43 @@ on:
...
+      - name: Use Node.js ${{matrix.node-version}}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{matrix.node-version}}
+      - name: Install modules
+        working-directory: ./atlas
+        run: yarn install --frozen-lockfile
+      - name: Run ESLint
+        working-directory: ./atlas
+        run: yarn run eslint . --ext .js,.jsx,.ts,.tsx
```
kdembler: I think it would make sense to have this as a script inside `package.json` and just call something like `yarn lint` here

.github/workflows/atlas-pr.yml
```diff
@@ -8,22 +8,43 @@ on:
...
       - name: Use Node.js ${{matrix.node-version}}
         uses: actions/setup-node@v1
         with:
           node-version: ${{matrix.node-version}}
-      - name: build
+      - name: Install modules
         working-directory: ./atlas
-        run: |
-          yarn install --frozen-lockfile
-          yarn build
```
kdembler, line 29: We don't need to do build here anymore because it's built by Netlify anyway, right? If a build fails we should get a failed check on the PR?

kdembler:
We still get failed Pioneer checks on this PR. Have you looked into what would need to be done to get rid of those?

.github/workflows/atlas-pr.yml
fcescob: Good idea, added it as a script in `package.json`

.github/workflows/atlas-pr.yml
fcescob: Correct! If there is an error during the build process, it will get cached by Netlify build process, and the preview will show as failed.

fcescob:
Pioneers build happen becuase how Netlify builds are set up, which run by default now. 
I think we would need some `netlify.toml` files to run a pre-build script and check nothing inside the Pioneer folder has changed, since we can't filter on braches for changes on Pioneer.
Maybe @mnaamani can take a look at this too, since changes need to be made to in the Netlify console it could mess with the current Pioneer build process and I would prefer to not mess with that for now

.github/workflows/atlas-pr.yml
kdembler: > it will get cached by Netlify build process

You mean that we will be able to see the build error?

kdembler:
Regarding Pioneer Netlify checks, we don't have to do that now but this will get PR checks for Atlas into usable state, until we do that, every PR will still fail. Changes made here are a huge improvement though!

atlas/__mocks__/svgrMock.js
```diff
@@ -0,0 +1,4 @@
+import * as React from 'react'
+export default 'SvgrURL'
+const SvgrMock = React.forwardRef((props, ref) => <span ref={ref} {...props} />)
```
kdembler, line 3: We have a linting issue here, we should start enforcing ESLint now so we need to either fix the issue or revise the rules behind it

kdembler (approved):
LGTM

.github/workflows/atlas-pr.yml
```diff
@@ -8,22 +8,43 @@ on:
...
+        node-version: [12.x]
+      fail-fast: true
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{matrix.node-version}}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{matrix.node-version}}
+      - name: Install modules
+        working-directory: ./atlas
```
mnaamani, line 46: The need to change working directory will be alleviated in future when the atlas/ directory is configured as a workspace as part of the root repo in the future.

mnaamani (approved):
LGTM

PR #879, opened 2020-07-03
add sidenav component

kdembler:
This PR tackles #483 by adding a sidenav component. We've decided that it won't currently be used inside the app so the component was added to the `components` package and the Storybook.

Screenshot 2020-07-03 at 11 49 43 AM

Screenshot 2020-07-03 at 11 49 49 AM


atlas/packages/components/stories/13-Sidenav.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+			<p>Sensorem, barcas, et fraticinida. Zeta manducares, tanquam barbatus gallus.</p>
+			<p>Sensorem, barcas, et fraticinida. Zeta manducares, tanquam barbatus gallus.</p>
+			<p>Sensorem, barcas, et fraticinida. Zeta manducares, tanquam barbatus gallus.</p>
+		</ContentWrapper>
+	</StoryStyles>
+);
+
+// this is needed because proper storybook styling isn't merged yet
+// TODO: remove
+const StoryStyles = styled.div`
```
kdembler, line 63: This will be removed once #870 is merged

atlas/packages/app/src/components/Layout.tsx
```diff
@@ -3,18 +3,21 @@ import { css, Global } from "@emotion/core";
...
-		box-sizing: border-box;
 		font-family: ${theme.typography.fonts.base};
 		background: ${theme.colors.black};
 		color: ${theme.colors.gray[500]};
 	}
 
+	*,
+	*::after,
+	*::before {
+		box-sizing: border-box;
```
kdembler, line 18: Just noticed this change breaks at least one existing component:

Screenshot 2020-07-03 at 7 37 29 PM

This is how it should look like:
Screenshot 2020-07-03 at 7 45 19 PM

However, I'd say we should keep this change and fix the component instead. `box-sizing: border-box` in general gives a much more predictable behavior and should be a default. That's also what all the CSS framework I've used do. @Gamaranto ping

fcescob: Yes definately, it is an easy fix anyway, we just need to adjust the margins back.

fcescob (approved):
Both the code and Storybook look good! Approved.

PR #882, opened 2020-07-03
Forum thread deletion

### LINKED ISSUE #760: Efficient thread deletion
- Organizing posts and threads in a double map has the benefit of allowing efficient removal of an entire thread. 
- Allow moderators to delete threads in categories they have control over.

ondratra:
Solves subtask of #766:
- Efficient thread deletion: #760

Fixes #760.

To minimize merge conflict potential, this PR assumes #871 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -503,7 +503,10 @@ decl_storage! {
         pub NextCategoryId get(next_category_id) config(): T::CategoryId;
 
         /// Map thread identifier to corresponding thread.
-        pub ThreadById get(thread_by_id) config(): map T::ThreadId => Thread<T::ForumUserId, T::CategoryId, T::Moment, T::Hash>;
+        pub ThreadById get(thread_by_id) config(): double_map T::CategoryId, blake2_256(T::ThreadId) => Thread<T::ForumUserId, T::CategoryId, T::Moment, T::Hash>;
+
+        // Map category to thread. Enables searching for threads in ThreadById without knowing to which category thread belongs to.
+        pub CategoryByThread get(category_by_thread) config(): map T::ThreadId => T::CategoryId;
 
         /// Thread identifier value to be used for next Thread in threadById.
         pub NextThreadId get(next_thread_id) config(): T::ThreadId;
@@ -769,14 +772,14 @@ decl_module! {
             Self::ensure_is_forum_user(origin, &forum_user_id)?;
 
             // Ensure forum user is author of the thread
-            Self::ensure_is_thread_author(&thread_id, &forum_user_id)?;
+            let thread = Self::ensure_is_thread_author(&thread_id, &forum_user_id)?;
 
             // Store the event
             Self::deposit_event(RawEvent::ThreadTitleUpdated(thread_id));
 
             // Update thread title
             let title_hash = T::calculate_hash(&new_title);
-            <ThreadById<T>>::mutate(thread_id, |thread| thread.title_hash = title_hash);
+            <ThreadById<T>>::mutate(thread.category_id, thread_id, |thread| thread.title_hash = title_hash);
 
             Ok(())
         }
@@ -811,7 +814,7 @@ decl_module! {
                 .collect();
 
             // Update thread with one object
-            <ThreadById<T>>::mutate(thread_id, |value| {
+            <ThreadById<T>>::mutate(thread.category_id, thread_id, |value| {
                 *value = Thread {
                     poll: Some( Poll {
                         poll_alternatives: new_poll_alternatives,
@@ -1031,7 +1034,10 @@ impl<T: Trait> Module<T> {
         };
 
         // Store thread
-        <ThreadById<T>>::mutate(new_thread_id, |value| *value = new_thread.clone());
+        <ThreadById<T>>::mutate(category_id, new_thread_id, |value| {
+            *value = new_thread.clone()
+        });
+        <CategoryByThread<T>>::mutate(new_thread_id, |value| *value = category_id);
 
         // Update next thread id
         <NextThreadId<T>>::mutate(|n| *n += One::one());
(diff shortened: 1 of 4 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -598,6 +616,9 @@ decl_module! {
                 ERROR_CATEGORY_DOES_NOT_EXIST
             );
 
+            // TODO: fix this:
+            // &who belongs to the lead that originated transaction
+            // moderator_id is separate account with no way to recover its accountId; yet required by runtime's `is_moderator`
```
ondratra: `set_moderator_category(origin, moderator_id, ...)` in this function lead user can set a moderator for a category. There might be a bad implementation. 

Now it first checks if `origin` is provided by lead and then controls `is_moderator(account_id, moderator_id)` which will always return false when `moderator_id` is not belonging to some lead user (because `account_id` belongs to the lead). There is no way to retrieve `account_id` for `moderator_id` now.

bedeho: oh yes, that is a bug, there is no need to check anything about the moderator.

runtime-modules/forum/src/lib.rs
```diff
@@ -769,14 +775,34 @@ decl_module! {
...
+            Self::ensure_data_migration_done()?;
+
+            let thread = Self::ensure_is_thread_moderator(origin, &moderator_id, &thread_id)?;
+
+            // Delete thread
+            <ThreadById<T>>::remove(thread.category_id, thread_id);
+            <CategoryByThread<T>>::remove(thread_id);
+
+            // Delete posts
+            <PostById<T>>::iter_prefix_values(thread_id).map(|post_id| <ThreadByPost<T>>::remove(post_id));
```
ondratra: This line won't compile now. `DoubleMap` has feature [iter_prefix_values](https://crates.parity.io/frame_support/storage/trait.StorageDoubleMap.html#tymethod.iter_prefix_values) (called `iter_prefix` in the past) which is exactly what we need here, but we are using very old version of Substrate that has no such feature.

I am not sure if there is any easy way to overcome this. Is there any Substrate version upgrade planned? This problem is a blocker for this issue.

bedeho: There are a number of things wrong here:

1) The signature of method must be altered to have user provide thread _and_ category ids.
2) `CategoryByThread` should not be there.
3) `PostById` involves iteration, this is not safe to do, as an attacker can grow set of posts in a thread so long as to make it impossible to have the computation complete in a timely manner. This is precisely why the underlying issue states the goal as `efficient removal of an entire thread`. This means constant time, and the correct operation for this is `remove_prefix`, which you use below.

runtime-modules/forum/src/lib.rs
```diff
@@ -503,13 +503,19 @@ decl_storage! {
         pub NextCategoryId get(next_category_id) config(): T::CategoryId;
 
         /// Map thread identifier to corresponding thread.
-        pub ThreadById get(thread_by_id) config(): map T::ThreadId => Thread<T::ForumUserId, T::CategoryId, T::Moment, T::Hash>;
+        pub ThreadById get(thread_by_id) config(): double_map T::CategoryId, blake2_256(T::ThreadId) => Thread<T::ForumUserId, T::CategoryId, T::Moment, T::Hash>;
+
+        // Map category to thread. Enables searching for threads in ThreadById without knowing to which category thread belongs to.
+        pub CategoryByThread get(category_by_thread) config(): map T::ThreadId => T::CategoryId;
```
bedeho: Why was `CategoryByThread get(category_by_thread) config(): map T::ThreadId => T::CategoryId` introduced? a thread already explicitly says what category it belongs to by `Thread::category_id`

runtime-modules/forum/src/lib.rs
```diff
@@ -503,13 +503,19 @@ decl_storage! {
...
 
         /// Thread identifier value to be used for next Thread in threadById.
         pub NextThreadId get(next_thread_id) config(): T::ThreadId;
 
         /// Map post identifier to corresponding post.
-        pub PostById get(post_by_id) config(): map T::PostId => Post<T::ForumUserId, T::ThreadId, T::Hash>;
+        pub PostById get(post_by_id) config(): double_map T::ThreadId, blake2_256(T::PostId) => Post<T::ForumUserId, T::ThreadId, T::Hash>;
+
+        // Map thread to post. Enables searching for posts in PostById without knowing to which thread post belongs to.
+        pub ThreadByPost get(thread_by_post) config(): map T::PostId => T::ThreadId;
```
bedeho: Why was `ThreadByPost get(thread_by_post) config(): map T::PostId => T::ThreadId` introduced? a post already explicitly says what thread it belongs to by `Post::thread_id`.

runtime-modules/forum/src/lib.rs
bedeho: I left you a review to respond to related points, but going forward, please do not use reviews as a way of raising a point of discussion. Reviews are for one purpose only, for a maintainer to review and sign off on the quality of code. For general discussion, either just make a comment in the PR or a standalone issue, or just get in touch the relevant maintainer outside of Github.

runtime-modules/forum/src/lib.rs
```diff
@@ -1224,6 +1258,22 @@ impl<T: Trait> Module<T> {
         Ok(())
     }
 
+    // Ensure moderator can manipulate thread.
+    fn ensure_is_thread_moderator(
```
bedeho: Nice construct, perhaps rename to `ensure_can_moderate_thread` for better semantic accuracy.

runtime-modules/forum/src/lib.rs
bedeho: If the problem is that you are unable to look up a thread by only its ID after the introduction of the double map, then the proper solution is to require the user to provide a combination of the thread id and category id when trying to fully identify a thread.

runtime-modules/forum/src/lib.rs
bedeho: If the problem is that you are unable to lookup a post by only its ID after the introduction of the double map, then the proper solution is to require the user to provide a combination of the post id and thread id when trying to fully identify a thread.

bedeho (changes requested):

ondratra:
The purpose of `ThreadByPost` and `CategoryByThread` was to provide two way lookup between Category <-> Thread <-> Post. Without it, a user needs to provide all `post_id`, `thread_id`, and `category_id` when, for example, editing a post. That's because editing posts can only be done when thread belongs to a category that is mutable (e.g., not archived, moderated, etc.), and we need to check that. 

Since it's ok to require all of that, as you mentioned in a review, I reworked that. All issues should be addressed now.

bedeho (approved):
LGTM.

PR #883, opened 2020-07-03
Forum move threads between categories

ondratra:
Solves subtask of #766:
- Allow moderators to move threads between categories, if they have power in both source and destination.

To minimize merge conflict potential, this PR assumes #882 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -1316,16 +1316,31 @@ impl<T: Trait> Module<T> {
         ensure!(category_id != new_category_id, ERROR_THREAD_MOVE_INVALID,);
 
         let info = Self::ensure_can_moderate_thread(origin, moderator_id, category_id, thread_id);
-        if info.is_err() {
-            return Err(ERROR_MODERATOR_MODERATE_ORIGIN_CATEGORY);
-        }
-        let (account_id, thread) = info.unwrap();
 
+        match info {
+            Err(_) => Err(ERROR_MODERATOR_MODERATE_ORIGIN_CATEGORY),
+            Ok((account_id, thread)) => {
+                Self::ensure_can_move_thread_to_destination(
+                    &account_id,
+                    moderator_id,
+                    new_category_id,
+                )?;
+
+                Ok((account_id, thread))
+            }
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -1257,19 +1283,53 @@ impl<T: Trait> Module<T> {
...
+        ),
+        &'static str,
+    > {
+        ensure!(category_id != new_category_id, ERROR_THREAD_MOVE_INVALID,);
+
+        let info = Self::ensure_can_moderate_thread(origin, moderator_id, category_id, thread_id);
+        if info.is_err() {
+            return Err(ERROR_MODERATOR_MODERATE_ORIGIN_CATEGORY);
+        }
+        let (account_id, thread) = info.unwrap();
```
bedeho: This is not statically panic-free, but it can be, ill leave it for you to find the best way.

bedeho (changes requested):
Almost there, otherwise good.

runtime-modules/forum/src/lib.rs
```diff
@@ -1316,16 +1316,31 @@ impl<T: Trait> Module<T> {
         ensure!(category_id != new_category_id, ERROR_THREAD_MOVE_INVALID,);
 
         let info = Self::ensure_can_moderate_thread(origin, moderator_id, category_id, thread_id);
-        if info.is_err() {
-            return Err(ERROR_MODERATOR_MODERATE_ORIGIN_CATEGORY);
+
+        match info {
```
bedeho: This is indeed safe, but its a bit hard to follow, try this

```Rust
let (account_id, thread) = Self::ensure_can_moderate_thread(..)
.map_err(ERROR_MODERATOR_MODERATE_ORIGIN_CATEGORY)?;

Self::ensure_can_moderate_category(..)
.map_err(ERROR_MODERATOR_MODERATE_DESTINATION_CATEGORY)?;

Ok(..)
```

bedeho (changes requested):

runtime-modules/forum/src/lib.rs
ondratra: Much better, thx.

bedeho (approved):
LGTM.

PR #884, opened 2020-07-03
Pioneer javascript playground inject joy types

mnaamani:
I'm finding the javascript toolbox quite useful for prototyping qurey and even tx submission code.

For examples: https://github.com/Joystream/joystream/pull/877#issuecomment-653601853

The sandbox allows us to create custom types through `types.createType(...)` but we only get instances and cannot access the core types to take advantage any helper static methods.

With this PR I added can access them on `joy.types`

so you can generate a content id for example like so:

```js
const new_id = joy.types.ContentId.generate()
```

Ttried to expose all the types but due to some name clashes  types from hiring and content-working-group are not exposed.
Renamed exported Stake type to ElectionStake from council types.

### DIFF
--- a/pioneer/packages/app-js/src/Playground.tsx
+++ b/pioneer/packages/app-js/src/Playground.tsx
@@ -41,8 +41,8 @@ interface Injected {
   util: typeof util;
   window: null;
   joy: {
-    types: typeof joy_types
-  }
+    types: typeof joy_types;
+  };
 }
 
 interface Props extends AppProps, I18nProps, RouteComponentProps<{}> {
--- a/pioneer/packages/app-js/src/Playground.tsx
+++ b/pioneer/packages/app-js/src/Playground.tsx
@@ -18,6 +18,7 @@ import uiKeyring from '@polkadot/ui-keyring';
 import * as types from '@polkadot/types';
 import * as util from '@polkadot/util';
 import * as hashing from '@polkadot/util-crypto';
+import * as joy_types from '@joystream/types';
 
 import makeWrapper from './snippets/wrapping';
 import allSnippets from './snippets';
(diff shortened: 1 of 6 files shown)


Lezek123:
Looks good overall.

Created a PR to your branch with a proposed solution for the types name clashes.

We can export the types preserving their "namespace" and use them like this:
`const id = joy.types.hiring.OpeningId(1);`

The drawback is that we then need to remember which namespace the type comes from, but this may be still a good tradeoff.

Lezek123 (changes requested):
https://github.com/mnaamani/joystream/pull/5

mnaamani:
Although this PR was meant really just to make types available in the javascript tool, I suppose it is making introducing a new interface, maybe for the better. Users no longer need to import types by using the long paths, and potentially in a bad order, possibly causing circular deps in the app itself?

However if user will depend on this new interface to access the types eg.:

```
import joystreamTypes from '@joystream/types'

const { DataObject } = joystreamTypes.media

// init api then register types
joystreamTypes.registerJoystreamTypes()

```
Then we need to insure all the types are available on that first module namespace. For a lot of the module groups it is the case, there is one index.ts with everything defined on it. In some others like the `content-working-group` the ChannelId is not exported on the index.ts and one would still have to import it with

 `import ChannelId from '@joystream/types/content-working-group/ChannelId`

mnaamani:
> The drawback is that we then need to remember which namespace the type comes from, but this may be still a good tradeoff.

Its not really any different to what we do when using the long import paths in the IDE, although we have intellisense there to help a bit, so its an acceptable trade-off.

mnaamani:
> [mnaamani#5](https://github.com/mnaamani/joystream/pull/5)

merged

Lezek123 (approved):

PR #886, opened 2020-07-03
Forum category depth

ondratra:
Solves subtasks of #766:
- Make maximum category depth parameter getable through runtime trait parameters, remove from storage: MaxCategoryDepth
- Drop set_max_category_depth and the value in storage, its not safe. Instead make it a parameter in the runtime trait.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -274,6 +274,17 @@ pub trait Trait: system::Trait + timestamp::Trait + Sized {
         + From<u64>
         + Into<u64>;
 
+    type MaxCategoryDepth: Parameter
+        + Member
+        + SimpleArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + PartialEq
+        + From<u64>
+        + Into<u64>;
+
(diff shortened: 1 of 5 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -282,6 +293,7 @@ pub trait Trait: system::Trait + timestamp::Trait + Sized {
     fn is_moderator(account_id: &Self::AccountId, moderator_id: &Self::ModeratorId) -> bool;
 
     fn calculate_hash(text: &[u8]) -> Self::Hash;
+    fn get_max_category_depth() -> Self::MaxCategoryDepth;
```
bedeho: This is one way to do it, but there is a native concept of parameters which is even more safe in that it statically guarantees that its a constant value, and it also works well with the runtime metadata system and Polkadotjs. Here is an example in the proposal system

https://github.com/Joystream/joystream/blob/master/runtime-modules/proposals/codex/src/lib.rs#L168

bedeho (changes requested):

bedeho:
Please fix merge conflict.

bedeho (approved):

PR #888, opened 2020-07-03
forum - rename of set_moderator_category

ondratra:
Solves subtasks of #766:
- Rename set_moderator_category to update_category_membership_of_moderator


bedeho (approved):
LGTM.

PR #891, opened 2020-07-03
ESLint and Code Style Standardization

mnaamani:
Attempt to standardize on how we do es linting in the monorepo and apply code style for js/ts packages

- Reusable eslint, prettier and tsconfig config package
- Update the workspaces' eslintc file to use @joystream/eslint-config (while still extending) - change to use `.eslintrc.js` not `.eslintrc`
- Add/Update npm scripts to execute linting (on master for now cli linter is run with the posttest script, there are a lot of linter errors currently)
- Change network-tests workspace to use eslint, currently it uses tslint
- Test prettier - make sure eslint and prettier play nice
- testing in VS Code

Apply linting and prettier fixes will be responsibility of maintainer of each package.

Some notes/guidlines on

**code style and formatting**
There is now one prettier config at the root which should be used by all projects. So you should avoid adding any prettier or editorconfig files anywhere else (prettier will traverse from the starting point of where you run prettier and stop at the first config file it finds)
All style related eslint rules are disabled, and the prettier plugin is configured to make style/format errors into linter errors so you can fix your code style in the IDE when you have eslint plugin enabled. (I think it means you don't need to explicitly also have the prettier extension enabled in the IDE).

If you enable "format on save" setting in your IDE with prettier plugin enabled, it should work. You probably want to re-format and fix linter warnings in one PR at some point before adding more code per project to make PRs simpler to review.

`yarn prettier . --write` 

will format all the files in the current directory.

**Linting:**
We should all now use eslint and there is a config file at the root with a slim set of rules. Basically just the recommended.
The parser used is `@typescript-eslint/parser` however the root package.json doesn't add it explicitly for now, so we will get the version that comes with `@polkadot/dev`, the newer version is breaking pioneer. So will have to bump it when/if we fix all the linter errors there first.

Each project should have its own `.eslintrc.js` file that will effectively inherit the root config (unless you specify root:true in your config. Please try to avoid doing that, otherwise the whole point of standardizing on eslint rules is pointless). You can still of course override some rules that are important for the specific project type, and add any specific plugins, env configurations appropriate, for example `env: { browser: true }` if its a web app.

There is a now a "blank" tsconfig at the root (only because eslint was complaining about it missing, I presume because of the eslint config file at the root that expects a ts project) I originally added a reusable tsconfig package that can be used and extended if required but as discussed with @Gamaranto and @kdembler typescript compiler configurations are usually so project specific its not very useful and maybe even a hindrance, so I dropped it. It also doesn't make sense since the root of the monorepo is not a ts project.

Currently the only linting check done as part of CI are for pioneer. Once code in all projects is lint and style fixes we can enable them.


devops/eslint-config/index.js
```diff
@@ -0,0 +1,40 @@
...
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true
+    },
+    ecmaVersion: 2019,
+    sourceType: 'module'
+  },
+  extends: [
+    'plugin:react/recommended',
+    'standard'
```
kdembler: `standard` is a styleguide package, we probably should use `eslint:recommended` instead

kdembler: Or alongside if you intended to use `standard`

mnaamani: If I include `standard` as first rule-set, followed by `eslint:recommended`, we get benefit of both with the eslint:recommended rules taking precedence over standard?

mnaamani: I'm actually going to remove any styleguide packages, because we need prettier for consistent code formatting.

kdembler: > If I include standard as first rule-set, followed by eslint:recommended, we get benefit of both with the eslint:recommended rules taking precedence over standard?

Yes, that's my understanding.

> I'm actually going to remove any styleguide packages, because we need prettier for consistent code formatting.

This makes sense, since prettier eslint plugin will disable any style rules anyway.

fcescob: Wouldn't `prettier eslint plugin` only disable the rules conflicting with prettier? So we would not get _all_ the benefits of the style guide, but still keep relevant checks like `eqeqeq`

tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+
```
fcescob, line 2: Would it be sensible to have some defaults here,  like:

```javascript
{
  "moduleResolution": "node"
}
```

It is hard to do such a general `tsconfig` file, but if left empty it will not contribute to any changes in configuration speed.

devops/prettier-config/index.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  singleQuote: true,
+  arrowParens: 'avoid',
```
fcescob: This one is a matter of personal preference (I prefer no parens in JS), however in TypeScript you are forced to add them back since you can't type an arrow function without parenthesis:

```javascript
// This won't work
let addOne = n:number => n + 1

let addOne = (n: number) => n + 1
```

devops/eslint-config/index.js
kdembler: Check like `eqeqeq` would be inside `eslint:recommended`, not `standard`. We could keep `standard` however, since looking at the rules it enables (https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json), some may not be covered by `prettier` since they aren't all code formatting related.

devops/prettier-config/index.js
kdembler: I agree with Francesco here, I prefer no parens in JS as well, but with Typescript I believe it's more convenient to have them since it's easier to add types later on. That's also why Prettier changed default value to `always` with 2.0 release, some reasoning here: https://prettier.io/docs/en/options.html#arrow-function-parentheses

mnaamani: Very reasonable argument, I will keep the entry and update it to 'always' to be explicit :+1:

devops/eslint-config/index.js
mnaamani: yes also I think standard works better for vanilla javascript. Its the standardjs guide if I'm not mistaken

devops/prettier-config/index.js
mnaamani: fixed in 60e74961a5973da577257a64532946aa4830bde9

tsconfig.json
mnaamani: I agree, I was not intending on adding anything at all really, because the root of the monorepo is not itself a ts project.
But for some reason VSCode was complaining about missing tsconfig.json - maybe via the linter.
I think a tsconfig file without specifying any source folders also is invalid. So as discussed I think its best to leave this blank for now.

fcescob: One possible trick would be to use `tsconfig.base.json` so TS and VSCode don't pick up on it but you can use it to extend your own tsconfig, but not sure it is worth it for our usecase

mnaamani: Yes I toyed with a reusable package earlier but dropped it.

That error I saw was only in pioneer/ so I think the culprit is this line, which seems un-necessary:

https://github.com/Joystream/joystream/blob/5700ce86922e2bbd34cab68488d603b5c06b7068/pioneer/.eslintrc.js#L9

commenting out the `projects` array seems to resolve it, and pioneer still builds.

any ideas on this @Lezek123

mnaamani:
Hmm seems despite having `root: true` in the `pioneer/.eslinrc.js` linting of pioneer picked up some rules from the config at the root and broke:

https://github.com/Joystream/joystream/pull/891/checks?check_run_id=844987807#step:4:1023

Unless this is a result of some `eslint-config-*` packages being bumped ... linting is hard

tsconfig.json
mnaamani: Removed it in 49daa93cc2c3e735aa5859835a999254c667741d

mnaamani:
> Hmm seems despite having `root: true` in the `pioneer/.eslinrc.js` linting of pioneer picked up some rules from the config at the root and broke:
> 
> https://github.com/Joystream/joystream/pull/891/checks?check_run_id=844987807#step:4:1023
> 
> Unless this is a result of some `eslint-config-*` packages being bumped ... linting is hard

Fixed by reverting to older version of `"eslint-plugin-react-hooks": "^2.3.0",` in 83a5cac6f25e9882378f73567e3179dd25cddd64

kdembler:
> Fixed by reverting to older version of "eslint-plugin-react-hooks": "^2.3.0", in 83a5cac

@mnaamani I'm not sure whether we should explicitly use an older version of a linting package at the root level. If a newer version breaks Pioneer then maybe we should overwrite this version in Pioneer instead? This way other apps can benefit from the more recent linting rules?

mnaamani:
> > Fixed by reverting to older version of "eslint-plugin-react-hooks": "^2.3.0", in [83a5cac](https://github.com/Joystream/joystream/commit/83a5cac6f25e9882378f73567e3179dd25cddd64)
> 
> @mnaamani I'm not sure whether we should explicitly use an older version of a linting package at the root level. If a newer version breaks Pioneer then maybe we should overwrite this version in Pioneer instead? This way other apps can benefit from the more recent linting rules?

Yes lets do that in nicaea branch, I don't want to touch pioneer code on master branch in this PR.
I know that atlas is using newer version of both this package as well as the `'@typescript-eslint/parser'` which also breaks pioneer. So we still have some tweaking to do.
We probably also need to normalize which version of eslint itself is used, if different packages are selecting different version.

tsconfig.json
mnaamani: Well that was a FAIL on my side!

```
Error: Error while loading rule '@typescript-eslint/await-thenable': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Occurred while linting /home/runner/work/joystream/joystream/pioneer/babel.config.js
```

IDE was happy, didn't check linter though.. will revert

Lezek123:
Tested with the current CLI. Most errors I got were just related to formatting and could be easily fixed with `--fix`.

Out of those that remained I think 2 rules could be a little more problematic in our TS projects:
- `@typescript-eslint/camelcase` - I remember having to disable this rule in Pioneer due to `@joystream/types` using snake-case for struct property names in order to be consistent with the runtime. I think I was also running into some issues when they were not consistent, but will try to reproduce to be sure. Even if we can make it work, this could force a lot of changes in projects that rely on `@joystream/types`.
- `no-console` - this rule is not suitable for all projects, will not work very well for the CLI and other Node.js projects (which is also described in the documatation at: https://eslint.org/docs/rules/no-console)

Lezek123:
I noticed that they are also some `prettier/prettier` rules that I don't think any of us were following before:

- `semi: false` - looks like `network-tests`, Atlas, Pioneer and CLI - all use semicolons, which I think makes the code more clear, so perhaps this should be set to `true`?
- `trailingComma: 'es5'` - here it seems like `none` is more commonly "agreed on" among our projects, this is also forced by some eslint rule in Pioneer (I think https://eslint.org/docs/rules/comma-dangle)
_Edit: Actually the second one has some advantages and only causes problems with very old browsers, like IE8, so perhaps it's worthing keeping_

mnaamani:
re: `@typescript-eslint/camelcase` and `no-console` rules
I think those two are perfectly legitimate to be overridden in the individual projects as required. There will always be rules in this shared config that don't apply well. If we find we are always overriding them, we can do it in the shared config.

I didn't realize camelcase also applied to object properties, I would have expected it to only apply to variables. But I can see how even that can get in the way if you want to construct object literals with shorthand:

```javascript
let some_field_a, some_field_b
let some_runtime_struct = {
  some_field_a,
  some_field_b
}
```

mnaamani:
>     * `semi: false` - looks like `network-tests`, Atlas, Pioneer and CLI - all use semicolons, which I think makes the code more clear, so perhaps this should be set to `true`?

:) we sort of came to this decision in our unofficial 'code style' committee meeting with @Gamaranto @kdembler and myself, where basically Klaudiusz and I were the majority. This is similar to tabs vs spaces argument where no team will ever agree, its just important to be consistent. So I can see how we are breaking the consistency rule here since most of the projects are already using semis, which Francesco did point out.

But in all seriousness, with the recommended eslinting rules, I'm assuming the linter will help us catch bugs that could result from unexpected semicolon insertion. If I'm mistaken, then I would be happy to set this rule to true.

Regarding pioneer I was concerned with actually changing formatting too much so we don't get a nightmare when we pull from upstream, but at the same time I don't want pioneer code style to hold us back. We can ofcourse override that rule in the pioneer eslintrc.js and introduce a prettier config just for pioneer to isolate it further.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
kdembler: Won't checking this file into the repo cause everybody to override it and commit (by mistake for example) their personal preferences?

fcescob: I think personal settings are set in the global `settings.json` and then VSCode takes care of mergin those with the workspace settings, so unless someone actively makes changes to this file, it shouldn't change.

kdembler: But people (me for example) may have workspace-specific settings they want to use.

package.json
```diff
@@ -1,38 +1,46 @@
...
+    "@polkadot/api-contract": "^0.96.1",
+    "@polkadot/keyring": "^1.7.0-beta.5",
+    "@polkadot/types": "^0.96.1",
+    "@polkadot/util": "^1.7.0-beta.5",
+    "@polkadot/util-crypto": "^1.7.0-beta.5",
+    "babel-core": "^7.0.0-bridge.0",
+    "typescript": "^3.7.2"
+  },
+  "devDependencies": {
+    "eslint-config-prettier": "^6.11.0",
```
kdembler: Shouldn't these be the dependencies of `@joystream/eslint-config` package instead?

kdembler: I'm still little lost in package management in monorepo with workspaces so please let me know if I'm wrong

devops/eslint-config/index.js
```diff
@@ -0,0 +1,54 @@
...
+    // pioneer has an isolated linting config so its safe to enable here.
+    // Lets try to support an accessible web in future
+    // 'plugin:jsx-a11y/recommended',
+
+    // Turns off all rules that are unnecessary or might conflict with Prettier.
+    // Allows us to do formatting separately from linting.
+    'prettier',
+    'prettier/@typescript-eslint',
+    'prettier/react',
+    'prettier/standard',
```
kdembler, line 31: For this to work we should add `standard` above as well

devops/eslint-config/index.js
```diff
@@ -0,0 +1,54 @@
...
+  settings: {
+    react: {
+      version: 'detect',
+    },
+  },
+  rules: {
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
+  },
+  plugins: ['@typescript-eslint', 'react', 'react-hooks'],
```
kdembler: Seems we should add` standard` here as well

package.json
```diff
@@ -1,38 +1,46 @@
...
+    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-react": "^7.16.0",
+    "eslint-plugin-react-hooks": "^2.3.0",
+    "husky": "^4.2.5",
+    "prettier": "2.0.2"
```
kdembler: This list lacks `standard` and its peer dependencies, full list: `eslint-config-standard eslint-plugin-standard eslint-plugin-promise eslint-plugin-import eslint-plugin-node`

.vscode/settings.json
mnaamani: Francesco is correct, the workspace settings checked in here will only override your global settings.
So you personal preferences can go into the `User` global settings.
If you have some good workspace settings then the rest of the team could benefit also by adding them here. What do you think?

package.json
mnaamani: Yes I think that would be neater, will try it out

package.json
mnaamani: noted

devops/eslint-config/index.js
mnaamani: Did you find the best placement for it in the array. Shall I make it the last one before `prettier` ?

mnaamani: For context.. this is to introduce the `eqeqeq` rule

kdembler: The example Prettier config has it at the very top

.vscode/settings.json
kdembler: > Francesco is correct, the workspace settings checked in here will only override your global settings.

Well, I'm not sure that's a good thing. For example, someone (from the team or from the outside) may not like formatting on save for some reason but they would be forced to either use it or keep enabling and disabling it to not check it into source control. I believe IDE settings are a pretty personal thing and it's easier if everybody just does them for themselves. Enabling format on save for a team member is a 10-second task and we wouldn't need to worry about someone committing their changes, different preferences, etc.

I don't use VSCode myself most of the time so I don't have that strong of an opinion but I believe it's a good practice to keep specific IDEs out of the source.

mnaamani:
I've been testing eslint plugin in VSCode, with the current setup it seems to be working well but with a couple of issues:

1 - There is still this issue with the plugin complaining about a missing tsconfig.json at the root (there is a an empty file now just to make the plugin actually work) - Francesco pointed at a possible solution here https://github.com/microsoft/vscode-eslint/issues/455

2 - I'm not 100% sure the rules the plugin is applying in the IDE are identical to when eslint is run on the commandline becase I had to add ` 'no-console': 'off'` explicitly for the rule to work in VSCode, but it wasn't required on the commandline. Hopefully other rules are applied consistently otherwise we are gonna have a frustrating experience!

I tried the eslint plugin in `Atom` but unfortunately it [doesn't support](https://github.com/AtomLinter/linter-eslint/issues/1340) `.eslintrc.js` it only looks for older `.eslintrc` which is a yaml file which sucks. We don't want to loose benefits of the .js file format so unless anyone is adamant about using Atom over VSCode will not try to fix.

devops/eslint-config/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // Turns off all rules that are unnecessary or might conflict with Prettier.
+    // Allows us to do formatting separately from linting.
+    'prettier',
+    'prettier/@typescript-eslint',
+    'prettier/react',
+    'prettier/standard',
+    // To Display prettier errors as ESLint errors, enable the following configuration,
+    // And make sure it is the last configuration in the extends array.
+    'plugin:prettier/recommended',
```
kdembler: Just noticed, in example config by Prettier this is actually above all other `prettier` rules (e.g. `prettier/react`, etc.)

devops/eslint-config/index.js
```diff
@@ -0,0 +1,50 @@
...
+    react: {
+      version: 'detect',
+    },
+  },
+  rules: {
+    // drop these when using newer versions of eslint-plugin-react-hooks
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
+  },
+  plugins: ['standard', '@typescript-eslint', 'react', 'react-hooks'],
```
kdembler: I think we should also add `prettier` here, per https://github.com/prettier/eslint-config-prettier#example-configuration

devops/eslint-config/index.js
kdembler: The docs say this line:
> - Enables eslint-plugin-prettier.
> - Sets the prettier/prettier rule to "error".
> - Extends the eslint-config-prettier configuration.

So it may be a replacement for `prettier` above

mnaamani: Yes I saw the example. I was basing the ordering on this [article](https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project)

The way I understand it is that `eslint-plugin-prettier` actually runs prettier itself and reports back the formatting errors based on your prettierrc config. So it seems natural to put it last. Perhaps the order of the plugin itself is not important because its not actually modifying any rules just executing and that happens after all the rules have been configured. I did test the behavior on both the command line and in the IDE and I got the expected results. But I guess it makes sense to follow the maintainers documentation :)

mnaamani: I asked out of curiosity https://github.com/prettier/eslint-plugin-prettier/issues/314

devops/eslint-config/index.js
mnaamani: I think having `'plugin:prettier/recommended'` in the extends array adds it for us, just as your hunch that it also:

> So it may be a replacement for prettier above

Based on this code: https://github.com/prettier/eslint-plugin-prettier/blob/master/eslint-plugin-prettier.js#L109

I added it anyway.

.vscode/settings.json
mnaamani: As I was initially hesitant to add this in the first place and your last argument, decided to revert it https://github.com/Joystream/joystream/pull/891/commits/ad88e6550203b8df508199f50f6365fd052d81e1

mnaamani: Happy to report the issue with eslint plugin in VSCode looking for tsconfig.json at the root is resolved by adding the following in the workspace settings.json:

```json
  "eslint.workingDirectories": [
    "./cli",
    "./pioneer",
    "./tests/network-tests",
    "./types"
  ]
```
as advised by @Gamaranto 

Is this a legit reason to keep .vscode/settings.json in the repo?

Alternative would be to just have a script that injects these settings into the workspace.

devops/eslint-config/index.js
mnaamani: Okay so I got an answer which makes sense as far as rules go, but didn't satisfy my curiosity about when prettier runs.
So I did a bit of debugging and can confirm that prettier runs (the `create()` method is called on the plugin) after all the rules in the extends array processed.

mnaamani:
> 2 - I'm not 100% sure the rules the plugin is applying in the IDE are identical to when eslint is run on the commandline becase I had to add ` 'no-console': 'off'` explicitly for the rule to work in VSCode, but it wasn't required on the commandline. Hopefully other rules are applied consistently otherwise we are gonna have a frustrating experience!

Somehow after re-testing this, I'm now seeing consistency between the IDE and the commandline linting results.

devops/eslint-config/index.js
kdembler: Awesome, good to know

kdembler (approved):
LGTM!

PR #900, opened 2020-07-06
Fix storage-node linter issues

### LINKED ISSUE #575: Update the storage-node project.
### Update the storage-node

1. Autofix linter warnings.
2. Fix other linter warnings (or add ignore rules).
3. Provide initial migration to the TypeScript using 'allow JS mode' (https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html)

'Atlas-committee' linter rules:

```
// .eslintrc.js
module.exports = {
// Example for a browser based project, if it is on the server, add node: true instead of browser: true
  env: {
    browser: true,
    es6: true,
  },
  globals: {
    Atomics: "readonly",
    SharedArrayBuffer: "readonly",
  },
// if using TS, otherwise leave blank
  parser: "@typescript-eslint/parser",
    ecmaVersion: 2019,
    sourceType: "module",
  },
  extends: [
    "esnext",
    "esnext/style-guide"
    "plugin:prettier/recommended",
  ],
};

//.prettierrc
{
	"semi": false,
	"trailingComma": "es5",
	"singleQuote": true
}


```

shamil-gadelshin:
Changes:
- introduced linter rules list 
- added prettier to the eslint
- fix numerous warnings.


storage-node/.eslintrc.js
```diff
@@ -1,290 +1,41 @@
...
-        "wrap-regex": "error",
-        "yield-star-spacing": "error",
-        "yoda": [
-            "error",
-            "never"
-        ]
-    }
+    extends: [
+        "esnext",
+        "esnext/style-guide",
```
mnaamani, line 13: With introduction of #891 We shouldn't add a unique code style for this project.
Also in general you will have a hard time if you add style rules, if they don't match with your prettier config, its best to drop all style rules, see the approach I took in #891, so all style comes purely from prettier.

I would recommend incorporating the above PR and dropping this `extends` array.
You will then need a few extra rule overrides during transition to TS:
```
    '@typescript-eslint/no-var-requires': 'warning',
    '@typescript-eslint/camelcase': 'warning',
```
You will then find a few more linter errors, most are easy to fix.

mnaamani: There is one error that didn't make a lot of sense, it just said delete the `.` in:

```javascript
module.exports = function (x,y) {

}
```
The problem is that its an anonymous function, giving it a name seems to make linter happy.

storage-node/.eslintrc.js
```diff
@@ -1,290 +1,41 @@
...
+				"**/test/fs/walk.js",
+				"**/test/storage.js",
+				"**/test/identities.js",
+				"**/test/balances.js",
+				"**/test/assets.js",
+			],
+			"rules": {
+				// Disabling Rules because of used chai lib:
+				// https://stackoverflow.com/questions/45079454/no-unused-expressions-in-mocha-chai-unit-test-using-standardjs
+				"no-unused-expressions": "off",
```
mnaamani, line 38: The dirty-chai approach seemed neater, at least reducing config in the eslint config file, although requiring some code changes.

storage-node/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+    "semi": false,
+    "trailingComma": "es5",
+    "singleQuote": true,
+	"arrowParens": "avoid",
+	"useTabs": false,
+	"tabWidth": 2
+}
```
mnaamani, line 8: we should drop project specific prettier configs in favor of global one which will come in #891 
If its premature to merge it now, perhaps you can copy the same settings.

storage-node/.eslintrc.js
```diff
@@ -1,290 +1,41 @@
...
+    extends: [
+        "esnext",
+        "esnext/style-guide",
+        "plugin:prettier/recommended"
+    ],
+	"rules": {
+		"import/no-commonjs": "off", // remove after converting to TS.
+		// Disabling Rules because of monorepo environment:
+		// https://github.com/benmosher/eslint-plugin-import/issues/1174
+		"import/no-extraneous-dependencies": "off",
```
mnaamani, line 20: I was interested to see this. I turned the rule to error but when I run it i get:

```
 | ackages/runtime-api/4m/Users/mokhtar/joystream/joystream/storage-node/packages/runtime-api/assets.js
 |   1:1  error  Definition for rule 'import/no-extraneous-dependencies' was not found  import/no-extraneous-dependencies
 | 
```

Do we need to add `eslint-config-import` package to dev dependencies ?

mnaamani (changes requested):

mnaamani:
last point, I had to add `eslint-plugin-prettier` package to dev dependencies in root package.json as I was getting error that the plugin was not found when running the linter. Not sure why.

mnaamani:
Flawless victory, not even a single warning ;)

Other than seeing this when running lint:
`The babel/generator-star-spacing rule is deprecated. Please use the built in generator-star-spacing rule instead.`

LGTM

PR #901, opened 2020-07-06
Forum category deletion

### LINKED ISSUE #761: Forum category deletion
- Safety is enforced by requiring that the number of child categories or threads is 0, and this can be checked
- A forum moderator can delete a category that is a descendant of a category to which they are assigned, but not the direct category itself.

ondratra:
Solves subtask of #766:
- Forum category deletion: #761

To minimize merge conflict potential, this PR assumes #883 will be merged first.

Fixes #761.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -1460,12 +1460,7 @@ impl<T: Trait> Module<T> {
         actor: &PrivilegedActor<T>,
         category_id: &T::CategoryId,
     ) -> Result<Category<T::CategoryId, T::ThreadId, T::Hash>, &'static str> {
-        /*
-        let (who, is_lead, moderator_id) = match actor {
-            PrivilegedActor::Moderator(moderator_id) => (Self::ensure_is_moderator(origin, &moderator_id)?, false, moderator_id),
-            PrivilegedActor::Lead => (Self::ensure_is_forum_lead(origin)?, true, &T::ModeratorId::from(0)),
-        };
-        */
+        // Check actor's role
         match actor {
             PrivilegedActor::Lead => Self::ensure_is_forum_lead(origin)?,
             PrivilegedActor::Moderator(moderator_id) => {
@@ -1473,6 +1468,7 @@ impl<T: Trait> Module<T> {
             }
         };
 
+        // Ensure category exists
         if !<CategoryById<T>>::exists(category_id) {
             return Err(ERROR_CATEGORY_DOES_NOT_EXIST);
         }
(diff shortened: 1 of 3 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -746,6 +752,26 @@ decl_module! {
             Ok(())
         }
 
+        fn delete_category(origin, moderator_id: T::ModeratorId, category_id: T::CategoryId) -> dispatch::Result {
+            // Ensure data migration is done
+            Self::ensure_data_migration_done()?;
+
+            Self::ensure_can_delete_category(origin, &moderator_id, &category_id)?;
+
+            let category = <CategoryById<T>>::get(category_id);
```
bedeho: This is not as safe as it can be. If `ensure_can_delete_category` has been successful, then we should be able to statically guarantee that the category instance exist, rather than looking it up again in the map, which can fail at runtime.

runtime-modules/forum/src/lib.rs
```diff
@@ -1414,24 +1440,72 @@ impl<T: Trait> Module<T> {
         }
     }
 
+    fn ensure_can_delete_category(
```
bedeho, line 1458: This appears to not allow the lead to act as a normal moderator, is that right?

There is something wrong here beyond that, because you seem to be checking that the moderator possibly is the lead, which is not possible. The lead does not have a moderator id.

bedeho (changes requested):
Good start.

runtime-modules/forum/src/lib.rs
ondratra: Redudant get was removed.

runtime-modules/forum/src/lib.rs
ondratra: No only moderator can delete category right now. I only wanted to provide the possibility for lead to deleting the root category. You are right, moderator + lead checks together are invalid, I didn't realize that before.

Should I create separate function `delete_category_lead` to be used by lead only (on any category?), or should I let update the function to be able to accept both lead and moderator (via union type `delete_category(..., moderator_or_lead: UnionTypeOfModeratorAndLead)`)?

bedeho: Your last suggestion is the right approach, just make the names more sensible, something like

```Rust
enum PrivilegedActor{
Lead,
Moderator(T::ModeratorId)
}
```

Now you can reuse this type for multiple extrinsic where you may need to allow both the lead and the moderator to act.

Well done on identifying the right approach right away ;)

bedeho (approved):

bedeho:
This looks good, but for some reason, only `delete_category` can directly allow the lead to act as a moderator. I think it makes the most sense to either give the lead full moderator powers directly or none. This halfway state is not a good idea.

I have created a new standalone issue for this that we can address when we have reviewed all outstanding PRs
https://github.com/Joystream/joystream/issues/946

Issue #477, opened 2020-05-27
Video Component [Atlas]


fcescob:
Implement  a functional video component, following the design in https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=451%3A1330

PR #903, opened 2020-07-06
Add video screen, rework video player

kdembler:
### Overview

This PR reworks `VideoPlayer` component to use `video.js` library and match (at least in a reasonable degree) the Figma design. It also adds a video screen as part of #477:

Screenshot 2020-07-06 at 11 15 31 PM

### Limitations

* Currently, the video component on the video screen is poorly responsive, the size of the player is dictated by the CSS, later we can add some media queries to adapt to different screen sizes better.

### Conventions started

As part of this PR, I've added some new conventions to Atlas. For clarity, I'm listing those here. Any feedback is welcome

* I've created a top-level `config` folder which I intended for any configuration that we may need. This currently includes routes config (paths mapping) and mocked data. In the future we could hold some tweakable variables here.
* I've created a top-level `types` folder in which I've created stubs of types for `Channel` and `Video`. These aren't widely used now and sort of guidance towards unifying the field names of properties we use. Maybe we should start passing objects typed by these as component props to better prepare for future state management. At the moment fields are passed directly. Any thoughts @Gamaranto?
* I've also created `utils` folder for keeping various utility functions, currently formatting date and number
* As discussed with @Gamaranto, I've added `luxon` as a library for handling dates


atlas/packages/app/src/utils/number.ts
```diff
@@ -0,0 +1,15 @@
+export const formatNumber = (num: number): string => {
```
fcescob, line 1: If this formats a number by adding a space every 3 non zero decimal digits, then it can be shortened to this:
```javascript
const formatNumber1 = n =>
  n
    .toLocaleString('en-US')
    .split(',')
    .join(' ');
```

kdembler: Ah, I knew there has to be something like this built in but couldn't find it. Awesome, will use that instead

atlas/packages/app/src/views/VideoView/VideoView.style.tsx
```diff
@@ -0,0 +1,68 @@
...
+	font-size: ${theme.typography.sizes.h5};
+`;
+
+export const MoreVideosGrid = styled.div`
+	display: grid;
+	grid-template-columns: repeat(auto-fit, minmax(320px, 1fr));
+	grid-gap: ${theme.spacing.xl};
+`;
+
+export const MoreVideosPreview = styled(VideoPreview)`
```
fcescob, line 67: Not sure if intentional, but on the `VideoView` page I see no hover effect on the grid of video previews, while I see them on the Carousels on the landing page.

kdembler: That's because there is no `onClick` handler set for those, so the hover animations is disabled. Once we add the handler, the animation should work properly.

fcescob:
A bit out of scope, this was likely introduced in a previous PR, but I get this when building the components package:

immagine

kdembler:
Yes, I wrote a bad import statement, partly because there wasn't a clear way on how to approach those. I have this fixed on my merge branch so let's not tackle this here

bedeho:
Haven't started a proper review, but the play button in the top left corner is not something we can have? perhaps I am misunderstanding?

atlas/packages/app/src/utils/number.ts
kdembler: Fixed in aa685c0041d3153b7e53e26d975fcf4034b215cb

kdembler:
@bedeho  This is a default from `video.js` I've left it so far because the design doesn't really specify how the player looks like before the playback starts. I could center it and replace with the play icon we use in video preview component, would you be fine with that?

fcescob:
Looks good overall, I really like how you organized the code on the app package!

Regarding passing whole objects to components instead of each props singularly, we would gain a stable interface Videos and channels are expected to have, we don't need to retype each component that gets a video or a channel

However, we would lose type hints on components, and only one prop would show from outside, making it harder imho to grasp how the prop is used by the component, it also makes it impossible to pick which fields to exclude and which ones to keep, so imho this weights against this pattern.

A possible solution with the benefits of the former and without the downside of the latter would be to pick specific fields from the predifined types and use them to type components. 
For example,  this is what a `VideoTitleProps` might be look like:

```typescript
type VideoPreviewProps = Pick<Video, 'title' | 'createdAt' | 'views'> & Channel & { 
  // other types like onClick , showMeta ...
}

```

kdembler:
Yes, the point you've made is a very good one, that's also partly why I didn't go for passing the objects at the time. I'd want to avoid situations in which we have to create fake objects just to pass them as a prop.

I really like sort of composing that you've proposed, however, I can see one issue with this approach. We would be fine as long as we have props that are part of one type (for example Video with `title`, `description` and so on). The problem arises once we want to have some info about other type (for example Channel) that may have the same property names (things like `name` can be a common conflict) and I'm currently not sure how we could alleviate that.

fcescob:
Good point, that would be an issue, I found this after some googling: https://stackoverflow.com/questions/59071058/how-to-pick-and-rename-certain-keys-using-typescript so it is doable to define a type that allows to pick and rename at will.

bedeho:
> @bedeho This is a default from `video.js` I've left it so far because the design doesn't really specify how the player looks like before the playback starts. I could center it and replace with the play icon we use in video preview component, would you be fine with that?

The video player has a few states
1. Paused
2. Finished
3. Playing
4. Buffering: Insufficient data to actually playback at the given time position.
5. Loading: This is before the initial metadata information has been loaded into the player so that the player can correctly display duration, and so on. 

I would say that before playback begins, it would make the most sense to be in 5, and that should not involve a button to click for starting playback. Then perhaps it should go to state 4, and then finally to 3. Sadly, the designer did not even provide a buffering screen. I would suggest you make all of these as simple as possible, but not involve needless action from the user, nor use foreign visual elements.

We will get another designer later in the process, they can fill some of these missing pieces.

kdembler:
Okay so what I did for now was adding an autoplay and centering the button. Autoplaying is a hard thing to accomplish across different browsers though, so it will work only if you navigate to the video within the app. If the video page is the first one that you visit (through an URL), the playback won't start automatically. Fixing that is a lot of work and if we want to, should be addressed separately. We can make it autoplay muted though easily.

bedeho:
1) I am not fully able to appreciate this issue, so lets discuss the difficulty at a later time, however, please make an issue about this with the  `atlas` tag.

2) While we dont have a designer at hand for this, I think we can try to make some best-effort attempts on getting thigns consistent. The play button you are currently using is not really consistent with the general visual style, nor the only other closest single screen for showing a major action button in the video display (see `player-ended`). Try to make something closer to this.

kdembler:
1. Sure thing, I've created #934 and also supplied some more detail there

2. I'm aware, that's also something I've mentioned in one of the DMs. This is what I've come up with for when the autoplay won't work:

Screenshot 2020-07-13 at 10 44 48 PM

3. I've also changed the sizing of the player a little bit, previously it would fill the full width and add black bars _inside_ the player. After the change, player size will fit the content size so the controls are more usable:

Screenshot 2020-07-13 at 10 46 41 PM

bedeho:
Merge conflict!

kdembler:
I'm aware but I want to wait till https://github.com/Joystream/joystream/pull/914 is merged, we will need to do a rebase then anyway

kdembler:
Rebase done and conflicts resolved! @bedeho @Gamaranto

fcescob (approved):

kdembler:
@bedeho waiting for your review

Issue #617, opened 2020-06-08
CLI: Storage working group - worker actions


Lezek123:
In contracst to the "lead-only" commands, "worker-only" ones will need an additional argument, which is `worker_id`. This cannot be just derived from currently selected account, since one account can be a role account of many active workers in given group simultaneously.

There can be something similar to `promptForAccount` method in `AccountsCommandBase` implemented as part of the "worker-only" gate that will be introduced in https://github.com/Joystream/joystream/issues/598. This will allow us to prompt the user to choose one of the workers associated with the currently selected account (if there is more than 1).

There are 3 simple actions an active worker will be able to perform:
- `working-groups:updateRoleAccount --group=storage` - the associated api method is `updateWorkerRoleAccount(worker_id, new_role_account_id)`. The worker id will be estabished based on users choice if there is more than 1 worker associated to the currently selected account.
- `working-groups:updateRewardAccount --group=storage` - implemented same way as above
- `working-groups:leaveRole --group=storage` - handled via `leaveWorkerRole(worker_id, rationale_text)` api method. As with termination, rationale could be provided either via flag (ie. `r="Some Rationale"`, when prompted for or both.

Lezek123:
Done in https://github.com/Joystream/joystream/pull/908

Issue #603, opened 2020-06-08
CLI: Manage storage working group workers (lead)


Lezek123:
Here I assume the list of workers will already be available through the group overview added in https://github.com/Joystream/joystream/issues/598 but a separate command with a more detailed workers overview can be considered if it turns out to be needed.

Commands to implement:
- `working-groups:slashWorker --group=storage WORKER_ID BALANCE` - handled via `slashWorkerStake(worker_id, balance)` api method. The input is quite straight-forward so it'll probably be easiest to just provide it via args.
- `working-groups:evictWorker --group=storage WORKER_ID` - handled via `terminateWorkerRole(worker_id, rationale_text)` api method. Rationale could be provided later (when prompted for) or via a flag (ie. `-r="Some rationale here"`)
- `working-groups:updateReward WORKER_ID` - I don't think there's is an api method to do this yet, but this shouldn't be hard to implement in the CLI once we have the  `fillWorkerOpening` command implemented (since the prompting and validation related to rewards could be reused from there).

Lezek123:
Done in https://github.com/Joystream/joystream/pull/908

PR #908, opened 2020-07-07
CLI: Worker management, worker actions and updates following hireable lead upgrade

Lezek123:
This PR covers issues https://github.com/Joystream/joystream/issues/603, https://github.com/Joystream/joystream/issues/617 and https://github.com/Joystream/joystream/issues/726 (CLI-related part only).

New commands (lead):
- `working-groups:decreaseWorkerStake WORKER_ID`
- `working-groups:evictWorker WORKER_ID`
- `working-groups:slashWorker WORKER_ID`
- `working-groups:updateWorkerReward WORKER_ID`

New commands (worker / lead as worker)
- `working-groups:updateRewardAccount ADDRESS` - address is optional. In case it's not provided - user can select one of the CLI accounts.
- `working-groups:updateRoleAccount ADDRESS` - as above, but requires member controller account to be selected (if it differs from current role account)
- `working-groups:increaseStake`
- `working-groups:leaveRole`

Other changes:
- Modified the prompting methods, like `promptForParam` etc. to be more customizable, so now it's possible to provide custom names, validation, default values or forced/locked values when prompting for extrinsic parameters, which, for example, allows us to skip prompts for `SlahsingTerms` and `OpeningType` in `working-groups:createOpening` and use the "locked" default values instead.
- Modified display/overview commands like `working-groups:overview`, `working-groups:openings`, `working-groups:opening` to include information about opening/worker type


cli/src/Api.ts
```diff
@@ -280,7 +306,9 @@ export default class Api {
         let groupMembers: GroupMember[] = [];
         for (let [index, worker] of Object.entries(workers.toArray())) {
             const workerId = workerIds[parseInt(index)];
-            groupMembers.push(await this.groupMember(workerId, worker));
+            if (worker.is_active) {
```
mnaamani, line 309: It occured to me now that this method should really be renamed from `groupMembers` to `activeGroupMembers()`. and similarly `workerByWorkerId()` should be renamed to `activeWorkerByWorkerId()` because of the filtering in the former, and throwing in the latter for non "active" workers.

cli/src/Types.ts
```diff
@@ -92,19 +93,27 @@ export const AvailableGroups: readonly WorkingGroups[] = [
   WorkingGroups.StorageProviders
 ] as const;
 
+export type Reward = {
+    totalRecieved: Balance;
+    value: Balance;
+    interval: number; // 0 = reward is not really "recurring"
```
mnaamani: wouldn't be better to make interval and nextPaymentBlock optional?
```suggestion
    interval?: number;
```

or

```suggestion
    interval: number | undefined;
```

cli/src/commands/working-groups/createOpening.ts
```diff
@@ -89,7 +79,7 @@ export default class WorkingGroupsCreateOpening extends WorkingGroupsCommandBase
         else {
             await this.requestAccountDecoding(account); // Prompt for password
             this.log(chalk.white('Sending the extrinsic...'));
-            await this.sendExtrinsic(account, apiModuleByGroup[this.group], 'addOpening', defaultValues!);
+            await this.sendExtrinsic(account, apiModuleByGroup[this.group], 'addOpening', Object.values(defaultValues!));
```
mnaamani: Its not guaranteed that `Object.values()` returns the values in the correct order expected by the extrinsic.
Although convenient, in this case it could be wrong to use it here.

`The ordering of the properties is the same as that given by looping over the property values of the object manually.` - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values)

cli/src/commands/working-groups/decreaseWorkerStake.ts
```diff
@@ -0,0 +1,58 @@
...
+            'decreaseStake',
+            [
+                new WorkerId(workerId),
+                balance
+            ]
+        );
+
+        this.log(chalk.green(
+            `${chalk.white(formatBalance(balance))} from worker ${chalk.white(workerId)} stake `+
+            `has been returned to worker's role account (${chalk.white(groupMember.roleAccount.toString())})!`
```
mnaamani, line 55: Should double check but I think the stake is returned to the member's controller account

mnaamani: Hmm that's, correct actually, but not what I expected.

cli/src/commands/working-groups/startAcceptingApplications.ts
```diff
@@ -26,21 +25,19 @@ export default class WorkingGroupsStartAcceptingApplications extends WorkingGrou
         // Lead-only gate
         await this.getRequiredLead();
 
-        const opening = await this.getApi().groupOpening(this.group, parseInt(args.wgOpeningId));
-
-        if (opening.stage.status !== OpeningStatus.WaitingToBegin) {
-            this.error('This opening is not in "Waiting To Begin" stage!', { exit: ExitCodes.InvalidInput });
-        }
+        const openingId = parseInt(args.wgOpeningId);
+        // We don't need the actual opening here, so this is just for validation purposes
```
mnaamani: this comment suggests factoring out the check from getOpeningForLeadAction() into its own function  ;)

cli/src/commands/working-groups/updateRewardAccount.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (!worker.reward) {
+            this.error('There is no reward relationship associated with this role!', { exit: ExitCodes.InvalidInput });
+        }
+
+        let newRewardAccount: string = args.accountAddress;
+        if (!newRewardAccount) {
+            const accounts = await this.fetchAccounts();
+            newRewardAccount = (await this.promptForAccount(accounts, undefined, 'Choose the new reward account')).address;
+        }
+        validateAddress(newRewardAccount);
```
mnaamani, line 38: this doesn't seem necessary since we are getting it from the stored accounts?

types/src/working-group/index.ts
```diff
@@ -243,22 +240,13 @@ export class WorkingGroupOpeningPolicyCommitment extends JoyStruct<IWorkingGroup
...
-        Worker: Null
-      },
-      value, index
-    );
-  }
-};
+export class OpeningType_Leader extends Null { };
+export class OpeningType_Worker extends Null { };
+export class OpeningType extends JoyEnum({
+  Leader: OpeningType_Leader,
```
mnaamani, line 247: I prefer this naming style, so perhaps use same for UnslashableTerms, and SlashableTerms. But I wonder if the linter will not like the underscore in a class name..

mnaamani (changes requested):
I did a code review but didn't get chance to try out all the commands yet.

cli/src/Api.ts
Lezek123: Actually the only thing `is_active` does in `Worker` class is checking whether the worker is not an empty object:
```
  get is_active(): boolean {
    return !Boolean(this.isEmpty);
  }
```

`LinkageResult`s coming from `workerById` may contain empty worker object if:
- No workers are available and we try to query all workers with `workerById()` (we then get a `LinkageResult` which is `[[0], [EMPTY_WORKER]]`)
- We query worker by id with `workerById(ID)` and it doesn't exist

In case of `groupMemebrs` method, this check is probably not even necessary, because the second case is not possible and we have a condition earlier that prevents the first case:
```
        if (nextId.eq(0)) {
            return [];
        }
```

I only named this getter `is_active` for consistancy with `contentWorkingGroup`, where workers actually can either be active or not. In case of `storageWorkingGroup` - inactive workers are just removed from the map (so inactive means "unexisting")

cli/src/commands/working-groups/updateRewardAccount.ts
Lezek123: We only get it from stored accounts in case it's not provided via command arg. You can also use ie.:
`./cli/bin/run working-groups:updateRewardAccount 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY`

cli/src/commands/working-groups/startAcceptingApplications.ts
Lezek123: I created an alias method called `validateOpeningForLeadAction`, since there isn't much to factor out, but this should make the code more readable without additional comments.

cli/src/commands/working-groups/createOpening.ts
Lezek123: Fixed by replacing `defaultValues` to an array of objects

cli/src/Types.ts
Lezek123: Fixed

types/src/working-group/index.ts
Lezek123: It looks like `@typescript-eslint/camelcase` would disallow it, which is currently a deprecated rule btw (https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md).

I was just forced to come up with some different names than `Worker` and `Leader` here to avoid name conflicts for `Worker` class.

I could also use `OpeningTypeWorker` or, since both classes are just `Null`, we could just abandon the requirement for registering classes for all enum keys and use `JoyEnum({ Leader: Null, Worker: Null } as const)`,

I think that now that we have a good, TS-compatible `JoyEnum` helper, we don't need to actually use `instanceof`, we can instead just check `this.type === typeKey` in `isOfType` and `asType`, since:
- `typeKey` is enforced to be one of the registered keys by TypeScript rules
- The same keys are beeing passed to `Enum` constuctor (`@polkadot/api`) and become available in `this._def` object, which is then used in `this.type` getter

So it should be a pretty type-safe and error-resistant way still (without the need for `instanceof`).

mnaamani (approved):

PR #909, opened 2020-07-07
Forum post thread moderation

ondratra:
Solves subtasks of #766:
- Moderation semantics for posts changed to mean that it is deleted by the moderator, and that some accompanying rationale is submitted, but is not itself stored.
- Moderation semantics for threads changed in the same way as posts. This can only be done after efficient thread deletion has been implemented.

To minimize merge conflict potential, this PR assumes #886 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -554,6 +554,7 @@ decl_event!(
         <T as Trait>::ThreadId,
         <T as Trait>::PostId,
         <T as Trait>::ForumUserId,
+        <T as system::Trait>::Hash,
     {
         /// A category was introduced
         CategoryCreated(CategoryId),
@@ -581,7 +582,7 @@ decl_event!(
         PostAdded(PostId),
 
         /// Post with givne id was moderated.
-        PostModerated(PostId),
+        PostModerated(PostId, Hash),
 
         /// Post with given id had its text updated.
         /// The second argument reflects the number of total edits when the text update occurs.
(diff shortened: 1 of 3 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -846,30 +847,21 @@ decl_module! {
...
-
-            // Get thread
-            let thread = Self::ensure_thread_exists(&category_id, &thread_id)?;
+            // Ensure moderator is allowed to moderate post
+            Self::ensure_can_moderate_thread(origin, &moderator_id, &category_id, &thread_id)?;
 
-            // ensure origin can moderate category
-            Self::ensure_can_moderate_category(&who, &moderator_id, &thread.category_id)?;
+            // Calculate rationale's hash
+            let rationale_hash = T::calculate_hash(rationale.as_slice());
```
bedeho: There is no need to calculate a hash just to put it in the event. The hash only serves some productive purpose when in storage.

runtime-modules/forum/src/lib.rs
```diff
@@ -846,30 +847,21 @@ decl_module! {
...
-            // Can mutate in corresponding category
-            let path = Self::build_category_tree_path(&thread.category_id);
-
-            // Path must be non-empty, as category id is from thread in state
-            assert!(!path.is_empty());
-
-            // Path can be updated
-            Self::ensure_can_mutate_in_path_leaf(&path)?;
+            // Delete thread
+            <ThreadById<T>>::remove(category_id, thread_id);
```
bedeho: Are you sure this is updating all state properly, what about `Category.num_direct_threads`. Be aware that you should not now lookup the category based on `category_id`, because this will already have been statically guaranteed prior to all the mutation commencing.

runtime-modules/forum/src/lib.rs
```diff
@@ -951,24 +943,21 @@ decl_module! {
...
 
-            // Get moderator id.
-            let who = Self::ensure_is_moderator(origin, &moderator_id)?;
+            // Ensure moderator is allowed to moderate post
+            Self::ensure_can_moderate_post(origin, &moderator_id, &category_id, &thread_id, &post_id)?;
 
-            // Make sure post exists and is mutable
-            let post = Self::ensure_post_is_mutable(&category_id, &thread_id, &post_id)?;
+            // Calculate rationale's hash
+            let rationale_hash = T::calculate_hash(rationale.as_slice());
```
bedeho: Same comment as above.

runtime-modules/forum/src/lib.rs
bedeho: This only deletes the thread, it does not also delete all posts that live in a thread, which is in `PostById`, which should be done efficiently.

bedeho (changes requested):

ondratra:
Rationale hash has been removed, and thread deletion/moderation has been united.

ondratra:
> Be aware that you should not now lookup the category based on category_id, because this will already have been statically guaranteed prior to all the mutation commencing.

If I understood it correctly, all extrinsic should be rewritten, so they use input parameters only before mutation?

```
fn my_function(category_id, thread_id, ...) {
    let (category, thread, ...) = Self::ensure_everything_is_ok(category_id, thread_id, ...); // only usage of category_id

    //
    // == MUTATION SAFE ==
    //
    // do not use category_id after this point

    <ThreadById<T>>::remove(category.id, thread.id);
}

```

If it is so, let's create a separate issue for that and merge this PR.

bedeho (approved):
Nice work.

bedeho:
> > Be aware that you should not now lookup the category based on category_id, because this will already have been statically guaranteed prior to all the mutation commencing.
> 
> If I understood it correctly, all extrinsic should be rewritten, so they use input parameters only before mutation?
> 
> ```
> fn my_function(category_id, thread_id, ...) {
>     let (category, thread, ...) = Self::ensure_everything_is_ok(category_id, thread_id, ...); // only usage of category_id
> 
>     //
>     // == MUTATION SAFE ==
>     //
>     // do not use category_id after this point
> 
>     <ThreadById<T>>::remove(category.id, thread.id);
> }
> ```
> 
> If it is so, let's create a separate issue for that and merge this PR.

Yes.

PR #911, opened 2020-07-07
Forum post reaction

### LINKED ISSUE #763: Generalize PostReaction
Currently, we have

```Rust
pub enum PostReaction {
    /// No any reaction
    NonReacton,

    /// Thumb up to a post
    ThumbUp,

    /// Thumb down to a post
    ThumbDown,

    /// Like a post
    Like,
}
```

Now, if anyone wanted to replace the application-level semantics of either variant to something else, such as `Like` to `Hate`, then they would either have to accept a naming conflict with the runtime or fork the module. More importantly, if they wanted to increase the number of reaction types, they would have to fork the module. We can treat this a bit more generally by just having the reaction encoded as some integer like type, and accept the number of different reactions accepted as a trait configuration for the module. This would allow for forkless customization without any penalty in clarity of performance.

ondratra:
Solves subtasks of #766:
- Generalize PostReaction: #763

Fixes #763.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -549,6 +549,7 @@ decl_event!(
         <T as Trait>::ThreadId,
         <T as Trait>::PostId,
         <T as Trait>::ForumUserId,
+        <T as system::Trait>::Hash,
     {
         /// A category was introduced
         CategoryCreated(CategoryId),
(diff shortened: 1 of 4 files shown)


bedeho (commented):
This looks about right, however, I think we should do this after looking closer at this again

https://github.com/Joystream/joystream/issues/948

This also means there is at present a bug in your code for deleting posts/threads, until we fix this issue. If you want to minimize merge conflicts with all the other open PRs built on top of this one, we can have the fix in a new fresh PR in the future.

ondratra:
There are multiple PRs depending on this one, so it will be better to create separate PR for this. I've added a note about the problem to the enhancements list issue.

bedeho:
This is fix
https://github.com/Joystream/joystream/pull/979

Issue #716, opened 2020-06-17
Pioneer: Create Lead Opening Proposal and generic working group proposal functionality


Lezek123:
## Create Lead Opening Proposal

For Nicaea release we will need a new type of proposal that will allow creating a storage workng group "Lead Opening", similar to the typical worker opening created with `add_worker_opening` extrinsic.

Considering how many parameters (of varying types) are required for creating such opening and the fact that the `human_readable_text` parameter (of `addWorkerOpening` extrinsic) is actually expected by the Pioneer to be a valid JSON, compatible with the `GenericJoyStreamRoleSchema` from `/types/src/hiring/schemas/role.schema.typings.ts`  (even though it is technically just one argument of the `Bytes` type), it could potentially get quite complex to implement a fully-featured form that will support creating such proposal.

Currently the `working-groups/admin` tool supports creating Content Working Group openings with the simplified form that only contains:
- `Activate at` field (`Current` and `Exact`. Switching to `Exact` displays additional input where user can provide block number)
- `Max review period length` field (number of blocks)
- Simplified `Application staking policy` section, toggleable by a checkbox, with `Stake mode` and `Stake value` fields
- Simplified `Role staking policy` section (same as above)
- A textarea for `human_readable_text`, where the entire json is just treated as "one value"

Creating such simplified form for the _Create Lead Opening Proposal_ will probably not be too time-consuming, although such implementation will obviosly be lacking compared to the potential fully-featured implementation that will allow providing values for all the extrinsic parameters and a decent UI for `human_readable_text` json schema.

There are a few minor adjustments that could make `human_readable_text` as textarea (containing json) a _good enough_ solution for now though: 
- Validation and error handling using `schemaValidator` from `/types/src/hiring/index.ts` (already implemented in the CLI: https://github.com/Joystream/joystream/pull/707) - a very low-effort, high-return solution, as it allows us to get a very specific error messages with a negligible amount of coding required
- A good default value/values (inside `/working-groups/admin` tool those are based on the selected "draft") that will allow the user to only slightly modify some parameters in case there's a need to do so
- A command in the CLI that will allow generating a json in a similar manner as in `working-group:createOpening` command and just copy-pasting the result into the textarea.

I would also consider adding `role_slashing_terms` as part of the form, as this seem to be an important part of the policy and in contrast to the other "skipped fields", it's actually a required parameter in the extrinsic (`working-groups/admin` tool currently doesn't allow specifying any value other than `Unslashable`)

## Reusable functionality

[...]

PR #912, opened 2020-07-07
Pioneer: Add Working Group Lead Opening Proposal

Lezek123:
Implements https://github.com/Joystream/joystream/issues/716


pioneer/packages/joy-proposals/src/forms/AddWorkingGroupOpeningForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+  validationSchema: (props: FormContainerProps) => Yup.object().shape({
+    ...genericFormDefaultOptions.validationSchema,
+    ...Validation.AddWorkingGroupLeaderOpening(props.currentBlock?.toNumber() || 0)
+  }),
+  handleSubmit: genericFormDefaultOptions.handleSubmit,
+  displayName: 'AddWorkingGroupOpeningForm'
+})(AddWorkingGroupOpeningForm);
+
+export default withCalls<ExportComponentProps>(
+  ['derive.chain.bestNumber', { propName: 'currentBlock' }]
```
mnaamani, line 354: This will cause the form to reload every 6s, I see the value is being passed to the validation logic. Can we  query the current block much less frequently, and always have a margin that the exact block should be at least N + 5min x blocks/min where N is the latest current blocknumber read from state.

mnaamani:
Did a quick check, trying to create a proposal with exact block:

Screen Shot 2020-07-09 at 6 29 30 PM

I just fired up a local dev chain, so the numbers don't seem quite correct.

mnaamani:
The issue I pointed to above is not a bug, its just the selection of the min value constant:

`const MIN_EXACT_BLOCK_MINUS_CURRENT = 14400 * 5; // ~5 days`

mnaamani (changes requested):

mnaamani:
Perhaps the HRT can span two columns for a better layout ?

Screen Shot 2020-07-09 at 7 59 28 PM

mnaamani:
It would be good to have the cursor jump to the field that has invalid input on it after clicking submit proposal button.
Since the form is long, and if a field that is invalid as at the top, user doesn't initially notice why the button isn't doing anything. Maybe even a popup dialog to indicate the proposal cannot be submitted at this time.

Last suggestion would be to have the form fields that are required and have a default/min value to be pre-populated so the user doesn't have to fill them in unnecessarily.

pioneer/packages/joy-proposals/src/forms/AddWorkingGroupOpeningForm.tsx
mnaamani: I see you are already doing the latter part in the validation, the issue with the currentBlock number frequency is maybe premature optimisation so we can postpone.

Lezek123: I'm not sure if that cause much issues regarding the optimalization. No DOM is actually beeing re-rendered etc., since React is able to figure out that this doesn't cause any changes in the UI, so it looks like a pretty fast process, compared to everything else that's going on when using the form (unless I'm missing something here).

Lezek123:
> The issue I pointed to above is not a bug, its just the selection of the min value constant:
> `const MIN_EXACT_BLOCK_MINUS_CURRENT = 14400 * 5; // ~5 days`

I remember we were discussing it already together with @bedeho and came to the conclusion that there's no need to create an opening which begins ie. 1 day after it is created (or even less). This can cause the proposal to fail if it gets voted-through after more than 1 day, so to avoid this we can add a check in the UI that in case `ExactBlock` is specified, it should be reasonably far into the future. I set it to 5 days just because that's the voting period for this proposal, but this can be set to any other value if needed.  

> Perhaps the HRT can span two columns for a better layout ?

I think there are quite a few issues with the current proposal parameters layout, which doesn't work very well for "more-complex" parameters. I already started tacking this issue while working on `FillLeaderOpening` proposal (which also has some HRT values), so perhaps we can get back to it then?

Lezek123:
> It would be good to have the cursor jump to the field that has invalid input on it after clicking submit proposal button.
Since the form is long, and if a field that is invalid as at the top, user doesn't initially notice why the button isn't doing anything. Maybe even a popup dialog to indicate the proposal cannot be submitted at this time.

I was able to come up with some solution by using ref for `GenericProposalForm` container div and then operating directly on DOM elements after the form gets validated to focus the first error field. It seemed like the simplest way to implement it and looks like it works pretty well too.

Lezek123:
> Last suggestion would be to have the form fields that are required and have a default/min value to be pre-populated so the user doesn't have to fill them in unnecessarily.

Added some default values for maxReviewPeriod (30 days) and terminate/leave role unstaking periods (7 days).

pioneer/packages/joy-proposals/src/forms/AddWorkingGroupOpeningForm.tsx
mnaamani: yes its not worth it.

Issue #906, opened 2020-07-07
Merge components and app packages [Atlas]


kdembler:
After some discussion we've had, to solve some build issues and improve developer experience, it was decided that for the time being, we should merge `@joystream/components` inside atlas `app`. While doing that, we should also try to separate "agnostic" components and maintain them as they were a separate package.

PR #914, opened 2020-07-08
move @joystream/components into atlas app, clean up build process

kdembler:
## Partly deprecated by https://github.com/Joystream/joystream/pull/914#issuecomment-658794277

This PR merges `@joystream/components` package with the `app` as discussed in #906.

### Summary

Since the diff is pretty big, this is a list of changes to make the review easier:

- Everything under `packages/components` was moved to `packages/app/src/shared`
- npm scripts of both packages were merged
- TS and Parcel paths were created so that we can easily import like `import { Button } from 'shared/components'`
- Removed rollup and cleaned up dependencies in general
- Moved app dependencies inside `app` package, left only build tools in the monorepo root
- Added Parcel transformers for importing svg files as React components
- Cleaned up icons under `shared/icons` with a `index.ts` for easy imports
- Updated Parcel and regenerated lockfiles

### Caveats

- It may be required to delete `.parcel-cache` since the build process was affected rather significantly.
- I've experienced Parcel crash at one occasion but wasn't able to reproduce it, if that's something that happens again we may need to downgrade

### Improvements

- With the testing I've done, I've seen great improvements in build/reload times. Now we don't need to build entire `components` package and then rebundle the `app` on every change, Parcel will just hot reload one specific module
- I feel like this greatly improves imports quality, before this change I've seen at least 3 different import styles, now we can clean that up. These are rules I've went by and I think will work good for us:
  - All imports inside `shared` should be relative, this way we don't rely on `app` aliases and can move the entire directory to a separate package with no problem
  - Imports outside of `shared` should use absolute imports using aliases, e.g. `shared/components`, `shared/theme`

### Feedback

I've had to make some decisions here and I feel like these are good improvements but any feedback and discussion is welcome


atlas/packages/app/tsconfig.json
```diff
@@ -2,8 +2,25 @@
...
 
 	"compilerOptions": {
+		"baseUrl": "./",
 		"declaration": true,
 		"declarationDir": "dist",
-		"lib": ["DOM", "ES2019"]
-	}
+		"lib": ["DOM", "ES2019"],
+		"paths": {
+			"components": ["src/components"],
```
fcescob: I think it would be clearer to prefix paths with `@` or `~` to indicate that they come from the same repo and not from an outside package

atlas/babel.config.json
```diff
@@ -4,6 +4,5 @@
 		"@babel/preset-react",
 		"@babel/preset-typescript",
 		"@emotion/babel-preset-css-prop"
-	],
-	"babelrcRoots": ["*", "packages/*"]
```
fcescob, line 8: Why was this removed?

atlas/packages/app/package.json
```diff
@@ -14,25 +14,60 @@
...
+		"chromatic": "^4.0.3",
+		"react-docgen-typescript-loader": "^3.7.1",
+		"storybook-addon-jsx": "^7.1.15",
+		"parcel": "^2.0.0-beta.1",
+		"@fortawesome/fontawesome-svg-core": "^1.2.28",
+		"@fortawesome/free-solid-svg-icons": "^5.13.0",
+		"@fortawesome/react-fontawesome": "^0.1.9",
+		"@parcel/transformer-svg-react": "^2.0.0-nightly.1739",
+		"@parcel/transformer-svgo": "^2.0.0-nightly.1739",
+		"@storybook/addon-actions": "^5.3.17",
```
fcescob: Given that we're going to separate the app from components again in the future and since components is meant to be published,  I would stick all devDependencies to the project root, keeping only things external to the build process here

atlas/packages/app/tsconfig.json
kdembler: Like we've talked, using `@` as a root was the first thing I wanted to do. However, Parcel allows neither `@` and `~` in the alias, as it reserves those for some specific use cases. That really sucks though, Webpack allows both :/

kdembler: So I guess we could use some other character (`#`?) for the root or just leave those be as they are. To be honest I don't like either of those solutions that much but it's still better than doing relative paths everywhere. We can also switch to Webpack haha

 What do you think?

atlas/packages/app/package.json
kdembler: In general, I agree, but I would say that what goes into the root `package.json` should be determined by what will be reused across packages after the split, rather than what is external to the build process.

I will try to make to clean it up in a way that makes sense.

atlas/babel.config.json
kdembler: By mistake I think, restored it in 0fcbed9e3a61588cb8ab10868bd5b27439cc0b01

atlas/packages/app/package.json
kdembler: Tried my best in 0fcbed9e3a61588cb8ab10868bd5b27439cc0b01, let me know if you're happy with the dependencies now

atlas/packages/app/tsconfig.json
fcescob: Let's leave it as is for now, and tackle a possible migration to webpack in a next PR if we start to see the benefits of doing so

atlas/packages/app/package.json
fcescob: LGTM!

fcescob (approved):

atlas/packages/app/tsconfig.json
kdembler: :+1:

mnaamani:
After merging #878 there are merge conflicts to be resolved.

I tested by running

```
yarn install
yarn test
yarn start
```
There are some tests failing which I assume are WIP.
Storybook renders correctly after yarn start.

I opened the built app but all I see is:

Screen Shot 2020-07-14 at 12 15 41 PM

Is that expected?

bedeho:
Har merge conflict.

kdembler:
## Updated description after move to webpack:

So after all to streamline the build process completely we decided to move over to Webpack and CRA. That means in this PR we go from Parcel, Rollup and Webpack to just Webpack. This allows us to for example share config between the app and the storybook (as we do).

### Some notable differences in comparison with the original description:

- Aliases were changed so that `@` represents project root, e.g. `@/shared/components`
- A lot of dependencies were removed - Rollup, Parcel, etc. Also Babel, since it's a part of `react-scripts` that Storybook leverages
- Version ranges of some root-level packages were updated (eslint, jest) to match the versions required by `react-scripts`.

### Caveats:

- Root-level `yarn dev` won't work for the time being because of some incompatibility between Lerna nad CRA on how stdin is handled. This is already fixed in CRA and should be a part of the next release. Until then we should do `yarn dev` inside the `app` package
- When running the `app` you can see a warning:
```
The following changes are being made to your tsconfig.json file:
- compilerOptions.paths must not be set (aliased imports are not supported)
```
This is because of an CRA config override we did and doesn't have any effect other than this warning. If we eject CRA at some point we can modify the scripts to get rid of it

atlas/packages/app/tsconfig.json
```diff
@@ -1,26 +1,29 @@
...
-	},
-
-	"exclude": [
-		"node_modules", "dist", "src/shared/stories"
-	],
-
-	"include": [
-		"src/shared/icons/types.d.ts"
-	]
+  "extends": "./tsconfig.paths.json",
```
fcescob, line 2: It is a bit of an overkill to have another file just for the paths imho, also looks like some options here are shared by the workspace root `tsconfig.json` , so those could be probably be moved all in the workspace root and inherited here.

atlas/packages/app/config-overrides.js
```diff
@@ -0,0 +1,46 @@
+const path = require('path')
+const { override, addBabelPreset, addWebpackAlias, disableEsLint } = require('customize-cra')
+const eslintConfig = require('../../.eslintrc.js')
+
+// based on https://github.com/arackaf/customize-cra/issues/175#issuecomment-610023655
+const useEslintConfig = (configRules) => (config) => {
+	const updatedRules = config.module.rules.map((rule) => {
+		// Only target rules that have defined a `useEslintrc` parameter in their options
+		if (rule.use && rule.use.some((use) => use.options && use.options.useEslintrc !== void 0)) {
```
fcescob, line 9: Any specific reason for `void 0` instead of `undefined` here?

atlas/packages/app/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
fcescob, line 1: What is this?

fcescob:
There are some issues with how tests are transpiled and run:


1.  Running `yarn test` from the root of the monorepo, gives the following error:

immagine

To me this looks like bad transpilation, I name Babel as the culprit here.

2. Running `yarn run jest` from `packages/app` gives the following one instead:

immagine


Let me know if you can reproduce those issues, if you feel like the scope would get too broad by investingating both, we can focus on addressing just one for now, as long as we have a way to run tests.

For reference, here is a branch with tests running, based on the pre-rework state of the repo: https://github.com/Gamaranto/joystream/tree/atlas-tests

atlas/packages/app/config-overrides.js
kdembler: This code is copied over from `customize-cra`, that's the check they use there, I didn't question whys

atlas/packages/app/src/react-app-env.d.ts
kdembler: This is a file generated by CRA on every run, we could possibly add it to `.gitignore`

atlas/packages/app/tsconfig.json
kdembler: This isn't really done for convenience, rather as a "hack". CRA doesn't allow some options in `tsconfig.json` (look at Caveats in my comment above). If it detects some options it doesn't allow, it will try to delete them. Keeping them in a separate config allows us to alleviate that, so the paths aren't removed. Once again, if we decide to eject we can tweak that and get rid of this.

Regarding moving some options to the root, I will look into that

kdembler:
I think the problem was that Jest wasn't using proper Babel config. I've fixed them by leveraging Jest supplied by CRA. I've moved the config to a section in `packages/app/package.json`. If we wanted to, we could also create the config programmatically by adding `jest` property to exports of `config-overrides.js`

atlas/package.json
```diff
@@ -7,11 +7,11 @@
 	"scripts": {
 		"start": "concurrently \"yarn dev\" \"yarn storybook\"",
 		"build": "lerna run build",
-		"dev": "lerna run --scope atlas-app --stream dev",
+		"dev": "lerna run --scope atlas-app dev",
 		"storybook": "lerna run --scope atlas-app --stream storybook",
 		"lint": "yarn run eslint . --ext .js,.jsx,.ts,.tsx",
 		"clean": "lerna clean && rm -rf node_modules",
-		"test": "jest"
+		"test": "lerna run --stream test"
```
fcescob: I suggest we change this to `yarn workspace atlas-app test` since that gives a nicer output and allows us to pass jest CLI options like `--coverage` directly, which currently doesn't work with lerna.
Otherwise LGTM!

atlas/packages/app/tsconfig.json
kdembler: > Regarding moving some options to the root, I will look into that

Maybe let's do that during rebasing on common linting

atlas/package.json
kdembler: Good idea, I've also used the same approach to fix `yarn dev` from the root

fcescob (approved):

PR #920, opened 2020-07-09
Forum archival status

ondratra:
Solves subtask of #766:
- Allow both the author and forum moderators to update archival status.

It allows categories and threads to be archived, making them (and their content) immutable. 

To minimize merge conflict potential, this PR assumes #901, #909, and #911 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -1269,7 +1269,7 @@ impl<T: Trait> Module<T> {
             Ok(())
         };
 
-        // Decide if actor can delete category
+        // Decide if actor can update thread
         match actor {
             PrivilegedActor::Lead => (),
             PrivilegedActor::Moderator(moderator_id) => ensure_moderator_can_update(moderator_id)?,


runtime-modules/forum/src/lib.rs
```diff
@@ -790,6 +787,38 @@ decl_module! {
...
+        fn update_thread_archival_status(origin, actor: PrivilegedActor<T>, category_id: T::CategoryId, thread_id: T::ThreadId, new_archival_status: bool) -> dispatch::Result {
+            // Ensure data migration is done
+            Self::ensure_data_migration_done()?;
+
+            // Ensure actor can update category
+            let (category, thread) = Self::ensure_can_update_thread_archival_status(origin, &actor, &category_id, &thread_id)?;
+
+            if let Some(tmp_parent_category_id) = category.parent_category_id {
+                // Get path from parent to root of category tree.
+                let category_tree_path = Self::ensure_valid_category_and_build_category_tree_path(&tmp_parent_category_id)?;
```
bedeho: `ensure_valid_category_and_build_category_tree_path(&tmp_parent_category_id)?` is invoked as if it can fail, but it cannot, unless our own code has a bug. This is because `tmp_parent_category_id` does not come from the user, but from our own state, which we must presume is correct. Hence we really need a way to build the path you need without exposing an error which is not possible.

runtime-modules/forum/src/lib.rs
```diff
@@ -790,6 +787,38 @@ decl_module! {
...
+
+                if new_archival_status && Self::ensure_can_mutate_in_path_leaf(&category_tree_path).is_err() {
```
bedeho: I don't understand this check, it returns `Ok`, despite no storage state being updated, something seems wrong here, and also too complex. This would in all cases need a better explanation.

runtime-modules/forum/src/lib.rs
```diff
@@ -790,6 +787,38 @@ decl_module! {
             Ok(())
         }
 
+        /// Update category
+        fn update_thread_archival_status(origin, actor: PrivilegedActor<T>, category_id: T::CategoryId, thread_id: T::ThreadId, new_archival_status: bool) -> dispatch::Result {
+            // Ensure data migration is done
+            Self::ensure_data_migration_done()?;
+
+            // Ensure actor can update category
+            let (category, thread) = Self::ensure_can_update_thread_archival_status(origin, &actor, &category_id, &thread_id)?;
```
bedeho: This name `ensure_can_update_thread_archival_status` seems to suggest that its conclusive single guarantor that the actor can update the archival state of a thread, however, there are all the remaining checks from line 798 which seem to be checking additional things relevant to that assessment. It seems that either those checks really should be part of this one check, or, the name of this check really is incorrect, as it providing some other check.

runtime-modules/forum/src/lib.rs
```diff
@@ -1198,12 +1228,56 @@ impl<T: Trait> Module<T> {
...
+
+        // Closure ensuring moderator can delete category
+        let ensure_moderator_can_update = |moderator_id: &T::ModeratorId| -> dispatch::Result {
+            Self::ensure_can_moderate_category_path(moderator_id, &category_id)
+                .map_err(|_| ERROR_MODERATOR_CANT_UPDATE_CATEGORY)?;
+
+            Ok(())
+        };
+
+        // Decide if actor can delete category
```
bedeho: Why is this comment talking about deletion?

bedeho (changes requested):

runtime-modules/forum/src/lib.rs
ondratra: Mistake - probably a line of code copied from elsewhere. It's fixed now

ondratra:
You are absolutely right about those issues. They are all fixed however in following PRs that changes whole section relating to commented lines. Changes can be seen here - https://github.com/Joystream/joystream/pull/929/files . I suggest you check the code from #929 and merge this PR if it is ok.

runtime-modules/forum/src/lib.rs
ondratra: This sub-issue is outdated. Check is done in a different way now.

runtime-modules/forum/src/lib.rs
ondratra: This strange condition was what got left after I cut out other parts of old code around it. Now this condition gone because a situation it covered cannot occur anymore - we cannot archive thread in an already archived category, it will cause error because category cannot be mutated.

runtime-modules/forum/src/lib.rs
ondratra: The big chunk of code you mentioned is removed now.

ondratra:
Changes from this PR were already accepted since one of the following PRs in a sequence was merged. Only a one-line comment change has remained. All issues from the review should already be addressed.

More info about what PR's were merged recently can be found at the bottom of page #766.

bedeho (approved):

bedeho:
Ok, since everything has already been dealt with I am closing this PR.

PR #921, opened 2020-07-09
Forum typed errors

ondratra:
Solves subtask of #766:
- Use new type safe error mechanism.

To minimize merge conflict potential, this PR assumes #920 will be merged first.

PR #929, opened 2020-07-10
Forum ensures refactoring

ondratra:
Solves subtask of #766:
- There are one or more cases where ensures should be introduced, in line with this [Ensure abstraction] (comment), for example for categories. Identify all such cases.

To minimize merge conflict potential, this PR assumes #921 will be merged first.


bedeho (approved):
LGTM.

PR #930, opened 2020-07-10
Working groups admin glitches

Lezek123:
An attempt to finalize https://github.com/Joystream/joystream/issues/741 by fixing an issue with React hooks and incorrect method of sending extrinsics in `Admin.controller.tsx`.

Note that most of the issues from https://github.com/Joystream/joystream/issues/741 have already been fixed at this point (either intentionally or as a side-effect of fixing other issues, like the `ts-nocheck` comment).

What should be improved after this PR is:
- No more `Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks...` error
- When extrinsic fails in the working groups Admin panel - it should display the correct error message, just like it does when using `TxButton` (due to the fact that Admin controller now uses `queueExtrinsic`)

I tried to also fix an additioanl issue I found, which is that form values are beeing reset on submit when adding a new opening via admin tool modal, but this turned out to be a bit more complex of an issue, due to bad design of the data flow etc., so I decided to just call `closeModal` on submit to prevent possible mistakes.

### DIFF
--- a/pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
 import { Link } from 'react-router-dom';
 import { formatBalance } from '@polkadot/util';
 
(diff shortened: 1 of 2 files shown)


pioneer/packages/joy-roles/src/tabs/Admin.controller.tsx
```diff
@@ -367,53 +368,35 @@ const newEmptyState = (): State => {
...
-        console.log('Successful transfer with hash ' + status.asFinalized.toHex());
-      } else {
-        console.log('Status of transfer: ' + status.type);
-      }
-
-      events.forEach(({ phase, event: { data, method, section } }) => {
-        console.log(phase.toString() + ' : ' + section + '.' + method + ' ' + data.toString());
-      });
-    });
+    this.queueExtrinsic({ extrinsic: tx, txSuccessCb: this.onTxSuccess, accountId });
```
mnaamani, line 399: should we handle `txFailureCb` as well ?

Lezek123: When transactions are correctly initialized using the transaction queue in Pioneer, the ExtrinsicFailed event is automatically handled and results in an error message beeing displayed. Other than that I don't think we need any additional handling.

mnaamani: I guess the view doesn't really need to adjust based on result anyway.

mnaamani:
nice fixs!

PR #935, opened 2020-07-13
Restore storage-node CLI commands.

### LINKED ISSUE #873: Fix Storage Node CLI commands
Storage node CLI contains several broken commands that need to be fixed:
- upload
- download
- head

This task was conceived [here](https://github.com/Joystream/joystream/pull/746).

shamil-gadelshin:
Changes:
- convert CLI package to TypeScript
- rewrite upload, download, head commands
- refactor commands to separate classes
- introduce parameter validation
- rewrite usage
- merge Nicaea linting rules and fix errors

Future changes:
- currently, the CLI package is based on the `meow` package which doesn't support subcommands - we should consider migration to the `oclif`. The current class-based implementation allows us to do that with less effort.
- the `helios` package looks like the candidate to separate CLI command. It doesn't need to be a separate package. It also contains the separate implementation of the 'head' command.
- also, we should continue with TypeScript migration because it allows us to get rid of multiple errors.
- we should introduce tests for the storage-node project


package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "test": "yarn && yarn workspaces run test",
     "test-migration": "yarn && yarn workspaces run test-migration",
-    "postinstall": "yarn workspace @joystream/types build",
-    "cargo-checks": "devops/git-hooks/pre-commit && devops/git-hooks/pre-push",
+    "postinstall": "yarn workspace @joystream/types build && yarn workspace storage-node run build",
```
mnaamani, line 9: Just leaving anote here, this will get long, so we can do `yarn workspaces run build` which will run the `build` npm script in all workspaces  sequentially. However the order is important because storage-node depends on types. For this we need to start using lerna soon.

storage-node/packages/cli/package.json
```diff
@@ -27,11 +27,12 @@
     "node": ">=10.15.3"
   },
   "scripts": {
-    "test": "mocha 'test/**/*.js'",
-    "lint": "eslint 'paths/**/*.js' 'lib/**/*.js'"
+    "test": "mocha 'dist/test/**/*.js'",
+    "lint": "eslint 'paths/**/*.js' 'lib/**/*.js'",
```
mnaamani: shouldn't we be linting the `*.ts` files instead ?

storage-node/packages/cli/src/cli.ts
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env node
```
mnaamani: Shouldn't this hashbang go in `packages/cli/bin/cli.js` instead ?

storage-node/packages/cli/src/cli.ts
```diff
@@ -0,0 +1,118 @@
...
+  },
+  // needs to be updated to take a content id and resolve it a potential set
+  // of providers that has it, and select one (possibly try more than one provider)
+  // to fetch it from the get api url of a provider..
+  download: async (api: any, url: string, contentId: string, filePath: string) => {
+    let downloadCmd = new DownloadCommand(api, url, contentId, filePath);
+
+    await downloadCmd.run();
+  },
+  // Shows asset information derived from request headers.
```
mnaamani:
```suggestion
  // Shows asset information derived from response headers.
```

storage-node/packages/cli/src/cli.ts
```diff
@@ -0,0 +1,118 @@
...
+    // dev accounts are automatically loaded, no need to add explicitly to keyring using loadIdentity(api)
+    return dev.init(api)
+  },
+  // Checks that the setup done by dev-init command was successful.
+  'dev-check': async api => {
+    // dev accounts are automatically loaded, no need to add explicitly to keyring using loadIdentity(api)
+    return dev.check(api)
+  },
+  // Uploads the file to the system. Registers new data object in the runtime, obtains proper colossus instance URL.
+  upload: async (api: any, filePath: string, dataObjectTypeId: string, keyFile: string, passPhrase: string, memberId: string) => {
```
mnaamani: perhaps make keyFile and passPhrase to be the last two arguments so we leave possibility to make them optional, but not being entered on the command line.

storage-node/packages/cli/src/cli.ts
```diff
@@ -0,0 +1,118 @@
...
+const FLAG_DEFINITIONS = {
+  // TODO: current version of meow doesn't support subcommands. We should consider a migration to yargs or oclif.
+}
+
+const usage = `
+  Usage:
+    $ storage-cli command [arguments..]
+
+  Commands:
+    upload            Upload a file to a Colossus storage node. Requires a
```
mnaamani:
```suggestion
    upload            Upload a file to the Joystream network. Requires a
```

mnaamani: and update order of arguments to match with suggestion below.

storage-node/packages/cli/src/commands/download.ts
```diff
@@ -0,0 +1,71 @@
...
+import chalk from "chalk"
+import fs from "fs";
+import {BaseCommand} from "./base";
+
+// Download command class. Validates input parameters and execute the logic for asset downloading.
+export class DownloadCommand extends BaseCommand{
+    private readonly api: any;
+    private readonly storageNodeUrl: string;
+    private readonly contentId: string;
+    private readonly filePath: string;
```
mnaamani:
```suggestion
    private readonly outputFilePath: string;
```

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+
+        let memberId: number = parseInt(this.memberId);
+        if (isNaN(dataObjectTypeId)) {
+            this.fail(`Cannot parse memberIdString: ${this.memberId}`);
+        }
+
+        return {
+            accountId,
+            ipfsCid: await this.computeIpfsHash(),
+            contentId: ContentId.generate().encode(),
```
mnaamani: do not `.encode()` here. We store the raw bytes in the runtime as a `H256` not a "string/text".
We used the encoded form, in URLs, and in the versioned-store only.

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+                this.fail(`File read failed: ${err}`);
+            });
+
+        return await ipfsHash.of(file);
+    }
+
+    // Read the file size from the file system.
+    private getFileSize(): number {
+        const stats = fs.statSync(this.mediaSourceFilePath);
+        return stats.size;
```
mnaamani: command should fail (maybe not right here) if file size exceeds MAX_CONTENT_LENGTH

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+        this.passPhrase = passPhrase;
+        this.memberId = memberId;
+    }
+
+    // Provides parameter validation. Overrides the abstract method from the base class.
+    protected validateParameters(): boolean {
+        return this.mediaSourceFilePath && this.mediaSourceFilePath !== ""
+            && this.dataObjectTypeId && this.dataObjectTypeId !== ""
+            && this.keyFile && this.keyFile !== ""
+            && this.passPhrase && this.passPhrase !== ""
```
mnaamani: we should allow for possibility that keyfile is not locked, as well as no keyfile being supplied, in which case we could have the development account `//Alice` be loaded as the "default" key.

We can determine if we are on a development chain by interrogating `await api.rpc.system.chain()` is equal to  `Development`

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    // Command executor.
+    async run() {
+        // Checks for input parameters, shows usage if they are invalid.
+        if (!this.assertParameters()) return;
+
+        let addContentParams = await this.getAddContentParams();
+        debug(`AddContent Tx params: ${JSON.stringify(addContentParams)}`);
+        debug(`Decoded CID: ${ContentId.decode(addContentParams.contentId).toString()}`);
```
mnaamani: see my earlier comment about how content id is stored in runtime.

```suggestion
        debug(`Decoded CID: ${addContentParams.contentId.encode()}`);
```

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+const debug = Debug('joystream:storage-cli:upload');
+
+// Defines maximum content length for the assets (files). Limits the upload.
+const MAX_CONTENT_LENGTH = 500 * 1024 * 1024; // 500Mb
+
+// Defines the necessary parameters for the AddContent runtime tx.
+interface AddContentParams {
+    accountId: string,
+    ipfsCid: string,
+    contentId: string,
```
mnaamani:
```suggestion
    contentId: ContentId,
```

storage-node/packages/cli/src/commands/upload.ts
```diff
@@ -0,0 +1,207 @@
...
+
+        // Upload file from the stream.
+        try {
+            const fileSize = this.getFileSize();
+            const config: AxiosRequestConfig = {
+                headers: {
+                    'Content-Type': '', // https://github.com/Joystream/storage-node-joystream/issues/16
+                    'Content-Length': fileSize.toString()
+                },
+                'maxContentLength': MAX_CONTENT_LENGTH,
```
mnaamani: As far as I know, this is the max length of the response, so it doesn't really make sense to have this option here. It is the storage node that will determine how much data it will accept :).

mnaamani (changes requested):

mnaamani:
Just a suggestion to also update the storage-node/package.json `checks` script to:

`"checks": "yarn lint && prettier ./ --check && tsc --noEmit --pretty",`

I noticed some formatting was missing, so you probably only did linter fixes, but not prettier formatting?

on a side note: I noticed different behavior between `prettier . --write` and `prettier ./ --write`

storage-node/packages/cli/src/cli.ts
shamil-gadelshin: Sure.

storage-node/packages/cli/src/cli.ts
shamil-gadelshin: Sure

storage-node/packages/cli/src/cli.ts
shamil-gadelshin: Ok.

storage-node/packages/cli/src/commands/download.ts
shamil-gadelshin: Ok.

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Without it I ran into the default axios limit. I could set it 10GB if necessary.

storage-node/packages/cli/src/cli.ts
shamil-gadelshin: Sure. It was an artifact.

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Ok.

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Ok

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Done.

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Yes, it is failing.

package.json
shamil-gadelshin: Ok

storage-node/packages/cli/package.json
shamil-gadelshin: Fixed.

storage-node/packages/cli/src/commands/upload.ts
shamil-gadelshin: Done

storage-node/packages/cli/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "test": "mocha 'dist/test/**/*.js'",
-    "lint": "eslint 'paths/**/*.js' 'lib/**/*.js'",
+    "lint": "eslint --ext .js,.jsx,.ts,.tsx . && tsc --noEmit --pretty",
```
mnaamani: I would leave the `&& tsc --noEmit --pretty` to put in the checks script.
So the lint script can take some extra args:

`yarn lint --fix`

storage-node/packages/cli/src/commands/upload.ts
mnaamani: But was the problem with the request (upload) or the response ?

storage-node/packages/cli/src/commands/upload.ts
mnaamani: are you only depending on the axios maxContentLength header option for this check? Still not sure we are using that correctly. Let me check HTTP docs.

mnaamani:
>     * currently, the CLI package is based on the `meow` package which doesn't support subcommands - we should consider migration to the `oclif`. The current class-based implementation allows us to do that with less effort.

I agree, we are using it in the joystream-cli package currently.


Merging, minor issue with maxContentLength can be addressed later.

PR #937, opened 2020-07-13
Forum fuzzy testing

ondratra:
Prototype for forum fuzzy testing. WIP


bedeho:
Conclusion: We are going to just leave this PR as a draft for now, as the cost-benefit and risk of doing fuzzing does not warrant any further work.

PR #938, opened 2020-07-13
Fix zero unstaking periods in the add_opening() parameters bug.

### LINKED ISSUE #913: Not all unstaking periods are checked for beeing > 0
Related issue: https://github.com/Joystream/joystream/issues/703
There was a fix introduced in https://github.com/Joystream/joystream/pull/858, but it only covered 3 unstaking periods:
- `fill_opening_successful_applicant_application_stake_unstaking_period`
- `fill_opening_failed_applicant_application_stake_unstaking_period`
- `fill_opening_failed_applicant_role_stake_unstaking_period`

Those are currently checked to be `> 0` when creating a working group opening to prevent creating an opening which cannot be filled.

The issue is that not only those, but **all** unstaking periods should be checked to be `> 0`, which can be traced to `runtime-modules/stake/src/lib.rs`:
```
    fn initiate_unstaking(
        &mut self,
        unstaking_period: BlockNumber,
        now: BlockNumber,
    ) -> Result<(), InitiateUnstakingError> {
        ensure!(
            unstaking_period > Zero::zero(),
            InitiateUnstakingError::UnstakingPeriodShouldBeGreaterThanZero
        );
```
This is probably why all unstaking periods are optional when creating an opening (so always `Option(null)` should be used instead of `0`).

Providing `0` value for some of the opening unstaking periods can cause unexpected results, ie.:
- `exit_role_stake_unstaking_period` / `exit_role_application_stake_unstaking_period` - inability to use `withdrawApplication` (throws `WithdrawWorkerApplicationUnstakingPeriodTooShort`)
- `crowded_out_unstaking_period_length` - runtime panick when tiggered. This can be tested by creating an opening with max_applicants set to `1` and some `AtLeast` application stake (with  `crowded_out_unstaking_period_length` set to `0`). Then we apply to this opening with minimum application stake (ie. `100`) and then - as a different member - with higher stake (ie. `101`). This should cause the unstaking to trigger and panick the runtime.

I haven't tested all the scenarios, but I believe there would be issues with `leaveRole` (`exit_role..._unstaking_period`), `terminateApplication`, `terminateRole` (`terminate..._unstaking_period`), etc. too.

**Conclusion:**
My understanding is that all unstaking periods should be checked to be `> 0` when creating an opening.

shamil-gadelshin:
Fixes [this bug](https://github.com/Joystream/joystream/issues/913).

Changes - add non-zero parameters checks (and tests) for:
- terminate_application_stake_unstaking_period
- terminate_role_stake_unstaking_period
- exit_role_application_stake_unstaking_period
- exit_role_stake_unstaking_period
- application_rationing_policy.max_applicant_numbers
- role_staking_policy.crowded_out_unstaking_period_length
- role_staking_policy.review_period_expired_unstaking_period_length
- application_staking_policy.crowded_out_unstaking_period_length
- application_staking_policy.review_period_expired_unstaking_period_length

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -3,7 +3,7 @@ authors = ['Joystream']
 build = 'build.rs'
 edition = '2018'
 name = 'joystream-node'
-version = '2.5.1'
+version = '2.6.0'
 default-run = "joystream-node"
 
 [[bin]]
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -3,7 +3,7 @@ authors = ['Joystream']
 build = 'build.rs'
 edition = '2018'
 name = 'joystream-node'
-version = '2.5.0'
+version = '2.5.1'
 default-run = "joystream-node"
 
 [[bin]]
--- a/runtime-modules/working-group/Cargo.toml
+++ b/runtime-modules/working-group/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'substrate-working-group-module'
-version = '1.0.2'
+version = '1.1.0'
 authors = ['Joystream contributors']
 edition = '2018'
 
--- a/runtime-modules/working-group/Cargo.toml
+++ b/runtime-modules/working-group/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'substrate-working-group-module'
-version = '1.0.1'
+version = '1.0.2'
 authors = ['Joystream contributors']
 edition = '2018'
 
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -297,6 +297,10 @@ decl_error! {
         /// Invalid OpeningPolicyCommitment parameter (application_staking_policy):
         /// review_period_expired_unstaking_period_length should be non-zero.
         ApplicationStakingPolicyReviewPeriodUnstakingPeriodIsZero,
+
+        /// Invalid OpeningPolicyCommitment parameter (application_rationing_policy):
+        /// max_active_applicants should be non-zero.
+        ApplicationRationingPolicyMaxActiveApplicantsIsZero,
     }
 }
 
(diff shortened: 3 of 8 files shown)


node/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ['Joystream']
 build = 'build.rs'
 edition = '2018'
 name = 'joystream-node'
-version = '2.5.0'
+version = '2.5.1'
```
mnaamani: I would bump it to `2.6.0` because the runtime spec version was changed.
`2.5.1` would make sense if we only changed implementation detail in runtime `impl` version

runtime-modules/working-group/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'substrate-working-group-module'
-version = '1.0.1'
+version = '1.0.2'
```
mnaamani: I'm going to be picky with the version numbers here because we are not publishing them, but technically the change although a "bug fix", is actually a change in behavior

runtime-modules/working-group/src/errors.rs
```diff
@@ -265,6 +265,42 @@ decl_error! {
         /// Invalid OpeningPolicyCommitment parameter:
         /// fill_opening_successful_applicant_application_stake_unstaking_period should be non-zero.
         FillOpeningSuccessfulApplicantApplicationStakeUnstakingPeriodIsZero,
+
```
mnaamani, line 268: I wonder wouldn't a single general `UnstakingPeriodIsZero` have been sufficient? 
Then you wouldn't have to pass the different variants as arguments in `fn check_unstaking_period()` and `fn check_staking_policy()`

The downside is the caller wouldn't be able to identify in which policy they made a mistake, but it would simplify the code, and reduce chance of returning the incorrect error variant.

node/Cargo.toml
shamil-gadelshin: Ok

runtime-modules/working-group/Cargo.toml
shamil-gadelshin: Ok

runtime-modules/working-group/src/errors.rs
shamil-gadelshin: I thought about it because it would be easier to implement, but I wanted to avoid the downside you mentioned.

mnaamani:
Changes look good, will just wait on travis then merge

mnaamani (approved):

PR #943, opened 2020-07-14
CLI: Linter fixes

Lezek123:
Introduces fixes to all linter errors following https://github.com/Joystream/joystream/pull/917

### DIFF
--- a/.github/workflows/joystream-cli.yml
+++ b/.github/workflows/joystream-cli.yml
@@ -3,7 +3,7 @@ on: [pull_request, push]
 
 jobs:
   cli_build_ubuntu:
-    name: Ubuntu Check
+    name: Ubuntu Checks
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -14,14 +14,14 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - name: check
+    - name: checks
       run: |
         yarn install --frozen-lockfile
         yarn madge --circular types/
-        yarn workspace joystream-cli check
+        yarn workspace joystream-cli checks
 
   cli_build_osx:
-    name: MacOS Check
+    name: MacOS Checks
     runs-on: macos-latest
     strategy:
       matrix:
@@ -32,8 +32,8 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - name: check
+    - name: checks
       run: |
         yarn install --frozen-lockfile --network-timeout 120000
         yarn madge --circular types/
-        yarn workspace joystream-cli check
+        yarn workspace joystream-cli checks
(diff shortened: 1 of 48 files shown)


mnaamani (changes requested):
Great! 

Can you also run `yarn format` and check-in the changed files.
The reason the `--fix` argument when you ran linter didn't sort those out is because the `lint` script is only linting .ts files `/src` folder I think.

Perhaps update it the lint script to cover the whole directory. So `format`, `lint`, and `checks` all cover the same files.

Perhaps also update `.github/workflows/joystream-cli.yml`

to run `yarn workspace joystream-cli checks` in place of `yarn workspace joystream-cli build`

mnaamani:
I think ideally

```
    "lint": "eslint . --quiet --ext .ts",
    "checks": "yarn lint && tsc --noEmit --pretty && prettier . --check",
    "format": "prettier . --write"
```

but will need to fix:

```
/Users/mokhtar/joystream/joystream/cli/test/commands/council/info.test.ts
  3:1  error  'describe' is not defined  no-undef

✖ 1 problem (1 error, 0 warnings)
```

I think adding mocha to the `env` in the eslintrc file should do the trick

Lezek123:
Fixed the mentioned issues.
While testing I noticed the lack of unused imports check for the CLI.
I found this check very helpful in Pioneer. so I added `@typescript-eslint/no-unused-vars` rule to CLI's `.eslintrc.js` too. Also set `"noUnusedLocals": true` in `tsconfig.json` (the latter will additionaly cause the build to fail if there are any unused imports spotted).

Perhaps this can be part of the common standard?

.github/workflows/joystream-cli.yml
```diff
@@ -14,14 +14,14 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - name: build
+    - name: check
       run: |
         yarn install --frozen-lockfile
         yarn madge --circular types/
-        yarn workspace joystream-cli build
+        yarn workspace joystream-cli check
```
mnaamani:
```suggestion
        yarn workspace joystream-cli checks
```

Its actually `checks`
`check` is a yarn [builtin command](https://classic.yarnpkg.com/en/docs/cli/check/#toc-yarn-check) that does sanity check on package.json and yarn.lock

mnaamani (commented):
> Fixed the mentioned issues.
> While testing I noticed the lack of unused imports check for the CLI.
> I found this check very helpful in Pioneer. so I added `@typescript-eslint/no-unused-vars` rule to CLI's `.eslintrc.js` too. Also set `"noUnusedLocals": true` in `tsconfig.json` (the latter will additionaly cause the build to fail if there are any unused imports spotted).
> 
> Perhaps this can be part of the common standard?

Yes those rules look very good and can catch a tonne of bugs. I'm surprised the no-unused-vars isn't on by default in the rule-sets we have.

Added reminder in https://github.com/Joystream/joystream/issues/664

.github/workflows/joystream-cli.yml
mnaamani: sorry seems I missed the fix in later commit

PR #919, opened 2020-07-09
Network testing: Nicaea proposals- leader hiring happy case

gleb-urvanov:
Includes implementation for following issues:
https://github.com/Joystream/joystream/issues/853
https://github.com/Joystream/joystream/issues/854
https://github.com/Joystream/joystream/issues/855
https://github.com/Joystream/joystream/issues/856

PR #782, opened 2020-06-24
Feature/nicaea runtime upgrade

gleb-urvanov:
Runtime upgrade test from Constantinople to Nicaea.
Includes https://github.com/Joystream/joystream/pull/769

Issue #685, opened 2020-06-12
Network testing: set a reward for the leader proposal


gleb-urvanov:
Implement network tests for https://github.com/Joystream/joystream/issues/629

Issue #684, opened 2020-06-12
Network testing: set bugdet for the working group proposal


gleb-urvanov:
Implement network tests for https://github.com/Joystream/joystream/issues/628

PR #952, opened 2020-07-15
Network testing: Nicaea proposals- leader management and mint capacity

gleb-urvanov:
implementation for following issues:
https://github.com/Joystream/joystream/issues/683
https://github.com/Joystream/joystream/issues/684
https://github.com/Joystream/joystream/issues/685
https://github.com/Joystream/joystream/issues/657

Depends on:
https://github.com/Joystream/joystream/pull/782
https://github.com/Joystream/joystream/pull/919


mnaamani:
can you merge nicaea to make this easier to review also

tests/network-tests/src/nicaea/tests/proposals/contentWorkingGroupMintCapacityProposalTest.ts
```diff
@@ -0,0 +1,41 @@
...
+import { initConfig } from '../../utils/config';
+import { Keyring, WsProvider } from '@polkadot/api';
+import BN from 'bn.js';
+import { setTestTimeout } from '../../utils/setTestTimeout';
+import tap from 'tap';
+import { registerJoystreamTypes } from '@nicaea/types';
+import { closeApi } from '../impl/closeApi';
+import { ApiWrapper } from '../../utils/apiWrapper';
+
+tap.mocha.describe('Validator count proposal scenario', async () => {
```
mnaamani, line 14: The name of the file, the description and the actual proposal being tested are all out of sync

tests/network-tests/src/nicaea/dto/workingGroupOpening.ts
```diff
@@ -183,10 +182,12 @@ export class WorkingGroupOpening {
     this.openingType = value;
   }
 
-  constructor() {}
+  constructor() {
```
mnaamani, line 185: I think passing in all values to the constructor would be a better pattern than depending on user to call each individueal set method. They may forget to set some values. Unless ofcourse you need to set the values at different times, in which case its probably not a good idea to bundle all the values in this single class.

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
```diff
@@ -0,0 +1,79 @@
...
+  public getSuccessfulApplicationId(): BN {
+    return this.successfulApplicationId;
+  }
+
+  public getWorkingGroup(): string {
+    return this.workingGroup;
+  }
+
+  public setAmountPerPayout(value: BN) {
+    this.amountPerPayout = value;
```
mnaamani, line 36: if you return `this` from each set method the user can more easily chain calls to the set methods, so instead of:

```javascript
let params = new FillOpeningParameters()
params.setAmountPerPayout()
params.setNextPayment()
params.setSomethingElse()
....
```

you can have cleaner/less code:

```javascript
let params = new FillOpeningParameters()
params.setAmountPerPayout()
    .setNextPayment()
    .setSomethingElse()

```

tests/network-tests/src/nicaea/tests/workingGroup/impl/workingGroupModule.ts
```diff
@@ -509,3 +504,90 @@ export async function awaitPayout(apiWrapper: ApiWrapper, membersKeyPairs: Keyri
...
+  apiWrapper: ApiWrapper,
+  leaderAddress: string,
+  module: WorkingGroups
+): Promise<BN> {
+  const leadWorkerId: BN = await apiWrapper.expectLeaderSet();
+  const worker: Worker = await apiWrapper.getWorkerById(leadWorkerId, module);
+  const leaderApplicationId = (await apiWrapper.getApplicationsIdsByRoleAccount(leaderAddress, module))[0];
+  const application: Application = await apiWrapper.getApplicationById(leaderApplicationId, module);
+  assert(
+    worker.role_account_id.toString() === application.role_account_id.toString(),
```
mnaamani, line 522: I think you can do:
```suggestion
    worker.role_account_id.eq(application.role_account_id)
```

If both are of type AccountId

tests/network-tests/src/nicaea/utils/apiWrapper.ts
```diff
@@ -41,6 +42,15 @@ export class ApiWrapper {
     this.api.disconnect();
   }
 
+  public getWorkingGroupString(workingGroup: WorkingGroups): string {
+    switch (workingGroup) {
+      case WorkingGroups.storageWorkingGroup:
+        return 'Storage';
+      default:
+        return 'Undefined';
```
mnaamani, line 50: probably better to throw instead

tests/network-tests/src/nicaea/utils/apiWrapper.ts
```diff
@@ -310,6 +325,156 @@ export class ApiWrapper {
...
+    return this.estimateTxFee(
+      this.api.tx.proposalsCodex.createAddWorkingGroupLeaderOpeningProposal(
+        0,
+        'some long title for the purpose of testing',
+        'some long description for the purpose of testing',
+        0,
+        {
+          activate_at: 'CurrentBlock',
+          commitment: {
+            application_rationing_policy: { max_active_applicants: '32' },
```
mnaamani, line 338: string '32' seems wrong for expected number

tests/network-tests/src/nicaea/utils/apiWrapper.ts
```diff
@@ -682,18 +870,90 @@ export class ApiWrapper {
...
+            resolve((record.event.data as unknown) as BN);
+          }
+        });
+      });
+    });
+  }
+
+  public expectLeaderTerminated(): Promise<void> {
+    return new Promise(async resolve => {
+      await this.api.query.system.events<Vec<EventRecord>>(events => {
```
mnaamani, line 887: in all these expectSomeEvent() methods, there is no need for async/await the way it is written, since we are relying on callback and then calling resolve(). Infact I think the api.query.system.events() method returns a promise for a function to unsubscribe from listening to events, which we should probably use before `resolve`(). As promises work it will only ever resolve to the first event found, but there is unnecessary events still being processed up until the api disconnects.

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
```diff
@@ -0,0 +1,79 @@
+import BN from 'bn.js';
+
+export class FillOpeningParameters {
+  private amountPerPayout!: BN;
```
mnaamani, line 4: my typescript knowledge is a bit rusty, does the ! mean required ? If so isn't it a problem that you can construct the class without giving all these fields a value at construction time?

tests/network-tests/src/nicaea/tests/workingGroup/impl/workingGroupModule.ts
```diff
@@ -274,23 +275,17 @@ export async function fillLeaderOpening(
     payoutInterval,
     module
   );
+
+  // Assertions
   const applicationIdToWorkerIdMap: ApplicationIdToWorkerIdMap = await fillOpeningPromise;
   applicationIdToWorkerIdMap.forEach(async (workerId, applicationId) => {
```
mnaamani, line 281: I missed this on prior PR review, but the assertions here are not sufficient.
You are not checking that any of the application ids filled are in fact the ones we intended to fill.

tests/network-tests/src/nicaea/utils/apiWrapper.ts
```diff
@@ -747,17 +1014,192 @@ export class ApiWrapper {
...
+    workingGroup: string
+  ): Promise<void> {
+    const memberId: BN = (await this.getMemberIds(account.address))[0];
+    return this.sender.signAndSend(
+      this.api.tx.proposalsCodex.createAddWorkingGroupLeaderOpeningProposal(
+        memberId,
+        title,
+        description,
+        proposalStake,
+        {
```
mnaamani, line 1043: The WorkingGroupOpening class can have a method that takes a workingGroup:string argument to construct this struct type

tests/network-tests/src/nicaea/tests/proposals/manageLeaderRole.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  setTestTimeout(apiWrapper, durationInBlocks);
+  membershipTest(apiWrapper, m1KeyPairs, keyring, N, paidTerms, sudoUri);
+  membershipTest(apiWrapper, m2KeyPairs, keyring, N, paidTerms, sudoUri);
+  membershipTest(apiWrapper, leadKeyPair, keyring, 1, paidTerms, sudoUri);
+  councilTest(apiWrapper, m1KeyPairs, m2KeyPairs, keyring, K, sudoUri, greaterStake, lesserStake);
+
+  let createOpeningProposalId: BN;
+  let openingId: BN;
+  tap.test(
```
mnaamani, line 71: I see now the mechanism you mentioned at one point for getting results out of a test. It works well and I suppose its the only way really. But it can be improved. I think we need to do a better approach of encapsulating the calls, context and results of executing them to make things cleaner.

What would be really nice to see is something like this:

```javascript

const context = { apiWrapper, keys, ... }
const params = { applicationStake, roleStake, ... } 
const expectations = { expectFailure, expectedValue/Events }

// asserts that state of runtime is what is expected for execution and prepares the execution to run
// in a tap.test call
const createOpeningFixture = new CreateWorkingGroupLeaderOpeningFixture(context, params)

// runs and asserts on expectations
tap.test('opening', createOpeningFixture.runner(expectations))

// at this point the createOpeningFixture has within it the result of execution, events, the porposalId for example.
// that can be used as inputs for the following tests.
createOpeningFixture.result 
createOpeningFixture.events
createOpeningFixture.openingId 
```

This is what I would have really liked to see as a pattern throughout. This makes constructing and writing scenarios very reusable.

mnaamani (changes requested):
I've left some comments and request for changes.
Overall the tests are running through till the end successfully.
I noticed that the nicaea tests don't cover the content working directory proposals. It is still part of nicaea runtime so they should not be left out. Please correct me if I'm wrong.

tests/network-tests/src/nicaea/dto/workingGroupOpening.ts
gleb-urvanov: Going to disagree with that point- one of the reasons why we have introduced a data transfer object is to mitigate the risk of passing the values in the wrong order: https://github.com/Joystream/joystream/pull/782#discussion_r447783424

If all the values will be passed to the constructor, that goal will not be achieved- instead of risk having the wrong order in the function call, we will have the same risk but for the constructor call.

mnaamani: Yes as in the original comment you linked to I wasn't implying re-introducing a long list of arguments. But this approach here of depending on methods might leave some of the private fields un-initialized?

mnaamani:
I'm going to merge this now to have a the tests in the nicaea branch, but please take a note of the suggested refactoring in the next iteration.

tests/network-tests/src/nicaea/tests/proposals/contentWorkingGroupMintCapacityProposalTest.ts
gleb-urvanov: Done in https://github.com/Joystream/joystream/pull/1001

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
gleb-urvanov: Done in https://github.com/Joystream/joystream/pull/1001

tests/network-tests/src/nicaea/tests/workingGroup/impl/workingGroupModule.ts
gleb-urvanov: Done in https://github.com/Joystream/joystream/pull/1001

tests/network-tests/src/nicaea/utils/apiWrapper.ts
gleb-urvanov: Done in https://github.com/Joystream/joystream/pull/1001

tests/network-tests/src/nicaea/utils/apiWrapper.ts
gleb-urvanov: Indeed. Done in https://github.com/Joystream/joystream/pull/1001

tests/network-tests/src/nicaea/utils/apiWrapper.ts
gleb-urvanov: Done in https://github.com/Joystream/joystream/pull/1001 , although it is worth to look at the solution. With `unsubscribe` introduction we have to await for unsubscription promise, thus the `await` is gaining sense. Please correct me if im wrong.

tests/network-tests/src/nicaea/tests/workingGroup/impl/workingGroupModule.ts
gleb-urvanov: Done in #1001

tests/network-tests/src/nicaea/utils/apiWrapper.ts
gleb-urvanov: Done in #1001

tests/network-tests/src/nicaea/tests/proposals/manageLeaderRole.ts
gleb-urvanov: Made a note for a future refactoring in Iznik

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
gleb-urvanov: With ! we asserting the value will be assigned before it will be used. As alternative, we could initialize the values with default values, but the object is not intended to be used without all fields being initialized.

tests/network-tests/src/nicaea/dto/workingGroupOpening.ts
gleb-urvanov: Yes, it might. Normally typescript won't allow that, that is why assignment assertion (exclamation sign) is used. What is the best approach you can suggest?

mnaamani: Yes so to this point exactly, I guess the approach here mainly solves problem of passing arguments in wrong order.
So passing an object was the correct solution, except now this new object can have null or underfined fields. So wouldn't simply passing arguments like so be the optimal approach:

```javascript
// original approach which I advised against
function withHugeArgsList(a, b, c, ...., z) {

}

// suggested method, takes a single object instead
function theMethod({ arg_a: A, arg_b: B, arg_c: C }) {
   do_something_with_args()
}

// calling the method
theMethod({arg_a: valueOfTypeA, arg_b: valueOfTypeB, ...})
```

This way will insure no incorrect ordering of arguments, and that all args are provided (optionally give some args a default value if you want them to be optional)

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
mnaamani: ok thanks for clarification. So we are infact using the `!` incorrectly, the fields should then be initialized in the constructor. No ?

tests/network-tests/src/nicaea/utils/apiWrapper.ts
mnaamani: Well yes if you are actually going to use it, then you need the await :) 
And I see you did refactor it in #1001 as you mentioned. :+1:

tests/network-tests/src/nicaea/dto/workingGroupOpening.ts
gleb-urvanov: Ok, will do.

tests/network-tests/src/nicaea/dto/fillOpeningParameters.ts
gleb-urvanov: With initialization in the constructor, the assignment assertion will be excessive (since the valued are guaranteed to be initialized). So, when the values will be initialized in the constructor, ! will be removed.

PR #954, opened 2020-07-15
Query node indexer start options

metmirr:
This PR adds the option to start indexer from a block number. Block number is passed as an argument:

```
yarn start --height 123
```

If block number is not provided either block number is `0` or retrieved from the database.

_Note_: `Bootstrapper.ts` file update is mostly white space removal so can be ignored.

### DIFF
--- a/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
+++ b/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
@@ -75,7 +75,7 @@ function banner():Command {
         .description("Joystream Indexer")
         .option('-b, --bootstrap', 'Load initial data using the boot* mappings')
         .option('-l, --logging <file>', 'Path to log4js config', null)
-        .option('-h, --height <height>', 'Block height to start from', 0)
+        .option('-h, --height <height>', 'Block height to start from')
         .option('-e, --env <file>', '.env file location', '../../.env')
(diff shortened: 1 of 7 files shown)


dzhelezov (approved):

dzhelezov:
> If block number is not provided either block number is 0 or retrieved from the database.

What happens if the provided height is greater than the one in the database?

metmirr:
If block height is provided then the block height in the database is ignored. I thought it as a forcing option.

dzhelezov:
> If block height is provided then the block height in the database is ignored. I thought it as a forcing option.

Isn't it a dangerous situation? Assume the indexer was stopped at block 200, and then started at block height 100. The events from block 100 and further on will be applied to the current state in the database (which is at block 200 and out sync). Probably one should just throw an error and allow a custom block height only if there's no previous history in the database.

metmirr:
> > If block height is provided then the block height in the database is ignored. I thought it as a forcing option.
> 
> Isn't it a dangerous situation? Assume the indexer was stopped at block 200, and then started at block height 100. The events from block 100 and further on will be applied to the current state in the database (which is at block 200 and out sync). Probably one should just throw an error and allow a custom block height only if there's no previous history in the database.

You are right, it is dangerous. Thanks, will do a fix!

bedeho (approved):
LGTM.

Issue #651, opened 2020-06-10
Pioneer: Runtime Proposal page heavy lag + proposed fix


mochet:
Currently the Runtime Proposal page causes a lot of lag for most users, this is because it reportedly calculates the hash on the browser on page load.
A user usually has to wait upwards of 5 minutes for the page to stabilize, even if they want to just check the status of current votes. The page becomes almost completely unresponsive while this happens.

An example is the current runtime proposal vote here: https://testnet.joystream.org/#/proposals/14

This is preventing users from voting or turning away users from voting on the current runtime upgrade proposal as they think the page is broken.

One suggestion to fix this issue would be to make the hash calculation only happen after a user has clicked a button, this is done on some routers, like the below pictures demonstrate:
![image](https://user-images.githubusercontent.com/50830161/84248355-5f06c500-ab1a-11ea-812a-1c1c9d95c5d9.png)
![image](https://user-images.githubusercontent.com/50830161/84248362-60d08880-ab1a-11ea-8b9a-11ad54cb7e5f.png)

bedeho:
I think this is a great idea, perhaps @Lezek123 can fix?

mnaamani:
In addition to the delay in calculating the hash, I expect there is also a short delay in downloading the ~1.4MB code as well, but I'm not sure how significant that is, it would depend on the user's bandwidth and specs.

Consider using background workers to do the check asynchronously?

mnaamani:
In future we might even have to consider storing the code in a separate state storage so the code can be downloaded only if user chooses to verify.

Lezek123:
I think this lag may be caused by the call to `api.query.proposalsCodex.proposalDetailsByProposalId(RUNTIME_PROPOSAL_ID)` itself (which can be also observed on https://testnet.joystream.org/#/chainstate currently with proposal id === `14`).

The details of runtime updrade proposal consist only of the full WASM file binary data, which has to be downloaded and decoded when this call is made. We can probably skip making this call until the user asks to verify the data (ie. using the button as proposed).

Lezek123:
I consider it fixed in https://github.com/Joystream/joystream/pull/957 (although the possibilities for optimalization are still kind of limited without the query node)

PR #957, opened 2020-07-16
Pioneer: Runtime upgrade proposal optimalization

Lezek123:
This PR fixes the problem with major lags on `RuntimeUpgrade` proposal details page which was raised in issue https://github.com/Joystream/joystream/issues/651

The fix also benefits proposal module in general, as it caches details for the `RuntimeUpgrade` proposal throught routes, which helps us avoid repeating a few time-consuming/computation-heavy tasks like fetching `RuntimeUpgrade` proposal details or re-calculating the WASM bytecode hash.

Further optimalizations are possible, but this already reduces the loading time greatly (especially when moving from one proposal-related page to another), gets rid of post-loading lags and seems like a good, quick optimalization in general.


pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -61,10 +64,44 @@ export default class ProposalsTransport extends BaseTransport {
     return (this.api.consts.proposalsEngine.cancellationFee as BalanceOf).toNumber();
   }
 
+  private runtimeUpgradeProposalCachedDetails (id: ProposalId) {
```
mnaamani: The name of the method `runtimeUpgradeProposalCachedDetails` and the private property `runtimeUpgradeProposalDetailsCache` are too similar and confused me a but while reading the code.
perhaps rename the method:
 `getRuntimeUpgradeProposalCachedDetails`

pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -61,10 +64,44 @@ export default class ProposalsTransport extends BaseTransport {
...
+    return this.runtimeUpgradeProposalDetailsCache[id.toNumber()];
+  }
+
+  private runtimeUpgradeProposalDetails (id: ProposalId, wasm: Bytes): [string, number] {
+    const cachedDetails = this.runtimeUpgradeProposalCachedDetails(id);
+
+    if (cachedDetails) {
```
mnaamani: this will never be truthy because it is only called by `proposalById()` on cache miss.

It also suggests the name of the method should be `updateRuntimeUpgradeProposalDetails` or `computeAndCacheRuntimeUpgradeProposalDetails`

mnaamani (changes requested):

mnaamani:
Tested works well, just had some suggestions.

There is a merge conflict to fix.

Also I'm wonder if you considered using the `APIQueryCache` as you did in the joy-roles app. Would it help or make things worse in caching the WASM blob in memory?

mnaamani:
Does it make sense to make a similar patch for master branch to benefit from this before the release?

Lezek123:
> Also I'm wonder if you considered using the APIQueryCache as you did in the joy-roles app. Would it help or make things worse in caching the WASM blob in memory?

`APIQueryCache` only caches the result of given query at given block and is reset every time there's a new block, so we can use it as an additional optimalization for the transports in case some component gets reloaded and for some reason tries to re-execute the samie queries that were already executed at given block.

> Does it make sense to make a similar patch for master branch to benefit from this before the release?

I think it does! Will prepare another PR to master

Lezek123:
- Fixed conflicts
- Added query cache for `joy-utils` transports
- Fixed a bug with transport instance beeing re-constructed throught proposal routes (which ie. prevented caching details when moving between `/proposals` and `/proposals/ID` routes)
- Implemented caching for all proposal details (not just runtime upgrade)
- Made the proposal details caching code cleaner

pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -112,13 +149,9 @@ export default class ProposalsTransport extends BaseTransport {
...
 
-  async proposalDetails (id: ProposalId) {
-    return this.proposalsCodex.proposalDetailsByProposalId(id);
-  }
-
   async voteByProposalAndMember (proposalId: ProposalId, voterId: MemberId): Promise<VoteKind | null> {
-    const vote = await this.proposalsEngine.voteExistsByProposalByVoter<VoteKind>(proposalId, voterId);
-    const hasVoted = (await this.proposalsEngine.voteExistsByProposalByVoter.size(proposalId, voterId)).toNumber();
+    const vote = (await this.proposalsEngine.voteExistsByProposalByVoter(proposalId, voterId)) as VoteKind;
+    const hasVoted = (await this.api.query.proposalsEngine.voteExistsByProposalByVoter.size(proposalId, voterId)).toNumber();
```
mnaamani, line 154: I see you are bypassing the cached api here, why exactly?

pioneer/packages/joy-utils/src/transport/council.ts
```diff
@@ -8,21 +8,22 @@ import { FIRST_MEMBER_ID } from '../consts/members';
...
+    super(api, cacheApi);
     this.membersT = membersTransport;
     this.chainT = chainTransport;
   }
 
   async councilMembersLength (atBlock?: number): Promise<number> {
     if (atBlock) {
       const blockHash = await this.chainT.blockHash(atBlock);
-      return ((await this.council.activeCouncil.at(blockHash)) as Seats).length;
+      return ((await this.api.query.council.activeCouncil.at(blockHash)) as Seats).length;
```
mnaamani, line 26: bypassing the cached api? Any special reason?

pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -177,7 +210,7 @@ export default class ProposalsTransport extends BaseTransport {
   }
 
   async subscribeProposal (id: number|ProposalId, callback: () => void) {
-    return this.proposalsEngine.proposals(id, callback);
+    return this.api.query.proposalsEngine.proposals(id, callback);
```
mnaamani, line 213: this one here I understand because you want to subsribe to updates..?

mnaamani (approved):
Looks good, just had a query on why you were bypassing the cached api in a few distinct places.
Tested against staging network works great

pioneer/packages/joy-utils/src/transport/proposals.ts
Lezek123: Cached api doesn't have `.size()` available

pioneer/packages/joy-utils/src/transport/council.ts
Lezek123: Cached api doesn't have `.at()` available

pioneer/packages/joy-utils/src/transport/proposals.ts
Lezek123: Yes, this is also not available via cached api. Only simple queries are supported.

mnaamani: ok thanks for clarification

PR #958, opened 2020-07-16
Rebase on common linting rules, fix linting errors, format with Prettier

kdembler:
This PR finalizes linting/formatting work we've been doing based on common Joystream linting rules. It fixes any ESLint/Prettier issues and fixes lint git hooks.


fcescob (approved):
I gave another look at config files, skipped all the other one since we did those in a pair programming session. LGTM!

fcescob (approved):

fcescob (approved):

.editorconfig
```diff
@@ -1,16 +1,14 @@
+# In case prettier plugin or eslint with autofix is not enabled in IDE
+# The fallback settings here should match with our prettierrc config
+# so we get consistency!
```
mnaamani, line 3: It will be a headache to maintain consistency although I don't expect these basic configs to change.
So in the rest of the repo I chose not to have and .editorconfig (which exception of pioneer)

mnaamani: Since .editorconfig doesn't have the same "coverage" as prettier, a developer that doesn't use the prettier or eslint plugin will inevitably have to run the formatting script and fix. But as a fallback it doesn't conflict with someone using prettier/eslint in the IDE since the prettier plugin will look at the prettier config files before .editorconfig I don't see the harm in keeping it.

mnaamani (approved):

mnaamani:
Tested linting, building :+1:

PR #960, opened 2020-07-16
My roles - add Worker ID

Lezek123:
SSIA

### DIFF
--- a/pioneer/packages/joy-roles/src/tabs/MyRoles.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/MyRoles.tsx
@@ -129,6 +129,7 @@ export const CurrentRoles = Loadable<CurrentRolesProps>(
             <Table.Header>
               <Table.Row>
                 <Table.HeaderCell>Role</Table.HeaderCell>
+                <Table.HeaderCell>Worker / Curator ID</Table.HeaderCell>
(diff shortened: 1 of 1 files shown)


mnaamani:
Is it normal not to see my role as content workinging group lead if I set it via sudo rather than going through hiring process?

mnaamani:
> Is it normal not to see my role as content workinging group lead if I set it via sudo rather than going through hiring process?

Same member id is also storage load, assigned through hiring process, and that showed up in my roles correctly

Lezek123:
> Is it normal not to see my role as content workinging group lead if I set it via sudo rather than going through hiring process?
> Same member id is also storage load, assigned through hiring process, and that showed up in my roles correctly

Storage Working Group lead is quite different, because technically he's part of the "workers" group, so it's easy to include this role in `my-roles`. Including Content Working Group lead role would require some addiotional handling, since it's kind of "separated", doesn't have `workerId` etc.

It can be done, but not sure if it's necessary.

mnaamani:
okay thanks for the clarification.

I think right now its not necessary.

mnaamani:
LGTM will merge unless @bwhm has any additional issue re showing content working group lead role info

Screen Shot 2020-07-16 at 6 39 31 PM

mnaamani (approved):

PR #961, opened 2020-07-16
Add Tests 

fcescob:
This PR adds a render test to each component, and adds code coverage to the codebase.
I've added a check to ensure coverage is above the current coverage minus `2%` so if a new component or function are added, they must be tested or the test suite will fail. 

Relevant issues: #748


kdembler:
@Gamaranto needs rebase

atlas/packages/app/config-overrides.js
```diff
@@ -56,4 +56,20 @@ module.exports = {
     paths.appBuild = path.resolve(__dirname, '..', '..', 'dist')
     return paths
   },
+  jest: function (config) {
+    config.coverageThreshold = {
+      global: {
+        statements: 48,
```
kdembler: Where do these values come from? I've seen you've mentioned that these should be current values - 2, but I don't see that reflected. When I run coverage I get the following values instead 68.66, 45.64, 57.26, 70.52.

Also, more importantly, I'm not sure we should enforce coverage at this point in development. I feel like we can enforce testing some more complex pieces of logic but I don't think we should spend time on writing tests for every component

kdembler (commented):
I'm not exactly sure what's the goal of these tests. I see that we are trying to do snapshot testing but there are some issues:

Firstly, our snapshots don't really work. If you look at a snapshot file you will see that the snapshots aren't really saved:
```
exports[`HamburgerButton component Should render Hamburger correctly 1`] = `ReactWrapper {}`;
```
So what I think actually happens is that we are just rendering the component and don't care about any of its outputs as long as it renders.

Secondly, I'm not sure we should do properly done snapshot tests at this point. I expect that during this period of development we may end up changing the components quite often so snapshots would be pretty inconvenient to use.

To summarise, I think that at this point we should just write shallow render tests for components (that's actualy what happens now, but we shouldn't pretend those are snapshots).

Please let me know what you think or whether's something I've said is unclear.

fcescob:
Good points. I will change the tests to be just shallow render tests, regarding coverage, I tend to agree here too that it might be pointless to have it now, so I will remove it for now, I do think it would be valuable to have this, closer to launch or after v1 is out.

atlas/packages/app/src/shared/__tests__/Button.test.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react'
 import { mount } from 'enzyme'
-import { Button } from '../components'
+import Button from '@/shared/components/Button'
```
kdembler, line 3: This is inside the `shared` package so I think it may be a better idea to use a relative import here

kdembler (approved):
Looks good. If I throw an error inside a component function, the tests fail so that's a good thing. I think we're more than fine for now with the testing. Just wondering whether you've looked at https://github.com/testing-library/react-testing-library. Approved!

bedeho (approved):

PR #962, opened 2020-07-16
Membership module refactoring.

shamil-gadelshin:
The new incarnation of [this PR](https://github.com/Joystream/joystream/pull/490). Merged with the Nicaea release branch.


runtime-modules/content-working-group/src/tests.rs
```diff
@@ -1184,6 +1183,7 @@ impl SetLeadFixture {
         let new_lead = LeadById::<Test>::get(new_lead_id);
 
         let expected_new_lead = Lead {
+            member_id: 0,
```
mnaamani:
```suggestion
            member_id: self.member_id,
```

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,658 @@
...
+        + SimpleArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + PartialEq
+        + Ord;
+
+    /// Initial balance of members created at genesis
+    type InitialMembersBalance: Get<BalanceOf<Self>>;
```
mnaamani: lets as well drop this now. The way we import members at genesis needs to be improved and one balance for all members isn't very useful.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,658 @@
...
+struct ValidatedUserInfo {
+    handle: Vec<u8>,
+    avatar_uri: Vec<u8>,
+    about: Vec<u8>,
+}
+
+#[derive(Encode, Decode, Debug, PartialEq)]
+pub enum EntryMethod<PaidTermId, AccountId> {
+    Paid(PaidTermId),
+    Screening(AccountId),
```
mnaamani, line 133: Another place where we can change to identify the Screening role with a memberid instead of an account id

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,658 @@
...
+// Default paid membership terms
+pub const DEFAULT_PAID_TERM_ID: u32 = 0;
+
+// Default user info constraints
+const DEFAULT_MIN_HANDLE_LENGTH: u32 = 5;
+const DEFAULT_MAX_HANDLE_LENGTH: u32 = 40;
+const DEFAULT_MAX_AVATAR_URI_LENGTH: u32 = 1024;
+const DEFAULT_MAX_ABOUT_TEXT_LENGTH: u32 = 2048;
+
+pub type MembershipOf<T> = Membership<
```
mnaamani: We've used this naming a few times in the library, but it makes the public api look odd with this `Of` suffix.
Its probably better to keep the public type that will appear in the api to be `Membership` and then the internal structure used can be named something different, similar to what I did here with [DataObject](https://github.com/Joystream/joystream/blob/7a350f81fea0b343c50403e2efe0c64cc41822dd/runtime-modules/storage/src/data_directory.rs#L127)

Also as I'm reviewing this, and thinking about public type names, I know in the latest substrate there is a membership pallet and I'm sure we might get some name clashes with types there. I guess we will cross that bridge when we get to it.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,658 @@
...
+                let user_info = ValidatedUserInfo {
+                    handle: handle.clone().into_bytes(),
+                    avatar_uri: avatar_uri.clone().into_bytes(),
+                    about: about.clone().into_bytes()
+                };
+
+                <Module<T>>::insert_member(&who, &user_info, EntryMethod::Genesis);
+
+                // Give member starting balance
+                T::Currency::deposit_creating(&who, T::InitialMembersBalance::get());
```
mnaamani: I think we should drop this initial balance, and later update the build() method to also take an initial balance for each account.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,9 +1,658 @@
...
+
+        Ok(ValidatedUserInfo {
+            handle,
+            avatar_uri,
+            about,
+        })
+    }
+
+    fn insert_member(
+        who: &T::AccountId,
```
mnaamani, line 539: In the past I think it was a mistake to only take one account in the arguments and set it as both root and controller.
We should pass both separately. This also means we will need to capture these two accounts in `buy_membership` and `add_screened_member`. On Pioneer side we can keep UI simple, and still provide the same account for both, and let user later change their controller account. But at least in the runtime it shouldn't force this.

runtime-modules/roles/src/role_types.rs
```diff
@@ -0,0 +1,22 @@
...
+    Curator,
+}
+
+/// Must be default constructable because it indirectly is a value in a storage map.
+/// ***SHOULD NEVER ACTUALLY GET CALLED, IS REQUIRED TO DUE BAD STORAGE MODEL IN SUBSTRATE***
+impl Default for Role {
+    fn default() -> Self {
+        Self::Curator
+    }
+}
```
mnaamani: do we event need the `Role` type anymore ? I can't find a reference to it anymore

runtime/src/integration/proposals/membership_origin_validator.rs
```diff
@@ -27,7 +27,7 @@ impl<T: crate::members::Trait>
         let account_id = ensure_signed(origin)?;
 
         // check whether actor_id belongs to the registered member
-        let profile_result = <crate::members::Module<T>>::ensure_profile(actor_id);
+        let profile_result = <membership::Module<T>>::ensure_membership(actor_id);
```
mnaamani, line 30: maybe rename actor_id to member_id as we are dropping the 'Actor' terminology

runtime/src/lib.rs
```diff
@@ -735,7 +734,7 @@ construct_runtime!(
         CouncilElection: election::{Module, Call, Storage, Event<T>, Config<T>},
         Council: council::{Module, Call, Storage, Event<T>, Config<T>},
         Memo: memo::{Module, Call, Storage, Event<T>},
-        Members: members::{Module, Call, Storage, Event<T>, Config<T>},
+        Members: membership::{Module, Call, Storage, Event<T>, Config<T>},
```
mnaamani, line 736: headsup @Lezek123 all `api.query.members.*` and `api.query.tx.members.*` in polkadot-js apps will need to be updated to `api.*.membership.*` after this change.

mnaamani (changes requested):
Nice to see this great refactoring.
Although large, PRs where code is mostly removed seems fundamentally easier to review :)

I think I left one small change request, otherwise looks good to me.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Ok

runtime-modules/content-working-group/src/tests.rs
shamil-gadelshin: Ok

runtime-modules/roles/src/role_types.rs
shamil-gadelshin: Merge artifact. Deleted.

runtime/src/integration/proposals/membership_origin_validator.rs
shamil-gadelshin: It's not related. I created these *validators structs during the work on the proposals. This `actor` could be the proposal author or the discussion thread author. So `actor` is just an abstraction.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: I'm not sure how can we change the EntryMethod to Membership to use member_id because we don't have it until we create the membership.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Renamed to MembershipObject and Membership.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: I think it is out of the scope for this task. Please, create a new issue - and we could deal with it during the work on the governance module.

runtime-modules/membership/src/lib.rs
mnaamani: The variant would still be `Screening(MemberId)` and the member id is that of the member who is the designated screening authority,  set by the council or sudo. So this involves also changing the the storage value `ScreeningAuthority` to `Option<MemberId>` and updating the signature of `fn add_screened_member()` dispatch.

I think we can deal with this at a later time when we add "screeing authority working group"..

runtime-modules/membership/src/lib.rs
mnaamani: Yes you are probably right here. There may even be an existing issue.. will look for it.

runtime/src/lib.rs
Lezek123: It seems like `@polkadot/api` still sees it as `members` module (ie. I can call `api.query.members.nextMemberId()`, but cannot `api.query.membership.nextMemberId()`)

PR #945, opened 2020-07-15
Pioneer: Remaining Nicaea proposals

Lezek123:
## Covered issues:
- https://github.com/Joystream/joystream/issues/725 - add SetWorkingGroupMintCapacity, remove SetStorageRoleParams, EvictStorageProvider
- https://github.com/Joystream/joystream/issues/717 - add BeginReviewWorkingGroupLeaderApplications
- https://github.com/Joystream/joystream/issues/718 - add FillWorkingGroupLeaderOpening
- https://github.com/Joystream/joystream/issues/719 - add SlashWorkingGroupLeaderStake and DecreaseWorkingGroupLeaderStake
- https://github.com/Joystream/joystream/issues/720 - add SetWorkingGroupLeaderReward
- https://github.com/Joystream/joystream/issues/724 - add TerminateWorkingGroupLeaderRole

## Other noticable changes:
**Validation schema**
Introduced some `validationSchema.ts` changes and normalization, allowing us to:
- Get the benefit of auto-suggested form field names etc., without having to define them spearately in 2 places (which always opens the room for errors)
- Easy way of providing additional parameters to validation schema from proposal's `FromContainer` component (ie. current block, constraints fetched from the runtime etc.)

**joy-utils**
The `joy-utils` package, which I planned to later extract out from Pioneer, so that other tools (ie. the CLI) can take advantage of it, now has some new useful functionality (transport, formatting, react components etc.), which I introduced there due to the fact that this package was already beeing used by `joy-proposals`. Unfortunately currently it's the 3rd place (I know of) which handles fetching working-groups related data (the other two are `joy-roles`' package transport and the CLI's `Api.ts`), but I hope this can be normalized soon (preferably after `content-working-group` is refactored to use the normalized `working-groups` module).

The immediate benefit of this though is that now we can, for example, use components that handle:
- Displaying wokring group leader data (`LeadInfo`)
- Displaying working group opening applications data and possibly even support filling an opening (`ApplicationDetails`)

Anywhere within Pioneer without creating dependency hell (since `joy-utils` is already pretty well separated). Currently we can, for example, use them allow viewing the applications from the perspecitve of openings view (`/opportunities`) etc. (will only work for `storageWorkingGroup` though)

**Parameters layout**
[...]

PR #963, opened 2020-07-16
@joystream/types - checks config, CI, linter fixes

Lezek123:
This PR:
- Adds yarn `lint`, `format` and `checks` (which also includes circular dep. check) scripts for `@joystream/types`
- Configures linter in the `@joystream/types` workspace and turns off some case-related rules that would require breaking changes
-  Fixes the remaining linter issues
- Adds CI checks specifically for `@joystream/types` and removes no-longer neccessary `madge --circular` check from other projects' CI checks.

Prettier is not yet configured, as there are some incoming changes in https://github.com/Joystream/joystream/pull/945 and https://github.com/Joystream/joystream/pull/884 which would cause unnecessary conflicts.

In order to enable prettier we would just need to change `.prettierignore` file to:
```
**/*.js
**/*.d.ts
!.eslintrc.js
hiring/schemas/role.schema.json
lib/
build/
```
_(This is based on .gitignore, already tested locally)_


types/.gitignore
```diff
@@ -1,6 +1,7 @@
 # Don't track build artifacts
 **/*.js
 **/*.d.ts
+!.eslintrc.js
```
mnaamani, line 4: what does the `!` here do?

mnaamani (commented):
Looks good, left a question about what exclamation mark in .gititnore means

I ran yarn checks and it is reporting a circular dependency:

```
MokoBook:types mokhtar$ yarn checks
yarn run v1.22.4
$ yarn build && madge --circular ./ && yarn lint && prettier ./ --check
$ tsc --build tsconfig.json
Processed 73 files (866ms) 

✖ Found 2 circular dependencies!

1) lib/content-working-group/index.js > lib/index.js
2) lib/index.js > lib/members.js

error Command failed with exit code 1.
```

Although in the github checks it doesn't seem to give the same result

mnaamani:
> Although in the github checks it doesn't seem to give the same result

Of course these were leftover artifacts from prior types library when output was being built into the `lib/` directory so this is a non issue

mnaamani:
Can we add a step to do the following:

```
npm pack
tar xzf  joystream-types-0.11.0.tgz
cd pacakge/
npm install
node ./index.js
```

As a sanity check that the published package will work as expected

types/.gitignore
Lezek123: It means this file is prevented from beeing ignored even if it's matched by some previous rule (`**/*.js` in this case)

mnaamani: ok

Lezek123:
I think we would need a workaround that will work for all versions, so we don't have to change the version in `tar xzf  joystream-types-0.11.0.tgz` each time manually. I'll try to look into that.

Lezek123:
Since the last line of output of `npm pack` should be the filename I think we can achieve this with:
`npm pack | tail -1 | xargs tar xzf`

Lezek123:
Added `npm pack` CI check and prettier formatting

mnaamani:
> Since the last line of output of `npm pack` should be the filename I think we can achieve this with:
> `npm pack | tail -1 | xargs tar xzf`

cool :+1:

PR #966, opened 2020-07-17
fix broken call to create identities api after linter autofix?

mnaamani:
- Small bug fix with passing keyFile option to create identities api
- discovery command, doesn't require keyFile, publicUrl, and providerId arguments


shamil-gadelshin (approved):
LGTM

PR #967, opened 2020-07-17
Feature: Hash commitment property type

### LINKED ISSUE #885: Hash commitment property type
# Background

For many types of data, such as the long description of a video, we do not need to have the data available in storage for any kind of processing. Here it is also not needed to be able to edit it it as a vector. In these cases, its sufficient for the storage to just hold a hash commitment over the raw data. The query node will be able to provide all values to user interfaces by grabbing the latest values from transaction/event history processing. Removing all these property values from the storage means that the content directory takes up much less space on full nodes, and it may even be feasible to do certain kinds of storage migrations in the future. The hash can be used for light clients to verify data provided by query nodes. Hence I propose

# Proposal

- Introduce a new property type `Hash` which accepts user input `Vec<u8>`, and it has an optional size limit defined in the type. Runtime computes a hash and stores it in storage.
- Introduce `Vec<Hash>` the same way.
- We still keep all other types for now, such as raw `Vec<u8>` and `Vec<Vec<u8>>`, in case people for some reason want to be able to treat such data as vectors, but I don't know if we will keep long term.

iorveth:
The scope of this PR includes hash commitment property type core logic implementation


bedeho (approved):
Looks good!

Can we address this in a later PR:

I think the name `StoredPropertyValue` is better than `output...`, because it reflects that this is the type we use to _store_ a property value in storage, sayings its "output" isn't really that clear about its role.

PR #968, opened 2020-07-17
Query node fetch block fix

metmirr:
When polkadotjs API connects to a Substrate/Polkadot chain it registers types for the latest block on the chain. Fetching old blocks becomes problematic because API tries to decode blocks with different polkadot types. The PR ensures metadata for each block before parsing the block.

To minimize merge conflict potential, this PR assumes #954 will be merged first.


dzhelezov (approved):
Looks good!

bedeho (approved):
LGTM.

PR #969, opened 2020-07-17
Dev proposal settings

mnaamani:
When running development chain we normally need 0 grace periods when testing. Instead of manually editing chainspec file after generating it to modify the default values, we can construct the chainspec file with appropriate values instead.


runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -802,7 +802,7 @@ fn set_default_proposal_parameters_succeeded() {
         // nothing is set
         assert_eq!(<SetValidatorCountProposalVotingPeriod<Test>>::get(), 0);
 
-        ProposalCodex::set_default_config_values();
+        ProposalCodex::set_config_values(p);
 
         assert_eq!(
```
mnaamani, line 807: I noticed while working on this PR that not all the config values are actually checked.

node/src/chain_spec.rs
```diff
@@ -140,6 +141,7 @@ impl Alternative {
                             get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
                             get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
                         ],
+                        node_runtime::ProposalsConfigParameters::development(),
```
shamil-gadelshin: Should it be "testing()" here?

runtime-modules/proposals/codex/src/proposal_types/mod.rs
```diff
@@ -348,3 +349,54 @@ impl Default for ProposalsConfigParameters {
         }
     }
 }
+
```
shamil-gadelshin, line 352: I suggest to leave with_grace_and_voting_periods() here and move the exact config types (development and testing) closer to the usage - in the node project probably in the separate file (proposal config) because testing and development grace and voting periods are likely to change more often than the rest of the code here.

runtime-modules/proposals/codex/src/tests/mod.rs
shamil-gadelshin: I'll create a fix for the Nicaea.

shamil-gadelshin (changes requested):

node/src/chain_spec.rs
mnaamani: I considered that, but for network integration testing @gleb-urvanov needs 0 grace periods :)

runtime-modules/proposals/codex/src/proposal_types/mod.rs
mnaamani: Thats a good suggestion 👍 will do

mnaamani: fixed in 2f8e54bdfb789beb7230b425ce84d8d175545046

shamil-gadelshin (approved):
LGTM

PR #971, opened 2020-07-17
CLI - prepare for publishing

Lezek123:
This PR prepares the CLI for beeing published into the npm:
- Updated `README.md`
- Added a few plugins based on #1014: `@oclif/plugin-autocomplete`, `@oclif/plugin-not-found`, `@oclif/plugin-warn-if-update-available`

Tested with `npm link`.
Unfortunately it turned out that `@oclif/plugin-autocomplete`:
1. Requires manual user installation (`$ joystream-cli autocomplete`)
1. Does not handle subcommands very well (see: https://github.com/oclif/plugin-autocomplete/issues/9), so it's only useful for autocompleting subcomand "family" names (ie. `working-groups`)

## Testing:
Probably the best way to test how the CLI behaves when it's installed by the user is to run `npm link` inside `/cli`. We can then use `$ joystream-cli` to run any command.


yarn.lock
```diff
@@ -2584,6 +2584,10 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
+"@nicaea/types@link:types":
```
Lezek123, line 2587: This seems to have happened (again) after executing `yarn workspace joystream-cli add ...`

Lezek123: I don't think it affects anything, so if yarn just keeps adding this entry perhaps we can leave it as is?

Lezek123:
Merged changes from https://github.com/Joystream/joystream/pull/983 and updated the README

Lezek123:
- Changed the package name to `@joystream/cli`
- Changed the version to `0.1.0`
- Implemented required adjustments in README and `StateAwareCommandBase` (app data folder package name normalization)
- Added `npm pack` + `npm link` test to CLI CI checks, to make sure the tarball works as expected (those checks can be observed here: https://github.com/Lezek123/substrate-runtime-joystream/runs/920071207)

mnaamani (approved):

Issue #819, opened 2020-06-26
Algebraic field types


bedeho:

dzhelezov:
Implemented in https://github.com/Joystream/joystream/pull/974, submitted for review

bedeho:
Done.

Issue #709, opened 2020-06-16
Algebraic types II


bedeho:
title: 

# Background

As already explained in the background section here

https://github.com/Joystream/joystream/issues/554

We have a problem domain which has algebraic types all over the place, and it would be a big benefit if they could be reflected neatly in our query infrastructure. As explained in that first post, doing it deeply (approach 4), is ideal, however a later post points out

https://github.com/Joystream/joystream/issues/554#issuecomment-640254649

The resulting GraphQL API must generate OpenCRUD like capabilities that tie into matching these algerbraic types, and this is a non-trivial task.

**This issue attempts to clarify how this can be done.**

# Proposal

# Input schema

# GraphQL API

# Database

dzhelezov:
Supporting variant types requires different design trade-offs in order to address both Postgres and GraphQL limitations.

The main limitation at the GraphQL side is the lack of union input types: https://github.com/graphql/graphql-js/pull/1196
At the database (e.g. Postgres) level the fundamental limitation is the "flat" structure of the tables, that is, the set of columns = entity properties is fixed by the table schema.

There are no mutations, and the API should offer a set of self-describing queries. A typical query returns a list of entities (more precisely, a specified set of fields) filtered by the `where` clause. 

## A query per variant type

The approach described above suggests reflecting the type system by defining a lot of separate queries for each combination of the fields the fully flattened entity may have. In the example above, the flattened version of`EntityA.status` have fields `hates, isMale, bank, father, mother`, coming from the different realizations of the `status` field. The main advantage of having a separate query for each possible realization is the "code completion" feature provided for each input type, tailored to each variant of the union type. The drawback is that the number of queries explodes if there are entities with multiple fields of algebraic type and/or if the input types form a complex tree as in the example above.

## A bag of input fields approach

An alternative approach is to combine all the possible input fields into single 'JUMBO-INPUT' type:
```gql
EnitityAWhereInput {
  hates_eq: String,
  isMale_eq: boolean,
  bank: String,
  father_hates_eq: String,
  father_isMale_eq: String, 
  mother_hates_eq: String,
  mother_isMale_eq: String,
  ...
}
```
Many combinations of the input fields are impossible, but that would simply return an empty list. 
The benefit of this approach is a conceptual and implementation simplicity (both at the Warthog-middleware and TypeORM level). The drawback is that the API consumer has to have an `out-of-band` knowledge of which input fields are actually compatible.

PR #974, opened 2020-07-20
Add algebraic types support for query nodes

dzhelezov:
This PR implements as described in https://github.com/Joystream/joystream/issues/709

In a nutshell, the following constructs are supported in the input schema:

```gql
type Miserable @variant {
  hates: String!
}

type HappyPoor @variant {
  isMale: boolean
}

union Poor = HappyPoor | Miserable

type MiddleClass @variant {
  father: Poor
  mother: Poor
}

type Rich @variant {
  bank: EntityC
}

union Status = Poor | MiddleClass | HappyPoor | Miserable

type Account @entity {
  id: ID!
  status: Status!
}
```

Such a construct would generate a query of the type 
```gql
query {
	accounts(limit: 5, orderBy: about_ASC, where: { status_json: { father: { isMale_eq: true }} }) {
    about
    status {
      __typename 
      ... on MiddleClass {
        father {
          ... on HappyPoor {
            isMale
          }
          ... on Miserable {
            hates
          }
        }
      }
    }
  }
}
```

Project Issue: https://github.com/Joystream/joystream/issues/819


query-node/substrate-query-framework/cli/src/templates/entities/model.ts.mst
```diff
@@ -91,5 +102,16 @@ export {{#isInterface}}abstract{{/isInterface}} class {{className}}
       {{#description}}description: `{{{description}}}`{{/description}} })
     {{camelName}}{{^required}}?{{/required}}{{#required}}!{{/required}}:{{tsType}} 
   {{/is.enum}}
+
+
+  {{#is.union}} 
+	@Column('jsonb')
```
metmirr, line 126: Could you use `@CustomField({ db: { type: 'jsonb' }, api: { type: 'json' } })` ?

dzhelezov: Based on what I have researched `api: { type: 'json' }` would make the resulting (scalar) GraphQL type `JSON`, while here we preserve more structure by making it a union type

dzhelezov: So the query _input_ field is JSON, while the result is of Union type and strongly typed

metmirr: Okay! Got it! 👍

metmirr (approved):

PR #975, opened 2020-07-20
Testing: Travis Skip Rust Build

mnaamani:
Just testing do not merge!

This should do a full rust build in the travis CI check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - |
     if [ "$TRAVIS_PULL_REQUEST" != "false" ]
       then
-      if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qE "(.rs|cargo.(lock|toml))$"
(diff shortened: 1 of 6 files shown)


mnaamani:
With Cargo.toml and Cargo.lock files changed the travis script should have run, but it didn't. 
Is the `grep -qE "(.rs|cargo.(lock|toml))$` not accounting for capital `C` in the toml and lock file correctly?
Will run another test with modified rust files.

mnaamani:
I think we can add a `-i` argument to grep to make it case insensitive, I'm not sure if cargo will work with lowercase cargo.toml files

from man pages for grep
```
 -i, --ignore-case
             Perform case insensitive matching.  By default, grep is case sensitive.
```

mnaamani:
The check did run after modifying `.rs` files

mnaamani:
This PR was just for testing purposes.. closing

Issue #902, opened 2020-07-06
Buttons have incorrect animation


bedeho:
The animations in the design provide a guide for how the secondary button should be animated

https://www.dropbox.com/s/mx2wminhpwxx7by/Screen%20Recording%202020-06-04%20at%2011.14.42.mov?dl=0

but our current components don't match this. We should probably try to have similar style animation for all buttons, hopefully generalizing without designer input is feasible?

PR #977, opened 2020-07-20
Atlas hover states

fcescob:
This PR implements the missing hover states mentioned in #902 
Buttons and Channel Preview now have a Hoverable prop, which adds a overlay effect, since the design was not exhaustive, I had to take some initiative on the various states. 

## What To Expect

Buttons and ChannelPreview have an `hoverable` prop (which can be set in storybook) and buttons have an additional `pressed` state where they have been selected. 

## What To Not Expect

`SeriesPreview` is not in the first release of Atlas, so hover effects have not been implemented.

## Feedback

I took some initiative here regarding how things look . Any feedback on how things should look is great feedback!


kdembler:
@Gamaranto needs rebase

atlas/packages/app/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
kdembler, line 1: I don't believe that file will get used any way, all the config should be done via `config-overrides.js`

atlas/packages/app/src/shared/.storybook/preview.jsx
```diff
@@ -8,6 +8,10 @@ import { GlobalStyle } from '../components'
 
 const wrapperStyle = css`
   padding: 10px;
+
+  & > * + * {
```
kdembler, line 12: What's the point of this style? I don't think we should add that for every story

atlas/packages/app/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -16,14 +16,7 @@ function initialsFromName(name: string): string {
...
 
-const Avatar: React.FC<Partial<AvatarProps>> = ({
-  img,
-  outerStyles,
-  onClick = () => {},
-  name,
-  className,
-  ...styleProps
-}) => {
+const Avatar: React.FC<Partial<AvatarProps>> = ({ img, outerStyles, onClick, name, className, ...styleProps }) => {
```
kdembler: I would advise against using `React.FC<Partial<...>>`. This makes props both:

a) harder to read
b) behave more unexpectedly when any prop can be `undefined`.

If there are some optional props that a component accepts, we should just mark them on a one-by-one basis with `?` in type declaration

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { css } from '@emotion/core'
+import styled from '@emotion/styled'
+import { BlockIcon } from '@/shared/icons'
+import { typography, colors, spacing } from '../../theme'
+
+export type ButtonStyleProps = Partial<{
+  variant: 'primary' | 'secondary' | 'tertiary'
+  full: boolean
+  size: 'regular' | 'small' | 'smaller'
+  children: React.ReactNode
```
kdembler: Why would styles need `children` and `onClick` props?

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,188 @@
...
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+  disabled: boolean
+  hoverable: boolean
+  clickable: boolean
+  pressedDown: boolean
```
kdembler: What does `pressedDown` mean?

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,188 @@
+import React from 'react'
+import { css } from '@emotion/core'
+import styled from '@emotion/styled'
+import { BlockIcon } from '@/shared/icons'
+import { typography, colors, spacing } from '../../theme'
+
+export type ButtonStyleProps = Partial<{
```
kdembler: I would advise against using `Partial` here as well and instead define optional props explicitly

atlas/packages/app/src/shared/components/Button/Button.tsx
```diff
@@ -1,31 +1,46 @@
...
+  variant = 'primary',
   disabled = false,
+  hoverable = false,
   containerCss,
-  onClick = () => {},
+  onClick,
   ...styleProps
 }) => {
-  const styles = useCSS({ disabled, type, children, ...styleProps })
+  const [pressedDown, setPressedDown] = useState(false)
```
kdembler: The same question regarding `pressedDown`, I don't understand what's its purpose

atlas/packages/app/src/shared/components/RadioButton/RadioButton.style.ts
```diff
@@ -1,109 +1,100 @@
...
-    top: '7px',
-    left: '7px',
-    '&:focus': {
-      borderColor: disabled ? colors.gray[200] : colors.gray[700],
-    },
-    '&:active': {
-      borderColor: disabled ? colors.gray[200] : colors.gray[700],
-    },
+const colorFromProps = ({ error, checked, disabled }: RadioButtonStyleProps) => {
+  console.log('{error, checked, disabled}', { error, checked, disabled })
```
kdembler: I don't think we need that `log` here

atlas/packages/app/src/shared/components/RadioButton/RadioButton.tsx
```diff
@@ -1,27 +1,54 @@
...
+      <StyledInput checked={checked} error={error} disabled={disabled}>
+        <Input type="radio" onClick={handleClick} disabled={disabled} {...props} checked={checked} />
+      </StyledInput>
+      {label && <span>{label}</span>}
+    </Label>
   )
 }
+
+export default RadioButton
+// export default function RadioButton({
```
kdembler: Commented out code

atlas/packages/app/src/shared/components/Button/Button.tsx
kdembler: Is that what's supposed to set the white border after the button has been clicked? Doesn't `:focus` selector work for that?

kdembler: Also, there's a (I presume) bug when you click 3 button in a row and each of them gets a border that doesn't disappear:

Screenshot 2020-07-30 at 5 56 23 PM

atlas/packages/app/src/shared/.storybook/preview.jsx
kdembler: If that's supposed to style `Button` stories then we should move that style inside those stories.

Also, `* + *` selector will apply to a second element so I think `margin-left` will actually do what's intentend

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,188 @@
...
+  size: 'regular' | 'small' | 'smaller'
+  children: React.ReactNode
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+  disabled: boolean
+  hoverable: boolean
+  clickable: boolean
+  pressedDown: boolean
+}>
+
+const colorsFromProps = ({ variant, hoverable = false, pressedDown }: ButtonStyleProps) => {
```
kdembler: There's a bug that will make secondary button look like a primary one after clicking:


Screen-Recording-2020-07-30-at-5 59 28-PM

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { typography, colors, spacing } from '../../theme'
+
+export type ButtonStyleProps = Partial<{
+  variant: 'primary' | 'secondary' | 'tertiary'
+  full: boolean
+  size: 'regular' | 'small' | 'smaller'
+  children: React.ReactNode
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+  disabled: boolean
+  hoverable: boolean
```
kdembler: What does `hoverable` mean? Is that related to whether the button should have a hover raise animation? If so, I think we should find a more appropriate name, e.g. `animated` or something like that. Every button is hoverable, even when disabled :P

kdembler (changes requested):

atlas/packages/app/src/shared/components/Button/Button.style.tsx
fcescob: Not a bug, that's the pressed state for a secondary button, as seen here: https://www.dropbox.com/s/mx2wminhpwxx7by/Screen%20Recording%202020-06-04%20at%2011.14.42.mov?dl=0

atlas/packages/app/src/shared/components/Button/Button.style.tsx
fcescob: Children are needed when calculating the size, onClick is not actually needed since there is clickable available.

atlas/packages/app/src/shared/components/Button/Button.style.tsx
fcescob: PressedDown stores if a button is currently selected or not, like checked for a checkbox.

atlas/packages/app/src/shared/components/Button/Button.tsx
fcescob: :focus targets elements that are currently on focus, it can't be used to target persistent state like a if a button has been selected. https://developer.mozilla.org/en-US/docs/Web/CSS/:focus

kdembler: Okay so now combined with your other answer I think I see the point. This is to enable sort of functionality of a toggle button, is that correct?

If so, then this functionality shouldn't be enabled for every button and should be instead enabled by props. I'm tempted to say that this should be even a separate component. I don't think a regular button:
a) should have a "pressed down" state
b) should need any state to operate

atlas/packages/app/src/shared/components/Button/Button.style.tsx
kdembler: I see, that looks like a "toggled" state I've talked about before though. I presumed it's a bug because other buttons on the same page don't have this behavior. So here I'd say the same thing about button being "toggleable" as here: https://github.com/Joystream/joystream/pull/977#discussion_r463435760

Per the "Components" screen in Figma, a regular secondary button should just have a different color of the text, not a blue background.

atlas/packages/app/src/shared/components/Button/Button.style.tsx
kdembler: All right, I've missed that. I'm tempted to ask however, do you see a situation in which we would use a button without any content?

kdembler: Genuinely curious, I don't think I've seen a situation like this before

atlas/packages/app/.babelrc
fcescob: Jest won't run without and will give the following error:
```Error: Uncaught [Error: Component selectors can only be used in conjunction with babel-plugin-emotion.]```

We can fix this later once we have the Jest config of the test PR merged

File: atlas/packages/app/src/shared/.storybook/preview.jsx (hunk 2 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/.storybook/preview.jsx:12]: Good catch changed it to `margin-left`, I added it to make elements have margins automatically, if they have siblings

File: atlas/packages/app/src/shared/components/Button/Button.style.tsx (hunk 4 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/Button/Button.style.tsx]: After thinking about your remark I've looked at the code again and found a bug.
`StyledButton` used `childrens` to know how to set the padding (to make Icon only buttons squared), but always caught the `Icon` itself since that was a child of the `StyledButton` component. 
I've added a `hasText` prop which is true if the button has text and false otherwise, this way Icon buttons are now squared.

File: atlas/packages/app/src/shared/components/RadioButton/RadioButton.style.ts (hunk 8 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/RadioButton/RadioButton.style.ts]: Removed, thanks

File: atlas/packages/app/src/shared/components/RadioButton/RadioButton.tsx (hunk 9 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/RadioButton/RadioButton.tsx]: Removed, thanks

File: atlas/packages/app/src/shared/components/Button/Button.style.tsx (hunk 10 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/Button/Button.style.tsx]: Other buttons on that page don't have this behaviour because they don't have an `onClick` prop, so `setPressedDown` is never called. 

Regarding the adherence to the design, if it is toggled, it should look like this: https://www.dropbox.com/s/mx2wminhpwxx7by/Screen%20Recording%202020-06-04%20at%2011.14.42.mov?dl=0

File: atlas/packages/app/src/shared/components/Button/Button.style.tsx (hunk 11 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/Button/Button.style.tsx]: Good point, `animated` is more descriptive, I replaced all mentions of `hoverable`.

File: atlas/packages/app/src/shared/.storybook/preview.jsx (hunk 2 above)
Comment (kdembler) [reply atlas/packages/app/src/shared/.storybook/preview.jsx:12]: But that may not be desired by _every story_ we will write. That may be a majority, I'm not sure, but even then that shouldn't be a global style IMO

File: atlas/packages/app/src/shared/components/Button/Button.tsx (hunk 7 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/Button/Button.tsx]: I agree having it on by default is not ideal. 
One solution is to enable the toggling only for animated buttons, in which case `pressedDown` is not updated if `animated` is false. 
The thing I like about having the `pressedDown` as internal state is that this way animated Buttons work like the regular ones and when we consume them we don't have to keep track of the pressedDown state.
What do you think?

File: atlas/packages/app/src/shared/components/Avatar/Avatar.tsx (hunk 3 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/components/Avatar/Avatar.tsx]: Removed the Partial here. I will also open an issue about removing Partial everywhere, but I would address it in a separate PR.

File: atlas/packages/app/src/shared/.storybook/preview.jsx (hunk 2 above)
Comment (fcescob) [reply atlas/packages/app/src/shared/.storybook/preview.jsx:12]: I think it's safe to assume that most stories we will ever write with more than one components are things like multiple states of the same UI element, like text inputs, buttons, radio buttons etc... For those, I would say the automatic margin is desirable.
If we need multiple elements in the same story without margins, we can wrap them in a `div` to prevent the margin from applying, this to me is a fair compromise since I believe we won't be writing this specific type of stories often.

File: atlas/packages/app/src/shared/components/Button/Button.tsx [hunk 12]
```diff
@@ -1,31 +1,44 @@
-import React from 'react'
+import React, { useState } from 'react'
```
kdembler: `'useState' is defined but never used`

atlas/packages/app/src/shared/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,164 @@
+import React from 'react'
+import { css } from '@emotion/core'
+import styled from '@emotion/styled'
+import { BlockIcon } from '@/shared/icons'
+import { typography, colors, spacing } from '../../theme'
```
kdembler: `'spacing' is defined but never used`

atlas/packages/app/src/shared/components/Button/Button.tsx
```diff
@@ -1,31 +1,44 @@
...
+  icon: boolean
+  disabled: boolean
+  toggle: boolean
   containerCss: SerializedStyles
-  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void
-} & ButtonStyleProps
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+} & Omit<ButtonStyleProps, 'clickable' | 'hasText'>
 
 const Button: React.FC<Partial<ButtonProps>> = ({
```
kdembler, line 13: Since we're refactoring button now, I believe we should get rid of `Partial` here as well

atlas/packages/app/src/shared/components/ChannelPreview/ChannelPreview.style.tsx
```diff
@@ -1,52 +1,63 @@
-import { makeStyles, StyleFn } from '../../utils'
-import { colors } from '../../theme'
+import { css } from '@emotion/core'
+import styled from '@emotion/styled'
+import { colors, spacing } from '../../theme'
+import Avatar from '../Avatar'
 
-export type ChannelPreviewStyleProps = {
-  channelAvatar: string
+export type ChannelPreviewStyleProps = Partial<{
```
kdembler: Let's not use `Partial`

atlas/packages/app/src/shared/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -1,27 +1,39 @@
 import React from 'react'
 import { SerializedStyles } from '@emotion/core'
-import { useCSS } from './ChannelPreview.style'
-import Avatar from '../Avatar'
+
+import { ChannelPreviewStyleProps, OuterContainer, InnerContainer, Info, ChannelAvatar } from './ChannelPreview.style'
```
kdembler: `'ChannelPreviewStyleProps' is defined but never used`

atlas/packages/app/src/shared/components/RadioButton/RadioButton.tsx
kdembler: @Gamaranto this still isn't removed

atlas/packages/app/src/shared/components/Avatar/Avatar.tsx
kdembler: :+1: I would say that we should remove it in the parts that are modified by this PR thoguh

atlas/packages/app/src/shared/components/Button/Button.style.tsx
kdembler: This isn't resolved

kdembler (changes requested):
Some `Partial`s still left and ESLint warnings

atlas/packages/app/src/shared/components/Button/Button.style.tsx
fcescob: thanks will remove it

atlas/packages/app/src/shared/components/Button/Button.tsx
fcescob: Thanks, will remove it

atlas/packages/app/src/shared/components/Button/Button.tsx
fcescob: I would argue that `Partial` is beneficial here, all props are optional when the button is used elsewhere, but `ButtonStyleProps` for example should be strictly typed, so one is forced to define defaults for all style properties

fcescob:
I've modified some type definitions, I did not remove `Partial` everywhere since it was actually quite useful in some places. For example `ButtonStyleProps` is now defined without `Partial`, so the `StyledButton` component inside `Button.tsx` is required to pass all the props down. But `Button` itself uses `Partial` as no props are required when using it elsewhere in the code base. 
I tried to improve the typing in `ChannelPreview` as well, removing `Partial` when possible. In general, I think it's reasonable to use it in cases like this where the type definition is used in multiple places and might be strict somewhere and "relaxed" elsewhere

atlas/packages/app/src/shared/components/Button/Button.tsx
kdembler: We can just as easily make the props optional the explicit way - with `?`. This way the code is more easily readable and we don't need to refactor in the future if we want to add just one required prop

atlas/packages/app/src/shared/components/Button/Button.tsx
```diff
@@ -1,31 +1,45 @@
...
+      css={containerCss}
+      onClick={handleClick}
+      disabled={disabled}
+      variant={variant}
+      clickable={clickable}
+      hasText={hasText}
+      full={full}
+      size={size}
+    >
+      {icon && <StyledIcon />}
```
kdembler, line 43: Thinking about this now I don't understand what's the purpose of that. I'm aware that this is carried from the old code but maybe we should get rid of that? I don't think we use this `block` icon anywhere and even if we did we should probably rather add support for any icon to the button. What do you think?

atlas/packages/app/src/shared/components/ToggleButton/ToggleButton.styles.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import Button from '../Button'
+import type { ButtonStyleProps } from '../Button/Button.style'
+import { spacing, colors } from '../../theme'
+
+export type ToggleButtonStyleProps = {
+  pressedDown: boolean
+} & ButtonStyleProps
+
+const hoverTransition = ({ pressedDown, disabled = false, variant }: ToggleButtonStyleProps) =>
```
kdembler, line 12: There are some UI issues I see here.

Most importantly, Chrome will automatically add an outline to a focused element. In the following gif I'm first toggling the button and then clicking outside the button:
togglebutton

As you can see, even though the button is no longer pressed down, Chrome will add an outline that makes it look very similar to the toggled state. We probably will need to improve that.

Another thing is that once the button is toggled it instantly jumps down from the hovered state (without a transition). This is a rather easy fix, it's a matter of moving the `transition` to the base style and not adding it conditionally.

However, I don't think we necessarily need to fix those right now. I think it'd be better to get this merged and open an issue describing those issues so we can tackle them later. I'll leave the decision to you

atlas/packages/app/src/shared/components/ToggleButton/ToggleButton.styles.tsx
```diff
@@ -0,0 +1,129 @@
...
+        }
+        &:active {
+          color: ${colors.white};
+          background-color: ${colors.gray[100]};
+          border-color: ${colors.gray[100]};
+        }
+      `
+    : null
+
+export const StyledToggleButton: React.FC<ToggleButtonStyleProps> = (props) => (
```
kdembler: Do we need to define a component explicitly here? I'd think `styled(Button)` would work fine. No need to change it though, just curious

kdembler (approved):
In general approving, just some minor things left, but one will require opening a followup issue

atlas/packages/app/src/shared/components/Button/Button.tsx
fcescob: Agree, I can add support for any Icon with no problems since it's something I've done on the search bar branch iirc

atlas/packages/app/src/shared/components/ToggleButton/ToggleButton.styles.tsx
fcescob: I thought so too, but for some reason using `styled(Button)` was not working and no styles were being applied. If you have any idea of why let me know and I'll revert back to that.

kdembler: `Button` needs to accept `className` prop and pass it to its container for styles to be applied

kdembler (approved):
LGTM!

bedeho (approved):

PR #979, opened 2020-07-20
Forum post reaction fix

### LINKED ISSUE #948: Fix forum post reaction
There is no way to have efficient thread deletion if we have reactions, at least as they are currently implemented.

```
/// Each account 's reaction to a post.
pub ReactionByPost get(reaction_by_post) config(): double_map T::PostId, blake2_256(T::ForumUserId) => T::PostReactionId;
```

This is because you would have to iterate over all posts in a thread and delete it from this map.

It's worth keeping in mind why we are even storing this in storage, which is that it should be possible for a user to _unreact_. 
I don't actually even see such an unreaction extrinsic at the moment, so we may not even be gaining the benefit of the extra complexity. So either we will have to sacrifice _unreaction_, and just allow users to submit reactions, and as many as they like, or we have to drop the feature entire. I think the former is more desirable.

ondratra:
Solves subtask of #766:
- rework post reaction storage

Fixes #948.

To minimize merge conflict potential, this PR assumes #929 will be merged first.

### DIFF
--- a/node/src/forum_config/from_serialized.rs
+++ b/node/src/forum_config/from_serialized.rs
@@ -46,7 +46,6 @@ pub fn create() -> ForumConfig {
 
         // TODO: get rid of mocks and setup valid values
         category_by_moderator: Vec::new(),
-        reaction_by_post: Vec::new(),
         poll_items_constraint: new_validation(1, 10),
         data_migration_done: true,
     }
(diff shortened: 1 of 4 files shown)


bedeho (approved):
LGTM.

PR #980, opened 2020-07-20
Forum lead moderator

### LINKED ISSUE #946: Forum lead should be able to act as moderator
Only `delete_category` can directly allow the lead to act as a moderator. I think it makes the most sense to either give the lead full moderator powers directly or none. This halfway state is not a good idea.

ondratra:
Solves subtask of #766:
-  unite lead/moderator privileges

Fixes #946.

To minimize merge conflict potential, this PR assumes #979 will be merged first.


bedeho (approved):
Nice work.

PR #988, opened 2020-07-21
Update @joystream/types to v0.12.0

mnaamani:
Update example code in @joystream/types README

Bump version to v0.12.0 because I published to npm before updating README, and `--force` publish is no longer supported so I couldn't force update the README.

Updated pioneer/package.json to use `link`ed path to @joystream/types this way if yarn is run from inside pioneer folder it still works and you get benefit of changes in types library showing up in IDE immediately


yarn.lock
```diff
@@ -2585,15 +2585,8 @@
...
-  version "0.11.0"
-  dependencies:
-    "@polkadot/keyring" "^1.7.0-beta.5"
-    "@polkadot/types" "^0.96.1"
-    "@types/lodash" "^4.14.157"
-    "@types/vfile" "^4.0.0"
-    ajv "^6.11.0"
-    lodash "^4.17.15"
-    moment "^2.24.0"
+  version "0.0.0"
```
Lezek123: I'm wondering what happened here

Lezek123 (approved):
Couldn't find any issues, but left a comment about some strange change inside `yarn.lock`

yarn.lock
mnaamani: I agree this looks quite odd. I did have to do some merge conflict resolution on yarn.lock but it wasn't at this line number. It was here: https://github.com/Joystream/joystream/pull/988/files#diff-8ee2343978836a779dc9f8d6b794c3b2R1831 but seems related.

Let me get rid of this section, re-run yarn install and see if all goes well:

```
 "@nicaea/types@link:types":
   version "0.0.0"
   uid ""
```

PR #1001, opened 2020-07-21
Network testing: linter and prettier fixes

gleb-urvanov:
Linter and prettier rules applied to the network tests codebase.
Also refactoring from https://github.com/Joystream/joystream/pull/952

PR #1020, opened 2020-07-22
Add github workflows for storage-node and network tests

mnaamani:
Added CI checks for linting and build for network-tests and storage-node.
The network-tests checks are expected to fail because the fixes are applied in a separate PR #1001

### DIFF
--- a/.github/workflows/storage-node.yml
+++ b/.github/workflows/storage-node.yml
@@ -1,0 +1,38 @@
+name: storage-node
+on: [pull_request, push]
+
+jobs:
+  storage_node_build_ubuntu:
+    name: Ubuntu Checks
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: checks
+      run: |
+        yarn install --frozen-lockfile
+        yarn workspace storage-node checks
+
+  storage_node_build_osx:
+    name: MacOS Checks
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: checks
+      run: |
+        yarn install --frozen-lockfile --network-timeout 120000
+        yarn workspace storage-node checks
+        yarn workspace @joystream/storage-cli build
--- a/.github/workflows/storage-node.yml
+++ b/.github/workflows/storage-node.yml
@@ -18,6 +18,7 @@ jobs:
       run: |
         yarn install --frozen-lockfile
         yarn workspace storage-node checks
+        yarn workspace @joystream/storage-cli build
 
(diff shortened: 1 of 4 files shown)


.github/workflows/storage-node.yml
```diff
@@ -0,0 +1,38 @@
...
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: checks
+      run: |
+        yarn install --frozen-lockfile --network-timeout 120000
+        yarn workspace storage-node checks
+        yarn workspace @joystream/storage-cli build
```
Lezek123: This one isn't part of the Ubuntu job, it also doesn't seem necessary, because `yarn workspace storage-node run build` (which does the same thing) is part of root `package.json` `postinstall` script already (so it's beeing run in each of the jobs after `yarn install`).

Lezek123 (changes requested):

.github/workflows/storage-node.yml
mnaamani: good point, will remove

mnaamani: also I thought I'd already pushed a second commit to add it to ubuntu job :)

mnaamani: fixed in 000e2db0157d1bb6e876c37a1c256d81d2cd6305

Lezek123:
It also looks like the `storage-node`'s `lint` script is ignoring `.ts` files.
I suggest changing:
`"lint": "eslint --ignore-path .gitignore .",`
To:
`"lint": "eslint --ext .js,.ts --ignore-path .gitignore .",`
In storage-node's `package.json` `scripts`

mnaamani:
> It also looks like the `storage-node`'s `lint` script is ignoring `.ts` files.
> I suggest changing:
> `"lint": "eslint --ignore-path .gitignore .",`
> To:
> `"lint": "eslint --ext .js,.ts --ignore-path .gitignore .",`
> In storage-node's `package.json` `scripts`

Aren't those extensions checked by default or do you always have to provide them explicitly?
I'm pretty sure those files have been scanned on prior runs, but I will test and confirm.

cc: @shamil-gadelshin

Lezek123:
I'm not sure how effective linter with Typescript-specific rules is for `.js` files, but since not all files are `.ts` in this workspace yet it may make sense to check both and just decrease some rules error level to `warning` (the way it's already done) until everything is converted to TypeScript.

mnaamani:
Yes we had to make quite a few rule tweaking to make the linter happy because that code base is still a hybrid. The plan is to gradually refactor all code to .ts and remove a lot of the rules that were turned off.

Lezek123:
> Aren't those extensions checked by default or do you always have to provide them explicitly?

The documentation says that the default is just `.js` (https://eslint.org/docs/user-guide/command-line-interface).
Perhaps you've been running `yarn workspace @joystream/storage-cli lint` which has them configured (but not `yarn workspace storage-node lint`)

mnaamani:
okay so with current lint script: I get `328 problems (0 errors, 328 warnings)`
and with your suggestion to add the extensions I get: `362 problems (0 errors, 362 warnings)` 
so clearly some files were not covered.
Will update.

mnaamani:
Yeah I think you were right about that.

> Perhaps you've been running `yarn workspace @joystream/storage-cli lint` which has them configured (but not `yarn workspace storage-node lint`)

Lezek123 (approved):
Looks good!

PR #1038, opened 2020-07-23
Use bn.js for numeric types

metmirr:
This PR sets `BN`(bn.js) for the database `numeric` type, it depends on the latest warthog release https://github.com/metmirr/warthog/releases/tag/v2.9.7

### DIFF
--- a/query-node/substrate-query-framework/cli/src/templates/entities/model.ts.mst
+++ b/query-node/substrate-query-framework/cli/src/templates/entities/model.ts.mst
@@ -17,6 +17,8 @@ import {
   StringField
 } from 'warthog';  {{! we don't need extra twists here }}
 
+{{#has.numeric}}import * as BN from 'bn.js' {{/has.numeric}}
+
 {{#has.union}}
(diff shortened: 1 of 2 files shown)


query-node/substrate-query-framework/cli/src/templates/entities/model.ts.mst
```diff
@@ -17,6 +17,8 @@ import {
   StringField
 } from 'warthog';  {{! we don't need extra twists here }}
 
+{{#has.numeric}}import * as BN from 'bn.js' {{/has.numeric}}
```
dzhelezov, line 20: Do we need to add `bn.js` as a dependency to the  `generated/graphql-server/package.json` ?

metmirr: Yes we do, for numeric columns `BN` is the typescript type.

dzhelezov (approved):

bedeho (approved):

PR #1044, opened 2020-07-23
Network testing: ansible to work on mac

gleb-urvanov:
Added support for mac, introduced proper ansible project structure


devops/ansible/roles/run_tests_single_node/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+      docker_container:
+        name: "joystream-node-testing"
+        image: "joystream/node-testing"
+        ports:
+          - "9944:9944"
+        entrypoint: ./node --chain=chainspec.json --alice --validator --ws-external --rpc-cors=all
+        state: started
+
+    - name: execute network tests
+      shell: yarn debug >> ../../.tmp/tests.log
```
mnaamani, line 14:
```suggestion
      shell: yarn test >> ../../.tmp/tests.log
```

mnaamani, line 14:
```suggestion
      shell: yarn test >> ../../.tmp/tests.log
```

mnaamani (approved):
The test for the single node was only running yarn debug instead of yarn test but that can be fixed in next iteration.

the install dependencies playbook successfully installed pip and python, but it messed with my existing version of python2.7 which was installed with brew. It kind of broke docker-compose and my docker desktop stopped responding. I had to mess around with brew and pip after that to get a working setup again. In the end I uninstalled python2 and python3 from brew, uninstalled ansible then reinstalled docker desktop and finally ansible with brew.

I think on a fresh machine these issues wouldn't come up so I think its okay.

I checked on the progress of the tests with `tail -f` so it all looks good.

Nice work

PR #1058, opened 2020-07-24
Prevent the block indexer from hanging after a websocket connection failure

dzhelezov:
This PR fixes the following bug observed on production. 

After the connection to the Substrate web-service (e.g. `wss://kusama-rpc.polkadot.io/`), the block producer stops processing new blocks, even after the connection is restored. 
This PR wraps each API call in `_withTimeout` call which forces the corresponding Promises to reject after 5 seconds. Then the main producer loop retries to produce the current block with exponentially increasing delays between the retries (up to 30 mins).

Note, that such a defensive mechanism would block the producer if there is e.g. a syntactically invalid block, in which case it will try to decode it for an indefinite time. In this case, the node operator is expected to manually resolve the issue and restart the indexer.

### DIFF
--- a/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
+++ b/query-node/substrate-query-framework/cli/src/templates/index-builder-entry.mst
@@ -143,7 +143,7 @@ async function main() {
       process.exit(0);
   }
 
-  node.start({
+  await node.start({
     wsProviderURI: providerUri,
     processingPack: getProcessingPack(),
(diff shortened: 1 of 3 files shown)


metmirr (approved):
Nice work!

I created this PR https://github.com/Joystream/joystream/pull/1065. It enables debug for the functions that fetch data from the chain. So we can have more information about the errors.

bedeho (approved):

PR #1003, opened 2020-07-21
Forum mutability threshold 2

ondratra:
Solves subtask of #766:
- ensure input parameters are not used after // == MUTATION SAFE == comment

Task originated in https://github.com/Joystream/joystream/pull/909#issuecomment-660870592 .

To minimize merge conflict potential, this PR assumes #980 will be merged first.

PR #1059, opened 2020-07-24
Forum efficient removal

ondratra:
Solves subtask of #766
- introduce limits on storage maps to prevent block gas exhaustion

To minimize merge conflict potential, this PR assumes #1003 will be merged first.

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -377,7 +377,10 @@ pub struct Poll<Timestamp, Hash> {
 /// Represents a thread post
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
 #[derive(Encode, Decode, Default, Clone, PartialEq, Eq)]
-pub struct Post<ForumUserId, ThreadId, Hash> {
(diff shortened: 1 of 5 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -577,6 +596,9 @@ decl_storage! {
         /// Category identifier value to be used for the next Category created.
         pub NextCategoryId get(next_category_id) config(): T::CategoryId;
 
+        /// Category identifier value to be used for the next Category created.
+        pub CategoryCounter get(category_counter) config(): T::CategoryId;
```
bedeho, line 611: This comment is incorrect.

bedeho (changes requested):
Looks good.

Can add a comment to each limit to explain which one of them must be small enough to allow safe iteration in a single extrinsic? This will help the runtime developer provide sane values. I think the upper bound on the size of a thread is at least one case.

runtime-modules/forum/src/lib.rs
ondratra: Fixed.

ondratra:
You are right. I added extra commentary to limits.

Issue #1021, opened 2020-07-22
Unique feature reworking


iorveth:
Introduce an  index in storage, something like `double_map (classidxpropertyid, property value) => ()`
whenever a `class` introduces a `property` which has _unique_ requirement, then all `entities` which get `schema` support, or have value updated, will have to update this map, and also respect it when setting new values.

This will only require lookups, which are log time, rather than linear over all entities.

iorveth:
implemented here https://github.com/Joystream/joystream/pull/1064

PR #1064, opened 2020-07-27
New uniqueness feature

iorveth:

1. New uniqueness feature core logic implementation.
1. Existing uniqueness failure cases coverage upgrade.
1. Uniqueness feature failure cases coverage for vector level operations.
1. Inbound entities reference counting logic related bug fix.
1. Uniqueness feature enhancement - store only property value hashes instead of actual values.
#1123

Closes: #1123, https://github.com/Joystream/joystream/issues/1021
To minimize merge conflict potential, this PR assumes #1035 will be merged first.

### DIFF
--- a/runtime-modules/content-directory/src/lib.rs
+++ b/runtime-modules/content-directory/src/lib.rs
@@ -153,6 +153,9 @@ decl_storage! {
         /// Map, representing  CuratorGroupId -> CuratorGroup relation
         pub CuratorGroupById get(curator_group_by_id) config(): map T::CuratorGroupId => CuratorGroup<T>;
 
+        /// Used to enforce uniqueness of a property value across all Entities that have this property in a given Class.
+        pub UniquePropertyValues get(unique_property_values) config(): double_map hasher(blake2_128) (T::ClassId, PropertyId), blake2_128(OutputPropertyValue<T>) => ();
+
         /// Next runtime storage values used to maintain next id value, used on creation of respective curator groups, classes and entities
 
         pub NextClassId get(next_class_id) config(): T::ClassId;
@@ -698,6 +701,8 @@ decl_module! {
 
             let class_properties = class.get_properties();
 
+            let class_id = entity.get_class_id();
+            
             let entity_property_values = entity.get_values();
 
             // Create wrapper structure from provided entity_property_values and their corresponding Class properties
@@ -731,14 +736,32 @@ decl_module! {
 
             // Ensure all provided `new_property_value_references_with_same_owner_flag_set` are valid
             Self::ensure_are_valid_references_with_same_owner_flag_set(
-                &new_values_for_existing_properties, &new_controller
+                new_values_for_existing_properties, &new_controller
             )?;
+
+            let new_output_property_value_references_with_same_owner_flag_set = Self::make_output_property_values(new_property_value_references_with_same_owner_flag_set);
+
+            let new_output_values_for_existing_properties = OutputValuesForExistingProperties::from(&class_properties, &new_output_property_value_references_with_same_owner_flag_set)?;
+
+            // Retrieve OutputPropertyValues, which respective Properties have unique flag set
+            // (skip PropertyIds, which respective property values under this Entity are default and non required)
+            let new_unique_property_values = new_output_values_for_existing_properties.compute_unique();
+
+            // Ensure all provided Properties with unique flag set are unique on Class level
+            Self::ensure_properties_unique_option_satisfied(class_id, &new_unique_property_values)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // Add property values, that should be unique on Class level
+            Self::add_unique_property_values(class_id, new_unique_property_values);
 
             // Make updated entity_property_values from parameters provided
             let entity_property_values_updated =
                     Self::make_updated_property_value_references_with_same_owner_flag_set(
                         unused_property_id_references_with_same_owner_flag_set, &entity_property_values,
-                        &new_property_value_references_with_same_owner_flag_set,
+                        &new_output_property_value_references_with_same_owner_flag_set,
                     );
 
             // Transfer entity ownership
@@ -748,12 +771,8 @@ decl_module! {
                 // Calculate entities reference counter side effects for current operation
                 let entities_inbound_rcs_delta =
                     Self::get_updated_inbound_rcs_delta(
-                        class_properties, entity_property_values, new_property_value_references_with_same_owner_flag_set
+                        class_properties, entity_property_values, new_output_property_value_references_with_same_owner_flag_set
                     )?;
-
-                //
-                // == MUTATION SAFE ==
-                //
 
                 // Update InboundReferenceCounter, based on previously calculated ReferenceCounterSideEffects, for each Entity involved
                 Self::update_entities_rcs(&entities_inbound_rcs_delta);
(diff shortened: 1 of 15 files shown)


runtime-modules/content-directory/src/lib.rs
```diff
@@ -1033,30 +1075,42 @@ decl_module! {
...
+
+            // Ensure all provided Properties with unique flag set are unique on Class level
+            Self::ensure_properties_unique_option_satisfied(class_id, &new_unique_property_values)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // Update property values, that should be unique on Class level
+            Self::add_unique_property_values(class_id, new_unique_property_values);
```
bedeho: I believe that there is cleanup processing missing that removes possible old values from the uniqueness index. The tests should be updated to catch this case, and all instances of such missing logic across the extrinsic should be considered.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1091,19 +1145,24 @@ decl_module! {
...
             let entities_inbound_rcs_delta = Self::make_side_effects_for_clear_property_vector_operation(&property_value_vector, property);
 
-            // Decrease reference counters of involved entities (if some)
-            Self::update_entities_rcs(&entities_inbound_rcs_delta);
-
             // Clear property_value_vector.
             let empty_property_value_vector = Self::clear_property_vector(property_value_vector);
 
+            // Ensure provided `Property` with `unique` flag set is `unique` on `Class` level
+            Self::ensure_property_unique_option_satisfied(class_id, in_class_schema_property_id, &empty_property_value_vector.simplify())?;
```
bedeho: This is basically a check that there is no empty vector for this, possible, unique property. Lets update the comment to say exactly that.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1091,19 +1145,24 @@ decl_module! {
...
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // Decrease reference counters of involved entities (if some)
```
bedeho, line 1217: Seems to be missing the same cleanup of old value from uniquness index.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1158,14 +1217,30 @@ decl_module! {
...
+
             //
             // == MUTATION SAFE ==
             //
 
-            let involved_entity_id = property_value_vector
-                .get_vec_value()
-                .get_involved_entities()
-                .and_then(|involved_entities| involved_entities.get(index_in_property_vector as usize).copied());
+            // Insert updated propery value into entity_property_values mapping at in_class_schema_property_id.
```
bedeho, line 1309: Same cleanup missing here.

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1246,12 +1312,31 @@ decl_module! {
...
+            let class_id = entity.get_class_id();
+
+            // Ensure provided `Property` with `unique` flag set is `unique` on `Class` level
+            Self::ensure_property_unique_option_satisfied(class_id, in_class_schema_property_id, &property_value_vector_updated.simplify())?;
+
             //
             // == MUTATION SAFE ==
             //
 
+            // Insert updated property value into entity_property_values mapping at in_class_schema_property_id.
```
bedeho, line 1414: Same here?

bedeho (changes requested):
- Seems like removal from index was missed a few places, that needs to be added, and tests added that would trigger on failure. Perhaps update tests to fail frist?

- Note that the `network-integration-test` is not for any sort of unit testing, even in ther untime, its explicitly for the network wide integration tests which Gleb & Mokhtar currently are working on, it does not apply to this PR, so I removed the label.

runtime-modules/content-directory/src/lib.rs
iorveth: Only default non required unique  value hashes can be removed on `update_entity_property_values` operation. Implemented here https://github.com/Joystream/joystream/pull/1064/commits/883863befbbdad929dd1bb0db11714ec96c555be#diff-0a165b042aa89a04d896ad36b8a33af5R1027
Additional test coverage for the previously discussed failure case https://github.com/Joystream/joystream/pull/1064/commits/fb19f628aa35f195d4bb94eb8d226129a59a1f5f

runtime-modules/content-directory/src/lib.rs
iorveth: Only default non required unique value hashes can be removed, otherwise we only update its respective hash

runtime-modules/content-directory/src/lib.rs
iorveth: Only default non required unique value hashes can be removed, otherwise we only update its respective hash.

runtime-modules/content-directory/src/lib.rs
iorveth: Only default non required unique value hashes can be removed, otherwise we only update its respective hash.

iorveth:
### Uniqueness feature enhancement - store only property value hashes instead of actual values. 
https://github.com/Joystream/joystream/issues/1123 

### Implementation
1. Core logic - https://github.com/Joystream/joystream/pull/1064/commits/e03886bc5786e8f75218251abe11152eb1ca7190
1. Default non required values potential bug fixed https://github.com/Joystream/joystream/pull/1064/commits/883863befbbdad929dd1bb0db11714ec96c555be.

runtime-modules/content-directory/src/lib.rs
iorveth: Done

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1643,6 +1825,98 @@ impl<T: Trait> Module<T> {
...
+        // Compute old hash from the old unique property value and its respective property id
+        let old_property_value_hash =
+            property_value_vector.compute_unique_hash(in_class_schema_property_id);
+
+        Ok((new_property_value_hash, old_property_value_hash))
+    }
+
+    /// Compute old and new unique property value hashes.
+    /// Ensure new property value hashes with `unique` flag set are `unique` on `Class` level
+    pub fn ensure_property_values_hashes(
```
bedeho: This ensure checkers has a name that does not really indicate a clear purpose.

Closer inspection shows that it seems to do three things
1) check that new property values respect uniquness constraint
2) if so, compute hashes representing these values values
3) if so, determine what hashes must be removed from the uniquness index because they represent values that will no longer be in directory.

It seems that only 1 or 2 are really part of the ensure check, and you could perhaps have one checker with a name like `ensure_new_property_values_respect_uniquness`. It can then return hashes in success case. Then for 3, which is only relevant if you go to mutation, you do the call to `compute_old_unique_hashes` somewhere else. Could be a standalone call, or it could be inside `remove_unique_property_value_hashes` or something else.

bedeho (changes requested):
This basically looks correct now, well done!

Just one comment about a minor refactor possibility for clarity, and then what we discussed about loud failure and asserting no error in extrinsics.

runtime-modules/content-directory/src/lib.rs
iorveth: Agree, refactored it here https://github.com/Joystream/joystream/pull/1064/commits/673fcb38407d012598e14158867d090e6a7846aa
Thank you for the good explanation!

bedeho (approved):

Issue #1002, opened 2020-07-21
Update @joystream/types, Pioneer and the CLI following runtime membership module refactoring


Lezek123:
Related PR: https://github.com/Joystream/joystream/pull/962
Required updates:
- ~~Change all occurences like `api.query.members`, `api.consts.members`, `api.tx.members`: `members` => `membership`~~ _Doesn't seem to be the case, as `api.query.members` still works._
- `api.query.members.membersCreated` => `api.query.members.nextMemberId`
- `api.query.members.memberProfile` => `api.query.members.membershipById`
- `api.query.members.handles` => `api.query.members.memberIdByHandle`
- `api.tx.members.updateProfile` => `api.tx.members.updateMembership`
- `UserInfo` in extrincis - replaced with "plain" parameters (`handle`, `avatar_uri`, `about`):
  - `buyMembership`
  - `updateMembership`
  - `addScreenedMember`
- `@joystream/types`:
  - Remove `Role`, `ActorInRole`, `ActorInRoleSet`, `UserInfo`, `CheckedUserInfo` (+ search for any imports)
  - Rename `Profile` to `Membership` and remove `roles` field (also fix in all imports)
   - Content working group `Lead` - add member id (this opens possibility for some optimalizations in Pioneer and CLI)
- Remove any references to `InitialMembersBalance` (in case there are any)

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1071

PR #1071, opened 2020-07-28
Pioneer, types, CLI: Changes following membership module refactorization

Lezek123:
Related issue: 
https://github.com/Joystream/joystream/issues/1002

Currently `membershipById` map seems to just return an empty `Membership` instance in case member doesn't exist.
Since this instance contains `Bool` field (`Membership.suspended`), we can't check if it's empty via `Membership.isEmpty`, because it will always return `false`. The current workaround to check for "emptiness" of the `Membership` struct is to use ie. `Membership.handle.isEmpty`. Perhaps there will be better ways to handle such cases in the latest version of `@polkadot/api`.


Lezek123:
Bumped the version of `@joystream/types` and adjusted the dependencies in order to fix `network-tests` CI checks.

mnaamani:
I guess the check `Membership.handle.isEmpty` will do for now and waiting to see what you find in latest version of polkadotjs api, but if the result is similar, ie. is there is no helper to check if a key exists in a runtime map, we should consider the general solution for all maps to get the raw value from storage, because it seems sub-optimal and even fragile each time to try to find which field to do an isEmpty check on.

```
const key = api.query.members.membershipById.key(memberId)
const maybeValue = (await api.rpc.state.getStorage(key)) // Option<ValueType>
const value = maybeValue.isSome ? new ValueType(maybeValue.unwrap()) : undefined
```

Tested working as expected

mnaamani (approved):

mnaamani:
The npm pack test for the cli still seems to be failing:

```
 npm ERR! code ETARGET
npm ERR! notarget No matching version found for @joystream/types@^0.13.0.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
npm ERR! notarget 
npm ERR! notarget It was specified as a dependency of 'package'
npm ERR! notarget 
```

Lezek123:
> The npm pack test for the cli still seems to be failing

I was trying to make all the CI chceks work, but this one now fails because the types are not yet published (I bumped the version for `Alexandria`).
This was a useful check before publishing the CLI, but now I consider removing it or making it "optional" if that's possible.
Or perhaps we should just perform it on `master` and make sure the types are published before there's a PR to master?

mnaamani:
> Or perhaps we should just perform it on `master` and make sure the types are published before there's a PR to master?

I think that would be good, but it would be good to find a way to test it before publishing. Is this something lerna can do, or perhaps a similar command in yarn, like yarn pack which would be able to use types from the workspace?

mnaamani:
I tried locally `yarn pack` and `yarn link` and it seems to work. 
So maybe for master branch we can do the check with npm, and for PRs and other branches we use yarn?

Lezek123:
The CI check fails after `yarn link` even though it says that the package is correctly linked (cannot find command `joystream-cli`): https://github.com/Lezek123/substrate-runtime-joystream/runs/952497411?check_suite_focus=true

I had the same issue locally when using the default configuration (without `.npmrc`)

mnaamani:
It seems to work on the mac but fails on ubuntu. Maybe we skip the `joystream-cli help` step for now. The key check is that it builds and all dependencies from npm are found.

Lezek123:
Adjusted the CI checks without `joystream-cli help` for now

mnaamani:
Does the failing of network tests checks have anything to do with merging of https://github.com/Joystream/joystream/pull/1091

Specifically this: https://github.com/Joystream/joystream/pull/1091/commits/87c14f62017e65f99e3ffd9e66f2b132c250586b

cc: @gleb-urvanov

Lezek123:
I think it has to do with the version of `@nicaea/types` inside `network-testing` `package.json` (that I set to `^0.12.0` in this PR) and also this: https://github.com/Joystream/joystream/pull/1085

I'll try to resolve.

PR #1073, opened 2020-07-28
Qnode new format for event name

metmirr:
This PR changes event handler naming convention from `handle<EventName>` to `<runtimeModule><EventName>`. There are same event names across runtime modules so we need to make sure each event handler name is unique. For example, in Kusama network, we can see that the `treasury` and `balances` module both have an event named `Deposit`.


dzhelezov:
It feels safer to me if we still leave a separator between the `module` and `method`, like an underscore. The method then would look like `treasury_TipProposed` and so is the event name. It might be useful to be able to decode the module and the method by simply looking at the event name, and if there is no delimiter it becomes ambiguous.

metmirr:
> It feels safer to me if we still leave a separator between the `module` and `method`, like an underscore. The method then would look like `treasury_TipProposed` and so is the event name. It might be useful to be able to decode the module and the method by simply looking at the event name, and if there is no delimiter it becomes ambiguous.

Done!

dzhelezov (approved):

bedeho (approved):

PR #1070, opened 2020-07-28
StorageNode - Log submitted txs

mnaamani:

PR #1074, opened 2020-07-29
Storage node/balance and synched checks

mnaamani:
Additional checks for chain state sync status and minimum free balance before proceeding with sending transactions.
Logs some warnings also on low balance.

Mainly to avoid situations as described in https://github.com/Joystream/joystream/pull/1070#pullrequestreview-456625480

During syncing, an improvement would be to do the balance check before each transaction submission rather than a catch all before the sync run. But this will have to do for now.

### DIFF
--- a/storage-node/packages/runtime-api/index.js
+++ b/storage-node/packages/runtime-api/index.js
@@ -31,6 +31,7 @@ const { DiscoveryApi } = require('@joystream/storage-runtime-api/discovery')
 const { SystemApi } = require('@joystream/storage-runtime-api/system')
 const AsyncLock = require('async-lock')
 const Promise = require('bluebird')
+const { sleep } = require('@joystream/storage-utils/sleep')
 
 Promise.config({
   cancellation: true,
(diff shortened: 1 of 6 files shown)


storage-node/packages/runtime-api/index.js
```diff
@@ -31,6 +31,7 @@ const { DiscoveryApi } = require('@joystream/storage-runtime-api/discovery')
 const { SystemApi } = require('@joystream/storage-runtime-api/system')
 const AsyncLock = require('async-lock')
 const Promise = require('bluebird')
+const { sleep } = require('@joystream/storage-utils/sleep')
```
mnaamani, line 34: the storage-utils packages doesn't have an index.js .. so importing like this, but we should fix in future.

Lezek123 (approved):
Didn't notice any issues

mnaamani:
Some observations

 - When storage-node starts before joystream-node, there is a short period when the joystream-node is spinning up, it is not yet actively syncing so the check at startup wouldn't work as expected.

- It might be also good to prevent storage-node sending transactions if the joystream-node doesn't have any peers connected as it indicates potential network issues so postponing tx submission might also make sense instead of dealing with tx's filling up the queue and timing out etc.

PR #1075, opened 2020-07-29
Migrate Runtime to the Substrate version 2.0.0 rc4

shamil-gadelshin:
Converted the runtime crate and all pallets.


runtime-modules/content-working-group/src/mock.rs
```diff
@@ -241,3 +234,15 @@ pub type System = system::Module<Test>;
 pub type Balances = balances::Module<Test>;
 pub type ContentWorkingGroup = Module<Test>;
 pub type Minting = minting::Module<Test>;
+
+// Recommendation from Parity on testing on_finalize
+// https://substrate.dev/docs/en/next/development/module/tests
+pub fn run_to_block(n: u64) {
+    while System::block_number() < n {
+        <System as OnFinalize<u64>>::on_finalize(System::block_number());
+        <ContentWorkingGroup as OnFinalize<u64>>::on_finalize(System::block_number());
```
mnaamani, line 243: should we also be doing this for other modules that are part of the this test runtime, like Balances and Minting?

runtime-modules/governance/src/lib.rs
```diff
@@ -9,3 +9,7 @@ mod sealed_vote;
 mod stake;
 
 mod mock;
+
+//TODO: Convert errors to the Substrate decl_error! macro.
+/// Result with string error message. This exists for backward compatibility purpose.
+pub type DispatchResult = Result<(), &'static str>;
```
mnaamani, line 15: This type alias is defined in each pallet lib.rs
Is there a particular reason you didn't just put it once in the common crate ?

runtime-modules/hiring/src/test/public_api/add_application.rs
```diff
@@ -115,7 +119,7 @@ impl AddApplicationFixture {
             let expected_application = Application {
                 opening_id: self.opening_id,
                 application_index_in_opening: expected_application_index_in_opening,
-                add_to_opening_in_block: 1,
+                add_to_opening_in_block: 0,
```
mnaamani, line 122: I couldn't quite figure out why this change was required.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -253,155 +253,125 @@ decl_error! {
...
-    fn from(error: proposal_engine::Error) -> Self {
-        match error {
-            proposal_engine::Error::Other(msg) => Error::Other(msg),
-            proposal_engine::Error::RequireRootOrigin => Error::RequireRootOrigin,
-            _ => Error::Other(error.into()),
-        }
-    }
-}
-
-impl From<proposal_discussion::Error> for Error {
```
mnaamani, line 276: As I understand these conversions are no longer required, because we are now returning the new dispatch result type that supports the error types defined in decl_error macro?

runtime-modules/proposals/codex/src/tests/mock.rs
```diff
@@ -209,19 +198,61 @@ parameter_types! {
...
+where
+    Call: From<LocalCall>,
+{
+    type OverarchingCall = Call;
+    type Extrinsic = Extrinsic;
 }
 
+pub type Extrinsic = TestXt<Call, ()>;
+
+/*
```
mnaamani, line 233: Can this commented code be removed?

runtime-modules/proposals/codex/src/tests/mock.rs
```diff
@@ -209,19 +198,61 @@ parameter_types! {
...
+    type NextNewSession = ();
+    type ElectionLookahead = ();
+    type Call = Call;
+    type MaxIterations = ();
+    type MinSolutionScoreBump = ();
+    type MaxNominatorRewardedPerValidator = ();
+    type UnsignedPriority = ();
+}
+
+impl<LocalCall> system::offchain::SendTransactionTypes<LocalCall> for Test
```
mnaamani, line 223: I can't spot what changes in the mocked runtime required implementing this trait.

runtime-modules/proposals/discussion/src/tests/mod.rs
```diff
@@ -34,12 +35,12 @@ struct TestThreadEntry {
...
 
 fn assert_thread_content(thread_entry: TestThreadEntry, post_entries: Vec<TestPostEntry>) {
-    assert!(<ThreadById<Test>>::exists(thread_entry.thread_id));
+    assert!(<ThreadById<Test>>::contains_key(thread_entry.thread_id));
 
     let actual_thread = <ThreadById<Test>>::get(thread_entry.thread_id);
     let expected_thread = DiscussionThread {
         title: thread_entry.title,
-        created_at: 1,
+        created_at: 0,
```
mnaamani, line 43: could you explain why 0 now and not 1. This is maybe similar to a prior comment.

runtime-modules/governance/src/lib.rs
mnaamani: I think I understand now why you are keeping it local.

runtime-modules/stake/src/lib.rs
```diff
@@ -268,7 +266,7 @@ where
     /// For all slahes that should be executed, will apply the Slash to the staked amount, and drop it from the ongoing slashes map.
     /// Returns a vector of the executed slashes outcome: (SlashId, Slashed Amount, Remaining Staked Amount)
     fn finalize_slashes(&mut self, minimum_balance: Balance) -> Vec<(SlashId, Balance, Balance)> {
-        let mut finalized_slashes: Vec<(SlashId, Balance, Balance)> = vec![];
+        let mut finalized_slashes: Vec<(SlashId, Balance, Balance)> = Vec::new();
```
mnaamani, line 269: I see the vec![] form is still used elsewhere so I gather its still valid?

runtime/Cargo.toml
```diff
@@ -4,372 +4,151 @@ edition = '2018'
...
+    'sp-runtime/std',
+    'sp-arithmetic/std',
+    'sp-offchain/std',
+    'sp-authority-discovery/std',
+    'sp-consensus-babe/std',
+    'sp-transaction-pool/std',
+    'sp-block-builder/std',
+    'sp-session/std',
+    'sp-staking/std',
+    'node-primitives/std',
```
mnaamani, line 104: This comes from the packages in substrate/bin/node crate.
Although we will use the same primitives, this should come from a similar package in our own node crate.

runtime/src/lib.rs
```diff
@@ -210,77 +192,153 @@ pub fn native_version() -> NativeVersion {
...
+{
+    type Extrinsic = UncheckedExtrinsic;
+    type OverarchingCall = Call;
+}
+
+parameter_types! {
+    pub const IndexDeposit: Balance = 0; // no minimum deposit
+}
+
+impl pallet_indices::Trait for Runtime {
```
mnaamani, line 330: I've always chosen not to include the account indices module because of its behavior when the minimum existential deposit was configured to be zero. Basically it would allow re-use of an index for new accounts which can be undesirable and lead to some confusion on the part of users.

runtime/src/tests/storage_integration.rs
```diff
@@ -2,12 +2,14 @@ use super::initial_test_ext;
...
 use crate::Runtime;
 
-use srml_support::{StorageLinkedMap, StorageMap};
+use frame_support::StorageMap;
 use working_group::{Instance2, Worker};
 
 #[test]
 fn storage_provider_helper_succeeds() {
     initial_test_ext().execute_with(|| {
+		// Bug in random module requires move the initial block number.
```
mnaamani, line 11: does the random module seed itself with some initial state from genesis or simply the block number? I imagine that is part of how it can be deterministic?

runtime/src/migration.rs
```diff
@@ -47,8 +37,8 @@ decl_storage! {
         /// initialized to Some(VERSION.spec_version). It is an Option because the first time the module
         /// was introduced was as a runtime upgrade and type was never changed.
         /// When the runtime is upgraded the spec version be updated.
-        pub SpecVersion get(spec_version) build(|_config: &GenesisConfig| {
-            VERSION.spec_version
+        pub SpecVersion get(fn spec_version) build(|_config: &GenesisConfig| {
```
mnaamani, line 40: There is a fundamental problem with this migration module, specifically to do with what the initial value set here will be. This has actually only caused a problem with unexpected behavior when working on our staging networks when. I won't go into the details, but we should as previously discussed move away from it to use the new builtin substrate on runtime upgraded callback.

runtime/src/lib.rs
```diff
@@ -290,48 +348,44 @@ parameter_types! {
...
     type ExistentialDeposit = ExistentialDeposit;
-    type TransferFee = TransferFee;
-    type CreationFee = CreationFee;
+    type AccountStore = System;
+}
+
+parameter_types! {
+    pub const TransactionByteFee: Balance = 1; // TODO: adjust fee
+    pub const TargetBlockFullness: Perquintill = Perquintill::from_percent(25);
+    pub AdjustmentVariable: Multiplier = Multiplier::saturating_from_rational(1, 100_000);
```
mnaamani, line 365: I'm not sure if these values match for our expected total issuance or if it even matters.

mnaamani (approved):
So tremendous job doing this conversion.
I left a few questions really. No major change requests everything seems to be in order.
Build and tests all going well.
Out of curiosity I did try to do a cargo update and see if a build with latest dependencies works, but I did run into the libp2p build error.
substrate rc5 is out and it does bump version of libp2p, perhaps it will help.

shamil-gadelshin:
Part of this [PR](https://github.com/Joystream/joystream/pull/1126)

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: Correct. The error model changed significantly.

runtime-modules/hiring/src/test/public_api/add_application.rs
shamil-gadelshin: The default block number just changed.

runtime-modules/proposals/codex/src/tests/mock.rs
shamil-gadelshin: Sure (conversion artifact).

runtime-modules/proposals/discussion/src/tests/mod.rs
shamil-gadelshin: The default block number just changed.

runtime-modules/stake/src/lib.rs
shamil-gadelshin: I encountered some weird build errors when using this kind of empty vec. Even after the latest rust compiler guaranteed that vec![] == Vec::new(). Probably it happens when you don't use 'std' feature. Anyway, now it's safe to use both types of empty vec initialization.

runtime/Cargo.toml
shamil-gadelshin: Yes, I already removed that during the node conversion.

runtime-modules/proposals/codex/src/tests/mock.rs
shamil-gadelshin: It's most likely just don't compile without it.

runtime/src/migration.rs
shamil-gadelshin: Ok. I will remove the migration module out of the runtime.

runtime/src/lib.rs
shamil-gadelshin: I've removed them in the latest PR. We will have to pick correct values when we choose correct fees and transactions.

runtime-modules/stake/src/lib.rs
mnaamani: ok great. Yes I think the difference has to do with whether you are building native or WASM runtime.

runtime/src/lib.rs
mnaamani: Yes as discussed in our call we will revisit all these runtime parameters :+1:

mnaamani:
Thanks for taking time to answer my questions.
Merging despite travis being unhappy, but that is okay because the node hasn't been updated in this PR.

Issue #1005, opened 2020-07-21
Network testing: fixture object encapsulation


gleb-urvanov:
Refactor tests to have fixtures encapsulated into objects

gleb-urvanov:
PR https://github.com/Joystream/joystream/pull/1080

PR #1080, opened 2020-07-29
Network testing: object encapsulation

gleb-urvanov:
PR for https://github.com/Joystream/joystream/issues/1005


tests/network-tests/src/iznik/tests/fixtures/membershipModule.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // Buying memebership
+    await this.apiWrapper.buyMembership(
+      this.aKeyPair,
+      this.paidTerms,
+      `late_member_${this.aKeyPair.address.substring(0, 8)}`,
+      true
+    )
+
+    // Assertions
```
mnaamani, line 98: For these assertions to really be useful, I think we should have also checked the same at the start of the runner.

mnaamani:
Nice work with this refactoring.
Some general points about the Fixtures.
Sometimes we pass the sudo keypair which is only used as a source of funds. In other cases we need to actually execute a sudo transaction. It might be better for clarity to give a better name to the arguments to reflect the actual use case.
The arguments m1Keys and m2Keys is loosing its meaning inside the fixture, maybe it would be better to give them better names like `councilMemberKeys`, `otherMemberKeys`. Sometimes you don't even need to send all keypairs depending on the fixture. Try to send as few arguments as possible really.
The `getResult()` method on the fixtures is very general, it might be better to give a name like `getCreatedProposalId()` for example for fixtures that create a proposal, which gives more clarity in the code of the Tests.

I tried running the test: `yarn debug` but got:

```
MokoBook:network-tests mokhtar$ yarn debug
yarn run v1.22.4
$ tap --files src/iznik/tests/workingGroup/*Test.ts -T
Error: EACCES: permission denied, unlink '/Users/mokhtar/joystream/joystream/tests/network-tests/.nyc_output/29dc554b-bc22-4202-a783-3b9faafd7bb8.json'
    at Object.unlinkSync (fs.js:1136:3)
```

Could you merge iznik again into your branch again.

mnaamani:
I had to delete the `.nyc_output` with sudo permissions because it was leftover from last run using ansible 

But now I get error running tests:

```
yarn run v1.22.4
$ tap --files src/iznik/tests/workingGroup/*Test.ts -T
src/iznik/tests/workingGroup/manageWorkerAsLeadTest.ts 2> Unknown types found, no types for Membership
 FAIL  src/iznik/tests/workingGroup/manageWorkerAsLeadTest.ts
 ✖ terms.unwrap is not a function

import { ApiPromise, WsProvider } from '@polkadot/api'
import { Option, Vec, Bytes, u32 } from '@polkadot/types'
import { Codec } from '@polkadot/types/types'

  test: Creating a set of members
  stack: >
    src/iznik/utils/apiWrapper.ts:1:55554

    BuyMembershipHappyCaseFixture.runner (src/iznik/tests/fixtures/membershipModule.ts:2:43)
  at:
    line: 1
    column: 55554
    file: src/iznik/utils/apiWrapper.ts
  type: TypeError
  tapCaught: returnedPromiseRejection
```

mnaamani:
> I had to delete the `.nyc_output` with sudo permissions because it was leftover from last run using ansible
> 
> But now I get error running tests:
> 
> ```
> yarn run v1.22.4
> $ tap --files src/iznik/tests/workingGroup/*Test.ts -T
> src/iznik/tests/workingGroup/manageWorkerAsLeadTest.ts 2> Unknown types found, no types for Membership
>  FAIL  src/iznik/tests/workingGroup/manageWorkerAsLeadTest.ts
>  ✖ terms.unwrap is not a function
> 
> import { ApiPromise, WsProvider } from '@polkadot/api'
> import { Option, Vec, Bytes, u32 } from '@polkadot/types'
> import { Codec } from '@polkadot/types/types'
> 
>   test: Creating a set of members
>   stack: >
>     src/iznik/utils/apiWrapper.ts:1:55554
> 
>     BuyMembershipHappyCaseFixture.runner (src/iznik/tests/fixtures/membershipModule.ts:2:43)
>   at:
>     line: 1
>     column: 55554
>     file: src/iznik/utils/apiWrapper.ts
>   type: TypeError
>   tapCaught: returnedPromiseRejection
> ```

My bad, was running latest runtime with modified runtime membership module..

mnaamani:
The work in this PR is included in #1091. If so we can close it. @gleb-urvanov

Issue #824, opened 2020-06-26
Atlas Product Release


bedeho:
# Purpose

The purpose of this document is to provide preliminary guidance for the front-end developers working on implementing the first release of Atlas based on the designs and design system currently provided here.

https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=730%3A5283

The reason extra guidance is provided is because the designs

- do not actually constitute a coherent product design, but in many places just presents installations of a modular design system.
- include functionality which should properly be out of scope for a first release.

# Goal

The goal of the release is to ship a minimal version of a consumer experience of the type and quality level that people can expect for the mainnet launch.

# Scope

- **No publishing GUI:** In fact, not only will there be no publishing experience as part of Atlas, in part because there are no designs for it, but we will not even update the publishing and channel creation experience that is already part of Pioneer. That will be dropped. Instead publishing and curation will only be possible through the Joystream CLI to begin with, as that requires substantially less effort. This will allow us to go to market with the viewing experience sooner, and also buy us more time to resolve what the publishing experience should look like, and whether it should even be part of Atlas.

- **No user consumption accounts:** The current design features signup and login screens, and a login button.

- **No personalisation:** By personalisation, we mean application behaviour which is informed by your past actions. While in most services, this requires an account, it can also be implemented with temporary client side storage. We already do this in Pioneer. This includes things such as
  - viewing history, including continuing to watch interrupted content
  - what channels you follow
  - what categories you follow
  - watchlist
  - onboarding state

- **Only video content:** Series, episodes, live content and playlists are not supported.

- **No personae metadata:** In the design there is list of personaes associated with each video, and there is also a standalone personae page that serves as a landing page for someone or something, by providing a bio and binding together all content they appear in. All of this will not be here.

- **No follower counts:** Channels do not have follow counts, as we do not have any following functionality.

- **Naive view counts:** We will have view counts, they will be hosted and managed by trusted view count server operated by Jsgenesis, works in a naive way by tallying view events from clients.

- **Simple playback:** We just use a simple native player, so no extra functionality for altering resolution, casting, scrubbing previews, subtitle language selection etc.

- **DMCA:** A simple link in the footer should. It would link to a separate page entirely which has the required DMCA information.

**Accommodating these requirements may involve changing or removing part of the design in non-obvious ways. As you run into these, please get in touch with product manager (Bedeho) for guidance.**

# Seed Content

[...]

PR #1090, opened 2020-07-30
cleanup mock data

### LINKED ISSUE #922: Clean up mock data and refactor components to use proper data types [Atlas]
Right now we use naive mock data (strings for view counts, formatted string with dates, etc.). Also, our mock data is rather repetitive and lots of images are broken.

We should improve the mock data by adding some different scenarios and make the components work with proper types/objects.

kdembler:
This PR cleans up mocked data used by Atlas by:

* adding set of images to use
* adding varied data for both videos and channels so we can observe how the app behaves with different data length, etc.
* cleaning up the home view to correspond to match what we agreed upon in https://github.com/Joystream/joystream/issues/824#issuecomment-653048676

Resolves #922


atlas/packages/app/src/utils/number.ts
```diff
@@ -1,3 +1,18 @@
 export const formatNumber = (num: number): string => {
   return num.toLocaleString('en-US').split(',').join(' ')
 }
+
+export const formatNumberShort = (num: number): string => {
```
fcescob, line 5: There is a tiny inconsistency here, this returns `3.0K` for `3001` and `3K` for `3000`, same things happens with milions

fcescob (commented):
Beside one minor thing, looks good to me. 

Atlas looks stunning with images!
immagine

atlas/packages/app/src/utils/number.ts
kdembler: Good catch! Changed the function to always skip `.0` suffix

atlas/packages/app/src/utils/number.ts
```diff
@@ -14,5 +14,10 @@ export const formatNumberShort = (num: number): string => {
     suffix = 'K'
   }
 
-  return `${value % 1 === 0 ? value : value.toFixed(1)}${suffix}`
+  let formattedValue = value.toFixed(1)
+  if (formattedValue.endsWith('.0')) {
+    formattedValue = formattedValue.slice(0, formattedValue.length - 2)
```
fcescob, line 19: Formatting is still a bit inconsistent,  `1000` and `1001` will get two different representations, `1000` and `1K` respectively, same happens for `1000000` and `1000001`


```javascript
// This evaluates to "1000" 
formatNumber(1000)
// This evaluates to "1K"
formatNumber(1001)
```

kdembler: ofc, that should have been `>=` instead of `>`. Fixed now!

fcescob (approved):
Good job!

PR #1092, opened 2020-07-30
Reading old chain state + example usage

Lezek123:
```
yarn workspace @joystream/js-examples build
yarn workspace @joystream/js-examples run-example readOldState
```
We can also provide endpoint and block number (which was before _Contantinopole => Nicaea_ upgrade in that case)
```
yarn workspace @joystream/js-examples run-example readOldState wss://rome-rpc-endpoint.joystream.org:9944/ 1500000
yarn workspace @joystream/js-examples run-example readOldState ws://localhost:3000/ 50
```

Feedback about more interesting cases of querying old chain data that could be tested with this script will be much appreciated.

### DIFF
--- a/joystream-js/.gitignore
+++ b/joystream-js/.gitignore
@@ -1,0 +1,3 @@
+lib
+examples/**/*.js
+examples/**/*.d.ts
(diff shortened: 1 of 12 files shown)


Lezek123:
This is no longer relevant since the much newer version of `@polkadot/api` we're using now exposes functionality like `api.at` (https://github.com/polkadot-js/api/issues/3871) for this purpose.

PR #1107, opened 2020-08-03
Search bar

fcescob:
This PR adds a SearchBar component and a Navbar to the App, it also implements a "search" screen with some random results. 

## Current Issues

There is an issue with mocking, so while search results should be relevant to the string searched, only one video repeated multiple times appears.

This PR builds off #1248 so any comments concerning Icons should be directed there.


atlas/packages/app/src/components/LayoutWithRouting.tsx
```diff
@@ -1,15 +1,20 @@
...
-import { HomeView, VideoView } from '@/views'
+import { Navbar } from '@/components'
+import { HomeView, VideoView, SearchView } from '@/views'
 import routes from '@/config/routes'
 import { Router } from '@reach/router'
 
 const LayoutWithRouting: React.FC = () => (
   <main>
     <GlobalStyle />
+    <Router primary>
```
kdembler, line 11: Why do we need this `<Router>`?

atlas/packages/app/src/components/Navbar.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { useState } from 'react'
+import styled from '@emotion/styled'
+import { animated, useSpring } from 'react-spring'
+import { ReactComponent as UnstyledLogo } from '../assets/logo.svg'
+import { Searchbar, Icon } from '@/shared/components'
+import { colors } from '@/shared/theme'
+import { navigate, Link, RouteComponentProps } from '@reach/router'
+
+const Header = styled.header<{ isSearching: boolean }>`
```
kdembler: There's a lot of `styled` components here, let's move them to a separate file to keep the code readable

atlas/packages/app/src/components/Navbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+  align-items: center;
+  justify-self: end;
+  :hover {
+    cursor: pointer;
+  }
+`
+
+const Navbar: React.FC<RouteComponentProps> = ({ location }) => {
+  const [search, setSearch] = useState('')
+  const [isSearching, setIsSearching] = useState(Boolean(search.trim()))
```
kdembler: Why `useState(Boolean(search.trim()))`? This will always be `false` - this is only the default value

atlas/packages/app/src/components/Navbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const [props, set] = useSpring(() => ({
```
kdembler: Please use more descriptive variable names

kdembler: Also, there is a warning that `props` is never used, what's the purpose of this then?

atlas/packages/app/src/components/Navbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+          </Link>
+          <Link to="/discover">
+            <StyledIcon icon="binocular" />
+          </Link>
+        </NavigationContainer>
+      )}
+
+      <StyledSearchbar
+        placeholder="Search..."
+        onChange={(e) => {
```
kdembler: It'd be good to move all of this handlers to functions outside of `return` to make the JSX more readable

atlas/packages/app/src/components/VideoBestMatch.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+type BestVideoMatchProps = {
+  video: VideoFields
+  onClick: (e: React.MouseEvent<HTMLImageElement>) => void
+}
+const Container = styled.div`
+  color: ${colors.gray[300]};
+  padding-right: 2rem;
+`
+const Header = styled.h3``
```
kdembler: We don't need that

atlas/packages/app/src/components/VideoBestMatch.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Button } from '@/shared/components'
+import { formatNumber } from '@/utils/number'
+import { formatDate } from '@/utils/time'
+import { colors } from '@/shared/theme'
+
+type BestVideoMatchProps = {
+  video: VideoFields
+  onClick: (e: React.MouseEvent<HTMLImageElement>) => void
+}
+const Container = styled.div`
```
kdembler: Like in the `Navbar`, let's move those to a separate file

atlas/packages/app/src/components/VideoBestMatch.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <InnerContainer>
+        <TitleContainer>
+          <Title>{title}</Title>
+          <span>
+            {formatNumber(views)} views • {formatDate(publishedOnJoystreamAt)}
+          </span>
+        </TitleContainer>
+        <ButtonsGroup>
+          <Button variant="secondary">Share</Button>
+          <Button icon="block" />
```
kdembler: Do we need those? I know they're part of the design but I don't think we'll be using them for now? Same for the share button

atlas/packages/app/src/components/VideoGallery.tsx
```diff
@@ -41,7 +41,9 @@ const VideoGallery: React.FC<VideoGalleryProps> = ({ title, action, videos, load
   }, [])
 
   const handleVideoClick = (id: string) => {
-    navigate(routes.video(id))
+    navigate(routes.video(id), {
+      replace: true,
```
kdembler: Why this change? If a user goes to video view from the landing page, we want them to be able to go back, right?

atlas/packages/app/src/mocking/server.ts
```diff
@@ -33,6 +34,38 @@ createServer({
...
             return shuffle(channels)
           },
+          // FIXME: This resolver is currently broken and returns the same result n times instead of the correct result.
+          search: (obj: unknown, { query_string }: SearchVariables, context: unknown, info: unknown) => {
+            const items = [...mockVideos, ...mockChannels]
+
+            let rankCount = 0
+            const matchQueryStr = (str: string) => str.includes(query_string) || query_string.includes(str)
+
+            const relevantItems = items.flatMap((item: any) => {
```
kdembler: This is overly complicated and hard to read with all the nesting. Also, when you need to build upon an array on some conditions it may make more sense to use `reduce` instead of `flatMap` and returning `[]`.

Consider the following example:
```tsx
const relevantItems = items.reduce((acc, item) => {
  const matched = item.__typename === 'Channel' ? matchQueryStr(item.handle) : matchQueryStr(item.description) || matchQueryStr(item.title)
  if (!matched) return
  return [
    ...acc,
    {
      __typename: 'FreeTextSearchResult',
      item,
      rank: rankCount++,
    }
  ]
}, [])
```

atlas/packages/app/src/views/SearchView.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { data, loading } = useQuery<Search, SearchVariables>(SEARCH, { variables: { query_string: search } })
+
+  if (loading || !data) {
+    return <p>Loading...</p>
+  }
+  if (!data.search) {
+    return <p>Something went wrong...</p>
+  }
+
+  const results = data.search.slice().sort((a, b) => a.rank - b.rank)
```
kdembler: This and the following 2 lines are potentially expensive and we shouldn't do them on every render. Consider using memoization here

atlas/packages/app/src/components/LayoutWithRouting.tsx
kdembler: Is this because of navigation being done inside the navbar?

kdembler (changes requested):
Apart from the comments, some things that will need improvement:

### UI
- We should indicate no search results, currently if there is nothing found, we're left with an empty screen with just the tabs
- The navbar changes its height when focusing/closing the searchbar, I don't think that's desired
- Should we present videos search result in a carousel? We have a lot of screen space, that seems like a weird choice. Especially in the videos tab when there is nothing else to browse. @bedeho 
- Navigation icons could use some more contrast - it's hard to spot them when they're this gray

### Code
- Please mind the warnings outputted by ESLint, there's a number of unused variables/imports
- Sometimes it's fine to have styled components alongside the main component but if there's a lot of them, we should keep them separate 
- You're repeatedly using raw pixel values. Shouldn't we use spacing values from the theme instead?

atlas/packages/app/src/mocking/server.ts
fcescob: This looks cleaner indeed, changed it.

atlas/packages/app/src/components/VideoGallery.tsx
fcescob: I forgot to remove that, removed it now.

atlas/packages/app/src/components/VideoBestMatch.tsx
fcescob: Removed them

atlas/packages/app/src/components/VideoBestMatch.tsx
fcescob: Done!

atlas/packages/app/src/components/Navbar.tsx
fcescob: Done!

atlas/packages/app/src/components/VideoBestMatch.tsx
fcescob: Removed

atlas/packages/app/src/components/Navbar.tsx
fcescob: It was used in an old animated version, removed it for now.

atlas/packages/app/src/components/Navbar.tsx
fcescob: Thanks for catching that, replaced that with `false`

atlas/packages/app/src/components/Navbar.tsx
fcescob: Done!

fcescob:
> Apart from the comments, some things that will need improvement:
> 
> ### UI
> * We should indicate no search results, currently if there is nothing found, we're left with an empty screen with just the tabs
> * The navbar changes its height when focusing/closing the searchbar, I don't think that's desired
> * Should we present videos search result in a carousel? We have a lot of screen space, that seems like a weird choice. Especially in the videos tab when there is nothing else to browse. @bedeho
> * Navigation icons could use some more contrast - it's hard to spot them when they're this gray


Regarding this:

- Let's handle this in #1289 
- This is part of the design, the height of the navbar goes from `72px` to `64px` when focused
- This is a good point, I will change that to be like in the `VideoView` , I've extracted a `VideoGrid` component and used that. 
- I used `gray[600]` for the icons as that is what the arrow icons where using in the design, so I think this is the intended contrast for icons on the navbar as specified by the design, but we can open an issue about it and discuss if we feel like it makes them hard to see.

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const [search, setSearch] = useState('')
+  const [isSearching, setIsSearching] = useState(false)
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearch(e.currentTarget.value)
+  }
+
+  const handleKeyPress = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter' || (e.key === 'NumpadEnter' && search.trim() !== '')) {
+      navigate(`/search/${search}`)
```
kdembler: Let's use the value from `routes` here

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -0,0 +1,31 @@
...
+  onBlur,
+  onSubmit,
+  ...hmtlProps
+}) => {
+  return (
+    <Input
+      value={value}
+      placeholder={placeholder}
+      type="search"
+      autoSave="some_unique_value"
```
kdembler: What is that value?

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -0,0 +1,31 @@
...
+      name="s"
```
kdembler: Why `s`?

kdembler (commented):
Also, there's still this: `You're repeatedly using raw pixel values. Shouldn't we use spacing values from the theme instead?`. What's your take on that?

kdembler:
One more comment to what you wrote:
> I used gray[600] for the icons as that is what the arrow icons where using in the design

That's actually not true. The design uses this color for a disabled arrow, when you cannot navigate forward. The go back arrow has white color, which we should use for the navigation icons

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: Good catch thanks!

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
fcescob: `autoSave` is a non standard attribute that's supposed to persist the input string between page refreshes, I had seen it and it looked useful, so I've added it, thinking about it, this is useless with React. So I'm removing it now.

https://www.wufoo.com/html5/autosave-attribute/

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
fcescob: Removed thanks.

fcescob:
> One more comment to what you wrote:
> 
> > I used gray[600] for the icons as that is what the arrow icons where using in the design
> 
> That's actually not true. The design uses this color for a disabled arrow, when you cannot navigate forward. The go back arrow has white color, which we should use for the navigation icons

I thought this was the case since the other clickable elements like the searchbar  had a similar contrast, and I assumed the navigation arrows being white as being focused or hovered. Anyway more contrast never hurts, so I'm moving them to be `gray[300]` like the non selected text in the tabs.

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
kdembler: @Gamaranto I think this isn't removed

kdembler (approved):
LGTM

bedeho (approved):

Issue #525, opened 2020-06-01
Initial membership module support in query node


bedeho:
We already have an active membership registry in the genesis block of the current testnet, so bootstrapping is required. Here we just want a simple search across all handles.

```graphql
type Member {

  id: ID!,

  handle: String!,

  avatar_uri: String,

  registered_at_block: BigInteger!,

  about: String,

  controller_account: Bytes!,

  root_account: Bytes!,
}
```

PR #1122, opened 2020-08-04
Query Node API Sketch Discussion PR

bedeho:
A PR created for the benefit of discussion in Atlas team.

### DIFF
--- a/atlas/query_node_API_sketch.graphql
+++ b/atlas/query_node_API_sketch.graphql
@@ -1,0 +1,196 @@
+
+enum Language {
+  Chinese
+  English
+  Arabic
+  Portugese
+  French
+  # ...
+}
+
+type Channel {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  owner: Member!
+
+  handle: String!
+
+  description: String!
+
+  coverPhotoURL: String!
+
+  avatarPhotoURL: String!
+
+  isPublic: Bool!
+
+  isCurated: Boolean!
+
+  language: Language
+
+  videos: [Video!]
+}
+
+type Category {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  name: String!
+
+  videos: [Video!]
+}
+
+type JoystreamVideoMediaLocation {
+  dataObjectID: BigInteger!
+}
+
+type HTTPVideoMediaLocation {
+  host: String!
+  port: Int
+}
+
+# In the future we can add IPFS, Dat, etc.
+union MediaLocation = JoystreamMediaLocation | HTTPMediaLocation
+
+# Mixed both encoding and containers, only having popular combos, may need to be changed later.
+enum VideoMediaEncoding {
+  H264_mpeg4
+  VP8_WEBM
+  Theroa_Vorbis
+}
+
+# Apparently there are lots of different Creative Commons licenses,
+# read about all here https://creativecommons.org/licenses/,
+# I haven't had the time.
+enum CreativeCommonsVersion {
+  CC_BY
+  CC_BY_SA
+  CC_BY_ND
+  CC_BY_NC
+  CC_BY_NC_SA
+  CC_BY_NC_ND
+}
+
+type CreativeCommonsLicense {
+  version:
+}
+
+type UserDefinedLicense {
+  text: String!
+}
+
+union License = UserDefinedLicense | CreativeCommonsLicense
+
+type VideoMedia {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  encoding: VideoMediaEncoding!
+
+  # Resolution width
+  pixelWidth: Int!
+
+  # Resolution height
+  pixelHeight: Int!
+
+  # Size in bytes
+  size: BigInteger
+
+  # where to find
+  location: MediaLocation!
+}
+
+type Video {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  channel: Channel!
+
+  category: Category!
+
+  title: String!
+
+  description: String!
+
+  # In seconds
+  duration: Int!
+
+  # In intro
+  skippableIntroDuration: Int
+
+  thumbnailURL: String!
+
+  Language: Language
+
+  media: VideoMedia!
+
+  hasMarketing: Boolean
+
+  # Timestamp of block
+  publishedOnJoystreamAtblockHeight: BigInteger!
+  publishedOnJoystreamAtTimeStamp: BigInteger!
+
+  # Possible time when video was published before Joystream
+  publishedBeforeJoystream: DateTime
+
+  isPublic: Bool!
+
+  isCurated: Boolean!
+
+  isExplicit: Boolean!
+
+  license: License!
+}
+```
+
+## Queries
+
+```graphql
+
+union FreeTextSearchResultItemType = Video | Channel
+
+type FreeTextSearchResult {
+
+   item: FreeTextSearchResultItemType!
+
+   rank: Int!
+}
+
+type Query {
+
+  # Lookup a channel by its ID
+  channel(id: ID!): Channel
+
+  # List all channel by given constraints
+  channels(order_by_creation_date: Boolean, ignore_curated: Boolean, ignore_non_public: Boolean, offset: Int, limit: Int): [Channel!]
+
+  # Lookup a channel by its ID
+  category(id: ID!): Category
+
+  # List all categories
+  categories(): [Category!]
+
+  # Lookup video by its ID
+  video(id: ID!): Video
+
+  # List all videos by given constraints
+  videos(order_by_publication_date: Boolean, ignore_curated: Boolean, ignore_non_public: Boolean, in_category_with_ID: ID, offset: Int, limit: Int): [Video!]
+
+  # List all top trending videos
+  featured_videos(): [Video!]
+
+  # Free text search across videos and channels
+  search(query_string: String!): [FreeTextSearchResult!]
+}


atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  owner: Member!
+
+  handle: String!
+
+  description: String!
+
+  coverPhotoURL: String!
```
kdembler, line 24: Should this and `avatarPhotoURL` be required?

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+   rank: Int!
+}
+
+type Query {
+
+  # Lookup a channel by its ID
+  channel(id: ID!): Channel
+
+  # List all channel by given constraints
+  channels(order_by_creation_date: Boolean, ignore_curated: Boolean, ignore_non_public: Boolean, offset: Int, limit: Int): [Channel!]
```
kdembler, line 177: I'm wondering whether we should maybe use some more generic ordering and filtering.

Currently, for ordering we have only `order_by_creation_date` but maybe in the future we would like to sort on some other value. Instead of adding `order_by_xxx` boolean later on, maybe we could just do `order_by` that would accept some generic value that would describe the desired sorting. I'm sure there are some existing formats we could use for that.

The same stands for `ignore_curated` and `ignore_non_public` - that could be some general `filter` field

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+
+  description: String!
+
+  # In seconds
+  duration: Int!
+
+  # In intro
+  skippableIntroDuration: Int
+
+  thumbnailURL: String!
```
kdembler, line 133: I'm wondering whether adding some metadata for the poster makes sense, like the image dimensions for example. What do you think?

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+
+  videos: [Video!]
+}
+
+type JoystreamVideoMediaLocation {
+  dataObjectID: BigInteger!
+}
+
+type HTTPVideoMediaLocation {
+  host: String!
```
kdembler, line 54: Is there some specific reason for which we'd need separate `host` and `port` fields? For all the other assets we just pass `URL` string (which can contain the port as well)

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+  French
+  # ...
+}
+
+type Channel {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
```
kdembler, line 16: Could you shed some light on `id` vs `entityID`? Both me and @Gamaranto seem to be not sure what these two represent

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+  category(id: ID!): Category
+
+  # List all categories
+  categories(): [Category!]
+
+  # Lookup video by its ID
+  video(id: ID!): Video
+
+  # List all videos by given constraints
+  videos(order_by_publication_date: Boolean, ignore_curated: Boolean, ignore_non_public: Boolean, in_category_with_ID: ID, offset: Int, limit: Int): [Video!]
```
kdembler, line 189: Usually when working with paginated data, I'd expect the returned object to contain both the data and pagination metadata. Based on the pagination scheme that metadata would for example be the total number of items, current offset and limit, etc. Is it a conscious decision to not return those?

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+
+  avatarPhotoURL: String!
+
+  isPublic: Bool!
+
+  isCurated: Boolean!
+
+  language: Language
+
+  videos: [Video!]
```
kdembler, line 34: Shouldn't those by paginated?

If so, I'd expect a different return type, more specifically what I've mentioned in https://github.com/Joystream/joystream/pull/1122#discussion_r465643870

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+type Category {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  name: String!
+
+  videos: [Video!]
```
kdembler, line 46: Same here, I guess those should be paginated

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+}
+
+type Channel {
+
+  id: ID!
+
+  # Id of underlying entity.
+  entityID: BigInteger!
+
+  owner: Member!
```
kdembler, line 18: The `Member` type seems to be missing?

atlas/query_node_API_sketch.graphql
```diff
@@ -0,0 +1,196 @@
...
+  channel(id: ID!): Channel
+
+  # List all channel by given constraints
+  channels(order_by_creation_date: Boolean, ignore_curated: Boolean, ignore_non_public: Boolean, offset: Int, limit: Int): [Channel!]
+
+  # Lookup a channel by its ID
+  category(id: ID!): Category
+
+  # List all categories
+  categories(): [Category!]
```
kdembler, line 183: This should probably be `[Category!]!` (notice one additional exclamation mark). What this says is that the result of the query will always be an array. It can be empty, but it won't be `null`. This also applies for other queries:
- `channels`
- `categories`
- `videos `
- `featured_videos`
- `search`

atlas/query_node_API_sketch.graphql
bedeho: Indeed, use some variation of this for now

https://github.com/Joystream/joystream/issues/525

atlas/query_node_API_sketch.graphql
bedeho: Probably not.

atlas/query_node_API_sketch.graphql
bedeho: You are correct.

With the current encoding, you could end up even having contradictory specifications about how to order, you could have `order_by_x` and `order_by_y` both be true, which is not semantically meaningful. The API we will get will follow the `OpenCRUD` GraphQL standard, at least in spirit, you can check it out here https://www.opencrud.org/.

In that system, there would be exactly the solution you suggested, see here

https://www.opencrud.org/#sec-Queries-Top-level.Example

I think the remedy here would be to get you an example of this from the query node (Hydra), so you can get it exactly right because my memory is obviously a poor guide.

atlas/query_node_API_sketch.graphql
bedeho: I don't see the immediate upside. Happy to add it if helps us.

atlas/query_node_API_sketch.graphql
bedeho: No, this actually even seems broken, because you cant even give a full path, so lets just have a single URL string.

atlas/query_node_API_sketch.graphql
bedeho: `id` is the instance identifier for `VideoMedia` as a concept in the query node, while `entityID` is the identifier used for the same concept as it exists in the underlying on-chain data model (the content directory). They are distinct. If you want to directly look something up in the content directory, e.g. talking to a full node, requires the latter.

atlas/query_node_API_sketch.graphql
bedeho: Again, same answer about `OpenCrud`, let me give you the actual schema structure Hydra spits out, then we don't need to guess.

atlas/query_node_API_sketch.graphql
bedeho: You mean  `videos`? if so, no, its just a collection of all videos in the channel. If you want a paginated list of all videos _in a given channel_, you have to do a video query, and filter on the channel membership property, and then add pagination.

atlas/query_node_API_sketch.graphql
bedeho: no.

atlas/query_node_API_sketch.graphql
bedeho: Agree.

atlas/query_node_API_sketch.graphql
kdembler: I was thinking about cases like handling vertical posters for example. But actually atm I'm not sure whether we will need this. Let's skip it for now I guess

atlas/query_node_API_sketch.graphql
kdembler: > `id` is the instance identifier for `VideoMedia` as a concept in the query node

Does that mean that this ID will be different for each query node instance?

atlas/query_node_API_sketch.graphql
kdembler: Okay, thanks

fcescob (approved):

atlas/query_node_API_sketch.graphql
bedeho: Yes in principle it could be, but in practice, I would not be surprised if it was not. But for safety, if you need to mix responses across nodes, it would be better to use the underlying entityID. If this creates nastiness, then we should look more closely at whether the guarantees about the `id` are stronger than I am assuming.

kdembler:
All of the raised points were addressed either in the merged schema or the one currently on WIP branches, alongside some other improvements. We can mark the discussion as done so closing the PR

Issue #1066, opened 2020-07-28
Network testing: types refactoring


gleb-urvanov:
Refactor the code to use explicit types as return values of api wrapper methods

gleb-urvanov:
PR: https://github.com/Joystream/joystream/pull/1125

Issue #1013, opened 2020-07-21
Network testing: update codebase


gleb-urvanov:
Update tests to work with new runtime and new types

gleb-urvanov:
https://github.com/Joystream/joystream/pull/1166

mnaamani:
Complete in https://github.com/Joystream/joystream/pull/1225

PR #1125, opened 2020-08-04
Network testing: explicit id types refactoring

gleb-urvanov:
Issue: https://github.com/Joystream/joystream/issues/1066
Since polkadot api 0.96.1 are in use, typecast is used to convert Call to Submittable Extrinsic. These casts should be removed after migration to newer version of the api as part of https://github.com/Joystream/joystream/issues/1013


mnaamani:
Refreshing to see this simple but much needed refactor. Just needs update from iznik branch and merge conflict resolution

Issue #896, opened 2020-07-04
Election System


bedeho:
# Election System

## Goal

The main goal in this iteration of the election system is to primarily make it possible to allow stake already deployed for other purposes, such as validation or in working groups, to be simultaneously used in the election system, both for voting and candidacy. A secondary goal is to remedy a collection of problems that on their own would not have required an invasive rewrite. This election system will be developed and deployed in concert with changes to other parts of the system, such as the council and membership, that facilitate this change, but the focus of this specification is the election system only.

### Locks

This design critically depends on currency locks on accounts, which stack on top of each other, and have independent expiry times. They support locking only for specific purposes, and in this design the broadest possible set of purposes is used in all election locks.

### Roles

There are two roles in an election cycle, a voter or a candidate. A given member can enter into, at most, one of these roles in a given election cycle (see below). A voter attempts to influence what candidates become the winners of the election, and a candidate is running to be a council member. Both must stake, and the stake of both impacts the outcome.

### Staking

The model takes as given that funds to be used for staking exist in one or more accounts that are associated with a given member. These accounts may have locks encumbering funds for independent purposes, and it is precisely the stacking property of locks which yield the desired property that those same funds can have independent staking commitments. The election system only needs a way to ask the runtime that a given account is indeed under the control of a given member. **It assumes that no two members can claim to operate the same account for staking**. Since stake held in some account may be transferred to another account, for example, due to loss of the staking key, the election system needs a way to learn about such transfers as they occur, so that it can reapply the election-related locking. This may occur at any time. The set of accounts provided as sources of funds for staking may change, for the same actor, at different points in time.

#### Locks

For a given account, the election system uses up to three distinct locks. One lock for the voting stake, one lock for candidacy in even-numbered election cycles, and another lock for candidacy in odd-numbered election cycles. The reason for having distinct locks is that it makes it much easier to handle the possible partial, or complete, reuse of funds staked in a prior election for the next one, and possibly for a different purpose. All of these locks are collectively referred to as _election locks_.

When anyone wishes to stake as part of the election process, they will provide a set of accounts, and for each account provide some amount of funds they wish to lock up for the given purpose. It will then be checked that

- the accounts are distinct
- the member does indeed control all such accounts
- each account has sufficient funds, free or not, to cover the stated amount for staking.
- that the total staked amount across all accounts covers the totals, or minimum bounds, required for the given purpose.

#### Invariants

[...]

PR #1133, opened 2020-08-06
Council rework referendum

ondratra:
Rework of elections / council according to #896.

This PR focuses on referendum module.


ondratra:
Builds on Travis are failing due to output timeout after the Substrate upgrade, but all checks pass when run on dev environment.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+use std::collections::HashSet;
+
+// declared modules
+mod mock;
+mod tests;
+
+/////////////////// Data Structures ////////////////////////////////////////////
+
+#[derive(Encode, Decode, PartialEq, Eq, Debug)]
+pub enum ReferendumStage {
```
bedeho: Can we rename from `Void`, to something more descriptive, such as `Inactive` or `NoRefferendum`.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    ) -> bool;
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait<I>, I: Instance> as Referendum {
+        /// Current referendum stage
+        pub Stage get(fn stage) config(): (ReferendumStage, T::BlockNumber);
+
+        /// Options of current referendum
+        pub ReferendumOptions get(fn referendum_options) config(): Option<Vec<T::ReferendumOption>>;
```
bedeho: This field has two problems

1) There is no need for any representation of the different voting alternatives themselves, you only need to know how many there are, and then use identifiers 0...#number-1 to identify them, and report the result. Too much information has leaked into this module abstraction.

2) This state information has certain static invariants w.r.t. the `stage` field, specifically, there can only be such a number defined for two of the three stages. There can be no value defined for current `Void` stage. Using a option that lives outside is not the safest representation you can pick. You could here consider doing something like 

```Rust

pub enum ReferendumStage {
    Inactive,
    Active(ActiveRefferendumState)
}

pub struct ActiveRefferendumState {

   pub nr_of_alternatives: T::ReferendumOption
   pub active_refferendum_stage: ActiveRefferendumStage
}

pub enum ActiveRefferendumStage {
  ...
}
```

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    fn free_currency(
+        account: &<Self as system::Trait>::AccountId,
+        balance: &Self::CurrencyBalance,
+    ) -> bool;
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait<I>, I: Instance> as Referendum {
+        /// Current referendum stage
+        pub Stage get(fn stage) config(): (ReferendumStage, T::BlockNumber);
```
bedeho: Its not clear to me what is being represented by the bocknumber dimension of this type.

If this is mean to hold the block where voting period started, or revealing period started -  which is indeed required due to the `on_finalized` procesing, then this representation is not correct, because there is no meaning to when we are in the idle/void stage.  See other review comment about improving representation of `ReferendumStage` below.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        pub ReferendumOptions get(fn referendum_options) config(): Option<Vec<T::ReferendumOption>>;
+
+        /// Votes in current referendum
+        pub Votes get(fn votes) config(): map hasher(blake2_128_concat) T::AccountId => SealedVote<T::Hash, T::CurrencyBalance>;
+
+        /// Revealed votes counter
+        pub RevealedVotes get(fn revealed_votes) config(): map hasher(blake2_128_concat) T::ReferendumOption => T::VotePower;
+
+        /// Target count of referendum winners
+        pub WinningTargetCount get(fn winning_target_count) config(): u64;
```
bedeho: Like other fields, this cannot be set when void, please coordinate with `ReferendumStage´ type construction.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+
+impl<T, U> Default for ReferendumResult<T, U> {
+    fn default() -> ReferendumResult<T, U> {
+        ReferendumResult::NoVotesRevealed
+    }
+}
+
+/////////////////// Trait, Storage, Errors, and Events /////////////////////////
+
+pub trait Trait<I: Instance>: system::Trait {
```
bedeho: I see many types trait constraints have conversion to/from `u64` as required, is it, why?

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + PartialEq
+        + From<u64>
+        + Into<u64>;
+
+    type VotePower: Parameter
```
bedeho, line 126: So far as I understand, this really should not be a separate type, as voting power really is already defined in terms of #coins , and this already has type coming from balances.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        type Error = Error<T, I>;
+
+        /// Setup events
+        fn deposit_event() = default;
+
+        /////////////////// Lifetime ///////////////////////////////////////////
+
+        // start voting period
+        #[weight = 10_000_000]
+        pub fn start_referendum(origin, options: Vec<T::ReferendumOption>, winning_target_count: u64) -> Result<(), Error<T, I>> {
```
bedeho: This is not an extrinsic, it's a public module routine. We probably also need a public trait defined by this module, e.g. `ReferrendumManager`  which exposes this call, so that other client modules can just depend on a runtime having this trait, rather than directly depending on this module. This makes mocking of this module easier during testing of those client modules.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            Mutations::<T, I>::start_revealing_period();
+
+            // emit event
+            Self::deposit_event(RawEvent::RevealingStageStarted());
+
+            Ok(())
+        }
+
+        #[weight = 10_000_000]
+        pub fn finish_revealing_period(origin) -> Result<(), Error<T, I>> {
```
bedeho: This is not an extrinsic, no key can trigger this, this should be triggered by `on_finalized`, as it happens when a certain numbe of blocks have passed.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+
+            // emit event
+            Self::deposit_event(RawEvent::ReferendumStarted(options, winning_target_count));
+
+            Ok(())
+        }
+
+        // finish voting period
+        #[weight = 10_000_000]
+        pub fn finish_voting_start_revealing(origin) -> Result<(), Error<T, I>> {
```
bedeho: This is not an extrinsic, no key can trigger this, this should be triggered by `on_finalized`, as it happens when a certain numbe of blocks have passed.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct SealedVote<Hash, CurrencyBalance> {
+    commitment: Hash,
+    stake: CurrencyBalance,
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, Debug)]
+pub enum ReferendumResult<ReferendumOption, VotePower> {
```
bedeho: I don't know whether you actually need this type, my next deeper review will look into this, however, if you want to represent outcomes, this is not the best type. Ill leave it for you to review based on your final rewrite.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+decl_storage! {
+    trait Store for Module<T: Trait<I>, I: Instance> as Referendum {
+        /// Current referendum stage
+        pub Stage get(fn stage) config(): (ReferendumStage, T::BlockNumber);
+
+        /// Options of current referendum
+        pub ReferendumOptions get(fn referendum_options) config(): Option<Vec<T::ReferendumOption>>;
+
+        /// Votes in current referendum
+        pub Votes get(fn votes) config(): map hasher(blake2_128_concat) T::AccountId => SealedVote<T::Hash, T::CurrencyBalance>;
```
bedeho: Votes should not be identified with an account, you vote as a member not as a raw account holder.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+
+        /// Revealed votes counter
+        pub RevealedVotes get(fn revealed_votes) config(): map hasher(blake2_128_concat) T::ReferendumOption => T::VotePower;
```
bedeho: Since the number of alternatives will always be quite small, we dont need to lift this map out of the safest possible place to put it, which is in ReferendumStage

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        + PartialEq
+        + From<u64>
+        + Into<u64>;
+
+    type VoteStageDuration: Get<Self::BlockNumber>;
+    type RevealStageDuration: Get<Self::BlockNumber>;
+
+    type MinimumStake: Get<Self::CurrencyBalance>;
+
+    fn is_super_user(account_id: &<Self as system::Trait>::AccountId) -> bool;
```
bedeho: There is no such concept in this module.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            // emit event
+            Self::deposit_event(RawEvent::ReferendumFinished(referendum_result));
+
+            Ok(())
+        }
+
+        /////////////////// User actions ///////////////////////////////////////
+
+        #[weight = 10_000_000]
+        pub fn vote(origin, commitment: T::Hash, stake: T::CurrencyBalance) -> Result<(), Error<T, I>> {
```
bedeho: The spec clearly describes how voting involves specifying in detail what accounts should be drawn upon for how much stake. It's also not clear where the locks are being applied for the actual voting, this is required, otherwise what would prevent different people voting with the same stake again and again? `can_vote` cannot do this without mutating locks, which is not what it should be doing. Lastly, be sure in the fixed version that the council module excludes council candidates from being able to vote in the implementation of this.

You are also not doing any checks as to whether there is sufficient free stake to do the voting. The call to `can_vote` cannot do any enforcement here, because it would have to update locks to reflect the fact that an actual vote had been made

bedeho (changes requested):
There are quite a few deep problems here, so I think you need to go back and read the initial implementation to make sure you are 100% onboard, and also to make sure it has no problems.

The biggest problem so far as almost all of the sophisticated stake management related logic has just been omitted entirely. I am not sure if you perhaps have put that into the council code? if so, then that would break the idea of having this be a reusable voting module.

I will review again from scratch when you have addressed these initial set of issues identified, also bare in mind:

1. In general, try to give better comments on what things represent.

2. Try to pick the safest most expressive types, and avoid anonymous types like tuples wrapped in vectors for example.

ondratra:
The Referendum module was created completely agnostic of the council. It is meant as an abstraction for a voting process that expects voters to stake some amount of currency to give value to their vote. Staking logic itself is not implemented inside of module - it only expects Trait will implement `has_sufficient_balance(...)`, `lock_currency(...)`, `free_currency(...)` that will actually handle the staking. **EDIT: I will look on staking again and possibly rework it**

The module is meant to be used in two main ways. The first use is as a standalone Referendum - some superusers can initialize the voting process for some referendum options. In this case, the referendum has only informational value without any programmatical consequences to the blockchain.

The second way to use the Referendum module is to wrap it's extrinsics in another module, thus enhance its features and create programmatical consequences to the vote results. This approach is used for the council. The council will handle all the staking logic, candidacy announcements, etc.

runtime-modules/referendum/src/lib.rs
ondratra: Definitely.

runtime-modules/referendum/src/lib.rs
ondratra: 1. The complex representation of voting options (convertible to `u64`) is useful mainly when the Referendum module is used as a standalone module. The mapping of vote results to 0..n is possible only when the module is wrapped in another module (council case).
2. ~~To maximize possibilities for extending this module, the block number when the referendum stage changed is saved for all stages. This is probably not useful for the council but can be useful for other potential uses.~~

runtime-modules/referendum/src/lib.rs
ondratra: I will try to revise all of them and remove these constraints from Trait where possible.

runtime-modules/referendum/src/lib.rs
ondratra: VotePower gives more flexibility to how votes should be counted. Depending on `fn caclulate_vote_power(account_id, stake)` implementation it can handle for example these situations:
- each account has one vote (`return 1;`)
- the voting power is linear with a stake (`return stake;`) - council's case
- 1st class citizen has higher vote power than 2nd class citizen (`return if account_id == MY_1ST_CLASS_CITIZEN {stake * 10} else {stake} `)
- possibly others

runtime-modules/referendum/src/lib.rs
ondratra: Good idea (`ReferrendumManager`). Will implement that.

runtime-modules/referendum/src/lib.rs
ondratra: `on_finalized` usage is missing right now on more places. I will rework these parts.

runtime-modules/referendum/src/lib.rs
ondratra: `on_finalized` usage is missing right now on more places. I will rework these parts.

runtime-modules/referendum/src/lib.rs
ondratra: Misunderstood this one. Will introduce members instead of accounts to all relevant places.

runtime-modules/referendum/src/lib.rs
ondratra: The superuser role is useful when the referendum module is used as a standalone module. I will revise this mechanic again after I introduce on_finalized logic to the module.

runtime-modules/referendum/src/lib.rs
ondratra: Stage has been reworked.

runtime-modules/referendum/src/lib.rs
ondratra: This has been reworked.

runtime-modules/referendum/src/lib.rs
ondratra: Done.

runtime-modules/referendum/src/lib.rs
ondratra: This enum handles all possible situations referendum can result in:
- there are X winners as requested
- X winners were expected but Xth winner has the same amount of votes as (X+1)th option (can't decide only X winners)
- X winners were expected but only Y (`Y < X`) options received any votes
- nobody reveals a valid vote in referendum

Of course, only the first 2 cases can possibly happen in the council because candidates' stakes in the council are added to total vote sum.

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Superuser can run an extrinsic version of `start_referendum`. As stated before, this role makes sense when the referendum is used as a standalone module. Council module can omit this feature simply by implementing `is_super_user(...) {false}` in the runtime.

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Reworked.

runtime-modules/referendum/src/lib.rs
ondratra: Reasons for these constraints were removed. No need for them now.

ondratra:
The following changes were introduced:
- Use LockableCurrency trait.
- One account one candidate voting. (no membership)
- Unstaking driven by voter.
- Unstaking during idle period possible for loosing stake. 
- Unstaking for winners, only possible after idle period.
- Fixes
 -- Drop panics
 -- Drop ReferendumOptions
 -- Represent only #candidates, not Ids

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+        pub Stage get(fn stage) config(): ReferendumStage<T::BlockNumber, T::VotePower>;
+
+        /// Votes in current referendum
+        pub Votes get(fn votes) config(): map hasher(blake2_128_concat) T::AccountId => SealedVote<T::Hash, Balance<T, I>>;
+
+        /// Index of the current referendum cycle.
+        pub CurrentCycle get(fn current_cycle) config(): u64;
+
+        /// Last cycle winning option(s)
+        pub PreviousCycleResult get(fn previous_cycle_result) config(): ReferendumResult<u64, <T as Trait<I>>::VotePower>;
```
bedeho: This value should not be set when we are in the voting or revealing stage because the vales no longer serve any purpose at that time, however, this representation still allows this. I suggest you bake this into the `Inactive` variant of `Stage`.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+decl_storage! {
+    trait Store for Module<T: Trait<I>, I: Instance> as Referendum {
+        /// Current referendum stage
+        pub Stage get(fn stage) config(): ReferendumStage<T::BlockNumber, T::VotePower>;
+
+        /// Votes in current referendum
+        pub Votes get(fn votes) config(): map hasher(blake2_128_concat) T::AccountId => SealedVote<T::Hash, Balance<T, I>>;
+
+        /// Index of the current referendum cycle.
+        pub CurrentCycle get(fn current_cycle) config(): u64;
```
bedeho: Rename to `CurrentCycleId`

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+impl<BlockNumber, VotePower> Default for ReferendumStage<BlockNumber, VotePower> {
+    fn default() -> ReferendumStage<BlockNumber, VotePower> {
+        ReferendumStage::Inactive
+    }
+}
+
+/// Representation for voting stage state.
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct ReferendumStageVoting<BlockNumber> {
+    start: BlockNumber,
```
bedeho: `started`

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+    winning_target_count: u64,
+    options_count: u64,
```
bedeho: When starting an election, you check that `options_count` is greater than 0 and separately greater than `winning_target_count`. You have an error check on this that the caller must possibly deal with.

You can rid of all of that, and also get rid of the possible risk at a later point that the invariant is broken by changing your representation. Instead of having two independent counts, have `options_count` be replaced by a variable that holds the number of alternatives _above and beyond the target_. So in this case, even a value of `0` is fine, and value of `1` means you have one more alternative than the target, and so on. This is safer all round.

This may seem like paranoia, but for runtime work we must squeeze out every inch of safety we can get from the type system.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+}
+
+/// Representation for revealing stage state.
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct ReferendumStageRevealing<BlockNumber, VotePower> {
+    start: BlockNumber,
+    winning_target_count: u64,
+    options_count: u64,
+    revealed_votes: Vec<VotePower>,
```
bedeho: Assuming this holds the cumulative voting power backing each alternative:
- the name needs to really improve
- we don't need `options_count` as a separate variable then, length of vector already sufficient.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+            return Err(Error::InsufficientStake);
+        }
+
+        // ensure account can lock the stake
+        if T::Currency::total_balance(&account_id) < *balance {
+            return Err(Error::InsufficientBalanceToStakeCurrency);
+        }
+
+        // ensure user haven't vote yet
+        if Votes::<T, I>::contains_key(&account_id) {
```
bedeho: I think its a bit impractical to not allow people to vote using funds that were locked for an old cycle. Many people will stake funds for a long time, and its not very ergonomic to require them to first unstake manually, then vote again for each account, all the time.
So I think we should allow reuse of an account that has funds staked for an expired cycle.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+    ) -> bool;
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait<I>, I: Instance> as Referendum {
+        /// Current referendum stage
+        pub Stage get(fn stage) config(): ReferendumStage<T::BlockNumber, T::VotePower>;
+
+        /// Votes in current referendum
+        pub Votes get(fn votes) config(): map hasher(blake2_128_concat) T::AccountId => SealedVote<T::Hash, Balance<T, I>>;
```
bedeho: Can we quickly explain how mappings enter and leave this map and leave it here? It's the most important dynamic that I have to keep reminding myself. My understanding is that an account is added when you vote, and it is removed only when you recover stake, and this removal can be as early as going into inactive stage if it backed a loser in the current election or any time after leaving the inactive stage for the first time. Correct?

This description perhaps shold also include reuse of a vote, if you add this feature as I mention below.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+pub struct ReferendumStageRevealing<BlockNumber, VotePower> {
+    start: BlockNumber,
+    winning_target_count: u64,
+    options_count: u64,
+    revealed_votes: Vec<VotePower>,
+}
+
+/// Vote casted in referendum but not revealed yet.
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct SealedVote<Hash, Currency> {
```
bedeho: The name of this type has a mismatch with semantics. I would just call it a `Vote`, which can be either in the sealed  or unsealed stage.

runtime-modules/referendum/src/lib.rs
bedeho: First off, the payload should be `BTreeMap<AlternativeId, VotePower>`, this provides a guarantee that the same alternative only has one specified voting power.

Second, we are keeping `ReferendumResult` internal to this module only because its required when checking whether a voter can recover stake during inactive period of same cycle. The problem is the details of what policy to apply in allowing someone to unstake, or not, should probably not be in this module to begin with. Look at `voted_for_winner_last_cycle`, you see that we - arbitrarly, have decided that if you voted for one among the `NotEnoughWinners`, then you cannot unstake! That does not make sense in the context of the council. Actually, even the policy for `ExtraWinners` does not make sense, because the council module may have some use case specific rule about whether to restart an election or just pick some subset at random. So both of these show that the details about whether to allow unstaking of a voter should live outside this module.

The best alternative would perhaps be to 
- move the logic for checking whether unstaking for a vote is allowed out of this module, and make it a runtime trait method, something like (roughly)
`fn can_unstake(SealedVote) -> bool`. This can allow rules like: if you did not unseal, then you must wait N blocks, if you unsealed, but voted for loser, you must wait M blocks, etc.
- Remove storage of `ReferendumResult` in this module.

Please check that this proposal makes sense, its a little bit non-trivial.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+
+        // ensure referendum is running
+        let stage_data = match stage {
+            ReferendumStage::Revealing(tmp_stage_data) => (tmp_stage_data),
+            _ => return Err(Error::RevealingNotInProgress),
+        };
+
+        let current_block = <system::Module<T>>::block_number();
+
+        // ensure voting stage is not expired (it can happend when superuser haven't call `finish_voting_start_revealing` yet)
```
bedeho: I think this should not be here anymore.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+        // ensure referendum is running
+        let voting_stage = match stage {
+            ReferendumStage::Voting(stage_data) => (stage_data),
+            _ => return Err(Error::ReferendumNotRunning),
+        };
+
+        let current_block = <system::Module<T>>::block_number();
+
+        // ensure voting stage is not expired (it can happend when superuser haven't call `finish_voting_start_revealing` yet)
+        if current_block >= T::VoteStageDuration::get() + voting_stage.start + One::one() {
```
bedeho: I think this should not be here anymore.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+
+        let current_block = <system::Module<T>>::block_number();
+
+        // ensure voting stage is not expired (it can happend when superuser haven't call `finish_voting_start_revealing` yet)
+        if current_block >= T::RevealStageDuration::get() + stage_data.start + One::one() {
+            return Err(Error::RevealingNotInProgress);
+        }
+
+        // ensure account haven't voted yet
+        if !Votes::<T, I>::contains_key(&account_id) {
```
bedeho: Is there room for creating a `ensure_vote_exists` that can be reused here?

runtime-modules/referendum/src/mock.rs
```diff
@@ -0,0 +1,504 @@
...
+        assert_eq!(
+            system::Module::<Runtime>::events().last().unwrap().event,
+            TestEvent::event_mod_Instance0(RawEvent::VoteCasted(account_id, commitment, balance))
+        );
+    }
+
+    pub fn reveal_vote(
+        origin: OriginType<<Runtime as system::Trait>::AccountId>,
+        account_id: <Runtime as system::Trait>::AccountId,
+        salt: Vec<u8>,
```
bedeho, line 477: I think most hashing is linear time, but not sure how expensive it can get for gigantic inputs, so perhaps good to introduce some upper bound limit on the salt size? What is a good length?

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+
+    fn ensure_regular_user(origin: T::Origin) -> Result<T::AccountId, Error<T, I>> {
+        let account_id = ensure_signed(origin)?;
+
+        Ok(account_id)
+    }
+
+    /////////////////// Action checks //////////////////////////////////////////
+
+    fn can_start_referendum_extrinsic(
```
bedeho: I think `can_start_referendum_extrinsic` can in its entirety be replaced by a runtime `EnsureOrigin` runtime trait. this is the typical way of making extrinsic access customizable. See here for example

https://github.com/paritytech/substrate/blob/master/frame/treasury/src/lib.rs#L146

The reason we do not use it alot is that we often want to to authenticate a memberid + origin together.

In the scheme you have currently implemented, the origin _must_. be an account, but this will generally not be the case, in general it will be some other part of the runtime, win which case the origin needs to be checked to be a `RootOrigin`.

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+                // calculate vote power
+                let vote_power = T::caclulate_vote_power(account_id, &sealed_vote.balance);
+                stage_data.revealed_votes[*option_index as usize] += vote_power;
+            };
+
+        // store revealed vote
+        Stage::<T, I>::mutate(|stage| {
+            match stage {
+                ReferendumStage::Revealing(stage_data) => distribute_vote(stage_data),
+                _ => panic!("Invalid state"), // will never happen
```
bedeho: This is avoidable. Recall that `reveal_vote` is invoked after `can_reveal_vote` has succeeded in the same code path, so you can provide required inputs for this function here with static guarantees about `stage_data` being available.

In fact, the problem is a bit deeper, look at this line above

`stage_data.revealed_votes[*option_index as usize] += vote_power;`

this is not safe, `option_index` can be out of bounds, again despite the fact that you have in fact earlier in the code path ensured that the `option_index` is valid. I suggest you combine this check for validity of `option_index` with a pure computation of the new `revealed_votes` vector based on `stage_data`.

runtime-modules/referendum/src/lib.rs
bedeho: Can you add quick rule about when this is incremented in comment?

runtime-modules/referendum/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+
+        // ensure there is some vote with locked stake
+        if !Votes::<T, I>::contains_key(&account_id) {
+            return Err(Error::NoStakeLocked);
+        }
+
+        let sealed_vote = Votes::<T, I>::get(&account_id);
+
+        // enable stake release only during
+        if cycle_id == sealed_vote.cycle_id {
```
bedeho: This appears to allow someone to release stake if they voted in this cycle and its during voting.

Some of this logic will likely be lifted out, but the rule for the council use case is
- if its a future cycle, its always ok.
- if its the same cycle, then we are in the idle period, and you voted for the loser.

bedeho (changes requested):
I left some changes that will really tighten things up to the final level, but I think overall this is looking excellent now. 

**Reading this module is really pleasant and easy, very well done.**

- We have been messing around with the name here for a while, `referrendum` does not really work because it conflicts with existing Substrate concept. I think one core aspect of our module is that its an _engine_, not a module to use in isolation, this is why it does not have any logic associated with picking candiate sets or doing something with winners/loosers. I therefore we call it `VotingEngine`. This is also analogous to the proposal engine, which plays a similar role in the proposal system. Unless... you have a better one.
- Add comments to struct fields and variants in enums.
- Try to make state names, in particular struct fields, more like human readable nouns or noun phrases.

runtime-modules/referendum/src/lib.rs
ondratra: Comment added `It is incremented everytime referendum ends.`

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Renamed to `intermediate_results`

runtime-modules/referendum/src/lib.rs
ondratra: Reuse enabled.

runtime-modules/referendum/src/lib.rs
ondratra: Renamed to `CastVote`.

runtime-modules/referendum/src/lib.rs
ondratra: Removed

runtime-modules/referendum/src/lib.rs
ondratra: `ensure_vote_exists` introduced

runtime-modules/referendum/src/lib.rs
ondratra: Those two condition will be satisfied in the council.

This particular condition is tied with the following `match` on the next line - it enables user to revoke his vote during voting period.

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Reworked

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/lib.rs
ondratra: Done

runtime-modules/referendum/src/mock.rs
ondratra: Implemented. Good length is something that ensures uniqueness - in our scenarios i think `std::u64::MAX` is optimal.

runtime-modules/referendum/src/lib.rs
ondratra: Description added

runtime-modules/referendum/src/lib.rs
ondratra: The issue continues here - https://github.com/Joystream/joystream/pull/1133#discussion_r477369394

ondratra:
The issues from the previous review were addressed with one exception being *rename Referendum -> VoteEngine* - this is a small change, but it will affect most of the code, so it will be better to create separate PR for that.

I mistakenly clicked on *Resolve conversation* in this issue https://github.com/Joystream/joystream/pull/1133#discussion_r477369394, and it can't be reverted. The things you requested there were implemented, but `Vec<AlternativeId, VotePower>` was not transferred to `BTreeMap<AlternativeId, VotePower>` because the vector is ordered from the winning alternative with the most votes to the winning alternative with the least votes (and BTree is unsorted). All alternatives' vote results (including those that didn't win) can be accessed in the new runtime function `process_results` that is being called after referendum ends.

bedeho (changes requested):
1. Drop `ReferendumResult` and input about
2. Drop sorting on voting power in result in order to use map over vector.

bedeho (changes requested):
I think I made one mistake here, likely because I took a comment from my old review and did not re-evaluate it properly in light of the change we were making.

When we are both representing, and reporting back, the currently accumulated VotingPower across the alternatives, we should only use `Vec<VotingPower>`. Not 

a. `Vec<CandidateId, VotingPower>`: because it allows duplication of the voting power.
b. `BTreeMap<CandidateId, VotingPower>`: because it allows specifying voting power for non-invalid alternatives to a much greater extent, e.g. the key `1231` could be added by mistake.

The most direct representation is just `Vec<VotingPower>`, both for reporting results and representing state underway. Of course, it's still possible to mess up the length, but it's still simpler.

ondratra:
Ok, I've just reverted the last commit and the result is `Vec<VotingPower>` now.

bedeho (approved):
Sweet, LGTM!

PR #1147, opened 2020-08-10
add graphql/miragejs data fetching

kdembler:
This PR adds logic for fetching the data using GraphQL. It also includes a MirageJS GraphQL for mocking the data client-side.

List of notable changes:
- Included support for Scalar types so that, for example, we can get `Date` object directly from the query result
- Added Apollo codegen so we get automatic typings for queries. Also removed some of the typings in favour of the generated ones
- Replaced `luxon` with `date-fns` as `luxon` dates had problems after being frozen by apollo cache mechanism


kdembler:
Tests fails currently with the following error:
```
 SyntaxError: Cannot use import statement outside a module

      1 | import { createServer } from 'miragejs'
    > 2 | import { createGraphQLHandler, mirageGraphQLFieldResolver } from '@miragejs/graphql'
```

I think this may be related to an issue raised here: https://github.com/miragejs/graphql/issues/3 There's currently a PR to the lib open regarding that so let's wait for now and maybe they'll be able to fix it

kdembler:
So the import issue in tests got resolved now thanks to a newer version of GraphQL MirageJS library. However, there's a new issue with tests:
```
    GraphQLError: Syntax Error: Cannot parse the unexpected character ".".

       8 | createServer({
       9 |   routes() {
    > 10 |     const graphQLHandler = createGraphQLHandler(schema, this.schema, {
         |                            ^
      11 |       resolvers: {
      12 |         Query: {
      13 |           videos: (...params: unknown[]) => {
```

Doesn't make much sense since the schema is correct and it works in a regular build. Won't investigate for now, let's maybe try and merge the tests PR, maybe that will help.

fcescob (approved):
LGTM!

bedeho (approved):
LGTM.

fcescob (approved):

PR #1148, opened 2020-08-10
Migrate the 'chain-spec-builder' utility project to the Substrate 2.0.0rc4

shamil-gadelshin:
Features:
- converted `chain-spec-builder` utility to the Substrate 2.0.0rc4
- all traces of the `pallet_contracts` were removed (runimte, rpc, node, chain-spec-builder)
- `AccountInfo` in the `service-discovery` was renamed to `StorageProviderAccountInfo`
- updated `build.rs` scripts in the runtime and node projects.


runtime-modules/service-discovery/src/lib.rs
```diff
@@ -10,8 +10,8 @@
 //!
 //! ## Supported extrinsics
 //!
-//! - [set_ipns_id](./struct.Module.html#method.set_ipns_id) - Creates the AccountInfo to save an IPNS identity for the storage provider.
-//! - [unset_ipns_id](./struct.Module.html#method.unset_ipns_id) - Deletes the AccountInfo with the IPNS identity for the storage provider.
+//! - [set_ipns_id](./struct.Module.html#method.set_ipns_id) - Creates the StorageProviderAccountInfo to save an IPNS identity for the storage provider.
```
mnaamani: I understand the need to change the name here because it conflicts with the balances::AccountInfo struct. And I recommend a better name of `ServiceProviderRecord` because it shouldn't be storage provider specific. I'm not sure how much longer we will keep this module around or if it will be used by other services, but until then we might as well use a general name.

mnaamani:
There is one piece of useful information which we seem to have lost now which is the commit hash in the version string like:

```
MokoBook:substrate mokhtar$ joystream-node --version
joystream-node 2.7.0-dc3b7624f-x86_64-macos
```

Is it possible to add it back?

runtime-modules/service-discovery/src/lib.rs
mnaamani: types library is prepared with this requested change https://github.com/Joystream/joystream/pull/1153

This is the only change request otherwise ready to merge. :+1:

shamil-gadelshin: Done.

shamil-gadelshin:
I restored the node version from the environment variable.

mnaamani (approved):

PR #1149, opened 2020-08-11
Pioneer: Improve proposal details loading flow

Lezek123:
This PR addresses an issue with slow proposal details page loading due to the fact that `api.rpc.state.getKeys` used to fetch all proposal votes is taking quite a long time on the current testnet (~6s during my tests) and blocks other queries that we try to execute simultaneously.

The fix is about:
- Starting to load votes **after** all the basic proposal data (title, rationale, description etc.) is loaded to prevent those queries from beeing blocked by `api.rpc.state.getKeys` and be able to display this data before everything else is loaded.
- Improving the flow  in `useProposalSubscription` to avoid executing queries twice during first load

This fixes the issue only partially, because the votes and proposal discussion will still take quite some time to load. Currently I can't think of any better way of fetching those, but perhaps this will be better optimized in Substrate 2.0 (as the new api provides a standarized way of fetching double-map entries by frist key).
The proposal state should still update after each council member vote without the need to refresh the page.

### DIFF
--- a/pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
+++ b/pioneer/packages/joy-proposals/src/Proposal/ProposalDetails.tsx
@@ -5,7 +5,7 @@ import Body from './Body';
 import VotingSection from './VotingSection';
 import Votes from './Votes';
 import { MyAccountProps, withMyAccount } from '@polkadot/joy-utils/MyAccount';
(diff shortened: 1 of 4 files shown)


mnaamani:
The improvement is noticeable and greatly improves the wait time/experience.

mnaamani (approved):

PR #1150, opened 2020-08-11
Pioneer: Working groups block link fix

Lezek123:
Fixes links in working groups opportinity boxes by using `api.rpc.chain.getBlockHash(height)` instead of `api.query.system.blockHash(height)` to fetch block hash in `joy-roles` transport (the latter doesn't work for older blocks)


mnaamani:
Test against our public endpoint (archival node) 

Screen Shot 2020-08-11 at 11 33 15 AM

We can now see the block hash correctly in the link at the bottom and navigating works correctly.

mnaamani (approved):

PR #1126, opened 2020-08-05
Migrate the node project. 

shamil-gadelshin:
Node project was converted to the Substrate version 2.0.0rc4

Comments:
- build scripts in both runtime and node will be changed during the work on the node_spec_builder.
- tracing, telemetry, weights, benchmarking are out of the scope of the current release
- runtime API, constants, primitives were moved to separate modules
- node executor, service, RPC modules were merged in the single node crate unlike the substrate base implementation

PR #1153, opened 2020-08-11
@joystream/types: Rename AccountInfo, make dependencies exact, add types required by new node

Lezek123:
Updated `@joystream/types` following the `AccountInfo` type rename in the runtime (to `StorageProviderAccountInfo`): https://github.com/Joystream/joystream/pull/1148/commits/e01eb1cfc625be6da976bd9903132c591457861e

The reason this rename was introduced is because the `AccountInfo` type name was colliding with one the new Substrate core types (see: https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/system/types.ts).

I also fixed `yarn.lock` which seemed outdated on the current `iznik` branch and was causing some dependency issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,13 +23,13 @@
     "devops/prettier-config"
   ],
   "resolutions": {
-    "@polkadot/api": "^1.26.1",
-    "@polkadot/api-contract": "^1.26.1",
-    "@polkadot/keyring": "^3.0.1",
-    "@polkadot/types": "^1.26.1",
-    "@polkadot/util": "^3.0.1",
-    "@polkadot/util-crypto": "^3.0.1",
-    "@polkadot/wasm-crypto": "^1.2.1",
+    "@polkadot/api": "1.26.1",
+    "@polkadot/api-contract": "1.26.1",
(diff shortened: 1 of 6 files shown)


Lezek123:
Since the `yarn.lock` fix wasn't enough, because the issue was still present during `npm pack` CI check, I changed `@polkadot` dependencies to exact versions to prevent incompatible versions like `1.28.1` from beeing installed.

mnaamani:
FYI: I suggested a better name for the renamed AccountInfo https://github.com/Joystream/joystream/pull/1148#discussion_r469028587

Lezek123:
Added types required by the node after https://github.com/Joystream/joystream/pull/1126/commits/fb9845beb7773e2282057572488259bf32487218
```
  Address: 'AccountId',
  LookupSource: 'AccountId',
```
Confirmed that this solves the issue with extrinsics discussed in https://github.com/Joystream/joystream/pull/1126

Lezek123:
Changed the name from `StorageProviderAccountInfo` to `ServiceProviderRecord` because of expected rename in the runtime.

mnaamani (approved):

PR #1155, opened 2020-08-12
Add api examples to monorepo

mnaamani:
Basic examples on using the polkadot js api and joystream types, ported from https://github.com/Joystream/joystream-api-examples


shamil-gadelshin:
Could we move the `api-examples` folder to the `tests`? The root directory space is a scarce resource and it is crowded already.

mnaamani:
> Could we move the `api-examples` folder to the `tests`? The root directory space is a scarce resource and it is crowded already.

Yes it is getting a bit crowded. I think `utils/` might be a better location. Will update

utils/api-examples/src/tohex.ts
```diff
@@ -0,0 +1,34 @@
+import { ApiPromise, WsProvider } from '@polkadot/api'
+import { CuratorApplicationId } from '@joystream/types/content-working-group'
+import { BTreeSet } from '@polkadot/types'
+import { types } from '@joystream/types'
+
+async function main() {
```
mnaamani, line 5: This script only needs to create some types and doesn't need to read state from chain, and therefore shouldn't require creating an api instance or connecting to a node. I this type of scenario, what is the proper way to create a type registry and register our types so we can construct them safely.

cc: @Lezek123

Lezek123: I added a function to do this in https://github.com/Joystream/joystream/pull/1176/files#diff-9e9b5c1632859d4c7bee5d17dd9f7c5c, it's not well-tested yet, but I think it should work fine. I can move this to another PR if needed.

utils/api-examples/scripts/list-data-directory.js
```diff
@@ -0,0 +1,33 @@
...
+  const ids = await api.query.dataDirectory.knownContentIds()
+
+  await Promise.all(
+    ids.map(async (id) => {
+      let obj = await api.query.dataDirectory.dataObjectByContentId(id)
+      if (obj.isNone) {
+        return
+      }
+      obj = obj.unwrap()
+      console.log(`contentId: ${new joy.media.ContentId(id).encode()}, ipfs: ${obj.ipfs_content_id}`)
```
Lezek123: Should be `api.createType('ContentId', id).encode()`
I'm not sure if the `joy` global is very useful at this point, since we should be able to create any type with `api.createType()`.

utils/api-examples/src/get-code.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  const api = await ApiPromise.create({ provider, types })
+
+  const currentBlockHash = await api.rpc.chain.getBlockHash(1)
+
+  console.log('getting code as of block hash', currentBlockHash.toString())
+
+  const substrateWasm = await api.query.substrate.code.at(currentBlockHash)
+
+  console.log(substrateWasm.toHex())
```
Lezek123, line 15: This gives me just `0x` as output (but also happens in Pioneer via `Extrinsics` tab)

mnaamani: Yes I was surprised by this as well. Does this have anything to do with the limit on the size of Vec you identified?

utils/api-examples/src/tohex.ts
mnaamani: Yes I spotted this also in pioneer react-api: 
https://github.com/Joystream/joystream/blob/iznik/pioneer/packages/react-api/src/typeRegistry.ts
https://github.com/Joystream/joystream/blob/iznik/pioneer/packages/react-api/src/Api.tsx#L179

So it looks "safe" and correct way.

utils/api-examples/scripts/list-data-directory.js
mnaamani: will fix :+1:

utils/api-examples/src/get-code.ts
Lezek123: While working on https://github.com/Joystream/joystream/pull/1177 I noticed that in the `augment-api.ts` auto-generated by `@polkadot/typegen` based on runtime metadata there is no `substrate` module at all.

So i did a little digging and found out that:
- This module is not part of the metadata retrieved via `api.rpc.state.getMetadata()` (even when I get it from our current testnet node)
- The same query also outputs `0x` on current Kusama
- It doesn't throw any errors (which happens in case of the `Vec` size limit issue that I ran into)

utils/api-examples/src/tohex.ts
```diff
@@ -1,19 +1,17 @@
...
-  // Is it not possible to create the registry without actually connecting to a node?
-  const set = new BTreeSet<CuratorApplicationId>(api.registry, CuratorApplicationId, [])
+  const registry = new TypeRegistry()
+  registry.register(types)
+
+  const set = new BTreeSet<CuratorApplicationId>(registry, CuratorApplicationId, [])
 
   wgId.forEach((id) => {
-    set.add(new CuratorApplicationId(api.registry, id))
+    set.add(createType(registry, 'CuratorApplicationId', id))
```
Lezek123, line 14: I found out we can actually use `registry.createType()` instead, which is probably a better approach (I'm going to change it in `@joystream/types` too)

Lezek123 (approved):

utils/api-examples/src/tohex.ts
mnaamani: Ah yes that is neater, and one less import :)

PR #1158, opened 2020-08-13
Expose missing types in registry for api.createType

Lezek123:
Exposed some missing types in the registy. They are not required for the api to work per se, but allow the script that creates `augment-types.ts` to be aware of them, which then allows us to use them with `api.createType`, ie.:
`api.createType('SlashingTerms', /* ... */)`

Exposing all types in the registry has additional benefit, because it makes `yarn workspace @joystream/types print:typedef` more effective in creating a typedef that is already almost fully compatible with `@polkadot/typegen` tool (there are some minor problems with `versioned-store` types, which are a spread over different files and seem to have some duplicates, but this can be probably resolved pretty quickly once we decide to migrate to `@polkadot/typegen` (or similar) solution)


gleb-urvanov (approved):
Seems good to me

mnaamani:
From what I understand the the types being added to the registry are not directly part of the runtime public api, and polkadot-js doesn't complain about them missing when connecting to the node. This has always been the case of course.

I clearly see the benefit of adding them to the registry as you described. Does it also benefit us when using the extrinsics app when filling in some struct fields that didn't previously have their type registered?

Have we considered the downside of registering these "non-public" runtime types?

mnaamani:
Need a minor merge conflict resolution

Lezek123:
> I clearly see the benefit of adding them to the registry as you described.

The additional benefit is described here https://github.com/Joystream/joystream/issues/1165, as registering those types will be required to take full advantage of `@polkadot/typegen`.

> Does it also benefit us when using the extrinsics app when filling in some struct fields that didn't previously have their type registered?

I can imagine we will get cleaner struct names (ie. `StakingPolicy` instead of json describing the struct), but other than that I'm not sure.

> Have we considered the downside of registering these "non-public" runtime types?

I think it is recommended to register them, that's what `@polkadot/typegen` expects and that's also the practice described in the api documentation. We could get away without doing it, since we use custom classes for the types and can pass some of them to the class instead of directly to the registry.

The obvious downside seems to be more risk for namespace clashes (since more types are exposed), but with the current tooling we can detect those clashes immediately (there are none currently) and in case of the types not required for the api to work we can just rename them, so it's not really an issue.

PR #1159, opened 2020-08-13
add placeholders for videos and channels loading

### LINKED ISSUE #1111: Implement placeholders for loading assets
Currently, Atlas does a poor job when loading assets like videos into the view. For example, images downloads are handled by the browser which makes the UX not so great with video poster loading one by one. Ideally, we should show some placeholder until all the data is ready to be displayed.

kdembler:
### Based on #1147, review that one first!

This PR partly resolves #1111 - it adds a basic placeholder for `VideoPreview` and `ChannelPreview` and uses those while data fetched via GraphQL is in a loading state. However, some further work will still be required to enable displaying the placeholders while the poster/channel avatars are being fetched. This should be a good start though.

It also does some small refactor of the `Carousel` component to improve readability and allow hiding controls 

Screenshot 2020-08-13 at 4 22 14 PM


fcescob (approved):

fcescob (approved):

bedeho (approved):
LGTM.

PR #1160, opened 2020-08-14
Pioneer: joy-election upgrade

Lezek123:
**Implements:**
-  https://github.com/Joystream/joystream/issues/1043

**Changes:**
- Continued `joy-utils` reactivation and upgrading - added transport provider for all packages that use the api (currently just `members` transport)
- Reactivated `joy-election` and did all the necessary adjustments.
- Fixed some UI issues with actions beeing displayed even though not available in given stage (ie. Applying form, Voting form, _Reveal Vote_ buttons)
- Added a few more adjustments to `polkadot-js/apps` core packages (ie. allowing council election stage as sidebar tab subtitle, just like before)
- Fixed linter issues

**Draft:**
Based on: https://github.com/Joystream/joystream/pull/1156

Commits of intrest:
- Almost all of the work: https://github.com/Joystream/joystream/commit/5d4a1949599cb2f04f2b8951c7aff10d0fc39a37
- Linter autofix: https://github.com/Joystream/joystream/commit/ea98e5dd637fd54e4b096e8e1d8554a50ea287b1
- Manual linter fixes (+ usePromise refactorization): https://github.com/Joystream/joystream/commit/2e655f7f8f3dca6500145641487e12c40e8a4c7b

**Testing:**
1. Run dev chain based on https://github.com/Joystream/joystream/tree/substrate_version_upgrade
1. `yarn workspace pioneer start` or use deploy preview
1. Change endpoint to local node (ie. by clicking the Joystream logo in the top-left corner)

Issue #1045, opened 2020-07-23
Pioneer upgrade: Upgrading and re-enabling joy-proposals


Lezek123:
`joy-proposals` seems to be a good candidate for a third package to re-enable. This is mostly because that's the package I'm the most familiar with, so I should be able to very easily spot any issues that could also affect other packages I'm going to re-enable later.

This issue will also cover adjusting the transports and a few related react components that live inside `joy-utils` (those were created with reusability in mind, but are currently only used inside `joy-proposals`).

Since this is a relatively big package that should be affected by a broad range of changes in `@polkadot/api`, `@joystream/types` and the runtime, working on it should allow me to better understand what kind of changes are going to be needed in other packages and what's the best way to address them.

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1195

PR #1176, opened 2020-08-18
Pioneer: joy-proposals upgrade

Lezek123:
**Implements:**
-  https://github.com/Joystream/joystream/issues/1045

**Changes:**
- Reactivating more of `joy-utils` - all transports inside `joy-utils` should now be compatible with the new runtime and api
- Reactivated `joy-proposals`
- Refactored `joy-proposals` to be typesafe (new linter rules are very strict when it comes to the use of `any` type, but this is good, since we can usually avoid that now that `@joystream/types` have been refactored)
- Adjusted some components like our custom `TxButton` to allow them to be used in more customizable way (introduced `SemanticTxButton` and `DefaultTxButton`) since they lost some of the features due to changes in `@polkadot/react-components`
- Fixed linter issues

**Draft:**
Based on: https://github.com/Joystream/joystream/pull/1160


Lezek123:
Covered by https://github.com/Joystream/joystream/pull/1195

PR #1185, opened 2020-08-24
connect video view to the data store

### LINKED ISSUE #1140: Connect video player view to a data source
At the moment, the video playback view always displays a mocked video. Instead, we should hook it up to the data store and enable id-based routing that will indicate which video to display

kdembler:
### Based on #1147, review that one first!

This PR resolves #1140 by connecting video view to use real data.


fcescob (approved):
Really like the work on the types! Looks like an amazing DX.

bedeho (approved):
LGTM.

PR #1179, opened 2020-08-21
Pioneer: joy-roles upgrade

Lezek123:
**Implements:**
-  https://github.com/Joystream/joystream/issues/1046

**Changes:**
- Moved some more code into the new `joy-utils`
- Reactivated `joy-roles`
- All the necessary adjustments (creating types, replaced linked maps usages in the transports etc.)
- Did some refactorization of the components structure inside `joy-utils` to avoid unnecessary re-renders that could lead to loss of data in the forms, problems when resizing the page and generally poor UX and optimalization. The main issue was that custom "render functions" were used in a few places instead of standard React components. When those functions were called after some state update - everything that they previously rendered was getting re-rendered from scratch and all the previous state in child components was lost. 
- Fixed lots of linter issues (mostly regarding the abuse of `any` type)
- Some minor adjustments to linter rules
- Some minor tooling adjustments (made `jest` available again which allows running `yarn workspace @polkadot/joy-roles test`, added `generate:json-schemas` scripting in `@joystream/types` to allow updating `hiring/schemas/role.schema.json`, configured webapck to handle `.scss` files, added back  `tsconfig-paths-webpack-plugin` to allow running storybook in the future)

I ignored the storybook files that already had `ts-nocheck`, the rest of the files I tried to upgrade as the cost wasn't very high. That means that a few of the stories will probably work and the rest can be updated later if needed. Currently the storybook fails because not all packages are upgraded yet (ie. `joy-media`), but I'll try to make it usable at the end of upgrading Pioneer.

**Draft:**
Based on: https://github.com/Joystream/joystream/pull/1176 with https://github.com/Joystream/joystream/pull/1177 merged (for a few additional types exposed in the registry)

Issue #1046, opened 2020-07-23
Pioneer upgrade: Upgrading and re-enabling joy-roles


Lezek123:
At this point `joy-proposals` should be already enabled which should make testing hiring scenarios and spotting any potential issues much easier. My familiarity with `joy-roles` package and the experience gained from upgrading `joy-proposals` (step 6) should also help in making the upgrade of this package go relatively smoothly at this point.

I'm predicting that this package is going to go through one of the biggest reworks (along with `joy-media`) w.r.t. the planned upgrades, as it intensely uses linked maps, `new SomeType()` constructs and other `@joystream/types` and `@polkadot/api` related "features" that may not be preserved.

It may turn out that a huge part of the required adjustments is going to revolve around the mocks and storybook, in which case I think a resonable solution would be to throw those away completely. Most of the `joy-roles` storybook is already outdated anyway and does no longer work, so keeping it and trying to implement all the required adjustments doesn't really seem justified (we can consider rewriting it when needed in the future. This should also be way easier once we have a complete, normalized mock-transport, possibly implemented in `joy-utils`)

_**While working on this issue and https://github.com/Joystream/joystream/issues/1045 I'll try to keep in mind that at some point (probably not yet due to complications related to content-working-group inconsistency with other working groups), both `joy-proposals` and `joy-roles` transports should be merged inside `joy-utils`. It may make sense to try to normalize them to the point that makes this transition go as smoothly as possible in the future.**_

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1195

PR #1188, opened 2020-08-26
Pioneer: joy-media upgrade

Lezek123:
**Implements:**
-  https://github.com/Joystream/joystream/issues/1046

**Changes:**
- Reactivated `joy-media`
- Added back some missing utils to the new `joy-utils` (ie. `withCurationActor` which I also had to upgrade to work with the new runtime)
- All the necessary adjustments in `joy-media` (creating types, linked maps usages in the transports etc.)
- **FIX:** problems with `instanceof` due to mismatching versions of `bn.js` in the monorepo (fixed by adding to `resolutions`)
- **FIX:** problems with `instanceof` due to `api.createType('ChannelId', 1)` (and some other ids) returning `u64` (which caused `instanceof ChannelId` to return `false` in those cases)
- Minor UI improvements (ie. displaying a loader while upload video extrinsic is waiting for the confirmation)
- Fixed linter issues according to the new rules

**Draft:**
Based on: https://github.com/Joystream/joystream/pull/1179


Lezek123:
Covered by https://github.com/Joystream/joystream/pull/1195

PR #1189, opened 2020-08-26
Storage node conversion to the runtime version based on Substrate 2.0.0-rc4

shamil-gadelshin:
Main changes:
- updated runtime types: "@joystream/types": "^0.13.0",
- updated API "@polkadot/api": "^1.26.1",
- transaction status  `IsFinalized` was changed to `asCompleted`

### DIFF
--- a/storage-node/packages/cli/src/commands/dev.ts
+++ b/storage-node/packages/cli/src/commands/dev.ts
@@ -188,7 +188,7 @@ const init = async (api: RuntimeApi): Promise<any> => {
   debug('Making Alice the storage Lead.')
   const leadOpeningId = await api.workers.devAddStorageLeadOpening()
   const leadApplicationId = await api.workers.devApplyOnOpening(leadOpeningId, aliceMemberId, alice, alice)
-  api.workers.devBeginLeadOpeningReview(leadOpeningId)
+  await api.workers.devBeginLeadOpeningReview(leadOpeningId)
   await api.workers.devFillLeadOpening(leadOpeningId, leadApplicationId)
 
   const leadAccount = await api.workers.getLeadRoleAccount()
--- a/storage-node/packages/cli/src/commands/dev.ts
+++ b/storage-node/packages/cli/src/commands/dev.ts
@@ -188,7 +188,7 @@ const init = async (api: RuntimeApi): Promise<any> => {
   debug('Making Alice the storage Lead.')
   const leadOpeningId = await api.workers.devAddStorageLeadOpening()
   const leadApplicationId = await api.workers.devApplyOnOpening(leadOpeningId, aliceMemberId, alice, alice)
-  await api.workers.devBeginLeadOpeningReview(leadOpeningId)
+  api.workers.devBeginLeadOpeningReview(leadOpeningId)
   await api.workers.devFillLeadOpening(leadOpeningId, leadApplicationId)
 
   const leadAccount = await api.workers.getLeadRoleAccount()
--- a/storage-node/packages/runtime-api/balances.js
+++ b/storage-node/packages/runtime-api/balances.js
@@ -51,7 +51,7 @@ class BalancesApi {
    */
   async freeBalance(accountId) {
     const decoded = this.base.identities.keyring.decodeAddress(accountId, true)
-    return this.base.api.query.balances.freeBalance(decoded)
+    return (await this.base.api.derive.balances.all(decoded)).freeBalance
   }
 
   /*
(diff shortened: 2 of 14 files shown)


storage-node/packages/runtime-api/balances.js
```diff
@@ -51,7 +51,7 @@ class BalancesApi {
    */
   async freeBalance(accountId) {
     const decoded = this.base.identities.keyring.decodeAddress(accountId, true)
-    return this.base.api.query.balances.freeBalance(decoded)
+    return (await this.base.api.derive.balances.all(decoded)).freeBalance
```
Lezek123: In this context I think `(await this.base.api.derive.balances.all(decoded)).availableBalance` may actually be more suitable, because `freeBalance = avaialble + locked` and if the balance is locked it seems like it cannot be transferred or used to pay the fee.

storage-node/packages/cli/src/commands/dev.ts
```diff
@@ -188,7 +188,7 @@ const init = async (api: RuntimeApi): Promise<any> => {
   debug('Making Alice the storage Lead.')
   const leadOpeningId = await api.workers.devAddStorageLeadOpening()
   const leadApplicationId = await api.workers.devApplyOnOpening(leadOpeningId, aliceMemberId, alice, alice)
-  api.workers.devBeginLeadOpeningReview(leadOpeningId)
+  await api.workers.devBeginLeadOpeningReview(leadOpeningId)
```
mnaamani: this `await` was left out so the two transactions can be finalized in the same block. 

It works because the second call doesn't need to get any values from event(s) emitted by the first transaction.
Of course the second one will fail if the first fails.

So its just a speed optimization :)

storage-node/packages/runtime-api/balances.js
mnaamani: @Lezek123 is correct. The way we are using it we were not accounting for locked balance. It is only directly called by `hasMinimumBalanceOf()` which is the call that is really interested in a minimum transferable balance. We can fix this in a later PR.

storage-node/packages/runtime-api/workers.js
```diff
@@ -121,31 +121,14 @@ class WorkersApi {
...
-        // const worker = (await this.base.api.query.storageWorkingGroup.workerById(id))[0]
-        value = value.unwrap()
-        // construct the Worker type from raw data
-        // const worker = createType('WorkerOf', value)
-        // const worker = new Worker(value)
-        ids.push(id)
-        providers[id] = new Worker(value)
-      }
-    }
+    const entries = (await this.base.api.query.storageWorkingGroup.workerById.entries());
```
mnaamani: love the new api

mnaamani:
The changes look good.

To build I had to fix call to `ContentId.generate()` to  `ContentId.generate(this.api.api.registry),` in:

```
MokoBook:storage-node mokhtar$ git diff packages/cli/src/commands/upload.ts 
diff --git a/storage-node/packages/cli/src/commands/upload.ts b/storage-node/packages/cli/src/commands/upload.ts
index 7a1ff8cfe..f0a9f7c34 100644
--- a/storage-node/packages/cli/src/commands/upload.ts
+++ b/storage-node/packages/cli/src/commands/upload.ts
@@ -96,7 +96,7 @@ export class UploadCommand extends BaseCommand {
     return {
       accountId,
       ipfsCid: await this.computeIpfsHash(),
-      contentId: ContentId.generate(),
+      contentId: ContentId.generate(this.api.api.registry),
       fileSize: new BN(this.getFileSize()),
       dataObjectTypeId,
       memberId,
```

mnaamani:
I'm seeing 9 linter errors that need to be fixed, we can fix in a separate PR?

Running `yarn storage-cli dev-init` works as expected after the build fix.

shamil-gadelshin:
That is strange - I had successfully uploaded the video file during the testing.

storage-node/packages/cli/src/commands/dev.ts
shamil-gadelshin: Ok. I'm rolling it back.

storage-node/packages/runtime-api/balances.js
shamil-gadelshin: Leszek's solution is only a single line. I'm including the fix in the current PR.

mnaamani:
> That is strange - I had successfully uploaded the video file during the testing.

Is it possible that the packages wasn't rebuilt and used some pre-compiled code you had?

I noticed storage-node is the only project we have where a `tsconfig.tsbuildinfo` is built. And sometimes its presence is preventing some code from being recompiled if the source doesn't change.

mnaamani:
Did some additional testing. LGTM.

One bug I just realized in the `async getAllProviders()` it returns all the worker ids (including the lead worker id). I opened an issue https://github.com/Joystream/joystream/issues/1211

Issue #949, opened 2020-07-15
Create DB schema via migrations


dzhelezov:
Currently, Query Node CLI assumes that there's an active DB connection during the codegen, and creates the entity tables and the indexer event table along with the sources (while one would expect only the latter). This has the following significant drawbacks

- Tight coupling between the code generation and DB schema creation: the DB connection should available during the initial setup
- Hard to deploy to a cloud/remote server ("production")
- Non-repeatable builds (once `save_entity_events` table is created, subsequent indexer builds fail)
- DB schema migrations are spread between the graphql server and indexer: graphql creates entities, indexer creates indexing tables. There are at least three places where the DB schema is updated
  -- DB synchronize (creates entity tables from the source classes; not recommended for production)
  -- Ad-hoc migrations from the input GraphQL schema (e.g. auxiliary views from Interface types or FTS queries)
  -- Events table created by the index-builder

A more flexible approach would be to generate TypeORM migrations and keep them in a single place. This will allow setting up the DB schema in one shot and rollback if an error occurs. 

In particular, we should avoid running `db:synchronize` altogether and replace it with auto-generated migrations source from the entity files.

PR #1191, opened 2020-08-26
Separate block indexing and processing the mappings

dzhelezov:
This PR is a major upgrade in the Hydra indexer architecture. The indexer is now split into two concurrent long-running promises, `MappingsProcessor` and `IndexBuilder`. Index builder is responsible for saving all the substrate events and extrinsics into the database, while `MappingsProcessor` queries only the events that match the handlers (typically very few).  

Implementation details:
- `MappingsProcessor` reads the indexed blocks from the database and filters out only the events matching the handlers. The event data is wrapped into `SubstrateEvent` interface. It has been moved into a separate `interfaces` module together with `SubstrateExtrinsic`

- `IndexBuilder` internally uses a pool of workers concurrently fetching the blocks. It also keeps updating the indexer "head" which is defined as the maximal height H such that all the blocks from 0 to H are indexed. Since the blocks are indexed concurrently some non-trivial logic is required to keep it up-to-date

- Events and Extrinsics are mapped by TypeORM entities in the `entities` module

- `MappingsProcessor` periodically checks the `IndexBuilder` head and queries the database in batches of 200 blocks (configurable through env variables). Perhaps a more efficient approach would be to turn `IndexBuilder` into an `EventEmitter` and listen to the indexed blocks containing the events of interest. However, due to the async nature of the indexer some extra care is needed to ensure the events are processed in the right order, so this improvement is left for future releases

In order to set up the required database table, a manual migration run is required. 
Run
```bash
yarn typeorm migration:run 
```
from the root folder of the indexer. This is a temporary solution until https://github.com/Joystream/joystream/issues/949 is implemented


metmirr (approved):
LGTM! I did my best for this quite large PR.

bedeho (approved):

PR #1193, opened 2020-08-27
Qnode separate processor indexer

dzhelezov:
This PR depends on https://github.com/Joystream/joystream/pull/1191

The main difference is how `processor/MappingsProcessor` handles the events. This PR removes all the indexer dependencies and instead relies on the database polling. Thus, `MappingsProcessor` and `IndexBuilder` are fully decoupled.

Further, `MappingProcessor` has been completely removed from `QueryNode`. Only the indexer is now managed by QueryNode and the mappings will _NOT_ be processed by default. Instead, in order to run the mappings (separately from the indexer), a separate entry point `QueryNodeManager.process()` should be called directly from the hydra-indexer tool.


metmirr (approved):
LGTM!

bedeho (approved):

PR #1156, opened 2020-08-12
Pioneer: Apps adjustments + joy-members reactivation

Lezek123:
**Implements:**
- https://github.com/Joystream/joystream/issues/1131
- https://github.com/Joystream/joystream/issues/1132

**Changes:**
- Adjusted styles and functionality of the new `polkadot-js/apps` based on our old, modified apps (ie. added providers in `Apps`, added our types to api, added `TopBar` with account dropdown-select, adjusted the routes/tabs). I tried to limit the amout of those changes to the absolute minimum.
- Partially re-activated `joy-utils`. Moved all required functionality to new `joy-utils` directory, did some refactoring and put the files inside a new, more clean directory structure.
- Reactivated `joy-members` and `joy-pages` and did all the necessary adjustments.
- Disabled some linter rules that could cause too much chaos during the upgrade 
- Fixed remaining linter issues

**Draft:**
Based on: https://github.com/Joystream/joystream/pull/1101
Merged https://github.com/Joystream/joystream/pull/1153 for new node compatibility

Commits of intrest:
- Almost all of the work: https://github.com/Joystream/joystream/commit/ea711c9e7731237bc987aa2e0f493d41488fb907
- Linter rules adjustment + autofix: https://github.com/Joystream/joystream/commit/15e1c69e5764f2b03f0a4eec6ce92560034f95ce
- Manual linter fixes: https://github.com/Joystream/joystream/commit/83d22c4016c293500857ce1093f65d0d31fcfcac
- Icons/title changes: https://github.com/Joystream/joystream/commit/8a8f85aee35153a905856dcfb4ab4b8bd4e97f11

**Testing:**
1. Run dev chain based on https://github.com/Joystream/joystream/pull/1126/commits/fb9845beb7773e2282057572488259bf32487218 (currently should already be part of https://github.com/Joystream/joystream/tree/substrate_version_upgrade)
1. `yarn workspace pioneer start` or use deploy preview at https://deploy-preview-1156--joystream-pioneer.netlify.app/
1. Change endpoint to local node (ie. by clicking the Joystream logo in the top-left corner)

Issue #1049, opened 2020-07-23
Pioneer upgrade: Upgrading and re-enabling joy-forum


Lezek123:
As far as upgrading Pioneer packages to work with the new runtime and `polkadot-js` goes, the forum package can probably be re-enabled at any point after `joy-members`. It's a rather small and independent package and both shouldn't affect and be affected by the re-activation of other packages. I suspect the workload to be comparable to this needed to re-activate `joy-members` or in the worst case `joy-election`.

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1195

Issue #1048, opened 2020-07-23
Pioneer upgrade: Upgrading and re-enabling joy-media


Lezek123:
`joy-media` is among the biggest `joy-` packages and looks like it may be heavily affected by the coming upgrades. It requires more setup that other packages in order to be tested (which has been fortunately very simplified before the Nicaea release) and is partially dependent on other external tools (like `versioned-store-js`). It also relies on a particular mechanism of encoding and decoding entities to/from `Codec` types described in https://github.com/Joystream/joystream/issues/1047, which will be affteted by the upgrade of `@joystream/types` that will cause the need for some additional adjustments.

Lezek123:
Updated the estimate as this issue now includes https://github.com/Joystream/joystream/issues/1047

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1195

Issue #1043, opened 2020-07-23
Pioneer upgrade: Upgrading and re-enabling joy-election


Lezek123:
Seems like upgrading and re-enabling `joy-election` may be a good choice after `joy-members`. This is still a relatively small and very useful app that may help uncover some additional issues and better prepare for enabling more complex apps (`joy-media`, `joy-proposals`, `joy-roles` etc.)

Lezek123:
Closed by https://github.com/Joystream/joystream/pull/1195

PR #1195, opened 2020-08-28
Pioneer upgrade - final adjustments

Lezek123:
This PR contains all Pioneer upgrade changes after the `joy-members` package upgrade (https://github.com/Joystream/joystream/pull/1156).

**Covered issues:**
- https://github.com/Joystream/joystream/issues/1043 (joy-election)
- https://github.com/Joystream/joystream/issues/1045 (joy-proposals)
- https://github.com/Joystream/joystream/issues/1046 (joy-roles)
- https://github.com/Joystream/joystream/issues/1048 (joy-media)
- https://github.com/Joystream/joystream/issues/1049 (joy-forum)

### `joy-election` (from 5d4a194 to 9fadcaf)
- Continued `joy-utils` reactivation and upgrading - added transport provider for all packages that use the api (currently just `members` transport)
- Reactivated `joy-election` and did all the necessary adjustments.
- Fixed some UI issues with actions beeing displayed even though not available in given stage (ie. Applying form, Voting form, _Reveal Vote_ buttons)
- Added a few more adjustments to `polkadot-js/apps` core packages (ie. allowing council election stage as sidebar tab subtitle, just like before)
- Fixed linter issues accoring to the new rules

### `joy-proposals` (from f7d87da to e19eb61 + 673855f)
- Reactivating more of `joy-utils` - all transports inside `joy-utils` should now be compatible with the new runtime and api
- Reactivated `joy-proposals`
- Refactored `joy-proposals` to be typesafe (new linter rules are very strict when it comes to the use of `any` type, but this is good, since we can usually avoid that now that `@joystream/types` have been refactored)
- Adjusted some components like our custom `TxButton` to allow them to be used in more customizable way (introduced `SemanticTxButton` and `DefaultTxButton`) since they lost(?) some of the customizability due to changes in `@polkadot/react-components`
- Fixed linter issues accoring to the new rules

### `joy-roles` (from 17933d4 to a9e4448)
- Moved some more code into the new `joy-utils`
- Reactivated `joy-roles`
- All the necessary adjustments (creating types, replaced linked maps usages in the transports etc.)
- Did some refactorization of the components structure inside `joy-utils` to avoid unnecessary re-renders that could lead to loss of data in the forms, problems when resizing the page and generally poor UX and optimalization. The main issue was that custom "render functions" were used in a few places instead of standard React components. When those functions were called after some state update - everything that they previously rendered was getting re-rendered from scratch and all the previous state in child components was lost. 
- Fixed lots of linter issues (mostly regarding the abuse of `any` type)
- Some minor adjustments to linter rules
- Some minor tooling adjustments (made `jest` available again which allows running `yarn workspace @polkadot/joy-roles test`, added `generate:json-schemas` scripting in `@joystream/types` to allow updating `hiring/schemas/role.schema.json`, configured webapck to handle `.scss` files, added back  `tsconfig-paths-webpack-plugin` to allow running storybook in the future)

### `joy-media` (from 03bd123 to 83a225c)
- Reactivated `joy-media`
- Added back some missing utils to the new `joy-utils` (ie. `withCurationActor` which I also had to upgrade to work with the new runtime)
- All the necessary adjustments in `joy-media` (creating types, linked maps usages in the transports etc.)
- **FIX:** problems with `instanceof` due to mismatching versions of `bn.js` in the monorepo (fixed by adding to `resolutions`)
- **FIX:** problems with `instanceof` due to `api.createType('ChannelId', 1)` (and some other ids) returning `u64` (which caused `instanceof ChannelId` to return `false` in those cases)
- Minor UI improvements (ie. displaying a loader while upload video extrinsic is waiting for the confirmation)
- Fixed linter issues according to the new rules

### `joy-forum` (from ad0377e to 5ce833f)
- Reactivated `joy-forum` + implemented the necessary adjustments
- Added last missing pieces to the new `joy-utils`
- Extended member preview components in `joy-utils` (ie. allowing to specify the size, displaying council badge etc.) to prevent imports between `joy-members` and `joy-forum`.
- Minor styling changes
- Linter fixes

### Remaining commits (after 5ce833f)
- Fixed storybook (not all of the components are working, which was already a problem before the upgrade, but now it's at least possible to succesfully run `yarn workspace pioneer storybook` again)
- Re-introduced a component to update the account memo (`MemoForm`) which was previously part of `app-accounts`
- Finalized the process of `joy-utils` migration
- Removed the no longer necessary direcotires (`old-apps`, `joy-utils-old`)
- Minor styling fixes
- Minor apps configuration fixes (default ui mode, removed unnecessary endpoints etc.)
- **FIX:** Fixed the problem with `RuntimePropsal` details by changing the type from `Vec<u8>` to `Bytes` (which works the same way, but doesn't have the `32768` length limit)


mnaamani:
There is a minor merge conflict in yarn.lock after merging https://github.com/Joystream/joystream/pull/1189

pioneer/packages/joy-media/src/transport.ts
```diff
@@ -217,26 +227,27 @@ export abstract class MediaTransport extends Transport {
...
     const idsOfCuratedCS = (await this.allCurationStatuses())
-      .filter(x =>
+      .filter((x) =>
         insensitiveEq(x.value, 'Under review') ||
         insensitiveEq(x.value, 'Removed')
-      ).map(x => x.id);
+      ).map((x) => x.id);
 
     const isPublicAndNotCurated = (video: VideoType) => {
       const isPublic = video.publicationStatus.id === idOfPublicPS;
```
mnaamani, line 241: I populated the content directory with some content and I'm seeing this error:

Screen Shot 2020-09-02 at 10 10 18 AM

mnaamani: For context, while exporting state from current chain, some items were dropped and their property values were dropped. This could explain this error for some particular videos: https://github.com/Joystream/joystream-api-examples/pull/12/files#diff-6d0009198c779316ce8a37be5c378e3cR118

mnaamani: Perhaps the media transport needs some improvement to determine when the property values don't exist and just exclude it from the array (similarly to where we did this check `entity.in_class_schema_indexes.toArray().length`)

pioneer/packages/apps-routing/src/joy-media.ts
```diff
@@ -0,0 +1,15 @@
+import Media from '@polkadot/joy-media/index';
+
+import { Route } from './types';
+
+export default function create (t: <T = string> (key: string, text: string, options: { ns: string }) => T): Route {
+  return {
+    Component: Media,
+    display: {
+      needsApi: ['query.storageWorkingGroup.workerById', 'query.dataObjectStorageRegistry.relationshipsByContentId']
```
mnaamani, line 9: media does need a few additional apis from other modules, should we be more comprehensive? I suppose this applies to other joy-* apps as well.

Lezek123:
Changed the implementation of `allChannelIds`, `allClassIds` and `allEntityIds` to use `.keys()` instead of returing all ids between first and last id, which should allow us to safely skip some entities during the export.

Lezek123:
Fixed `RuntimeUpgrade` and `Text` proposal validation to make use of exposed runtime consts instead of using `validationSchema.js` hardcoded values (in preparation for coming https://github.com/Joystream/joystream/issues/1216 fix)

pioneer/packages/joy-media/src/Upload.tsx
```diff
@@ -205,7 +222,7 @@ class Upload extends React.PureComponent<Props, State> {
       });
     } else {
       this.setState({ file, computingHash: true });
-      this.startComputingHash();
+      void this.startComputingHash();
```
mnaamani, line 225: I don't understand the presence of this `void` here? Is it to drop or ignore the return value of the call ?

mnaamani:
Just some feedback so far, as I'm still going through the PR:

* Suggestion to change react-api/src/typeRegistry.ts to use registry created in @joystream/types

```
import { mockRegistry } from '@joystream/types';
export default mockRegistry;
```
Reason: In some components we grab the `{createMock}` from joystream/types and in other places we use `api.createType`
So this will at least ensure the same instance of the registry is used in both those instances.

* Suggested renaming in @joystream/types/src/index

rename `createMock` to `createType`, and `mockRegistry` to `registry`

The term "Mock" in the name is misleading but the approach is perfectly reasonable, and that is in fact how react-api initializes the Api and types registry.

* Drop the `EditForumSudo` component in the forum app (which we don’t render anyway)
Reason: It was never implemented correctly. Only the system sudo account via the sudo module can set the new forum sudo. This can be easily done in the sudo app.

I tested media app again with dropped entities work nicely and I was able to navigate and play content.
Uploaded a new media file to my channel, but after media upload, the form to fill details has a field `Explicit`.

* Explicit field should be treated as boolean but displays as a string input field. Was expecting a drop-down with true, false options.

Screen Shot 2020-09-02 at 6 20 27 PM

pioneer/packages/joy-media/src/Upload.tsx
Lezek123: This is a way to mark a promise which we intentionally don't await. It's required by the new `no-floating-promises` rule (see: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md#ignorevoid)

I think it's a good rule, because it prevents omitting `await` by accident. Marking with `void` also gives a a nice hint that the function/method we execute is async.

pioneer/packages/joy-media/src/common/MediaPlayerWithResolver.tsx
```diff
@@ -24,56 +24,46 @@ function newCancelSource (): CancelTokenSource {
...
-    readyProviders = _.uniqBy(readyProviders, provider => provider.toString());
+    readyProviders = _.uniqBy(readyProviders, (provider) => provider.toString());
 
     if (!readyProviders.length) {
       setError(new Error('No Storage Providers found storing this content'));
+
       return;
     }
 
+    const activeProviders = (await transport.workingGroups.allWorkers('Storage')).map(([id]) => id);
```
mnaamani, line 60: In this implementation we need to filter out the Storage Group Lead id, so we get only the storage providers.

pioneer/packages/joy-roles/src/transport.substrate.ts
```diff
@@ -256,27 +249,19 @@ export class Transport extends TransportBase implements ITransport {
...
-    const groupOpenings = new MultipleLinkedMapEntry<GroupOpeningId, GroupOpening>(
-      OpeningId,
-      workingGroupsApiMapping[group].openingType,
-      await this.cachedApiMethodByGroup(group, 'openingById')()
+    const groupOpenings = await this.entriesByIds<GroupOpeningId, GroupOpening>(
+      this.apiMethodByGroup(group, 'openingById')
     );
 
-    for (const groupOpening of groupOpenings.linked_values) {
+    for (const [/* id */, groupOpening] of groupOpenings) {
```
mnaamani, line 256: If we don't need the key (id), which we get from value.hiring_opening_id.id, perhaps better to get `.values()` from map only instead of `.entries()`. Its a minor point, but it would give a cleaner Destructuring assignment

pioneer/packages/joy-roles/src/transport.substrate.ts
```diff
@@ -510,9 +458,9 @@ export class Transport extends TransportBase implements ITransport {
         maxNumberOfApplications: opening.max_applicants,
         requiredApplicationStake: stakes.application,
         requiredRoleStake: stakes.role,
-        defactoMinimumStake: new u128(0)
+        defactoMinimumStake: this.api.createType('u128', 0)
```
mnaamani: Balance instead of u128 ?

pioneer/packages/joy-media/src/common/MediaPlayerWithResolver.tsx
mnaamani: After raising this in chat, we can ignore this comment.

ref: https://github.com/Joystream/joystream/issues/1211

pioneer/packages/joy-roles/src/transport.substrate.ts
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/1195/commits/728bb4c90c228828a77325f387efa9787f720711

Lezek123:
> Suggestion to change react-api/src/typeRegistry.ts to use registry created in @joystream/types

Done in https://github.com/Joystream/joystream/pull/1195/commits/7edff5e06635ab2f3f0e80469474e4350e6ea719

> rename createMock to createType, and mockRegistry to registry

Done in https://github.com/Joystream/joystream/pull/1195/commits/1d7d84dda4f9ba1038b8e15e29dcd0b01081aefa

pioneer/packages/joy-roles/src/transport.substrate.ts
Lezek123: I dind't notice `.values()` method to be available. We could write an implementation similar to `entiresByIds`, but it would actually do the same thing (execute `entiresByIds` and then throw away ids), since we still need ids to sort the entries (by default they are returned in unpredictable order)

Lezek123:
> Explicit field should be treated as boolean but displays as a string input field. Was expecting a drop-down with true, false options.

This was an issue before the upgrade too, but fixed it now in https://github.com/Joystream/joystream/pull/1195/commits/8b2b4840e69016737c9839f35e925d579732db9c

> Drop the EditForumSudo component in the forum app (which we don’t render anyway)

Done in https://github.com/Joystream/joystream/pull/1195/commits/c8f2d1faa3c2160998fc5536235b513813bf835a

pioneer/packages/apps-routing/src/joy-media.ts
Lezek123: Those are not very strict even in the original `polkadot-js/apps` I think.
The `needsApi` itself is required as it affects how components are rendered by `apps/src/Content/index.tsx`, but including **all** the methods that are actually used by given module would be hard to maintain.
It makes sense only if there are multiple chains user can connect to, which have different runtime modules.

mnaamani (approved):

PR #1199, opened 2020-08-31
add infinite video grid component

kdembler:
### ~Based on #1159, review that one first!~

This PR adds an infinite scroll component that will load more videos as the user scrolls.

atlas-scroll

It will require some tweaking once we:
a) add responsiveness,
b) connect to a real data source

Places that will require a change are indicated by TODO comments


fcescob (commented):
This implementation works fine but I think we could make this more general with small effort, for example in the future  we might want to have an infinite list of channels or series. 
One possible approach to this might be:
- Make the component accept a query and call it with `useLazyQuery` (basically moving the logic inside the `Home` view here) or just use a more general function name instead of `loadVideos`
- Remove all specific references to videos and make `ITEMS_PER_ROW` set by a prop 

What do you think?

kdembler:
Sure, the reusability of this for other assets crossed my mind as well and I agree it shouldn't be much effort. Although I would say it may not be worth doing it at this moment. We probably won't be using this for anything other than videos for the initial release so let's not do a premature abstraction. We can do that once we have a use case and then make sure the abstraction is good enough. Would you agree?

fcescob:
Sounds good to me. Approved!

fcescob (approved):

PR #1204, opened 2020-08-31
Pioneer: Video sorting fix

### LINKED ISSUE #1203: Pioneer: Nondeterministic video sorting
It seems like due to invalid use of `Array.sort()` in `joy-media` module, the videos that appear in the _**Latest videos**_ section at https://testnet.joystream.org/#/media are not always in fact the latest videos.

On Firefox those videos seem to be sorted from oldest to newest, which means that currently there is always the same set of videos beeing displayed in the  _**Latest videos**_ section on that browser (even though there is already a lot of newer content uploaded to the platform).

Lezek123:
Fixes https://github.com/Joystream/joystream/issues/1203

### DIFF
--- a/pioneer/packages/joy-media/src/transport.ts
+++ b/pioneer/packages/joy-media/src/transport.ts
@@ -158,7 +158,7 @@ export abstract class MediaTransport extends Transport {
   async videosByChannelId (channelId: ChannelId, limit?: number, additionalFilter?: (x: VideoType) => boolean): Promise<VideoType[]> {
     let videos = (await this.allVideos())
       .filter(x => channelId && channelId.eq(x.channelId) && (additionalFilter || (() => true))(x))
-      .sort(x => -1 * x.id);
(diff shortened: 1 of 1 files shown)


pioneer/packages/joy-media/src/transport.ts
```diff
@@ -158,7 +158,7 @@ export abstract class MediaTransport extends Transport {
   async videosByChannelId (channelId: ChannelId, limit?: number, additionalFilter?: (x: VideoType) => boolean): Promise<VideoType[]> {
     let videos = (await this.allVideos())
       .filter(x => channelId && channelId.eq(x.channelId) && (additionalFilter || (() => true))(x))
-      .sort(x => -1 * x.id);
+      .sort((a, b) => b.id - a.id);
```
mnaamani, line 161: Good catch, `x => -1 * x.id` doesn't seem to make any sense at all :)

mnaamani (approved):

PR #1212, opened 2020-09-02
Query node Joystream proposal system

metmirr:
This PR creates a schema for the Joystream proposal system.

### DIFF
--- a/query-node/joystream-query-node/schema.graphql
+++ b/query-node/joystream-query-node/schema.graphql
@@ -1,21 +1,22 @@
+"Stored information about a registered user"
 type Member @entity {
   memberId: BigInt!
 
-  # The unique handle chosen by member
+  "The unique handle chosen by member"
   handle: String @fulltext(query: "handles")
 
-  # A Url to member's Avatar image
+  "A Url to member's Avatar image"
   avatarUri: String
 
-  # Short text chosen by member to share information about themselves
+  "Short text chosen by member to share information about themselves"
   about: String
-
-  # Blocknumber when member was registered
+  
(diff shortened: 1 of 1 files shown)


metmirr:
@bedeho I need your input on some fields.

1. You mention algebraic types could be used for `Proposal.finalizedAtBlock` can you please give more detail about it?
2. Could you help me with these proposal types `TerminateWorkingGroupLeaderRole`, `FillWorkingGroupLeaderOpening`, and `AddWorkingGroupLeaderOpening`? I can't create proper entities for them.

bedeho:
1. I see that no derived relationships are used, could it not? for example couldn't the council derive what council members are in it based on the `CouncilMember.council` relationship? 
2. The about algebraic types in `Proposal` is not about `finalizedAtBlock`, it's actually about the structure of all fields below the comment
```
  finalizedAtBlock: BigInt
  status: ProposalState
  "Amount which was staked"
  stakeBalance: BigInt!
  abstentionVotes: BigInt!
  approvalVotes: BigInt!
  rejectionVotes: BigInt!
  slashVotes: BigInt!
```

So now that we do have algebraic types I think it makes sense to use them to model the proposal state properly.
3. Did you look at the extrinsics for creating these proposals, that is a good place to start. Compare with other proposals for where you already have entities defined.
4. I think the current way proposals are modeled is not ideal, because its possible to associate multiple specific proposals, like `TextProposal` and `RuntimeUpgradeProposal`, with the same underlying  `Proposal`. this is bad. I think perhaps the best approach is to make `Proposal` into an interface and implement it for each proposal. This will also allow us to query across proposal types.

PR #1209, opened 2020-09-01
'Process' and 'Index' target for the Hydra indexer entry point

dzhelezov:
This PR depends on https://github.com/Joystream/joystream/pull/1193
It incorporates the following changes:
- Refactoring of the indexer entry point template. 
- Introduce new commands: "index", "process", "migrate" for indexing, processing the mappings and performing the initial db schema migration (TBD)
- The lib is now published as `@dzlzv/hydra-indexer-lib` until migrated to `@joystream`. The templates have been updated accordingly

PR #1224, opened 2020-09-02
Refactoring and migrations target for Hydra indexer

dzhelezov:
This PR depends on https://github.com/Joystream/joystream/pull/1209 and should be merge after. It brings about the following refactorings and features:

- Actualized 'migrate' target for `indexer-cli`. It triggers the migrations auto-generated by typeorm;
- Simplified logic for events processing based on eventIDs. All events IDs are now fixed-length strings of the form `padZeros(16, <blockHeight>)-padZeroes(6, <index>)` so that the natural string order coincides with the order the events should be processed;
- All processed event references are saved to the `processed_event_log` table for better bookkeeping (only the last processed event was previously saved);
- Additional options are added to the MappingProcessor allowing multiple processors to be run simultaneously as long as the names are different;


metmirr (approved):
LGTM!

bedeho (approved):

PR #1229, opened 2020-09-03
Implement tokenomics page

DzhideX:
Couple of things:

- I saw (fairly late) that you use a completely different approach when taking data from the api, to keep on track, I've left it as is but am happy to change, if necessary!
- The graph code is very vanillaJS-esque, I hope that isn't much of a problem.
- I'm still fairly new to all of this so a thorough code review would be much appreciated!


bedeho:
Excited to see this one in production!

DzhideX:
Happy to have been able to contribute. Hope everyone likes it 👍

Lezek123:
Hello @DzhideX,

First of all - great work, I took a quick look at it yesterday and I'll be happy to make a review.

Couple of things to start with:

- We took the approach of naming our custom packages `joy-{package_name}` in order to differentiate between the packages that are part of the forked repository (`polkadot-js/apps`) and the ones that are completely Joystream-specific (like `Memberships`, `Council`, `Proposals` etc.)

- When you make a pull request, our Continuous Integration checks are automatically run on GitHub. You can see the results below. As you can notice, the `build` (`Pioneer / Ubuntu/MacOS Build`) and `lint` (`Pioneer / Ubuntu/MacOS Linting`) checks have failed. You can see the reason by clicking on the `Details` link on the right-hand side of the check (ie. the Ubuntu build resuts are here: https://github.com/Joystream/joystream/pull/1229/checks?check_run_id=1067928627). You can run those checks locally before making a PR:
```
yarn workspace pioneer build
yarn workspace pioneer lint-only-errors
```
Some cosmetic linter errors (like presence/lack of `;`, indentation etc.) can be fixed automatically with:
```
yarn workspace pioneer lint-autofix
```


I'll post a more thorough review of the code in the next few hours.

pioneer/packages/app-tokenomics/src/Overview/OverviewTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+          {item}
+          {help && <OverviewTableHelp position={{ right: '2rem' }} help={help} />}
+        </div>
+      </Table.Cell>
+      <Table.Cell>{value}</Table.Cell>
+    </Table.Row>
+  );
+};
+
+const OverviewTable: React.FC<{data: TokenomicsData | undefined; statusData: StatusServerData | undefined | null}> = ({ data, statusData }) => {
```
Lezek123: You could use `data?: TokenomicsData` as a shortcut for `data: TokenomicsData | undefined`

pioneer/packages/app-tokenomics/src/Overview/OverviewTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <Table.Body>
+        <OverviewTableRow
+          item='Total Issuance'
+          help='The current supply of tokens.'
+          value={data ? `${data.totalIssuance} JOY` : 'Loading...'}
+        />
+        <OverviewTableRow
+          item='Fiat Pool'
+          help='The current value of the Fiat Pool.'
+          value={statusData === null ? 'Data currently unavailable..' : statusData ? `${round(statusData.dollarPool.size)} USD` : 'Loading...'}
```
Lezek123: You could use `statusData.dollarPool.size.toFixed(2)` and at the same time avoid defining custom `round()` method.

This series of conditions:
```
statusData === null ? 'Data currently unavailable..' : statusData ? `${value} ${unit}` : 'Loading...'
```

Is also repeated quite a few times later in the code, so perhaps can be factored out to something like:
```
const displayStatusData = (val?: string, unit?: string): string => (
    statusData === null ? 'Data currently unavailable...' : statusData ? `${val} ${unit}` : 'Loading...'
);
```
And then used like this:
```
value={displayStatusData(statusData?.dollarPool.size.toFixed(2), 'USD')}
```

pioneer/packages/app-tokenomics/src/Overview/Help.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from 'react';
+import { Icon, Label, Transition } from 'semantic-ui-react';
+
+type HelpProps = {help: string; className?: string; position: { left?: string; right?: string}};
+
+export default function Help (props: HelpProps): React.ReactElement {
```
Lezek123: There are already a few components that are available and serve the same puprose:
- In `semantic-ui-react`: https://react.semantic-ui.com/modules/popup
- `@polkadot/react-components/src/LabelHelp.tsx`
- A very similar implementation is inside `joy-proposals/src/forms/LabelWithHelp.tsx` (you probably even reused some code from there)

It would be best to avoid creating the same components over and over, so perhaps you could either:
- Use the one that comes from one of the 2 libraries mentioned above
- Put this one inside `joy-utils/src/react/components` (the `joy-utils` package is currently a bit of a mess, but perhaps in the future it could serve as library of resuable react stuff that other packages can import without worrying about circular dependencies etc.)

pioneer/packages/app-tokenomics/src/Overview/PieChart.tsx
```diff
@@ -0,0 +1,52 @@
...
+import styled from 'styled-components';
+import { Icon, Label } from 'semantic-ui-react';
+
+const ChartContainer = styled('div')`
+  display:flex;
+  flex-direction:column;
+  align-items:center;
+`;
+
+const PieChart: React.FC<{icon: any; typeOfChart: string; percentages: Array<{ percent: any; color: string}>}> = ({ icon, typeOfChart, percentages }) => {
```
Lezek123: Putting aside the way DOM elements are handled here, this is a similar situation to the one with `Help.tsx` component.

We already have `react-chartjs-2` available in the project (dependency of `@polkadot/react-components`) which can be used to create pie charts (https://medium.com/@rkbhatia89/how-to-make-pie-charts-in-react-with-chart-js-64edd03fc922), so perhaps you could take advantage of it (it looks like it also provides some additional features too!)

pioneer/packages/app-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+const StyledTable = styled(Table)`
+  border: none !important;
+  width: 70% !important;
+  margin-bottom:1.5rem;
+  @media (max-width: 1400px){
+    width:100% !important;
+  }
+  & > tbody > tr{
+    td:nth-of-type(1), td:nth-of-type(3), td:nth-of-type(6){
```
Lezek123: I feel there's a bit too much going on here and this solution doesn't seem very flexible if we decide to add rows/columns to the table.

I understand the purpose is to create a table that has an outside border and is also divided internally.
The internal division of columns can currently be achieved with something like:
```
  & tr {
    td:nth-of-type(3n),
    th:nth-of-type(3n) {
      border-left: 0.12rem solid rgba(20,20,20,0.3) !important;
    }
  }
```
Or the styled component could take `divideColumsAt` for more flexibility (ie. `divideColumsAt={[3, 6, 9]}`).

I see there is some issue with the outer table border if you set it directly to the table, but you could get around this by creating a `div` container

pioneer/packages/app-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+        <SpendingAndStakeTableRow
+          role={width <= 1050 ? 'Validators' : 'Validators (Nominators)'}
+          helpContent='The current set of active Validators (and Nominators), and the sum of the sets projected rewards and total stakes (including Nominators).'
+          helpPosition={width <= 767 ? 'right' : 'left'}
+          numberOfActors={data && `${data.validators.number} (${data.validators.nominators.number})`}
+          groupEarning={data && `${Math.round(data.validators.rewardsPerWeek)}`}
+          groupEarningDollar={statusData === null ? 'Data currently unavailable..' : (data && statusData) && `${round(data.validators.rewardsPerWeek * Number(statusData.price))}`}
+          earningShare={data && `${round(data.validators.rewardsShare * 100)}`}
+          groupStake={data && `${data.validators.totalStake}`}
+          groupStakeDollar={statusData === null ? 'Data currently unavailable..' : (data && statusData) && `${round(data.validators.totalStake * Number(statusData.price))}`}
```
Lezek123: Those are repeated conditions similar to the ones inside `OverviewTable.tsx` so you could consider factoring it out to an arrow function too. Something like the `parseData` function that you defined inside `SpendingAndStakeTableRow`

pioneer/packages/app-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+          stakeShare={data && `${round(data.validators.stakeShare * 100)}`}
+          color='rgb(246, 109, 68)'
+        />
+        <SpendingAndStakeTableRow
+          role={width <= 1015 ? 'Council' : 'Council Members'}
+          helpContent='The current Council Members, and the sum of their projected rewards and total stakes (including voters/backers).'
+          helpPosition={width <= 767 ? 'right' : 'left'}
```
Lezek123: The number here implies that this depends on a `@media` breakpoint.
The problem with mixing media breakpoints with `window.innerWidth`/`window.innerHeight` is that those values don't always match (depending on the browser etc.), which can result in broken design on some screen sizes.
If this indeed depends on `media` breakpoint I'd advise to also put it in a `media` breakpoint.

pioneer/packages/app-tokenomics/src/Overview/TokenomicsCharts.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import PieChart from './PieChart';
+import { Icon } from 'semantic-ui-react';
+import styled from 'styled-components';
+
+import { TokenomicsData } from '../lib/getTokenomicsData';
+
+const TokenomicsChartsContainer = styled('div')`
+  width:30%;
```
Lezek123: Since this is a separate component I'd generally avoid setting a width which only works in one specific context.

pioneer/packages/app-tokenomics/src/Overview/TokenomicsCharts.tsx
```diff
@@ -0,0 +1,95 @@
...
+  display:flex;
+  align-items:center;
+  justify-content:space-evenly;
+  padding: 2rem 0;
+  svg{
```
Lezek123: `svg` is very generic (it could also mean an icon that's part of the component too), so ideally I'd narrow it down with a class or create `StyledPieChart` component, but since I already recommended using `react-chartjs-2`, perhaps this comment will not apply anymore.

pioneer/packages/app-tokenomics/src/Overview/TokenomicsCharts.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+  @media (max-width: 550px){
+    flex-direction:column;
+    & > div {
+      margin-bottom: 1.5rem;
+    }
+  }
+`;
+
+const TokenomicsCharts: React.FC<{data: TokenomicsData | undefined}> = ({ data }) => {
```
Lezek123: `data?: TokenomicsData`, it would be best to use this syntax consistently instead of `| undefined`

pioneer/packages/app-tokenomics/src/Overview/hooks/useWindowDimensions.ts
```diff
@@ -0,0 +1,24 @@
+import { useState, useEffect } from 'react';
+
+function getWindowDimensions () {
```
Lezek123: For the reason I mentioned before I think it would be best to stick with `@media` queries for consistency, but if you really see the need to use this hook I'd place it in `joy-utils/src/react/hooks` (as it is not something that's specific to this package)

pioneer/packages/app-tokenomics/src/Overview/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    flex-direction:column;
+  }
+`;
+
+const Title = styled('h2')`
+  border-bottom: 1px solid #ddd;
+  margin: 0 0 2rem 0;
+`;
+
+export type StatusServerData = {
```
Lezek123: Since we're already fetching data from the status server it may make sense to use more of that data and limit the number of additional `api` queries that are executed if possible (ie. we can get `totalIssuance`, the count of members in each role etc. from the status server too).

pioneer/packages/app-tokenomics/src/Overview/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const getApiData = async (): Promise<void> => {
+    try {
+      const data = await getTokenomicsData(api);
+      setData(data);
+    } catch (e) {
+      console.log(e);
+    }
+  };
+
+  useEffect(() => {
```
Lezek123: There is a `usePromise` hook inside `@polkadot/joy-utils/react/hooks` that can be used here instead. It would prevent calling `setStatusData` after the component is unmounted.

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Reward } from './getTokenomicsData';
+
+type ContentCurator = {
+  reward_relationship: number;
+  role_stake_profile: {
+    stake_id: number;
+  };
+}
+
+const getCurators = async (api: ApiPromise): Promise<Array<ContentCurator>> => {
```
Lezek123: As you mentioned in the PR description, there is already a more-or-less standarized way of fetching data from the `api` in Pioneer and this is via the "transport". Not all of the packages use this approach yet, but there a few classes inside `joy-utils/src/transport` that provide a way to query and parse commonly used data from the chain. I made a few changes there recentlly to make it more accessible, but currently you could, for example, wrap the app inside `TransportProvider` (reference: `joystream/joystream/pioneer/packages/joy-proposals/src/index.tsx`) and then use `useTransport()` hook to access those methods inside a component (and define the new ones that could be usfeul for other apps).

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
```diff
@@ -0,0 +1,58 @@
...
+  const curators: Array<ContentCurator> = [];
+  const [workerIds, curatorData] = ((await api.query.contentWorkingGroup.curatorById()).toJSON() as [Array<number>, Array<object>]);
```
Lezek123: Currently it's hard to figure out the right types when querying the data from the `api`, but hopefully this will be way easier soon due to `@polkadot/api` upgrade, types augmentation and the removal of `LinkedMaps`, which are all part of the `Alexandria` realase that our team has been working on lately.

So I'm not sure if it's worth fixing now, but the right, type-safe way currently would be do something like:
```
    import { MultipleLinkedMapEntry } from '@polkadot/joy-utils/LinkedMapEntry';
    import { Curator, CuratorId } from '@joystream/types/content-working-group';

    /* ... */

    const mapEntries = new MultipleLinkedMapEntry(
      CuratorId,
      Curator,
      await api.query.contentWorkingGroup.curatorById()
    );

    const activeCurators = mapEntries.linked_values.filter(c => !c.isEmpty && c.is_active);
```

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+  const curatorStakeData = await api.query.stake.stakes.multi(stakeIds);
+  curatorStakeData.map((stakeData: any) => {
+    totalContentCuratorStake += stakeData.toJSON()[0].staking_status.Staked.staked_amount;
+  });
+  return totalContentCuratorStake;
+};
+
+export default async (api: ApiPromise, recurringRewards: Array<Reward>): Promise<any> => {
+  const currentLead = (await api.query.contentWorkingGroup.currentLeadId()).toJSON() as number;
```
Lezek123: So, for example, this is available via `joy-utils/src/transport/contentWorkingGroup.ts`.
The type-safe approach to get the lead id is:
```
    const optLeadId = (await this.contentWorkingGroup.currentLeadId()) as Option<LeadId>;
    const leadId = optLeadId.unwrapOr(null); // The type of this is: LeadId | null
```

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (reward && reward.amount_per_payout && reward.payout_interval) {
+      rewardsPerBlock += reward.amount_per_payout / reward.payout_interval;
+    }
+  });
+  return rewardsPerBlock;
+};
+
+const calculateStake = async (api: ApiPromise, curators: Array<ContentCurator>): Promise<number> => {
+  const stakeIds: Array<number> = []; let totalContentCuratorStake = 0;
+  curators.forEach((curator: any) => {
```
Lezek123: The use of `any` type should be generally avoided and this is actually not needed here, as the type of `curator` should be already derived from `curators` by the compailer (if you remove `any` it will be `ContentCurator`)

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (curator.role_stake_profile) {
+      stakeIds.push(curator.role_stake_profile.stake_id);
+    }
+  });
+  const curatorStakeData = await api.query.stake.stakes.multi(stakeIds);
+  curatorStakeData.map((stakeData: any) => {
+    totalContentCuratorStake += stakeData.toJSON()[0].staking_status.Staked.staked_amount;
```
Lezek123: Here you could take a look at the `stakeValue` method from `pioneer/packages/joy-utils/src/transport/workingGroups.ts` for reference on how to get stake value in a type-safe way (this is a bit complicated, because of the `LinkedMap`, which won't be available after `Alexandria` as I mentioned before).

pioneer/packages/app-tokenomics/src/lib/getCouncilData.ts
```diff
@@ -0,0 +1,70 @@
+import { Vec, u32 } from '@polkadot/types';
+import { ApiPromise } from '@polkadot/api';
+import { BlockNumber, BalanceOf } from '@polkadot/types/interfaces';
+
+const getCouncilMembers = async (api: ApiPromise): Promise<any> => {
+  let totalStake = 0;
+  const activeCouncil = await api.query.council.activeCouncil() as Vec<u32>;
+  const payoutInterval = (await api.query.council.payoutInterval()).toJSON() as number;
```
Lezek123: This will be number | null, because the query returns `Option<BlockNumber>`. You can see the return type of any query on `https://testnet.joystream.org/#/chainstate` and assert the result to be of this type (as you already tried to do with `activeCouncil` above which is of type `Seats` from `@joystream/types/council`)

pioneer/packages/app-tokenomics/src/lib/getCouncilData.ts
```diff
@@ -0,0 +1,70 @@
...
+  const amountPerPayout = (await api.query.council.amountPerPayout() as BalanceOf).toNumber();
+  console.log(amountPerPayout, payoutInterval);
```
Lezek123: This can be removed I guess.

pioneer/packages/app-tokenomics/src/lib/getCouncilData.ts
```diff
@@ -0,0 +1,70 @@
...
+  activeCouncil.map((member: any) => {
```
Lezek123: Avoid asserting `any`, especially because `activeCouncil` has been already asserted to be `Vec<u32>`. Skipping `any` here would reveal an error, because the right type for `activeCouncil` should in fact be `Seats` (you can see `packages/joy-utils/src/transport/council.ts` for reference)

pioneer/packages/app-tokenomics/src/lib/getCouncilData.ts
```diff
@@ -0,0 +1,70 @@
...
+    totalStake += stakeAmount;
+  });
+  return {
+    numberOfCouncilMembers: activeCouncil.length,
+    totalCouncilRewardsPerBlock: (amountPerPayout * activeCouncil.length) / payoutInterval,
+    totalCouncilStake: totalStake
+  };
+};
+
+const calculateCouncilRewards = async (api: ApiPromise, totalCouncilRewardsPerBlock: number): Promise<any> => {
```
Lezek123: `Promise<number>` (it would require a minor tweak inside the function by moving the `return` statement below `while`, but this should make it more clear anyway)

pioneer/packages/app-tokenomics/src/lib/getCouncilData.ts
```diff
@@ -0,0 +1,70 @@
...
+      weekInBlocks -= announcingPeriod;
+    } else {
+      return councilRewardsInOneWeek;
+    }
+  }
+};
+
+export default async (api: ApiPromise): Promise<any> => {
+  const { numberOfCouncilMembers, totalCouncilRewardsPerBlock, totalCouncilStake } = await getCouncilMembers(api);
+  const totalCouncilRewardsInOneWeek = await calculateCouncilRewards(api, totalCouncilRewardsPerBlock) as number;
```
Lezek123: You could get rid of the assertion after fixing `calculateCouncilRewards` return type.

pioneer/packages/app-tokenomics/src/lib/getStorageProviderData.ts
```diff
@@ -0,0 +1,102 @@
...
+type StorageProvider = {
+  reward_relationship: number;
+  role_stake_profile: {
+    stake_id: number;
+  };
+  workerId: number;
+  lead?: boolean;
+}
+
+const getStorageProviders = async (api: ApiPromise): Promise<any> => {
```
Lezek123: Here you could perhaps take a look at `/cli/src/Api.ts` method like `groupMembers` for example of a typesafe implementation (I really hope we will have this normalized across projects soon)

pioneer/packages/app-tokenomics/src/translate.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright 2017-2019 @polkadot/app-123code authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { withTranslation } from 'react-i18next';
+
+export default withTranslation(['app-123code']);
```
Lezek123: The package name seems wrong here

pioneer/packages/apps-routing/src/tokenomics.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import Tokenomics from '@polkadot/app-tokenomics/';
+
+export default ([
+  {
+    Component: Tokenomics,
+    display: {
+      needsAccounts: true,
+      needsApi: [
+        'tx.balances.transfer'
```
Lezek123: Here you would usually put the name of some method that is actually used by the package (ie. `query.stake.stakes`, but this doesn't need to be very strict.

pioneer/packages/app-tokenomics/src/lib/getValidatorData.ts
```diff
@@ -0,0 +1,51 @@
+import { ApiPromise } from '@polkadot/api';
+import { Vec } from '@polkadot/types';
+import { AccountId, IndividualExposure } from '@polkadot/types/interfaces';
+
+const getValidators = async (api: ApiPromise): Promise<any> => {
```
Lezek123: The return value could be perhaps derived by TypeScript compiler itself if you omitted the `: Promise<any>`

pioneer/packages/app-tokenomics/src/lib/getValidatorData.ts
```diff
@@ -0,0 +1,51 @@
...
+  let [totalValidatorStake, slotStake, numberOfNominators, numberOfValidators] = [0, 0, 0, 0];
+  const validatorIds = await api.query.staking.currentElected() as Vec<AccountId>;
+  await Promise.all(validatorIds.map(async (id: AccountId, index: number) => {
+    const nominators = (await api.derive.staking.info(id)).stakers?.others as Vec<IndividualExposure>;
+    const totalStake = (await api.derive.staking.info(id)).stakers?.total.toNumber() as number;
```
Lezek123: It seems that (accoring to TypeScript) `.stakers` can be `undefined` here, so perhaps a better way to deal with it would be:
```
    const nominators = (await api.derive.staking.info(id)).stakers?.others.toArray() || [];
    const totalStake = (await api.derive.staking.info(id)).stakers?.total.toNumber() || 0;
```

Lezek123 (changes requested):
Overall it seem to work pretty well, my comments are mostly regarding best practices and type-safety on the coding-side (I know the latter is very hard to achieve now, because `api` queries lack return types very badly, but I was working on that lately and I believe we will have this available pretty soon!).

I think the `Alexandria` upgrade that we've been working on latetly will affect this code a lot, so most of the queries etc. will have to be changed anyway, but I hope my input will provide you some hints on how this code could be improved now.

Thanks for contributing!

pioneer/packages/app-tokenomics/src/Overview/OverviewTable.tsx
DzhideX: Do I just leave these as is? (As they're reliant on both data & statusData)

```
statusData === null ? 'Data currently unavailable..' : (data && statusData) ? `${(data.totalWeeklySpending * Number(statusData.price)).toFixed(2)} USD` : 'Loading...'}
```

There are only two of these so I don't think it'd be much of a problem but maybe there's a sleeker solution that I'm missing?

pioneer/packages/app-tokenomics/src/Overview/PieChart.tsx
DzhideX: I'm going to assume this means you want me to make a PieChart component inside of '@polkadot/react-components/Chart' and then import it?

bwhm:
As @Lezek123 is on vacation, I propose that @DzhideX will: 
- fix the "important" stuff
- fix the low hanging fruit
- address all the remaining comments
- push his changes
Then I test it, and if it looks good:
- @mnaamani does a quick review and:
     - if satisfied, merges and deploys
     - if not, we'll leave it for Alexandria

---

Regardless @DzhideX moves on to make it ready for Alexandria.

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
DzhideX: I think it's not worth trying to implement this at the moment.

pioneer/packages/app-tokenomics/src/lib/getContentCuratorData.ts
DzhideX: This one is incredibly nice, but it would take a very long time to implement all the logic. I'll keep this one(and some of the other tips you've had that I may not have commented on) in mind when adapting this for `Alexandria`.

DzhideX:
This works great but I thought I'd push this commit to see where to go next regarding general code arrangement, cleanliness, and type safety.

Note:

- I wrote all of the functions inside of `joy-utils/src/transport/tokenomic.ts`. I did this because I felt it better to delegate to @Lezek123  to decide which of these are best kept here and which should/can be moved to their respective transport.

- There is a new error in apps-routing which shows that there is some sort of incompatibility with Props that I tried to fix but wasn't quite sure how to.

- There were some unavoidable(from what I can tell) typescript linting errors in `transport/tokenomics/ts` on lines 139, 142, etc.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -0,0 +1,362 @@
...
+      leadNumber,
+      leadRewardId,
+      leadStakeId
+    };
+  }
+
+  async calcuateStorageProvider (stakeIds: StakeId[], leadStakeId: StakeId | null, rewardIds: RewardRelationshipId[], leadRewardId: RewardRelationshipId | null) {
+    let totalStorageProviderStake = 0; let leadStake = 0; let storageProviderRewardsPerBlock = 0; let storageLeadRewardsPerBlock = 0;
+
+    (await this.api.query.stake.stakes.multi(stakeIds) as Stake[]).forEach((stake) => {
```
Lezek123: Try `this.api.query.stake.stakes.multi<Stake>(stakeIds)`, you can also use syntax like that for other queries that cause the errors.

pioneer/packages/joy-tokenomics/src/index.tsx
```diff
@@ -1,33 +1,37 @@
...
+        </header>
+        <Switch>
+          <Route component={Overview} />
+        </Switch>
+      </main>
+    </TransportProvider>
   );
 }
 
-export default translate(App);
```
Lezek123: That's the change that caused `TypeScript` error in `packages/apps-routing/src/joy-tokenomics.ts`.
A simple way to fix it:
- Revert this (along with `pioneer/packages/joy-tokenomics/src/translate.ts` changes)
- Change Props interface to:
```
import { RouteProps as AppMainRouteProps } from '@polkadot/apps-routing/types';

interface Props extends AppMainRouteProps, I18nProps {}
```

Lezek123 (changes requested):
The changes look great, I didn't go through them thoroughly yet, but leaving some tips how to fix the build.

pioneer/packages/joy-tokenomics/src/index.tsx
DzhideX: The reason I initially changed this was because the error was immediately there after moving to Iznik but I couldn't figure out what it was and just left it like that. 

I tried reverting the changes and doing exactly like you wrote but it still didn't work. Same error. But just adding these two lines to the file as it is now allows it to work properly. Should I just go along with that, then?

DzhideX: Still getting an error. While hovering translate(App):
```
Types of parameters '__0' and 'props' are incompatible.
      Type 'PropsWithChildren<WithTranslation>' is missing the following properties from type 'Props': basePath, onStatusChange
```
While hovering Component in routing:
```
Type 'ComponentType<Pick<WithTranslation, never> & WithTranslationProps>' is not assignable to type 'ComponentType<RouteProps>'.
```
I got it working by not reverting the changes and just deleting and not extending I18nProps. I assume that is far from optimal but I thought it'd be good to mention.

bwhm:
This worked fine by just doing `yarn workspace pioneer run (start?)`, but the build did in fact fail...

So I think we may have to sort it out after Alexandria is live.

bwhm:
Small fixes only:
- After the upgrade, we changed the recurring reward from weekly to biweekly, in order to follow the council term.
This means that the "Weekly top-ups" should be half of the `replenishAmount`.
 - As the council term is now 10 days (excluding "announcing", "voting" and "revealing" stage), or two weeks (including stages), I'm guessing that the council reward will also be calculated incorrectly.

If you can fix those two, I think we should merge this.
The "Overview" will get more tabs later, and small improvements to the "tokenomics" can be made later!

pioneer/packages/joy-tokenomics/src/index.tsx
Lezek123: Taking another look at it I think both my and your solution works, because `withTranslation(['joy-tokenomics'])(App)` should basically strip `I18nProps` (more specifically: `WithTranslation`) props in the resulting component and just add some optional ones (I'm not sure why it caused error in your IDE).

If you're using hooks instead of `withTranslation` then just removing `I18nProps` from `Props` seems perfectly valid and is probably even better, so if it works - go for it!

Lezek123 (changes requested):
Excellent, checked the code and tested functionality, everything looks and works great!

Created PR to your branch with just a few minor fixes (https://github.com/DzhideX/joystream/pull/1)

- Fixed build error based on your previous comment
- We don't have any linter rules to enforce this in Pioneer yet, but generally very long lines (ie. beyond full hd screen width) should be avoided, because they're hard to read and debug in case of errors. I fixed a few of those in `joy-utils/src/transport/tokenomics.ts`
- I factored out staking-related constants and a function to calculte validators total reward per era, as it seems quite complex (took me some time to parse), but very useful. I hope it's now more readable and can also be sort of "unit tested". I also added some comments explaining where those values / formulas come from.
- Removed `TransportProvider` in `joy-tokenomics/src/index.tsx`, as it is now provided globally for all apps (see: `apps/src/Content/index.tsx`)
- Fixed a few linter errors and some other very minor stuff

I think calculating working groups rewards and stakes can be simplified with `workingGroups` transport, but we'll leave it as is for now, because this will have to be changed in the next release (`Babylon`) anyway

After you merge my PR and implement fixes mentioned by @bwhm, I think we can merge to `master`!

Lezek123:
> As the council term is now 10 days (excluding "announcing", "voting" and "revealing" stage), or two weeks (including stages), I'm guessing that the council reward will also be calculated incorrectly.

In that case it will be just: `(amountPerPayout * numberOfCouncilMembers) / payoutInterval * BLOCKS_PER_WEEK` no matter which stage we're currently in. Should it depend on the current stage?

bwhm:
> > As the council term is now 10 days (excluding "announcing", "voting" and "revealing" stage), or two weeks (including stages), I'm guessing that the council reward will also be calculated incorrectly.
> 
> In that case it will be just: `(amountPerPayout * numberOfCouncilMembers) / payoutInterval * BLOCKS_PER_WEEK` no matter which stage we're currently in. Should it depend on the current stage?

This will always be a projection anyway, as there is no way to know any of these values in advance. I think showing what will happen on "average" based on the current state is better, so one should assume the council will receive their recurring reward on 10/14 -> 5/7 days.

This makes it easier to evaluate current spending, and let individuals and the Council make informed decisions.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -44,8 +44,8 @@ export default class TokenomicsTransport extends BaseTransport {
   }
 
   async calculateCouncilRewards (numberOfCouncilMembers: number) {
```
Lezek123: I think the idea here was make it sort of universal, ie.:
```
const councilTermDurationRatio = termDuration / (termDuration + announcingPeriod + votingPeriod + revealingPeriod);
const avgCouncilRewardPerBlock = councilTermDurationRatio * totalCouncilRewardsPerBlock;
const avgCouncilRewardPerWeek = avgCouncilRewardPerBlock * (7 * 24 * 60 * 60) / 6;
```
Pinging @bwhm to confirm

bwhm: Not sure I understand, but:

To be precise, the `councilMembers` only earn rewards during their `newTermDuration`, ie. when there are no election ongoing, but they still have the same "power" during `announcingPeriod+votingPeriod+revealingPeriod`.

So to get the average (weekly) spending on CM rewards:
```
const weeklyAvgCouncilReward = activeCouncil.length * amountPerPayout * (blocksPerWeek / payoutInterval) * newTermDuration / (newTermDuration+announcingPeriod+votingPeriod+revealingPeriod) 
```
-> weeklyAvgCouncilReward = 6*18500*(100800/3600*)144000/201600 = 2220000

Lezek123: I think our calculations are pretty much the same then.
The point is that this way we can get rid of the entrie `while` loop in `calculateCouncilRewards` that requires setting some arbitrary period in order to run the calculation (which doesn't seem necessary and is not very flexible) @DzhideX

DzhideX: Ok. I am very much for this as that was my least favorite part of the code as well.

Lezek123 (approved):
Great, I think this is ready for a merge.

The last thing I would suggest changing (which is not a high priority) is the naming of the transport methods to be more descriptive, ie.:
- `getCouncilMembers` - `councilSizeAndStake`
- `calculateCouncilRewards` - `councilRewardsPerWeek`
etc.

You could also mark methods that aren't very usable "on the outside" with `private` (ie. `calcuateStorageProvider`).

DzhideX:
Agreed. I think that also in case of the tab being slow(er), there are optimizations that can be done that I've sacrificed for readability, like: 
```
const { new_term_duration, voting_period, revealing_period, announcing_period } = await this.councilT.electionParameters();
```
This method along with the data used in `tokenomics.ts` also gets data for 
```
min_council_stake,min_voting_stake,candidacy_limit,council_size
```
I think there are more similar instances but I think it's fine for now!

mnaamani:
Nice work @DzhideX !
All CI checks are green,
Deploy preview looks good,
based on approval and testing by martin and leszek.. merging.

PR #1231, opened 2020-09-04
Genesis initial balances

mnaamani:
Option to include initial balances for accounts at genesis from exported json file.

Corresponding export script: https://github.com/Joystream/runtime-state-export-tools/pull/1/commits/0b787aa7d6da4196351d4e3a1177a265ef0f75bb

### DIFF
--- a/utils/chain-spec-builder/src/main.rs
+++ b/utils/chain-spec-builder/src/main.rs
@@ -24,8 +24,8 @@ use rand::{distributions::Alphanumeric, rngs::OsRng, Rng};
 use structopt::StructOpt;
 
 use joystream_node::chain_spec::{
-    self, chain_spec_properties, content_config, forum_config, initial_members, proposals_config,
-    AccountId,
+    self, chain_spec_properties, content_config, forum_config, initial_balances, initial_members,
+    proposals_config, AccountId,
 };
 
 use sc_chain_spec::ChainType;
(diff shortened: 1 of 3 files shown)


bwhm (approved):

utils/chain-spec-builder/src/main.rs
```diff
@@ -258,9 +287,11 @@ fn genesis_constructor(
         versioned_store_permissions_cfg,
         data_directory_config,
         content_working_group_config,
+        initial_account_balances,
     )
 }
```
shamil-gadelshin, line 292: Please add a small comment on why we disable the lint here.

utils/chain-spec-builder/src/main.rs
```diff
@@ -191,6 +212,7 @@ impl ChainSpecBuilder {
     }
 }
```
shamil-gadelshin, line 214: I noticed you disabled this lint at least twice. Do you mind to convert it to module-scope lint with a comment?

utils/chain-spec-builder/src/main.rs
```diff
@@ -241,6 +264,12 @@ fn genesis_constructor(
         )
     };
```
shamil-gadelshin, line 266: What do you think about functional style or you prefer the imperative style more?
let initial_account_balances = initial_balances_path.map(|path| {initial_balances::from_json(path.as_path())}).unwrap_or_else(vec![]);

shamil-gadelshin (changes requested):
Just a couple of minor comments.

utils/chain-spec-builder/src/main.rs
mnaamani: Applied suggestion to other code in same function as well.

utils/chain-spec-builder/src/main.rs
mnaamani: done

shamil-gadelshin (approved):
LGTM

PR #1236, opened 2020-09-04
Proposal error message hotfix

Lezek123:
Probable cause: `Vec<u8>.toString()` now seems to return the json representation of the array instead of hex string.


Lezek123:
I used `Bytes` to avoid issues like https://github.com/Joystream/apps/issues/419, but since this data can be trusted, `Text` will be a better choice.

It's possible that the issue with `Text` decoding is already solved in the new api and `bytesToString` workaround can be removed completely (will have to test).

PR #1246, opened 2020-09-07
Node: re-enabled some unit tests

mnaamani:
Re-enabled the node sc-service-test based unit tests for the node.
Incresed time allowed with travis_await to allow enough time for those tests to run, and enabled them in travis.yml

Locally the tests run and succeed but I still see this in logs:

`[2020-09-07T07:59:06Z ERROR sc_network::protocol::sync] 💔 Called on_block_announce with a bad peer ID`

I think we might have some configuration error perhaps, because when running the node in production on staging network @bwhm observed this:

```
You are using the peer ID QmasDuhJ3AJRYCwUerpmKHw9aN8eae8NNeEuKKGrNyToSn.
This peer ID uses a legacy, deprecated representation that will no longer be supported in the future.
Please refresh it by performing a RPC query to the appropriate node,
by looking at its logs, or by using `subkey inspect-node-key` on its private key.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
 script:
   - export WASM_BUILD_TOOLCHAIN=nightly-2020-05-23
   - BUILD_DUMMY_WASM_BINARY=1 cargo clippy --release --all -- -D warnings
-  - travis_wait 75 cargo test --release --verbose --all --ignored
+  - travis_wait 75 cargo test --release --verbose --all -- --ignored
   - cargo build --release
   - ls -l ./target/release/wbuild/joystream-node-runtime/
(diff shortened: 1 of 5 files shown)


mnaamani:
I think the error from running the node is because the chainspec is using the old format of the peerid in the bootnodes list:

```
  "bootNodes": [
    "/ip4/139.162.227.183/tcp/30333/p2p/QmasDuhJ3AJRYCwUerpmKHw9aN8eae8NNeEuKKGrNyToSn"
  ],
```

mnaamani:
I think I need to fix the placement of the `--ignored` argument. That is why travis failed.

shamil-gadelshin (approved):
LGTM

PR #1248, opened 2020-09-07
Atlas icon component

fcescob:
This PR implements an Icon component to improve icon consumption through Atlas.

Here is an example to illustrate the difference with the current method of importing icons:
Currently:
```javascript
import PlayIcon from '../path/to/PlayIcon.svg'

function myComponent(){
// ...
return <PlayIcon />
}
```
With Icon component: 
```javascript
import Icon from '../path/to/Icon'

function myComponent(){
// the icon prop here gets autocomplete
return <Icon icon="play" />
}

```

## Feedback

It would be very good to have some feedback on the implementation of the `Icon` component itself.
It has at least one incovenience, each time a new icon is added to the codebase a manual change is needed to reflect that,  so available icons and actual icons in the codebase could go out of sync.
Builds on top of #977


atlas/packages/app/src/shared/components/Checkbox/Checkbox.tsx
```diff
@@ -25,7 +25,8 @@ export default function Checkbox({
       <div css={styles.outerContainer}>
         <div css={styles.innerContainer}>
           <input css={styles.input} type="checkbox" checked={selected} disabled={disabled} onChange={onChange} />
-          {selected && (icon === 'check' ? <CheckIcon css={styles.icon} /> : <DashIcon css={styles.icon} />)}
+          {selected &&
+            (icon === 'check' ? <Icon icon="check" css={styles.icon} /> : <Icon icon="dash" css={styles.icon} />)}
```
kdembler: Since it's the same component now it'd be more readable to write it as:
```tsx
{selected && <Icon icon={icon === 'check' ? 'check' : 'dash'} css={styles.icon} />}
```

atlas/packages/app/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,30 @@
...
+type IconProps = {
+  icon: IconType
+} & React.SVGProps<SVGSVGElement>
+
+const capitalize = (s: string) => s.slice(0, 1).toUpperCase() + s.slice(1)
+const pascalCase = (s: string) => capitalize(camelCase(s))
+
+const Icon: React.FC<IconProps> = ({ icon, ...svgProps }) => {
+  const iconProp = pascalCase(icon) as keyof typeof Icons
+  if (!Object.keys(Icons).includes(iconProp)) return null
```
kdembler: Two things here:
1) It would be useful for debugging to add some console warning/error here in addition to returning `null`
2) Since it will be constant, let's move `Object.keys(Icons)` outside of the render function

atlas/packages/app/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { css } from '@emotion/core'
+import * as Icons from '../../icons'
+import { camelCase } from 'lodash'
+
+type IconType = Icons.IconType
+type IconProps = {
+  icon: IconType
```
kdembler: Maybe `name` would be more fitting since we already have `Icon` as the component name?

atlas/packages/app/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,30 @@
...
+} & React.SVGProps<SVGSVGElement>
+
+const capitalize = (s: string) => s.slice(0, 1).toUpperCase() + s.slice(1)
+const pascalCase = (s: string) => capitalize(camelCase(s))
+
+const Icon: React.FC<IconProps> = ({ icon, ...svgProps }) => {
+  const iconProp = pascalCase(icon) as keyof typeof Icons
+  if (!Object.keys(Icons).includes(iconProp)) return null
+
+  const iconStyles = css`
```
kdembler: Same as 2. above, we can move this outside the render function

atlas/packages/app/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,30 @@
...
+    stroke: currentColor;
+    & > * {
```
kdembler: Also FYI, I don't think you need the `&` in the selector, `> *` should work exactly the same. Or is there some difference that you know of?

atlas/packages/app/src/shared/components/NavButton/NavButton.tsx
```diff
@@ -13,7 +13,7 @@ export default function NavButton({ direction = 'right', onClick, outerCss, ...s
   const styles = useCSS(styleProps)
   return (
     <button css={[styles, outerCss]} onClick={onClick}>
-      {direction === 'right' ? <ChevronRightIcon /> : <ChevronLeftIcon />}
+      {direction === 'right' ? <Icon icon="chevron-right" /> : <Icon icon="chevron-left" />}
```
kdembler: Same story here, let's just do the ternary inside the `icon` prop

atlas/packages/app/src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -38,6 +39,7 @@ export const CoverHoverOverlay = styled.div`
   align-items: center;
 `
 
+const PlayIcon = () => <Icon icon="play" />
```
kdembler: Something broke here, the video preview no longer shows the play icon on hover:
Screenshot 2020-09-07 at 12 55 19 PM

kdembler (changes requested):
Overall looking good. I'm a bit concerned about the maintainability of this in the long term because of how you need to define an icon in more than one place but it's a step forward so let's iterate on this in the future.

atlas/packages/app/src/shared/components/VideoPlayer/VideoPlayer.style.tsx
```diff
@@ -134,7 +135,9 @@ export const PlayOverlay = styled.div`
   cursor: pointer;
 `
 
-export const StyledPlayIcon = styled(PlayIcon)`
+const StyledIcon = styled(Icon)`
   height: 72px;
   width: 72px;
+  color: ${colors.white};
 `
+export const StyledPlayIcon = ({ ...svgProps }) => <StyledIcon name="play" {...svgProps} />
```
kdembler, line 143: FYI, I _think_ you should be able to do this inline with defining the styled component, like so:

```tsx
const StyledIcon = styled(({ ...props }) => <Icon name="play" {...props} />)`
  height: 72px;
  width: 72px;
  color: ${colors.white};
`
```

Your solution is perfectly fine as well so don't bother changing it

kdembler (approved):
Just one FYI comment, otherwise looks good!

atlas/packages/app/src/shared/components/VideoPlayer/VideoPlayer.style.tsx
fcescob: good catch! Will keep it in mind for the next time something like this comes out!

bedeho (approved):

Issue #1201, opened 2020-08-31
Discussion: Stake reuse logic implementaion


bedeho:
# Background

We are currently going to be staking for the following purposes in our runtime, with additional purposes undoubtedly on the horizon:

1. voting
2. candidacy in an election
3. submitting proposal
4. applying for a role in the working group
5. holding a role in a working group
6. validation
8. nomination

We already know that we want stake deployed for purposes 3-8 should also be countable towards 1 and 2, although not simultaneously. It's not entirely clear whether we should accept any other stake reuse, but my current hunch is no.

The question is then, what sort of approach should we apply at the module level to implement this?

Currently, we are working on a new final iteration of the working group and council+election, and we are also using an old staking module for the proposal system and content working group, as well as a totally different staking system in the existing council.

# Analysis

## How to stake

One important question is just how to stake. If you can deploy stake from any account, and you are staking in a context associated with membership, then we have to settle on some consistent way of authenticating for both (the account and the membership) simultaneously. There are basically two approaches:

1. For an account to be used for staking, one must bind it to a membership upfront in a separate step. An account could only be bound to one member at any given time. In order to avoid an attack where the stake is unbinded, and then rebinded to a different member and then staking is initiated again, in an attempt to double count these tokens, the staking module must keep track of what accounts have been used for active staking at any given time, and reject reuse of the same account.

2. It becomes a two-step process, where the member first does one step to register the intent to stake in one extrinsic, and then does a second extrinsic with a signed origin corresponding to the staking account, where they actually stake. It need not be two actual transactions, it can be made doing one transaction using the new `Utility` Pallet and its batching call.

## How to avoid stake reuse

For example, how to prevent funds in an account locked for the purposes of a proposal to be reused for working group activities?

The simplest possible approach appears to be that each module uses a specific lock ID at all times, which should be configurable by the runtime. The runtime is responsible for ensuring uniqueness across all modules. Then, whenever staking is an attempt in a module, it asks the runtime whether it can use the given account for staking. The runtime implements the extremely simple rule of checking all locks on the account, which it can do since it directly observes the implementation of the `balances` module, and only signal OK to the module if there is a valid combination attempted.

So for example, if staking for voting is being attempted, then the runtime will always say OK. However, if staking for a role in working group X is being attempted, then the runtime will only say OK if the only other lock is for voting. It can tell what purpose each lock is for since it assigned them to each module.

bedeho:
# Conclusion

## How to stake

We go for 1.

[...]

Issue #1157, opened 2020-08-12
Idea: Proposal vote reasoning


mochet:
Currently a proposal is made, and discussion takes place either within the proposal itself or elsewhere. Council members can vote to approve, deny, slash or abstain.

In the distant future, when an interested user is looking through hundreds of prior proposals, they would basically have to read through many pages of discussion, possibly in multiple areas of the platform in order to understand why a proposal failed or passed. This still would not directly answer why a council member voted a certain way unless that council member specifically created a post explaining why they voted the way they did.

I would suggest that a field (~120-500 characters) is added to the vote option which allows for a council member to fill in a brief reasoning for their vote. Having this would improve readability of the voting process.

The reasoning could be hidden by a collapse/expand system for each proposal in the proposal overview and for things like council reports, the vote reasoning could be used directly.
This also means that possibly a system is implemented where a user can see the voting history of a specific council member, and quickly digest what kind of reasoning they have employed while a member of the council.

bedeho:
This is brilliant! This is easy and a must-have!

mochet:
Thanks.
One other use of this system would be for council members and users to establish the reputation more clearly of users who have a history of very successful or very unsuccessful proposals.
example: A user has submitted 12 spending proposals in recent history, the first 3 of these were successful and the subsequent 9 were rejected or slashed. A user would be able to see all of these proposals with council member reasoning and quickly be able to digest that council members began rejecting this users proposals because they can rapidly see reasoning like _"Voted to reject as user failed to deliver on previous proposals"_ without having to read through likely a combined hundreds of pages of discussion spanning all of these proposals.

bedeho:
Very true indeed, this is exactly the sort of way we should be building our products: making it easy for participants to evaluate the credibility and effectivness of their peers.

mochet:
This issue could perhaps be expanded to include two additional types of reasoning.
Firstly an incoming reasoning, where a user applying to be a council member can state their mission and goals for the particular election cycle they are applying for. We currently have a thread where a potential council member can state their interests in the platform, but this is an overly broad piece of information and won't relate to specific election cycles unless they continuously update their post.
Secondly an outgoing reasoning, where a CM having completed an election cycle, can provide feedback on the election cycle.

I discussed this a bit here, in relation to how we currently solicit feedback from council members manually, but if this were included in the election process it would hopefully be a great way to standardize the input of this information and this would hopefully mean you could create almost entirely automated council reports.
[...]

PR #1251, opened 2020-09-07
Council rework council

ondratra:
Rework of council according to #896.


ondratra:
The lifecycle of the council is implemented and ready to review.

Things to discuss. Currently:
- there is no way to select initial council members; module starts in announcement stage. Is that ok or should there be possibility for runtime to select initial members?
- passing of election results from referendum to council module is made via runtime "bridge". (see `process_results(...)` in `mocks.rs`)
  Runtime function `process_results(...)` is invoked from referendum and it should call council's `receive_referendum_results(...)`. Maybe this is not ideal way to do this, but I'm not aware of better solution now.
- `... An announced candidate also gets to provide a rich candidate profile, which presumably matches some schema following social convention, however this is not stored. ...` - this is not addressed. Should we simply allow saving of arbitrary hash?
- candidacy lock can be released when user requests to do so during *any* stage if he is neither candidate or council member. Could this approach possibly replace the proposed even/odd cycle locks system?

Things I'm going to implement now:
- `and all announce candidates are removed, and all candidate locks just set or updated in this period are removed.` - this cleanup after announcement is not done, right now announcement period get's prolonged instead of reset
- membership feature is missing; account_id based interactions for now
- more tests - right now only one test for the whole lifecycle exist

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+    stake: Balance,
+}
+
+/////////////////// Type aliases ///////////////////////////////////////////////
+
+// `Ez` prefix in some of the following type aliases means *easy* and is meant to create unique short names
+// aliasing existing structs and enums
+
+// Alias for referendum's storage.
+pub(crate) type Referendum<T> = referendum::Module<T, ReferendumInstance>;
```
bedeho: We should not bind statically to module like this, we should just use a trait that exposes the handling of voting-related logic, similar to how we do not bind to `balances`, but to `Currency` trait, this helps with unit testing also.

bedeho: Was there not some trait for this already even? it should just be on the runtime trait.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        pub fn release_candidacy_stake(origin) -> Result<(), Error<T>> {
+            let account_id = EnsureChecks::<T>::can_release_candidacy_stake(origin)?;
+
+            // update state
+            Mutations::<T>::release_candidacy_stake(&account_id);
+
+            Ok(())
+        }
+
+        /////////////////// Referendum Wrap ////////////////////////////////////
```
bedeho: Why are these wrappers here? These are already extrinsics in the referendum module?

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        /////////////////// Lifetime ///////////////////////////////////////////
+
+        // No origin so this is a priviledged call
+        fn on_finalize(now: T::BlockNumber) {
+            Self::try_progress_stage(now);
+        }
+
+        /// Subscribe candidate
+        #[weight = 10_000_000]
+        pub fn candidate(origin, stake: Balance<T>) -> Result<(), Error<T>> {
```
bedeho: All methods should be verbs.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+            // ensure action can be started
```
bedeho, line 337: Even if you include membership here, is the plan to require council candidacy to have staking only from one account? This is a big limitation, e.g. I may be doing nomination, proposals, and many other things. I should be able to redeploy all of that. I think you need to be able to announce with a vector of accounts at least.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+
+/// Possible council states.
+#[derive(Encode, Decode, PartialEq, Eq, Debug)]
+pub enum CouncilStage<AccountId, Balance> {
+    /// Candidacy announcement period.
+    Announcing(CouncilStageAnnouncing<AccountId, Balance>),
+    /// Election of the new council.
+    Election(CouncilStageElection<AccountId, Balance>),
+    /// The idle phase - no new council election is running now.
+    Idle,
```
bedeho, line 89: I think we should be a bit more opinionated here and reject any attempt to provide a new council more than once per cycle. With this single `Idle` state you will keep accepting new councils over and over, despite no election having been triggered. This may conceal a badly written voting engine. If I am correct about this, then we should have a separate stage `WaitingForNewCouncil`.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
+use system::{ensure_root, ensure_signed, RawOrigin};
+
+use referendum::Instance as ReferendumInstanceGeneric;
+use referendum::ReferendumManager;
+use referendum::Trait as ReferendumTrait;
+
+// declared modules
+mod mock;
```
bedeho, line 66: Why are mock and test modules here?

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+// aliasing existing structs and enums
+
+// Alias for referendum's storage.
+pub(crate) type Referendum<T> = referendum::Module<T, ReferendumInstance>;
+
+pub type Balance<T> = <<T as referendum::Trait<ReferendumInstance>>::Currency as Currency<
+    <T as system::Trait>::AccountId,
+>>::Balance;
+
+pub type EzCandidate<T> = Candidate<<T as system::Trait>::AccountId, Balance<T>>;
```
bedeho: We use the pattern `_Of<T>` for these kinds of aliases, so e.g. `BalanceOf<T>` or `CandidateOf<T>`.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        // No origin so this is a priviledged call
+        fn on_finalize(now: T::BlockNumber) {
+            Self::try_progress_stage(now);
+        }
+
+        /// Subscribe candidate
+        #[weight = 10_000_000]
+        pub fn candidate(origin, stake: Balance<T>) -> Result<(), Error<T>> {
+            // ensure action can be started
+            let (stage_data, candidate) = EnsureChecks::<T>::can_candidate(origin, &stake)?;
```
bedeho: There is something wrong here because you are establishing whether someone can become a valid candidate before doing any comparison on the total stake they are offering compared to possibly a full list of existing candidates. I see you are doing some involved stuff in `Mutate::candidate` that does compare, but at this point, it has to bee too late. What is going on there?

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        #[weight = 10_000_000]
+        pub fn recieve_referendum_results(origin, all_options_results: Vec<T::VotePower>) -> Result<(), Error<T>> {
+            ensure_root(origin)?;
+
+            // ensure this method was called during election stage
+            let stage_data = match Stage::<T>::get().stage {
+                CouncilStage::Election(stage_data) => stage_data,
+                _ => return Err(Error::InvalidRuntimeImplementation),
+            };
+
```
bedeho: Mutation safe marker missing.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+            // emit event
+            Self::deposit_event(RawEvent::NewCandidate(candidate));
+
+            Ok(())
+        }
+
+        // TODO: reconsider this function. It is meant as temporary solution for recieving referendum results via runtime bridge
+        /// Process candidates' results recieved from the referendum.
+        #[weight = 10_000_000]
+        pub fn recieve_referendum_results(origin, all_options_results: Vec<T::VotePower>) -> Result<(), Error<T>> {
```
bedeho: I am not sure if this should really be an extrinsic, it should only ever be called from inside the runtime. I know we can hook into this with the root origin, but since we don't want to use SUDO or other stuff I wonder if it just causes confusion. Just having the module implement a trait like `ElectioinResultSink` or something, where it can accept the result via a function, and then have the runtime itself connect the call from referendum into a forwarded call to that function.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+            // conclude election
+            Self::end_election_period(stage_data, all_options_results.as_slice());
+
+            Ok(())
+        }
+
+        #[weight = 10_000_000]
+        pub fn release_candidacy_stake(origin) -> Result<(), Error<T>> {
+            let account_id = EnsureChecks::<T>::can_release_candidacy_stake(origin)?;
+
```
bedeho, line 363: Mutation safe marker.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+    type LockId: Get<LockIdentifier>;
+
+    /// Duration of annoncing period
+    type AnnouncingPeriodDuration: Get<Self::BlockNumber>;
+    /// Duration of idle period
+    type IdlePeriodDuration: Get<Self::BlockNumber>;
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Referendum {
```
bedeho: Why does this have name `as Referendum` here, should it not be `as Council` ?

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+mod mock;
+mod tests;
+
+/////////////////// Data Structures ////////////////////////////////////////////
+
+/// Information about council's current state and when it changed the last time.
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct CouncilStageInfo<AccountId, Currency, BlockNumber> {
+    stage: CouncilStage<AccountId, Currency>,
+    changed_at: BlockNumber,
```
bedeho, line 78: What does this hold if the council stage could not advance/change from announcing to voting? is it still update? if so,then the name has incorrect semantic.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+
+// declared modules
+mod mock;
+mod tests;
+
+/////////////////// Data Structures ////////////////////////////////////////////
+
+/// Information about council's current state and when it changed the last time.
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct CouncilStageInfo<AccountId, Currency, BlockNumber> {
```
bedeho: Is there a better name here? `*Info` vs `*` names have no good meaningful difference.

bedeho (changes requested):
This looks like a good start, I have left reasonably detailed review comments, and also pointed out some missing features that you were not aware of.

# Missing

- Periodic rewards paid out to the active council. Payouts should keep occurring so long as the council is in session, even after multiple failed elections for a new council.

- A budget out of which all spending proposals will be financed. The budge is reset automatically periodically to give replenished resources.

- A council member can provide an "exit rationale" : https://github.com/Joystream/joystream/issues/1157#issuecomment-679119760

# Answers

> there is no way to select initial council members; module starts in the announcement stage. Is that ok or should there be a possibility for the runtime to select initial members?

Let's just have the stage be configurable for now, that should do.

> ... An announced candidate also gets to provide a rich candidate profile, which presumably matches some schema following social convention, however this is not stored. ... - this is not addressed. Should we simply allow saving of arbitrary hash?

Let's do what we did before: so data is submitted as input in the transaction, so its part of chain history, but we store nothing in storage. I think for the forum, storing the hash in storage makes some sense, because it's permanently there (unless the post is deleted), but a candidate will be removed every cycle, so the vast majority of historical profiles cannot benefit from any light client validation over time.

> passing of election results from referendum to council module is made via runtime "bridge"

I am not sure I understood what is going on here, but there is a comment in the review about not using an extrinsic. The routing should happen via the runtime itself, not in the mocking module.

> candidacy lock can be released when user requests to do so during any stage if he is neither candidate or council member. Could this approach possibly replace the proposed even/odd cycle locks system?

If a council member is staking X for the current council and is trying to stake Y for the next, the logic associated with rolling back and managing what the actual stake should be for different paths (announced-but kicked out, announced-but lost, announced-and won). Also, what do you do if Y < X, you cannot really reduce the stake to a lower, possibly near 0 level, for an active council member just because they are interested in the next election. This is also complicated by the fact that the set of accounts used across two consecutive elections need not perfectly overlap.

Having distinct locks for distinct cycles is to me the simplest way for me to get a mental model that convinces me that things are working and that the logic an all the edge cases are relatively simple.

I think what we need to do here is to take a step back and come up with a model that cleanly deals with stake across different accounts, different purposes (voting, candidacy), and elections (current, next). It may be that my mental model is not fully coherent either because I am thinking of the original approach of accounts with fractional staking, which was even more complex.

Please try to come up with a model, one way or the other, that you feel is the simplest one to deal with this and describe it in a comment prior to further implementation. This will likely save us time, and make my review faster.

runtime-modules/council/src/lib.rs
ondratra: Changed to `CouncilStageUpdate` as it (now) only contains information about current stage and when it changed.

runtime-modules/council/src/lib.rs
ondratra: It was overlooked copied code from referendum when initiating this module. Changed to `as Council`.

runtime-modules/council/src/lib.rs
ondratra: Done

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: This is just a standard way of informing rust that we have a code in different files - it has to be there, or tests won't be available.

runtime-modules/council/src/lib.rs
ondratra: This scenario can't happen in the current implementation. The only time when the new council can be appointed is inside of `recieve_referendum_results()` that can only be called during  `Election` stage, but that also changes stage to `Idle`, thus preventing of multiple council appointment in one cycle.

runtime-modules/council/src/lib.rs
ondratra: The word *candidate* is both noun and verb. But I changed it to `announce_candidacy` to better express what it does.

runtime-modules/council/src/lib.rs
ondratra: Added

bedeho:
I don't know what the status of this work is, but many comments have not been addressed, and there are seemingly other critical issues there as well, such as both 

- council members and candidates not being bound to memberships,
- set of council candidates is stored as a storage value
- misplaced mutation marks w.r.t. fallible calls

```Rust
            //
            // == MUTATION SAFE ==
            //

            // call referendum vote extrinsic
            <Referendum<T>>::vote(origin, commitment, balance)?;
```

Please write a little note on the status of the work here.

runtime-modules/council/src/lib.rs
ondratra: After later discussions membership design was simplified. To recapitulate: 
a) not allowing council stake to be reused by anything else, except voting.
b) requiring that only a single account is used to fund the stake, as its no longer required to redeploy existing stake in multiple other accounts in order to stake.
c) not automatically counting council stake towards outcome.
d) allow voting with with stake that is used by a existing council member or candidate council member.

Membership has been implemented accordingly. Point a) should be achieved by proper setting of referendum/council locks in runtime.

runtime-modules/council/src/lib.rs
ondratra: The original idea behind wrappers was to have a place where extra checks can be made before calling a referendum module itself. After some reworks in both referendum and council, it is redundant with an exception being `release_vote_stake()/can_release_vote_stake()`, but that logic can be transferred into the runtime. 

`Referendum wrap` section will be removed.

runtime-modules/council/src/lib.rs
ondratra: Relevant scenarios:
- when not enough candidates announced their candidacy -> restart announcement period
- when not enough council members were voted during voting/reveal period -> restart announcement period
- due to bad runtime implementation referendum's `start_referendum()` can fail during council's `end_announcement_period()`(it contains `.unwrap()` call). But only reason `start_referendum()` can fail is if referendum is already running - that should never happen with proper runtime, because council itself doesn't start referendum outside of expected period. There is currently no recovery from that and I'm not sure if can statically ensure it is implemented well.

ondratra:
Council membership has been added.

What is remaining:
- unlimited candidates
- connect referendum and council with patter `T::Referendum` instead of current `Trait = system::Trait + referendum::Trait<ReferendumInstance>`
- two locks for even/odd election rounds
- minor issues from review discussions above (remove referendum wrappers and `ElectionResultSink`)

Ad displaced mutation mark) All problematic places in referendum wrappers that are going to be removed.

runtime-modules/council/src/lib.rs
ondratra: Extrinsic was removed, and the new trait `ReferendumConnection` was created. It contains 2 functions that need to be called by the runtime on proper places.

runtime-modules/council/src/lib.rs
ondratra: Wrappers have been removed.

runtime-modules/council/src/lib.rs
ondratra: This issue is obsolete now since we allow unlimited candidates now, and it works slightly different.

ondratra:
Short description of the two significant parts:

1) Infinite candidates:
- a user announces candidacy -> added to `Candidates` map
- a candidate receives vote -> `CurrentCycleCandidatesOrder` is updated (it serves as a list of intermediate winners though it is actually a map indexed by a position)
- failed candidate unstakes -> his record is removed from `Candidates`
- successful candidates have their records removed from `Candidates` and `CurrentCycleCandidatesOrder` automatically when set up as new council members

2) Candidacy stake locks:
- user announces candidacy -> stake is locked in candidacy lock
- election ends -> elected members have their candidacy locks unlocked, and new council locks with the same amount staked are established. Old council members have their council stakes released automatically (unless they were reelected).
- failed candidates can unlock their stakes via extrinsic, or they can keep stake locked and reuse it for candidacy in the next cycle

There is no iteration over unconstraint array, and monetary incentives to clear the state are in place.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+        pub Stage get(fn stage) config(): CouncilStageUpdate<T::BlockNumber>;
+
+        /// Current council members
+        pub CouncilMembers get(fn council_members) config(): Vec<CouncilMemberOf<T>>;
+
+        /// Map of all candidates that ever candidated and haven't unstake yet.
+        pub Candidates get(fn candidates) config(): map hasher(blake2_128_concat) T::CouncilUserId => Candidate<T::AccountId, Balance::<T>>;
+
+        /// Order of candidates in current announcement period.
+        pub CurrentCycleCandidatesOrder get(fn current_cycle_candidates_order) config(): map hasher(blake2_128_concat) u64 => T::CouncilUserId;
```
bedeho: As discussed, this should not be needed, as the referrendum module uses membership identifiers as a way to identify the different alternatives itself.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+        /////////////////// Lifetime ///////////////////////////////////////////
+
+        // No origin so this is a priviledged call
+        fn on_finalize(now: T::BlockNumber) {
+            Self::try_progress_stage(now);
+        }
+
+        /// Subscribe candidate
+        #[weight = 10_000_000]
+        pub fn announce_candidacy(origin, council_user_id: T::CouncilUserId, stake: Balance<T>) -> Result<(), Error<T>> {
```
bedeho, line 336: Where in this code path is it ensured that the user is not already a candidate? It appears to just overwrite the candidate, but what does that doe with

bedeho (changes requested):
I made two initial review comments, but these changes will be substantial enough that I am pausing further review, as the `StakingHandler` trait from @shamil-gadelshin needs to be introduced.

runtime-modules/council/src/lib.rs
ondratra: Repeated candidacy is prevented now.

runtime-modules/council/src/lib.rs
ondratra: Removed.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+            T::CandidacyLock::is_member_staking_account(&council_user_id, &account_id),
+            Error::CouncilUserIdNotMatchAccount,
+        );
+
+        Ok(account_id)
+    }
+
+    /////////////////// Action checks //////////////////////////////////////////
+
+    fn can_candidate(
```
bedeho: - Can we rename to something simpler `can_announce_candidacy`

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+                == CurrentAnnouncementCycleId::get()
+        {
+            return Err(Error::CantCandidateTwice);
+        }
+
+        if stake < &T::MinCandidateStake::get() {
+            return Err(Error::CandidacyStakeTooLow);
+        }
+
+        let candidate = Candidate {
```
bedeho: This construction, while useful, should be done outside ensure check, it is not directly related to the check, and can be invoked by caller, either inline or in a separate method.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+        // ensure user's membership
+        let account_id = Self::ensure_user_membership(origin, &council_user_id)?;
+
+        let stage_data = match Stage::<T>::get().stage {
+            CouncilStage::Announcing(stage_data) => stage_data,
+            _ => return Err(Error::CantCandidateNow),
+        };
+
+        // prevent user from candidating twice in the same election
+        if Candidates::<T>::contains_key(council_user_id)
```
bedeho, line 707: This implies that you can announce candidacy, even if there is an old one hanging around from a prior election. You are sure this  works fine in terms of managing the old stake and everything else?
Either way, please add something to the comment about this case.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+
+    /// Announce user's candidacy.
+    fn announce_candidacy(
+        stage_data: &CouncilStageAnnouncing,
+        council_user_id: &T::CouncilUserId,
+        candidate: &CandidateOf<T>,
+        stake: &Balance<T>,
+    ) {
+        // insert candidate to candidate registery
+        Candidates::<T>::mutate(council_user_id, |value| *value = candidate.clone());
```
bedeho: Just insert/overwrite, rather than mutate, here, simpler and safer.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+
+        // prepare new stage
+        let new_stage_data = CouncilStageAnnouncing {
+            candidates_count: stage_data.candidates_count + 1,
+        };
+
+        // lock candidacy stake
+        T::CandidacyLock::set_stake(&candidate.account_id, *stake);
```
bedeho: This line breaks the story of what is going on, move it out either before or after `Stage` updating.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+
+        // store new candidacy list
+        Stage::<T>::mutate(|value| {
+            *value = CouncilStageUpdate {
+                stage: CouncilStage::Announcing(new_stage_data),
+                changed_at: value.changed_at, // keep changed_at - stage phase haven't changed
```
bedeho: no need to specify `changed_at`, you can use `...` syntax.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct CouncilStageElection {
+    candidates_count: u64,
+}
+
+/// Candidate representation.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default, Clone)]
+pub struct Candidate<AccountId, Balance> {
+    account_id: AccountId,
```
bedeho, line 116: We identify candidates with a member, not an account.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+    cycle_id: u64,
+    order_index: u64,
+    stake: Balance,
+}
+
+/// Council member representation.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, PartialEq, Eq, Debug, Default, Clone)]
+pub struct CouncilMember<AccountId, CouncilUserId, Balance> {
+    account_id: AccountId,
```
bedeho, line 125: We identify council members with a member, not an account.

runtime-modules/council/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+//! - [announce_candidacy](./struct.Module.html#method.announce_candidacy)
+//! - [release_candidacy_stake](./struct.Module.html#method.release_candidacy_stake)
+//!
+//! ## Important functions
+//! These functions have to be called by the runtime for the council to work properly.
+//! - [recieve_referendum_results](./trait.ReferendumConnection.html#method.recieve_referendum_results)
+//! - [can_release_vote_stake](./trait.ReferendumConnection.html#method.can_release_vote_stake)
+//!
+//! ## Dependencies:
+//! - [proposals engine](../referendum/index.html)
```
bedeho: referendum, not proposals engine, right?

bedeho (changes requested):

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: Changed on multiple places.

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: Comment fixed

runtime-modules/council/src/lib.rs
ondratra: Operation has been moved to the end of the function.

runtime-modules/council/src/lib.rs
ondratra: Old stakes can be reused - even current StakingHandler expects such behavior. I added extra commentary.

runtime-modules/council/src/lib.rs
ondratra: We actually identification Candidates (and elected members) by `council_user_id`, but we need to know `account_id` as well so we can use it for staking a currency. It is necessary for an account never to switch `council_user_id` id in the runtime, as we discussed. I've added commentary about this to the module's description.

runtime-modules/council/src/lib.rs
ondratra: See the reply in discussion above this one.

ondratra:
This PR also fixes #1499 and #1201 .

bedeho (approved):
LGTM! Great.

PR #1258, opened 2020-09-08
Fix processing block events in one db transaction

metmirr:
This PR fixes; processing block events in one database transaction and replace async foreach function defination with for loop.


dzhelezov (approved):
Looks good!

bedeho (approved):

PR #1259, opened 2020-09-08
Cont directory debug trait implementation

iorveth:
This PR aims to remove `Debug` Trait inheritance from main module Trait and `ActorAuthenticator` Trait


shamil-gadelshin (changes requested):
I'm not sure why do you need to implement the `Debug` trait for some entities: `CreateEntityOperation`,` UpdatePropertyValuesOperation`, etc. 

I tried to remove those and I have successfully run both `cargo test` and `cargo build --release`.

Parity substrate doesn't generally use that in their pallets. But when they implement the `Debug` they implement it with `std` feature condition:

```
impl sp_std::fmt::Debug for AccountId32 {
	#[cfg(feature = "std")]
	fn fmt(&self, f: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
		let s = self.to_ss58check();
		write!(f, "{} ({}...)", crate::hexdisplay::HexDisplay::from(&self.0), &s[0..8])
	}

	#[cfg(not(feature = "std"))]
	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
		Ok(())
	}
}
```

iorveth:
Good points!

shamil-gadelshin (approved):
LGTM

bedeho (approved):

PR #1279, opened 2020-09-10
runtime: Update the chain constants.

shamil-gadelshin:
Changes:
- SlashDeferDuration changed to the BONDING_DURATION - 1
- Staking history_depth was increased to 336

### DIFF
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -260,6 +260,7 @@ pub fn testnet_genesis(
                 .collect(),
             invulnerables: initial_authorities.iter().map(|x| x.0.clone()).collect(),
             slash_reward_fraction: Perbill::from_percent(10),
+            history_depth: 336,
             ..Default::default()
         }),
         pallet_sudo: Some(SudoConfig { key: root_key }),
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -296,8 +296,8 @@ parameter_types! {
 
 parameter_types! {
     pub const SessionsPerEra: sp_staking::SessionIndex = 6;
-    pub const BondingDuration: pallet_staking::EraIndex = 24;
-    pub const SlashDeferDuration: pallet_staking::EraIndex = 6; // 1/4 the bonding duration.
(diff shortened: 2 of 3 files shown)


node/src/chain_spec/mod.rs
```diff
@@ -260,6 +260,7 @@ pub fn testnet_genesis(
                 .collect(),
             invulnerables: initial_authorities.iter().map(|x| x.0.clone()).collect(),
             slash_reward_fraction: Perbill::from_percent(10),
+            history_depth: 336,
```
bwhm, line 263: Where did 84 come from? Was it just some default value?

shamil-gadelshin: Yes.

bwhm (approved):

runtime/src/lib.rs
```diff
@@ -296,8 +296,8 @@ parameter_types! {
 
 parameter_types! {
     pub const SessionsPerEra: sp_staking::SessionIndex = 6;
-    pub const BondingDuration: pallet_staking::EraIndex = 24;
-    pub const SlashDeferDuration: pallet_staking::EraIndex = 6; // 1/4 the bonding duration.
+    pub const BondingDuration: pallet_staking::EraIndex = BONDING_DURATION;
+    pub const SlashDeferDuration: pallet_staking::EraIndex = BONDING_DURATION - 1; // 'slightly less' than the bonding duration.
```
mnaamani, line 300: If we ever (mistakenly) set BONDING_DURATION to 0 we might get problems (compiler would catch attempting to set a negative number).
, or even setting it to 1 then `SlashDeferDuration` = 0.

So maybe setting `pub const Slash_Defer_Duration = 23` then `BondingDuration` = `Slash_Defer_Duration + 1`

mnaamani:
So I only had a small comment, but otherwise the numbers look good.

runtime/src/lib.rs
shamil-gadelshin: The compiler prevents it with the `arithmetic operation will overflow` message.

mnaamani (approved):

PR #1280, opened 2020-09-10
Working team pallet. Hiring flow.

shamil-gadelshin:
Comments:
- hiring flow was simplified: add_opening, apply_on_opening, fill_opening.
- after filling the opening the related state is deleted (opening, applications)
- no rewards, stakes yet
- the leader concept was implemented.
- there are tests for hiring flow & leadership


runtime-modules/working-team/src/checks.rs
```diff
@@ -0,0 +1,160 @@
...
+use super::Error;
+use frame_support::dispatch::{DispatchError, DispatchResult};
+use frame_support::{ensure, StorageMap, StorageValue};
+use sp_std::collections::btree_set::BTreeSet;
+use system::{ensure_root, ensure_signed};
+
+use crate::types::ApplicationInfo;
+
+// Check opening: verifies opening description length.
+pub(crate) fn ensure_opening_description_is_valid<T: Trait<I>, I: Instance>(
```
bedeho: We dont need to limit this, as we are no longer storing this.

runtime-modules/working-team/src/checks.rs
```diff
@@ -0,0 +1,160 @@
...
+        Error::<T, I>::OpeningDoesNotExist
+    );
+
+    let opening = <crate::OpeningById<T, I>>::get(opening_id);
+
+    Ok(opening)
+}
+
+// Check application: verifies application description length.
+pub(crate) fn ensure_application_description_is_valid<T: Trait<I>, I: Instance>(
```
bedeho: Same here.

runtime-modules/working-team/src/checks.rs
```diff
@@ -0,0 +1,160 @@
...
+        .ensure_valid(
+            text.len(),
+            Error::<T, I>::JobApplicationDescriptionTooShort.into(),
+            Error::<T, I>::JobApplicationDescriptionTooLong.into(),
+        )
+        .map_err(|e| DispatchError::Other(e))
+}
+
+// Check application: ensures that nobody applies for the same opening twice.
+pub(crate) fn ensure_member_has_no_active_application_on_opening<T: Trait<I>, I: Instance>(
```
bedeho: We are no longer rationing access to apply based on the application stake. If we still insist on having a limit, then an opening can be attacked by having someone just fill up all spots, excluding anyone else. The solution is that we should have no limit, nor do we actually need one for any reason. This means all applicants live in a separate map, not as elements in an inline set in the opening. If we want to maintain this rule of applying only once per member to each opening, then a double map `OpeningIdxMemberId -> Application` could be sufficient, but I am not sure its even worth enforcing this rule. In some cases, you cuold even want to ahve one member having multiple worker roles in one group, so why even bother with trying to prevent applying more than once in one opening?

However we store the applications, they are cleaned out of storage as people unstake the role stake when they lost, or if they are winners. The only edge case is if there is no stake for the role, in that case it would live in storage forever. To avoid this would could hard code a lower bound > tx fee of cleanup.

We no longer need to iterate all applications, as we are not managing their stake. This means we no longer should attempt to limit the number of applicants. This means the number of applicants should be unconstrained, i.e.

runtime-modules/working-team/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+        /// Maps identifier to job opening.
+        pub OpeningById get(fn opening_by_id): map hasher(blake2_128_concat)
+            T::OpeningId => JobOpening<T::BlockNumber, T::ApplicationId>;
+
+        /// Count of active workers.
+        pub ActiveWorkerCount get(fn active_worker_count): u32;
+
+        /// Opening human readable text length limits.
+        pub OpeningDescriptionTextLimit get(fn opening_description_text_limit):
```
bedeho: Not needed, see above.

runtime-modules/working-team/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+                /// Maps identifier to worker application on opening.
+        pub ApplicationById get(fn application_by_id) : map hasher(blake2_128_concat)
+            T::ApplicationId => JobApplication<T, I>;
+
+        /// Next identifier value for new worker application.
+        pub NextApplicationId get(fn next_application_id) : T::ApplicationId;
+
+        /// Job application description text length limits.
+        pub ApplicationDescriptionTextLimit get(fn application_description_text_limit):
```
bedeho: Not needed, see above.

bedeho (approved):
Very nice start ;) and an excellent level of abstraction for first PR, fantastic judgement.

# Edit

I actually meant to request changes, not approve, see comments.

runtime-modules/working-team/src/checks.rs
shamil-gadelshin: Removed.

runtime-modules/working-team/src/checks.rs
shamil-gadelshin: Removed.

runtime-modules/working-team/src/checks.rs
shamil-gadelshin: Removed.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Removed.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Removed.

bedeho (approved):
Looks great!

PR #1298, opened 2020-09-10
referendum - reintroduce multiple winners

ondratra:
This PR adds the possibility to provide an unlimited amount of options for the referendum and set an expected number of winners. Options can be restricted via runtime function `is_valid_option_id(...)`.

### DIFF
--- a/runtime-modules/referendum/src/lib.rs
+++ b/runtime-modules/referendum/src/lib.rs
@@ -155,7 +155,7 @@ pub trait Trait<I: Instance>: system::Trait /* + ReferendumManager<Self, I>*/ {
 
     /// Checks if user can unlock his stake from the given vote.
     /// Gives runtime an ability to penalize user for not revealing stake, etc.
-    fn can_unstake(vote: &CastVote<Self::Hash, Balance<Self, I>>) -> bool;
+    fn can_release_voting_stake(vote: &CastVote<Self::Hash, Balance<Self, I>>) -> bool;
 
     /// Gives runtime an ability to react on referendum result.
     fn process_results(winners: &[OptionResult<Self::VotePower>]);
@@ -251,8 +251,11 @@ decl_error! {
         /// Trying to reveal vote that was not cast
         VoteNotExisting,
 
+        /// Trying to vote multiple time in the same cycle
+        AlreadyVotedThisCycle,
+
         /// Invalid time to release the locked stake
-        InvalidTimeToRelease,
+        UnstakingVoteInSameCycle,
 
         /// Salt is too long
         SaltTooLong,
@@ -685,6 +688,23 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
     }
 
     fn can_vote(origin: T::Origin, stake: &Balance<T, I>) -> Result<T::AccountId, Error<T, I>> {
+        fn get_existing_stake<T: Trait<I>, I: Instance>(
+            account_id: &T::AccountId,
+        ) -> Result<Balance<T, I>, Error<T, I>> {
+            if !Votes::<T, I>::contains_key(&account_id) {
+                return Ok(0.into());
+            }
+
+            let existing_vote = Votes::<T, I>::get(&account_id);
+
+            // don't allow repeated vote
+            if existing_vote.cycle_id == CurrentCycleId::<I>::get() {
+                return Err(Error::<T, I>::AlreadyVotedThisCycle);
+            }
+
+            Ok(existing_vote.stake)
+        }
+
         // ensure superuser requested action
         let account_id = Self::ensure_regular_user(origin)?;
 
@@ -696,13 +716,18 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
             _ => return Err(Error::ReferendumNotRunning),
         };
 
+        // get current lock balance if still staking for some past cycle
+        let existing_stake: Balance<T, I> = get_existing_stake(&account_id)?;
+
         // ensure stake is enough for voting
         if stake < &T::MinimumStake::get() {
             return Err(Error::InsufficientStake);
         }
 
+        let stake_diff = *stake - existing_stake;
+
         // ensure account can lock the stake
-        if T::Currency::total_balance(&account_id) < *stake {
+        if T::Currency::total_balance(&account_id) < stake_diff {
             return Err(Error::InsufficientBalanceToStakeCurrency);
         }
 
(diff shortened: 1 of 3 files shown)


runtime-modules/referendum/src/lib.rs
```diff
@@ -46,15 +47,24 @@ impl<BlockNumber, VotePower> Default for ReferendumStage<BlockNumber, VotePower>
...
 /// Representation for revealing stage state.
 #[derive(Encode, Decode, PartialEq, Eq, Debug, Default)]
 pub struct ReferendumStageRevealing<BlockNumber, VotePower> {
-    started: BlockNumber,                 // block in which referendum started
-    intermediate_results: Vec<VotePower>, // votes that options have recieved at a given time
+    started: BlockNumber,      // block in which referendum started
+    winning_target_count: u64, // target number of winners
+    intermediate_winners: Vec<OptionResult<VotePower>>, // intermediate winning options
+    //intermediate_results: HashMap<u64, VotePower>, // votes that options have recieved at a given time
+    intermediate_results: BTreeMap<u64, VotePower>, // votes that options have recieved at a given time
```
bedeho: I am assuming this holds the amount of staking backing a candidate?

If so, then we cannot represent it like this safely for two reasons
- the number of alternatives, i.e. mappings in this map, is unconstrained. That is in fact the whole point of the change we are making:  not to cap. the number of alternatives. Storing it as a map inline like this will be an attack vector.
- the number of votes per alternative is unconstrained.

bedeho: Please let me know what you think could be an alternative approach prior to staring implementing improvement.

bedeho (changes requested):
There is one major issue here, please have a look.

I was a bit surprised that this reintroduces a basic problem we keep trying to avoid, please check in with me if you feel you are not totally on top of why this is a problem ;)

runtime-modules/referendum/src/lib.rs
ondratra: To clarify: 
- `intermediate_winners` is a vector of length less than or equal `winning_target_count` (`0 <= intermediate_winners.len() <= winning_target_count`) representing ordered list of currently winning options. Since `winning_target_count` is parameter received at the start of referendum, it is up to calling module (e.g. council module) to constrain this value.
- `intermediate_results` is a map `option_id -> sum_of_votes_for_option`; possible `option_id`s can be restricted via `is_valid_option_id(...)`. The values are needed for the time when a user is voting for an option currently not listed among winners so that new total votes for the option can be compared to the current winners. This map never needs to be iterated over in referendum; thus, it shouldn't be an attack vector (or am I wrong here?). If it needs to be iterated in the council module (or others), it is up to that module to constraint the number of options.

bedeho: - `intermediate_winners`: yes, I have no problem with this, I was only commenting on `intermediate_results`.
- `intermediate_results`: yes, despite not being explicitly iterated by you, a BTreeMap is not a normal storage map. While it is indeed a map, it is encoded as a single value in the underlying database. This means _any_ attempt to read, update, count, or do anything with a btreemap involves reading all mappings from the database. This is not the case with a storage map, which is a genuine map in the underlying database model

bedeho: Fixed in https://github.com/Joystream/joystream/pull/1298/commits/212fb6711ba48cdc72d5bd9d275bafcca9d4f0ab.

runtime-modules/referendum/src/lib.rs
```diff
@@ -504,32 +520,127 @@ impl<T: Trait<I>, I: Instance> Mutations<T, I> {
...
+                &current_winners[0..target_index],
+                &tmp[..],
+                &current_winners[target_index..current_index],
+                &current_winners[current_index + 1..current_winners.len()],
+            ]
+            .concat()
+        }
+
+        /// Tries to insert option to the winners list
+        fn try_winner_insert<T: Trait<I>, I: Instance>(
```
bedeho: The overall approach here, seems like it has gotten very complex. Do you think there is room to simplify this?

The logic should be quite simple
1. Iterate list to find whether winner is on it, if it is, add to voting power in vector. If it was top of list, quit, otherwise remember index and go to 3.
2. If it wasn't on list, then just compare its value with lowest ranked current winner, if its less, quit, otherwise remember index and go to 3.
3. Whatever index you have, keep comparing it to what is above, and swap places as long as you can (just as in bubble sort).

This should be quite simpler than what you currently have, but correct me if not.

bedeho (changes requested):
1. Update `can_vote` to prevent updating a prior vote.
2. Refactor `can_release_stake`to
  - Remove exception in unstaking extrinsic that allow removal of vote from same cycle if in voting period.
  - Make `if cycle_id < cast_vote.cycle_id` check pure equality check, and update error name.
  - Make `!T::can_unstake` call last, and rename to reflect that it really is a gatekeeper for whether voting stake has passed any unbonding period.
  - Improve comments describing each check.

runtime-modules/referendum/src/lib.rs
ondratra: Yes this is likely less complex / better to read algorithm. Done.

runtime-modules/referendum/src/lib.rs
```diff
@@ -685,6 +688,23 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
     }
 
     fn can_vote(origin: T::Origin, stake: &Balance<T, I>) -> Result<T::AccountId, Error<T, I>> {
+        fn get_existing_stake<T: Trait<I>, I: Instance>(
```
bedeho: I don't understand what is going on here.

Why are we returning the stake of a vote that was for a old cycle? Why do we care about that?
Isn't `can_vote` just supposed to say `yes` if the vote does not exist, or is from an old cycle, and `no` otherwise? This is what `Mutations::vote` does.
If it returned `yes`, then the voting should just overwrite the vote and the lock, no?

bedeho (changes requested):

runtime-modules/referendum/src/lib.rs
```diff
@@ -587,56 +548,76 @@ impl<T: Trait<I>, I: Instance> Mutations<T, I> {
...
-            // option is already in the list and only needs to change it's position?
-            if let Some(current_index) = already_existing_index {
-                return Some(move_winner::<T, I>(
-                    current_winners,
-                    current_index,
-                    insert_index,
-                    option_result.vote_power,
-                ));
-            }
+            let mut new_winners = current_winners.to_vec();
```
bedeho: - This def. looks simpler, I think it could be a bit safer.
  1. `already_existing_index` can be immutable, and in fact can be found using `iter::find`, so no need to do it yourself even.
  2. try a bit more descriptive names. `new_virtual_index` does not help the reader much, even `already_existing_index` could be improved to `current_winners_index_of_vote_recipient`.
  3. The entire block from 554->592 is doing a lot, it computes a value, has side-effects and even has path that exists the method entirely. It takes a lot of time to get convinced that it works. Is it possible to do this in a few more steps that have a bit narrower purpose.

- Can we move out of `reveal_vote`, really obscures readability when an internal utility routine gets this long. Make it a standalone non-public function instead.

bedeho (changes requested):

runtime-modules/referendum/src/lib.rs
ondratra: Code has been flattened and simplified.

runtime-modules/referendum/src/lib.rs
ondratra: I mistakenly thought about `total_balance()` as if it was `free_balance()`; subtracting existing locked stake value from this ~~free_balance~~. It has been fixed - no need to retrieve existing stake now.

bedeho (approved):
Excellent!

PR #1299, opened 2020-09-11
Fix uncle rewards.

shamil-gadelshin:
Our current network configuration has unexpected side-effect. In the list of validator who get reward-points for participating in network activities we have special account `5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM` in each era.

Example of rewards points for an era:

```
{
  total: 12,198,
  individual: {
    5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM: 66,
    5D5E3egY12qYrWP7tiAQHYTBzTVfUfFeV68EsKYSQz9FWadN: 6,236,
    5Gmi8oxKM4oV18RmUDyL7HVg494zNDdCE2HGAA12Vewp6ykw: 5,896
  }
```
}

Also neither Kusama nor Polkadot networks have same effects. In those networks all validators got 20xN points where N is blocks for a validator.

This behavior could affect validator payouts but appears to not cause any problems in the network itself. 

After an investigation we found out that `5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM` is '0x0' account (zero) and it got rewarded by uncle verification only.  

0x0 is a outcome of using `Default::default()` for account in case of any errors.

in some cases, 0x0 is because the node cannot determine an author() of the uncle block, which in turn is set by ImOnline EventHandler implementation.

Further investigation shows that both Kusama and Polkadot have a different configuration. 

```
parameter_types! {
	pub const UncleGenerations: u32 = 0;
}

// TODO: substrate#2986 implement this properly
impl authorship::Trait for Runtime {
	type FindAuthor = session::FindAccountFromAuthorIndex<Self, Babe>;
	type UncleGenerations = UncleGenerations;
	type FilterUncle = ();
	type EventHandler = (Staking, ImOnline);
}

```
Having set this configuration for our network we eliminate the issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,7 +1993,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
-version = "3.0.1"
+version = "3.1.0"
 dependencies = [
  "frame-benchmarking",
  "frame-benchmarking-cli",
@@ -2053,7 +2053,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node-runtime"
-version = "7.1.0"
+version = "7.2.0"
 dependencies = [
  "frame-benchmarking",
  "frame-executive",
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -3,7 +3,7 @@ authors = ['Joystream contributors']
 build = 'build.rs'
 edition = '2018'
 name = 'joystream-node'
-version = '3.0.1'
+version = '3.1.0'
 default-run = "joystream-node"
 
 [[bin]]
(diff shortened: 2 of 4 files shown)


mnaamani (changes requested):
Looks good.
Just minor request: bump the runtime spec version.

bwhm (approved):
Looks like these changes are in line with the values from [kusama](https://github.com/paritytech/polkadot/blob/master/runtime/kusama/src/lib.rs#L247) and [polkadot](https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L262), so I'm happy :)

mnaamani:
Clearly omitting ImOnline module as an EventHandler for the authorship module was a mistake in the past.

With respect to the `UncleGenerations` which is described as: 

```
	/// The number of blocks back we should accept uncles.
	/// This means that we will deal with uncle-parents that are
	/// `UncleGenerations + 1` before `now`.
```

I can't claim to fully understand the significance of 5 vs 0 so selecting the same value used in polkadot and kusama production networks is fine by me.

mnaamani (approved):

Issue #1257, opened 2020-09-08
WIP: Pioneer issues for Alexandria


bwhm:
As I think there will be more items here, I'm marking this as WIP for now.
New entries will be added as replies. I've tried to assign most things to @DzhideX, but @mnaamani may be needed if they are more difficult than they appear...

**Note**
When I say "remove", I mean just commenting them out, so it's easy to put back in once we know what they do, and/or need them :)

## General
When typing in a balance, decimal points are mishandled - e.g. 12.91 returns 9112
- [x] This applies to many different situations so I **hope** it's enough to fix once - @DzhideX
- Confirmed to happen when:
   - balance transfers
   - stake/vote on elections
   - staking for validators

## Chain State
**`balances.account(AccountId): AccountData`**
For any address, this returns:
```
{
  free: 0,
  reserved: 0,
  miscFrozen: 0,
  feeFrozen: 0
}
```
Is it something to do with types?
- [x] Investigate, and fix - @mnaamani 

## Accounts
General comment:
- The top right `?` has some info. Not sure we want to keep this.

**Rename back to `My Keys`**
- [x] This has been called `My Keys` on previous network. I suggest we do it again (without changing the URL) - @DzhideX

### My Accounts
**Remove polkadot-js extension "ad"**
- [ ] Although this apparently works, I think we should remove it for now. - @DzhideX

**Remove `Add via QR`**
- [ ] I haven't tried if it works even, but still, remove - @DzhideX

**Remove `Multisig` and `Proxied`**
- [ ] They are greyed out anyway. Remove for now - @DzhideX

**Passwords**
We have always allowed empty passwords, and even generate some ourselves.
- [x] Allow empty passwords - @DzhideX
- creating accounts
- importing jsons
- exporting jsons

## My Memo
**Move to `My keys / Accounts`**
- [x] Add as a new tab - @DzhideX

## Staking
**Question mark**
- [ ] The top right `?` has some info, and it's not really correct anymore. Remove - @DzhideX

**Rename back to `Validators`**
- [x] This has been called `Validators` on previous network. I suggest we do it again (without changing the URL) - @DzhideX

### Account Actions
**`+ Nominator` and `+ Validator` buttons**
What are these buttons meant to do?
- [ ] Investigate and consider removing - @DzhideX

**`Validate->Set validator preferences`**
- [ ] Value > 100 breaks page - @DzhideX

### Payouts
**All `Payout *` buttons greyed out** 
What are these buttons meant to do? 
[...]

PR #1313, opened 2020-09-13
Runtime: add pallet-utility

mnaamani:
Pioneer Staking Payouts requires the `utility.batch()` dispatchable to do batch payouts.
So this PR adds the runtime module to our configuration.

Background: https://github.com/Joystream/joystream/issues/1257#issuecomment-691617912

It is used here: https://github.com/Joystream/joystream/blob/afd7f8da4c7145e17d96af683d69a2efbf0c9b9c/pioneer/packages/page-staking/src/Payouts/PayButton.tsx#L37

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2053,7 +2053,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node-runtime"
-version = "7.2.0"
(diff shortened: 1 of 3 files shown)


bwhm (approved):
LGTM!

PR #1343, opened 2020-09-14
Colossus/route to ipfs gateway

mnaamani:
Storage Node:

~Route new asset URL `/asset/v1/{contentID}` to the IPFS HTTP Gateway running on the storage node.~
Routing `/asset/v0/{contentId}` path to the IPFS HTTP Gateway running on the storage node.

Currently the way the `/asset/v0/` api works, handles range requests very inefficiently. Colossus opens a readable file stream and must read the stream to the point where it needs to 'seek' to return the requested range of bytes. This is due to a limitation in the streams api and lack of support for seeking in that stream. This was causing the storage node to consume a large amount of memory especially.

So the alternative solution was to just proxy the http requests for accessing content to the IPFS HTTP gateway.

Pioneer, Helios and the storage cli tool was updated to use the new endpoint.

For correct functioning of the proxy middle ware the ipfs gateway needs to be configured to use paths and not subdomains which is the default config.

The recommended configuration when used with colossus can be applied with this command. (daemon should be restarted to take affect):

`ipfs config --json Gateway.PublicGateways '{"localhost": null }' `


shamil-gadelshin (changes requested):
We should keep our architecture consistent, so please, refactor the 'middleware + path rewriting' approach of V1 to the Open API similar to existing V0.

mnaamani:
> We should keep our architecture consistent, so please, refactor the 'middleware + path rewriting' approach of V1 to the Open API similar to existing V0.

Updated to stick with the open api structure.
Also added handling of downloading file as attachement.

We no longer have a dependency on the `utils/ranges` library so it can/should be removed in the future unless we find some other need for it.

shamil-gadelshin (approved):
LGTM

Issue #1301, opened 2020-09-11
Pioneer: Account passwords


bwhm:
ref #1257 

We have always allowed empty passwords, and even generate some ourselves.
Allow empty passwords when:
- creating accounts
- importing jsons
- exporting jsons
- changing passwords for accounts
- others?

PR #1345, opened 2020-09-14
Allow empty passwords.

DzhideX:
Initial changes for issue #1301.

Note:
- This should work for the most part fairly well. Where it falls short is in components that utilize some internal polkadot code. When this happens, one should refresh the page and it should work again.

The code in question: 
```
page-accounts/src/Accounts/modals/Backup.tsx

          const addressKeyring = address && keyring.getPair(address);
          const json = addressKeyring && keyring.backupAccount(addressKeyring, password);
          const blob = new Blob([JSON.stringify(json)], { type: 'application/json; charset=utf-8' });

          FileSaver.saveAs(blob, `${address}.json`);

page-accounts/src/Accounts/modals/ChangePass.tsx

          account.decodePkcs8(oldPass);      
```

If it's ok with @bwhm for the most part, it might be a good idea to have someone else check this out further.

### DIFF
--- a/pioneer/packages/page-accounts/src/Accounts/PasswordInput.tsx
+++ b/pioneer/packages/page-accounts/src/Accounts/PasswordInput.tsx
@@ -4,6 +4,7 @@
 
 import { Modal, Password } from '@polkadot/react-components';
 // import keyring from '@polkadot/ui-keyring';
+import { isPasswordValid } from '@polkadot/joy-utils/functions/accounts';
 
 import React, { useCallback, useState } from 'react';
 
@@ -22,6 +23,8 @@ export default function PasswordInput ({ onChange, onEnter, password }: Props): 
 
   const _onPasswordChange = useCallback(
     (password: string) => {
+      const isPassValid = isPasswordValid(password);
+
       setPassValid(isPassValid);
 
       const isValid = isPassValid && isPass2Valid;
@@ -33,7 +36,7 @@ export default function PasswordInput ({ onChange, onEnter, password }: Props): 
 
   const onPassword2Change = useCallback(
     (password2: string) => {
-      const isPass2Valid = password2 === password;
+      const isPass2Valid = isPasswordValid(password2) && (password2 === password);
 
       setPassword2({ isPass2Valid, password2 });
 
--- a/pioneer/packages/page-accounts/src/Accounts/PasswordInput.tsx
+++ b/pioneer/packages/page-accounts/src/Accounts/PasswordInput.tsx
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { Modal, Password } from '@polkadot/react-components';
-import keyring from '@polkadot/ui-keyring';
+// import keyring from '@polkadot/ui-keyring';
 
 import React, { useCallback, useState } from 'react';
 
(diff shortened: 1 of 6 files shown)


pioneer/packages/page-accounts/src/Accounts/PasswordInput.tsx
```diff
@@ -17,13 +17,11 @@ type Props = {
...
 export default function PasswordInput ({ onChange, onEnter, password }: Props): React.ReactElement {
   const { t } = useTranslation();
-  const [isPassValid, setPassValid] = useState<boolean>(false);
-  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: false, password2: '' });
+  const [isPassValid, setPassValid] = useState<boolean>(true);
+  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: true, password2: '' });
 
   const _onPasswordChange = useCallback(
     (password: string) => {
-      const isPassValid = keyring.isPassValid(password);
```
Lezek123, line 25: Just noticed we have Joystream-specific `isPasswordValid` function inside `joy-utils/src/functions/accounts.ts`.

It may be cleaner/safer to use that one, because:
1. It preserves the `keyring.isPassValid` check (I know that currently it just checks if length > 0, but it's possible that in some future version it will also include some other security checks we may need)
2. This way we can add our custom rules to password validation later and will only have to change this in one place.
3. If we decide to pull from upstream again this should give us a clearer indication of what was the reason behind this change.

pioneer/packages/page-accounts/src/Accounts/modals/Derive.tsx
```diff
@@ -83,8 +83,8 @@ function Derive ({ className = '', from, onClose }: Props): React.ReactElement {
   const [isConfirmationOpen, toggleConfirmation] = useToggle();
   const [{ isLocked, lockedError }, setIsLocked] = useState<LockState>({ isLocked: source.isLocked, lockedError: null });
   const [{ isNameValid, name }, setName] = useState({ isNameValid: false, name: '' });
-  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: false, password: '' });
-  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: false, password2: '' });
+  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: true, password: '' });
+  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: true, password2: '' });
   const [{ isRootValid, rootPass }, setRootPass] = useState({ isRootValid: false, rootPass: '' });
```
Lezek123: This is the root account password, so it needs to be adjusted to work with empty passwords too

pioneer/packages/page-accounts/src/Accounts/modals/Derive.tsx
```diff
@@ -83,9 +84,9 @@ function Derive ({ className = '', from, onClose }: Props): React.ReactElement {
   const [isConfirmationOpen, toggleConfirmation] = useToggle();
   const [{ isLocked, lockedError }, setIsLocked] = useState<LockState>({ isLocked: source.isLocked, lockedError: null });
   const [{ isNameValid, name }, setName] = useState({ isNameValid: false, name: '' });
-  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: false, password: '' });
-  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: false, password2: '' });
-  const [{ isRootValid, rootPass }, setRootPass] = useState({ isRootValid: false, rootPass: '' });
+  const [{ isPassValid, password }, setPassword] = useState({ isPassValid: true, password: '' });
+  const [{ isPass2Valid, password2 }, setPassword2] = useState({ isPass2Valid: true, password2: '' });
+  const [{ isRootValid, rootPass }, setRootPass] = useState({ isRootValid: true, rootPass: '' });
```
Lezek123, line 89: There is also `setRootPass({ isRootValid: !!rootPass, rootPass });` below (line 129)
You can change it to `setRootPass({ isRootValid: isPasswordValid(rootPass), rootPass });` and should be good to go.

Lezek123:
I found an issue with `Derive account via derivation path` yesterday while testing this PR, see: https://github.com/Joystream/joystream/issues/1359.
It wasn't introduced in this PR, but leaving a comment here in case @bwhm runs into it too during testing.

Lezek123 (approved):
Works great, LGTM!

bwhm (approved):

PR #1347, opened 2020-09-14
Integration tests/exported state

mnaamani:


testnets/nicaea-exported-state/chain_spec.json
```diff
@@ -29,7 +29,8 @@
           [
             "5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY",
             5000
-          ]
+          ],
+          ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY", 100000000]
```
mnaamani: This was the reason the tests were failing using the exported state. The chainspec was not build with appropriate arguments to give the account `//Alice` enough funds. We should change the mechanism by which we give newly generated accounts funds. `sudo(balances::setBalance())` is more appropriate. This we we also don't need to estimate ahead of time how much total funds the tests will need.

mnaamani: Any way the chain_spec.json and raw_chain_spec.json should not be added to the repo, but generated from the exported state using the chain-spec-builder while running the tests.

PR #1351, opened 2020-09-15
Atlas channel screen

fcescob:
This PR implements a Channel screen, I removed all elements that were present in the design but that are not in Atlas v1.


atlas/packages/app/src/shared/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -25,6 +33,7 @@ const ChannelPreview: React.FC<ChannelPreviewProps> = ({ name, avatarURL, views,
       nameNode={nameNode}
       metaNode={metaNode}
       animated={animated}
+      onClick={onClick}
```
kdembler, line 36: We need to add `cursor: pointer` when `onClick` is defined

atlas/packages/app/src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -0,0 +1,45 @@
...
+`
+
+export const VideoSectionGrid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(320px, 1fr));
+  grid-gap: ${theme.spacing.xl};
+`
+export const StyledAvatar = styled(Avatar)`
+  max-width: 136px;
+  float: left;
```
kdembler: I'd advise against slipping `float` into our CSS, we can easily achieve the same with flex

kdembler: Also, I think it'd make sense to add something like `align-items: center` so that avatar<->title are centered if the title wraps into multiple lines

atlas/packages/app/src/api/queries/channels.ts
```diff
@@ -18,3 +20,22 @@ export const GET_NEWEST_CHANNELS = gql`
...
+
+export const GET_CHANNEL = gql`
+  query GetChannel($id: ID!) {
+    channel(id: $id) {
+      ...ChannelFields
+    }
+  }
+  ${channelFieldsFragment}
+`
+export const GET_CHANNEL_VIDEOS = gql`
```
kdembler: Some specific reason to run this as a separate query instead of just adding `videos` field to `GET_CHANNEL `?

kdembler (changes requested):
Looks good in general.

Two additional things:
- Same story as with the search bar, shouldn't we use values from the theme for paddings, etc.? Let's make a decision and stick with it
- I guess now that we have `VideosGrid` (my guess is at some point it should probably be used / merged with `InifiniteVideoGrid`), I don't think we need to keep the old `Grid` any longer?

atlas/packages/app/src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -0,0 +1,46 @@
...
+  height: 430px;
+  padding: 0 ${theme.sizes.b8}px;
+`
+export const TitleSection = styled.div`
+  display: flex;
+  align-items: center;
+  padding-top: ${theme.sizes.b10}px;
+`
+export const Title = styled.h1`
+  font-size: ${theme.typography.sizes.h2}px;
```
kdembler: `typography.sizes.h2` already specifies a unit so adding the `px` at the end actually breaks the style. This may be my fault because I introduced `sizes` which are number values and now we get these mixed up. We probably should clean that up soon

kdembler (approved):
LGTM

bedeho (approved):

PR #1352, opened 2020-09-15
Colossus - accountNextIndex() instead of nonce caching

mnaamani:
Use `api.rpc.system.accountNextIndex(accountId)` now in place of old nonce caching mechanism.


shamil-gadelshin (approved):
LGTM

PR #1353, opened 2020-09-15
cli: fix dependency to correct version of polkadot/api package

mnaamani:
In making preparations to publish new npm packages for Alexandria I was doing some double checking.
Fixed the polkadot api package version in joystream-cli

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -15,7 +15,7 @@
     "@oclif/plugin-help": "^2.2.3",
     "@oclif/plugin-not-found": "^1.2.4",
     "@oclif/plugin-warn-if-update-available": "^1.7.0",
-    "@polkadot/api": "^0.96.1",
+    "@polkadot/api": "1.26.1",
     "@types/inquirer": "^6.5.0",
     "@types/proper-lockfile": "^4.1.1",
     "@types/slug": "^0.9.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3361,7 +3361,7 @@
     memoizee "^0.4.14"
     rxjs "^6.6.0"
 
-"@polkadot/api@1.26.1", "@polkadot/api@^0.96.1", "@polkadot/api@^1.26.1":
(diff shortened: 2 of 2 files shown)


mnaamani:
I'm considering changing the versioning schema for types and maybe even the cli packages so the major version number matches the runtime spec version. Currently we are only incrementing the minor version on each release which is semantically incorrect. 

What do you think @Lezek123

Lezek123:
Progress from `0.1.0` to `0.2.0` should be correct for initial development releases, during which there are many breaking changes and backward compatibility may not be preserved. This is, for example, how `@polkadot/api` was intially versioned, so in case of `0.x.y`, `x` would mean something like a "major release" (this is also how it's treated by `npm`, so ie. `^0.1.0` will not catch `0.2.0` but `^1.1.0` will catch `1.2.0`: https://semver.npmjs.com/).

If we decide we're not in "initial development" stage anymore, we could move to `1.0.0` and then increment the first digit after each release. It would be nice to have the major version match runtime spec version, but moving from `0.1.0` to `3.0.0` may also be a little confusing.

Lezek123 (approved):

mnaamani:
> Progress from `0.1.0` to `0.2.0` should be correct for initial development releases, during which there are many breaking changes and backward compatibility may not be preserved. This is, for example, how `@polkadot/api` was intially versioned, so in case of `0.x.y`, `x` would mean something like a "major release" (this is also how it's treated by `npm`, so ie. `^0.1.0` will not catch `0.2.0` but `^1.1.0` will catch `1.2.0`: https://semver.npmjs.com/).
> 
> If we decide we're not in "initial development" stage anymore, we could move to `1.0.0` and then increment the first digit after each release. It would be nice to have the major version match runtime spec version, but moving from `0.1.0` to `3.0.0` may also be a little confusing.

Okay that sounds reasonable. So lets stick with current convention :+1:

Issue #1354, opened 2020-09-15
Pioneer: Allow exploring old (previous) chain proposals after Alexandria release


Lezek123:
The idea is to make all the current testnet proposals (https://testnet.joystream.org/#/proposals) data easily accessible for Pioneer users after the Alexandria release (which requires starting a new chain). Since the old (previous chain) proposals will not be part of this new chain state (migrating them would be quite diffucult), the idea is to fetch all the data expected by the related UI components (`ProposalDetails` etc.) and put it in a JSON file inside Pioneer's codebase. We can then use this data to display the old proposals in a similar fashion as the current ones, but on a separate route (ie. https://testnet.joystream.org/#/proposals/old)

More specifically this can be achieved by:

- Using a transport method (ie. `transport.proposals.proposalsBatch('All', 1, 9999)`) to fetch and parse all proposals data the same way this is currently done by Pioneer
- For each of those proposals using `transport.proposals.votes` to fetch all the votes and `transport.proposals.discussion` to fetch the discussion posts
- Combining those results into an array of objects like: `{ proposal, discussion, votes }`
- Saving it as a json file inside the Pioneer's codebase
- Adjusting the proposals transport so that it can either fetch the data from the chain or use the json file depending on the type of proposals we want to display (old/current)
- Creating a separate tab/route where we only display the old proposals
- Implementing necessary adjustments to components like `ProposalDetails` etc. (ie. disabling block number links for old proposals)

bwhm:
@mnaamani !

Lezek123:
Done in https://github.com/Joystream/joystream/pull/1381

PR #1381, opened 2020-09-16
Pioneer: Exploring old proposals

Lezek123:
Implements https://github.com/Joystream/joystream/issues/1354

In order to re-create `historical-proposals.json` file:
1. Checkout https://github.com/Lezek123/substrate-runtime-joystream/tree/proposals-export-script locally
2. `yarn && yarn workspace pioneer start`
3. Go to http://localhost:3000/#/js (connect to the current testnet node)
4. Select `Export proposals data`
5. Run the script

### DIFF
--- a/pioneer/packages/joy-proposals/src/Proposal/Body.tsx
+++ b/pioneer/packages/joy-proposals/src/Proposal/Body.tsx
@@ -33,6 +33,7 @@ type BodyProps = {
   proposerId: number | MemberId;
   isCancellable: boolean;
   cancellationFee: number;
+  historical?: boolean;
 };
 
 function ProposedAddress (props: { accountId?: AccountId }) {
@@ -97,7 +98,7 @@ class ParsedParam {
 }
 
 // The methods for parsing params by Proposal type.
-const paramParsers: { [k in ProposalType]: (params: SpecificProposalDetails<k>) => ParsedParam[]} = {
+const paramParsers: { [k in ProposalType]: (params: SpecificProposalDetails<k>, historical?: boolean) => ParsedParam[]} = {
   Text: (content) => [
     new ParsedParam(
       'Content',
@@ -237,12 +238,14 @@ const paramParsers: { [k in ProposalType]: (params: SpecificProposalDetails<k>) 
     new ParsedParam('Working group', (group as WorkingGroup).type),
     new ParsedParam('Mint capacity', formatBalance((capacity as Balance)))
   ],
-  BeginReviewWorkingGroupLeaderApplication: ([id, group]) => [
+  BeginReviewWorkingGroupLeaderApplication: ([id, group], historical) => [
     new ParsedParam('Working group', (group as WorkingGroup).type),
     // TODO: Adjust the link to work with multiple groups after working-groups are normalized!
     new ParsedParam(
       'Opening id',
-      <Link to={`/working-groups/opportunities/storageProviders/${id.toString()}`}>#{id.toString()}</Link>
+      historical
+        ? `#${id.toString()}`
+        : <Link to={`/working-groups/opportunities/storageProviders/${id.toString()}`}>#{id.toString()}</Link>
     )
   ],
   FillWorkingGroupLeaderOpening: ({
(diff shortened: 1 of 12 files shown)


bwhm (approved):
Looks great to me!

*this link won't work for long, but:
https://alexandria-testing-4.joystream.app/#/proposals/historical

mnaamani:
Solution looks great.
I am having trouble running the export script though.

In browser console:
```
Uncaught (in promise) TypeError: _transport.default is not a constructor
    _runJs http://localhost:3000/main.00476fde.js:184626
    apply http://localhost:3000/vendor.4e747629.js:20094
    baseInvoke http://localhost:3000/vendor.4e747629.js:21755
    apply
```

Getting some build errors in pioneer:
```
Version 3.7.2
app-js/src/Playground.tsx:30:8 - error TS2613: Module '"/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' has no default export. Did you mean to use 'import { Transport } from "/Users/mokhtar/joystream/joystream/pioneer/packages/joy-utils/src/Transport"' instead?

30 import Transport from '@polkadot/joy-utils/transport';
          ~~~~~~~~~
```

I will try to remove all my node_modules/ folders and try to rebuild. Its probably something local because I see all the CI checks have passed.

edit:

Hmm I think this may be another case of import resolution due to case sensitive issue with the name of the file `joy-utils/src/Transport.tsx` vs `joy-utils/src/transport/` folder.  @Lezek123

mnaamani:
> Looks great to me!
> 
> *this link won't work for long, but:
> https://alexandria-testing-4.joystream.app/#/proposals/historical

Tested it there.. looks great, almost wish the live view was as snippy. 👍 

Although the Query node will help, it might not hurt to use a similar mechanism to display the live content.
Just have the "status-server" do regular exports and serve a live.json which pioneer can fetch.

mnaamani:
Only thing that breaks and maybe be worth fixing, are links to old proposals in a proposal text like `https://testnet.joystream.org/#/proposals/83` which is referenced from the historical proposal: https://alexandria-testing-4.joystream.app/#/proposals/historical/88

Some regex match and replace magic perhaps?

mnaamani:
> Hmm I think this may be another case of import resolution due to case sensitive issue with the name of the file `joy-utils/src/Transport.tsx` vs `joy-utils/src/transport/` folder. @Lezek123

The build issue is on my Mac

mnaamani:
I fixed it by just adding extra `/index` at the end to be more explicit

`import Transport from '@polkadot/joy-utils/transport/index';`

mnaamani:
One suggestion is to perhaps use the `FileSaver` api to download the exported state instead of pasting it to console.

const data = JSON.stringify(allProposalsData, null, 4)
const blob = new Blob([data], { type: 'application/json; charset=utf-8' });
FileSaver.saveAs(blob, `historical-proposals.json`);

I think we do something like this when users apply for a role and we generate a new key for them.

mnaamani (approved):

Lezek123:
`@polkadot/joy-utils/src/Transport.tsx` is no longer part of the repository on `iznik` branch, but I had to place the export script on a branch based on `master`, because of the dependencies like `@polkadot/api`, `@polkadot/types` etc. (and there this file is still present). I didn't notcie this, but actually the Mac OS check indeed caught it: https://github.com/Lezek123/substrate-runtime-joystream/runs/1123661907. Will fix.

> Although the Query node will help, it might not hurt to use a similar mechanism to display the live content.

Maybe we could leverage this somehow, but this would require factoring the `joy-utils` out into `joystream-js` first (then it could be "consumed" by the status server). I'm planning to do this as soon as possible, because this will have a lot of potential benefits.

>Only thing that breaks and maybe be worth fixing, are links to old proposals (...) Some regex match and replace magic perhaps?

Will do

> One suggestion is to perhaps use the `FileSaver` api to download the exported state instead of pasting it to console.

Good idea, will implement.

mnaamani:
Nice https://github.com/Joystream/joystream/pull/1381/commits/3c978576ccd19f5789c961440f602d3ac19fb905
Just wondering if there could also be links in the proposal discussion posts?

Lezek123:
> Just wondering if there could also be links in the proposal discussion posts?

Done (but couldn't find any examples)

Lezek123:
> One suggestion is to perhaps use the FileSaver api to download the exported state instead of pasting it to console.

Implemented on https://github.com/Lezek123/substrate-runtime-joystream/tree/proposals-export-script

mnaamani:
> Implemented on https://github.com/Lezek123/substrate-runtime-joystream/tree/proposals-export-script


Screen Shot 2020-09-17 at 12 53 19 PM

Exporting Works.

And Proposals links updated:
Screen Shot 2020-09-17 at 1 01 18 PM

PR #1385, opened 2020-09-16
Merge content directory second try into babylon

iorveth:


mnaamani:
Shouldn't the old versioned store modules be removed?

```
        VersionedStore: versioned_store::{Module, Call, Storage, Event<T>, Config},
        VersionedStorePermissions: versioned_store_permissions::{Module, Call, Storage, Config<T>},
```

Or at least remove the `Call` so no extrinsics can be dispatched. There is no point really keeping them except for maybe reading storage values with older joystream/types library. But even that is going to be tricky after the runtime upgrade.

I recommend we export the state with https://github.com/Joystream/runtime-state-export-tools/blob/master/src/export_content.ts before updating to Babylon runtime and drop all the old versioned-store/permissions code.

mnaamani:
@bedeho @iorveth Shall we merge this ?

Issue #1383, opened 2020-09-16
Alexandria: Key generation


bwhm:
When we ran the first test, lots of keys were generated in pioneer for each of the testers.
Some were also imported from running testnet.

When we launched the second test, with same url for pioneer, some keys were still "there" but a lot of them were gone. They still exist in local storage, but they are not shown to the user.

Here are the three types of tested:
1. Keys imported from actual running testnet
2. Keys created in pioneer using "add account"
3. Keys created in pioneer automatically by applying for a role (the "role key")

Only the ones from "3." was displayed in pioneer.

I did some research, and created new keys without setting a password:
#### From 1. 
sr key .json file:
```
{"address":"5GZYTuC9S5mUjPFAJoEHNj2NdcgabjuhmYAjeHaWZM64RFod","encoded":"0x3053020101300506032b6570042204202074a41b276960ce99d3ecfbdc266c0e0dd25f09e1cb33fa58dacebc1b4ad77422b52664e52d1d944254acf0353c16d326e7f2dce86ee721f348418ca6ae2a21a123032100c6ef6eb0e60dc5fe0d6ee2072be6d746882def9551c61db4bc1d34b66c26d420","encoding":{"content":["pkcs8","sr25519"],"type":"none","version":"2"},"meta":{"name":"sr-key","tags":[],"whenCreated":1600258195404}}
```
ed key .json file:
```
{"address":"5FpjwJDbW7jLTwgAHesAvurTeshQ6HisAi6EFqo48CRJgQb1","encoded":"0x3053020101300506032b6570042204202d952efd931b55531a646f24c1caba79fa95eb08fab5bedebf70a567c7190988a64a7fd96e75b2a6eb48439a12ed62d9c21f9346eb5df7cd721aa92f07c3726da123032100a64a7fd96e75b2a6eb48439a12ed62d9c21f9346eb5df7cd721aa92f07c3726d","encoding":{"content":["pkcs8","ed25519"],"type":"none","version":"2"},"meta":{"name":"ed-key","tags":[],"whenCreated":1600259396709}}
```

#### From 2.
Applied for a role on alexandria test 2 (defaults to ed):
```
[...]

PR #1390, opened 2020-09-17
Pioneer: Ignore "genesisHash" when loading/creating accounts

Lezek123:
Related issue: https://github.com/Joystream/joystream/issues/1383

PR content:
- Ignore `genesisHash` in local storage keys
- Don't save `genesisHash` when creating new keys
- Remove "Only this network" option from the accounts UI.


mnaamani (approved):
Works as expected.

mnaamani:
For any keys already in storage created before these changes, backing up the key will still include the genesis hash. I think that is acceptable (just pointing it out so when you do testing martin)

bwhm (approved):

Issue #1340, opened 2020-09-14
Content directory schema: define `Class` entity 


metmirr:
`Class` entity has relationships with the following entities:

- `Property`
- `Schema`
- `CuratorGroup`
- `EntityCreationVoucher`

_The issue https://github.com/Joystream/joystream/issues/644 is used as a reference to estimate time._

PR #1399, opened 2020-09-17
Content directory schema: `Class` entity defination

metmirr:
This PR introduces entity definition for `Class` type to the query node schema and addresses this issue https://github.com/Joystream/joystream/issues/1340.


query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,212 @@
...
+	maximumEntitiesCount: BigInt!
+
+	"The current number of entities which exist"
+	currentNumberOfEntities: BigInt!
+
+	"How many entities a given controller may create at most"
+	defaultEntityCreationVoucherUpperBound: BigInt!
+
+	"For this permission, the individual member is allowed to create the entity and become controller"
+	anyMemberCanCreate: Boolean
```
iorveth: Shouldn't we have a separate type for the `Class` permissions related fields here?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,212 @@
...
+
+	"Whether to prevent everyone from creating an entity"
+	entityCreationBlocked: Boolean!
```
iorveth, line 55: Why `entityCreationBlocked` & `allEntityPropertyValuesLocked` fields should not be null, but `anyMemberCanCreate` flag haven`t this restriction?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,212 @@
...
+	schemas: [Schema!] @derivedFrom(field: "class")
+
+	"Current class maintainer curator groups"
+	maintainers: [CuratorGroup!]
+
+	"All vouchers for this class"
+	vouchers: [EntityCreationVoucher!] @derivedFrom(field: "class")
+}
+
+enum PropertyType {
```
iorveth, line 73: Please, add Hash type here https://github.com/iorveth/joystream/blob/content_directory_second_try/runtime-modules/content-directory/src/schema.rs#L56.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,212 @@
...
+	description: String!
+
+	"Property is locked from maintainer"
+	isLockedFromMaintainer: Boolean!
+
+	"Property is locked from controller"
+	isLockedFromController: Boolean!
+
+	"Upper bound for vector length when property is a vector"
+	maximumLength: BigInt
```
iorveth, line 122: Shouldn't we create a separate schema for `VecPropertyType` ?https://github.com/iorveth/joystream/blob/content_directory_second_try/runtime-modules/content-directory/src/schema.rs#L91

Afterwards we possibly can utilize union type to choose between `Type` or `VecType`, can't we? 
It`s seems much clearer, IMHO. https://graphql.org/learn/schema/#union-types

```
"Optional field, it must reference a class then PropertyType is Reference"
	referencedClass: Class
	"Optional field, Only will be set when PropertyType is Reference"
	sameControllerStatus: Boolean
```

This fields should go to `VecPropertyType`

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,212 @@
...
+	groupId: BigInt!
+
+	"Activity stats of group"
+	isActive: Boolean!
+
+	"All member curators of this group"
+	curators: [Curator!]
+
+	"All classes this group maintains"
+	maintainerOfClasses: [Class!] @derivedFrom(field: "maintainers")
```
iorveth, line 182: We only should know the number of classes, maintained by this curator group, don`t we?

```
pub struct CuratorGroup<T: Trait> {
    /// Curators set, associated with a iven curator group
    curators: BTreeSet<T::CuratorId>,

    /// When `false`, curator in a given group is forbidden to act
    active: bool,

    /// Used to count the number of `Class`(es), given curator group maintains
    number_of_classes_maintained: u32,
}
```

iorveth (changes requested):

query-node/joystream-query-node/schema.graphql
metmirr: For the GraphQL API purposes, I am not sure if someone would want to query only class permissions if he/she wants then they can query the fields (permissions related) explicitly.
Is there any case you can think of?

query-node/joystream-query-node/schema.graphql
metmirr: It should have actually, I must have forgotten it. `ClassPermissions` fields aren't optional as I understand.

query-node/joystream-query-node/schema.graphql
metmirr: Yeah, agree with that. We have `maximumLength` field when the property type is `Vector` and [`referencedClass`, `sameControllerStatus`] fields when the property type is `Reference`. So that means we have two options if we use union type (We have `variant` directive that can be used) ie:

```graphql
type ReferenceTypeData @variant {
  referencedClass: Class
  sameControllerStatus: Boolean
}

type VectorTypeData @variant {
  maximumLength: BigInt
}

union PropertyData =  ReferenceTypeData | VectorTypeData

type Property @entity {
   #...
   propertyData: PropertyData
}
```

But a property type has 3 variants actually `Reference`, `Vector` and (`Int*, `Text`). If you notice there is no extra field exists when property type is (`Int*, `Text`). I mean if i could have at least a field for (`Int*, `Text`) ie:

```
type TextIntTypeData @variant {
  someField: SomeType
}
```
then it would be okey to do that. I hope I could explain it :)

query-node/joystream-query-node/schema.graphql
metmirr: Yes right, but instead of having only a number we create a sql reference to those maintainers. Let's say someone needs to get all maintainers data related to maintainers ie: 
```graphql
query {
  class(classId: 123) {
    maintainers {
      groupId
      isActive
    }
  }
}
```

query-node/joystream-query-node/schema.graphql
metmirr: Thanks, done!

query-node/joystream-query-node/schema.graphql
metmirr: Thanks, done!

query-node/joystream-query-node/schema.graphql
iorveth: I don`t think someone will query only class permissions, though it seems to be easier to select all permissions related fields, when he/she need them. 

```
query {
    class {
      classId
      classPermissions
  }
}
```

instead of 

```
query {
    class {
      classId,
      anyMemberCanCreate,
      entityCreationBlocked,
      allEntityPropertyValuesLocked,
  }
}
```

Will it complicate things a lot?

query-node/joystream-query-node/schema.graphql
iorveth: ok, good to know!

query-node/joystream-query-node/schema.graphql
iorveth: Oh, i see, type system is much less flexible here. 
Thank you for the clarification.

iorveth (changes requested):

iorveth (approved):
LGTM

metmirr:
Superseded by #1458

Issue #1341, opened 2020-09-14
Content directory schema: define `Entity` entity


metmirr:
`Entity` entity has relationships with the following entities:

- `Class`
- `EntityController`
- `PropertyValue`
- `Schema` 

This issue depends on #1340 

_The issue #644 is used as a reference to estimate time._

PR #1400, opened 2020-09-17
Content directory schema: `Entity` entity defination

metmirr:
This PR introduces entity definition for `Entity` type to the query node schema and addresses this issue https://github.com/Joystream/joystream/issues/1341.

The PR is built on #1399


query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,291 @@
...
+"""
+type Entity @entity {
+	"Runtime identifier for entity"
+	entityID: BigInt!
+
+	"Current controller, which is initially set based on who created entity"
+	controller: EntityController!
+
+	"Forbid groups to mutate any property value"
+	isFrozen: Boolean!
```
iorveth, line 226:
```
"Current controller, which is initially set based on who created entity"
	controller: EntityController!
	"Forbid groups to mutate any property value"
	isFrozen: Boolean!
"Prevent from being referenced by any entity (including self-references)"
	isReferenceable: Boolean!
```
Can we move `EntityPermissions` related fields to separate type, please?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,291 @@
...
+
+	hash: Bytes
+
+	reference: Entity
+}
+
+"""
+Fields encode a flattened representation of vector properties, only one will be non-null
+"""
+type VectorValue @variant {
```
iorveth, line 266: Don't sure, if i fully understand how `@variant` directive works. 
Though i concerned about the `nonce: BigInt!` field here. 
We should have this field for each of `VectorValue` variants.

iorveth (changes requested):

query-node/joystream-query-node/schema.graphql
metmirr: If we have `EntityPermissions` type there is no use except in `Entity` so I think it's good to keep the fields here.

query-node/joystream-query-node/schema.graphql
metmirr: Yes, we will have `nonce` (non-null) field for each variants. You can find more on `@variant` directive there https://dzhelezov.gitbook.io/hydra-docs/docs/schema-spec/variant-types

iorveth: ok, great

query-node/joystream-query-node/schema.graphql
iorveth: I don`t think someone will query only entity permissions, though it seems to be easier to select all permissions related fields, when he/she need them. 

```
query {
    entity {
      entityId
      entityPermissions
  }
}
```

instead of 

```
query {
    entity {
      entityId,
     controller: EntityController!
	"Forbid groups to mutate any property value"
	isFrozen: Boolean!
"Prevent from being referenced by any entity (including self-references)"
	isReferenceable: Boolean!
  }
}
```

Will it complicate things a lot?

iorveth (approved):
LGTM!

metmirr:
Superseded by #1458

PR #1253, opened 2020-09-07
Fix chain spec builder.

shamil-gadelshin:
Add `authority_discovery` key to the chain_spec_builder.

PR #1401, opened 2020-09-18
runtime: bump version and build chainspec builder in docker image

mnaamani:
After #1253 we should have updated spec version.

I'm also including chain-spec-builder binary in joystream/node docker image build to be used in the integration tests.


shamil-gadelshin (approved):
LGTM

Issue #810, opened 2020-06-25
Use a Proper CDN for fonts [Atlas]


fcescob:
We should set up a CDN to host our fonts instead of checking them in to source control (required for premium fonts like our headers font, nice to have for the others).

kdembler:
For the time being, we can bundle this with #1269 and use the same storage for hosting the fonts

PR #1404, opened 2020-09-18
use mocked atlas assets hosted on jsgenesis infra

### LINKED ISSUE #1269: Host mocked media assets on Jsgenesis infrastructure
We want to host media assets on our Jsgenesis infrastructure so we don't need to rely on 3rd part providers to display mocked data inside Atlas. To do that let's use some static storage that will provide public access to the assets:
- Find suitable storage (coordinate with Martin)
- Figure out an access scheme that makes sense - avoid situations when one person goes on vacation and the rest of the team is cut off
- Select free assets and upload them to the storage, compose lists of URLs and save it in the codebase

kdembler:
Resolves #1269 and #810


fcescob (approved):
LGTM

bedeho (approved):

PR #1411, opened 2020-09-18
Pioneer: Alexandria deployment landing page

Lezek123:
Allows building static landing page with link to release blog post instead of the entire app if `IS_LIVE=false` is provided in `env`.

Examples:
- `IS_LIVE=false yarn workspace pioneer build` - builds just the static page
- `IS_LIVE=false yarn workspace pioneer start` - serves the static page on `localhost:3000`
- `yarn workspace pioneer build` - builds Pioneer

**Static page desktop:**
alexandria-landing

**Static page mobile:**
alexandria-landing-mobile


mnaamani:
Excellent. This will work nicely.

mnaamani (approved):
Implementation wise this is exactly why we needed, a simple way to render a "site under maintenance" page. Just waiting for okay from @bwhm and @bedeho regarding design/layout.

bedeho:
In order to judge this, I would need to see it fully rendered, not sure how I do that, but I am quite skeptical that I will say this is better than simply having no Pioneer button. This is a non-trivial design task, getting deliverable from designers should have been scheduled.

bedeho:
Someone could still have a raw link I guess, we could also remove the entire hosted app, but that is a bit rough? I am happy to ACK this now, given that my complaint is so late in the day, but this _really_ should have been scheduled before with a designer.

mnaamani:
> Someone could still have a raw link I guess, we could also remove the entire hosted app, but that is a bit rough? I am happy to ACK this now, given that my complaint is so late in the day, but this _really_ should have been scheduled before with a designer.

If someone has a raw link to any part of pioneer app they will get this same page rendered. I agree we probably should have scheduled this.

We chose this approach to avoid tinkering with domain redirects (as we did last time which ended up in mistakenly doing HTTP 301 Redirect -Moved Permanently, which forced users to clear their local browser cache/data and loosing their stored keys) to be able to visit the site gain without being redirected.

PR #1414, opened 2020-09-19
Runtime membership module: Fix set_root_account

mnaamani:
Unfortunately the unit test was incorrect and we didn't have test coverage for managing keys in the integration tests.
Currently member is unable to change the root account!

We don't provide a direct way to change the root account in pioneer it would have do be done via extrinsics app, so I don't expect many have tried to do this anyway. There would have been an inconsistency in the state that you could not lookup a member by their root account. Not sure if that has had any broken experiences.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,7 +1993,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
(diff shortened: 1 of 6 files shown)


bedeho (approved):
LGTM

PR #1430, opened 2020-09-21
cargo update yanked crossbeam-channel

mnaamani:
While fixing README instruction for joystream-node on how to do cargo install, because build fails (since it wasn't looking at the Cargo.lock file) fixed by adding `--lock`. It did however display a warning:

```
warning: package `crossbeam-channel v0.4.3` in Cargo.lock is yanked in registry `crates.io`, consider running without --locked
```

The version of crossbeam-channel in our cargo lock file was yanked: https://crates.io/crates/crossbeam-channel/0.4.3
Although crates.io says dependants on this package can still download it, I don't for how long that is supported so I bumped the version to latest v0.4.4.

Since this looks like a patch version change, it didn't seem to warrant a runtime spec version bump, just impl version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -711,12 +711,12 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.4.3"
+version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09ee0cc8804d5393478d743b035099520087a5186f3b93fa58cec08fa62407b6"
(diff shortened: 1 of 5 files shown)


mnaamani:
I can't find reason why the crate was originally yanked, but it might be worth investigating. Could it affect stability of the node? it is a dependency of `async-std` which is a dependency of `libp2p-mdns`, `libp2p-tcp` and `substrate-prometheus-endpoint` to it looks like network messaging related.

shamil-gadelshin (approved):
LGTM

Issue #1288, opened 2020-09-10
Discuss removal of buttons outline on Chrome


fcescob:
As noted here by @kdembler 

https://github.com/Joystream/joystream/pull/977#discussion_r483523606

Buttons on Chrome have a light blue, rounded outline. This is good for accessibility, but breaks the design. 

We should discuss if its worth to remove it or not.

PR #1431, opened 2020-09-21
Remove Outline On Chrome

fcescob:
This PR addresses #1288

### DIFF
--- a/atlas/packages/app/src/shared/components/Button/Button.style.ts
+++ b/atlas/packages/app/src/shared/components/Button/Button.style.ts
@@ -134,10 +134,7 @@ export const StyledButton = styled.button<ButtonStyleProps>`
 	align-items: center;
 	&:hover {
 		cursor: ${(props) => (!props.disabled && props.clickable ? 'pointer' : '')}
-	}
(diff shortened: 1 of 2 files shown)


atlas/packages/app/src/shared/components/Button/Button.style.ts
```diff
@@ -135,6 +135,9 @@ export const StyledButton = styled.button<ButtonStyleProps>`
 	&:hover {
 		cursor: ${(props) => (!props.disabled && props.clickable ? 'pointer' : '')}
 	}
+  &:focus {
```
kdembler: Maybe we should make that rule general and just remove outline for all the focused elements?

fcescob: Good idea, if we do it in the styles that are specific to the App itself, it will be easier to separate the components later as well.

bedeho: On that note, do we have some sort of coding style guide or something?

bedeho (approved):

Issue #1293, opened 2020-09-10
Remove useless UI elements


fcescob:
Currently we have some useless UI elements, like  the "Share" button on the landing page, which have no purpose.

We should go through each page of Atlas and remove elements that will have no purpose in the first release.

PR #1432, opened 2020-09-21
Remove Useless Elements

fcescob:
This PR addresses #1293


atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -37,7 +37,7 @@ const Navbar: React.FC<RouteComponentProps> = () => {
           <Link to="/">
             <StyledIcon name="home" />
           </Link>
-          <Link to="/discover">
+          <Link to="/browse">
```
kdembler: We should use the value from `routes` object here

fcescob: Fixed, thanks!

kdembler (approved):

bedeho (approved):

Issue #1391, opened 2020-09-17
Babylon: @joystream/types - adapt for the new content directory


Lezek123:
Adapt `@joystream/types` for the new runtime changes intorduced in https://github.com/Joystream/joystream/pull/1385 (related issue: https://github.com/Joystream/joystream/issues/1275)
Will require a big (complete?) rework of `@joystream/types/versioned-store`.

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1433

PR #1433, opened 2020-09-21
@joystream/types: adapt for new content directory

Lezek123:
Implements https://github.com/Joystream/joystream/issues/1391 (adapting `@joystream/types` for new content directory).
The types should work with the runtime built from `content_directory_second_try` branch, commit https://github.com/Joystream/joystream/pull/1385/commits/a3e59df01b9b12f9143ff29cdb018edda6aea28b (see: https://github.com/Joystream/joystream/pull/1385)

### Changes:
- Removed outdated `versioned-store` and `versioned-store-permissions` types to avoid namespace clashes with the new `content-directory` module.
**_Note: This renders `versioned-store` and `versioned-store-permissions` modules useless, but there's not much that can be done to avoid this. Those modules will be removed soon anyway, so it shouldn't be a problem._**
- Added new content directory types in `@joystream/types/content-directory`
- Created custom `Hash` type in ``@joystream/types/common` for better readability
- Created custom `JoyBTreeSet` type which fixes encoding of `BTreeSet` (adds sorting expected by the runtime)
_**Note: To take full advantage of it we would have to create custom types for all `BTreeSet<X>` types in the runtime that are passed to extrinsics as arguments (like `ApplicationIdSet`, which we already have)**_
- Auto-generated new api augmentation files etc. (required minor tweak in one script)
- Added a script in `api-examples` to initialize and test some basic extrinsics of the new content directory: `yarn workspace api-examples script initNewContentDir` (can be run on fresh dev chain after registering `Alice` as a first member)
- Excluded media app from Pioneer in order to allow it to build/start and make testing easier
- Bumped types version


pioneer/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "@polkadot/dev/config/tsconfig",
   "exclude": [
     "build/**/*",
-    "**/build/**/*"
+    "**/build/**/*",
+    "packages/joy-media/**"
```
mnaamani, line 6: why not just remove `packages/joy-media/` folder completely ?

types/src/common.ts
```diff
@@ -1,20 +1,55 @@
...
+
+  public toU8a(isBare?: boolean): Uint8Array {
+    const encoded = new Array<Uint8Array>()
+
+    if (!isBare) {
+      encoded.push(Compact.encodeU8a(this.size))
+    }
+
+    Array.from(this)
+      .sort()
```
mnaamani: Will the default [sort()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) - "The default sort order is ascending, built upon converting the elements into strings, then comparing their sequences of UTF-16 code units values."  correspond to how the runtime expects the sorting. 

In the past I've only confirmed this to be correct for BTreeSet<number> how well will this work for more complex Codec types?

types/src/common.ts
```diff
@@ -1,20 +1,55 @@
...
+}
+
 export class Credential extends u64 {}
 export class CredentialSet extends Vec.with(Credential) {} // BtreeSet ?
 
 // common types between Forum and Proposal Discussions modules
 export class ThreadId extends u64 {}
 export class PostId extends u64 {}
 
+export class Hash extends U8aFixed implements H256 {}
```
mnaamani: If I'm not mistaken there is already a type named `Hash` that comes from `@polkadot/types` or is that now `H256` ?

types/src/content-directory/index.ts
```diff
@@ -0,0 +1,297 @@
...
+}) {}
+
+// Versioned store relicts - to be removed:
+export class ClassPermissionsType extends Null {}
+export class ClassPropertyValue extends Null {}
+export class Operation extends Null {}
+export class ReferenceConstraint extends Null {}
+
+export const contentDirectoryTypes = {
+  Nonce,
```
mnaamani, line 246: Can we confirm there is no substrate type named Nonce ? I think the nonce type name for accounts in the system module is `Index` but we should double check.

types/src/content-directory/index.ts
```diff
@@ -0,0 +1,297 @@
...
+export class SameController extends bool {}
+
+export class ClassPermissions extends JoyStructDecorated({
+  any_member: bool,
+  entity_creation_blocked: bool,
+  all_entity_property_values_locked: bool,
+  maintainers: JoyBTreeSet.with(CuratorGroupId),
+}) {}
+
+// Named just "Type" in the runtime, but this name conflicts with @polkadot/types/primitive/Type.ts
```
mnaamani, line 25: As long as this type is only used as a type of of field in a struct it shouldn't be an issue when reading storage values from chain.

Perhaps `StoredPropertyType`  might be better name choice, it corresponds with `StoredPropertyValue` .. ?

Maybe not much of an improvement but I have an issue with the term `Single`. Its not a big deal though.

types/src/content-directory/index.ts
mnaamani: confirmed.

types/src/council/index.ts
```diff
@@ -52,7 +53,7 @@ export type ISealedVote = {
 export class SealedVote
   extends JoyStructDecorated({
     voter: AccountId,
-    commitment: U8aFixed, // Hash
+    commitment: Hash,
```
mnaamani, line 56: Why did we need to use a class constructor here and not the `Hash` interface from `@polkadot/types/interfaces` ?

mnaamani (changes requested):
Looks great.
I double checked the types match the runtime, didn't find any discrepancy.

I think you forgot to check-in the updated yarn.lock file

Ran the test script:

Screen Shot 2020-09-24 at 9 39 10 AM

No issues, except when I pasted in js-toolbox it failed because of the `joy` global was missing (fixed by removing it)

pioneer/tsconfig.json
Lezek123: I just thought it would be cleaner to do it in a separate PR

types/src/common.ts
mnaamani: I can think of two 'hacks' to help in constructing BTreeMap/Set that are properly sorted.

Option 1. custom RPC calls in the joystream-node
Option 2. In rust replicate the BTreeMap/Set implementations for the types in question, and compile into WASM libraries that can be used from our javascript code?

mnaamani: Another workaround is for dispatchables that expect a BtreeSet<T>, they should take a Vec<T> instead and construct the BTreeSet themselves. This will not work however when the set is a field of a struct which is the argument type.

types/src/content-directory/index.ts
Lezek123: It's a type that describes `Single` variant of `PropertyType` enum, ie.:
```
export class PropertyType extends JoyEnum({
  Single: SinglePropertyType,
  Vector: VecPropertyType,
}) {}
```
Maybe we should use `PropertyTypeSingle` and `PropertyTypeVector`?

types/src/council/index.ts
Lezek123: We cannot use an interface here, because interfaces are a compilation-time-only constructs in TypeScript.
Here we're passing values that are later passed to `Struct.with()`, so they have to be `Record<string, Constructor | string>`.
To `Struct.with()` we could just pass `'Hash'` (a string), but this will not work with `JoyStructDecorated` which needs `Constructor`s to do the "decorating"

types/src/common.ts
Lezek123: This is just a "helper" for cases when class constructor is needed instead of an interface (`@polkadot/types` only expose `Hash` interface, while the underlying class is `U8aFixed`) 
`Hash` and `H256` interfaces should be identical, but I can implement the original `Hash` one for more clarity.

types/src/content-directory/index.ts
Lezek123: Any situation like this should cause conflicts in files like `augment/augment-types.ts` and fail the CI checks

types/src/common.ts
Lezek123: Fixed a `sort()` issue and enforced `JoyBTreeSet` values to extend `UInt` (compile-time) to make sure the sorting will always be correct.
With that we can be sure that when executing extrinsics like `api.tx.storageWorkingGroup.fillOpening` we can provide application ids in any order (this can be tested ie. in Pioneer's `js` console).
As I mentioned before, runtime-side it just requires that any `BTreeSet` extrinsics arguments (unless part of structs) should have their own type name (just like `ApplicationIdSet` or `CredentialSet`)
Struct fields should be covered by that solution without any required tweaks in the runtime.

mnaamani: Excellent.

types/src/common.ts
mnaamani: Ok thanks for clarification. I think H256 more explicitly matches the length 32-bit length we use in our runtime so its fine.

types/src/content-directory/index.ts
mnaamani: > Maybe we should use `PropertyTypeSingle` and `PropertyTypeVector`?

Thats a better suggestion.

mnaamani (approved):

PR #1434, opened 2020-09-21
Working team pallet. Stakes and extra hiring flow extrinsics.

shamil-gadelshin:
Comments: 
- added stake support for the hiring flow extrinsics
- added multiple extrinsics:
   - increase_stake
   - decrease_stake
   - slash_stake
   - withdraw_application
   - cancel_opening
- unstaking period is not implemented yet

### DIFF
--- a/runtime-modules/working-team/src/lib.rs
+++ b/runtime-modules/working-team/src/lib.rs
@@ -145,6 +145,12 @@ decl_event!(
         /// - regular worker/lead id.
         /// - new stake amount
         StakeDecreased(TeamWorkerId, Balance),
+
+        /// Emits on increasing the regular worker/lead stake.
+        /// Params:
+        /// - regular worker/lead id.
+        /// - new stake amount
+        StakeIncreased(TeamWorkerId, Balance),
     }
 );
 
@@ -478,6 +484,38 @@ decl_module! {
             )?;
 
             Self::deposit_event(RawEvent::StakeDecreased(worker_id, new_stake_balance));
+        }
+
+        /// Increases the regular worker/lead stake, demands a worker origin.
+        /// Locks tokens from the worker staking_account_id equal to new stake. No limits on the stake.
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn increase_stake(origin, worker_id: TeamWorkerId<T>, new_stake_balance: BalanceOfCurrency<T>) {
+            // Checks worker origin and worker existence.
+            let worker = checks::ensure_worker_signed::<T, I>(origin, &worker_id)?;
+
+            ensure!(
+                new_stake_balance != <BalanceOfCurrency<T>>::zero(),
+                Error::<T, I>::StakeBalanceCannotBeZero
+            );
+
+            T::StakingHandler::ensure_can_increase_stake(
+                T::LockId::get(),
+                &worker.staking_account_id,
+                new_stake_balance,
+            )?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // This external module call both checks and mutates the state.
+            T::StakingHandler::increase_stake(
+                T::LockId::get(),
+                &worker.staking_account_id,
+                new_stake_balance,
+            )?;
+
+            Self::deposit_event(RawEvent::StakeIncreased(worker_id, new_stake_balance));
         }
     }
 }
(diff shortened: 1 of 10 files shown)


runtime-modules/working-team/src/lib.rs
```diff
@@ -209,33 +255,50 @@ decl_module! {
...
+            staking_account_id: T::AccountId,
+            description: Vec<u8>,
+            stake: Option<BalanceOfCurrency<T>>,
         ) {
-            // Ensure origin which will server as the source account for staked funds is signed
+            // Ensure origin which will server as the source account for staked funds is signed.
             let source_account = ensure_signed(origin)?;
 
-            // Ensure the source_account is either the controller or root account of member with given id
+            // Ensure the source_account is either the controller or root account of member with given id.
```
bedeho: 1. Why are we checking the root account?
2. Why are we coupled to the membership module? 

There should just be a simple `is_member` check on the runtime trait.

runtime-modules/working-team/src/types.rs
```diff
@@ -101,14 +113,88 @@ pub struct Worker<AccountId, MemberId> {
...
+
+    /// Verifies that the stake could be decreased to a given amount.
+    fn ensure_can_decrease_stake(
+        lock_id: LockIdentifier,
+        account_id: &T::AccountId,
+        new_stake: BalanceOfCurrency<T>,
+    ) -> DispatchResult;
+
+    /// Decreases the stake for to a given amount.
+    fn decrease_stake(
```
bedeho: Why do decrease and increase return dispatchable results, they should not be able to fail - just like slash, as they just set a lock on the given account. It resulting in possible errors _MUTATION SAFE_ markers.

runtime-modules/working-team/src/types.rs
```diff
@@ -101,14 +113,88 @@ pub struct Worker<AccountId, MemberId> {
...
+    /// Unstaking period for the stake. Zero means no unstaking period.
+    pub unstaking_period: BlockNumber,
+}
+
+/// Defines abstract staking handler to manage user stakes for different activities
+/// like adding a proposal. Implementation should use built-in LockableCurrency
+/// and LockIdentifier to lock balance consistently with pallet_staking.
+pub trait StakingHandler<T: system::Trait + GovernanceCurrency> {
+    /// Locks the specified balance on the account using specific lock identifier.
+    fn lock(lock_id: LockIdentifier, account_id: &T::AccountId, amount: BalanceOfCurrency<T>);
```
bedeho: Can we drop these lock_ids ? I know you wanted to post pone, but it really does not make sense to me to have them even now

runtime-modules/working-team/src/types.rs
```diff
@@ -101,14 +113,88 @@ pub struct Worker<AccountId, MemberId> {
...
+
+    /// Removes the specified lock on the account.
+    fn unlock(lock_id: LockIdentifier, account_id: &T::AccountId);
+
+    /// Verifies that stake could be placed using given account.
+    fn ensure_can_make_stake(
```
bedeho: The semantics of this call is too vague, the name is overly broad. A consequence appears to be that no check is made that the staker is providing an account that is actually under their control. The signature of this interface could not check this even in principle, as it does not have the identity of the staker.

I suggest we make things crystal clear by having two distinct calls
- `is_member_staking_account(member_id, account_id) -> bool`
- `is_account_free_of_conflicting_staking(account_id) -> bool`

runtime-modules/working-team/src/types.rs
bedeho: Let me add to that: why are there even checkers and mutators even here?

I don't see a need for either.

bedeho (changes requested):
- I know this is not from this PR, but: Why does `JobOpening` have `created` field?

runtime-modules/working-team/src/types.rs
shamil-gadelshin: Ok.

runtime-modules/working-team/src/types.rs
shamil-gadelshin: Ok.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Fixed.

runtime-modules/working-team/src/types.rs
shamil-gadelshin: I changed the semantics.

shamil-gadelshin:
Among other fixes I remove abstract `ensure_* `methods from the `StakingHandler` trait and introduced fine grained methods. Also, `decrease_stake()` is infallible now.

bedeho (approved):
LGTM.

Issue #1270, opened 2020-09-09
Expand and tweak the mocked dataset


kdembler:
To do as part of this task:
- Make the dataset deterministic - currently assets are shuffled around, let's make it static
- Add more data points - we should have enough assets to enable infinite scrolling without duplication (around 30 videos I think)
- Use media assets hosted on Jsgenesis infrastructure (thumbnails + videos) - blocked by #1269 - all the URLs should be defined in a single place so it's easy to switch them out
- Make sure the data is complete and diversified - videos have different categories, some channels have videos, others don't

PR #1437, opened 2020-09-22
expand the mocking dataset, add mocking tooling

kdembler:
This PR expands the mocked dataset as part of #1270 

Some scripts were added so we can easily regenerate the mocked data if we want it to meet some different criteria in the future. I've also added a script for parsing the video files so that we don't have to do this by hand for every video.

There is one issue though, due to some strangness in MirageJS, all the returned `video.media.location` objects are the same so the video playback will serve the same file for now.


fcescob (approved):
LGTM!

bedeho (approved):

Issue #1348, opened 2020-09-14
Content directory schema: define `Transaction` and `EntityOwnership` entities


metmirr:

bedeho:
Out of scope for Babylon, is not needed because `transaction` extrinsic will still trigger the same events that we already have mappings for.

PR #1458, opened 2020-09-24
Introduce content directory graphql schema for query node

metmirr:
This PR introduces the content directory graphql input schema for the query node. All the types are defined in the schema expect https://github.com/Joystream/joystream/issues/1348


bedeho (commented):
Why are the changes in https://github.com/Joystream/joystream/pull/1458/commits/8a13fa71f4d1d37003a12f688a09c0c133727f1f here?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+	"Member's controller account id"
+	controllerAccount: Bytes!
 
-  # Member's root account id
-  rootAccount: Bytes!
+	"Member's root account id"
+	rootAccount: Bytes!
+
+	"When a member is worker"
+	worker: Worker @derivedFrom(field: "member")
```
bedeho: A member can occupy 0 or many worker roles, both at the same time, and over time.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	"Virtual field required for 1:1 relationship"
+	entityController: EntityController @derivedFrom(field: "controllerWhenMember")
```
bedeho: Why is this here?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	"Current class maintainer curator groups"
+	maintainers: [CuratorGroup!]
+
+	#### ClassPermissions ####
+
+	# Commenting out this line because the properties are accessible from the schemas
+	# query { classes { schemas {properties { ... } } } }
+	# "All properties that have been used on this class across different class schemas"
+	# properties: [Property!]
```
bedeho: Why is this commented out? I don't think the comment provides a good rationale, may have been me who wrote it long ago.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+	VECUINT64
+	VECINT16
+	VECINT32
+	VECINT64
+	VECTEXT
+	VECREFERENCE
+}
+
+"`Property` representation, related to a given `Class`"
+type Property @entity {
```
bedeho: The `Property` and  `PropertyType` types need to be coordinated better. we have fields such as ".... when the property is a vector, this is a symptom of not using algebraic types, which we should and can do. Look at runtime for guidance. @iorveth  will have valuable input on this.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	"Total number of inbound references from another entities"
+	totalNumberOfInboundReferences: BigInt!
+
+	"Number of inbound references from another entities with 'SameOwner' flag set"
+	sameOwnerInboundReferences: BigInt!
+}
+
+"Fields encode a flattened representation of non-vector properties, only one will be non-null"
+type SingleValue @variant {
```
bedeho: This should be modelled differently, this is not how to model this algebraically, @iorveth may help.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+	hash: Bytes
+
+	reference: Entity
+}
+
+"""
+Fields encode a flattened representation of vector properties, only one will be non-null
+"""
+type VectorValue @variant {
+	boolVec: [Boolean!]
```
bedeho: This should be modelled differently, this is not how to model this algebraically, @iorveth may help.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	"If schema can be added to an entity"
+	isActive: Boolean!
+
+	"Virtual field, supported entities"
+	entities: [Entity!] @derivedFrom(field: "supportedSchemas")
+}
+
+"Working group participant: working-group/src/types.rs"
+type Worker @entity {
```
bedeho: This should probably be an interface, as there are many kinds of workers, not just curators (even though we will only include curators in Babylon). The `Curator` entity can implement this interface. This would also require dropping `curator` field.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	referenceVec: [Entity!]
+
+	"Used to avoid vector data race update conditions"
+	nonce: BigInt!
+}
+
+union StoredValue = SingleValue | VectorValue
+
+type PropertyValue @entity {
```
bedeho: This is not being correctly modelled, please reevaluate. Include @iorveth here.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+	"Runtime identifier of property value in entity"
+	propertyValueId: BigInt! @unique
+
+	"Property to which this value corresponds"
+	property: Property!
+
+	value: StoredValue!
+
+	"Virtual field for entity property values"
+	entities: [Entity!] @derivedFrom(field: "values")
```
bedeho: A property value only corresponds to a single entity, and the `values`field should derive from `PropetyValue`, not the other way around.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,489 @@
...
+
+	"Used to avoid vector data race update conditions"
+	nonce: BigInt!
+}
+
+union StoredValue = SingleValue | VectorValue
+
+type PropertyValue @entity {
+	"Runtime identifier of property value in entity"
+	propertyValueId: BigInt! @unique
```
bedeho: There is no unique property value id from the chain, the id in the chain is just an index in the given class, which will not be unique.

A property value is unique for a given combination and entity&property _of the same class_.

bedeho (changes requested):
- I think @iorveth should be more deeply involved in making and reviewing this, he could have cought some of these issus.
- Please introduce explicit `Id` field in all entities, and explain where value comes from. This is unclear here.

query-node/joystream-query-node/schema.graphql
metmirr: Done in 162095e all roles extends `Worker` interface will have a reference to the `Member` type.

query-node/joystream-query-node/schema.graphql
metmirr: Done 74b9b3d. It is my mistake, a member can control more than one entity.

query-node/joystream-query-node/schema.graphql
metmirr: Done in f1bfa3b. No, it's my comment actually, after fixing `Property` entity (each property has a reference to a class) I uncomment it and it is a reverse lookup now.

query-node/joystream-query-node/schema.graphql
metmirr: I refactored it to the model you suggested here https://github.com/Joystream/joystream/issues/644 waiting @iorveth input on this.

query-node/joystream-query-node/schema.graphql
metmirr: I refactored it to the model you suggested here #644 waiting for @iorveth input on this.

query-node/joystream-query-node/schema.graphql
metmirr: I refactored it to the model you suggested here #644 waiting for @iorveth input on this.

query-node/joystream-query-node/schema.graphql
metmirr: I refactored it to the model you suggested here #644 waiting for @iorveth input on this.

query-node/joystream-query-node/schema.graphql
metmirr: Done in 9540c34

query-node/joystream-query-node/schema.graphql
metmirr: Done in d8b9033

query-node/joystream-query-node/schema.graphql
metmirr: Done in d8b9033. Since `PropertyValue` has a reference to `Property` no longer needed that field.

metmirr:
>     * I think @iorveth should be more deeply involved in making and reviewing this, he could have cought some of these issus.
> 
>     * Please introduce explicit `Id` field in all entities, and explain where value comes from. This is unclear here.

Every entity defined in `schema.graphql` has an `Id: ID!` field by default.

query-node/joystream-query-node/schema.graphql
bedeho: There needs to be a `propertyValueId`, because it tells you the index of the value for the given entity. However, This value is no _unique_ across all `PropertyValue` instances, precisely because lots of entities will have a property with index 0,1,2,...

If this is not clear, please review content director data model again.

query-node/joystream-query-node/schema.graphql
bedeho: #644 is not a solution, that draft was made before we even had algebraic types as a features.

query-node/joystream-query-node/schema.graphql
bedeho: #644 is not a solution, that draft was made before we even had algebraic types as a features.

query-node/joystream-query-node/schema.graphql
bedeho: #644 is not a solution, that draft was made before we even had algebraic types as a features.

query-node/joystream-query-node/schema.graphql
bedeho: #644 is not a solution, that draft was made before we even had algebraic types as a features.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+It represents a conceptual category, similar - but not identical,
+to tables in relational databases and classes in object-oriented programming
+Examples of classes in our example could be:
+- Person: represents the concept of a person.
+- Place: represents the concept of a place.
+- PersonLivedInPlace: represents the concept that a person lived in a place.
+"""
+type Class @entity {
+	"Runtime identifier for class"
+	classId: BigInt! @unique
```
bedeho: replace with `id: ID!`, and the value should be "Runtime identifier for class".

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+
+	"Name of class"
+	name: String!
```
bedeho: Are class names non-null? are they unique? I forget.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	VECINT32
+	VECINT64
+	VECTEXT
+	VECREFERENCE
+}
+
+"'Property' representation, related to a given 'Class'"
+type Property @entity {
+	"Runtime identifier for property in class"
+	propertyId: Int!
```
bedeho: Drop field, replace with new field "id:ID!`, and put value: ClassId+":"+PropertyId from runtime. Comment is then "Runtime class identifier concatenated with property runtime identifier, using colon as separator".

Correct me if I am wrong, ut runtime PropertyId is just an index relative to the class, so its not sufficient in of itself.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	sameControllerStatus: Boolean
+
+	"Schemas which involve this property"
+	schemas: [Schema!] @derivedFrom(field: "properties")
+}
+
+"Represents a standard for what information must be associated with any entity of a given class"
+type Schema @entity {
+	"Runtime identifier of schema in class"
+	schemaId: Int!
```
bedeho: Same idea here, replace with "id: ID!" where value is ClassId+":"+SchemaId, so comment says "Runtime class identifier concatinated with schema runtime identifier, using colon as separator".

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	"If schema can be added to an entity"
+	isActive: Boolean!
+
+	"Virtual field, supported entities"
+	entities: [Entity!] @derivedFrom(field: "supportedSchemas")
+}
+
+"Working group participant: working-group/src/types.rs"
+interface Worker @entity {
+	member: Member!
```
bedeho: I think this needs Id ?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	"Account used to authenticate in this role"
+	roleAccount: Bytes!
+
+	groups: [CuratorGroup] @derivedFrom(field: "curators")
+}
+
+"A group, that consists of 'curators' set"
+type CuratorGroup @entity {
+	"Runtime identifier for group"
+	groupId: BigInt! @unique
```
bedeho: Look at other Id comments.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	curators: [Curator!]
+
+	"All classes this group maintains"
+	maintainerOfClasses: [Class!] @derivedFrom(field: "maintainers")
+}
+
+"A voucher for 'Entity' creation"
+type EntityCreationVoucher @entity {
+	"Class to which voucher applies"
+	class: Class!
```
bedeho: What is Id ?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+enum EntityControllerType {
+	LEAD
+	MEMBER
+	MAINTAINERS
+}
+
+"Owner of an 'Entity'"
+type EntityController @entity {
+	"Current controller, which is initially set based on who created entity"
+	controllerType: EntityControllerType!
```
bedeho: What is Id?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+	voucher: EntityCreationVoucher @derivedFrom(field: "controller")
+}
+
+"""
+Represents a particular instance of a class,
+hence each entity corresponds to exactly one class
+"""
+type Entity @entity {
+	"Runtime identifier for entity"
+	entityId: BigInt! @unique
```
bedeho: See Id comments.

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+
+	"Total number of inbound references from another entities"
+	totalNumberOfInboundReferences: BigInt!
+
+	"Number of inbound references from another entities with 'SameOwner' flag set"
+	sameOwnerInboundReferences: BigInt!
+}
+
+type PropertyValue @entity {
+	entity: Entity!
```
bedeho: What is Id?

query-node/joystream-query-node/schema.graphql
```diff
@@ -1,21 +1,492 @@
...
+
+"The stage of the involvement of a lead in the working group"
+enum LeadRoleState {
+	ACTIVE
+	EXITED
+}
+
+"Working group lead: curator lead content-working-group/src"
+type Lead implements Worker @entity {
+	leadId: BigInt! @unique
```
bedeho: See Id comments.

bedeho (changes requested):
- The problem with algebraic type has not been resolved.
- It needs to be explicitly stated what the `Id` field is for each entity, I believe a random one will be injected by Warthog or some other layer of the data model stack if we do not explicitly control it.

Issue #1454, opened 2020-09-23
Pioneer v1: Show "total" stake pr validator "Staking overview"


bwhm:
For some reason, this tab only shows "own stake" and "other stake".
I checked on Polkadots [UI](https://polkadot.js.org/apps/#/staking), and it's the same there. Seems like adding a new column, should be easy enough, and make it a lot simpler for both existing and aspiring Validators and Nominators to make decisions...

<img width="1492" alt="screenshot" src="https://user-images.githubusercontent.com/35237943/93989935-cb849780-fd8a-11ea-9693-b1b24218bb33.png">

One could even argue the overview below would be appropriate to inject as well:
<img width="1492" alt="screenshot-2" src="https://user-images.githubusercontent.com/35237943/93991421-92e5bd80-fd8c-11ea-87ef-f0070fe9f328.png">

...although I suspect it might cause some issues with responsiveness..?

---

I don't think this should be high on any list of priorities, but just something to keep in mind.

DzhideX:
Are we 100% set on the total stake tab being all the way to the left? It is definitely possible but when we open the nominators on other stake it doesn't look as nice as it pushes them all vertically right underneath instead of also pushing some horizontally. 

Maybe between own stake and commission would be a better place?

bwhm:
> Are we 100% set on the total stake tab being all the way to the left? 

Not at all, but it was easier to inject it there using only screenshots :D

> It is definitely possible but when we open the nominators on other stake it doesn't look as nice as it pushes them all vertically right underneath instead of also pushing some horizontally.

Good point.

> Maybe between own stake and commission would be a better place?

That was actually what I thought would be best as well, so go for it :D

PR #1459, opened 2020-09-24
Implement total stake in staking overview

DzhideX:
This PR addresses issue #1454 

Note:

- I am not sure if `stakeOwn` and `stakeOther` will ever really be undefined, but I put checks in place. If anyone thinks these might not be necessary, I'm happy to take them out.

### DIFF
--- a/pioneer/packages/page-staking/src/Overview/Address/index.tsx
+++ b/pioneer/packages/page-staking/src/Overview/Address/index.tsx
@@ -131,6 +131,14 @@ function Address ({ address, className = '', filterName, hasQueries, isElected, 
         )}
       </td>
       <td className='number'>
+        {(stakeOwn && stakeOther)
+          ? <FormatBalance value={stakeOther.add(stakeOwn)} />
+          : stakeOwn ? <FormatBalance value={stakeOwn} />
+            : stakeOther ? <FormatBalance value={stakeOther} />
+              : null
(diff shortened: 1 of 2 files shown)


bwhm (approved):
Tested, and works fine :)

pioneer/packages/page-staking/src/Overview/Address/index.tsx
```diff
@@ -130,6 +130,14 @@ function Address ({ address, className = '', filterName, hasQueries, isElected,
           <FormatBalance value={stakeOwn} />
         )}
       </td>
+      <td className='number'>
+        {(stakeOwn && stakeOther)
```
Lezek123: `stakeOther` seems to be calculated based on `stakeTotal`, so perhaps we can just use the latter here and avoid too many unnecessary checks (only leaving one `stakeTotal ? /* ... */ : null` or `stakeTotal && /* ... */`)

Lezek123 (changes requested):

Lezek123 (approved):
Seems to work well. Doesn't make responsiveness much worse, as there similar problems with it before anyway, so I think it's good to go!

bwhm:
@mnaamani 
Can we merge this?

mnaamani:
LGTM go for it @bwhm

Issue #1393, opened 2020-09-17
Babylon: @joystream/types: api agumentation with custom class types


Lezek123:
As this is a relatively easy step and may greatly simplify further work, I think it's worth implementing in the current release.

The idea is that we can then make queries etc. without the need for type assertions (which take some time to get right) in Pioneer (which heavily depends on custom-class types) and other projects.

We can already kind-of do this using interfaces generated with `@polkadot/typegen`, but this solution has a few downsides that can be eliminated:
- Refactoring Pioneer to use interfaces instead of class-types would take a long time
- We cannot access getters like `Entity.values` in auto-generated interfaces (because `values` is already part of `Struct` and all interfaces extend `Struct`. In our custom-class types we change the getter name to ie. `entity_values` to sidestep this problem)
- If our class types define getters that return modified field value (ie. `return this.getField('name').toString()`), the type in the auto-generated interface will not match

The solution is to add a script to generate `agument-api` in `types/augment-codec`, which will augment the `api.query`, `api.tx` and `api.consts` types with our own classes instead of interfaces generated by `@polkadot/typegen`. The easiest way to do this seems to be just replacing the imports in the files that we already generate (ie. take `augment-api` files from `types/agument` and replace imports section with imports analogous to those from `augment-codec/augment-types.ts`).

Adding a script to do this should be relatively easy, but considering there may be some unexpected circumstances, I'm adding `estimate-6h`.

Lezek123:
Done in https://github.com/Joystream/joystream/pull/1460

PR #1460, opened 2020-09-24
@joystream/types tooling improvements - augment-codec and defs.json

Lezek123:
Draft based on https://github.com/Joystream/joystream/pull/1433
Implements: https://github.com/Joystream/joystream/issues/1393

Changes:
- Full augmentation support for custom types (`@joystream/types/augment-codec`) - replaced `generateCodecDefs` script with `generateAugmentCodec`
- Modified `generateRegistryJson` script to create perfectly valid `defs.json`. Tweaks to make it compatible with `@polkadot/typegen` are implemented at later stage (in `types/augment/all/definitions.ts`), which should make it possible to use `defs.json` file to register joystream types without any issues (this may be helpful for applications that expect types in json string format or for applications that want to use `@joystream/types/augment` with 100% compatibility)

Unfortunately it turned out that using the new augmentation in Pioneer will still be very tricky, because packages like `@polkadot/react-signer` etc. are often trying to access modules/methods that don't exist in Joystream chain (ie. `api.tx.proxy.proxy`), resulting in a lot of typescript errors in those packages. To some extend this provides valueable information, but fixing all those errors would take a lot of time and probably make it very hard or almost impossible to pull from upstream `@polkadot-js/apps` again (if we ever plan to do this)

A potential solution may be to have an additional signature in augmented apis like:
```
    [module: string]: {
      [method: string]: QueryableStorageEntry<ApiType>;
    },
``` 
So that, for example, any unknown `api.query` method fallsback to `QueryableStorageEntry<ApiType>`.
Further investigation required.


types/src/mint/index.ts
```diff
@@ -49,10 +51,10 @@ export class Mint
   implements IMint {}
 
 export const mintTypes: RegistryTypes = {
-  MintId: 'u64',
+  MintId,
   Mint,
-  MintBalanceOf: 'Balance',
-  BalanceOfMint: 'Balance',
```
mnaamani, line 55: Strictly speaking we want the underlying u128 to be inferred by the type of the Balance in the system. It is u128 but that is configured in the runtime. I doubt we will change it but as a general point how can we avoid explicitly using u128 here?

All currencies in our runtime use this type: https://github.com/Joystream/joystream/blob/master/runtime-modules/common/src/currency.rs#L11

The Balance trait type comes from the balances module:
https://github.com/Joystream/joystream/blob/master/runtime/src/lib.rs#L215

The primitive type for Balance is defined here: https://github.com/Joystream/joystream/blob/master/runtime/src/primitives.rs#L33

mnaamani:
Which version of the runtime did you run the `yarn generate:all` against? should i do it against the runtime built from `content_directory_second_try` branch?

Lezek123:
> Which version of the runtime did you run the `yarn generate:all` against? should i do it against the runtime built from `content_directory_second_try` branch?

Yes, I used `content_directory_second_try` branch.

mnaamani:
Ok great I've run the augment generate scripts and I'm getting the same results.

Looks good.

> Unfortunately it turned out that using the new augmentation in Pioneer will still be very tricky, because packages like `@polkadot/react-signer` etc. are often trying to access modules/methods that don't exist in Joystream chain

Thats too bad. There are a couple of modules that we might consider adding like Proxy and Multisig to our runtime because we could end up using them, but unfortunately it doesn't make sense for many of the other modules which we haven't included in our runtime. 

I can't say I understand exactly how you would implement the suggested solution for pioneer, but clearly if the fallback works then it is just as good as when we didn't have any of these definitions, and we at least cover the apis and types we do have in our runtime. 

It is certainly not worth the effort to try to fix the  typescript errors. I would rather we are able to pull from upstream if/when we need to.

So with this change I take it, there is no need to update the paths in tsconfig.json files of dependant packages, since the generated files are still in the same location? `types/augment-codec`

And pardon my ignorance, but why is there also a similar `types/augment/` and when would that be used instead of `types/augment-codec/` ?

Lezek123:
> Thats too bad. There are a couple of modules that we might consider adding like Proxy and Multisig to our runtime because we could end up using them, but unfortunately it doesn't make sense for many of the other modules which we haven't included in our runtime.

We could exclude some packages we don't need through `tsconfig` and `.eslintignore`, but this still leaves us with quite a few issues. I'll try to think about it, maybe I'll be able to find some quick and reasonable solution.

> So with this change I take it, there is no need to update the paths in tsconfig.json files of dependant packages, since the generated files are still in the same location? types/augment-codec

Yes, but now some projects may also want to add:
```
      "@polkadot/api/augment": ["path/to/types/augment-codec/augment-api.ts"]
```
to get the api qu(which I initially planned to do in Pioneer)

> And pardon my ignorance, but why is there also a similar types/augment/ and when would that be used instead of types/augment-codec/ ?

Inside `augment` we have the augmentation that `@polkadot/typegen` tool generates. It takes types like those inside `defs.json` as input and splits out typescript interfaces for them. Then it is able to decorate the api using those interfaces.

`augment-codec` basically replaces the interfaces automatically generated by `@polkadot/typegen` tool with our custom classes, which may have different or additional methods. I mostly added this with Pioneer in mind, because it would be hard to make it compatible with `augment` (ie. no `isOfType`, `getField` or any other custom methods would be recognized by typescript. There is also no exact `1:1` corelation in getters between `augment` and `augment-codec`, which could cause some hard to spot issues)

Lezek123:
Another thing about `agument` vs `augment-codec` is that:
- If someone uses `defs.json` file to register the types (which should be now be perfectly plausible), `agument` should be 100% compatible with that (this is generally the recommended way of registering types in `@polkadot/api` dependent projects)
- If someone uses `import { types } from '@joystream/types'` to register the types (like pretty much all our projects now), then `augment-codec` should now give 100% compatible typescript augmentation (including custom methods defined in our types classes)

mnaamani:
Excellent. Thanks for the detailed explanation. I feel we should probably compile all this usage and behavior information somewhere. Perhaps in the README of the types package?

The user developer experience is improving with each iteration of this tooling :+1:

PR #1461, opened 2020-09-24
Network Integration Tests - refactoring workflow

mnaamani:
Some basic cleanup of the scripts used to run the "network-tests".

- [x] Dropped use of ansible scripts in Github Actions, not adding any value for CI.
- [x] Tests wait for api and node to be "ready" before starting
- [x] Updating Dockerfile and docker-compose.yml files
- [x] Removed old runtime test code (only current runtime tests should be in the branch) and re-organized folder structure in network-test package
- [x] Fixed issue with properly locating shared `db.json` lowdb database file (used to share keys between tap test instances)
- [x] Relocated Docker related files to root to simplify context path and path to docker file in docker-compose files.


mnaamani:
Some things you can do/test more easily:

**Development Environment:**
Running a dev chain:
```
# Start a dev chain (will build joystream-node from source if local docker image not found)
docker-compose up -d

# Stop the chain
docker-compose down

# To rebuild from source and start new chain
docker-compose up -d --build
```

Running a dev chain and ipfs-node (for a test environment for storage-node)
```
# startup the chain and ipfs node
docker-compose --file docker-compose-with-storage.yml up -d

# init the chain with content dir and storge node provider
yarn storage-cli dev-init

# start a colossus server in dev mode
yarn colossus --dev
```

**New github workflows:**

- [joystream-node-checks](https://github.com/Joystream/joystream/actions?query=workflow%3Ajoystream-node-checks): just place holder - use it to checks for valid changes in runtime version + matching version change in Cargo.toml file
- [joystream-node-docker](https://github.com/mnaamani/joystream/actions?query=workflow%3Ajoystream-node-docker): on branch pushes (only on joystream repo) builds and caches a docker image of joystream/node for use in other workflows
- [run-network-tests](https://github.com/Joystream/joystream/actions?query=workflow%3Arun-network-tests): prepares joystream/node docker image (can reuse pre-built cached image from prior runs) then runs 3 jobs in parallel, which are the integration tests (runtime tests, query node and storage node). The integration test will run only if the label: `run-network-tests` is used on the pull request. This also effectively removes the need for our travis CI which was only used to run `cargo test, clippy, build` - now this flows does this for us.

mnaamani:
**Caching mechanism:**
joystream/node docker image generally takes around 1h to build so to speed up CI checks we save the docker image after it is built into a cached `~/docker-images` folder on the runners. A cache is created for each 'unique' docker image. The name of the cache is generated using a `shasum` over all the files that correspond to files used to build the image. A copy is also uploaded to docker hub (caches only retained for 7 days and limited to 5GB). This can be retrieved to "refresh" the cache in a future run.

This way after an image is created lets say on babylon branch, and PRs that are opened that do not modify the rust source code, do not need to rebuild the code, and can use the cached images to run joystream/node containers for their tests.


The joystream/node image produced in an workflow is available as an "artifact" and can be downloaded so you can run it locally to test your work locally against it (if someone is working on a new feature PR and you don't want to necessarily checkout their branch and build the node yourself)

eg. Download the joystream/node image artifact from this run: https://github.com/Joystream/joystream/actions/runs/281241904

Load it into your docker images locally:
```
unzip e9a49a97a8bea5b50ddd98f4ed6442d0ae4a2b1b-joystream-node-docker-image.tar.gz.zip
docker load --input joystream-node-docker-image.tar.gz

28c2d301a17f: Loading layer [==================================================>]  1.536kB/1.536kB
2d12a540f913: Loading layer [==================================================>]  54.65MB/54.65MB
ab79f50815f5: Loading layer [==================================================>]  2.082MB/2.082MB
92f3f89f913e: Loading layer [==================================================>]  14.84MB/14.84MB
Loaded image: joystream/node:latest
```
.. yes the .tar.gz.zip is ugly but correct (this is because I used the .tar.gz in the name of the artifact which was not necessary)

```
docker images

REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE
joystream/node             latest              e6a8b7e52819        14 hours ago        172MB
```

runtime/Cargo.toml
```diff
@@ -4,7 +4,7 @@ edition = '2018'
 name = 'joystream-node-runtime'
 # Follow convention: https://github.com/Joystream/substrate-runtime-joystream/issues/1
 # {Authoring}.{Spec}.{Impl} of the RuntimeVersion
-version = '7.5.1'
```
shamil-gadelshin, line 7: Why did you update the runtime version?

mnaamani: I was testing correct behavior when runtime code changes, with respect to rebuilding joystream/node image, and caching new images.

shamil-gadelshin (approved):
LGTM

Issue #1271, opened 2020-09-09
Make the infinite video grid work with non-infinite data


kdembler:
As in the title, update the infinite video grid once the mocked backend provides pagination metadata (#1268). There are some comments in the code that should help with achieving that

PR #1462, opened 2020-09-25
switch to relay pagination, improve the infinite scroller

### LINKED ISSUE #1268: Switch to relay-style pagination
Since it seems relay-style pagination is superior to what we're currently using and also will be supported by GraphQL framework Hydra is using we should switch to it.

As part of this task, we should do the following:
- Rewrite app logic that handles fetched data
- Update Apollo Client's cache behaviour (this may be the most complex part)
- Update MirageJS logic:
  - Use the new pagination style
  - Return paginated data from the DB instead of just repeating the dataset

Reference:
- https://www.apollographql.com/docs/react/data/pagination/#relay-style-cursor-pagination
- https://relay.dev/graphql/connections.htm
- https://github.com/goldcaddy77/warthog/issues/338

kdembler:
Resolves #1268 and #1271 

This PR moves Atlas to use Relay-style pagination and adds some improvements to the infinite scroller.

Notable changes:
- updated schema to match exactly how the one generated by the query node would look like
- cleaned up the MirageJS server code
- moved the query inside the infinite scroller component
- made infinite scroller component properly react to category change and handle end of data


fcescob (approved):
The InfiniteVideoGrid API improved a lot. Good Job!

bedeho (approved):

Issue #1402, opened 2020-09-18
Babylon: Create JSON input files for creating classes/schemas in the new content directory


Lezek123:
Create JSON `input` files for generating new content directory classes and schemas based on Atlas API expectations: https://github.com/Joystream/joystream/issues/824
Current (old) inputs reside in: https://github.com/Joystream/versioned-store-js/tree/master/src/inputs

This issue involves creating such JSON schemas for classes and schemas of objects like:
- `Language`
- `Channel`
- `Category`
- `MediaLocation`
- `VideoMediaEncoding`
- `License`
- `VideoMedia`
- `Video`
- etc.

bedeho:
Just remember that you need to take the current mocked API that the Atlas team is assuming will hold as the authoritative guide on the data model. The old schemas will have lots of junk that is not needed (e.g. related to music), and it will be missing other important stuff.

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1463

PR #1463, opened 2020-09-25
Content directory: JSON schemas and inputs + related tooling

Lezek123:
Implements https://github.com/Joystream/joystream/issues/1402 and https://github.com/Joystream/joystream/issues/1484

The documentation can be found in the `README` file: https://github.com/Lezek123/substrate-runtime-joystream/blob/content-directory-schemas/content-directory-schemas/README.md


Lezek123:
Fixed conflicts and some issues with importing files into projects that use different type augmentations (ie. CLI)

mnaamani:
Some really nice work in this PR.
Everything is working great. I ran the schema generation scripts, tested the tooling in VSCode by editing the input .json files with incorrect properties etc. :+1: 

Initialized the content dir with `initialize:dev` 

I have to say my favorite feature is how the entity batching input .json file supports 'new' and 'existing' entities.
Obviously after releasing babylon and we do an initial initialization of the content dir, members of the content working group will need to prepare a bunch of new batches of entities mostly. 
So batches .json syntax needs support for the 'existing' where you can either 

- specify the entity id directly
- similar unique lookup but the lookup would need to be done on chain if not found in the current batched operations.

One concern I just have is that on previous occasions we had issues if the number of operations to the `api.tx.contentDirectory.transaction` call was too large it caused the block production to fail when trying to include the extrinsic, the reason being that it takes too long to process the block.

We should try to split the batch into multiple transaction extrinsics and maybe do some performance testing to see what the maximum number of Operations can fit into one extrinsic.

content-directory-schemas/scripts/inputSchemasToEntitySchemas.ts
```diff
@@ -47,6 +44,7 @@ const ReferencePropertyDef = ({ Reference: ref }: ReferenceProperty): JSONSchema
   'oneOf': [
     onePropertyObjectDef('new', { '$ref': `./${ref.className}Entity.schema.json` }),
     onePropertyObjectDef('existing', { '$ref': `../entityReferences/${ref.className}Ref.schema.json` }),
+    PRIMITIVE_PROPERTY_DEFS.definitions.Uint64 as JSONSchema7,
```
mnaamani, line 47: I'm guessing this is the main change needed to reference existing entitites in the chain?

PR #1467, opened 2020-09-28
actions: make all linter checks --quiet

mnaamani:
Reduce amount of "noise" from linter warnings when reviewing a PR.

All the `checks` scripts now run the eslint checks last, so we can easily pass an optional `--quiet` arg to the linter.
Removed the `--quiet` arg from `lint` scripts so by default we see all linter errors, including warnings.

To supress warnings:
`yarn checks --quiet` 
or 
`yarn lint --quiet`

All github actions have been configured to run with `--quiet` arg


Lezek123 (approved):
Very nice improvement, I didn't notice any issues, all CI checks passed and the diffs don't contain any unrelated warnings anymore!

Issue #1290, opened 2020-09-10
Improve Searchbar animation and usability


fcescob:
Those are little improvements on the current Searchbar

- Animate the width transition when clicking

- Add a fallback for the cancel button inside the searchbar, it is only supported in Chrome right now, so add an element as a fallback for other browsers.

- Add keyboard shortcuts for getting out of focus of the spacebar and reverse the animation

PR #1477, opened 2020-09-28
Add Animation To SearchBar, Add CancelButton, Add Keyboard Shortcuts

fcescob:
This PR addresses #1290 . The searchbar design has also been corrected to resemble more closely the one on Figma.

### DIFF
--- a/atlas/src/components/LayoutWithRouting.tsx
+++ b/atlas/src/components/LayoutWithRouting.tsx
@@ -9,19 +9,21 @@ import routes from '@/config/routes'
 import { sizes } from '@/shared/theme'
 
 const LayoutWithRouting: React.FC = () => (
-  <MainContainer>
+  <>
     <GlobalStyle />
     <Router primary>
       <Navbar default />
     </Router>
-    <Router primary={false}>
-      <HomeView default />
-      <VideoView path={routes.video()} />
-      <SearchView path={routes.search()} />
-      <BrowseView path={routes.browse()} />
-      <ChannelView path={routes.channel()} />
-    </Router>
-  </MainContainer>
+    <MainContainer>
+      <Router primary={false}>
+        <HomeView default />
+        <VideoView path={routes.video()} />
+        <SearchView path={routes.search()} />
+        <BrowseView path={routes.browse()} />
+        <ChannelView path={routes.channel()} />
+      </Router>
+    </MainContainer>
+  </>
 )
 
 const MainContainer = styled.main`
--- a/atlas/src/components/Navbar/Navbar.style.tsx
+++ b/atlas/src/components/Navbar/Navbar.style.tsx
@@ -4,15 +4,9 @@ import { Searchbar, Icon } from '@/shared/components'
 import { colors, sizes } from '@/shared/theme'
 import { ReactComponent as UnstyledLogo } from '@/assets/logo.svg'
 
-export const Header = styled.header<{ isSearching: boolean }>`
-  display: grid;
-  grid-template-columns: ${(props) => (props.isSearching ? `134px 1fr 134px` : `repeat(3, 1fr)`)};
-  grid-template-areas: ${(props) => (props.isSearching ? `". searchbar cancel"` : `"navigation searchbar ."`)};
-  width: 100%;
-  padding: ${(props) => (props.isSearching ? `${sizes.b2}px` : `${sizes.b3}px ${sizes.b8}px`)};
-  border-bottom: 1px solid ${colors.gray[800]};
-  background-color: ${(props) => (props.isSearching ? colors.gray[900] : colors.black)};
-`
+type NavbarStyleProps = {
+  hasFocus: boolean
+}
 
 export const Logo = styled(UnstyledLogo)`
   width: ${sizes.b12}px;
@@ -21,7 +15,6 @@ export const Logo = styled(UnstyledLogo)`
 
 export const NavigationContainer = styled.div`
   display: flex;
-  grid-area: navigation;
   align-items: center;
   > * + * {
     margin-left: ${sizes.b6}px;
@@ -29,10 +22,13 @@ export const NavigationContainer = styled.div`
 `
 
 export const StyledSearchbar = styled(Searchbar)`
-  width: 100%;
-  grid-area: searchbar;
+  transition: width 0.6s cubic-bezier(0.165, 0.84, 0.44, 1);
+  will-change: width;
 `
-
+export const SearchbarContainer = styled.div`
+  display: flex;
+  justify-content: center;
+`
 export const StyledIcon = styled(Icon)`
   color: ${colors.gray[300]};
   &:hover {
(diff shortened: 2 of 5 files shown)


atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
-  const [search, setSearch] = useState('')
-  const [isSearching, setIsSearching] = useState(false)
+type NavbarProps = RouteComponentProps
 
-  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    setSearch(e.currentTarget.value)
-  }
+const Navbar: React.FC<NavbarProps> = () => {
+  const [search, setSearch] = useState('')
+  const [focus, setFocus] = useState(false)
```
kdembler: Maybe something like `isFocused` would be more descriptive about what this variable holds?

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
 
   const handleKeyPress = (e: React.KeyboardEvent<HTMLInputElement>) => {
-    if (e.key === 'Enter' || (e.key === 'NumpadEnter' && search.trim() !== '')) {
+    if (e.key === 'Enter' || (e.key === 'NumpadEnter' && search.trim())) {
```
kdembler: I don't think I understand this condition. We will navigate to search results if any of the following is true:
1) user pressed `Enter`
2) user pressed `NumpadEnter` **and** search isn't empty.

Is that the desired behaviour? Why do we have an additional condition only in the second case?

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
       navigate(routes.search(search))
     }
+    if (e.key === 'Escape' || e.key === 'Delete' || e.key === 'Esc') {
```
kdembler: Couple of issues I see here:
1) Why would we want to exit search on `Delete` press? That's a regular key used to edit the field value, like `Backspace`
2) The `Escape` doesn't work for me on Chrome unfortunately. It does in FF, but seems Chrome handles this somewhat differently. Maybe it'd make sense to investigate. Seems the input fields gets blurred on `Escape` press and our handler doesn't get fired

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
+      setFocus(false)
+      setSearch('')
```
kdembler: Maybe it'd make sense to also blur the input?

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
+    }
+  }
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (search !== '' && e.currentTarget.value === '') {
+      setFocus(false)
```
kdembler: I wonder whether we actually should do that. When you first click the searchbar, it doesn't have any value but is in the focused state. Exitting that state when the input gets emptied seems like an inconsistency to me. What do you think?

atlas/packages/app/src/components/Navbar/Navbar.tsx
```diff
@@ -1,60 +1,87 @@
...
-        onKeyPress={handleKeyPress}
-        onFocus={handleFocus}
-      />
-      {isSearching && (
-        <CancelButton onClick={handleCancel}>
-          <Icon name="times" />
-        </CancelButton>
-      )}
+    <Header hasFocus={focus}>
+      <div>
```
kdembler: Do we need that `div` here? It will be empty when `focus` is `true`

kdembler (changes requested):
A couple of comments mostly regarding the behavior of the search bar. Also, maybe we could add simple transitions to the navbar background color and height (although this would probably require specifying the height explicitly). That would make the transition between states more seamless.

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
```diff
@@ -1,29 +1,57 @@
...
   placeholder,
   onChange,
   onFocus,
-  value,
+  onCancel,
+  controlled = false,
+  value: externalValue,
   onBlur,
   onSubmit,
   ...hmtlProps
```
kdembler: `hmtl` -> `html`

kdembler:
There's also one more thing I've just noticed - when clicking on a video from the search results, the search query persists in the search bar. Maybe we should remove that?

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: I was indecisive on this one, but I decided to pull it in, after reading your thoughts I am leaning on removing it. It could give some eyestrain since the searchbar is constantly jumping between expanded and not expanded,  for example if you mistype the first character of your query

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: Auto-format grouped them, I changed it to be:
1. User pressed `Enter` **or** `NumpadeEnter`
2. **And** search isn't empty

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: I've added those divs to make animating the `Searchbar` easier, they allow me to keep the same grid structure even when one element is missing, having the elements themselves instead of enclosing them in divs makes the `Searchbar`  jump around (it is the 2nd column when unfocused, 1st when focused). An alternative was to hide the `CancelButton` (or the `NavButtons` respectively) when `Searchbar` is unfocused, but that looked to me like a worse solution.

fcescob:
> There's also one more thing I've just noticed - when clicking on a video from the search results, the search query persists in the search bar. Maybe we should remove that?

Youtube does this as well, I would keep it if there is not big reason for removing it.

atlas/packages/app/src/shared/components/Searchbar/Searchbar.tsx
fcescob: Thanks fixed.

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: Good idea, I will add that.

atlas/packages/app/src/components/Navbar/Navbar.tsx
fcescob: Good catch, changed it.

atlas/packages/app/src/components/Navbar/Navbar.tsx
kdembler: Okay, let's keep it

atlas/packages/app/src/components/Navbar/Navbar.tsx
kdembler: FYI, this probably wasn't because of the formatter. My guess is your original condition looked like so: `e.key === 'Enter' || e.key === 'NumpadEnter' && search.trim()`. However, the `AND` operator has higher precedence than the `OR` so when formatted added grouping around `e.key === 'NumpadEnter' && search.trim()` it was only to improve the readability of what the condition actually represents, the logic remained intact.

atlas/packages/app/src/components/Navbar/Navbar.tsx
kdembler: Regarding the comment you've sent me on the chat: the second point still isn't really resolved, because even though we have `blur` now, the behaviour still differs between Chrome and Firefox - on Firefox on `Esc` the field will empty out and blur and on Chrome it will only blur. We should keep that consistent

atlas/packages/app/src/components/Navbar/Navbar.tsx
kdembler: I think this still isn't changed?

kdembler:
> Youtube does this as well, I would keep it if there is not big reason for removing it.

All rightey.

Just noticed that when typing, there are two "close" buttons available, one inside the text field and one on the far right of the navbar. Do we need both? It can be confusing for the user to figure out which one to click

kdembler (approved):
LGTM

bedeho (approved):

Issue #1423, opened 2020-09-21
Add fading opacity animation to placeholders


fcescob:
Add a fading opacity animation to placeholders as seen in the design.

Issue #1416, opened 2020-09-19
Fade in from loader to image 


bedeho:
Currently the transition from loading to showing assets is instant, this is a bit rough, in particular when you have a full screen of loading assets. Could we introduce a fast transition where the media asset fades in with increasing opacity? Just play around with parameters to find a value you like, it may need to be different for different scenes.

PR #1480, opened 2020-09-28
Add Placeholder animation

fcescob:
This PR addresses #1416 and #1423

### DIFF
--- a/atlas/src/shared/components/Placeholder/Placeholder.tsx
+++ b/atlas/src/shared/components/Placeholder/Placeholder.tsx
@@ -1,3 +1,4 @@
+import { keyframes } from '@emotion/core'
 import styled from '@emotion/styled'
 import { colors } from '@/shared/theme'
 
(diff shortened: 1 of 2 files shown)


atlas/packages/app/src/shared/components/VideoPreview/VideoPreviewBase.styles.tsx
```diff
@@ -19,6 +28,7 @@ export const CoverContainer = styled.div`
   transition-timing-function: cubic-bezier(0.165, 0.84, 0.44, 1);
 
   position: relative;
+  animation: ${fadeIn} 2s ease-in -0.5s;
```
kdembler: This feels too long, I'd actually aim for something <0.5s

kdembler:
One more thing I've noticed: the image entry blur doesn't happen only when finishing the loading. For example, if you go to a video playback view to watch a video and then move back to the home view, the animation still plays out even though the items are already loaded. It would probably make sense to not animate the previews in that case

kdembler (approved):
LGTM

bedeho (approved):

Issue #1294, opened 2020-09-10
Make all elements clickable


fcescob:
As mentioned by @kdembler  in #824 some elements like the video previews on the infinite grid are not clickable. 
A possible fix is to add a wrapper around those elements.

PR #1482, opened 2020-09-28
Make All Elements Clickable

fcescob:
This PR addresses #1294


kdembler:
Please add the atlas label in the future or request a review, this has been open for a week and I just discovered it accidentally

kdembler:
For the review itself: the approach you've taken will require us to add click handlers for all the places where we use respective previews. This will make us duplicate the logic and make the potential future refactor more complicated.

What do you think about adding some Atlas-specific wrapper elements for `*Preview` components that would contain the navigation logic? Then all the views wouldn't need to worry about navigating but just use these components.

atlas/src/components/VideoGrid.tsx
```diff
@@ -18,10 +19,13 @@ const VideoGrid: React.FC<VideoGridProps> = ({ videos }) => {
     <Grid>
       {videos.map((v, idx) => (
         <StyledVideoPreview
+          id={v.id}
+          channelId={v.channel.id}
           key={idx}
           title={v.title}
           channelName={v.channel.handle}
           channelAvatarURL={v.channel.avatarPhotoURL}
+          onChannelClick={() => navigate(routes.channel(v.channel.id))}
```
kdembler: Do we still need that?

atlas/src/shared/components/InfiniteVideoGrid/InfiniteVideoGrid.tsx
```diff
@@ -15,6 +17,7 @@ type InfiniteVideoGridProps = {
   ready?: boolean
   className?: string
 }
+console.log(Grid)
```
kdembler: 😛

atlas/src/shared/components/InfiniteVideoGrid/InfiniteVideoGrid.tsx
```diff
@@ -156,5 +161,8 @@ const StyledVideoPreviewBase = styled(VideoPreviewBase)`
   margin: 0 auto;
   width: 100%;
 `
+const StyledGrid = styled(Grid)`
+  padding-top: ${sizes.b2}px;
```
kdembler: Why this change? In the design there is a 16px margin between title and the grid, like it is right now

atlas/src/components/VideoGrid.tsx
fcescob: Nice catch, removed.

atlas/src/shared/components/InfiniteVideoGrid/InfiniteVideoGrid.tsx
fcescob: Removed :D

kdembler (approved):
LGTM!

bedeho (approved):

PR #1485, opened 2020-09-29
atlas cleanup

kdembler:
Changes as per commits:
- remove atlas multi-package structure, update README
– remove unused packages
- remove icon prop from TextField - it was the only instance where we were using the old `@fortawesome` packages. If we want to use icons with the text field, we will need to rework that a bit anyway
- remove pre-push test hook


atlas/src/shared/components/TextField/TextField.tsx
```diff
@@ -1,25 +1,15 @@
-import React, { useState, useRef, useEffect } from 'react'
-import { useCSS, TextFieldStyleProps } from './TextField.style'
-import { IconProp } from '@fortawesome/fontawesome-svg-core'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import React, { useEffect, useRef, useState } from 'react'
```
fcescob: We never use this component anywhere, and if we decided to use it, it would need to reworked anyway. I suggest we just get rid of it for now

kdembler: Removed 👍

fcescob (approved):
LGTM

bedeho (approved):

PR #1490, opened 2020-09-30
Multiple working team pallet improvements.

### LINKED ISSUE #1479: Add working group document
# Background

In practice, there is a need for the lead to broadcast some public mutable information to the working group, wider community, and the world, on various subjects. This is stuff like.

1. A one line status message on the subsystem.
2. New upcoming expected positions.
3. A link to a subsection of the forum devoted to the subsystem.
4. A message feed including information updates.

# Proposal

A new storage field is introduced which holds an updatable hash, set by the working group lead, of some document provided through an extrinsic, or a council proposal. The document should follow some standard encoding the runtime does not care about, which allows user facing applications to properly render a rich experience base on it.

shamil-gadelshin:
Changes:
- change membership validity check to `ensure_actor_origin`
- pay when we can unpaid money automatically
- slashing with a sum on termination
- slashing text on slashing and termination with slashing
- [new storage field which is set by leader](https://github.com/Joystream/joystream/issues/1479)
- `staking_account_id` should be demanded only when an opening has stakes
- pay unpaid money during the worker deactivation (termination or leaving)
- add missing extrinsics: `update_reward_account` and `update_reward_amount`
- payment should be once in several blocks
- if stakes decrease to zero we should prevent unstaking period to happen

### DIFF
--- a/runtime-modules/working-team/src/checks.rs
+++ b/runtime-modules/working-team/src/checks.rs
@@ -11,7 +11,7 @@ use sp_arithmetic::traits::Zero;
 use sp_std::collections::btree_set::BTreeSet;
 use system::{ensure_root, ensure_signed};
 
-use crate::types::ApplicationInfo;
+use crate::types::{ApplicationInfo, StakeParameters};
 
 // Check opening: verifies origin and opening type compatibility.
 pub(crate) fn ensure_origin_for_opening_type<T: Trait<I>, I: Instance>(
@@ -224,25 +224,24 @@ pub(crate) fn ensure_valid_reward_policy<T: Trait<I>, I: Instance>(
 
 // Check application: verifies free balance for a given account for staking.
 pub(crate) fn ensure_enough_balance_for_staking<T: Trait<I>, I: Instance>(
-    account: &T::AccountId,
-    stake: &Option<BalanceOfCurrency<T>>,
+    stake_parameters: Option<StakeParameters<T::AccountId, BalanceOfCurrency<T>>>,
 ) -> DispatchResult {
-    let stake_balance = stake.unwrap_or_default();
+    if let Some(sp) = stake_parameters {
+        if sp.stake > (Zero::zero()) {
+            let free_balance = T::Currency::free_balance(&sp.staking_account_id);
 
-    if stake_balance > (Zero::zero()) {
-        let free_balance = T::Currency::free_balance(account);
+            if free_balance < sp.stake {
+                return Err(Error::<T, I>::InsufficientBalanceToCoverStake.into());
+            } else {
+                let new_balance = free_balance - sp.stake;
 
-        if free_balance < stake_balance {
-            return Err(Error::<T, I>::InsufficientBalanceToCoverStake.into());
-        } else {
-            let new_balance = free_balance - stake_balance;
-
-            return T::Currency::ensure_can_withdraw(
-                account,
-                stake_balance,
-                WithdrawReasons::all(),
-                new_balance,
-            );
+                return T::Currency::ensure_can_withdraw(
+                    &sp.staking_account_id,
+                    sp.stake,
+                    WithdrawReasons::all(),
+                    new_balance,
+                );
+            }
         }
     }
 
@@ -252,14 +251,15 @@ pub(crate) fn ensure_enough_balance_for_staking<T: Trait<I>, I: Instance>(
 // Check application: verifies that proposed stake is enough for the opening.
 pub(crate) fn ensure_application_stake_match_opening<T: Trait<I>, I: Instance>(
     opening: &JobOpening<T::BlockNumber, BalanceOfCurrency<T>>,
-    stake: &Option<BalanceOfCurrency<T>>,
+    stake_parameters: &Option<StakeParameters<T::AccountId, BalanceOfCurrency<T>>>,
 ) -> DispatchResult {
     let opening_stake_balance = opening
         .stake_policy
         .clone()
         .unwrap_or_default()
         .stake_amount;
-    let application_stake_balance = stake.unwrap_or_default();
+
+    let application_stake_balance = stake_parameters.clone().unwrap_or_default().stake;
 
     if application_stake_balance < opening_stake_balance {
         return Err(Error::<T, I>::ApplicationStakeDoesntMatchOpening.into());
--- a/runtime-modules/working-team/src/checks.rs
+++ b/runtime-modules/working-team/src/checks.rs
@@ -1,6 +1,6 @@
 use crate::{
-    BalanceOfCurrency, Instance, JobOpening, JobOpeningType, MemberId, StakePolicy, TeamWorker,
-    TeamWorkerId, Trait,
+    BalanceOfCurrency, Instance, JobOpening, JobOpeningType, MemberId, RewardPolicy, StakePolicy,
+    TeamWorker, TeamWorkerId, Trait,
 };
 
 use super::Error;
@@ -195,6 +195,20 @@ pub(crate) fn ensure_valid_stake_policy<T: Trait<I>, I: Instance>(
         ensure!(
             stake_policy.stake_amount != Zero::zero(),
             Error::<T, I>::CannotStakeZero
+        )
+    }
+
+    Ok(())
+}
+
+// Check opening: verifies reward policy for the opening.
+pub(crate) fn ensure_valid_reward_policy<T: Trait<I>, I: Instance>(
+    reward_policy: &Option<RewardPolicy<BalanceOfCurrency<T>>>,
+) -> Result<(), DispatchError> {
+    if let Some(reward_policy) = reward_policy {
+        ensure!(
+            reward_policy.reward_per_block != Zero::zero(),
+            Error::<T, I>::CannotRewardWithZero
         )
     }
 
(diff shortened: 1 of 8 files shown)


runtime-modules/working-team/src/checks.rs
```diff
@@ -203,25 +224,24 @@ pub(crate) fn ensure_valid_stake_policy<T: Trait<I>, I: Instance>(
 
 // Check application: verifies free balance for a given account for staking.
 pub(crate) fn ensure_enough_balance_for_staking<T: Trait<I>, I: Instance>(
```
bedeho: What is the difference between this routine and `StakingHandler::is_enough_balance_for_stake` ?

runtime-modules/working-team/src/types.rs
```diff
@@ -181,4 +251,60 @@ pub trait StakingHandler<T: system::Trait + membership::Trait + GovernanceCurren
...
+    /// Stake information for the application.
+    pub stake_parameters: Option<StakeParameters<AccountId, Balance>>,
+}
+
+/// Contains information for the stakes when applying for opening.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Debug, Clone, Default, PartialEq, Eq)]
+pub struct StakeParameters<AccountId, Balance> {
+    /// Stake balance.
+    pub stake: Balance,
```
bedeho, line 287: Why is this provided? Every applicant _must_ stake the exact same amount.

runtime-modules/working-team/src/types.rs
```diff
@@ -107,37 +133,81 @@ impl<AccountId: Clone, MemberId: Clone> Application<AccountId, MemberId> {
...
+
+    /// Reward account id.
+    pub reward_account_id: AccountId,
+
+    /// Specifies the block when the worker chose to leave.
+    pub started_leaving_at: Option<BlockNumber>,
+
+    /// Unstaking period when the worker chooses to leave the role.
+    /// It is defined by the job opening.
+    pub job_unstaking_period: BlockNumber,
```
bedeho, line 154: 1. Given that this unstaking period only applies to leaving, perhaps `leaving_unstaking_period` is sharper.

2. This will be 0 if there was no stake to begin with?

runtime-modules/working-team/src/lib.rs
```diff
@@ -695,4 +918,115 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
         let slashed_balance = T::StakingHandler::slash(staking_account_id, balance);
         Self::deposit_event(RawEvent::StakeSlashed(worker_id, slashed_balance));
     }
+
+    // Reward a worker using reward presets and working team budget.
+    fn reward_worker(worker_id: &TeamWorkerId<T>, worker: &TeamWorker<T>) {
```
bedeho, line 957: Perhaps you can simply this by factoring out considerations related to spending from the budget in general, as this is something that will be reused also with the extrinsic the lead can trigger to spend out of the budget on discretion.

runtime-modules/working-team/src/lib.rs
```diff
@@ -670,7 +886,8 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
     }
 
     // Fires the worker. Unsets the leader if necessary. Decreases active worker counter.
-    fn deactivate_worker(worker_id: &TeamWorkerId<T>, worker: &TeamWorker<T>) {
+    // Deposits an event.
+    fn deactivate_worker(worker_id: &TeamWorkerId<T>, worker: &TeamWorker<T>, event: Event<T, I>) {
```
bedeho: Deactivation is not a clear term in the current implementation, as its no longer a state, so perhaps we can just call it `remove_worker`. Also, perhaps event submission should be kept outside, because this allows sending any random event, not removal events only. Caller can send contextual event?

runtime-modules/working-team/src/lib.rs
```diff
@@ -695,4 +918,115 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+                // Update worker missed reward.
+                WorkerById::<T, I>::mutate(worker_id, |worker| {
+                    worker.missed_reward = None;
+                });
+            }
+        }
+    }
+
+    // Returns a collection of workers with active unstaking period.
+    fn get_leaving_workers() -> Vec<WorkerInfo<T>> {
```
bedeho: Leaving worker implies its everyone leaving, that is they are unstaking from a leaving operation, but really this is those who have completed their unstaking from leaving precisely in this block. So perhaps something like `workers_just_done_unstaking`.

runtime-modules/working-team/src/lib.rs
```diff
@@ -695,4 +918,115 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+
+        // Don't reward at genesis.
+        if current_block == Zero::zero() {
+            return false;
+        }
+
+        let reward_period = T::RewardPeriod::get();
+
+        // If reward period is not set or equals one - reward every block.
+        if reward_period <= One::one() {
```
bedeho: I am not sure I follow why this precheck is needed? Type of `reward_period` is `u32`, so worst case it is `0`, so to staticaly avoid we can just change its semantics to allow us to do `current_block % (reward_period.into() + 1) == Zero::zero()`. So if you want to pay out every 100 blocks, you pass in value 99. This side steps issue of modulo being undefined for 0.

runtime-modules/working-team/src/lib.rs
```diff
@@ -695,4 +918,115 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+                // Update worker missed reward.
+                WorkerById::<T, I>::mutate(worker_id, |worker| {
+                    worker.missed_reward = Some(new_missed_reward);
+                });
+            }
+        }
+    }
+
+    // Tries to pay missed reward if the reward is enabled for worker and there is enough of team budget.
+    fn try_to_pay_missed_reward(worker_id: &TeamWorkerId<T>, worker: &TeamWorker<T>) {
```
bedeho: I think it makes sense to try to pay as much as you can, rather than making it all or nothing. For example, if a worker is leaving, it does not make sense to have them lose _all_ their owed money, just because you could not pay some of it. I would say that applies to normal payouts also. It also discouraged debts from growing extremely large to some extent.

runtime-modules/working-team/src/lib.rs
bedeho: In general actually, this whole routine could do with getting simplified a bit.

bedeho (changes requested):
Overall this looks very good, the module looks very crisp and clean, can't tell you how much more confident this makes me that we have no serious bugs, and have good maintainabilty, compared to the module hierarchy we had before.

A few notes:

- Can you implement leader spending out of budget, its a small addition, and it meshes well with required change to reward payout anyway. 

- Can I confirm that the lock on the staking account of a worker exists, through its life cycle, always and only when the opening initially had a staking requirement? It presence is not impacted by sequences of reductions and increases in stake changes.

- Almost always we are using optional wrappers around integers, which both gives duplicate semantics for nothing: `None`/0, and makes the code overall slightly more complex in many places in having to deal with unwrapping, or manually enforcing that `Some(x)` for `x==0` is not allowed. I see no benefits to this.

runtime-modules/working-team/src/types.rs
shamil-gadelshin: 1. Ok.
2. Yes.

runtime-modules/working-team/src/types.rs
shamil-gadelshin: We agreed that it's possible to have a stake greater than required.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Changed: Only `zero` now is a special case.

runtime-modules/working-team/src/checks.rs
shamil-gadelshin: It was an artifact. I removed it.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Renamed it. But kept the event variable - there 3 events that could be throw, if I remove this variable - I should duplicate twice `deposit_event` call.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Renamed to `get_workers_with_finished_unstaking_period`

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Created `try_to_pay_from_budget` private method.

runtime-modules/working-team/src/lib.rs
shamil-gadelshin: Done.

runtime-modules/working-team/src/lib.rs
```diff
@@ -774,19 +777,43 @@ decl_module! {
...
+            // Ensure team leader privilege.
+            checks::ensure_origin_is_active_leader::<T,I>(origin)?;
+
+            ensure!(amount > Zero::zero(), Error::<T, I>::CannotSpendZero);
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            Self::try_to_pay_from_budget(&account_id, amount)?;
```
bedeho: I don't love this ? check after mutation check. Can we not avoid this? if the check creates imbalance for example?

bedeho (changes requested):
Looks good, just one last point.

bedeho (approved):
Looks good!

Issue #1276, opened 2020-09-10
Orion initial implementation


kdembler:
### Functionality
We keep it bare-bones for now and possibly expand in the future. This means:
- Simple GraphQL API allowing to:
  - Add one view count to a specific video
  - Get a view count for a specific video
  - (Possibly) get a batch view counts for a list of videos
- No authentication of any kind for now

### API
```gql
type VideoInfo {
  id: ID!
  views: Int!
}

### API
type Query {
  # Get views count for a single video
  videoViews(videoID: ID): VideoInfo

  # Get views counts for a list of videos
  batchedVideoViews(videoIDList: [ID!]!): [VideoInfo]!
}

type Mutation {
  # Add a single view to target video's count
  addVideoView(videoID: ID!): VideoInfo!
}
```

PR #1493, opened 2020-09-30
initial Orion functionality

kdembler:
This PR adds Orion with its initial functionality as described in #1276. It uses Apollo Server, Express and MongoDB to save view counts for the specified video ids.


orion/schema.graphql
```diff
@@ -0,0 +1,18 @@
...
+# !!! THIS FILE WAS GENERATED BY TYPE-GRAPHQL !!!
+# !!!   DO NOT MODIFY THIS FILE BY YOURSELF   !!!
+# -----------------------------------------------
+
+type Mutation {
+  addVideoView(videoID: ID!): VideoInfo!
+}
+
+type Query {
+  batchedVideoViews(videoIDList: [ID!]!): [VideoInfo!]!
```
fcescob: I think this array might be sparse, for example if there are some IDs which don't exist in the list, also as it is right now it is not consistent with `VideoViews` which can be `null`, if there is a good reason for `batchedVideoViews` to be defined like it is then I think we should change `videoViews` to be consistent with it

orion/src/resolvers/videoInfos.ts
```diff
@@ -0,0 +1,42 @@
...
+    return VideoInfoModel.find({
+      'videoID': {
+        $in: videoIDList,
+      },
+    })
+  }
+
+  @Mutation(() => VideoInfo)
+  async addVideoView(@Args() { videoID }: AddVideoViewArgs) {
+    return VideoInfoModel.findOneAndUpdate({ videoID: videoID }, { $inc: { views: 1 } }, { new: true, upsert: true })
```
fcescob: I think we can use the shorthand syntax here for `videoID`

fcescob (changes requested):
Some minor things to fix, but otherwise looks good.

orion/schema.graphql
kdembler: Yeah, I spent some time thinking about how this should look like. My initial approach was the one you're proposing I think: return `[VideoInfo]!`.

However, I figured that I'm not sure if this makes sense. The issue is that returning `null` doesn't really say anything. I think the client shouldn't rely just on the order of the returned array. Parsing the result will actually require doing a search - go through the results and match the video IDs we're looking for with the returned `VideoInfo`. In that case, returning additional `null` values from the resolver doesn't really bring any benefit or additional info.

For the `videoViews` I think it actually makes sense - with that endpoint you're asking for a specific ID so you either get a result or `null`. I guess I see those two endpoints as having a different purpose - `videoViews` is asking for something specific, while the `batchedVideoViews` is more like a search.

Let me know if that makes sense or whether you're like to discuss this on a call. Based on the result of our discussion I'll add some comments to those resolvers to better explain the rationale behind them.

orion/src/resolvers/videoInfos.ts
kdembler: Yes, cleaned that up

fcescob (approved):
LGTM

bedeho (approved):

PR #1500, opened 2020-10-02
Tokenomics Stats Page

DzhideX:
Notes:

* I had everything in one function but I thought it'd be better to seperate them into seperate function as it is _much_ more readable compared to what I think should be a trivial(?) performance hit. If we notice that it's considerably slow, we could always make it more performant. I'm open for suggestions!
 
* There are two functions in here which check for different proposal types but have identical functionality. I kept it this way in case we want to add some differing functionality in the future, if not, this is easily changed!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703467049498) by [Unito](https://www.unito.io)


bwhm (changes requested):
As discussed before, I'm not sure whether my idea of including the last line of proposal statistics -`Tokens Created/Burned` - works well. I would suggest:
- Cut this one (perhaps comment out the affected lines of code)
- Add question marks that explains which proposals belongs to which groups

Thoughts?

DzhideX:
In the meantime I've also worked on the historical proposals which we didn't have burned tokens for and by taking it out from the first table it would be more in-line with the second one, meaning much easier to implement. Therefore:

1. I fully agree with the decision, but I think it wouldn't be a problem to refactor the code as a whole(not only commenting unused lines), because I had to make certain sacrifices which can now be fixed making the code much more simple.

2. I don't fully understand what you mean by this?

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
       }
     };
   }
+
+  async networkStatistics () {
+    const blockHeight = (await this.api.derive.chain.bestNumber()).toNumber();
+    const numberOfMembers = (await this.api.query.members.nextMemberId() as MemberId).toNumber();
+    const content = (await this.api.query.versionedStore.classById(7) as Class).properties;
+    const numberOfChannels = (await this.api.query.contentWorkingGroup.nextChannelId() as ChannelId).toNumber();
+    const proposalCount = Number(await this.api.query.proposalsEngine.proposalCount());
```
Lezek123: It's probably better (and more consistent) to use `(await this.api.query.proposalsEngine.proposalCount() as u32).toNumber()`

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
+    const numberOfForumCategories = (await this.api.query.forum.nextCategoryId() as CategoryId).toNumber();
+    const numberOfForumPosts = (await this.api.query.forum.nextPostId() as PostId).toNumber();
+    const [councilMintId, contentCuratorMintId, storageMintId] = await Promise.all([await this.api.query.council.councilMint() as MintId, await this.api.query.contentWorkingGroup.mint() as MintId, await this.api.query.storageWorkingGroup.mint() as MintId]);
+    const [councilMint, contentMint, storageMint] = (await this.api.query.minting.mints.multi<Mint>([councilMintId, contentCuratorMintId, storageMintId]));
+
+    return {
+      blockHeight,
+      numberOfMembers,
+      content: content.length,
```
Lezek123: If this is supposed to represent the number of videos on the platform, then it is not correct.

One way to find out this number (which unfortunately takes quite a lot of time) would be to use `this.api.query.versionedStore.entityById.entries()` and count all entities where `classId === 7` (for Babylon that would then need to be changed to `contentDirectory.entityById`)

The other way would be to use `this.api.query.dataDirectory.knownContentIds` and get the `Vec` length. This would probably include videos that have no associated entities (and may not be visible on the platform), but it would be incomparably faster (that's the query that our status server executes to get this number)

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
   }
+
+  async networkStatistics () {
+    const blockHeight = (await this.api.derive.chain.bestNumber()).toNumber();
+    const numberOfMembers = (await this.api.query.members.nextMemberId() as MemberId).toNumber();
+    const content = (await this.api.query.versionedStore.classById(7) as Class).properties;
+    const numberOfChannels = (await this.api.query.contentWorkingGroup.nextChannelId() as ChannelId).toNumber();
+    const proposalCount = Number(await this.api.query.proposalsEngine.proposalCount());
+    const numberOfForumCategories = (await this.api.query.forum.nextCategoryId() as CategoryId).toNumber();
+    const numberOfForumPosts = (await this.api.query.forum.nextPostId() as PostId).toNumber();
```
Lezek123: Both forum categories and forum posts have ids that start from `1`, so you need to substract `1` from this number.
For reference you can try executing:
```
console.log(await api.query.forum.nextPostId())
console.log(await api.query.forum.nextCategoryId())
console.log((await api.query.forum.postById.keys()).length)
console.log((await api.query.forum.categoryById.keys()).length)
```
In the JavaScript playground (https://testnet.joystream.org/#/js)

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
+    const [councilMintId, contentCuratorMintId, storageMintId] = await Promise.all([await this.api.query.council.councilMint() as MintId, await this.api.query.contentWorkingGroup.mint() as MintId, await this.api.query.storageWorkingGroup.mint() as MintId]);
```
Lezek123: Split overly long lines into multiple ones (ie. each promise can be in a separate line)

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      councilMintSpent: councilMint.total_minted.toNumber(),
+      contentCuratorMintCapacity: contentMint.capacity.toNumber(),
+      contentCuratorMintSpent: contentMint.total_minted.toNumber(),
+      storageProviderMintCapacity: storageMint.capacity.toNumber(),
+      storageProviderMintSpent: storageMint.total_minted.toNumber()
+    };
+  }
+
+  async proposalStatistics () {
+    const proposal = {
```
Lezek123: **Minor:** Rename to `proposalCounters`?
Also we don't seem to have `Vetoed` proposals

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      returnData[definedProposalType][proposalStatus]++;
+      returnData.all.all++;
+      returnData.all[proposalStatus]++;
+    });
+
+    return returnData;
+  }
+
+  historicalProposalStatistics () {
+    const proposal = {
```
Lezek123: I'm guessing that's the duplicate code that you mentioned in PR description.
I understand this is to prevent issues in case proposal module changes in the runtime?

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,125 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      text: { ...proposal },
+      spending: { ...proposal },
+      workingGroups: { ...proposal },
+      networkChanges: { ...proposal },
+      all: { ...proposal }
+    };
+
+    HISTORICAL_PROPOSALS.map(({ proposal }) => {
+      let proposalType: 'text' | 'spending' | 'networkChanges' | 'workingGroups';
+      const proposalStatus = proposal.status.Finalized
```
Lezek123, line 403: You could use `createType('ProposalStatus', proposal.status)` and treat it the same way as current proposal statuses but as mentioned before, maybe this way is actually more furute-proof in case there will be some backward-incompatible changes in the runtime (though that would cause lots of other issues in Pioneer)

Lezek123 (changes requested):
I requested a few changes, I'm also considering what would be the best practice of splitting `transport`/`types`/`consts` etc. (since, for example, `ProposalStatistics` could belong to both `proposals` and `tokenomics`), but I'm not clear on that yet, so for now we can probably leave it as is and see how this works out when doing upgrades etc.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
DzhideX: Vetoed proposals weren't in the Wireframe so I just assumed we didn't want them in there. The implementation of these should be fairly trivial though. It's up to @bwhm .

pioneer/packages/joy-utils/src/transport/tokenomics.ts
DzhideX: I'll answer for both this one and the last comment on here as they are related. In my opinion due to `maybe this way is actually more furute-proof in case there will be some backward-incompatible changes in the runtime ` I would rather keep it as is. If you've changed your mind in the meantime and think we should try to implement it all in one function let me know!

DzhideX:
Done!

I also realized that it wasn't properly counting the `contentCuratorLeadNumber` and quickly fixed that as well.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,133 @@ export default class TokenomicsTransport extends BaseTransport {
       }
     };
   }
+
+  async networkStatistics () {
+    const blockHeight = (await this.api.derive.chain.bestNumber()).toNumber();
+    const numberOfMembers = (await this.api.query.members.nextMemberId() as MemberId).toNumber();
+    const content = await this.api.query.dataDirectory.knownContentIds() as Vec<ContentId>;
+    const numberOfChannels = (await this.api.query.contentWorkingGroup.nextChannelId() as ChannelId).toNumber();
```
Lezek123: Just noticed this is a similar case to forum categories and posts (`1` needs to be substracted):
```
console.log((await api.query.contentWorkingGroup.nextChannelId()))
console.log((await api.query.contentWorkingGroup.channelById.keys()).length)
```

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +350,133 @@ export default class TokenomicsTransport extends BaseTransport {
...
+
+      if (proposalType === 'Text') {
+        definedProposalType = 'text';
+      } else if (proposalType === 'Spending') {
+        definedProposalType = 'spending';
+      } else if (proposalType === 'RuntimeUpgrade' || proposalType === 'SetElectionParameters' || proposalType === 'SetValidatorCount') {
+        definedProposalType = 'networkChanges';
+      } else {
+        definedProposalType = 'workingGroups';
+      }
```
Lezek123: I thought I mentioned this before (perhaps my comment didn't get submitted), but it would probably be better to not have a "fallback" like that. We most likely will be adding new proposal types in the future, so perhaps we can have an object like `{ [k in ProposalType]: GenereicProposalType }` to make sure all existing types are always mapped to a corresponding `'text' | 'spending' | 'networkChanges' | 'workingGroups'` (see for example: `/joy-utils/src/consts/proposals.ts`)

Lezek123 (commented):
The fixes look good, added 2 more small comments and I think we can start to prepare this for Babylon!

pioneer/packages/joy-utils/src/transport/tokenomics.ts
DzhideX: Wow. Such a nice solution!!

Should we make a separate one for `HISTORICAL_PROPOSALS` though or do we leave that as is?

DzhideX:
Done!

Lezek123 (approved):
Great job! Looks ready to mege and update for Babylon!

bwhm (approved):
Looks great!
Did a few spotchecks, and everything seems correct :)

mnaamani:
Was this work ever merged into Antioch?

mnaamani:
I've retargeted the PR to sumer branch. It needs some rework specifically related to content directory since it has changed significantly in sumer. 

The work should also be rebased on sumer branch.

```
packages/joy-utils/src/transport/tokenomics.ts:308:51 - error TS2344: Type 'EntityId' does not satisfy the constraint 'UInt'.
  Type 'EntityId' is missing the following properties from type 'UInt': #private, isUnsigned, bitLength, isMax, and 85 more.

308     const entityEntries = await this.entriesByIds<EntityId, Entity>(this.api.query.contentDirectory.entityById);
                                                      ~~~~~~~~

packages/joy-utils/src/transport/tokenomics.ts:311:65 - error TS2339: Property 'class_id' does not exist on type 'Entity'.

311     const content = entityEntries.filter(([, entity]) => entity.class_id.toNumber() === 10).length;
                                                                    ~~~~~~~~

packages/joy-utils/src/transport/tokenomics.ts:312:74 - error TS2339: Property 'class_id' does not exist on type 'Entity'.

312     const numberOfChannels = entityEntries.filter(([, entity]) => entity.class_id.toNumber() === 1).length;
                                                                             ~~~~~~~~

```

mnaamani (changes requested):
Rebase on sumer and fixed needed for content directory changed in new runtime.

DzhideX:
Just finished my outstanding work for `joystream-org`. Will move on to this starting tomorrow!

PR #1501, opened 2020-10-05
Basic Storage Node Integration Tests

mnaamani:
Primary purpose of this PR was to "Dockerize" the apps in the yarn workspace of our monorepo, to make it easier to run services such as the colossus storage node in a container along with an ipfs node and joystream-node chain.
There are some issues with the size of the image but that can be resolved later.
While testing I realized some limitations of colossus with both running in a container and if/when it is started before the joystream-node is up. So I also made some adjustments to colossus:

- colossus will attempt to retry connecting to chain on startup instead of just exiting with error
- if colossus is started with provider id that is not yet registered as a worker it suspends certain operations
   like syncing and publishing identity until it becomes registered (this is particularly useful for when colossus starts up
   and dev-init hasn't completed yet)
- in a containeriszd setup we need to be able to also provide as an argument the hostname of the ipfs node, so I added that which also required a bit of refactoring of the discovery/publishing package into classes. Updated dependents (api discovery path handler, helios and the cli pacakges)
- updated the network-tests github workflow to startup a storage node and test by simply uploading a file.

Testing:
Its now possible to run a "full-stack" dev environment like so:

```
# Startup the services (ipfs node, chain, and colossus)
docker-compose --file docker-compose-with-storage.yml up -d

# configure dev storage node on chain
yarn storage-cli dev-init

# Initialize the content directory with classes, schemas and entitites
yarn workspace cd-schemas initialize:dev
```

wait about 2 minutes for colossus to register identity in ipfs and onchain..
then you can upload a file (data object id 1 and member id 0 == Alice)
`yarn storage-cli upload somefile.mp4 1 0`

Try playing the file from the assert url provided after successful upload.

```
# Finally shutdown all the nodes
docker-compose --file docker-compose-with-storage.yml down
```


apps.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN yarn install --frozen-lockfile
+
+# ENV NODE_ENV=production
+# Pioneer is failing to build only on github actions workflow runner
+# Error: packages/page-staking/src/index.tsx(24,21): error TS2307: Cannot find module './Targets' or its corresponding type declarations.
+# RUN yarn workspace pioneer build
+RUN yarn workspace @joystream/cli build
+RUN yarn workspace storage-node build
+
+# The image is huge ~ 3GB!
```
shamil-gadelshin: It looks like the reasoning rather than file comments. Do you want to keep it here?

storage-node/packages/cli/src/commands/dev.ts
```diff
@@ -100,9 +100,8 @@ const init = async (api: RuntimeApi): Promise<any> => {
     debug('Alice is already a member.')
   }
```
shamil-gadelshin, line 102: Please, remove or comment the commented code.

storage-node/packages/discovery/publish.js
```diff
@@ -1,6 +1,5 @@
-const ipfsClient = require('ipfs-http-client')
-
-const ipfs = ipfsClient('localhost', '5001', { protocol: 'http' })
+// const ipfsClient = require('ipfs-http-client')
```
shamil-gadelshin: Please, remove or comment the commented code.

storage-node/packages/discovery/publish.js
```diff
@@ -33,56 +32,62 @@ function encodeServiceInfo(info) {
   })
 }
```
shamil-gadelshin: Could you copy comments from the deleted code to annotate the class itself?

storage-node/packages/discovery/publish.js
```diff
@@ -33,56 +32,62 @@ function encodeServiceInfo(info) {
...
+      key: PUBLISH_KEY,
+      resolve: false,
+      // lifetime: // string - Time duration of the record. Default: 24h
+      // ttl:      // string - Time duration this record should be cached
+    })
 
-  // Return the key id under which the content was published. Which is used
-  // to lookup the actual ipfs content id of the published service information
-  return servicesKey.id
+    // The name and ipfs hash of the published service information file, eg.
```
shamil-gadelshin: Do we need this comment?

shamil-gadelshin (changes requested):

apps.Dockerfile
mnaamani: I take them out and put them in a new issue

mnaamani: clearned in d7cb35bea56883e0a9e5d1cc157dd4a1fa8e41b1
add to issue: https://github.com/Joystream/joystream/issues/1541

storage-node/packages/cli/src/commands/dev.ts
mnaamani: done in d607a24f8eb7591013a5aae44a506fe63268bafa

storage-node/packages/discovery/publish.js
mnaamani: done in d607a24f8eb7591013a5aae44a506fe63268bafa

storage-node/packages/discovery/publish.js
mnaamani: done in d607a24f8eb7591013a5aae44a506fe63268bafa

storage-node/packages/discovery/publish.js
mnaamani: removed it in d607a24f8eb7591013a5aae44a506fe63268bafa but left a small useful note.

shamil-gadelshin (approved):
LGTM

Issue #1394, opened 2020-09-17
Babylon: adapt Pioneer after runtime and types changes (content directory, working groups, proposals)


Lezek123:
The goal is to make Pioneer work after changes introduces in https://github.com/Joystream/joystream/issues/1391 and https://github.com/Joystream/joystream/issues/1392.

This will require:
- Removing media app (to be replaced with media stats page)
- Changing `packages/joy-roles/src/transport.substrate.ts` to use the same logic for storage and content working group (ideally: move it to `joy-utils`/`joystream-js`)
- Adjusting `joy-roles` components affected by the transport changes
- Implementing changes in `joy-proposals` (remove curator-specific proposals and allow choosing `Curators` group in generic `Wroking Groups` proposals)

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1547

Issue #1392, opened 2020-09-17
Babylon: @joystream/types - adapt for working group and porposal module changes


Lezek123:
More changes in `@joystream/types` will be required after https://github.com/Joystream/joystream/issues/1295:
- Removing `content-working-group` specific types
- Some minor changes in `working-groups` types
- Some minor changes in `proposals` types
- Possibly other (`versioned-store`/`content-directory`?)

Lezek123:
It seems possible that all the required changes have been already implemented in https://github.com/Joystream/joystream/pull/1505 and https://github.com/Joystream/joystream/pull/1521, but due to limited testing and the possibility that a few minor changes in `@joystream/types` will be required after https://github.com/Joystream/joystream/pull/1529, I'm leaving a small estimate just in case.

PR #1505, opened 2020-10-06
@joystream/types - PoroposalDetails and WorkingGroup enums update (Babylon)

Lezek123:
Minor changed in `@joystream/types` following last 3 commits in https://github.com/Joystream/joystream/pull/1385:
- Added `Content` variant in `WorkingGroup` enum
- Removed no longer supported proposals from `ProposalDetails` enum
- Ran `generate:all` using the latest node

I additionally implemented `toArray()` method in `JoyBTreeSet` (noticed that the original `BTreeSet` types doesn't have it while working on the CLI)

This partially implements https://github.com/Joystream/joystream/issues/1392
I didn't remove `types/content-working-group` yet to prevent some issues with Pioneer and auto-genrated augmentations (since this module is still exposed by the runtime).

I'll adjust Pioneer in a separate PR (see issue: https://github.com/Joystream/joystream/issues/1394)


mnaamani:
>     * Removed no longer supported proposals from `ProposalDetails` enum

Actually removing the variants from the enum in the runtime here: https://github.com/Joystream/joystream/pull/1385/files#diff-d826bd63d452cdb624a717e8355d24a3L58

Is not safe especially since this will be a runtime upgrade, because it will change values of the variants following it and this will cause issues reading state before the runtime upgrade. They should be kept and just a code comment left like for:

```
    /// ********** Deprecated during the Nicaea release.
    /// It is kept only for backward compatibility in the Pioneer. **********
    /// AccountId for the `evict storage provider` proposal
    EvictStorageProvider(AccountId),
```

Lezek123:
> Actually removing the variants from the enum in the runtime here (...) is not safe

I agree that this could cause issues by generating one of those cases where reading old chain state will become impossible unless using old types and metadata (though I believe the new version of `@polkadot/api` implements some mechanism of switching types&metadata based on which block is beeing queried, which could help in situations like this).

It would also require some tweaks for `Historical proposals` in Pioneer, since they "mock" the `ProposalDetails` type based on json input and this will not work if some enum variants no longer exist.

mnaamani:
So just a correction, with respect to the runtime, after checking the code, in its current form it is "safe" to change the variant ordering (as long as the runtime doesn't read the map that stores the ProposalDetails value). It is still an issue for pioneer.

If it is not in scope now to update pioneer, cli and historical proposals to deal with this, then we can just add back those variants. I prepared a PR here https://github.com/Joystream/joystream/pull/1510

types/src/proposals.ts
```diff
@@ -254,8 +250,6 @@ export class ProposalDetails extends JoyEnum({
   RuntimeUpgrade: Bytes,
   SetElectionParameters: ElectionParameters,
   Spending: SpendingParams,
-  SetLead: SetLead,
-  SetContentWorkingGroupMintCapacity: u128,
```
mnaamani, line 258: So I guess we will keep these and rename to with `Deprecated` prefix. As per 

https://github.com/Joystream/joystream/pull/1510/files#diff-d826bd63d452cdb624a717e8355d24a3R58

And same for other deprecated variants: `EvictStorageProvider`, `SetStorageRoleParameters`

mnaamani:
> I'll adjust Pioneer in a separate PR (see issue: #1394)

I guess that explains why the pioneer check is failing.

Otherwise changes look good.

PR #1508, opened 2020-10-06
Proposals update. Iteration 1. Delete obsolete code and add exact block execution.

shamil-gadelshin:
### Changes
- Drop some archival state - proposals and discussions after the proposal finalization and execution.
- Alter withdrawal/cancellation: only proposals with no votes could be canceled.
- Remove `create_set_election_parameters_proposal` proposal.
- Introduce block height triggering to lock in execution time.


runtime-modules/proposals/codex/src/proposal_types/mod.rs
```diff
@@ -47,6 +47,8 @@ pub enum ProposalDetails<
     /// The wasm code for the `runtime upgrade` proposal
     RuntimeUpgrade(Vec<u8>),
 
+    /// ********** Deprecated.
+    /// It is kept only for backward compatibility in the Pioneer. **********
```
bedeho, line 51: Why?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -539,6 +577,17 @@ impl<T: Trait> Module<T> {
             return Err(Error::<T>::StakeShouldBeEmpty.into());
         }
 
+        if let Some(execution_block) = exact_execution_block {
+            if execution_block == Zero::zero() {
+                return Err(Error::<T>::ZeroExactExecutionBlock.into());
+            }
+
+            let now = Self::current_block();
+            if execution_block <= now + parameters.grace_period {
```
bedeho: This does not match the handbook:

https://joystream.gitbook.io/joystream-handbook/governance/proposals#submit-proposal

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -835,125 +785,25 @@ impl<T: Trait> Module<T> {
...
-            params.stake_balance,
-            params.proposal_code,
-        )?;
+            proposer_id: params.member_id,
+            proposal_parameters: params.proposal_parameters,
+            title: params.title,
+            description: params.description,
+            stake_balance: params.stake_balance,
+            encoded_dispatchable_call_code: params.proposal_code,
+            exact_execution_block: None,
```
bedeho: Why is this `None` ?

bedeho (changes requested):
Looks like an excellent start,  I assume all the other archival state is removed later, since you said "Drop some archival state".

runtime-modules/proposals/codex/src/proposal_types/mod.rs
shamil-gadelshin: Here is a good answer by Mokhtar: https://github.com/Joystream/joystream/pull/1510

shamil-gadelshin:
Correct. I will remove 'staking errors' and other archival data during the staking implementation.

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: This is a bug. Thanks for catching, forgot to remove the intermediate stub.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Fixed - changed the check and the algorithm.

bedeho (changes requested):
I think `ensure_create_proposal_parameters_are_valid` is not checking that the trigger block is valid in the correct way according to the handboook.com

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -822,12 +822,8 @@ impl<T: Trait> Module<T> {
 
                 let now = Self::current_block();
 
-                // If exact block is set - cancel the grace period.
-                let ready_for_execution = if proposal.exact_execution_block.is_none() {
-                    proposal.is_grace_period_expired(now)
-                } else {
-                    proposal.is_execution_block_reached(now)
-                };
+                let ready_for_execution = proposal.is_grace_period_expired(now)
```
bedeho: Can we replace with `Proposal::is_ready_for_execution`.

bedeho (changes requested):

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Done

bedeho (approved):
Perfect!

PR #1510, opened 2020-10-07
Keep old enum variants for ProposalDetails and mark deprecated

mnaamani:
The `ProposalDetails` enum is used to store the proposal details when a proposal is initially created and inserted into the `ProposalDetailsByProposalId` map. The runtime never reads or iterates this map from that point on, and therefore it is safe to remove/re-order the variants of `ProposalDetails`. (The proposal also stored as an encoded `Call` which is what actually gets read later to execute)

Pioneer and the joystream/cli are the clients that actually read from `ProposalDetailsByProposalId` map. If variant ordering changes between runtimes, the clients must be aware of it to decode the value correctly. We haven't yet added support for this.. see relevant [discussion here](https://github.com/Joystream/joystream/pull/1505#issuecomment-704165931)

So this PR just adds back the variants removed in https://github.com/Joystream/joystream/pull/1385 and bumps the runtime spec version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -562,7 +562,7 @@ dependencies = [
 
 [[package]]
 name = "chain-spec-builder"
-version = "3.0.0"
+version = "3.1.0"
 dependencies = [
  "ansi_term 0.12.1",
  "enum-utils",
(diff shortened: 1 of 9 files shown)


iorveth (approved):
LGTM!

shamil-gadelshin (approved):
LGTM

bedeho (approved):
OK, this will be dropped anyway with query node and the work Shamil has already done.

PR #1515, opened 2020-10-09
Proposals update. Iteration 2. New staking implementation.

shamil-gadelshin:
This PR contains staking handler implementation for the proposals system. It implements staking via locks from the Substrate `balances` pallet.

Comments: 
- deleted joystream `stake` dependency from the proposals engine
- introduced `StakingManager` as an implementation of the `StakingHandler`
- member-to-staking-account-id is skipped for now


runtime-modules/proposals/engine/src/types/mod.rs
```diff
@@ -419,13 +379,64 @@ pub struct ProposalCreationParameters<BlockNumber, Balance, MemberId, AccountId>
...
 
+/// Contains staking balance and account id.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Stake<AccountId, Balance> {
+    /// Staking account id.
+    pub account_id: AccountId,
+
+    /// Staking balance.
+    pub balance: Balance,
```
bedeho: If staking for proposals is exact, then why do we need this?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -898,16 +845,101 @@ type ApprovedProposal<T> = ApprovedProposalData<
...
+    BalanceOf<T>,
     <T as system::Trait>::AccountId,
 >;
+
+pub struct StakingManager<T: Trait, LockId: Get<LockIdentifier>> {
+    trait_marker: PhantomData<T>,
+    lock_id_marker: PhantomData<LockId>,
+}
+
+impl<T: Trait, LockId: Get<LockIdentifier>> StakingHandler<T> for StakingManager<T, LockId> {
```
bedeho, line 868: This should not be in the engine itself, that makes the engine as a software module, tightly coupled to the balances and membership modules. It should be in the runtime, the same applies to the working group case.

bedeho, line 868: Why do we need to implement this again and again for each module, that seems wrong.

bedeho (changes requested):

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: I agree with both. It is situated here until we integrate at least 2 modules.

shamil-gadelshin: It is situated here until we integrate at least 2 modules. I don't expect copies with this handler.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: It is an artifact. We could keep it as a possible agility feature in order to make a stake 'equal or greater than' preset someday or remove it for all proposals.

runtime-modules/proposals/engine/src/lib.rs
bedeho: OK, merging in that case.

runtime-modules/proposals/engine/src/types/mod.rs
bedeho: No I don't think that is advisable, going to have to insist here.

runtime-modules/proposals/engine/src/lib.rs
bedeho: Ok.

bedeho: Ok.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: Sure.

runtime-modules/proposals/engine/src/types/mod.rs
```diff
@@ -379,15 +379,12 @@ pub struct ProposalCreationParameters<BlockNumber, Balance, MemberId, AccountId>
     pub exact_execution_block: Option<BlockNumber>,
 }
 
-/// Contains staking balance and account id.
+/// Contains staking account id.
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, Default, Clone, Copy, PartialEq, Eq, Debug)]
-pub struct Stake<AccountId, Balance> {
+pub struct Stake<AccountId> {
```
bedeho: :D Does this struct really make sense at this point! you are desperate not to go back to just a simple account? :P

bedeho (changes requested):

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: Ok. I removed it.

bedeho (approved):

PR #1509, opened 2020-10-06
CLI: Managing content directory classes, groups and permissions

Lezek123:
**Implements:**
- https://github.com/Joystream/joystream/issues/1439 - creating classes and schemas via CLI (with optional json input/output specified via `--input` and `--output` flags)
- https://github.com/Joystream/joystream/issues/1440 - manage groups and class permissions
- https://github.com/Joystream/joystream/issues/1442 - listing of classes and entities and their details
- https://github.com/Joystream/joystream/issues/1397 - `-g curators` support for `working-groups` commands

_Not all issues are implmented in full scope yet, there are still a few minor things missing, like, for example, `contentDirectory:updateClassSchemaStatus` command_

_**NOTE: All `working-groups` commands should now work with `-g curators`, ie.: `./cli/bin/run working-groups:createOpening -g curators`. so it should be possible to handle curator hiring process in a similar way to how storage providers can be hired** (except Pioneer is not yet updated, so `Extrinsics` tab must be used when applying)_

I updated `README.md` using the `oclif-dev readme` script to include the new commands. This documentation can be found here: https://github.com/Lezek123/substrate-runtime-joystream/tree/cli-content-directory-classes/cli

As always commands can also be inspected by running:
```
./cli/bin/run content-directory
./cli/bin/run content-directory:createClass --help
./cli/bin/run content-directory:entities --help
```
etc.

The draft PR is based on https://github.com/Joystream/joystream/pull/1463, so in order to initialize content directory on dev chain it's possible to run:
```
yarn workspace cd-schemas initialize:dev
```
_This will set ALICE as lead and initialize content directory with classes, schemas and entities based on inputs in `content-directory-schemas/inputs`_

Issue #1443, opened 2020-09-22
CLI: Creating and updating channels, uploading and updating videos


Lezek123:
Since some entity-related actions are more complex and will be hard or impossible to cover with commands introduced in https://github.com/Joystream/joystream/issues/1441, there should be a few higher-level commands that will reuse some of thier logic, but provide additional functionality.

There should be already a lot of lower-level commands in the `contentDirectory:` family at this point, perhaps it would be best to create a new family for higher-level commands intended mostly for standard members trying to manage their content (ie. `media:`)

Commands to cover:
- `media:uploadVideo [path]` - handle the process of uploading a video, which will require
    - interacting with the storage node
    - probably creating multiple entities (`Video`, `MediaLocation`, `VideoMedia`), where some input will be provided by the user and some will be derived from the provided file or the storage node interaction
- `media:createChannel` - this can be just a shortcut for `contentDirectory:createEntity` with some pre-defined input as described in https://github.com/Joystream/joystream/issues/1441
- `media:updateChannel` - a shortcut for `contentDirectory:updateEntity`
- `media:updateVideo` - probably can also be just some sort of a shortcut for `contentDirectory:updateEntity`

Optionally can add some other low-effort shortcut-commands here, like `media:videos` as shortcut for `contentDirectory:entities Video` etc.

Lezek123:
I'm updating the estimate due to a few factors:
- I started working on this issue earlier than expected, which means there wasn't all that much functionality related to creating and updating entities to rely on yet
- The initial estimate turned out to be too optimistic, since the process of uploading a video involves quite a few interactions with the Joystream infrstructure and the operations on the content directory itself also involve working with relatively complex relations between multiple entities.

I think it's also worth to spend some more time on this task in order to make the experience of uploading a video as user-friendly as possible, otherwise it can easily become too complicated and prone to many errors.

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1516

Issue #1441, opened 2020-09-22
CLI: Managing entities (creating, updating, removing etc.)


Lezek123:
As I mentioned in https://github.com/Joystream/joystream/issues/1398, I think it's worth to implement commands for handling more universal, lower-level interactions like entity creation and updates in the CLI, instead of trying to cover only the creation of some specific entities like `Video` or `Channel` and specific updates like marking content as curated. It doesn't mean that we can't do both, since if we start from more universal commands, it should be trivial to then create more specific ones that can reuse existing logic (ie. `contentDirectory:createChannel` could be a shortcut for `contentDirectory:createEntity` with some specific `classId`, `actor` and `supportedSchemas` as input)

### `actor` context
Entity-related commands will require providing value for `actor` argument of the underlying extrinsics. This means that when executed, the CLI should probably display a list of possible `actor` contexts that the user can choose from based on the currently selected key (the way it's already the case with worker context in some of the `working-groups` commands).

### Commands covered by this issue:
* `contentDirectory:createEntity [ClassID / ClassName] [SchemaId]` -  will combine `createEntity` and `addSchemaSupportToEntity` extrinsics. It can also allow the user to choose a class and schema based on exsisting ones queried from the chain. Then it should prompt for and validate property values based on data fetched from the chain using provided `ClassID` and `SchemaID`. Ideally when prompting for relationship value (ie. `Language`) it would display a list of entities that the user can select from. It should also integrate an "extended draft system" described in https://github.com/Joystream/joystream/issues/1439 (allowing it to take `json` files as input and also save input provided via UI into `json` files) 
* `contentDirectory:updateEntity [EntityID]` - will prompt for and validate property values based on data fetched from the chain. It should prompt only for values that can be provided considering choosen `actor` context. It can use current entity values as defaults.
* `contentDirectory:removeEntity [EntityID]`

### Not covered extrinsics (unless there is a strong use-case for current release):
* `clearEntityPropertyVector`
* `insertAtEntityPropertyVector`
* `removeAtEntityPropertyVector`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703465796846) by [Unito](https://www.unito.io)

Lezek123:
Moving out of `Babylon` scope, since most of the common use-cases were covered by higher-level commands:
- `media:createVideo`
- `media:updateVideo`
- `media:removeVideo`
- `media:createChannel`
- `media:updateChannel`
- `media:removeChannel`

[...]

PR #1516, opened 2020-10-09
CLI: uploading videos, creating channels, updating videos and channels

Lezek123:
Implements: https://github.com/Joystream/joystream/issues/1443
Based on: https://github.com/Joystream/joystream/pull/1509

Commands added:
- `media:createChannel`
- `media:updateChannel`
- `media:uploadVideo PATH`
- `media:updateVideo`
- `media:myVideos`
- `media:myChannels`

Commands modified:
- `content-directory:class` - added `Index` (`PropertyID`) column
- `content-directory:entities` - added `--filters` flag allowing for basic filtering of the results. 

_See updated `README` at: https://github.com/Lezek123/substrate-runtime-joystream/blob/cli-media/cli/README.md for further description of the new commands. Note that most of the arguments/flags are optional, in which case entities can be selected via a CLI choice-list prompt_

#### Uploading a video - features
- `ffmpeg.ffprobe` is used to retrieve some information (metadata) about the video, like: codec name, width, height and duration, and provide them as defaults / hints when prompting for `Video` / `VideoMedia` entity values. https://www.npmjs.com/package/@ffmpeg-installer/ffmpeg is used here for cross-platform support. Since it seems like this approach may be prone to some errors and is not essential for the command to do its job, the failure to retrieve this data will only result in a warning (the command itself will not be terminated).
- To resolve the upload url - either the local ipfs node will be used (if available) or random bootstrap discovery endpoint (fetched via `api.query.discovery.bootstrapEndpoints`). The information about which one was used will be logged in the console.
- There is some basic error handling for cases like too large file or no storage provider available. Other errors (like invalid mime type) are handled during the upload (caught from `axios.put` when sending a request to upload endpoint)
- Specifying custom user license is not yet implemented (currently only one of the `KnownLicense`s can be chosen)


cli/src/base/AccountsCommandBase.ts
```diff
@@ -224,6 +224,16 @@ export default abstract class AccountsCommandBase extends ApiCommandBase {
...
   }
 
+  async getRequiredMemberId(): Promise<number> {
+    const account = await this.getRequiredSelectedAccount()
+    const memberIds = await this.getApi().getMemberIdsByControllerAccount(account.address)
+    if (!memberIds.length) {
+      this.error('Membership required to access this command!', { exit: ExitCodes.AccessDenied })
+    }
+
+    return memberIds[0].toNumber() // FIXME: Temporary solution (just using the first one)
```
mnaamani, line 251: I wanted to raise this issue in previous PR about member id selection and clearly you are aware of it hence your "FIXME" comment. Just wanted to assert that yes it would be good to give use option to select the member id of their choice when there are multiple matches..

cli/src/base/ApiCommandBase.ts
```diff
@@ -405,11 +406,26 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
...
+            .filter(({ event }) => event.section === 'system')
+            .forEach(({ event }) => {
+              if (event.method === 'ExtrinsicFailed') {
+                const dispatchError = event.data[0] as DispatchError
+                let errorMsg = dispatchError.toString()
+                if (dispatchError.isModule) {
+                  try {
+                    // Need to assert that registry is of TypeRegistry type, since Registry intefrace
+                    // seems outdated and doesn't include DispatchErrorModule as possible argument for "findMetaError"
+                    const { name, documentation } = (this.getOriginalApi().registry as TypeRegistry).findMetaError(
```
mnaamani, line 428: wonderful, I will try to update storage-node with same code, there i just pass the raw dispatch error which is quite useless :)

mnaamani:
Wroked great. Uploaded a video successfully and played it back over asset url.
And nice fixes from last PR too.

Just couple observations:

- When prompted to enter video metadata after upload:
```
File uploaded!
 Choose Video encoding MPEG4
? pixelWidth 
? pixelHeight 
 Choose Video language English
 Choose Video category Cartoon
 Choose License CC_BY
? title my first video
? description zoom 4
? thumbnailURL 
? duration dunno.  
? isPublic Yes
? isExplicit No
? hasMarketing No
```

for `duration` I entered text without being warned that it should be a number.

- Inspecting `Video` entity shows a strange values for `curation status`, `skippableIntroDuration`, `publishedBeforeJoystream`
```
MokoBook:joystream mokhtar$ yarn joystream-cli content-directory:entity 29
yarn run v1.22.4
$ /Users/mokhtar/joystream/joystream/node_modules/.bin/joystream-cli content-directory:entity 29
ID                            29        
Class name                    Video     
Class ID                      11        
Supported schemas             [0]       
Controller                    Member(0) 
Frozen                        false     
Refrecencable                 true      
Same owner references         0         
Total references              0         

________________ Property values _________________

channel                       24 Single<Reference>        
category                      2 Single<Reference>         
title                         my first video Single<Text> 
description                   zoom 4 Single<Text>         
duration                      0 Single<Uint32>            
skippableIntroDuration        false Single<Uint16>   // expected number
thumbnailURL                   Single<Text>               
language                      10 Single<Reference>        
media                         28 Single<Reference>        
hasMarketing                  false Single<Bool>          
publishedBeforeJoystream      false Single<Uint32>       // uhm ?
isPublic                      true Single<Bool>           
isExplicit                    false Single<Bool>          
license                       25 Single<Reference>        
curationStatus                false Single<Reference>.   // expected entity id but showing 
```

- Prompt for account password
I'm always prompted for password for account which doesn't have one. perhaps function that loads the key can try to unlock the kerying without a password first or check if it is locked, only if it fails to unlock then prompt for a password? Less hassle for user.

I'll try to do a bit more testing like with no ipfs node, storage node being down...

Lezek123:
Added a few fixes:
- > for duration I entered text without being warned that it should be a number.
   
  Fixed in `JSONSchemaPrompt` by constructing `Ajv` with `allErrors: true` (which is required when we prompt for values in non- 
  standard order,  ie. by using `promptMultipleProps`) and changing normalizers for `integer` and `number` to only return the normalized value if it matches the original input when converted `.toString` 
- > Inspecting Video entity shows a strange values for curation status, skippableIntroDuration, publishedBeforeJoystream
  
  When the value for a property is not set (which can be the case when the property is not `required`), the runtime seems to return `Bool(false)`. I added an additional check to handle this case in commands like `entity` and `entities` and display a more clear message (`[not set]`)
- > I'm always prompted for password for account which doesn't have one

Fixed + added a loop when prompting for password (when it is actually set to non-empty string), so that providing invalid one doesn't exit the command right away (which I found to also be quite bad UX)

- > I'll try to do a bit more testing like with no ipfs node, storage node being down...

I noticed a small error in `isAnyProviderAvailable` check which I probably introduced accidently at some point, but it should be fixed now.

cli/src/Api.ts
```diff
@@ -481,16 +490,18 @@ export default class Api {
   }
 
   // Content directory
-  availableClasses(): Promise<[ClassId, Class][]> {
-    return this.entriesByIds<ClassId, Class>(this._api.query.contentDirectory.classById)
+  async availableClasses(useCache = true): Promise<[ClassId, Class][]> {
+    return useCache && this._cdClassesCache
+      ? this._cdClassesCache
+      : await this.entriesByIds<ClassId, Class>(this._api.query.contentDirectory.classById)
```
mnaamani: shouldn't we set `_cdClassesCache` here and then return it after querying it from chain. I don't see any place where it is ever set.

cli/src/base/AccountsCommandBase.ts
```diff
@@ -216,12 +216,34 @@ export default abstract class AccountsCommandBase extends ApiCommandBase {
   }
 
   async requestAccountDecoding(account: NamedKeyringPair): Promise<void> {
-    const password: string = await this.promptForPassword()
+    // First - try decoding using empty string
```
mnaamani, line 224: We could also return early if `!account.isLocked`

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -163,4 +244,33 @@ export default abstract class ContentDirectoryCommandBase extends AccountsComman
       return columns
     }, {} as Record<string, { value: Codec; type: string }>)
   }
+
+  async parseToKnownEntityJson<T>(entity: Entity, entityClass?: Class): Promise<FlattenRelations<T>> {
```
mnaamani: It seems safer to never take the optional argument `entityClass` since it can be fetched by class_id from Entity.
`parseEntityPropertyValues()` will not give proper result if wrong entityClass is passed to it, and that could be hard to debug?

cli/src/helpers/InputOutput.ts
```diff
@@ -4,9 +4,13 @@ import ExitCodes from '../ExitCodes'
 import fs from 'fs'
 import path from 'path'
 import Ajv from 'ajv'
-import { JSONSchema7 } from 'json-schema'
+import $RefParser, { JSONSchema } from '@apidevtools/json-schema-ref-parser'
+import { getSchemasLocation } from 'cd-schemas'
 import chalk from 'chalk'
 
+// Default schema path for resolving refs
+const DEFAULT_SCHEMA_PATH = getSchemasLocation('entities') + path.sep
```
mnaamani, line 12: Here we are assuming that the schemas we have at this location match the schemas on chain.
That may not always a valid expectation and the cli should probably not rely on them, but instead read them/reconstruct them from the chain.

mnaamani (changes requested):

cli/src/helpers/InputOutput.ts
mnaamani: Also on this point, if cli package depends on the cd-schemas package, we would have to make sure to publish that to npm as well. Is it something we want to do?

mnaamani:
> Added a few fixes:
> 
>     * > for duration I entered text without being warned that it should be a number.
>       
>       
>       Fixed in `JSONSchemaPrompt` by constructing `Ajv` with `allErrors: true` (which is required when we prompt for values in non-
>       standard order,  ie. by using `promptMultipleProps`) and changing normalizers for `integer` and `number` to only return the normalized value if it matches the original input when converted `.toString`
> 
>     * > Inspecting Video entity shows a strange values for curation status, skippableIntroDuration, publishedBeforeJoystream
>       
>       
>       When the value for a property is not set (which can be the case when the property is not `required`), the runtime seems to return `Bool(false)`. I added an additional check to handle this case in commands like `entity` and `entities` and display a more clear message (`[not set]`)
> 
>     * > I'm always prompted for password for account which doesn't have one
> 
> 
> Fixed + added a loop when prompting for password (when it is actually set to non-empty string), so that providing invalid one doesn't exit the command right away (which I found to also be quite bad UX)
> 
>     * > I'll try to do a bit more testing like with no ipfs node, storage node being down...
> 
> 
> I noticed a small error in `isAnyProviderAvailable` check which I probably introduced accidently at some point, but it should be fixed now.

Tested :+1:

cli/src/base/AccountsCommandBase.ts
Lezek123: Good idea, added the check.

cli/src/base/ContentDirectoryCommandBase.ts
Lezek123: That's true, I actually forgot to remove this, the purpose of the optional argument was to avoid an additional api call, but after I implmeneted `_cdClassesCache` it is no longer necessary.

cli/src/Api.ts
Lezek123: Right, I overlooked that, thank you for noticing!

cli/src/helpers/InputOutput.ts
Lezek123: I'm not sure if this still applies after https://github.com/Joystream/joystream/pull/1530, but:

1. I think it would be very useful to have `cd-schemas` published either way. Perhaps it can become part of `@joystream/js` if I manage to get into that before the release. If not, it can be just a standalone library that makes interactions with `content-directory` runtime module easier.

1. > Here we are assuming that the schemas we have at this location match the schemas on chain.

Since this is used by higher-level commands like `uploadVideo` etc. that make some assumptions about the `Video` schema anyway (when populating `VideoEntity` interface with data provided by the user) I think it may be acceptable? 

The alternative would be to avoid any assumptions about classes/schemas at all in the CLI and just have a command like `createEntity Video 0` (where `0` would be `schemaIndex`), This was the idea behind https://github.com/Joystream/joystream/issues/1441 which I postponed a bit due to change of priorities, but I hope we'll have both of those solutions implmented soon and we could then compare their pros and cons (for now I think having a nice, customizable UI for uploading a video is a good thing, because CLI will temporarly be the only place where this could be done)

Lezek123:
Added missing `removeCuratorFromGroup` command

Lezek123:
Added a small api initialization speed improvement by implementing metadata cache, I was expecting commands to execute ~2-3 seconds faster after that, but turned out it only saves ~300-800 ms on local chain (which is still something).

I suspect the difference may be greater when fetching data from a remote node.

mnaamani:
> Added a small api initialization speed improvement by implementing metadata cache, I was expecting commands to execute ~2-3 seconds faster after that, but turned out it only saves ~300-800 ms on local chain (which is still something).
> 
> I suspect the difference may be greater when fetching data from a remote node.

Nice addition. Yes might be more noticeable when using a remote endpoint.

PR #1523, opened 2020-10-12
pioneer: add missing Content key in apiModuleByGroup

mnaamani:


Lezek123 (approved):
This is just enough to make the CI checks pass and the transport inside `joy-utils` work with the new content working group.

Some adjustments in `joy-roles/src/transport.substrate.ts` and a few react components are still needed in order to make Pioneer fully compatible with the new runtime, but I'll introduce those in another PR.

PR #1527, opened 2020-10-13
Make Navbar Sticky

### LINKED ISSUE #1524: Stable navigation in Atlas
When you scroll in ATlas, at least according to the designs, then all major navigation elements go away, like search, the left hand side menu (which we currently have removed anyway). This is a really bad interaction in my opinion, it means you always ahve to navigate all the way back up again to transition screens! I am not sure how I did not react to this before.

This is a non-starter I think, we need to change it,

fcescob:
This PR fixes #1524

### DIFF
--- a/atlas/src/components/LayoutWithRouting.tsx
+++ b/atlas/src/components/LayoutWithRouting.tsx
@@ -13,36 +13,45 @@ type RouteProps = {
 } & RouteComponentProps
 const Route: React.FC<RouteProps> = ({ Component, ...pathProps }) => {
   return (
-    <MainContainer>
-      <ErrorBoundary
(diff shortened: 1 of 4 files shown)


kdembler:
This doesn't work well on the Browse screen - the navbar gets shadowed by the category picker.

Also, I'm a little worried about the fact that we'll have to manually specify the navbar height in the styling. If we change the navbar and forget to update `NAVBAR_HEIGHT`, screens may break.

What if instead of having the navbar fixed, we would have it be a part of a layout and have the main container be scrollable? I'm not sure what drawbacks this creates

kdembler (approved):
LGTM

PR #1530, opened 2020-10-13
Content directory tooling as library

Lezek123:
This PR includes:
- a few more entities in `entityBatches` json inputs (now there are ~40 languages, ~20 encodings and ~15 categories)
- updates in `InputParser` allowing to reference entities with `{ existing: { /* unique prop query */ } }` even if they are not part of the same batch, but already exist on chain. This currently only works when creating new entities (`getEntityBatchOperations`)
- added documentation in the `README` regarding the usage of `cd-schemas` workspace as library and a few examples in `content-directory-schemas/examples` that can be ran with `yarn workspace cd-schemas example:[EXAMPLE_NAME]` (further described in the `README`)

The new `README` can be found here: [LINK](https://github.com/Lezek123/substrate-runtime-joystream/tree/cd-schemas-lib/content-directory-schemas) (or [Direct link to the new section](https://github.com/Lezek123/substrate-runtime-joystream/tree/cd-schemas-lib/content-directory-schemas#using-as-library))


mnaamani:
This is extremely valuable, and with reference to my comment [here](https://github.com/Joystream/joystream/pull/1516#discussion_r505812118) it would be great to have the library actually load the schemas from chain rather than use the "harcoded" ones. Because it will be impractical to update the library when new classes and schemas are updated on-chain. Please correct me if I'm missing some details here.

Lezek123:
>  it would be great to have the library actually load the schemas from chain rather than use the "harcoded" ones. Because it will be impractical to update the library when new classes and schemas are updated on-chain.

This is the discussion we had with @bedeho, but in the end I'm not sure if it makes all that much of a difference whether we load schema data from files or from chain. There are pros and cons on both sides and we need to update the library anyway (which for the most part will be just about running a script) if there are some new schemas/classes added in the runtime, since the higher-level interfaces (like `VideoEntity` or `ChannelEntity`) and json schemas that provide the validation for entity inputs will become outdated (we should still be able to succesfully use the library to create "older-schema" `Video`, because the parser assumes we always want to add support for schema `0` to the entity that we create). In that case we would also most likely have to update other projects (like query node, Atals etc.) that expect a certain schema for those entities.

Having said that, we can generate json schemas and TypeSciprt interfaces for entities based directly on what we have on chain instead on relying on the up-to-date schema input files to be present inside `inputs/schemas` (which was my initial idea). This theoretically saves us one step (adding those files), but in reality this shouldn't be much of a problem, because those files were either created in the first place in order to put the new schemas on chain or the CLI was used to do that (in which case we just need to make sure to use `--output` and we'll have those files available too). The benefit of having up-to-date schema input files as part of the library is that we can then always be sure that running `initialize:dev` script will produce exactly the classes and schemas that we expect. Now if we assume that schemas in the files match those in the runtime, loading them from files could theoretically have the benefit of using some of the library functionality without even connecting to the chain, but this is not really the case ATM, since we almost always need to do that in order to get `className => classId` map

So this is obviously not the final solution, because the library is still a work in progress, ultimately I think we'd probably want:
- Full export/import logic for runtime classes, schemas and possibly entities
- Some logic of defining what's the currently supported schema for each class, or if multiple ones can be supported at the same time - some logic related to handling that
- Perhaps some mechanism of detecting inconsistencies between the schema in the runtime and the schema that we expect

It's possible that I'm missing something here and there exists a much more universal way to approach this, but for the purpose of Babylon where we make quite strict assumptions about what classes, schemas and perhaps even some entities (like. Language) will be available, I think it may be enough to have it done this way.

mnaamani:
Thanks for the deep analysis. I'm glad you have already discussed this and I agree for babylon what we have is already a great starting point and advanced enough for what we need.

On the point:
> This theoretically saves us one step (adding those files), but in reality this shouldn't be much of a problem, because those files were either created in the first place in order to put the new schemas on chain or the CLI was used to do that (in which case we just need to make sure to use --output and we'll have those files available too.

Yes and it will most likely be the responsibility of the curator team to ensure the library is kept upto date, so when they add new classes and and schemas they need to make sure to add the new outputed files with a pull-request.

mnaamani:
Additional note since the cd-schemas package will need to be published to npm we probably want to add some CI checks similar to the cli to re: npm pack etc.

mnaamani:
Tested working great, just needs a merge conflict fix after merging https://github.com/Joystream/joystream/pull/1516

mnaamani (changes requested):
Update with merge conflict fixes and its good to go.

PR #1532, opened 2020-10-13
createType - TypeScript support

Lezek123:
Adds typescript support for `@joystream/types`' `createType` function.
We can now use it to more safely create very complex types, for example:
```
    createType('OperationType', {
      AddSchemaSupportToEntity: {
        entity_id: { ExistingEntity: 1 },
        parametrized_property_values: [
          {
            in_class_index: 1,
            value: {
              InputPropertyValue: {
                Single: { Reference: 2 },
              },
            },
          },
        ],
        schema_id: 0,
      },
    })
```
Assuming we're using agumentations from `/agument-codes/augment-types.ts` (which almost all TypeScript projects now use for `createType`), the object provided as second argument to `createType` in this example will be fully validated by TypeScript (even if we provide invalid value for `Reference`, which is very deeply nested, we will still get an error)

### DIFF
--- a/content-directory-schemas/src/helpers/InputParser.ts
+++ b/content-directory-schemas/src/helpers/InputParser.ts
@@ -143,16 +143,13 @@ export class InputParser {
         let value = customHandler && customHandler(schemaProperty, propertyValue)
         if (value === undefined) {
           value = createType('ParametrizedPropertyValue', {
-            InputPropertyValue: this.parsePropertyType(schemaProperty.property_type)
-              .toInputPropertyValue(propertyValue)
-              .toJSON(),
+            InputPropertyValue: this.parsePropertyType(schemaProperty.property_type).toInputPropertyValue(
+              propertyValue
+            ),
           })
         }
 
-        return {
-          in_class_index: schemaPropertyIndex,
-          value: value.toJSON(),
-        }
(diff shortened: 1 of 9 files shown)


mnaamani:
This enhancement for our `createType` is brilliant. But honestly the advanced syntax is beyond me. I'm happy to merge it as is, but might be prudent for someone else on the team with more typescript experience to have a closer look?

mnaamani (approved):

Lezek123:
One small issue I noticed is that it takes some time for the compiler to validate the second argument of `createType`, so the errors in IDE come up after a short delay. This isn't all that bad and the overall build time doesn't seem to increase noticeably, so it may be a non-issue, but leaving a note here that this may be perhaps something worth looking into.

mnaamani:
> One small issue I noticed is that it takes some time for the compiler to validate the second argument of `createType`, so the errors in IDE come up after a short delay. This isn't all that bad and the overall build time doesn't seem to increase noticeably, so it may be a non-issue, but leaving a note here that this may be perhaps something worth looking into.

The pros of having it far outways this short delay. I've noticed VSCode does get a little slow in general when there are multiple extensions doing code validation/linting etc so its okay.

PR #1533, opened 2020-10-14
testing init of new content directory in github workflow

mnaamani:


mnaamani:
[Failing network integration tests](https://github.com/Joystream/joystream/pull/1533/checks?check_run_id=1252168741) fail here is expected because the runtime has been updated and the tests in this branch still expect some proposals types.
Those are fixed in https://github.com/Joystream/joystream/pull/1506

Lezek123 (approved):
Very useful check before we implement the full query node testing scenario, LGTM!

Issue #1200, opened 2020-08-31
Proposal System: Final Major iteration


bedeho:
# Overview

We should refactor the proposal system, meaning the engine, the codex and discussion modules, to be in line with the functional description currently live in the Handbook: https://joystream.gitbook.io/joystream-handbook/governance/proposals

It's important to bear in mind that this is an integrated functional description that crosses many layers of abstraction, not only the multiple modules but even into what would be found in a query node. So it must be extracted from this how the modules should be organized to facilitate this. Some of the key points that pertain specifically to how to reorganize the runtime modules are therefore

- Drop all archival stuff, both in codex, engine, and discussion system.
- Use locks for staking, not `staking` module, recall
   - one account can be used per proposal
   - verify with runtime that account is bound to member
   - has sufficient free balance
   - verify with runtime that account is not already locked for purpose for which stake is not reusable: anything except voting
- Drop withdrawal/cancellation.
- Introduce block height triggering to lock in execution time.
- Discussion permissions: https://github.com/Joystream/joystream/issues/1227
- Add constitutionality: https://github.com/Joystream/joystream/issues/144
- Clearly separate creation and execution checks: https://github.com/Joystream/joystream/issues/1205, this requires going through each proposal creator one-by-one.
- Add rationales: https://github.com/Joystream/joystream/issues/1157
- Bug: https://github.com/Joystream/joystream/issues/828
- Update budget: https://github.com/Joystream/joystream/issues/1487
- Update new proposals
   - Add constitution: https://github.com/Joystream/joystream/issues/799
   - Add council blog
   - Comment out `create_set_election_parameters_proposal` proposal.
   - Update all proposals to 
- Proposal parameters are constants https://github.com/Joystream/joystream/issues/862

# Update Handbook

The current state of the handbook is quite complete, but the details of each proposal is not filled in, and as many proposals will change with this change, filling in the details has been postponed.

**After completing this work, please update the signature and information for each active proposal, and update any incorrect claims about business logic and state management in general. There is no need to fill in all the missing constant values, this can be done by someone else**

# Implementation Path

## Starting Point

Should build on a new council & election system & working group system that should be ready to go into production. Only work on the codex directly needs these most recent versions, hence work on the engine and discussion modules can still commence if these dependencies are still not fully completed.

## Process

This is a complex set of changes, here are some ideas for how I think we can minimize confusion and false starts:

1. It should be bottom-up in terms of dependency graph: 
  a. Engine module
  b. Discussion module
[...]

Issue #862, opened 2020-07-02
Remove storage representation of immutable parameters in codex


bedeho:
These should not be in storage, as they should not be genuinely mutable, but instead be module parameters which futher are configurable in the runtime build for integration testing purposes. Implementing this will have cascading implications for the runtime and integration testing tool.

```Rust
        /// Voting period for the 'set validator count' proposal
        pub SetValidatorCountProposalVotingPeriod get(set_validator_count_proposal_voting_period)
            config(): T::BlockNumber;

        /// Grace period for the 'set validator count' proposal
        pub SetValidatorCountProposalGracePeriod get(set_validator_count_proposal_grace_period)
            config(): T::BlockNumber;

        /// Voting period for the 'runtime upgrade' proposal
        pub RuntimeUpgradeProposalVotingPeriod get(runtime_upgrade_proposal_voting_period)
            config(): T::BlockNumber;

        /// Grace period for the 'runtime upgrade' proposal
        pub RuntimeUpgradeProposalGracePeriod get(runtime_upgrade_proposal_grace_period)
            config(): T::BlockNumber;

        /// Voting period for the 'set election parameters' proposal
        pub SetElectionParametersProposalVotingPeriod get(set_election_parameters_proposal_voting_period)
            config(): T::BlockNumber;

        /// Grace period for the 'set election parameters' proposal
        pub SetElectionParametersProposalGracePeriod get(set_election_parameters_proposal_grace_period)
            config(): T::BlockNumber;

        /// Voting period for the 'text' proposal
        pub TextProposalVotingPeriod get(text_proposal_voting_period) config(): T::BlockNumber;

        /// Grace period for the 'text' proposal
        pub TextProposalGracePeriod get(text_proposal_grace_period) config(): T::BlockNumber;

        /// Voting period for the 'set content working group mint capacity' proposal
        pub SetContentWorkingGroupMintCapacityProposalVotingPeriod get(set_content_working_group_mint_capacity_proposal_voting_period)
            config(): T::BlockNumber;

        /// Grace period for the 'set content working group mint capacity' proposal
        pub SetContentWorkingGroupMintCapacityProposalGracePeriod get(set_content_working_group_mint_capacity_proposal_grace_period)
            config(): T::BlockNumber;

        /// Voting period for the 'set lead' proposal
        pub SetLeadProposalVotingPeriod get(set_lead_proposal_voting_period)
            config(): T::BlockNumber;

        /// Grace period for the 'set lead' proposal
[...]

PR #1535, opened 2020-10-14
Proposal new features and new constitution pallet.

shamil-gadelshin:
Comments:
- new `constitution` pallet was added with storage for text hash and extrinsic for changes
- new `amend constitution` proposal was added
- a migration for the amend constitution proposal was added
- vote rationale was added to the `vote` extrinsic
- `constitutionality` feature was added to proposals `engine`
- `ActiveStake` archival state was removed
- multiple tests were added


runtime-modules/constitution/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// Contains constitution text hash and its amendment number.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, Default)]
+pub struct ConstitutionInfo {
+    /// Constitution text hash.
+    pub text_hash: Vec<u8>,
+
+    /// Constitution amendment number.
+    pub amendment_number: u32,
```
bedeho: This is never read, except to set the new number, so it has no non-archival function.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -303,6 +303,14 @@ decl_storage! {
         /// Grace period for the 'terminate working group leader role' proposal
         pub TerminateWorkingGroupLeaderRoleProposalGracePeriod get(fn terminate_working_group_leader_role_proposal_grace_period)
             config(): T::BlockNumber;
+
+        /// Voting period for the 'amend constitution' proposal
+        pub AmendConstitutionProposalVotingPeriod get(fn amend_constitution_proposal_voting_period)
```
bedeho, line 308: I think we want to not have these in storage, ala https://github.com/Joystream/joystream/issues/862.

As you can see from the handbook, all these parameters are constants, not mutable.

bedeho: I am seeing the orignal issue #1200 did not reference it, this was my mistake.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -318,6 +326,15 @@ decl_module! {
         /// Exports max wasm code length of the runtime upgrade proposal const.
         const RuntimeUpgradeWasmProposalMaxLength: u32 = T::RuntimeUpgradeWasmProposalMaxLength::get();
 
+        // Runtime upgrade summary:
+        //  - add values for the new 'amend constitution' proposal.
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
bedeho, line 331: Drop when storage values are dropped.

runtime-modules/proposals/engine/src/types/mod.rs
```diff
@@ -133,17 +129,24 @@ pub struct Proposal<BlockNumber, ProposerId, Balance, AccountId> {
     /// Proposal body
     pub description: Vec<u8>,
 
-    /// When it was created.
-    pub created_at: BlockNumber,
+    /// When it was created or reactivated to get constitutionality.
```
bedeho: My understanding is that the value here is either
a) when the proposal was first created.
b) the beginning of a new council where the proposal was automatically activated due to constitutionality

If so, can we write this out explicitly in the comment?

runtime-modules/constitution/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Memo {
+        Constitution get(fn constitution) : ConstitutionInfo;
+        NextAmendmentNumber get(fn next_amendment_number) : u32 = 0;
```
bedeho: Drop for same reason.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -468,30 +474,25 @@ impl<T: Trait> Module<T> {
...
-                if let Some(staking_account_id) = creation_params.staking_account_id {
-                    T::StakingHandler::lock(&staking_account_id, stake_balance);
-
-                    Some(ActiveStake {
-                        source_account_id: staking_account_id,
-                    })
-                } else {
-                    return Err(Error::<T>::EmptyStake.into());
-                }
+        if let Some(stake_balance) = creation_params.proposal_parameters.required_stake {
```
bedeho, line 478: Can you add a comment on what is being done here, or even isolate out as ensure check? I know this may not be reused multiple times, but it may help readability.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -318,7 +314,11 @@ decl_storage! {
             T::ProposalId=> ();
 
         /// Ids of proposals that were approved and theirs grace period was not expired.
-        pub PendingExecutionProposalIds get(fn pending_proposal_ids): map hasher(blake2_128_concat)
+        pub PendingExecutionProposalIds get(fn pending_execution_proposal_ids): map hasher(blake2_128_concat)
+            T::ProposalId=> ();
+
+        /// Ids of proposals that were approved and needed more councils approvals to be executed.
+        pub PendingConstitutionalityProposalIds get(fn pending_consitutionality_proposal_ids): map hasher(blake2_128_concat)
```
bedeho, line 321: I understand this new state serve the purpose of avoiding iterating over all proposals when reactivating them at the beginning of a new council? If so, does this very minor performance improvement warrant new state. This iteration, over a bounded set, happens very rarely, it does not seem to warrant adding new state that must be correctly managed.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -589,13 +590,35 @@ impl<T: Trait> Module<T> {
 
     /// Resets voting results for active proposals.
     /// Possible application includes new council elections.
-    pub fn reset_active_proposals() {
+    pub fn reset_votes_for_active_proposals() {
```
bedeho, line 595: Why are `reset_votes_for_active_proposals` and `reactivate_pending_constitutionality_proposals` two distinct API calls? Wouldn't you just want a single call which you invoke whenever you want dormant proposals to be reactivated?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -589,13 +590,35 @@ impl<T: Trait> Module<T> {
...
+        <Proposals<T>>::mutate(proposal_id, |proposal| {
+            proposal.reset_proposal_votes();
+            <VoteExistsByProposalByVoter<T>>::remove_prefix(&proposal_id);
+        });
+    }
+
+    /// Reactivate proposals with pending constitutionality.
+    /// Possible application includes new council elections.
+    pub fn reactivate_pending_constitutionality_proposals() {
+        <PendingConstitutionalityProposalIds<T>>::iter().for_each(|(proposal_id, _)| {
```
bedeho, line 607: Can this be done more functionally, just computing new proposals and reinsert them into `Proposals`.

bedeho (changes requested):
Two overarching issues here

1. I think this PR combined too many unrelated changes, such as a particular proposal and constitutionality. I hope the naming did not suggest that they were some how related? That was not the intention.

2. It seems before we proceed to adding new proposal types, or even adding new features to the engine (rationale, constitutionality), it would be advisable to clean it up and drop the state representations which no longer apply.

3. I think we need to limit the total number of proposals, not just active proposals. Currently, we are excluding dormant proposals from the constraint, which is not safe, or clearly and simple.

runtime-modules/constitution/src/lib.rs
shamil-gadelshin: We don't store the history of the amendment numbers. It is always the last one.  This is actual data rather than archival data. Without it the 'query node' has more information than the runtime, which is incorrect.

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: I coded it similar to existing proposals. I will include the #862 in the backlog and implement in the next PR.

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: Ok. Also in the next PR.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Because they are different:
- reset_votes_for_active_proposals - is a call to reset proposals votes on council elections. To clear obsolete votes for a new council.
- reactivate_pending_constitutionality_proposals - makes active again proposals with constitutionality

Even if I introduce a new combined method I would keep both `reactivate_pending_constitutionality_proposals` and `reset_votes_for_active_proposals` for clarity.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: It does several activities like increasing the proposal counter. I'm not sure I understand your note, we should probably discuss it.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: Sure.

shamil-gadelshin:
Comments:

- `Amend the constitution` proposal and `constitutionality` property are not related.

- I clear the archival state during the changes. This way I don't change the state twice. Eg.: I cleared the `Stake` and `ActiveStake` archival data during the work on staking.

- I agree. I will change the total number proposal counter to include proposals with `PendingExecution` and `PendingConstitutionality`.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Added comments for the code.

runtime-modules/constitution/src/lib.rs
shamil-gadelshin: I removed the field after our discussion.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: I will remove such caches in the next PR.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: I refactored it after our discussion.

runtime-modules/constitution/src/lib.rs
shamil-gadelshin: Dropped.

bedeho (approved):

Issue #1522, opened 2020-10-12
CLI: Curating content


Lezek123:
Since marking content as curated needs to involve at least 2 separate operations (creating `CurationStatus` entity and updating `Video`/`Channel` entity) with the current content directory schemas structure, and since the lower-level commands that would allow creating, updating and removing any entity don't exist yet, it seems reasonable to have a simple higher-level command for specifying/changing the curation status:

- `content-directory:setCurationStatus CLASS ENTITY_ID STATUS [RATIONALE]`, ie. `content-directory:setCurationStatus Channel 15 1` or `content-directory:setCurationStatus Video 20 0 "This video breaks some rules"` (currently the status will be either `0` or `1`, signaling whether the content was accepted or rejected)

By also exposing already existing commands like `updateVideo` and `updateChannel` for maintainers, we would already have basic moderation availble, which together with already implemented commands should give enough tools to handle the core activities planned for `Babylon` (which can be later expanded by implementig https://github.com/Joystream/joystream/issues/1442)

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1536

PR #1536, opened 2020-10-14
CLI - content curation

Lezek123:
Implements https://github.com/Joystream/joystream/issues/1522
Contains changes from: https://github.com/Joystream/joystream/pull/1516, https://github.com/Joystream/joystream/pull/1530 and https://github.com/Joystream/joystream/pull/1532

**Brief summary:**
- Added `media:curateContent` command with `--id` (entity id), `--status` (Accepted/Rejected), `--className` (Video/Channel) and `--rationale` flags (the last one is optional)
- Added `--asCurator` flag for `media:updateChannel` and `media:updateVideo` commands

Running `media:curateContent` or one of the update commands with `--asCurator` flag will cause the extrinsics to be executed in `Curator` (`Maintainer`) actor context. If multiple contexts are available for given account, user will be able to select the `Curator` and/or `Group`. Access will be denied unless correct `Maintainer` privileges will be available in given actor context. 

**Testing:**

Some setup is required in order to get access to the curation actions:
1. `content-directory:createCuratorGroup` (_In the next steps I'll assume that the created group id is 1_)
1. `content-directory:addMaintainerToClass CurationStatus 1`
1. `content-directory:addMaintainerToClass Channel 1` - at this point the members of this group will be able to: update any channel, set curation status of any channel
1. `content-directory:addMaintainerToClass Video 1`
1. `content-directory:addMaintainerToClass License 1` - required in order to update video license, which is part of `media:updateVideo`
1. `content-directory:addCuratorToGroup 1 0` - here I assume the `groupId` to be `1` and `curatorId` to be `0`. This command can also be ran without arguments (just like all the previous commands) in order to manually select the group and curator.
1. `content-directory:setCuratorGroupStatus 1 1` - activate the group

At this point the curator added to the group should be able to execute all the commands described in _Brief summary_ section.
The curator group can be verified by running: `content-directory:curatorGroup 1`


mnaamani:
So this works great. One small fix needed when displaying curation status when updating, the comment field is showing as boolean rather than text:

```
yarn joystream-cli media:curateContent -c Video -s Accepted --id 98
yarn run v1.22.4
$ /Users/mokhtar/joystream/joystream/node_modules/.bin/joystream-cli media:curateContent -c Video -s Accepted --id 98
Existing curation status found!
{
    approved: false,
    comment: false,
    entityId: 98
}
It will be updated to...
{
    approved: true,
    comment: false,
    entityId: 98
}
 Do you confirm this operation? Yes

Sending contentDirectory.transaction extrinsic...
Extrinsic successful!
✨  Done in 9.62s.
MokoBook:cli mokhtar$ yarn joystream-cli content-directory:entity 99
yarn run v1.22.4
$ /Users/mokhtar/joystream/joystream/node_modules/.bin/joystream-cli content-directory:entity 99
ID                            99             
Class name                    CurationStatus 
Class ID                      3              
Supported schemas             [0]            
Controller                    Maintainers    
Frozen                        false          
Refrecencable                 true           
Same owner references         0              
Total references              1              

________________ Property values _________________

approved                      true Single<Bool>      
comment                       [not set] Single<Text> 
entityId                      98 Single<Uint64>      
```

Perhaps when displaying list of my videos we can show a column with curation status? Otherwise user needs to do two calls to `content-directory:entity video_entity_id` first to find the entity id of the curation status, and then `content-directory:entity cuation_status_entity_id` .. but this is just a convenience. However it brings up a general point about the output of the cli and how we can enhance it by optionally outputting JSON instead of a pretty table (for human eyes).
This makes the cli more "unixy" and usable in a scripts. I'll create an issue to elaborate and show some examples of what I mean.

mnaamani (approved):
Not sure if it is worth fixing now, the rendering of the curation status, since you are planning to update the schema?

mnaamani:
>  I'll create an issue to elaborate and show some examples of what I mean.

https://github.com/Joystream/joystream/issues/1571

Lezek123:
Simplified `CurationStatus` into `isCensored` field of `Video` and `Channel` entities.

This is a less flexible solution, but temporarly allows us to avoid additional complexity on the CLI and query node part.

In the future it would really help to have `is_locked_from_maintainer` field that cannot be initialized to any value during `addSchemaSupportToEntity` (this way we could easily have multiple curation statuses like `NotReviewed`, `Verified` and `Censored` and we wouldn't have to worry that the user will, for example, initialize the entity with curation status already set to `Verified`)

PR #1538, opened 2020-10-15
add orion/atlas integration

### LINKED ISSUE #1488: Orion and Atlas integration
Once the initial implementation of Orion is ready (#1276), we should implement the integration with Atlas. It's not yet clear how exactly that integration should look like. There are a few approaches to consider:
- Run two separate Apollo Client instances, have one responsible only for Orion requests. This way video counts will probably need to be fetched on one by one basis
- Try out client-side stitching, although we will need to consider how much effort that requires and whether that's worth doing for this release. Some links:
  - https://hasura.io/blog/client-side-graphql-schema-resolving-and-schema-stitching-f4d8bccc42d2/
  - https://github.com/hasura/client-side-graphql/tree/master/examples/schema-stitching

kdembler:
This PR adds an initial integration of Orion into Atlas, allowing the basic view count functionality. Orion was also mocked via MirageJS so that the functionality can be easily tested.

**Current limitations:**
- View counts are resolved synchronously only after the video resolving completes, potentially stretching out the complete request time. This can be further improved and should be possible to achieve in a parallel matter without much effort
- Views counts are fetched on a one-by-one basis, because of the limitation of the first point. Again, we should be able to switch to the batched request without much effort
- After the discussion on schema files, I've decided to drop the extension schema for now and instead have one file that describes the full schema used by Atlas (`extendedQueryNode.graphql`). This has both pros and cons. The advantage is that the tooling should be able to handle that no problem - current codegen works and some doc generation should be the same. The disadvantage is that it's not clear what's the exact schema used by the Query Node. Let's see how this works out and potentially revisit in the future

Resolves #1488


fcescob (approved):
Great work on the stitching! Lots of non trivial stuff required to make it work!

bedeho (approved):

PR #1543, opened 2020-10-15
Orion deployment improvements

### LINKED ISSUE #1486: Orion deployment
Figure out the possibilities for Orion deployment:
- Where does it run?
- What's the Devops (some CI?) for it?

Once figured out, do that

kdembler:
Resolves #1486: makes Orion configurable via env variables and adds a Docker setup

### DIFF
--- a/orion/.dockerignore
+++ b/orion/.dockerignore
@@ -1,0 +1,5 @@
+*.log
+node_modules/
+dist/
+Dockerfile
+docker-compose.yml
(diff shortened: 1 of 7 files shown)


fcescob (approved):
Looks good.
It might be out of the scope of this PR, but I was wondering if we should use something like [pm2](https://github.com/Unitech/pm2) for restarting Orion if an exception occurs?

kdembler:
That's a good idea, will add that here

kdembler:
Not gonna add it here, after all, sent the explanation to Francesco over chat

bedeho (approved):

PR #1548, opened 2020-10-16
CLI: Working-groups UX improvements

Lezek123:
A small CLI UX improvement for `working-groups` commands, now that they can be executed in context of either `curators` or `storageProviders` group.

Users will be able to use `working-groups:setDefaultGroup` in order to set the default context for `working-groups` commands and avoid having to specify `-g storageProviders` or `-g curators` every time, which can get hard to manage (as I experienced while doing some tests).


mnaamani:
Code works as expected.

I can see how having a default group saved in the state can be convenient, but I personally favor explicit selection on the command line.

If the selected account is a worker in multiple groups they could inadvertently make the call to the wrong group (although for some command where you need to specify args to reference exisiting application ids, openings etc they wouldn't be correct). 

But since they can equally specify the wrong group on the command line I'm not sure if this is any better or worse. User could end up having to check what the current group is, by calling `working-groups:setDefaultGroup` without args.

Am I over thinking this? What do you think Martin @bwhm

mnaamani:
needs merge conflict fix.
I think I will merge this after rethinking it. Its up to the user to decide if they want to rely on the feature or not.

mnaamani:
I did a bad job here fixing the merge conflict. Will try to resolve

mnaamani:
@Lezek123 I think I will leave it to you to fix, and force push an update.

mnaamani (approved):

PR #1554, opened 2020-10-19
Integrate Gliderjs

fcescob:
This PR removes the old carousel and replaces it with one using Gliderjs


kdembler:
Is this supposed to be like this (with a lot of blank space left)?

Screenshot 2020-10-19 at 2 44 07 PM

Also, the navigation arrows are pretty small and it's hard to press them

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -1,94 +1,52 @@
...
-  disableControls?: boolean
-  onScroll?: (direction: 'left' | 'right') => void
-} & CarouselStyleProps
+import Icon from '../Icon'
 
+export type CarouselProps = {
+  trackPadding?: string
+} & GliderProps
+
+const Track: React.FC<any> = ({ className = '', ...props }) => (
```
kdembler: Why `any`?

kdembler:
One more thing I've noticed: the scrolling arrows don't work for featured videos and channels carousels - only the first one

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -1,94 +1,52 @@
...
+  <div className={`glider-track ${className}`} {...props} />
+)
+
+const defaultBreakpoints: BreakPoint[] = [
+  {
+    breakpoint: 400,
```
kdembler: At this point, it would probably make sense to define some app-wide breakpoints. We have those and also the breakpoint for the navbar, I'm sure we'll have some more. What do you think?

atlas/package.json
```diff
@@ -80,6 +80,7 @@
     "eslint-plugin-react-hooks": "^4.0.4",
     "faker": "^5.1.0",
     "fluent-ffmpeg": "^2.1.2",
+    "glider": "^0.1.0",
```
kdembler: This is a library for PostgreSQL, let's remove that and regenerate `yarn.lock`

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -1,94 +1,70 @@
...
-      case 'right': {
-        // Prevent overscroll on the right
-        const newOffset = carouselOffset + containerWidth
-        scrollAmount = newOffset > maxScrollOffset ? maxScrollOffset : newOffset
-        onScroll('right')
-        break
-      }
+const Carousel: React.FC<CarouselProps> = ({ children, trackPadding = '0', className, ...gliderProps }) => {
+  // Using any because the GliderMethods type only has methods and I need the full instance
+  const gliderRef = React.useRef<any>()
```
kdembler: You import `useRef`, so you don't need `React.` prefix here. We always import needed methods directly, so let's keep that consistent and remove `React.` prefix across this file

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -1,94 +1,70 @@
...
+  const [arrows, setArrows] = React.useState<
+    { prev: string | HTMLButtonElement; next: string | HTMLButtonElement } | undefined
+  >(undefined)
+
+  React.useLayoutEffect(() => {
+    if (gliderRef.current) {
+      const glider = gliderRef.current.ele
+      const prevArrow = glider.previousSibling
+      const nextArrow = glider.nextSibling
+      const INSTANCE_KEY = Math.round(Math.random() * 1000)
```
kdembler: If I understand correctly, this has a 1/1000 chance of being a duplicate, hence potentially breaking some other carousel in the same view?

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -1,94 +1,70 @@
...
-        const newOffset = carouselOffset + containerWidth
-        scrollAmount = newOffset > maxScrollOffset ? maxScrollOffset : newOffset
-        onScroll('right')
-        break
-      }
+const Carousel: React.FC<CarouselProps> = ({ children, trackPadding = '0', className, ...gliderProps }) => {
+  // Using any because the GliderMethods type only has methods and I need the full instance
+  const gliderRef = React.useRef<any>()
+  const [arrows, setArrows] = React.useState<
+    { prev: string | HTMLButtonElement; next: string | HTMLButtonElement } | undefined
```
kdembler: Looking at the code, those always will be refs, not strings, why this type?

atlas/src/shared/components/Carousel/Carousel.tsx
kdembler: Maybe it'd be better to use something like `GliderMethods & {ele: HTMLDivElement}` instead of `any`?

atlas/src/shared/components/Carousel/Carousel.tsx
fcescob: This was mainly because the actual correct type of `gliderjs` accepts both a string and an element, I've changed this and added a type conversion when passing them to the `Glider` component (since `react-glider` expects them both to be strings)

atlas/src/shared/components/Carousel/Carousel.tsx
fcescob: Correct. Being such a small chance `#carousels on the same page / 1000` I didn't bother to add a real library like `uuid` to take care of this, but if you think this might be an issue, I'll add it.

atlas/src/shared/components/Carousel/Carousel.tsx
fcescob: Thanks, removed

atlas/package.json
fcescob: Thanks, removed it

atlas/src/shared/components/Carousel/Carousel.tsx
kdembler: We currently have 3 carousels in the home view, so statistically, every 333rd visitor will encounter a bug, that's not a very good score

fcescob: I just realised that an `ISTANCE_KEY` is only useful if the `arrows` object holds strings, but since we're passing the elements directly (and those are tied to a specific instance) there was no need to add the class and have the `INSTANCE_KEY`

kdembler (approved):
LGTM

PR #1558, opened 2020-10-20
network-tests: improve test runner

mnaamani:
script can take scenarion arg
initialize content-dir before executing scenario
add content-directory scenario
default now doesn't do a runtime upgrade, specify starting runtime with RUNTIME=alexandria if you want to test with a runtime upgrade.

From inside the `tests/network-tests` directory

To run the `full` test scenario: `src/scenarios/full.ts`
`./run-tests full`

To run the draft content-directory scenario: `src/scenarios/content-directory.ts`

`./run-tests content-directory`

I also added a job to the github workflows to run the new content-directory scenario


iorveth (approved):
LGTM!

PR #1561, opened 2020-10-21
Query Node Integration in CI

mnaamani:
Setup a Query-Node/Hydra dev instance before running content-directory integration tests
Will need to update the script to copy custom joystream mapping and graphaql schema files once they are added by @metmirr

To test:

`./tests/query-node/run-tests.sh`

per-requisits:
`docker`

`joystream/node` docker image:
```
docker pull joystream/node:babylon
docker image tag joystream/node:babylon joystream/node:latest
```

`npx`: install with `npm -g npx`


mnaamani:
@dzhelezov  One improvement I can suggest for the indexer is to have it retry connecting to the joystream-node a few times, instead of immediately exiting. I know the docker contrainer is configured to restart, but reconnecting might be better. I've had to do that in our testing scripts as well because you can't always guarantee the endpoint is ready because of startup delays. (doing "curl polling" and/or `sleep` also in some cases is adequate but makes the scripts a little messy) -- eg: https://github.com/Joystream/joystream/blob/7713c5840a8f423d185fba5579930b0f52aff6ff/tests/network-tests/src/Api.ts#L48

iorveth (approved):
LGTM!

metmirr (approved):

tests/query-node/run-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+npx @dzlzv/hydra-cli scaffold --wsProviderUrl=ws://joystream-node:9944/ --projectName=Test
+
+cp ../../../types/augment/all/defs.json ./mappings/typedefs.json
+export TYPES_JSON=../../mappings/typedefs.json
+
+# Joystream mappings and graphsql schema
+# cp ../../../query-node/mappings/* ./mappings/
+# cp ../../../query-node/schema.graphql ./
+
+npx @dzlzv/hydra-cli codegen
```
dzhelezov: Probably a more isolated and thus safer way would be to run `db:migrate` from a docker container, and in then  we don't have to run `codegen`:

```
docker build . -f docker/Dockerfile.hydra -t hydra-indexer:latest
docker-compose up $1 -d db
docker run --env DB_HOST=db --env TYPEORM_HOST=db --env-file .env --network=hydra-test_default hydra-indexer:latest yarn db:migrate 
```
and then
```
docker-compose up $1 -d
```

tests/query-node/run-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+function cleanup() {
+    docker-compose down -v
+}
+
+trap cleanup EXIT
+
+yarn db:migrate
+
+# Bring up remaining services
+docker-compose up $1 -d
```
dzhelezov: We have to be sure that `hydra-indexer` image is always fresh. One option is that  I can add `:latest` tag to the compose file template, so, assuming the latest image is built at the previous step, ensures it is fresh. Or perhaps simply clean all the images afterwards?

dzhelezov (changes requested):

dzhelezov:
> @dzhelezov One improvement I can suggest for the indexer is to have it retry connecting to the joystream-node a few times, instead of immediately exiting. I know the docker contrainer is configured to restart, but reconnecting might be better. I've had to do that in our testing scripts as well because you can't always guarantee the endpoint is ready because of startup delays. (doing "curl polling" and/or `sleep` also in some cases is adequate but makes the scripts a little messy) -- eg:
> 
> https://github.com/Joystream/joystream/blob/7713c5840a8f423d185fba5579930b0f52aff6ff/tests/network-tests/src/Api.ts#L48

Thanks! Will introduce retries at startup in the next indexer-builder-lib release

tests/query-node/run-tests.sh
mnaamani: I had the same thought, which is why I added the `$1` there so you can pass an optional `build` argument when calling the script, to force rebuilding if you choose to. Or maybe the default should be to always build and optionally skip rebuild.

tests/query-node/run-tests.sh
mnaamani: sounds like a better approach yes, will update

tests/query-node/run-tests.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_PATH="$(dirname "${BASH_SOURCE[0]}")"
+cd $SCRIPT_PATH
+mkdir -p hydra-test/
+cd hydra-test
+
+npx @dzlzv/hydra-cli scaffold --wsProviderUrl=ws://joystream-node:9944/ --projectName=Test
```
mnaamani: @dzhelezov should this initial execution of the scaffold command be specific version of hydra-cli perhaps?

Is there a way to include the version in the package name when calling npx? like:  `npx @dzlzv/hydra-cli:v0.0.14 ...`

mnaamani: ah i think its like this:
`npx @dzlzv/hydra-cli@v0.0.14`

dzhelezov: Actually, the scaffold here is going to be used only to 1) create .env file 2) create Dockerfile 3) create a docker-compose file. Perhaps we can replace this dependency altogether and use fixed files?

dzhelezov: The actual dependencies are all hidden in the hydra-cli docker image in Dockerfile.hydra

tests/query-node/run-tests.sh
dzhelezov: FYI, I just published hydra-cli@0.0.16 where I added a few convenenience yarn targets. The whole process can now be a bit streamlined: 
```
yarn docker:db:up # runs db from docker-compose
yarn docker:build # builds an image with the schema, mappings and types.json, it's called hydra-kit
yarn docker:db:migrate # migrate the db
yarn docker:up # runs docker-compose up -d
```

There scripts above are more or less repeat what we discuss above, so it may or may not be convenient to use them

mnaamani:
Updated with latest hydra-cli release and used recommended steps.

I'm seeing the following logs though for the indexer:

```
2020-10-24T08:57:57.229Z index-builder:processor Spawned the processor
Error: Current indexer head block is not yet published: {"response":{"errors":[{"message":"Current indexer head block is not yet published","locations":[{"line":3,"column":3}],"path":["indexerStatus"],"extensions":{"code":"INTERNAL_SERVER_ERROR"}}],"data":null,"status":200},"request":{"query":"\nquery {\n  indexerStatus {\n    head\n  }\n}\n"}}
    at GraphQLClient.<anonymous> (/home/hydra/generated/indexer/node_modules/graphql-request/src/index.ts:101:13)
    at step (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:63:23)
    at Object.next (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:44:53)
    at fulfilled (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:35:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
And I don't see it logs anything about processing any blocks after that.

Note the joystream-node is started after the hydra services. Is something off?

dzhelezov:
> Updated with latest hydra-cli release and used recommended steps.
> 
> I'm seeing the following logs though for the indexer:
> 
> ```
> 2020-10-24T08:57:57.229Z index-builder:processor Spawned the processor
> Error: Current indexer head block is not yet published: {"response":{"errors":[{"message":"Current indexer head block is not yet published","locations":[{"line":3,"column":3}],"path":["indexerStatus"],"extensions":{"code":"INTERNAL_SERVER_ERROR"}}],"data":null,"status":200},"request":{"query":"\nquery {\n  indexerStatus {\n    head\n  }\n}\n"}}
>     at GraphQLClient.<anonymous> (/home/hydra/generated/indexer/node_modules/graphql-request/src/index.ts:101:13)
>     at step (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:63:23)
>     at Object.next (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:44:53)
>     at fulfilled (/home/hydra/generated/indexer/node_modules/graphql-request/dist/index.js:35:58)
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
> ```
> 
> And I don't see it logs anything about processing any blocks after that.
> 
> Note the joystream-node is started after the hydra services. Is something off?

This error is thrown when the indexer endpoint the processor is connecting to haven’t reported a single block. It should resolve itself after the indexer starts the work (and the processor restarts). I will think how to handle it more gracefully, probably throwing an error is not the best approach here

PR #1562, opened 2020-10-21
Add Mobile And Table Screens

fcescob:


atlas/src/shared/components/Grid/Grid.tsx
```diff
@@ -1,21 +1,42 @@
 import React from 'react'
 import styled from '@emotion/styled'
 import useResizeObserver from 'use-resize-observer'
+import { useMediaQuery } from 'react-responsive'
 import { spacing } from '../../theme'
 
-const Container = styled.div<GridProps>`
+const toPx = (n: number | string) => (typeof n === 'number' ? `${n}px` : n)
+
+const LARGE_VIEWPORT_BREAKPOINT = toPx(2000)
```
kdembler: Maybe let's move that to `theme.breakpoints` so we don't have a bunch of breakpoints scattered across the source files

atlas/src/shared/components/Grid/Grid.tsx
```diff
@@ -25,11 +46,14 @@ const Grid: React.FC<GridProps> = ({ children, className, gap = spacing.xl, repe
       }
     },
   })
+  const isLargeViewport = useMediaQuery({ query: `(min-width: ${LARGE_VIEWPORT_BREAKPOINT})` })
```
kdembler: Why not use a regular CSS media query here, without JS?

atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -13,6 +13,9 @@ type ChannelProps = {
 
 export const CoverImage = styled.img<CoverImageProps>`
   display: block;
+  position: absolute;
```
kdembler, line 16: I don't think the image should be `absolute` here. What's stopping us from having it be a part of the layout?

atlas/src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -19,10 +20,14 @@ export const TitleSection = styled.div`
   display: flex;
   align-items: center;
   padding-top: ${theme.sizes.b10}px;
+  @media (max-width: ${theme.breakpoints.mobile}) {
```
kdembler: I see we both made improvements to that view, that's probably my bad. I added a couple more things though I think so let's get this merged and I'll rebase

atlas/src/components/VideoBestMatch/VideoBestMatch.style.tsx
```diff
@@ -8,8 +8,12 @@ export const Container = styled.div`
 
 export const Content = styled.div`
   display: grid;
-  grid-template-columns: 650px 1fr;
+  grid-template-columns: minmax(50%, 650px) 1fr;
```
kdembler: This isn't related to this piece of code. But the best match video still suffers from the problem of changing the image aspect ratio when resizing. Also take a look at how it looks like at `500px`, this needs improvement:

Screenshot 2020-10-28 at 12 43 39 PM

atlas/src/shared/components/Grid/Grid.tsx
fcescob: Good Idea, moved.

atlas/src/shared/components/Grid/Grid.tsx
fcescob: For some reason it was not working properly when I tried, but now looks fine, so I removed this and went back to regular CSS.

atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
fcescob: I'm setting the image to be `absolute` because I'm using a "trick" to preserve the aspect ratio of the images.
I absolutely position each image in a `div` which has a width of 100% and a `padding-top` of `100% * aspectRatio`,  so the absolute positioning is required in this case. 
I'm happy to change this if there is a better way though.

atlas/src/components/VideoBestMatch/VideoBestMatch.style.tsx
fcescob: Didn't catch that, thanks. I've made some changes and now it should look better. The trick to preserve the aspect ratio I've described above is used here as well, again happy to chance if a better solution comes up.

atlas/src/views/ChannelView/ChannelView.style.tsx
fcescob: Let's leave this as is then and I'll leave it up to you to rebase / add your changes on top.

atlas/src/components/VideoBestMatch/VideoBestMatch.style.tsx
```diff
@@ -8,24 +9,36 @@ export const Container = styled.div`
 
 export const Content = styled.div`
   display: grid;
-  grid-template-columns: 650px 1fr;
+  grid-template-columns: min(650px, 50%) 1fr;
   grid-column-gap: 24px;
+
+  @media (max-width: 750px) {
```
kdembler: Shouldn't that use one of the defined breakpoints?

atlas/src/shared/theme/breakpoints.ts
```diff
@@ -1,5 +1,5 @@
 export default {
-  small: '21rem',
-  medium: '48rem',
-  large: '72rem',
+  mobile: '480px',
```
kdembler: I've seen you've renamed `large`, maybe we should rename the others as well at the same time?

atlas/src/components/VideoBestMatch/VideoBestMatch.style.tsx
```diff
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
```
kdembler, line 1: Relating to the styling since I can't comment the specific line:

Looks much better now, but I think we should also tweak the padding of `InnerContainer` once the title moves underneath the poster. Also the `Container` has right padding that doesn't seem to fit in mobile view. Take a look:

Screenshot 2020-10-30 at 12 53 48 PM

There is excess white space to the right and between the poster and the title

kdembler (approved):
LGTM

bedeho (approved):

PR #1564, opened 2020-10-21
Council candidacy withdraw

### LINKED ISSUE #1499: Council election candidates can withdraw candidacy
Announcement period lasts a long time, and circumstances can change. It may be that some people do not catch the message that someone actually does not want to stand if the candidacy still exists. They may just vote for a brand they trust or something.

ondratra:
Fixes #1499.

To minimize merge conflict potential, this PR assumes #1251 will be merged first.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -113,7 +113,7 @@ pub struct CouncilStageElection {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, PartialEq, Eq, Debug, Default, Clone)]
 pub struct Candidate<AccountId, Balance> {
-    account_id: AccountId,
+    staking_account_id: AccountId,
     cycle_id: u64,
     stake: Balance,
 }
@@ -122,7 +122,7 @@ pub struct Candidate<AccountId, Balance> {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, PartialEq, Eq, Debug, Default, Clone)]
 pub struct CouncilMember<AccountId, CouncilUserId, Balance> {
-    account_id: AccountId,
+    staking_account_id: AccountId,
     council_user_id: CouncilUserId,
     stake: Balance,
 }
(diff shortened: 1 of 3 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -738,7 +790,44 @@ impl<T: Trait> EnsureChecks<T> {
...
+        let account_id = Self::ensure_user_membership(origin, council_user_id)?;
+
+        // escape when no previous candidacy stake is present
+        if !Candidates::<T>::contains_key(council_user_id) {
+            return Err(Error::NotCandidatingNow);
+        }
+
+        let candidate = Candidates::<T>::get(council_user_id);
+
+        // ensure candidacy was announced in on of previous election cycles
```
bedeho: I think this is typo, should be "was not announced" ?

runtime-modules/council/src/lib.rs
```diff
@@ -738,7 +790,44 @@ impl<T: Trait> EnsureChecks<T> {
...
+        }
+
+        // ensure candidacy announcing period is running now
+        match Stage::<T>::get().stage {
+            CouncilStage::Announcing(_) => (),
+            _ => return Err(Error::CantWithdrawCandidacyNow),
+        };
+
+        // ensure there is no mismatch between saved and current `council_user_id`/`account_id` combination
+        if candidate.account_id != account_id {
```
bedeho: `account_id` is the controller account of the member, but `Candidate::account_id` is the staking account of the candidate, these are not the same in general. Perhaps renaming `account_id` to `staking_account_id` would increase clarity to avoid such possible confusion.

runtime-modules/council/src/lib.rs
```diff
@@ -671,15 +722,15 @@ impl<T: Trait> EnsureChecks<T> {
             return Err(Error::StakeStillNeeded);
         }
 
-        if &candidate.account_id != account_id {
-            return Err(Error::InvalidAccountToUnstake);
-        }
+        // prevent membership and account combination change
+        Self::prevent_ids_mismatch(&candidate, account_id)?;
```
bedeho: I am not sure exactly what this test is trying to do, but it has the same problem as the other comment, namling comparing membership controller accounts and staking accounts.

bedeho (changes requested):
Requires fixing merge conflict as well.

runtime-modules/council/src/lib.rs
ondratra: Yes that was a typo. Fixed

runtime-modules/council/src/lib.rs
ondratra: `staking_account_id` name was introduced and withdrawal no longer needs to compare those ids (you can authorize withdrawal with different `account_id` as long as you are using the same membership id).

runtime-modules/council/src/lib.rs
ondratra: This function is no longer needed. Due to changes mentioned in the previous review-topic above.

bedeho (approved):

Issue #1396, opened 2020-09-17
Babylon: Create initial, consumable joystream-js library


Lezek123:
During the last few weeks/months I was aiming to extract some of the potentially reusable functionality from Pioneer's packages to `pioneer/packages/src/joy-utils`. This process began with https://github.com/Joystream/apps/pull/465 and the main idea is also further described in that PR.

I think Babylon release may be a good time to extract most of the code in `joy-utils` into the first version of long-awaited `joystream-js` library.

# Benefits
## Benefit 1: Transports
Currently the main benefit of doing this is that it will make the so-called "transports" (see: https://github.com/Joystream/joystream/tree/iznik/pioneer/packages/joy-utils/src/transport) accessible for all our JavaScript projects (Pioneer, CLI, storage node, status server etc.).

Basically all of those projects currently have some custom wrapper class/classes for the api to simplify fetching, combining and parsing commonly used data. Having this normalized across projects could greatly simplify adapting them after runtime changes that affect the api/types (which currently generates a need to update the same / similar query scenarios in all projects separately).

Some examples of data querying scenarios and how they are simplified via transports:

* Fetching all of given working group opening applications along with applicants' profiles and their stakes.
_This currently requires executing tons of separate queries, talking to lots of different runtime modules and parsing a lot of the data, but using the existing transport we can just do it with: `const applications = await transport.workingGroups.openingApplications(group, openingId)`_
* Fetching the entire proposal discussion thread along with all the posts and involved users' profiles.
_This requires fetching the related thread, then executing some still relatively complex logic of fetching all double map entries by first key, then fetching member profiles, combining results, parsing dates etc. With transport we just do: `const discussion = await transport.proposals.discussion(proposalId)`_
* Fetching a batch of proposals filtered by status, containing data like: proposal details, proposal type (comes from ProposalDetails enum), approval/slashing quorum, proposer profile etc. (it also contains information that allows easy pagination of the results, like total number of proposals that match the filter)
_Available via `transport.proposals.proposalsBatch(status, batchNumber, batchSize)`. Some of this data we can't even query from the chain, that's where `joy-utils/consts` kick in (see below in `Common constants`)_

There are lots of other similar common scenarios that are already covered by the transport classes inside `joy-utils` (some of them very simple, like querying timestamp based on block number)

Other benefits that shared transports provide:

[...]

PR #1572, opened 2020-10-22
Initial @joystream/js library

Lezek123:
Partialy implements https://github.com/Joystream/joystream/issues/1396

- Moves most of the reusable functionality (transports, helpers, constants etc.) from `pioneer/joy-utils` to `joystream-js`
- Moves all working-groups-related transport methods from CLI to `joystream-js`
- Introduces a way to use augmented `api.tx` / `api.query` without the need to globally change `paths` in project's `tsconfig`:
This can be now done with:
```
import Transport from '@joystream/js/transport'
/* ... */
const transport = new Transport(api)
transport.query.contentDirectory.entityById // Augmentation provided
transport.tx.contentDirectory.transaction // Augmentation provided
```
- Adds basic CI checks for `@joystream/js`

Having `@joystream/js` as a separate library gives us a place where we can put some reusable functionality once we decide it may be valueable for other projects. This seems to be quite important feature, because I think it is often the case that a lot of code ends up beeing copy-pasted or rewritten between CLI, Pioneer, Storage Node, Network Tests etc. (soon there will also be query node mappings merged and they may also benefit from having access to shared methods of parsing some data like content directory entities etc. in the future)

**Not yet implemented:**

Due to the fact that the process takes more time than expected and this PR has already grown substantially, I wasn't able to implement all the things I planned in the original issue yet, ie.:

- **Documentation** (requires some research w.r.t. implementing a tool like http://typedoc.org/ in order to make it easier to generate docs)
- **Re-use of the working groups transport in Pioneer (`joy-utils`)** - the transport methods are currently shared between CLI and Pioneer's `joy-proposals`, while `joy-roles` still rely on `joy-roles/transport.substrate.ts`. It is not easy to switch the implementation in `joy-roles` in a way that wouldn't require refactoring a huge chunk of code (and generate a need for quite a lot of testing), so it seemed risky to include it as part of the Babylon scope.

**Integration with `cd-schemas`**
If we don't want to manage `cd-schemas` and `joystream-js` as 2 separate libraries we can probably merge them at some point.
Here's an idea of what the structure may look like:
```
cd-schemas/inputs/ => joystream-js/json/content-directory/inputs/
cd-schemas/schemas/ - joystream-js/json/content-directory/schemas/
cd-schemas/types/ => joystream-js/types/content-directory

scripts/devInitAliceLead => move the extrinsics part to joystream-js/transport/workingGroups and the script itself perhaps to CLI?
scripts/initializeContentDir => cli/commands/content-directory/initialize
scripts/inputSchemasToEntitySchemas - joystream-js/scripts/content-directory/generateEntitySchemas.ts
scripts/validate - joystream-js/scripts/validate

(rename joystream-js/functions => joystream-js/helpers)
helpers/InputParser - joystream-js/helpers/content-directory/InputParser
helpers/inputs.ts - joystream-js/helpers/content-directory/inputs.ts
helpers/propertyType.ts - joystream-js/helpers/content-directory/propertyType.ts
helpers/schemas.ts - joystream-js/helpers/content-directory/schemas.ts
helpers/extrinsics.ts - joystream-js/helpers/extrinsics.ts
```
I didn't try to do this yet, because `cd-schemas` is quite dynamic part of the monorepo at this moment and will likely undergo a few changes in the near future. I'm also not entirely sure if it's a good idea to merge them completely (perhaps separating some of the tooling from `@joystream/js` may actually be of value to prevent it from becoming too huge)


mnaamani:
Babylon branch is no longer active, work here will need to be re targeted to another branch.

PR #1573, opened 2020-10-22
Implement landing page in Overview

DzhideX:
- Some functions in the transport could be moved to other transports, I left it all here for now and we can move them after the review.
- To keep the clutter out of the visual part of the landing, I decided to do all of  the calculations and preparations for data visualization in the transports. I thought this was the easiest way to do it but if it goes against what transports are for then we can change that (relatively) easily as well.

### DIFF
--- a/pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
+++ b/pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import styled from 'styled-components';
+import usePromise from '@polkadot/joy-utils/react/hooks/usePromise';
+import { useTransport } from '@polkadot/joy-utils/react/hooks';
+
+const Subtitle = styled('h3')`
+    font-weight: 600;
+    color: rgba(0,0,0, 0.65);
+    margin-bottom: 1rem;
+`;
+
+const GetStarted = () => {
+  const transport = useTransport();
+  const [data] = usePromise(() => transport.proposals.proposalsBatch('Active'), undefined, []);
+
+  return <>
+    <p>If you&apos;re new to Joystream, go to our <a href='https://github.com/Joystream/helpdesk/'>Helpdesk</a> to learn what you can do, and how to get started. If you&apos;re not there already, join our <a href='https://t.me/joinchat/CNyeUxHD9H56m3e_44hXIA'>telegram group</a>, get some tokens, and join our community!</p>
+    <Subtitle>Membership</Subtitle>
+    <p>Most actions on the platform requires you to become a <a href='/#/members/edit'>Member</a>.</p>
+    <Subtitle>Council</Subtitle>
+    <p>As &quot;The video platfrom DAO&quot;, Joystreams day to day operations are controlled by the <a href='/#/council/members'>Council</a> through a Proposal system.</p>
+    <Subtitle>Proposals</Subtitle>
+    <p>There are currently {data?.proposals.length || 'no'} <a href='/#/proposals'>Proposals</a> open for voting.</p>
+    {data?.proposals.length
+      ? <ul>
+        {data.proposals.map((proposal, index) => <li key={index}><a href={`/#/proposals/${proposal.id.toString()}`}>{proposal.title}</a> opened by <a href={`/#/members/${proposal.proposer.handle}`}>{proposal.proposer.handle}</a></li>)}
+      </ul> : null
+    }
+    <Subtitle>Roles</Subtitle>
+    <p>In addition to being a Council Member, the other &quot;official&quot; roles are:</p>
+    <ul>
+      <li><a href='/#/staking'>Validator</a> (and Nominators)</li>
+      <li><a href='/#/working-groups'>Storage Provider</a></li>
+      <li><a href='/#/working-groups'>Content Curator</a></li>
+    </ul>
+    <Subtitle>Media</Subtitle>
+    <p>Go <a href='/#/media'>here</a> to..</p>
+    <p>Although there are no on-chain rewards for doing so, you can also create a <a href='/#/media/channels/new'>Channel</a> and upload videos, just make sure they are in line with the platforms <a href='https://www.joystream.org/privacy-policy/'>Terms of Service</a>.</p>
+  </>;
+};
+
+export default GetStarted;
--- a/pioneer/packages/joy-tokenomics/src/Landing/Roles/Role.tsx
+++ b/pioneer/packages/joy-tokenomics/src/Landing/Roles/Role.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Card } from 'semantic-ui-react';
+import RoleCardContent from './RoleCardContent';
+import RoleCardDescription from './RoleCardDescription';
+import { WorkingGroupExtra, CouncilExtra } from './RoleCardExtra';
+import { LandingPageCouncilExtra, LandingPageWorkingGroupExtra } from '../../../../joy-utils/src/types/tokenomics';
+
+const StyledRoleCard = styled(Card)`
+  width: 100% !important;
+  margin: 0 !important;
+  .circle.notched.loading.icon {
+    margin: 2rem auto !important;
+  }
+`;
+
+const StyledRoleCardContent = styled(RoleCardContent)`
+  margin-top: 1rem;
+  display: grid;
+  grid-template-columns: repeat(${(props) => (props ? props.cellData.length : 1)}, 1fr);
+  div:first-child {
+    border-left: none;
+  }
+  @media (max-width: 700px) {
+    grid-template-columns: none;
+    grid-template-rows: repeat(${(props) => (props ? props.cellData.length : 1)}, 1fr);
+    grid-row-gap: 1rem;
+  }
+`;
+
+const StyledRoleCardHeader = styled(Card.Content)`
+  background-color: rgba(0, 0, 0, 0.05) !important;
+`;
+
+const Role: React.FC<{
+  title: string
+  roleData: [string, string][]
+  extraData?: LandingPageWorkingGroupExtra[] | LandingPageCouncilExtra[]
+}> = ({ title, roleData, extraData }) => {
+  return (
+    <StyledRoleCard>
+      <StyledRoleCardHeader textAlign='center' header={title} />
+      {extraData && (extraData as any[]).map((data: LandingPageWorkingGroupExtra | LandingPageCouncilExtra, index: number) => (
+        <div style={{ margin: '0.5rem 0 1rem', padding: '0.5rem' }} key={index}>
+          <h3 style={{ textAlign: 'center', color: data.title.color }}>{data.title.value}</h3>
+          <StyledRoleCardContent cellData={data.cellData} />
+          {data.type === 'WorkingGroup' ? (
+            <WorkingGroupExtra href='/#/working-groups/opportunities' />
+          ) : (
+            <CouncilExtra
+              termReward={data.extra.termReward}
+              termMaxBonus={data.extra.termMaxBonus}
+              text={data.extra.text}
+              buttonText={data.extra.button.text}
+              buttonHref={data.extra.button.href}
+            />
+          )}
+        </div>
+      ))}
+      {extraData?.[0]?.type === 'Council' && (
+        <h3 style={{ textAlign: 'center', marginTop: '0', fontStyle: 'italic', fontWeight: 600 }}>Current Council</h3>
+      )}
+      <StyledRoleCardContent cellData={roleData} />
+      <RoleCardDescription group={title}/>
+    </StyledRoleCard>
+  );
+};
+
+export default Role;
--- a/pioneer/packages/joy-tokenomics/src/Landing/Roles/RoleCardExtra.tsx
+++ b/pioneer/packages/joy-tokenomics/src/Landing/Roles/RoleCardExtra.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Button } from 'semantic-ui-react';
+
+const WorkingGroupExtraButton = styled(Button)`
+    margin-top: 1.5rem !important;
+    margin-left: 0.5rem !important;
+    @media (max-width: 700px){
+        width: 100% !important;
+        margin-left: 0rem !important;
+    }
+`;
+
+export const WorkingGroupExtra: React.FC<{href: string}> = ({ href }) => {
+  return (
+    <WorkingGroupExtraButton href={href}>Learn More and Apply</WorkingGroupExtraButton>
+  );
+};
+
+const CouncilExtraWrapper = styled('div')`
+    display:flex;
+    justify-content: space-evenly;
+    margin-top: 1rem;
+    text-align:center;
+    @media (max-width: 700px){
+        flex-direction: column;
+        margin-top:0rem;
+    }
+`;
+
+const CouncilExtraCell = styled('div')`
+    display: flex;
+    align-items: center;
+    justify-content: space-evenly;
+    flex-direction: column;
+    margin:0 1rem;
+    h2{
+        margin: 0;
+    }
+    @media (max-width: 700px){
+        margin-top:1rem;
+        .ui.button{
+            width: 100%;
+        }
+    }
+`;
+
+export const CouncilExtra: React.FC<{termReward: string, termMaxBonus: string, text: string, buttonText: string, buttonHref: string}> = ({ termReward, termMaxBonus, text, buttonText, buttonHref }) => {
+  return (
+    <CouncilExtraWrapper>
+      <CouncilExtraCell>
+        <p>TERM REWARD</p>
+        <h2>{termReward}</h2>
+      </CouncilExtraCell>
+      <CouncilExtraCell>
+        <p>TERM MAX BONUS</p>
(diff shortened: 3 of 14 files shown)


bwhm (commented):
Looks like a great start to me!

@blrhc 
Do you mind having a look, and come with your suggestions wrt to info and links?
You can have a look [here](https://alexandria-testing-2.joystream.app/#/overview)

blrhc:
I'm really glad we're adding this - at the moment it is basically impossible for participants to know where to start with Pioneer.

I think the current section headings are good, but it would also be helpful to add:
- Keys and Addresses
- Get Tokens
- Redeeming Tokens
- Advanced

**Keys and Addresses** might say something like:
```
In order to do anything useful, you will need to generate a key. You can do that by visiting this page.

Click Add Account and give your key a name and password. Make sure to download the JSON 
file to your computer so you have a backup of your account!
```

**Get Tokens** could say:
```
Now that you have an account to store your tokens on, you can use this address to request tokens. 
Your address should look something like this:

5CZmyWhCcQ5fKbbmyFX458iG4TjuHbi8D6akqNdsbMYnzBwD

The easiest and quickest way to get tokens is through our Telegram group. Simply leave a message 
requesting tokens and one of our helpful community members will send them to you!
```

**Redeeming Tokens** could say:
```
After earning tokens on our testnet through one of the roles, these can be exchanged for real money!

All tokens (tJOYs) issued on the Constantinople network are backed by USD, held in trust by Jsgenesis, 
the team developing Joystream.

To redeem tokens, simply set the memo of your account to your Monero address. You can then visit this 
page which explains how to burn tokens to redeem them for XMR.
```

**Advanced**:
- explains how to change Pioneer from "Basic Features Only" to "Fully Featured"
- provides a (very) brief explanation of extrinsics tab and chain state

I am also going to comment some spelling/grammar suggestions below.

blrhc:
I also think the first line under `Get Started` can be removed, as we'll be providing instructions on all of the relevant stuff below. We should provide a link to the helpdesk somewhere, but probably in the roles section rather than the first sentence.

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
```diff
@@ -0,0 +1,42 @@
...
+`;
+
+const GetStarted = () => {
+  const transport = useTransport();
+  const [data] = usePromise(() => transport.proposals.proposalsBatch('Active'), undefined, []);
+
+  return <>
+    <p>If you&apos;re new to Joystream, go to our <a href='https://github.com/Joystream/helpdesk/'>Helpdesk</a> to learn what you can do, and how to get started. If you&apos;re not there already, join our <a href='https://t.me/joinchat/CNyeUxHD9H56m3e_44hXIA'>telegram group</a>, get some tokens, and join our community!</p>
+    <Subtitle>Membership</Subtitle>
+    <p>Most actions on the platform requires you to become a <a href='/#/members/edit'>Member</a>.</p>
```
blrhc, line 19:
```suggestion
    <p>Most actions on the platform require you to become a <a href='/#/members/edit'>Member</a>.</p>
```

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <Subtitle>Council</Subtitle>
+    <p>As &quot;The video platfrom DAO&quot;, Joystreams day to day operations are controlled by the <a href='/#/council/members'>Council</a> through a Proposal system.</p>
```
blrhc, line 21:
```suggestion
    <p>As &quot;The video platform DAO&quot;, the testnet's day to day operations are managed by the <a href='/#/council/members'>Council</a> through the Proposal system.</p>
```

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <Subtitle>Proposals</Subtitle>
+    <p>There are currently {data?.proposals.length || 'no'} <a href='/#/proposals'>Proposals</a> open for voting.</p>
```
blrhc, line 23: Small thing here but would be nice to have the numbers in words (e.g. one two three four five) rather than figures.

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </ul> : null
+    }
+    <Subtitle>Roles</Subtitle>
+    <p>In addition to being a Council Member, the other &quot;official&quot; roles are:</p>
+    <ul>
+      <li><a href='/#/staking'>Validator</a> (and Nominators)</li>
+      <li><a href='/#/working-groups'>Storage Provider</a></li>
+      <li><a href='/#/working-groups'>Content Curator</a></li>
+    </ul>
+    <Subtitle>Media</Subtitle>
```
blrhc, line 36: @bwhm Since the media consumption part of Pioneer is being removed (as far as I understand), is it worth including this section?

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
DzhideX: Since the data here is dynamic and not manually inserted by someone, we would need to write a function which turns numbers into words. After some research doesn't look like it would be even remotely simple. 

One saving grace here would be if there is already such a function in the codebase. Maybe @Lezek123 knows?

blrhc:
Just thought of something else too.
Might be nice to show five most recent forum posts in a **Forum** section?

And in the roles section, as well as linking to the working group/appropriate Pioneer page maybe have a button at the side of each link `GUIDE` with a link to the relevant helpdesk guide?

DzhideX:
I have one problem with our current approach of just piling on information in this one page. 

I implemented all of the "non-disputable" sections suggested by @blrhc and the way it is now one needs to scroll fairly long until they get to the Roles section which may lead to people not even being able to find it initially at all. One more thing, in my experience, the longer the wall of text is, the lower the chance a person/beginner would want to read it is.

I think we should look into making all of this data more presentable. I have a couple of suggestions that I've been thinking about:

- We could hide the "Get Started" part and just leave the title, with a small hovering text box next to it (that would maybe exist only for a couple of seconds?) which could say something along the lines of : "Click here if you're new to Joystream!". By clicking on the title, all of the information would then open up(animate into existence?). One problem with this approach is that it doesn't feel very Pioneer-y in the sense that there aren't many things like that from what I've seen.

- Move Get Started(or Roles) into it's own tab or somehow isolate the data not to feel as overwhelming..

Then again, this is just my own opinion and maybe it's just a premature/unnecessary optimization.

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
```diff
@@ -0,0 +1,42 @@
...
+    color: rgba(0,0,0, 0.65);
+    margin-bottom: 1rem;
+`;
+
+const GetStarted = () => {
+  const transport = useTransport();
+  const [data] = usePromise(() => transport.proposals.proposalsBatch('Active'), undefined, []);
+
+  return <>
+    <p>If you&apos;re new to Joystream, go to our <a href='https://github.com/Joystream/helpdesk/'>Helpdesk</a> to learn what you can do, and how to get started. If you&apos;re not there already, join our <a href='https://t.me/joinchat/CNyeUxHD9H56m3e_44hXIA'>telegram group</a>, get some tokens, and join our community!</p>
```
Lezek123, line 17: Since it's a `Get started` page, we should perhaps use `target="_blank"` in links here. I assume the user will probably want to explore multiple links without beeing moved from this page.

Also as I pointed out a few times: Please split overly long lines into multiple ones, it's making it way easier to review/update the code.

pioneer/packages/joy-tokenomics/src/Landing/GetStarted.tsx
Lezek123: Perhaps we can use some existing npm package to convert this to text (as @blrhc suggested). For example: https://www.npmjs.com/package/number-to-words

pioneer/packages/page-staking/src/Targets/index.tsx
```diff
@@ -177,9 +182,10 @@ function Targets ({ className = '', isInElection, ownStashes, targets: { calcWit
         totalStaked={totalStaked}
       />
       <Button.Group>
+        {/* !validators?.length || !ownNominators?.length */}
         <Button
           icon='check'
-          isDisabled={!validators?.length || !ownNominators?.length}
+          isDisabled={false}
```
Lezek123, line 188: What is the reason behind this change?

pioneer/packages/page-staking/src/Targets/index.tsx
```diff
@@ -59,6 +59,11 @@ function Targets ({ className = '', isInElection, ownStashes, targets: { calcWit
   const [withIdentity, setWithIdentity] = useState(false);
   const [{ sortBy, sortFromMax }, setSortBy] = useState<SortState>({ sortBy: 'rankOverall', sortFromMax: true });
 
+  console.log(selected);
```
Lezek123, line 62: Some leftover console logs to remove

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +454,346 @@ export default class TokenomicsTransport extends BaseTransport {
...
+    const payoutInterval = (await this.api.query.council.payoutInterval()) as Option<BlockNumber>;
+
+    if (payoutInterval.isSome) {
+      termReward = (amountPerPayout.toNumber() / payoutInterval.unwrap().toNumber()) * new_term_duration.toNumber();
+    }
+
+    if (stage.isOfType('Announcing')) {
+      const applicants = (await this.api.query.councilElection.applicants()) as Vec<AccountId>;
+
+      return [
```
Lezek123, line 645: Too much data related to UI here (stuff like `color`, `cellData`, `href` etc. doesn't belong here, this is something that should be handled by the components that make use of this data).
This needs to be refactored.

Perhaps while refacoring this code you can also consider how to split it to reusable methods to avoid repeating a lot of code related to fetching the same data etc. For example - if you have a method to get council memeber reward per block you can use it to calculate both total council reward per week, council member reward per term etc.

If you often need to convert `JOY` to `USD`, it may make sense to create a transport method just for this purpose and perhaps implement some caching for the price (which needs to be fetched from the status server). So for example, you can have a method like: `joyToUSD()`. Perhaps it would make even more sense to create a react context that will provide the current price to all the `joy-tokenomics` components (and potentially other apps in the future), so that you can always easily convert between USD and JOY inside any component.

pioneer/packages/joy-tokenomics/src/Landing/Roles/Role.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const Role: React.FC<{
+  title: string
+  roleData: [string, string][]
+  extraData?: LandingPageWorkingGroupExtra[] | LandingPageCouncilExtra[]
+}> = ({ title, roleData, extraData }) => {
+  return (
+    <StyledRoleCard>
+      <StyledRoleCardHeader textAlign='center' header={title} />
+      {extraData && (extraData as any[]).map((data: LandingPageWorkingGroupExtra | LandingPageCouncilExtra, index: number) => (
```
Lezek123, line 43: The `as any[]` part doesn't seem to be needed

pioneer/packages/joy-tokenomics/src/Landing/Roles/RoleCardExtra.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <CouncilExtraCell>
+        <p>TERM REWARD</p>
+        <h2>{termReward}</h2>
+      </CouncilExtraCell>
+      <CouncilExtraCell>
+        <p>TERM MAX BONUS</p>
+        <h2>{termMaxBonus}</h2>
+      </CouncilExtraCell>
+      <CouncilExtraCell>
+        <p dangerouslySetInnerHTML={{ __html: text }}/>
```
Lezek123, line 60: I would avoid that. Passing a HTML string for the component to render is not a good practice (unless it cannot be avoided). We can, for example, pass `jsx` here through `children` (or as `prop`)

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -44,34 +64,43 @@ export default class TokenomicsTransport extends BaseTransport {
...
+    return {
+      councilTermLength,
+      avgCouncilRewardPerWeek
+    };
   }
 
   async getCouncilData () {
     const { numberOfCouncilMembers, totalCouncilStake } = await this.councilSizeAndStake();
-    const totalCouncilRewardsInOneWeek = await this.councilRewardsPerWeek(numberOfCouncilMembers);
+    const { avgCouncilRewardPerWeek, councilTermLength } = await this.councilRewardsPerWeek(numberOfCouncilMembers);
```
Lezek123, line 97: Since this method no longer returns council rewards per week, perhpas the name should be changed?
Although I think it would be best to make two separate methods out of it.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +454,346 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      ).toNumber();
+
+      if (!accountBalanceData.freeBalance.isZero()) {
+        userFreeBalance = accountBalanceData.freeBalance.toNumber();
+      }
+    } else {
+      profitPerEra = (await this.optimalNomination()).toNumber();
+    }
+
+    return {
```
Lezek123, line 553: Avoid returning data formatted for the specific UI in the transports.
The role of the transport is to fetch, combine and perhaps parse the data into a format that's easier to handle, but the data itself should be reusable enough so that it can be used by multiple different components in multiple scenarios.

The way return data is formatted here doesn't seem very reusable. Things like headers, number formatting or what to display in case the data is not available - all of this should be left for the component that makes use of the data to decide.

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +454,346 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      ];
+    }
+
+    return undefined;
+  }
+
+  async openingData (group: 'storageWorkingGroup' | 'contentWorkingGroup'): Promise<LandingPageWorkingGroupExtra[]> {
+    let openingsToShow: Opening[];
+
+    if (group === 'storageWorkingGroup') {
```
Lezek123, line 753: For Babylon and in the future you should use `workingGroupsT.queryByGroup` for any queries that can be executed in multiple working group contexts (this was probably not yet available since you based your work on `master`, but just pointing for once you get to upgrade this code)

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -341,4 +454,346 @@ export default class TokenomicsTransport extends BaseTransport {
...
+      const contentOpenings = await this.entriesByIds<OpeningId, CuratorOpening>(
+        this.api.query.contentWorkingGroup.curatorOpeningById
+      );
+
+      openingsToShow = await this.api.query.hiring.openingById.multi<Opening>(
+        contentOpenings.map(([_, curatorOpening]) => curatorOpening.opening_id)
+      );
+    }
+
+    return openingsToShow
```
Lezek123, line 771: Just like before - this is mostly UI data. Perhaps you don't even need any methods other than the ones that already exist in `workingGroupsTransport`?

pioneer/packages/joy-utils/src/transport/tokenomics.ts
```diff
@@ -262,38 +330,83 @@ export default class TokenomicsTransport extends BaseTransport {
...
-    const { numberOfStorageProviders, storageProviderLeadNumber, totalStorageProviderStake, totalStorageProviderLeadStake, storageProviderLeadRewardsPerWeek, storageProviderRewardsPerWeek } = await this.getStorageProviderData();
-    const { numberOfContentCurators, contentCuratorLeadNumber, totalContentCuratorStake, contentCuratorRewardsPerWeek } = await this.getContentCuratorData();
-    const { numberOfValidators, numberOfNominators, totalValidatorStake, validatorRewardsPerWeek, totalIssuance } = await this.getValidatorData();
-    const currentlyStakedTokens = totalCouncilStake + totalStorageProviderStake + totalStorageProviderLeadStake + totalContentCuratorStake + totalValidatorStake;
-    const totalWeeklySpending = totalCouncilRewardsInOneWeek + storageProviderRewardsPerWeek + storageProviderLeadRewardsPerWeek + contentCuratorRewardsPerWeek + validatorRewardsPerWeek;
-    const totalNumberOfActors = numberOfCouncilMembers + numberOfStorageProviders + storageProviderLeadNumber + numberOfContentCurators + contentCuratorLeadNumber + numberOfValidators;
+    const {
```
Lezek123, line 370: Good job splitting the lines here!

Lezek123 (changes requested):
The main point is to refactor the code in such a way that everything UI-specific is inside react compontents instead of transport methods. I explained in more detail in the comments what's the role of the transports and what doesn't belong there.

I also added a few other small suggestions, but I didn't do very thorough review of the queries yet, since the transport code needs quite substantial refactoring and I assumed it will be easier to review that part later.

Issue #1575, opened 2020-10-22
Make channel and video views use infinite scroller


kdembler:
Currently, both video playback view (more videos section) and channel view use a regular video grid for displaying the channel's videos. Furthermore, they don't use paginated queries but rather just request all videos from a specified channel via the channel's `videos` field. This works for now, but once channels start having more videos, pagination will be a must.

As part of this task, support for passing channel ID to the InfiniteVideoGrid should be added and that component should be used for the aforementioned views (unless we don't want infinite grids there, then we only need to use paginated queries)

PR #1576, opened 2020-10-22
use real channel videos in video playback

### LINKED ISSUE #1525: Static more videos in video playback view
In the video playback view, under the player, there's a grid of videos called "More from <channel name>". Some time ago we statically mocked this data inside the component and haven't changed that since. We should do a proper data fetching/mocking for it and also figure out what should be displayed there - probably videos of that specific channel.

Screenshot 2020-10-12 at 1 50 48 PM

kdembler:
Resolves #1525 

This is a fix for now, the real solution is described here - #1575


fcescob (approved):
LGTM!

bedeho (approved):

PR #1578, opened 2020-10-22
CLI - fix bugs arising from entities without a schema

Lezek123:
While doing some tests with `createEntity` and `addSchemaSupportToEntity` I noticed that entities without a schema can cause some issues in the CLI:
- `content-directory:entities` command would sometimes fail because it was running under the wrong assumption that all entities (presented in a table) will have the same properties in `entity.values` (while entities with no schema support have no `values` at all)
- Commands like `updateChannel`/`updateVideo` etc. didn't check if the entity has any schema support, which resulted in some unexpected behaviour

This PR tries to address those issues


mnaamani:
LGTM, reminds me of similar case we had to deal with in pioneer.

Issue #1534, opened 2020-10-14
Add weights and benchmarking to working group


bedeho:
# Background

The working group organizational concept in the Joystream runtime, and there are many instance of it in the runtime. An explanation of its function and organization can be found here

https://joystream.gitbook.io/joystream-handbook/governance/working-groups

The most recent implementation can be found in the `working_team_pallet` branch of this repository, in the directory `runtime-modules/working-team`. Be aware that this directory is just a temporary location, the sister folder `working_group` holds an old implementation which will be removed soon, at which point the `working-team` directory will be renamed to this.

All routines this module currently have degenerate weighting, this must be changed, and the documentation in each extrinsic must be updated to reflect the fee and complexity associated with each module.

# Proposal

Introduce weights and benchmarking of the working group extrinsics.

## Practical (for beginner)

1. Review the working group handbook documentation.
2. Review Substrate introductory materials and tutorials in order to understand the organization of a runtime module.
3. Review material on weights and benchmarking of Substrate modules.
4. Review the implementation.
5. Implement weights and benchmarking in a new PR targeting the `working_team_pallet` branch, including updating tests.
6. Update network integration tests to work with new fees.
7. Update CLI to work with new fees.

PR #1587, opened 2020-10-24
Benchmarking working_team

conectado:
Pull Request to track progress in #1534 
---

We need to benchmark `working_team` to implement the weights for each extrinsic(and `on_initialize`)

In order to test the benchmark, I added a dummy implementation of `working_team::Trait` for the `Runtime` we are using.

**Note**: It's useless to calculate the weights in the reference hardware until the acual impl for the `Runtime` is done. (Benchmarking only make sense in the context of a given `Runtime`)

To complete the PR I need to do the following:

- [x] Add benchmarking to the Runtime(when the feature is enabled)
- [x] Add dummy `working_team::Trait` impl for `Runtime`
- [x] Add benchmark function for each branch of each extrinsic and `on_initialize`
  - [x] `leave_role`
  - [x] `set_budget`
  - [x] `add_opening`
  - [x] `update_reward_account`
  - [x] `set_status_text`
  - [x] `update_reward_amount`
  - [x] `spend_from_budget`
  - [x] `decrease_stake`
  - [x] `increase_stake`
  - [x] `terminate_role`
  - [x] `slash_stake`
  - [x] `withdraw_application`
  - [x] `cancel_opening`
  - [x] `update_role_account`
  - [x] `fill_opening`
  - [x] `apply_on_opening`
  - [x]  `on_initialize`
- [x] Add code to run verifications with the tests
- [x] Implement the verify function for each benchmark function
- [x] Whitelist appropiate keys
- [ ] Write the weight function for each extrinsic (**Note**:  in substrate current version this is done automatically by the CLI) 
- [ ] Run the code in [reference hardware](https://substrate.dev/docs/en/knowledgebase/learn-substrate/weight) (**Blocked until `working_team::Trait` is implemented**)
- [ ] Add weights in the extrinsics(and `on_initialize`)

### DIFF
--- a/runtime-modules/working-team/Cargo.toml
+++ b/runtime-modules/working-team/Cargo.toml
@@ -37,5 +37,4 @@ std = [
 	'common/std',
 	'membership/std',
 	'balances/std',
-  'frame-benchmarking/std',
 ]
--- a/runtime-modules/working-team/Cargo.toml
+++ b/runtime-modules/working-team/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-working-team"
-version = "1.0.0"
+version = "1.0.1"
 authors = ['Joystream contributors']
 edition = '2018'
 
@@ -15,7 +15,6 @@ sp-std = { package = 'sp-std', default-features = false, git = 'https://github.c
 common = { package = 'pallet-common', default-features = false, path = '../common'}
 membership = { package = 'pallet-membership', default-features = false, path = '../membership'}
 balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = '00768a1f21a579c478fe5d4f51e1fa71f7db9fd4'}
-
 frame-benchmarking = { package = 'frame-benchmarking', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = '00768a1f21a579c478fe5d4f51e1fa71f7db9fd4', optional = true}
 
 [dev-dependencies]
--- a/runtime-modules/working-team/Cargo.toml
+++ b/runtime-modules/working-team/Cargo.toml
@@ -16,6 +16,7 @@ common = { package = 'pallet-common', default-features = false, path = '../commo
 membership = { package = 'pallet-membership', default-features = false, path = '../membership'}
 balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = '00768a1f21a579c478fe5d4f51e1fa71f7db9fd4'}
 
+frame-benchmarking = { package = 'frame-benchmarking', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = '00768a1f21a579c478fe5d4f51e1fa71f7db9fd4', optional = true}
 
 [dev-dependencies]
 sp-io = { package = 'sp-io', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = '00768a1f21a579c478fe5d4f51e1fa71f7db9fd4'}
@@ -24,6 +25,7 @@ pallet-timestamp = { package = 'pallet-timestamp', default-features = false, git
 
 [features]
 default = ['std']
+runtime-benchmarks = ["frame-benchmarking"]
 std = [
 	'serde',
 	'codec/std',
@@ -35,4 +37,5 @@ std = [
 	'common/std',
 	'membership/std',
 	'balances/std',
+  'frame-benchmarking/std',
 ]
--- a/runtime-modules/working-team/src/benchmarking.rs
+++ b/runtime-modules/working-team/src/benchmarking.rs
@@ -584,7 +584,7 @@ benchmarks_instance! {
         );
     }
 
-    fill_opening_worker { // We can actually fill an opening with 0 applications?
+    fill_opening_worker {
         let i in 1 .. T::MaxWorkerNumberLimit::get();
         let (lead_id, lead_worker_id) = insert_a_worker::<T, I>(
             StakingRole::WithoutStakes,
--- a/runtime-modules/working-team/src/benchmarking.rs
+++ b/runtime-modules/working-team/src/benchmarking.rs
@@ -54,7 +54,13 @@ fn add_opening_helper<T: Trait<I>, I: Instance>(
     T::OpeningId::from(id.try_into().unwrap())
 }
 
-fn apply_on_opening_helper<T: Trait<I>, I: Instance>(id: u32, staking_role: &StakingRole, applicant_id: &T::AccountId, member_id: &T::MemberId, opening_id: &T::OpeningId) -> T::ApplicationId {
+fn apply_on_opening_helper<T: Trait<I>, I: Instance>(
+    id: u32,
+    staking_role: &StakingRole,
+    applicant_id: &T::AccountId,
+    member_id: &T::MemberId,
+    opening_id: &T::OpeningId,
+) -> T::ApplicationId {
     let stake_parameters = match staking_role {
         StakingRole::WithStakes => Some(StakeParameters {
             stake: BalanceOfCurrency::<T>::max_value(),
@@ -79,6 +85,30 @@ fn apply_on_opening_helper<T: Trait<I>, I: Instance>(id: u32, staking_role: &Sta
     T::ApplicationId::from(id.try_into().unwrap())
 }
 
+fn add_opening_and_n_apply<T: Trait<I>, I: Instance>(
+    ids: &Vec<u32>,
+    add_opening_origin: &T::Origin,
+    staking_role: &StakingRole,
+    job_opening_type: &JobOpeningType,
+) -> (T::OpeningId, BTreeSet<T::ApplicationId>, Vec<T::AccountId>) {
+      let opening_id = add_opening_helper::<T, I>(1, add_opening_origin, &staking_role, job_opening_type);
+
+      let mut successful_application_ids = BTreeSet::new();
+
+      let mut account_ids = Vec::new();
+      for id in ids.iter()  {
+        let (applicant_account_id, applicant_member_id) = member_funded_account::<T>("member", *id);
+        let application_id = apply_on_opening_helper::<T, I>(*id, &staking_role, &applicant_account_id,
+          &applicant_member_id, &opening_id);
+
+        successful_application_ids.insert(application_id);
+        account_ids.push(applicant_account_id);
+
+      }
+
+    (opening_id, successful_application_ids, account_ids)
+}
+
 fn add_and_apply_opening<T: Trait<I>, I: Instance>(
     id: u32,
     add_opening_origin: T::Origin,
@@ -87,9 +117,11 @@ fn add_and_apply_opening<T: Trait<I>, I: Instance>(
     member_id: T::MemberId,
     job_opening_type: JobOpeningType,
 ) -> (T::OpeningId, T::ApplicationId) {
-    let opening_id = add_opening_helper::<T, I>(id, &add_opening_origin, &staking_role, &job_opening_type);
+    let opening_id =
+        add_opening_helper::<T, I>(id, &add_opening_origin, &staking_role, &job_opening_type);
 
-    let application_id = apply_on_opening_helper::<T, I>(id, &staking_role, &applicant_id, &member_id, &opening_id);
+    let application_id =
+        apply_on_opening_helper::<T, I>(id, &staking_role, &applicant_id, &member_id, &opening_id);
 
     (opening_id, application_id)
 }
@@ -138,6 +170,15 @@ fn member_funded_account<T: membership::Trait>(
     (account_id, T::MemberId::from(id.try_into().unwrap()))
 }
 
+fn force_missed_reward<T: Trait<I>, I: Instance> () {
+    let curr_block_number =
+        System::<T>::block_number().saturating_add(T::RewardPeriod::get().into());
+    System::<T>::set_block_number(curr_block_number);
+    WorkingTeam::<T, I>::set_budget(RawOrigin::Root.into(), Zero::zero()).unwrap();
+    WorkingTeam::<T, I>::on_initialize(curr_block_number);
+}
+
+
 fn insert_a_worker<T: Trait<I>, I: Instance>(
     staking_role: StakingRole,
     job_opening_type: JobOpeningType,
@@ -172,14 +213,9 @@ where
     )
     .unwrap();
 
-    // Force a missed_reward for the created worker
     // Every worst case either include or doesn't mind having a non-zero
     // remaining reward
-    let curr_block_number =
-        System::<T>::block_number().saturating_add(T::RewardPeriod::get().into());
-    System::<T>::set_block_number(curr_block_number);
-    WorkingTeam::<T, I>::set_budget(RawOrigin::Root.into(), Zero::zero()).unwrap();
-    WorkingTeam::<T, I>::on_initialize(curr_block_number);
+    force_missed_reward::<T, I>();
 
     (caller_id, TeamWorkerId::<T>::from(id.try_into().unwrap()))
 }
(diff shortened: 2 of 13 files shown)


bedeho:
What do you think is the best way to review this?

Should this all be reviewed when this is all done, in one gigantic review? or something else?

conectado:
> What do you think is the best way to review this?
> 
> Should this all be reviewed when this is all done, in one gigantic review? or something else?

I thought it over and I think the best way to approach and I think the best way is to do a review each time a certain "milestone" is hit. That way we can catch problems early and don't have to re-write a lot of code, but not so often that you review code that will be discarded or changed later.

I was thinking of these milestones.
When each one of these are completed:

- [x] Add benchmark function for each branch of each extrinsic and `on_initialize`
- [ ] Implement the verify function for each benchmark function
- [ ] Whitelist appropiate keys
- [ ] Everything

Does this make sense?

conectado:
> > What do you think is the best way to review this?
> > Should this all be reviewed when this is all done, in one gigantic review? or something else?
> 
> I thought it over and I think the best way to approach and I think the best way is to do a review each time a certain "milestone" is hit. That way we can catch problems early and don't have to re-write a lot of code, but not so often that you review code that will be discarded or changed later.
> 
> I was thinking of these milestones.
> When each one of these are completed:
> 
>     * [x]  Add benchmark function for each branch of each extrinsic and `on_initialize`
> 
>     * [ ]  Implement the verify function for each benchmark function
> 
>     * [ ]  Whitelist appropiate keys
> 
>     * [ ]  Everything
> 
> 
> Does this make sense?

The extrinsic/`on_initialize` is ready to be reviewed, a couple of pointers:

* I added a dummy implementation for the `working_team::Trait` that was needed to test the benchmarks
  * This meant that I also needed to implement a dummy implementation of `working_team::StakingHandler`
    * `current_stake` returns a value different than 0, this is needed for `leave_role_later` to take the correct branch
  * I also needed to expose publicly `types::BalanceOfCurrency` to implement this dummy quickly
* **It's very important that we check that every extrinsic benchmarked takes the worst possible scenario or when it's not obvious every not-obviously worst-case is tested**
  * I always erred in the side of caution trying to test every branch when I wasn't sure
* **It's also very important to see that we identify every parameter which makes the function takes more time** (such as the length of a string that we are hashing or the number of workers over which we are iterating)
  * For `on_initialize` it might appear that there are 2 parameters: the number of leaving workers and the number of workers to reward. However, when there are two parameters when the benchmarks do its linear regression it fixes the other parameter in the biggest case, in this case `T::MaxWorkerNumberLimit::get()`, this means that there is no space for leaving workers because all the workers would be in the rewarded set or vice versa. Still, we can take this as 2 branches with only 1 parameter, this being `leaving_workers` + `rewarding_workers`(with the two branches being all are leaving or all rewarded). This works because one of the branches has to be worse than the other and the more workers taking the "bad" branch the worse it will be.
  * Some of the parameters seem to be arbitrarily going from 1 to 10, these are the constant cases but a parameter is needed for the benchmarking to work(afaik in the latest substrate version this is no longer needed)
  * Some parameters go into an arbitrarily big number, these are some input parameters that were unbounded, in every example I've found every parameter was bounded and used that as the upper limit. As I understand the weight of the parameter would prevent this to be a problem, but I'm not sure, we might have to look more deeply into that.

conectado:
For whitelisting there are two cases:
* A key is read/written in every block 
* A key is transient

In both cases the since the key is cached and we don't need to count it on our read/write count

With the version we are using for Substrate we can't add a pallet scoped key because the function to do this(`add_to_whitelist`) was added in [this commit](https://github.com/paritytech/substrate/commit/368903f7aa9ef652bf8157d476dc13cb36e3affc#diff-0531fde3df80cc3c15bc09edd247ac8cbd6106ec9adfd406a49a07528ffea663R126) which was merged after the version we are using. (This shouldn't matter since we care more about using the worst case scenario rather than tightening the weight to the actual computation time)

Still, in this case I think we shouldn't whitelist any key regarding this pallet, since the only candidates are the keys in `WorkerById` which are accessed through the function `Self::get_workers_with_finished_unstaking_period`. However then we can add and access another key in `WorkerById` in the same block, so we have no way of knowing a-priori if the key in the current extrinsic was previously accessed.

For these reasons, I'm marking the checkpoint **Whitelist appropiate keys** as done.

conectado:
Note: To implement the benchmark tests we need to use `DefaultInstance` as explained in #1652

runtime-modules/working-team/Cargo.toml
```diff
@@ -35,4 +37,5 @@ std = [
 	'common/std',
 	'membership/std',
 	'balances/std',
```
shamil-gadelshin: Do we need 'frame-benchmarking' in the std?

runtime/src/lib.rs
```diff
@@ -65,6 +65,95 @@ pub use versioned_store;
 pub use versioned_store_permissions;
 pub use working_group;
```
shamil-gadelshin: We probably need to remove this after the benchmarking is done and prior to merging.

runtime-modules/working-team/src/lib.rs
```diff
@@ -47,7 +48,12 @@ use sp_std::vec::Vec;
 use system::{ensure_root, ensure_signed};
 
 pub use errors::Error;
-use types::{ApplicationInfo, BalanceOfCurrency, MemberId, TeamWorker, TeamWorkerId, WorkerInfo};
+/*
```
shamil-gadelshin: We probably need to remove this after the benchmarking is done and prior to merging.

runtime/src/lib.rs
```diff
@@ -618,5 +707,9 @@ construct_runtime!(
         // --- Working groups
         // reserved for the future use: ForumWorkingGroup: working_group::<Instance1>::{Module, Call, Storage, Event<T>},
         StorageWorkingGroup: working_group::<Instance2>::{Module, Call, Storage, Config<T>, Event<T>},
+        /*
```
shamil-gadelshin: We probably need to remove this after the benchmarking is done and prior to merging.

runtime/src/runtime_api.rs
```diff
@@ -8,6 +8,8 @@ use sp_core::crypto::KeyTypeId;
 use sp_core::OpaqueMetadata;
 use sp_runtime::traits::{BlakeTwo256, Block as BlockT, NumberFor};
 use sp_runtime::{generic, ApplyExtrinsicResult};
+#[cfg(feature = "runtime-benchmarks")]
```
shamil-gadelshin: We could move this inside the `impl frame_benchmarking::Benchmark<Block> for Runtime {` similar to other `use` statements.

runtime/src/runtime_api.rs
```diff
@@ -201,4 +203,53 @@ impl_runtime_apis! {
...
+             */
+            use frame_benchmarking::{Benchmarking, BenchmarkBatch, add_benchmark};
+            use frame_system_benchmarking::Module as SystemBench;
+            impl frame_system_benchmarking::Trait for Runtime {}
+
+            use crate::WorkingTeam;
+
+            let whitelist: Vec<Vec<u8>> = vec![
+                // Block Number
+                // frame_system::Number::<Runtime>::hashed_key().to_vec(),
```
shamil-gadelshin: Remove the commented code.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+            description: vec![],
+            stake_parameters,
+        },
+    )
+    .unwrap();
+
+    T::ApplicationId::from(id.try_into().unwrap())
+}
+
+fn add_opening_and_n_apply<T: Trait<I>, I: Instance>(
```
shamil-gadelshin: What does `n` correspond to? Please rename it.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+          System::<T>::block_number().saturating_add(T::RewardPeriod::get().into());
+      System::<T>::set_block_number(curr_block_number);
+
+      // Sets budget so that we can pay it
+      WorkingTeam::<T, _>::set_budget(RawOrigin::Root.into(), BalanceOfCurrency::<T>::max_value()).unwrap();
+
+    }: { WorkingTeam::<T, _>::on_initialize(curr_block_number) }
+
+    apply_on_opening {
+      let i in 1 .. 50000;
```
shamil-gadelshin: We should make the max description length consistent with the UI restrictions or limit this value in the runtime.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+
+      let (lead_account_id, lead_member_id) = member_funded_account::<T>("lead", 0);
+      let (opening_id, application_id) = add_and_apply_opening::<T, I>(0, &RawOrigin::Root.into(), &StakingRole::WithoutStakes, &lead_account_id,
+        &lead_member_id, &JobOpeningType::Leader);
+
+      let mut successful_application_ids: BTreeSet<T::ApplicationId> = BTreeSet::new();
+      successful_application_ids.insert(application_id);
+    }: fill_opening(RawOrigin::Root, opening_id, successful_application_ids)
+
+    fill_opening_worker { // We can actually fill an opening with 0 applications?
```
shamil-gadelshin: > We can actually fill an opening with 0 applications?

Yes, this is a bug. I will fix it myself during the pallet integration.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+            stake: BalanceOfCurrency::<T>::max_value(),
+            staking_account_id: lead_account_id.clone(),
+          }
+        ),
+      };
+
+    }: _ (RawOrigin::Signed(lead_account_id.clone()), apply_on_opening_params)
+
+    fill_opening_lead {
+      let i in 0 .. 10;
```
shamil-gadelshin: Do we need "i" variable here?

shamil-gadelshin: I saw your comment `// TODO: test not running if we don't set a range of values`. If this is really the case, please, add the explicit comment near each usage or a single one in the module description.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+    // to leave this branch and use tha max between these 2
+    leave_role_later {
+        let i in 0 .. 10;
+
+        // Workers with stake can't leave immediatly
+        let (caller_id, caller_worker_id) = insert_a_worker::<T, I>(StakingRole::WithoutStakes, JobOpeningType::Leader, 0, None);
+    }: leave_role(RawOrigin::Signed(caller_id), caller_worker_id)
+    verify { assert_eq!(true, true); }
+}
+
```
shamil-gadelshin, line 957: Yes, please add tests for each benchmark functions.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+    // Regular worker is the worst case scenario since the checks
+    // require access to the storage whilist that's not the case with a lead opening
+    slash_stake {
+      let i in 0 .. 10;
+
+      let (lead_id, lead_worker_id) = insert_a_worker::<T, I>(StakingRole::WithoutStakes, JobOpeningType::Leader, 0, None);
+      let (caller_id, worker_id) = insert_a_worker::<T, I>(StakingRole::WithStakes, JobOpeningType::Regular, 1, Some(lead_id.clone()));
+      let penalty = Penalty {
+        slashing_text: vec![],
```
shamil-gadelshin: Should we place a major vector here?

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    terminate_role_lead {
+      let i in 0 .. 10;
+
+      let (_, lead_worker_id) = insert_a_worker::<T, I>(StakingRole::WithStakes, JobOpeningType::Leader, 0, None);
+      let current_budget = BalanceOfCurrency::<T>::max_value();
+      // To be able to pay unpaid reward
+      WorkingTeam::<T, _>::set_budget(RawOrigin::Root.into(), current_budget).unwrap();
+      let penalty = Penalty {
+        slashing_text: vec![],
```
shamil-gadelshin: Should we place a major vector here?

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,599 @@
...
+      let i in 0 .. 10;
+
+      let new_budget = BalanceOfCurrency::<T>::max_value();
+
+    }: _(RawOrigin::Root, new_budget)
+
+    // Regular opening is the worst case scenario since the checks
+    // require access to the storage whilist that's not the case with a lead opening
+    add_opening{
+      let i in 0 .. 50000; // TODO: We should have a bounded value for description
```
shamil-gadelshin: @bedeho Despite the fact we don't use the text fields in the runtime because of the query node - we need to bound the length of these values in order to properly benchmark such processes like text hashing.

shamil-gadelshin (changes requested):
- Please, refactor the code in order to be closer to rust recommended 99 symbols per line.
- Did you consider 'weight-refunds' for any function?

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,921 @@
...
+        &JobOpeningType::Regular
+      );
+    }: fill_opening(RawOrigin::Signed(lead_id.clone()), opening_id, successful_application_ids.clone())
+    verify {
+      assert!(!OpeningById::<T, I>::contains_key(opening_id), "Opening still not filled");
+
+      for i in 1..successful_application_ids.len() {
+        assert!(WorkerById::<T, I>::contains_key(TeamWorkerId::<T>::from(i.try_into().unwrap())), "Not all workers
+        added");
+      }
```
shamil-gadelshin: It seems that you missed check for the `OpeningFilled` event.

shamil-gadelshin: Also, a successful outcome contains a new worker in the storage.

runtime-modules/working-team/src/benchmarking.rs
```diff
@@ -0,0 +1,921 @@
...
+    }: terminate_role(RawOrigin::Root, lead_worker_id, Some(penalty))
+    verify {
+      assert!(!WorkerById::<T, I>::contains_key(lead_worker_id), "Worker not terminated");
+      assert_last_event::<T, I>(RawEvent::TerminatedLeader(lead_worker_id).into());
+    }
+
+    // Regular worker is the worst case scenario since the checks
+    // require access to the storage whilist that's not the case with a lead opening
+    increase_stake {
+      let i in 0 .. 10;
```
shamil-gadelshin: For stake related 'verify' functions, you could check the actual change of the stake.

runtime-modules/working-team/src/benchmarking.rs
shamil-gadelshin: Nevermind, the latest commit includes the tests.

shamil-gadelshin:
Please, introduce the '4 spaces' indentation instead of the '2 spaces'.

shamil-gadelshin:
Also, please avoid double line spacing like in the `add_opening` or `on_initialize_leaving` and similar places.

runtime-modules/working-team/Cargo.toml
conectado: I think we don't need it. It is a leftover from earlier tests.

Will erase it.

runtime-modules/working-team/src/benchmarking.rs
conectado: I believe just like in `add_opening` we need to restrict this value in the runtime. (Will add a similar comment)

runtime-modules/working-team/src/benchmarking.rs
conectado: Sorry, I don't know what's a major vector, could you point me in a direction of a doc on the topic?

runtime-modules/working-team/src/benchmarking.rs
conectado: > Also, a successful outcome contains a new worker in the storage.

Aren't we checking that the outcome contains a new worker in the storage by checking that:
```rs
WorkerById::<T, I>::contains_key(TeamWorkerId::<T>::from(i.try_into().unwrap()))
```

is true?

runtime-modules/working-team/src/benchmarking.rs
conectado: I think a problem with that it's that it depends on the specific implementation of `StakingHandler` in the runtime or in the verification test in the `Test`.

In fact, I think we can't check for `current_stake` since that (almost)always returns 100  in the current mock implementation. I'm not sure but maybe we can for the locked amount?

runtime-modules/working-team/src/benchmarking.rs
conectado: This is needed in Substrate 2.0.0-RC-4 in the latest version(2.0.0) it seems that it is no longer needed. Will change this to `let i in 0 ..  1` and a comment to erase it when we update the version.

runtime-modules/working-team/src/benchmarking.rs
bedeho: I think there is a way to do this without needing to bound it, I believe the extrinsic caller needs to commit to a length. Please elaborate @conectado.

runtime-modules/working-team/src/benchmarking.rs
bedeho: What about this

https://github.com/Joystream/joystream/pull/1587#discussion_r520490279

bedeho (commented):
Very impressive. I don't have the ability to look through this in any detail, but before any non-trivial changes are implemented in the working group module to accommodate the benchmarking, lets schedule a shared call going through the major issues, as I assume the initial set of issues will be reoccurring across the runtime.

runtime-modules/working-team/src/benchmarking.rs
conectado: That way of doing things works for weights that depend on an stored item.

Say, for example, an extrinsic that depends on the number of stored workers, then instead of assuming worst-case scenario(that there are max number of workers stored) you can ask the caller to pass the number of workers as a parameter and then ensure that it is correct before executing the code. That way you can use that parameter to calculate the weight instead of assuming worst-case and refunding later.

Here, it is not the case, because we depend directly on the parameter passed, the problem is that this parameter is unbounded and for what I've seen in FRAME is that they actually bound with a "max" value all these linear parameters (like [here](https://github.com/paritytech/substrate/blob/900d034fc1ba0ccb30dd54b2dd7498ad5caa7fb5/frame/treasury/src/lib.rs#L633)).

However, I don't fully understand why this is needed. What I understand is that since we are doing linear interpolation, if the extrinsic (meaning, in this case, the hash function) is really linear we can use an arbitrarily large value as an upper-bound only for the linear interpolation and then depend on the Weight of the extrinsic call to bound the function running-time. But I've not seen it done like that anywhere.

conectado:
>     * Did you consider 'weight-refunds' for any function?

`leave_role` can have weight returns depending on the branch(leaving immediatly or later) taken and we could implement  weight refunds for `terminate_role` in a similar fashion, depending if the terminated role is a lead or a worker, the same goes for `fill_opening`.

What do you think?

shamil-gadelshin:
Please, remove all TODO code from the PR before merging. We are going to create an integration branch and this mock code in the runtime is redundant. 

Regarding the 'weight-refunds' - I was thinking about the major difference between code branches weights and which is hidden in the subtle 'parameters and storage state' combination. Since there are no such obvious places - we should wait for production weight costs of extrinsic and consider 'weight-refunds' as an advanced tweaking technique.

As to leave_role, termination_role and similar: you had already split their benchmarks into several functions and I think we could use just max of them because the weight difference which brings the 'leader case' is not very significant.

runtime-modules/working-team/src/benchmarking.rs
shamil-gadelshin: Nevermind, probably I saw not the latest commit version.

runtime-modules/working-team/src/benchmarking.rs
shamil-gadelshin: I agree. Let's skip it then.

runtime-modules/working-team/src/benchmarking.rs
shamil-gadelshin: Ok. Let's skip the upper limits and rely on block length fees and linear dependency on the text length.

shamil-gadelshin:
Also, since we are not going to limit the text fields, please, remove the comments in the benchmarks about it.

shamil-gadelshin (changes requested):
See the comments above.

shamil-gadelshin (approved):
LGTM

PR #1588, opened 2020-10-25
Council naming improvements

ondratra:
This PR renames `...stake` to `...vote_stake` to better distinguish referendum and council actions.  The naming change is in accord with https://github.com/Joystream/handbook/pull/14

~Plus, an unnecessary check for the situation that couldn't happen regarding the candidacy stake was deleted.~

### DIFF
--- a/runtime-modules/referendum/src/lib.rs
+++ b/runtime-modules/referendum/src/lib.rs
@@ -23,7 +23,7 @@
 //!
 //! - [vote](./struct.Module.html#method.vote)
 //! - [reveal_vote](./struct.Module.html#method.reveal_vote)
-//! - [release_stake](./struct.Module.html#method.release_stake)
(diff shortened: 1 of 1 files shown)


bedeho (commented):
Reading this I am actually failing to see why release candidacy stake would work to begin with?

What is making this only unlock the correct lock, and nothing else? When releasing stake, we want to allow removal of a council lock which either is for a candidate that failed in the last election, or it is for an election that is not the immediately prior one. What am I missing?

Also `ensure_previous_stake_reusable` is being reused in a context where that name, at least, does not really make sense as a check.

ondratra:
The scope of this PR was reduced to renaming `...stake` to `...vote_stake` in referendum's code.

bedeho (approved):

PR #1607, opened 2020-10-26
remove_entity: Update reference counter of referenced entities after removal performed

iorveth:
Closes: https://github.com/Joystream/joystream/issues/1600

Issue #1589, opened 2020-10-26
CLI - remaining content directory functionality for Babylon


Lezek123:
Remaining functionality to implement in the CLI before Babylon release:

**`media`**
- [x] Allow removing video (will involve removing multiple entities like `VideoMedia`, `VideoMediaLocation` etc)
- [x] Allow removing a channel
- [x] Create/update video - allow providing some missing data: `userDefinedLicense` (path to file?), `publishedBeforeJoystream`, `skippableIntroDuration`

**`content-directory`**
- [x] Add a command to run content directory initialization script as Curator Lead **(!)**
- [x] Remove any entity by id (it should be no problem if we already have a higher-level functionality to remove channel/video)

Lezek123:
Closed with https://github.com/Joystream/joystream/pull/1614

PR #1614, opened 2020-10-27
CLI: Remaining content-directory/media functionality for Babylon

Lezek123:
Implements https://github.com/Joystream/joystream/issues/1589

- Adds content directory initialization script as CLI command (`content-directory:initialize`)
- Adds `media:removeVideo` command which allows the user to delete `Video` along with associated `VideoMedia` and `License` entities (requires a fix from https://github.com/Joystream/joystream/pull/1607 in order to work)
- Adds `media:removeChannel` command
- Adds `content-directory:removeEntity` command allowing the user to remove any entity using choosen actor context (`Member` / `Curator` / `Lead`)
- Allows specifying either `KnownLicense` or `UserDefinedLicense` when uploading a video
- Introduces a separate command to update video license (`media:updateVideoLicense`), due to complexity of this operation (this also allows the curator to use `media:updateVideo --asCurator` without maintainer access to `License` class).
- Adds prompts for 2 other previously missing fields during video upload/update - `skippableIntroDuration` and `publishedBeforeJoystream`
- Fixes a bug in CLI's `Api.entityById()` which caused removed entities to not be skipped (invalid `exists` check)


cli/src/commands/content-directory/initialize.ts
```diff
@@ -0,0 +1,50 @@
...
+  async run() {
+    const account = await this.getRequiredSelectedAccount()
+    await this.requireLead()
+    await this.requestAccountDecoding(account)
+
+    const classInputs = getInputs<CreateClass>('classes').map(({ data }) => data)
+    const schemaInputs = getInputs<AddClassSchema>('schemas').map(({ data }) => data)
+    const entityBatchInputs = getInputs<EntityBatch>('entityBatches').map(({ data }) => data)
+
+    const currentClasses = await this.getApi().availableClasses()
```
mnaamani, line 20: Glad you improved the "emptiness" check here. In the cd-schemas initialize script it assumed the first id's will be equal to 1 which will not be the case when we perform a runtime upgrade, in which case the first id will in fact be zero.

https://github.com/Joystream/joystream/blob/130644e989355b68734a19bf9a1c8ae8a08016d3/content-directory-schemas/scripts/initializeContentDir.ts#L28

mnaamani (changes requested):
There is an issue with `media:updateVideo` when prompting for `publishedBeforeJoystream` 

```
Do you want to update first publication date (publishedBeforeJoystream)? No
Provided value: false
ERROR: This value should be integer
Try providing the input for publishedBeforeJoystream again...
```

`skippableIntroDuration` as far as I can see is not optional, can we give it a default value of 0 when being prompted?

Tested removing entities worked, I also tested removing a channel that contained videos and it correctly failed. Perhaps the command can check if there are any videos in the channel before submitting the extrinsic to avoid a wasted transaction?

Tested updating license and that worked as well when selecting a known license, but when trying to specify a path to a custom license i had to struggle a bit to realize it was best to use a absolute path, and not use a `~` shortcut for example, maybe we can add a comment about it, or support a relative paths and "`~`" substitution for the home directory?


```
MokoBook:joystream mokhtar$ yarn joystream-cli media:updateVideoLicense
yarn run v1.22.4
$ /Users/mokhtar/joystream/joystream/node_modules/.bin/joystream-cli media:updateVideoLicense
 Select a video to update ID:104
Current license: { knownLicense: 21, userDefinedLicense: false }
 Choose license type Custom (user-defined)
 Path to license file: ./LICENSE
 ›   Warning: The license content cannot be more than 4096 characters long
 Path to license file: ~/joystream/joystream/LICENSE
 ›   Warning: The file was not found or couldn't be accessed, try again...
 Path to license file: /Users/mokhtar/joystream/joystream/LICENSE
 ›   Warning: The license content cannot be more than 4096 characters long
 Path to license file: 
```

cli/src/commands/content-directory/initialize.ts
mnaamani: Perhaps we can fix it there as well?

Lezek123:
**New changes:**

cd-schemas:
- Added a way to set optional values back to `[not set]` by providing entity input with propertyValue === `null`
- Fixed small issue with `isExplicit` beeing locked from controller (video schema)
- `initializeContentDir` - fixed "non-empty" check to rely on `.classById.keys()` instead of `nextClassId()`

CLI:
- Small update in `JsonSchemaPrompt` to make it compatible with the new way of representing optional values (`null`)
- Fixed issue with `publishedBeforeJoystream` and allowed updating it to `[not set]` even after it was already set before
- Default value for `skippableIntroDuration`
- Added support for `~` and relative path resolution when specifying path to `UserDefinedLicense`
- Fixed `parseToKnownEntityJson` behaviour with optional fields

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -322,7 +322,7 @@ export default abstract class ContentDirectoryCommandBase extends RolesCommandBa
   async parseToKnownEntityJson<T>(entity: Entity): Promise<FlattenRelations<T>> {
     const entityClass = (await this.classEntryByNameOrId(entity.class_id.toString()))[1]
     return (_.mapValues(this.parseEntityPropertyValues(entity, entityClass), (v) =>
-      v.value.toJSON()
+      v.type !== 'Single<Bool>' && v.value.toJSON() === false ? null : v.value.toJSON()
```
mnaamani, line 325: Adding some `(` brackets `)` will go a long way to help understand the logic of this expression

mnaamani (commented):
I'm finding the approach to set InputPropertyValue to `Single(InputValue::Bool(false))` a bit hard to understand. Is that propagating to constructing an `Operation` to update property values on an entity, or is it simply for display reasons?

Apart from this point, everything else looks good.

Lezek123:
> I'm finding the approach to set InputPropertyValue to Single(InputValue::Bool(false)) a bit hard to understand. Is that propagating to constructing an Operation to update property values on an entity, or is it simply for display reasons?

It's a way to "clear" a non-required field: https://github.com/Joystream/joystream/blob/babylon/runtime-modules/content-directory/src/schema/property.rs#L466

`Bool(false)` represent a "lack of value", unless the underlying type of the field is `Bool` (in which case I think there is no way to differentiate between "no value" and "value set to `false`")

cli/src/base/ContentDirectoryCommandBase.ts
Lezek123: I added them, but they are automatically stripped due to linter/prettier rules

mnaamani: yes I was expecting that hehe. Okay now with your clarification in https://github.com/Joystream/joystream/pull/1614#issuecomment-721551034 it makes sense and confirms my memory of logical operator precedence is still intact ;)

mnaamani (approved):

iorveth:
@Lezek123 you are right. Previously in old `versioned-store` we had None enum variant to specify such case, but afterwards @bedeho suggested this approach to not sacrifice type safety.

PR #1618, opened 2020-10-27
channels responsive improvements

### LINKED ISSUE #1603: Channel Titles should not be Wrapped

kdembler:
Resolves #1603 and #1604

### DIFF
--- a/atlas/src/components/Hero.tsx
+++ b/atlas/src/components/Hero.tsx
@@ -2,7 +2,6 @@ import React from 'react'
 import { fluidRange } from 'polished'
 import { css } from '@emotion/core'
 import { Button, Header } from '@/shared/components'
-import sizes from '@/shared/theme/sizes'
 
 type HeroProps = {
   backgroundImg: string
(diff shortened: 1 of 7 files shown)


kdembler:
@Gamaranto rebased on top of the Glider.js changes. There's one issue though - because the `ChannelPreview` doesn't have a specified width now, it will grow to take the full row inside the carousel. I think this is the same kind of problem as with the `VideoPreview` and will be fixed by your other responsiveness changes, right?

fcescob:
> @Gamaranto rebased on top of the Glider.js changes. There's one issue though - because the `ChannelPreview` doesn't have a specified width now, it will grow to take the full row inside the carousel. I think this is the same kind of problem as with the `VideoPreview` and will be fixed by your other responsiveness changes, right?

Yes, you are correct, it will be fixed in the responsiveness changes.

fcescob (approved):
LGTM!

bedeho (approved):

Issue #1227, opened 2020-09-03
Proposal discussion permissioning


bedeho:
Basic logic is as described in the corresponding section of the handbook, but the discussion is deleted from storage, not just locked from posting.

PR #1619, opened 2020-10-27
Proposals modules changes.

shamil-gadelshin:
### Change list
- Include early rejection of the proposals similar to approval and slashing.
- [Discussion permissions](https://github.com/Joystream/joystream/issues/1227).
- Refactor origin clones: remove custom implementation when Substrate supported it via Clone trait.
- Copy parameters from `runtime upgrade` to `amend constitution` proposal.
- Drop title and description from the proposals engine state, but keep the validation logic.
- Remove caches: `ActiveProposalsIds`, `PendingExecutionIds` and `PendingConstitutionalityIds`.
- If the proposal is not elected during the council change - reject it.
- Drop thread title and post text from the proposals discussion state, but keep the validation logic.


runtime-modules/proposals/discussion/src/types.rs
```diff
@@ -71,6 +68,23 @@ impl<ThreadAuthorId: Clone> ThreadCounter<ThreadAuthorId> {
...
 
+/// Discussion thread permission modes.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum ThreadMode<MemberId> {
+    /// Every member can post on the thread.
+    Open,
+
+    /// Only author, councilor or white member list could post on the thread.
+    Closed(Vec<MemberId>),
```
bedeho, line 38: I think we need a static cap for this container.

runtime-modules/proposals/discussion/src/types.rs
```diff
@@ -8,24 +8,21 @@ use sp_std::vec::Vec;
...
 
 /// Post for the discussion thread
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
 #[derive(Encode, Decode, Default, Clone, PartialEq, Eq)]
 pub struct DiscussionPost<PostAuthorId, BlockNumber, ThreadId> {
-    /// Text
-    pub text: Vec<u8>,
-
     /// When post was added.
     pub activated_at: BlockNumber,
```
bedeho: What does this do now? I don't see it get read for transaction processing, is it not archival?

runtime-modules/proposals/discussion/src/types.rs
```diff
@@ -8,24 +8,21 @@ use sp_std::vec::Vec;
 /// Represents a discussion thread
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
 #[derive(Encode, Decode, Default, Clone, PartialEq, Eq)]
-pub struct DiscussionThread<ThreadAuthorId, BlockNumber> {
-    /// Title
-    pub title: Vec<u8>,
-
+pub struct DiscussionThread<ThreadAuthorId, BlockNumber, MemberId> {
     /// When thread was established.
     pub activated_at: BlockNumber,
```
bedeho, line 13: What does this do now? I don't see it get read for transaction processing, is it not archival?

runtime-modules/proposals/discussion/src/types.rs
bedeho: Same goes for `updated_at` and  `edition_number`.

runtime-modules/proposals/engine/src/types/mod.rs
```diff
@@ -875,4 +912,66 @@ mod tests {
         proposal.exact_execution_block = Some(3);
         assert!(!proposal.is_execution_block_reached_or_not_set(2));
     }
+
+    #[test]
```
bedeho: Why are tests inlined here? I thought we had abandoned this approach?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -629,12 +625,14 @@ impl<T: Trait> Module<T> {
...
+        // Filter active proposals and gather finalization data.
         // Skip proposals with unfinished voting.
-        <ActiveProposalIds<T>>::iter()
-            .filter_map(|(proposal_id, _)| {
-                // load current proposal
-                let proposal = Self::proposals(proposal_id);
+        <Proposals<T>>::iter()
+            .filter_map(|(proposal_id, proposal)| {
+                // Filter only active proposals.
+                if !proposal.status.is_active_proposal() {
```
bedeho: This segment can be made clearer. 

A proposal is either active, or it is not, in which case there is a decision status for it. This code however does not statically capture this, it appears as if `decision_status` could theoretically be `None`, but this will never happen at runtime. This can be improved to statically capture this, raising clarity and safety when someone tries to update this in the future. Comment should also be changed.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -433,7 +421,7 @@ decl_module! {
 
             // Check vote results. Approved proposals with zero grace period will be
             // transitioned to the PendingExecution status.
-            for  proposal_data in finalized_proposals {
+            for proposal_data in finalized_proposals {
```
bedeho: I don't understand what is happening here, why are you writing the proposal write back, and then running `finalize_proposal`, which mutates it. Moreover`finalize_proposal` does checks like `is_active_or_pending_execution` which trivially must hold given that there is a valid `decision_status` provided.
Later you even go in `get_approved_proposal_ready_for_execution` and reread same proposals from storage.

In general, reading the entire code path in `on_finalize` is not very easy, there are so mange branches with side-effects and stages where things are happening here.

The entire process should be a simple single pure computation `{p | proposal p} => {what to do about proposal p + require static information | proposal p}` and then one big `match` side-effect computation where you take the output of this and do what you need case by case, whether its overwriting an old proposal or deleting one, or executing. Right now I cant easily convince myself that this works, and you are almost certainly avoiding lots of checks and rechecks

runtime-modules/proposals/engine/src/types/mod.rs
```diff
@@ -279,8 +280,43 @@ where
...
+        );
+
+        let required_threshold_fraction =
+            Perbill::from_percent(self.proposal.parameters.approval_threshold_percentage);
+
+        possible_approval_votes_fraction.deconstruct() >= required_threshold_fraction.deconstruct()
+    }
+
+    // Verifies that both approval and slashing thresholds cannot be achieved.
+    pub fn is_rejection_imminent(&self) -> bool {
```
bedeho, line 303: I think this is too strict, remember, rejection is imminent if one cannot satisfy the constraint on both quorum AND threshold. This is equivalent to failing on one or the other (or both), for both approval and slashing. You are now requiring that you specifically cannot achieve the threshold, but this can in principle not be determinable as early.

bedeho (changes requested):
- It seems there is a good amount of archival state still to be removed.
- As an extension to the prior point, there is also some input length validation no longer needed when things are not stored in storage.
- Do we really need a global post counter when posts live in a double map with the thread? dropping it would also drop need for thread id in each post.
- We do not constrain the number of posts in a thread, yet we expect to delete them all when the proposal is finalized, this is not safe.
- `LastThreadAuthorCounter` does not seem to do anything? lets drop.
- `vote` is not doing finalization logic, do you think this is advisable? I know `on_finalize` does it later in the same block, but it leaves the proposal in a state that does not correspond to reality while any intermediate transactions are executed.

runtime-modules/proposals/discussion/src/types.rs
shamil-gadelshin: I removed activated_at and updated_at, but kept edition_number because we have a 'max allowed edits' policy.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: That's an artifact. These tests are very small and don't depend on the Substrate. I moved them to the 'tests' folder.

runtime-modules/proposals/discussion/src/types.rs
shamil-gadelshin: I removed it.

shamil-gadelshin:
- We use the global post counter as an indicator for the next post ID. But the thread_id in the post is unnecessary indeed (removed).
- `LastThreadAuthorCounter` is part of the anti-spam mechanism: it allows check for max thread in a row by the same author. But it looks like an artifact because we don't create threads outside the codex module, for example via some extrinsic. Let me know if we need to modify or remove it.
- Related to the 'intermediate proposal' state - currently, we calculate the proposal state on the block finalization, and 'during the block' a proposal doesn't have any state. If we allow vote extrinsic to do the finalization logic we could encounter a situation that in the same block one extrinsic would be successful because of the active proposal and another failed because of the finalized proposal.

runtime-modules/proposals/discussion/src/types.rs
shamil-gadelshin: Sure.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: Quorum calculation could be omitted because when dealing with for example "approve threshold achievability" we calculate possible outcome when all remaining councilors vote "Approve". Obviously 100% of votes satisfy any quorum requirements.

Rejection is imminent when (approve threshold is not achievable) AND (slash threshold is not achievable).

For example (test `proposal_status_resolution_approval_achievable_works_correctly`):
    Given:

```
    slashing_threshold_percentage: 50
    approval_threshold_percentage: 50
    council size: 600
```

    1. Only one councilor approves.
    2. 301 other councilors abstain.

So, even if the remaining 298 councilors also approve the proposal the approval threshold can not be achieved.

Thus, having 1 approval and 300 abstains we could decide that the proposal should be rejected.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Ok. I implemented the suggested dichotomy (active/others) using `match proposal.status` and commented 'others' branch.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: 'Write back a proposal before the finalization' was a harmless bug (I removed it). It is an artifact from the 'archival proposal world'.

The process in the on_finalize() is very similar to your description with one addition: on_finalize also gathers and executes the proposals with an expired grace period. I renamed types and methods and added more comments to improve code readability

Currently, on_finalize contains 4 steps:
1. Get proposals that ready for finalization.
2. Finalize them with side-effects.
3. Get proposals ready for execution (including proposals from the previous step).
4. Execute proposals that ready for execution.

I hope these stages became clear after the last refactoring.

runtime-modules/proposals/discussion/src/types.rs
bedeho: Lets drop that policy, and that number. We don't have such a policy in the forum, nor is it something I have really seen elsewhere, not sure why I suggested it to begin with.

runtime-modules/proposals/engine/src/types/mod.rs
bedeho: I don't fully understand your reply, but is your point that its actually never possible to rule out early that quorum is unsatisfiable, because everyone could still show up in time to satisfy any quorum requirement. The only thing you can rule out early is that a given threshold could be satisfied?

bedeho:
Before I do another review, lets just drop `LastThreadAuthorCounter` as you ask.

runtime-modules/proposals/engine/src/types/mod.rs
shamil-gadelshin: Correct.

runtime-modules/proposals/discussion/src/types.rs
shamil-gadelshin: Ok. Removed it.

shamil-gadelshin:
> Before I do another review, lets just drop `LastThreadAuthorCounter` as you ask.

Done.

bedeho (approved):
Excellent job, well done.

PR #1620, opened 2020-10-27
always display navigation in the navbar

### LINKED ISSUE #1613: Navbar/searchbar tweaks
After a discussion following the UI review (#1584), we decided to tweak the navbar to:
- keep the searchbar aligned to the right instead of the center
- if possible, keep displaying the navigation links even when the searchbar is expanded

kdembler:
Resolves #1613

Screenshot 2020-10-27 at 5 37 16 PM
Screenshot 2020-10-27 at 5 37 23 PM
Screenshot 2020-10-27 at 5 37 36 PM

### DIFF
--- a/atlas/src/components/Navbar/Navbar.style.tsx
+++ b/atlas/src/components/Navbar/Navbar.style.tsx
@@ -80,9 +80,9 @@ export const Header = styled.header<NavbarStyleProps>`
 
   grid-template-columns: 1fr 2fr;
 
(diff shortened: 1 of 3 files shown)


fcescob (changes requested):
When pressing the searchbar, the Logo jumps to the left a bit, I think it would be better to keep it in place and just expand the search bar

kdembler:
That's actually how it worked before. I've removed the padding changes altogether, let me know what you think

fcescob (approved):
LGTM!

PR #1621, opened 2020-10-28
Add Error Boundaries To Routes and Galleries

fcescob:
This PR adds Error Boundaries to all routes and to Video and Channel Galleries in the Home screen. The `ErrorFallback` just displays the error message and has a `Try again` button. 
Clicking `Try again` redirects back to the home screen if the error is in a Route; If it's in a gallery, it will cause a refetch of the gallery data.
Addresses #1606


atlas/src/components/ChannelGallery.tsx
```diff
@@ -10,13 +10,17 @@ type ChannelGalleryProps = {
...
+  error?: Error
 }
 
 const PLACEHOLDERS_COUNT = 12
 
-const ChannelGallery: React.FC<ChannelGalleryProps> = ({ title, action, channels, loading }) => {
+const ChannelGallery: React.FC<ChannelGalleryProps> = ({ title, action, channels, loading, error }) => {
   const displayPlaceholders = loading || !channels
 
+  if (error) {
```
kdembler: It doesn't make much sense to pass an error to a component just so it can throw it.

atlas/src/components/ErrorFallback.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const StyledButton = styled(Button)`
+  color: ${colors.white};
+`
+
+const ErrorFallback: React.FC<FallbackProps> = ({ error, resetErrorBoundary }) => {
+  return (
+    <Container>
+      <p>Something went wrong:</p>
+      <pre>{error?.message}</pre>
```
kdembler: I don't think we should present a JS error to the user

atlas/src/components/VideoGallery.tsx
```diff
@@ -39,6 +41,9 @@ const VideoGallery: React.FC<VideoGalleryProps> = ({ title, action, videos, load
     }
   }, [])
 
+  if (error) {
```
kdembler: Same thing as with `ChannelGallery`

atlas/src/views/HomeView.tsx
```diff
@@ -34,13 +46,42 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
       <Hero backgroundImg={backgroundImg} />
       <Container>
-        <VideoGallery title="Newest videos" loading={newestVideosLoading} videos={newestVideos} />
-        <VideoGallery
-          title="Featured videos"
-          loading={featuredVideosLoading}
-          videos={featuredVideosData?.featured_videos}
-        />
-        <ChannelGallery title="Newest channels" loading={newestChannelsLoading} channels={newestChannels} />
+        <ErrorBoundary
```
kdembler: Instead of passing the error down and using `ErrorBoundary` to catch it back, let's just do a switch based on `error` right here

atlas/src/views/HomeView.tsx
```diff
@@ -34,13 +46,42 @@ const HomeView: React.FC<RouteComponentProps> = () => {
...
+          FallbackComponent={ErrorFallback}
+          resetKeys={[newestVideos, newestVideosLoading]}
+          onReset={() => refetchNewestVideos()}
```
kdembler: I don't think that fully works, I'd expect the carousel to go back to the loading state once I press that but that doesn't happen

atlas/src/views/SearchView.tsx
```diff
@@ -38,8 +38,11 @@ const SearchView: React.FC<SearchViewProps> = ({ search = '' }) => {
   if (loading || !data) {
     return <p>Loading...</p>
   }
+  if (error) {
+    throw error
+  }
   if (!data.search) {
-    return <p>Something went wrong...</p>
+    return <p>There was a problem loading your search...</p>
```
kdembler: If `data.search` is null, something went wrong so we shouldn't handle that case separately - from the user's point of view this should be the same case as when `error` is defined

atlas/src/components/ErrorFallback.tsx
```diff
@@ -0,0 +1,30 @@
...
+  display: grid;
+  place-items: center;
+`
+
+const StyledButton = styled(Button)`
+  color: ${colors.white};
+`
+
+const ErrorFallback: React.FC<FallbackProps> = ({ error, resetErrorBoundary }) => {
+  return (
```
kdembler, line 22: There is a screen in Figma showing how the "error page" should look like: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=2155%3A180756

Also, we probably need a distinction between a full view error (for example if the video fails to load in video view) and a component-level error as the one we'd display instead of a carousel

atlas/src/components/ErrorFallback.tsx
fcescob: You're right, I removed it and logged the error to the console in case we need it for debugging purposes.

atlas/src/components/ErrorFallback.tsx
fcescob: I had totally missed that, thanks. I've changed the error fallback for a route to that and kept the previous ones for galleries.

atlas/src/components/ChannelGallery.tsx
fcescob: Agree, I removed it here and from the Video Gallery

atlas/src/views/HomeView.tsx
fcescob: I changed this to work without the Error Boundary, let me know your thoughts

atlas/src/views/HomeView.tsx
fcescob: Added an option to `useQuery` and changed the logic a bit, now placeholders should load again if `Try again` is clicked

atlas/src/views/SearchView.tsx
fcescob: Agree. Changed this to throw here as well

atlas/src/views/SearchView.tsx
```diff
@@ -21,7 +21,7 @@ const VideosHeader = styled.h5`
 `
 const SearchView: React.FC<SearchViewProps> = ({ search = '' }) => {
   const [selectedIndex, setSelectedIndex] = useState(0)
-  const { data, loading } = useQuery<Search, SearchVariables>(SEARCH, { variables: { query_string: search } })
+  const { data, loading, error } = useQuery<Search, SearchVariables>(SEARCH, { variables: { query_string: search } })
 
   const getChannelsAndVideos = (loading: boolean, data: Search | undefined) => {
```
kdembler: I know this is old code but why not have this below loading and error handlers? Then we don't need to handle loading case there at all

kdembler (commented):
Heading in the right direction but there are still some things to improve:
• You haven't added `notifyOnNetworkStatusChange: true` for the featured videos query so retrying still doesn't work
• The first carousel has some weird behaviour - initially it's in the loading states, then the "something went wrong" component flashes for a second, carousel goes back to loading and then finally display the error component for good. There shouldn't be that flash
• The infinite grid on home view still doesn't have any error indication
• Also the video grid in the "Browse" view doesn't indicate a failure in any way (I guess we could use a "full view" error there)
• Maybe it would make sense to keep the section header when the section fails to load, so the user can see what actually failed to load?
• The search view seems broken - I only see "Loading..." and nothing ever happens, even if I cause an error in the resolver

atlas/src/components/ErrorFallback.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+
+import styled from '@emotion/styled'
+import { FallbackProps } from 'react-error-boundary'
+
+import { Button } from '@/shared/components'
+import { sizes, colors } from '@/shared/theme'
+
+const Container = styled.div`
+  padding: ${sizes.b4};
```
kdembler: This lacks `px` doesn't it?

kdembler (commented):
Just a reminder that the error view still doesn't trigger when error happens in search view or in a channel view.

atlas/src/components/ViewErrorFallback.tsx
```diff
@@ -0,0 +1,50 @@
...
+`
+
+const ErrorFallback: React.FC<FallbackProps> = ({ error, resetErrorBoundary }) => {
+  console.error(error)
+  return (
+    <Container>
+      <ErrorIllustration />
+      <Message>
+        <Title variant="h3">Ops! An Error occurred</Title>
+        <p>We could not aquire expected results. Please try reloading or return to the home page.</p>
```
kdembler: This has too low contrast. Check the color in the design

atlas/src/components/ErrorFallback.tsx
kdembler: @Gamaranto this still doesn't seem resolved

kdembler (commented):
LGTM!

bedeho (approved):

Issue #771, opened 2020-06-23
Add a fade to each last element in viewport in the VideoGallery Component [Atlas]


fcescob:
As in Figma, the last video in the viewport of a VideoGallery component should have a fade on the thumbnail. 

See point 7 here for details: https://github.com/Joystream/joystream/pull/693#pullrequestreview-432470782

and here: https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=1037%3A1082

PR #1637, opened 2020-11-02
Add Fade To Carousel Edges

fcescob:
Solves #771 , review after #1562 is merged


kdembler:
Try clicking some element in the carousel :P

kdembler (commented):
Much better now :P Shouldn't there be a fade for the left arrow as well when it appears though?

atlas/src/shared/components/Button/Button.tsx
```diff
@@ -11,40 +11,48 @@ export type ButtonProps = {
...
-  containerCss,
-  className,
-  onClick,
-}) => {
-  const clickable = !!onClick
-  const hasText = !!children
-  const handleClick = (e: React.MouseEvent<HTMLButtonElement>) => {
-    if (!onClick) return
-    onClick(e)
+const Button = React.forwardRef<HTMLButtonElement, Partial<ButtonProps>>(
```
kdembler: I think it'd be best to separate component declaration and `React.forwardRef`, for readability

kdembler (commented):
To summarize the points from the call:
- let's remove the old `Glider` component
- let's try to abstract some Glider-specific logic outside of `Carousel` component (like adding event handlers and so on)
- let's use pure CSS for adding the shadows so that we don't need any kind of scroll event handler

fcescob:
I believe #1709 is fixed too in this PR

kdembler (approved):
LGTM!

bedeho (approved):

bedeho (approved):

PR #1640, opened 2020-11-03
storage-node integration tests: retry on wait for ipfs registration

mnaamani:
Allow multiple tries when testing upload in storage-node integration tests, because ipfs name registration sometimes takes more than allotted time.


shamil-gadelshin (approved):
LGTM

PR #1641, opened 2020-11-03
Query node types

Lezek123:
**Scope**
- Fixes issue with `@joystream/types` augmentation in `query-node`.
- Updates `classId` mappings in `query-node/mappings/content-directory/content-dir-consts.ts` to match the ids of classes after they are generated by `yarn workspace cd-schemas initialize:dev` script (they probably became outdated after `CurationStatus` class was removed)

**Types issue - the cause**
The reason why the build was failing with `query-node/tsconfig.json` paths set to:
```
"@polkadot/types/augment": ["../types/augment-codec/augment-types.ts"]
```

Was mainly because in `query-node/tsconfig.json` we didn't have `esModuleInterop` set to `true`, so we had to use imports like:
```
import * as BN from 'bn.js'
```

Due to lack of `esModuleInterop` flag in `query-node/tsconfig.json`, imports like:
```
import BN from 'bn.js';
```
Were probably not correctly recognized, so any type that indirectly extended `BN` inside `@joystream/types` became incompatible with definitions that expected this type to contain, for example, `toNumber()` method.

**Type issue - the fix**
The issue was fixed by adding:
```
"esModuleInterop": true
```
to `query-node/tsconfig.json` and reverting the changes in `tsconfig` paths that caused the augmentation to not work at all.

Now the augmentations work correctly and I was able to remove the no longer necessary assertions in `query-node/mappings/content-directory/decode.ts`


iorveth (approved):

mnaamani (approved):
LGTM

metmirr (approved):

PR #1642, opened 2020-11-03
CLI: uploadVideo - use relative paths when resolving endpoints

Lezek123:
Solves issues pointed out by @bwhm which arised when storage node base api url included some additional path, ie.:
```
http://localhost/storage/
```
In which case endpoints need to be resolved by using relative path, like:
```
http://localhost/storage/discover/v0/${storageProviderId}
```
Instead of absolute path:
```
http://localhost/discover/v0/${storageProviderId}
```


mnaamani (approved):
Will just wait for CI check to complete before merging.

bwhm (approved):

PR #1645, opened 2020-11-04
Add MediaLocation and License entities

metmirr:
This PR:
- Add two new entities to input schema `License`, `MediaLocation` and related operations(create, remove entity and update entity property)
- Replace referenced entity id with the actual entity instance (ie. `languageId: Int!` -> `language: Language!`). Graphql query will look like this: 
```graphql
query {
  channels {
    title
    description
    language {
      id
      name
    }
  }
}```

### CODE REVIEW

**Comment by mnaamani** [comment]
looks like you brought use of `happenedIn` property but its not defined on the types in question in the actual graphql-schema.json file?

https://github.com/Joystream/joystream/runs/1351475258?check_suite_focus=true#step:7:23

**Comment by metmirr** [comment]
> looks like you brought use of `happenedIn` property but its not defined on the types in question in the actual graphql-schema.json file?
> 
> https://github.com/Joystream/joystream/runs/1351475258?check_suite_focus=true#step:7:23

These are types generated by hydra-cli (`generated/graphql-server`) based on the input schema (`schema.graphql`). I am not sure what is causing this. Locally I can build without error.

**Comment by mnaamani** [comment]
> > looks like you brought use of `happenedIn` property but its not defined on the types in question in the actual graphql-schema.json file?
> > https://github.com/Joystream/joystream/runs/1351475258?check_suite_focus=true#step:7:23
> 
> These are types generated by hydra-cli (`generated/graphql-server`) based on the input schema (`schema.graphql`). I am not sure what is causing this. Locally I can build without error.

Ah okay that property was missing in earlier version of the schema, I can try to test locally with `yarn build` and report back.

Perhaps the fix in https://github.com/Joystream/joystream/pull/1641 might be needed ?

**Comment by mnaamani** [comment]
I'm getting the same errors locally.. 

> These are types generated by hydra-cli (`generated/graphql-server`) based on the input schema (`schema.graphql`). I am not sure what is causing this. Locally I can build without error.

@metmirr Are you running codegen by using the hydra-cli that is now a dependency in our repo with simply:

```
cd query-node/
yarn build
```

The "build" script will do `yarn hydra-cli codegen` then followed by updating tsconfig.json file in indexer, and finally `tsc --build tsconfig.json` on the actual mappings which generates these errors above.

**Comment by mnaamani** [comment]
Looks good, just waiting for your update to fix the error

```
TypeError: Cannot read property 'title' of undefined
    at Object.channelCreation [as default] 
```

From chat I see you already identified the issue with @iorveth which required un-commenting a line in the schema?
https://github.com/Joystream/joystream/blob/babylon/query-node/schema.graphql#L99

PR #1647, opened 2020-11-05
query-node: clean build

mnaamani:
- properly ignore `query-node/generated/` folder when building docker image `joystream/apps`
- clean generated folder before rebuilding to avoid using files generated by prior codegen runs which may have had a different schema file.
- moved the build script steps into a bash script to for clarity


iorveth (approved):
LGTM!

PR #1656, opened 2020-11-06
council - candidacy note

ondratra:
This PR adds the possibility to council candidates to add a short description about their candidacy.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -29,6 +29,7 @@
 //! ## Supported extrinsics
 //! - [announce_candidacy](./struct.Module.html#method.announce_candidacy)
 //! - [release_candidacy_stake](./struct.Module.html#method.release_candidacy_stake)
+//! - [set_candidacy_note](./struct.Module.html#method.set_candidacy_note)
 //!
 //! ## Important functions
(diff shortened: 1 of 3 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -399,6 +406,28 @@ decl_module! {
...
+        #[weight = 10_000_000]
+        pub fn set_candidacy_note(origin, council_user_id: T::CouncilUserId, note: Vec<u8>) -> Result<(), Error<T>> {
+            // ensure action can be started
+            EnsureChecks::<T>::can_set_candidacy_note(origin, &council_user_id)?;
+
+            // calculate note's hash
+            let note_hash = T::Hashing::hash(note.as_slice());
+
+            //
+            // == MUTATION SAFE ==
```
bedeho, line 417: I think this note can be a bit higher up, right after `can_set_candidacy_note`.

bedeho (changes requested):
Looks good, just a minor cosmetic change needed.

runtime-modules/council/src/lib.rs
ondratra: Done

bedeho (approved):

PR #1658, opened 2020-11-07
api-scripts: increase retry attempts on connect

mnaamani:
These scripts are used by tests to wait for joystream-node to come up,
even after the process/service is up it may take some time to initialize
genesis state. Increasing the retry count will help when running on systems with limited resources.


mnaamani:
included in https://github.com/Joystream/joystream/pull/1659

PR #1659, opened 2020-11-07
Docker/unified docker compose

mnaamani:
One docker-compose file to rule them all.

Fixed docker build for pioneer.
Simplified joystream/node Dockerfile (removed dependency on intermediate rust-builder image)

Because ipfs http gateway and graphql were configured to listen on port 8080, I moved graphql to port 8081 and updated .env in network tests

Modified query-node/package.json docker related scripts

Also updated to hydra-cli v0.0.20 and enabled esModuleInterop as per changes in https://github.com/Joystream/joystream/pull/1645 

Will wait for fixes in that branch first to merge into this update.

Various improvement to setup.sh, now installs volta to manage node and yarn version used by project/monorepo.

Added `build.sh` and `start.sh` scripts. So now with a newly cloned repo (provided docker and docker-compose is installed) a developer can run a local development instance of joystream with following steps:

```
./setup.sh
yarn
yarn build
yarn start
```

Then use joystream-cli and/or open pioneer at `localhost:3000/`

During build step you will be prompted if you want to build native binaries (you can skip that step) but the joystream/node docker image is required. If you don't want to build the joystream/node docker image, you can instead pull down the babylon version and re-tag it so it is used by the "start" script like so:

```
docker pull joystream/node:babylon
docker image tag joystream/node:babylon joystream/node:latest
```


shamil-gadelshin (approved):
I couldn't build the joystream-node locally for an unknown reason, but it worked when I pulled the babylon.

mnaamani:
> I couldn't build the joystream-node locally for an unknown reason, but it worked when I pulled the babylon.

Thanks for making the effort to test. It seems from our chats that the build failure was not clear, the rust compiler was being killed "(signal: 9, SIGKILL: kill)". It could be something specific to your mac docker desktop settings. Would be interesting to see if anyone else faces similar problems.

I was planning to update the build.sh to try to automatically fetch a pre-built image from docker hub matching the code shasum. There are more improvements such as using cached/downloaded creates from host machine. Will add in future.

PR #1660, opened 2020-11-08
add cover featured video

kdembler:
Resolves #1644

Screenshot 2020-11-08 at 6 51 26 PM

PR #1664, opened 2020-11-09
improve channel view cover

### LINKED ISSUE #1612: Lock the channel cover and hero image so they don't move around
As per UI review (#1584), hero image and channel covers should be locked in place so they always display the same content regardless of the viewport size

kdembler:
### Review after #1660 

Resolves #1612 - makes the channel cover image follow roughly the same policy as featured video cover - keep the aspect ratio of the image and let the page content cover it slightly

### DIFF
--- a/atlas/src/views/ChannelView/ChannelView.style.tsx
+++ b/atlas/src/views/ChannelView/ChannelView.style.tsx
@@ -1,33 +1,87 @@
 import styled from '@emotion/styled'
 import { fluidRange } from 'polished'
 import { Avatar, Placeholder } from '@/shared/components'
-import theme from '@/shared/theme'
+import theme, { breakpoints } from '@/shared/theme'
 import { css } from '@emotion/core'
 
 const SM_TITLE_HEIGHT = '48px'
 const TITLE_HEIGHT = '56px'
 
-type ChannelHeaderProps = {
-  coverPhotoURL: string
+type CoverImageProps = {
+  src: string
 }
-export const Header = styled.section<ChannelHeaderProps>`
-  background-image: linear-gradient(0deg, #000000 10.85%, rgba(0, 0, 0, 0) 88.35%),
-    ${(props) => `url(${props.coverPhotoURL})`};
-  background-size: cover;
-  background-position: center center;
-  background-repeat: no-repeat;
-  height: 430px;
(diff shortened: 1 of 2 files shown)


atlas/src/components/FeaturedVideoHeader/FeaturedVideoHeader.style.ts
```diff
@@ -0,0 +1,131 @@
+import styled from '@emotion/styled'
+import { fluidRange } from 'polished'
+
+import { Avatar, Button } from '@/shared/components'
+import { breakpoints, colors, spacing, typography } from '@/shared/theme'
+import { Link } from '@reach/router'
+
+export const Container = styled.section`
+  position: relative;
+  @media screen and (min-width: ${breakpoints.small}) {
```
fcescob: It looks to me that margins and gradient % need to be in sync based on what "looks good" this is hard to grasp without context, so maybe let's add a comment here describing that.

atlas/src/components/FeaturedVideoHeader/FeaturedVideoHeader.style.ts
```diff
@@ -0,0 +1,131 @@
...
+    margin-bottom: -75px;
+  }
+  @media screen and (min-width: ${breakpoints.medium}) {
```
fcescob: Not about code, but a big part of the image gets covered on bigger screens 
image

atlas/src/views/ChannelView/ChannelView.style.tsx
```diff
@@ -1,34 +1,88 @@
...
-  background-repeat: no-repeat;
-  height: 430px;
-  padding: 0 var(--global-horizontal-padding);
+
+export const Header = styled.section`
+  position: relative;
+  margin-bottom: 60px;
+
+  @media screen and (min-width: ${breakpoints.small}) {
+    margin-bottom: -75px;
```
fcescob, line 21: Same comments as in the `FeaturedVideoHeader` apply here

fcescob (commented):
Some minor fixes needed, otherwise looks good.

fcescob (approved):
LGTM!

bedeho (approved):

PR #1665, opened 2020-11-09
Make Preview Elements Resize Accordingly To Preview Width

### LINKED ISSUE #1623: Make Typography Fluid in Video and Channel Previews

fcescob:
This fixes #1623 #1622 . It adds a `scalingFactor` prop which ranges between a `MIN_SCALING_FACTOR` of 1 and a `MAX_SCALING_FACTOR` of 1.375 (this was chosen arbitrarily, it produces a `font-size` for the title of 1.375rem and I thought it looked good)


kdembler (commented):
So like I've mentioned in the call today I was curious to see how this looks like but I'd avoid using JS to do that if possible. Doing that adds a lot of unnecessary logic and that means potential failure points. I think this should be achievable with pure CSS - `%` of width for the avatar size, and we can go the `fluidRange` way for the font size. Unless there's some major obstacle with doing just CSS, I'd be hesitant to move forward this way here

fcescob:
> So like I've mentioned in the call today I was curious to see how this looks like but I'd avoid using JS to do that if possible. Doing that adds a lot of unnecessary logic and that means potential failure points. I think this should be achievable with pure CSS - `%` of width for the avatar size, and we can go the `fluidRange` way for the font size. Unless there's some major obstacle with doing just CSS, I'd be hesitant to move forward this way here

I've thought about doing that, but I think JS is necessary here, for a couple of reasons:

- `fluidRange` scales with the viewport, but we want to scale text based on the preview size 
- For the avatar, `%` values work for the width but not for the `height`, viewport units also don't work for the reason mentioned above. The alternative to JS was setting `width` based on a `%` and then use a pseudo element to preserve the 1:1 aspect ratio. But since JS is needed for the point above, I went for the JS alternative

I might have missed something, so if you find a better solution I'm happy to go for it.

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -71,6 +72,25 @@ const VideoPreview: React.FC<VideoPreviewProps> = ({
...
+  const MAX_COVER_WIDTH = 600
+  const MIN_SCALING_FACTOR = 1
+  const MAX_SCALING_FACTOR = 1.375
+  // Linear Interpolation, see https://en.wikipedia.org/wiki/Linear_interpolation
+  const calculateScalingFactor = (coverWidth: number) =>
+    MIN_SCALING_FACTOR +
+    ((coverWidth - MIN_COVER_WIDTH) * (MAX_SCALING_FACTOR - MIN_SCALING_FACTOR)) / (MAX_COVER_WIDTH - MIN_COVER_WIDTH)
+
+  const [scalingFactor, setScalingFactor] = useState(MIN_SCALING_FACTOR)
+  const { ref: imgRef } = useResizeObserver<HTMLImageElement>({
```
kdembler: Since we're using hooks here I think it'd be best to move this code fragment above the click handlers. In general, I think we should follow the following component code layout:
1. Hooks
2. Event handlers
3. Return

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -71,6 +72,25 @@ const VideoPreview: React.FC<VideoPreviewProps> = ({
     onClick(e)
   }
 
+  const MIN_COVER_WIDTH = 300
```
kdembler: Where does this and `MAX_COVER_WIDTH` come from? Should this one by correlated with the default value of `Grid` prop `minWidth`? I'm afraid of having constant values all over the place. Also, how does `MAX_COVER_WIDTH` correlate with `MAX_VIDEO_PREVIEW_WIDTH` from `VideoPreviewBase.styles.tsx`? My guess would be that we only need one of them

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
fcescob: Good Idea. I've moved the hooks to the start of the component and the interpolation logic out of it since its static.

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
fcescob: This choice for `MAX_COVER_WIDTH` is arbitrary. I set it to `600` as I've notice that previews on smaller screen almost never get wider than 600px so I've used to test how the Avatar and Title would look like in that case.
Note that, when previews get wider than 600px, for example on a _very_ wide screen,  the `scalingFactor` will keep increasing past `1.375` and so Title and Avatar will get bigger. 
I think `MAX_VIDEO_PREVIEW_WIDTH` in `VideoPreviewBase.styles.tsx` is stale code from when we had a `max-width` for video previews, and it was indeed that value. I don't think its used anywhere now.

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
kdembler: Looks much better now :+1:

kdembler (approved):
LGTM!

bedeho (approved):

PR #1676, opened 2020-11-11
Council staking accounts

ondratra:
This PR focuses on distinguishing between authorizing and staking accounts. A candidate can select a staking account during candidacy announcement. A voter can specify a staking account in the same fashion. During stake release, there is no need to specify the staking account as it will be read from the storage.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -203,11 +203,6 @@ pub trait Trait: system::Trait {
         membership_id: &Self::MembershipId,
         account_id: &<Self as system::Trait>::AccountId,
     ) -> bool;
-
-    fn is_account_free_of_conflicting_stakes(
-        staking_account_id: &<Self as system::Trait>::AccountId,
-        stake: &Balance<Self>,
-    ) -> bool;
 }
 
 /// Trait with functions that MUST be called by the runtime with values received from the referendum module.
@@ -753,7 +748,8 @@ impl<T: Trait> EnsureChecks<T> {
             return Err(Error::MembershipIdNotMatchAccount);
         }
 
-        if !T::is_account_free_of_conflicting_stakes(&staking_account_id, &stake) {
+        // ensure there are no conflicting stake types for the account
+        if !T::CandidacyLock::is_account_free_of_conflicting_stakes(&staking_account_id) {
             return Err(Error::ConflictingStake);
         }
 
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -287,6 +287,9 @@ decl_error! {
         /// Council member and candidates can't withdraw stake yet.
         StakeStillNeeded,
 
+        /// User tried to release stake when no stake exists.
+        NoStake,
+
         /// Insufficient balance for candidacy staking.
         InsufficientBalanceForStaking,
 
@@ -374,14 +377,14 @@ decl_module! {
         /// Release candidacy stake that is no longer needed.
         #[weight = 10_000_000]
         pub fn release_candidacy_stake(origin, council_user_id: T::CouncilUserId) -> Result<(), Error<T>> {
-            let account_id = EnsureChecks::<T>::can_release_candidacy_stake(origin, &council_user_id)?;
+            let staking_account_id = EnsureChecks::<T>::can_release_candidacy_stake(origin, &council_user_id)?;
 
             //
             // == MUTATION SAFE ==
             //
 
             // update state
-            Mutations::<T>::release_candidacy_stake(&account_id, &council_user_id);
+            Mutations::<T>::release_candidacy_stake(&staking_account_id, &council_user_id);
 
             // emit event
             Self::deposit_event(RawEvent::CandidacyStakeRelease(council_user_id));
@@ -734,10 +737,10 @@ impl<T: Trait> EnsureChecks<T> {
     fn ensure_previous_stake_reusable(
         council_user_id: &T::CouncilUserId,
         account_id: &T::AccountId,
-    ) -> Result<(), Error<T>> {
+    ) -> Result<Option<T::AccountId>, Error<T>> {
         // escape when no previous candidacy stake is present
         if !Candidates::<T>::contains_key(council_user_id) {
-            return Ok(());
+            return Ok(None);
         }
 
         let candidate = Candidates::<T>::get(council_user_id);
@@ -752,7 +755,7 @@ impl<T: Trait> EnsureChecks<T> {
             return Err(Error::InvalidAccountToUnstake);
         }
 
-        Ok(())
+        Ok(Some(candidate.staking_account_id))
     }
 
     /////////////////// Action checks //////////////////////////////////////////
@@ -819,9 +822,14 @@ impl<T: Trait> EnsureChecks<T> {
         }
 
         // ensure that previous stake is unlockable (if any present)
-        Self::ensure_previous_stake_reusable(council_user_id, &account_id)?;
+        let mb_staking_account_id =
+            Self::ensure_previous_stake_reusable(council_user_id, &account_id)?;
 
-        Ok(account_id)
+        if let Some(staking_account_id) = mb_staking_account_id {
+            return Ok(staking_account_id);
+        }
+
+        Err(Error::NoStake)
     }
 
     /// Ensures there is no problem in withdrawing already announced candidacy.
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -362,7 +362,7 @@ decl_module! {
         #[weight = 10_000_000]
         pub fn announce_candidacy(origin, council_user_id: T::CouncilUserId, staking_account_id: T::AccountId, stake: Balance<T>) -> Result<(), Error<T>> {
             // ensure action can be started
-            let stage_data = EnsureChecks::<T>::can_announce_candidacy(origin, &council_user_id, &staking_account_id, &stake)?;
+            let (stage_data, previous_staking_account_id) = EnsureChecks::<T>::can_announce_candidacy(origin, &council_user_id, &staking_account_id, &stake)?;
 
             // prepare candidate
             let candidate = Self::prepare_new_candidate(staking_account_id, stake);
@@ -370,6 +370,9 @@ decl_module! {
             //
             // == MUTATION SAFE ==
             //
+            if let Some(tmp_account_id) = previous_staking_account_id {
+                Mutations::<T>::release_candidacy_stake(&council_user_id, &tmp_account_id);
+            }
 
             // update state
             Mutations::<T>::announce_candidacy(&stage_data, &council_user_id, &candidate, &stake);
@@ -757,25 +760,6 @@ impl<T: Trait> EnsureChecks<T> {
         Ok(account_id)
     }
 
-    /// Checks any previous candidacy stakes has been unstaked or is reusable.
-    fn ensure_previous_stake_is_reusable(
-        council_user_id: &T::CouncilUserId,
-    ) -> Result<Option<CandidateOf<T>>, Error<T>> {
-        // escape when no previous candidacy stake is present
-        if !Candidates::<T>::contains_key(council_user_id) {
-            return Ok(None);
-        }
-
-        let candidate = Candidates::<T>::get(council_user_id);
-
-        // prevent user from candidating twice in the same election
-        if candidate.cycle_id == CurrentAnnouncementCycleId::get() {
-            return Err(Error::StakeStillNeeded);
-        }
-
-        Ok(Some(candidate))
-    }
-
     /////////////////// Action checks //////////////////////////////////////////
 
     /// Ensures there is no problem in announcing candidacy.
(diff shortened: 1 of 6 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -761,24 +782,33 @@ impl<T: Trait> EnsureChecks<T> {
     fn can_announce_candidacy(
         origin: T::Origin,
         council_user_id: &T::CouncilUserId,
```
bedeho: Rename `council_user_id` to `membership_id`, same with any other place where the id is identifying a member who is not yet a candidate.

runtime-modules/council/src/lib.rs
```diff
@@ -761,24 +782,33 @@ impl<T: Trait> EnsureChecks<T> {
...
             _ => return Err(Error::CantCandidateNow),
         };
 
         // ensure that previous stake is unlockable (if any present)
-        Self::ensure_previous_stake_reusable(council_user_id, &account_id).map_err(|error| {
-            if error == Error::StakeStillNeeded {
-                return Error::CantCandidateTwice;
-            }
+        let old_candidate =
+            Self::ensure_previous_stake_is_reusable(council_user_id).map_err(|error| {
```
bedeho: This name does not really make sense. I cannot sensibly ask that the stake of a member is reusable if they potentially have nothing staked even.

What this routine really does is check if a member has active candidacy, and it returns an old candidacy in case it had an old one as well. So name should reflect that, and error should be changed as well. Currently the error is being used directly in one caller extrinsic (`can_release_candidacy_stake`), and its being substituted in this caller. But errors from the module decl should only be used for module errors. However, if this routine is really a utility called by different callers, then there should be no error type (`()`), as there is no need to differentiate the case: there is only one way to fail in `ensure_previous_stake_is_reusable`.

runtime-modules/council/src/lib.rs
```diff
@@ -761,24 +782,33 @@ impl<T: Trait> EnsureChecks<T> {
...
+                    return Error::CantCandidateTwice;
+                }
+
+                error
+            })?;
 
-            error
-        })?;
+        // ensure existing and current staking account are the same
+        if let Some(tmp_old_candidate) = old_candidate {
```
bedeho: Why do these have to be the same? I should be able to stake with a new, or same, staking account associated with my same membership. 

Please explain.

What we do however have to do is to remove the lock that exists on that candidacy by virtue of failing to get elected from it _and_ not recovering the stake in the past. I am not 100% on whether my assumptions about cleanup are 100% correct, given that I am not on top of how you are using locks, but some variation of that should be true.

runtime-modules/council/src/lib.rs
```diff
@@ -807,21 +837,24 @@ impl<T: Trait> EnsureChecks<T> {
         council_user_id: &T::CouncilUserId,
     ) -> Result<T::AccountId, Error<T>> {
         // ensure user's membership
-        let account_id = Self::ensure_user_membership(origin, council_user_id)?;
+        Self::ensure_user_membership(origin, council_user_id)?;
+
+        // ensure stake is unlockable
+        let staking_account_id = match Self::ensure_previous_stake_is_reusable(council_user_id)? {
```
bedeho: Please split this up into two separate steps for clarity.

runtime-modules/council/src/lib.rs
```diff
@@ -807,21 +837,24 @@ impl<T: Trait> EnsureChecks<T> {
...
+            Some(candidate) => candidate.staking_account_id,
+            None => return Err(Error::NoStake),
+        };
 
         // ensure user is not current council member
```
bedeho: If I get here, then there must have been a candidate for this member, hence this member cannot be serving on the current council, as there is only at most one candidate object per member at any time, and it is removed any time the member wins a seat in the council.

runtime-modules/council/src/lib.rs
```diff
@@ -723,21 +735,35 @@ impl<T: Trait> EnsureChecks<T> {
...
+            T::is_council_member_account(council_user_id, &account_id),
             Error::CouncilUserIdNotMatchAccount,
         );
 
         Ok(account_id)
     }
 
-    /// Checks there are no problems with existing candidacy record (if any present).
-    fn ensure_previous_stake_reusable(
+    fn ensure_user_membership_staking(
```
bedeho: This method does two unrelated things, and the name gives ambiguous semantics. Please split out check for membership signer on origin from check that given member controls a staking account.

bedeho (changes requested):
While trying to evaluate the correctness of certain parts of this implementation I had to start understanding how so called "cycle ids" worked.

- The first issue I faced was that the state variable `CurrentAnnouncementCycleId` is not great. There is no such thing as a announcement cycle, there is no definition of what the id of such a cycle is, and to the extent that this is the announcement _period_, then most of the time there is no _current_ announcement period. Obviously I am being difficult here, but the name was an obstacle. What this really _should be_, is "the number of started announcement periods". Likewise, the variable name `cycle_id` should be something like "announcement_period_nr". This is more direct.
- The storage variable has comment
`Index of the current candidacy period. It is incremented everytime announcement period is.`  but the method `next_announcement_cycle` has the following comment when icnreasing the "id": `/// Increases announcing cycle index. It should be called after election phase ends or before announcing period is reset.` These appear to be in conflict.
- The truth appears to be the last comment, but I don't understand why you would increment the id when the election ends? as long as you do it whenever an announcement period ends, then you are uniquely identifying all periods, but now you are in fact skipping over ids? If I am right, this may not only be surprising, but may have security issues by making it seem that candidates have ids further in the past than they really are.

runtime-modules/council/src/lib.rs
ondratra: In the current implementation, announcing candidacy can be done after releasing a previous candidacy stake is released with an exception when the user wants to reuse the stake (setting different stake amount is possible) -> that's why only the same account id is allowed.

We could change this, and announcing candidacy may release previous candidacy stake. Originally, I was afraid it could create some side effects or edge cases, but after giving it a thought now, I think there is no reason to do so.

runtime-modules/council/src/lib.rs
ondratra: You are right, I'll remove this condition.

ondratra:
Ad `CurrentAnnouncementCycleId`:
1) I'll find a better name for the variable because it is not a *cycle* as you pointed out. 
2) One of the commentaries is incomplete (ending `... announcement period is.`), I'll fix it. 
3) The reason why the counter is increased after the election ends (or the announcement phase is restarted) rather than at the start of each announcing phase is the Idle period. If that was not a case, a user might call extrinsic that checks for `CurrentAnnouncementCycleId` during the Idle period and experience unexpected behavior - for example, a user couldn't release the candidacy stake because it would appear election is still running when it already finished.

bedeho:
3. I don't think the answer to this concrete situation is to introduce this strange semantic for this state variable ( `CurrentAnnouncementCycleId`), I cant even say what it represents. Instead you just need to have the check be that _either_ the cycle id of the candidate is in the past, or its the present one AND the Idle phase is set. Not only is this very straight forward, and preserves a reasonable interpretation of the state variable, but the current way seems unclear if its even safe.

runtime-modules/council/src/lib.rs
bedeho: Please do :)

runtime-modules/council/src/lib.rs
bedeho: Missing response.

runtime-modules/council/src/lib.rs
bedeho: Missing response.

runtime-modules/council/src/lib.rs
bedeho: Missing response.

runtime-modules/council/src/lib.rs
ondratra: The function has been removed by 'candidacy unstaking' changes https://github.com/Joystream/joystream/pull/1676#discussion_r521504113

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: The function has been removed by 'candidacy unstaking' changes https://github.com/Joystream/joystream/pull/1676#discussion_r521504113

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs

ondratra:
All issues have been addressed.

bedeho (commented):
This comment is just about commit fd42c31.

Where is this change coming from? I don't recall any recent discussion about changing referendum for any reason, nor does the change appear to make sense. 

Voting is already done directly by accounts, not by members, so why would separate staking accounts be introduced. For example `release_stake` now accepts a staking account as separate from the origin, the checker for this extrinsic is `can_release_stake`, it basically 

- checks that origin signed, in `ensure_regular_user`
- checks that there is a vote assocaited with the staking account
- checks other unrelated stuff

if this passes, its Ok. This means that anyone can release the stake of anyone else?

Also, looked at `can_reveal_vote`, isn't it supposed to check that you havent already done so?

runtime-modules/council/src/lib.rs
```diff
@@ -718,74 +716,52 @@ impl<T: Trait> EnsureChecks<T> {
...
-            return Err(Error::InvalidAccountToUnstake);
-        }
-
-        Ok(())
-    }
-
     /////////////////// Action checks //////////////////////////////////////////
 
     /// Ensures there is no problem in announcing candidacy.
     fn can_announce_candidacy(
```
bedeho, line 737: One cannot resuse stake from, for example validation or working groups, in candidacy, as shown here

https://joystream.gitbook.io/joystream-handbook/key-concepts/staking#reuse

Hence there needs to be a check for this (`is_account_free_of_conflicting_stakes` ?)

runtime-modules/council/src/lib.rs
```diff
@@ -810,8 +804,11 @@ impl<T: Trait> EnsureChecks<T> {
         let candidate = Candidates::<T>::get(council_user_id);
 
         // prevent user from releasing candidacy stake during election
-        if candidate.cycle_id == CurrentAnnouncementCycleId::get() {
-            return Err(Error::StakeStillNeeded);
+        if candidate.cycle_id == AnnouncementPeriodNr::get() {
```
bedeho: Can you rewrite this as a direct `&&` check, where second check is `matches` macro.

runtime-modules/council/src/lib.rs
```diff
@@ -832,14 +829,15 @@ impl<T: Trait> EnsureChecks<T> {
...
-        if candidate.cycle_id != CurrentAnnouncementCycleId::get() {
-            return Err(Error::NotCandidatingNow);
-        }
-
         // ensure candidacy announcing period is running now
         match Stage::<T>::get().stage {
-            CouncilStage::Announcing(_) => (),
+            CouncilStage::Announcing(_) => {
+                // ensure candidacy was announced in current election cycle
+                match Stage::<T>::get().stage {
```
bedeho: What is going on here? you are matching on the same thing `Stage::<T>::get().stage` nested? Doesn't the second math really need to be a check that the candidacy is from the same cycle?

bedeho (changes requested):
I think it was introduced at a prior point but missed, can you create a separate issue and future PR to fix this

```

        // TODO: try to find way how to get rid of unwrap here or staticly ensure it will not fail here
        // update state
        Mutations::<T>::finalize_announcing_period(&stage_data).unwrap(); // starting referendum should always start if implementation is valid - unwrap
```

runtime-modules/council/src/lib.rs
ondratra: Probably unintentional rewrite happened when copy/pasting the snippet. Check for `AnnouncementPeriodNr` belongs there. Fixed

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: The function `is_account_free_of_conflicting_stakes` has been introduced.

ondratra:
A new issue #1725 has been created to address `finalize_announcing_period`'s `unwrap` issue.

runtime-modules/council/src/lib.rs
```diff
@@ -204,6 +203,11 @@ pub trait Trait: system::Trait {
         membership_id: &Self::MembershipId,
         account_id: &<Self as system::Trait>::AccountId,
     ) -> bool;
+
+    fn is_account_free_of_conflicting_stakes(
```
bedeho: This needs to be on `StakeHandler` for consistency please review that trait.

runtime-modules/council/src/lib.rs
```diff
@@ -793,11 +805,13 @@ impl<T: Trait> EnsureChecks<T> {
         let candidate = Candidates::<T>::get(membership_id);
 
         // prevent user from releasing candidacy stake during election
-        if candidate.cycle_id == AnnouncementPeriodNr::get() {
-            match Stage::<T>::get().stage {
-                CouncilStage::Idle => (),
-                _ => return Err(Error::StakeStillNeeded),
+        if candidate.cycle_id == AnnouncementPeriodNr::get()
```
bedeho, line 804: Please use `matches` macro, as prior review said, this is needlessly verbose, macro gives pure variant condition test.

bedeho (changes requested):

runtime-modules/council/src/lib.rs
ondratra: Done; macro used. `if candidate.cycle_id == AnnouncementPeriodNr::get()
            && !matches!(Stage::<T>::get().stage, CouncilStage::Idle)
        {`

runtime-modules/council/src/lib.rs

bedeho (approved):

PR #1680, opened 2020-11-11
Proposals update. Parameters migration from the Storage to the Trait.

shamil-gadelshin:
### Summary

We created a separate runtime module(file) In order to be able to pass the proposal parameter constants during the compilation time. The primary goal for the conditional compilation is integration tests. All parameters for the proposal could be set via `ALL_PROPOSALS_PARAMETERS_JSON` environment variable. The value of the variable should be a valid JSON file (a sample file is included in the PR). The PR contains a migration of the single proposal parameter set for demo purposes.

### Comments

- We used `lite-json` crate as "no_std" option for parsing JSON recommended for Substrate. `serde_json` is not 'no_std' ready.  `serder_json_core` library is not working. 
- Tests are ready for the code but disabled (marked `ignore`).
- Corrupted JSON resulted in panic
- JSON with incorrect types (eg.: string instead of integer) resulted in panic
- 0 (zero) value for the optional type becomes None.
- Non-existent fields and parameter groups are set with default values.

The previous draft PR could be found [here](https://github.com/shamil-gadelshin/substrate-runtime-joystream/pull/5).


mnaamani (approved):

mnaamani:
Only issue here is that the CI checks specifically for network integration tests are from an old branch.
After merging we should probably update `proposals_update` branch from babylon.

PR #1692, opened 2020-11-12
Ui review

fcescob:


bedeho:
- Loaders for channels are clipped

Screenshot 2020-11-12 at 09 52 16

- The resizing policy on videos still breaks when max item cap (6) is reached, as margins start growing and coursel buttons drift off:

Screenshot 2020-11-12 at 14 58 26


- Carousel buttons for video previews are still not aligned vertically centered w.r.t. image preview.

- Carousel buttons for channel previews need to be vertically centered on the grey bounding box, see design on Figma for this.

- Lets have carousel button outward facing vertical edge align with content vertical edge in the carousel, see design on Figma. We have gone a bit back and forth on this, but I think none of the alternatives we have had in mind really work better for now. Both of the left and right buttons below are misaligned w.r.t. this new policy.

Screenshot 2020-11-12 at 13 16 39

- Top of image previews, at least when loading, have a non gradient dark border

Screenshot 2020-11-12 at 10 34 11

- The entire top cover has no linear left(black)-to-center(transparent) gradient.

- Lots of fonts, in particular when bold, look very blurred and bad.

- Hover effect on left most preview gets clipped on the left hand side.

Screenshot 2020-11-12 at 11 11 54

- Channel preview margins keeps drifting as one resizes the viewport, and the view count preview
 and bottom margin is gone

Screenshot 2020-11-12 at 11 15 43

- The resizing policy some how makes the top area jump, it even appears to impact the gradient between the first row and the top cover, just look at Arnold's face.

- The gradient becomes very strong for some viewport ranges, may be related to the prior point. We should try to have the cover be slightly visible between the top videos, just like the design, unless I am forgetting some problem with this.

weird_jumping

- I don't know if this is related to the prior point, probably is: the placement of the left hand metadata & logo is positioned in a way I cant quite wrap my head around. I suspect that maintaining a relatively stable, or perhaps just fixed, px gap between the video previews is the cleanest approach, even though it results in some of it disappearing/getting occluded, just like the videos, when the viewport becomes very wide.

- The fade in transition from loading to ready on video previews should also fade in the text and text, not just the cover.

- Lets have carousels rotate out full row, not just one item at a time.

- Play button should have play icon.

- Title text & icon & buttons in header section needs to be more sensitive to screen viewport size. I know the design does not give much indication about this

- The subtext on the header:
  - is too close to the main title, see original design.
  - changes in size substantially, and continuously, while the main title does not, this incongruence looks strange, and breaks the visual hierarchy of the design. See hierarchy in original design.
  - does not adjust the line height in conjunction with change in the size, which results in an unattractive crowding between wrapping lines when the font size gets big. If we are going to have wrapping, then make sure to keep a distance in line with the original design. It only shows one font size, but a distance informed by font-size should be derivable.
  - has lots of dynamic wrapping effects as viewport gets bigger, even when its much wider horizontally than the area the text needs. Can we maintain a static wrapping and just adjust the font-size to play well with the viewport size. Right now the dynamic wrapping does not look good, and it also affects the layout when going between 1 and 3 lines.
  - is a few pixels to far left aligned w.r.t. main title text.

- The top menu should look more like the design https://files.slack.com/files-pri/T04RX1Z0P-F01D12UKW0M/image.png, specifically
  - use full logo, not just the mark, and use correct size and position, and mind left margin w.r.t. viewport
  - increase height. The original design had shorter height, but still longer than what is in the app currently.
  - highlighted tabs should have solid icon
  - improve space between label icons and buttons.
  - right margin between search bar and viewport should be smaller.
  - increase contrast of "Search..." text and add search icon.

- When the backend was not working, I did not get a nice error message, I just got some text saying something went wrong.

- Channel name shows scroll bar at times

Screenshot 2020-11-12 at 11 37 28

- Channel cover image is resized properly, but lets make the first row be about 50% of the displacement it is now. Align the channel avatar + name with a fixed px gap w.r.t. first row. Unfortunately the design did not say much about the resize policy, but the one here seems the most functional w.r.t. having a tab bar here eventually.

- I am unable to exit empty search results back to whatever scene I was on, in whatever state it was in. This is a bad limitation and feels very unnatural. Only exit is to jump back to a top level  screen, losing all scroll state or other specific things I may be looking at. We should go back to having escape button and exit button as in the design.

- When searching, the transition from loading to showing results seems to have a sort of vertical drop for everything on the screen.

- The entire top area is clickable in search results, but it feels very unatural, and the click courser triggers substantial uncertainty of what would happen if you just click on the black void (at least for me). Lets stick to just have the huge video preview be the play trigger.

- Empty state for search result, or just one of the tabs on search, like channels being empty, shows a blank scene.

Issue #1697, opened 2020-11-13
CLI / content-directory-schemas: Featured videos


Lezek123:
Add support for featured videos in the content directory schemas and related functionality to CLI.

- add `FeaturedVideo` class and schema inputs. Each entity would contain a reference to `Video`
- add `joystream-cli media:setFeaturedVideos [IDS]` and `joystream-cli media:featuredVideos` (list) commands

iorveth:
> Add support for featured videos in the content directory schemas and related functionality to CLI.
> 
> * add `FeaturedVideo` class and schema inputs. Each entity would contain a reference to `Video`
> * add `joystream-cli media:setFeaturedVideos [IDS]` and `joystream-cli media:featuredVideos` (list) commands

Hey, can we close this issue? cc @Lezek123

PR #1702, opened 2020-11-13
Featured videos schemas + CLI support

Lezek123:
- implements https://github.com/Joystream/joystream/issues/1697
- introduces a small change in `content-directory-schemas` and `@joystream/cli` to make it more easy to initialize classes in a predictable order, which means `FeaturedVideo` class will always have an id of `13` (when using initialization scripts)
- adds support for providing custom path to inputs that will be used for the initialization in `joystream-cli content-directory:initialize`
- auto-updated the `README`

Adding support to already running chain should be possible without re-runing the initialization, just by running the following 2 commands:
- `yarn joystream-cli content-directory:createClass -i ./content-directory-schemas/inputs/classes/FeaturedVideoClass.json`
- `yarn joystream-cli content-directory:addClassSchema -i ./content-directory-schemas/inputs/classes/FeaturedVideoSchema.json`

**Commands added:**
- `yarn joystream-cli media:setFeaturedVideos [...]`
- `yarn joystream-cli media:featuredVideos` (list)

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -87,17 +87,26 @@ When using the CLI for the first time there are a few common steps you might wan
 * [`joystream-cli content-directory:curatorGroups`](#joystream-cli-content-directorycuratorgroups)
 * [`joystream-cli content-directory:entities CLASSNAME [PROPERTIES]`](#joystream-cli-content-directoryentities-classname-properties)
 * [`joystream-cli content-directory:entity ID`](#joystream-cli-content-directoryentity-id)
+* [`joystream-cli content-directory:initialize`](#joystream-cli-content-directoryinitialize)
+* [`joystream-cli content-directory:removeCuratorFromGroup [GROUPID] [CURATORID]`](#joystream-cli-content-directoryremovecuratorfromgroup-groupid-curatorid)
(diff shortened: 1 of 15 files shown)


mnaamani:
Tested and all works as expected.

Just minor correction in your instructions the path should be to the `inputs/schemas/` folder
>     * yarn joystream-cli content-directory:addClassSchema -i ./content-directory-schemas/inputs/classes/FeaturedVideoSchema.json


I'm glad you are addressing the possible lack of determinism of `fs.readdirSync` I had wondered about whether it was indeed a possible source of bugs in trying to sync the class ids with query-node and atlas.

Perhaps a simpler approach might be to have a file in `inputs/classes/index.js` where we explicitly force an ordering of json files to be processed:

rough idea:
```javascript
const ordering = [
  'ChannelClass',
  'ContentCategory',
  ...
]

// Array of sorted filenames (to get the classname) and the raw json
module.exports = ordering.map((fileName) => ({ 
     fileName,
     data: require(`./${filename}`)
}))
```

Lezek123:
> Perhaps a simpler approach might be to have a file in inputs/classes/index.js where we explicitly force an ordering of json files to be processed:

Pushed an update using this approach

PR #1704, opened 2020-11-13
Upgrade Substrate version from 2.0.0-rc4 to 2.0.0

shamil-gadelshin:
Upgrade comments:
- affected crates: node, runtime, chain-spec-builder, runtime-modules/*
- all affected crates versions were bumped
- system pallet was renamed to `frame-system` because of changed internal naming in the `decl_module!` macro
- rust compiler version was upgraded from nightly-2020-05-23 to nightly-2020-10-06.


runtime/src/lib.rs
```diff
@@ -70,12 +72,22 @@ pub use content_directory::{
...
+    )
+}
+
 /// This runtime version.
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("joystream-node"),
     impl_name: create_runtime_str!("joystream-node"),
-    authoring_version: 7,
-    spec_version: 7,
+    authoring_version: 8,
```
mnaamani: By bumping the authoring version we are making the statement that this runtime will not be a runtime upgrade for babylon, and will have to be a new chain. Is that the case?

The significant changes I am seeing seem around BABE consensus protocol seem to indicate so.

We certainly need to bump polkadot-api version on client side as there is a new metadata version, but I think we should be able to test runtime upgrade. This will require a minor fix to the CI checks because docker build is failing as its depending on an intermediate image (rust-builder) which isn't updated with latest rust version needed. This has been resolved in https://github.com/Joystream/joystream/pull/1659 so I will try to push a branch that combines this new work with the updated version of substrate.

runtime/src/weights/frame_system.rs
```diff
@@ -0,0 +1,57 @@
...
+
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 2.0.0-rc5
+
+#![allow(unused_parens)]
+
+use frame_support::weights::{constants::RocksDbWeight as DbWeight, Weight};
+
+pub struct WeightInfo;
+impl frame_system::WeightInfo for WeightInfo {
+    // WARNING! Some components were not used: ["b"]
```
mnaamani, line 26: What components are being referred to in this Warning ?

node/src/command.rs
```diff
@@ -84,17 +91,98 @@ pub fn run() -> Result<()> {
...
-                );
-                Ok(())
+                    .into())
             }
         }
-        Some(Subcommand::Base(subcommand)) => {
-            let runner = cli.create_runner(subcommand)?;
-
-            runner.run_subcommand(subcommand, |config| Ok(new_full_start!(config).0))
+        Some(Subcommand::Key(cmd)) => cmd.run(),
```
mnaamani, line 99: I tried to test the `key` command and `inspect` command. According to changes in https://github.com/paritytech/substrate/releases/tag/v2.0.0  the `inspect-key` subcommand was renamed to `inspect` but I still get the following message: 

```
error: The subcommand 'inspect' wasn't recognized
	Did you mean 'inspect-key'?

If you believe you received this message in error, try re-running with 'joystream-node key -- inspect'

USAGE:
    joystream-node key <SUBCOMMAND>
```

node/src/command.rs
```diff
@@ -84,17 +91,98 @@ pub fn run() -> Result<()> {
 
                 runner.sync_run(|config| cmd.run::<Block, Executor>(config))
             } else {
-                println!(
-                    "Benchmarking wasn't enabled when building the node. \
+                Err("Benchmarking wasn't enabled when building the node. \
```
mnaamani, line 94: I wanted to test running the benchmarking command, but i clearly forgot to build the node with that feature.
I tried to pass that argument at the root like so:

`cargo build --release --features runtime-benchmarks`

But i get error:

```
error: --features is not allowed in the root of a virtual workspace
note: while this was previously accepted, it didn't actually do anything
```

What is the correct way to build the node with the benchmarking feature?

runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -621,25 +621,15 @@ fn create_set_validator_count_proposal_common_checks_succeed() {
 #[test]
 fn create_set_validator_count_proposal_failed_with_invalid_validator_count() {
     initial_test_ext().execute_with(|| {
-        assert_eq!(
```
mnaamani, line 624: Could you briefly explain what was broken about this test that required these changes?

runtime-modules/versioned-store/src/lib.rs
```diff
@@ -697,7 +697,9 @@ impl<T: Trait> Module<T> {
         }
     }
 
+    // Preserve special formatting.
     #[rustfmt::skip]
+    #[allow(clippy::match_like_matches_macro)]
```
mnaamani, line 702: Seems to me that we should drop these deprecated modules especially when we launch a new chain for olympia.

runtime/src/lib.rs
```diff
@@ -70,12 +72,22 @@ pub use content_directory::{
     HashedTextMaxLength, InputValidationLengthConstraint, MaxNumber, TextMaxLength, VecMaxLength,
 };
 
+#[cfg(feature = "std")]
+/// Wasm binary unwrapped. If built with `BUILD_DUMMY_WASM_BINARY`, the function panics.
+pub fn wasm_binary_unwrap() -> &'static [u8] {
+    WASM_BINARY.expect(
+        "Development wasm binary is not available. This means the client is \
```
mnaamani, line 79: Isn't the message backwards? We only set `BUILD_DUMMY_WASM_BINARY` when running clippy linter, and therefore any artifact built is not intended for production.

wasm_binary_unwrap() during chainspec construction because a wasm blob must be present. So I would expect the message to say: "WASM binary is not available. This means the client was built with `BUILD_DUMMY_WASM_BINARY` flag and is not intended for production chains. Please rebuild with the flag disabled."

mnaamani (commented):
Fantastic work with this update as usual.

Had a couple of queries and wondering about whether this can work as runtime upgrade. Will need to do some tests.

I was able to run linter, tests and finally build and run the node.

However there seems to be an issue building initially:

```
     Compiling wasm-build-runner-impl v1.0.0 (/Users/mokhtar/joystream/joystream/target/release/wbuild-runner/joystream-node-runtime625570859927018581)
      Finished release [optimized] target(s) in 3m 28s
       Running `/Users/mokhtar/joystream/joystream/target/release/wbuild-runner/joystream-node-runtime625570859927018581/target/x86_64-apple-darwin/release/wasm-build-runner-impl`
      Updating crates.io index
  error: failed to select a version for `parity-util-mem`.
      ... required by package `sp-core v2.0.0 (https://github.com/paritytech/substrate.git?rev=a200cdb93c6af5763b9c7bf313fa708764ac88ca#a200cdb9)`
      ... which is depended on by `frame-executive v2.0.0 (https://github.com/paritytech/substrate.git?rev=a200cdb93c6af5763b9c7bf313fa708764ac88ca#a200cdb9)`
      ... which is depended on by `joystream-node-runtime v8.0.0 (/Users/mokhtar/joystream/joystream/runtime)`
      ... which is depended on by `joystream-node-runtime-wasm v1.0.0 (/Users/mokhtar/joystream/joystream/target/release/wbuild/joystream-node-runtime)`
  versions that meet the requirements `^0.7.0` are: 0.7.0

  the package `parity-util-mem` links to the native library `parity-util-mem-ban-duplicates`, but it conflicts with a previous package which links to `parity-util-mem-ban-duplicates` as well:
  package `parity-util-mem v0.6.1`
      ... which is depended on by `substrate-test-runtime v2.0.0-rc4 (/Users/mokhtar/.cargo/git/checkouts/substrate-7e08433d4c370a21/00768a1/test-utils/runtime)`
      ... which is depended on by `substrate-test-runtime-wasm v1.0.0 (/Users/mokhtar/joystream/joystream/target/release/wbuild/substrate-test-runtime)`

  failed to select a version for `parity-util-mem` which could resolve this conflict
warning: build failed, waiting for other jobs to finish...
```

But running build again works .. happened on two separate machines (which did have some cached crates from previous builds of babylon) Will need to test with a clean build.

runtime/src/lib.rs
mnaamani: Interestingly, the wasm builder version is bumped to v2.0.0 so now I was actually able to run clippy without need of setting `BUILD_DUMMY_WASM_BINARY=1` and it seemed to work without issues.

So we might be able to drop it from our build scripts.

runtime/src/lib.rs
shamil-gadelshin: I think it is rather risky to have it as a runtime upgrade. Because we cannot guarantee (neither can Parity) that 2.0.0rc4 libs are fully compatible with 2.0.0. The worst-case scenario would be "it works most of the time but fails on rare occasions".

mnaamani: Doing some testing here https://github.com/Joystream/joystream/pull/1717

shamil-gadelshin:
> However there seems to be an issue building initially:

I encountered similar build errors and clearing the local build cache works every time. I suspect a conflict between some libraries built with different compiler versions.

runtime/src/weights/frame_system.rs
shamil-gadelshin: I copied the entire file from the Parity Substrate `runtime` folder. Also this file is autogenerated.

node/src/command.rs
shamil-gadelshin: Try to run it in the `node` folder.

runtime-modules/versioned-store/src/lib.rs
shamil-gadelshin: Yes. You are correct. I was going to drop these modules during the integration of multiple modules in the olympia.

runtime/src/lib.rs
shamil-gadelshin: It is used only during the genesis config creation. This comment states that "Development wasm binary is not available" - I understand it as "You could not use 'development wasm' in this context". Anyway, I copied it from the parity node project and could use your suggested comment. Let me know if you want it changed.

runtime-modules/proposals/codex/src/tests/mod.rs
shamil-gadelshin: I noticed test duplication, which is likely to be the merge artifact. During the test debugging I changed the stake to the required value for this proposal type (it was unnecessary). And finally, the test was broken because of the default `staking::MinValidatorCount` changes. It looks like it is not set by default when using the latest Substrate libraries.

node/src/command.rs
shamil-gadelshin: I'm not sure of your initial command, but this commands work in my local enviroment:
```
./target/release/joystream-node inspect --dev block 1
./target/release/joystream-node key inspect-key
```

Also, the link you provided contains changes for the subkey utility.

runtime/src/weights/frame_system.rs
mnaamani: Is this what the benchmarking command produces ?

node/src/command.rs
mnaamani: Ah perhaps I mixed up the subkey command name with the subcommands for the node..

I tried:

```
./target/release/joystream-node key inspect
```

And as you pointed out the rename of `inspect-key` to `inspect` was for subkey utility not the node client :+1:

runtime-modules/versioned-store/src/lib.rs
mnaamani: Okay sounds like a good plan.

mnaamani:
> > However there seems to be an issue building initially:
> 
> I encountered similar build errors and clearing the local build cache works every time. I suspect a conflict between some libraries built with different compiler versions.

Yes I can confirm with a clean cache it works without issues!

runtime/src/lib.rs
mnaamani: I suppose there is not much point changing it, the key thing is the chainspec build step will fail when we need it to which is all that matters. Although in the past I have made the mistake of trying to run a node where the wasm bytes was empty, a panic occurs on execution of the node, I like that it is now caught earlier.

And the fact that we are interpreting the statement differently suggests to me that trying to open a PR upstream to fix the message is going to lack objectivity so I will spare myself :)

mnaamani:
Could you confirm that you can run clippy without need to set `BUILD_DUMMY_WASM_BINARY=1` I've done that locally on mac and linux, but I remember taking it off once, but you had to add it back because it failed on your system.

runtime/src/lib.rs
mnaamani: I remember we discussed this briefly, I agree since we don't have 100% test coverage in the integration tests there is a risk. However didn't polkadot and kusama do runtime upgrades recently to latest version? Can we ask or look into how they managed it.

mnaamani: So far my testing shows that it will be possible to upgrade the chain.
- The native binary can sync the current chain. So in terms of consensus, storage compatibility and networking it has no issues.

Screen Shot 2020-11-16 at 5 23 51 PM

- I ran a dev babylon chain initialized it with some state and did a runtime upgrade (kept the same authoring version and bumped the spec version) without issues.. block production, reading state, executing transactions all works as expected. (Using hosted polkadot-js/apps website for a working "pioneer")

mnaamani (approved):
As we discussed, will try to update the apps and integration tests to use compatible version of polkadot-js/api to get them working again so we can have higher confidence that the runtime upgrade will be reliable.

PR #1717, opened 2020-11-16
Runtime upgrade ci and build update

mnaamani:
For purpose of testing #1704 combined work from https://github.com/Joystream/joystream/pull/1659

I rolled back runtime version to remain at authoring version 7, and just bumped spec version relative to babylon, to test the runtime as an upgrade of a babylon chain.

Updated the docker image build config to use the new rust toolchain versions.

I do not expect the tests to pass, because the polkadot-api version and types library isn't updated, I just want to see if a chain "survives" the updated runtime :)


shamil-gadelshin (approved):
I tested it. It works and contains similar content with the previous similar PR.

Issue #1703, opened 2020-11-13
Atlas/Orion integration improvements


kdembler:
As part of the initial Atlas/query node integration, it turned out that the current Atlas/Orion integration needs some improvements. Specifically, Atlas currently sends a request to the query node containing the `views` field handled by Orion. Our mocking server didn't mind that and just returned the response without the `views` field, but the query node actually errors out with a message that it's not possible to query `views` field on the `Video` type. We will need to modify the stitching resolver so that it transforms the query node request and removes the `views` field from it. This way `views` can be later properly populated by Orion

Issue #1698, opened 2020-11-13
Atlas/Query node naming/types mismatch


kdembler:
During the initial integration of Atlas and the query node, some incompatibilities were found. This issue compiles the ones I've found so far. Please mind that it's possible there are some that I've missed and will be found only later once it's actually possible to test Atlas with the real query node.

### Connection queries naming

Atlas uses a pluralized form of a noun with a `Connection` suffix, the query node uses a singular form of that same noun. E.g. Atlas - `videosConnection`, query node - `videoConnection`.

__Recommendation__: Update the query node to use the plural form.
This isn't clearly defined by the Relay connection spec (it constantly uses just the plural noun, without the suffix), but the GraphQL docs use the form used by Atlas (https://graphql.org/learn/pagination/#complete-connection-model)

### Mismatch on the field name

List of mismatches with recommendation on which version to settle on:
- `Video.publishedOnJoystreamAt` _(Atlas)_ <-> `Video.createdAt` _(query node)_ - use __query node__ version
- `Video.thumbnailURL` _(Atlas)_ <-> `Video.thumbnailUrl` _(query node)_ - use __query node__ version
- `Channel.coverPhotoURL` _(Atlas)_ <-> `Channel.coverPhotoUrl` _(query node)_ - use __query node__ version
- `Channel.avatarPhotoURL` _(Atlas)_ <-> `Channel.avatarPhotoUrl` _(query node)_ - use __query node__ version
- `Channel.handle` _(Atlas)_ <-> `Channel.title` _(query node)_ - use __Atlas__ version

### Mismatch on the field type

- `Channel.coverPhotoUrl`, `Channel.avatarPhotoUrl`:
In the schema used by Atlas, both of these are nullable and that was our assumption during implementation (e.g. provide default channel avatar/cover). The query node marks both of those as non-nullable.
__Recommendation__: Make it consistent with the constraints of the content directory - if the fields aren't required, make them nullable

- `Video.channel`:
Atlas assumes the channel will never be null for a video, i.e. every video is always published by _some_ channel. The query node allows the channel to be nullable - that would have a lot of UI implications in Atlas.
__Recommendation__: Make the channel field non-nullable

- `Video.category` - Nullable in the query node, required in Atlas.
__Recommendation:__ Changing this doesn't have any implication for Atlas at this point, so I think it's mostly a product question @bedeho 

- `Video.publishedBeforeJoystream` - `String` in Atlas, `Int` in the query node.
[...]

PR #1722, opened 2020-11-16
rename schema fields, fix Orion integration

kdembler:
First part of changes related to #1698 - just renames of the fields
And fix for #1703


fcescob (approved):
LGTM.
Before we merge,  let's wait for #1698 to be resolved on the query node side and see if any type mismatch we might have missed pops up. 
I'm also okay with merging this as is and then coming back to add more types if turns out we missed any.

bedeho (approved):

PR #1723, opened 2020-11-16
Change Grid to use auto-fill by default

### LINKED ISSUE #1651: Add max width to videos on the Search View
Video Previews in the video section of the search view can get quite big if there is only a few of them. Fix this by given them a reasonable max-width

fcescob:
Fixes #1651


kdembler (approved):
LGTM!

bedeho (approved):

PR #1729, opened 2020-11-17
Empty search states

### LINKED ISSUE #1716: Add no search results view
We have a design in Figma telling what we should show when there are no search results. Let's implement and use that

fcescob:
Fixes #1716


atlas/src/views/SearchView/EmptyFallback.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const Title = styled(Typography)`
+  line-height: 1.25;
+`
+
+const EmptyFallback: React.FC = () => (
+  <Container>
+    <EmptyStateIllustration />
+    <Message>
+      <Title variant="h3">Sorry, we couldn&apos;t find any matches.</Title>
```
kdembler, line 38: Is there some reason to use `&apos;` instead of just inserting `'` insside the HTML?

atlas/src/views/SearchView/EmptyFallback.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <p></p>
```
kdembler: I don't think we need that here

atlas/src/views/SearchView/EmptyFallback.tsx
fcescob: ESlint was complaining, https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md

atlas/src/views/SearchView/EmptyFallback.tsx
fcescob: I forgot to add the subtext, added it now.

atlas/src/views/SearchView/EmptyFallback.tsx
kdembler: Ah, good to know

kdembler (commented):
LGTM

bedeho (approved):

PR #1731, opened 2020-11-17
Rebase with olympia and benchmark proposals engine and add weights for proposals 

conectado:
# Pull Request to track the progress of #1696 

This also includes the contents of PR #1764 meaning this PR also includes #1751 (find more discussion in the original PR)

This PR includes the following:
* Rebasing with branch olympia
* Move `on_finalize` function for `proposals_engine` to `on_initialize`
  * Refactor `on_initialize` to take into account weights for the executed proposals
* Weights for `proposal_discussion`
* Benchmark and weights for `proposals_engine`

The task for adding the benchmarks for `proposal_engine` includes: 
* [x] Add benchmarking to the runtime
* [x] Implement **benchmark** and **verfy** for these functions: **In-Progress**
  * [x] `vote`
  * [x] `cancel_proposal`
  * [x] `veto_proposal`
  * [x] `on_finalize`
* [x] Add weights in the `proposal_engine` extrinsics


runtime-modules/proposals/codex/src/tests/mock/mod.rs
```diff
@@ -132,13 +165,30 @@ parameter_types! {
     pub const MaxWhiteListSize: u32 = 20;
 }
 
+pub struct MockProposalWeight;
```
shamil-gadelshin: Consider reflecting the proposal dicussions pallet name: MockProposalDiscussionsWeight

runtime-modules/proposals/engine/Cargo.toml
```diff
@@ -16,13 +16,24 @@ sp-runtime = { package = 'sp-runtime', default-features = false, git = 'https://
 balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
 membership = { package = 'pallet-membership', default-features = false, path = '../../membership'}
 common = { package = 'pallet-common', default-features = false, path = '../../common'}
+frame-benchmarking = { package = 'frame-benchmarking', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca', optional = true}
+# We need to add the following dependencies(governance, recurringrrewards and minting) for benchmarking
```
shamil-gadelshin: Can't we add these dependencies as part of the 'benchmarking' feature?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -119,23 +119,38 @@ pub use types::{
...
 use frame_system::{ensure_root, RawOrigin};
-use sp_arithmetic::traits::Zero;
+use sp_arithmetic::traits::{Saturating, Zero};
+use sp_std::convert::TryInto;
 use sp_std::vec::Vec;
 
 use common::origin::ActorOriginValidator;
 use membership::staking_handler::StakingHandler;
 
+// proposals_engine
```
shamil-gadelshin: Consider changing the comment to rust-doc format with a note about autogenerated code.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -348,8 +369,28 @@ decl_module! {
         /// Exports const -  max simultaneous active proposals number.
         const MaxActiveProposalLimit: u32 = T::MaxActiveProposalLimit::get();
 
+        /// Block Initialization. Perform voting period check, vote result tally, approved proposals
+        /// grace period checks, and proposal execution.
+        fn on_initialize() -> Weight {
```
shamil-gadelshin, line 386: Add notes about the weight calculation in the on_initialize() comment.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -348,8 +369,28 @@ decl_module! {
...
+                        ::WeightInfo
+                        ::on_initialize_approved_pending_constitutionality(max_active_proposals)
+                )
+                .max(
+                    <T as Trait>::WeightInfo::on_initialize_rejected(max_active_proposals)
+                )
+                .max(
+                    <T as Trait>::WeightInfo::on_initialize_slashed(max_active_proposals)
+                )
+                .saturating_add(Self::process_proposals())
```
shamil-gadelshin: Consider separating code execution from the weight calculation to clarify the function.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -348,8 +369,28 @@ decl_module! {
         /// Exports const -  max simultaneous active proposals number.
         const MaxActiveProposalLimit: u32 = T::MaxActiveProposalLimit::get();
 
+        /// Block Initialization. Perform voting period check, vote result tally, approved proposals
+        /// grace period checks, and proposal execution.
+        fn on_initialize() -> Weight {
+            let max_active_proposals = T::MaxActiveProposalLimit::get();
+
```
shamil-gadelshin: This function is hard to read. Consider these improvements:
- separate weight calculations from the code execution
- separate each code path weight in its own variable and merge them in using the short expression
- consider introducing an alias for the `<T as Trait>::WeightInfo`
- comment each part of the weight formula

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -381,7 +422,7 @@ decl_module! {
         }
 
         /// Cancel a proposal by its original proposer.
-        #[weight = 10_000_000] // TODO: adjust weight
+        #[weight = <T as Trait>::WeightInfo::cancel_proposal(T::MaxLocks::get())]
```
shamil-gadelshin: Consider introducing an alias for the `<T as Trait>::WeightInfo`

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -780,18 +829,25 @@ impl<T: Trait> Module<T> {
...
+                            decision_status,
+                        ));
                     }
                 }
                 // Execute the proposal code if the proposal is ready for execution.
                 ProposalStatus::PendingExecution(_) => {
                     if proposal.is_ready_for_execution(now) {
-                        Self::execute_proposal(proposal_id);
+                        executed_weight =
+                            executed_weight.saturating_add(Self::execute_proposal(proposal_id));
```
shamil-gadelshin, line 922: Initial executed_weight = 0, you don't need to use `saturating_add` here, an assignment is enough.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -780,18 +829,25 @@ impl<T: Trait> Module<T> {
 
                     // If decision is calculated for a proposal - finalize it.
                     if let Some(decision_status) = decision_status {
-                        Self::finalize_proposal(proposal_id, proposal, decision_status);
+                        executed_weight.saturating_add(Self::finalize_proposal(
```
shamil-gadelshin, line 911: Initial executed_weight = 0, you don't need to use `saturating_add` here, an assignment is enough.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -685,13 +729,16 @@ impl<T: Trait> Module<T> {
 
             // immediately execute proposal if it ready for execution or save it for the future otherwise.
             if finalized_proposal.is_ready_for_execution(now) {
-                Self::execute_proposal(proposal_id);
+                executed_weight =
```
shamil-gadelshin: Initial executed_weight = 0, you don't need to use `saturating_add` here, an assignment is enough.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -398,7 +439,7 @@ decl_module! {
         }
 
         /// Veto a proposal. Must be root.
```
shamil-gadelshin, line 472: Parity uses this format:
```
T::WeightInfo::set_members(
				*old_count, // M
				new_members.len() as u32, // N
				T::MaxProposals::get() // P
			),
```
Can we use `T::WeightInfo` instead of `<T as Trait>::WeightInfo` as well?

runtime-modules/proposals/engine/src/benchmarking.rs
```diff
@@ -0,0 +1,533 @@
...
+                assert_in_events::<T>(
+                    RawEvent::ProposalExecuted(
+                        proposal_id.clone(),
+                        ExecutionStatus::failed_execution("Not enough data to fill buffer")).into()
+                );
+            }
+        }
+    }
+
+    on_initialize_approved_pending_constitutionality {
```
shamil-gadelshin, line 432: Did you cover Approved-PendingExecution case?

runtime-modules/proposals/discussion/src/lib.rs
```diff
@@ -228,7 +243,13 @@ decl_module! {
        }
 
         /// Changes thread permission mode.
```
shamil-gadelshin, line 264: Consider use Parity style T::WeightInfo and if it is impossible - introduce an type alias.

shamil-gadelshin (changes requested):
As a general comment - please, modify comments of all affected methods/extrinsics to reflect your changes.

runtime-modules/proposals/engine/src/lib.rs
conectado: [Here](https://github.com/paritytech/substrate/blob/v2.0.0/frame/babe/src/lib.rs#L266) I saw that they used this style since it's a similar case where `Trait` is bound by `pallet_timestamp` which has its own `WeightInfo` trait.

runtime-modules/proposals/engine/src/lib.rs
conectado: Here it's within the `for (proposal_id, proposal) in proposals` loop and `executed_weight` might not be 0 if another proposal was executed in the loop.

runtime-modules/proposals/engine/src/lib.rs
conectado: Here it's within the `for (proposal_id, proposal) in proposals` loop and `executed_weight` might not be 0 if another proposal was executed in the loop.

runtime-modules/proposals/engine/src/benchmarking.rs
conectado: I didn't because the Immediate Execution case calls `execute_proposal` and does more stuff before so it has to be a worse case. But just to be safe I will add it.

runtime-modules/proposals/engine/src/lib.rs
conectado: Let me know if you still think it's better to introduce an alias such as `ProposalsEngineWeightInfo` to solve this. 

I tried using `type WeightInfo<T> = <T as Trait>::WeightInfo` and I can't because the name overlaps with the `trait WeightInfo`.

shamil-gadelshin: I think the alias is a better option because you can use it also in the on_initialize() to simplify code.

runtime-modules/proposals/engine/Cargo.toml
conectado: The only way I found to do that is to change [here](https://github.com/Joystream/joystream/pull/1731/files#diff-7f65dc0a272bf88b83e4d7361ba52d17af2e8ad908f0d9cab5c84b4ae70de091L41) by this:
```rs
#[cfg(features = "runtime-benchmarks")]
impl_outer_event! {
    pub enum TestEvent for Test {
        balances<T>,
        engine<T>,
        membership_mod<T>,
        frame_system<T>,
        council<T>,
    }
}

#[cfg(not(features = "runtime-benchmarks"))]
impl_outer_event! {
    pub enum TestEvent for Test {
        balances<T>,
        engine<T>,
        membership_mod<T>,
        frame_system<T>,
    }
}
```

Otherwise, it would break the test compilation when **not** using using `--features runtime-benchmarks`.

If you think that's the best way to go I will change it.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -348,8 +373,57 @@ decl_module! {
...
+                .max(rejected_branch_weight)
+                .max(slashed_branch_weight);
+
+            // total_weight = executed_proposals_weight + maximum_branch_weight
+            executed_proposals_weight.saturating_add(maximum_branch_weight)
+        }
+
         /// Vote extrinsic. Conditions:  origin must allow votes.
-        #[weight = 10_000_000] // TODO: adjust weight
+        #[weight = WeightInfoEngine::<T>::vote(_rationale.len().try_into().unwrap())]
```
shamil-gadelshin: unwrap() and expect() are dangerous outside of the testing context. Please try to replace them with non-panicking alternatives like saturated_into() or another one.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Yes, you are correct.

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: Yes, you are correct.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -348,8 +373,57 @@ decl_module! {
...
+            // `on_initialize` does. Therefore, to get the total weight of `on_initialize` we need
+            // to add the weight of the execution of `on_intialize` to the weight returned by
+            // `process_proposal`.
+            // To be safe, we use the worst possible case for `on_initialize`, meaning that there
+            // are as many proposals active as possible and they all take the worst possible branch.
+
+            let max_active_proposals = T::MaxActiveProposalLimit::get();
+
+            // Weight when all the proposals are immediatly approved and executed
+            let immediate_execution_branch_weight =
```
shamil-gadelshin: Much better now. But I see the possibility to move the weight-code to a separate function. Do you think it would be better?

runtime-modules/proposals/engine/Cargo.toml
shamil-gadelshin: Isn't it possible to put dependencies both in "dev-dependencies" and to the feature scope?

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -615,13 +684,17 @@ impl<T: Trait> Module<T> {
     }
 
     // Executes proposal code.
-    fn execute_proposal(proposal_id: T::ProposalId) {
+    fn execute_proposal(proposal_id: T::ProposalId) -> Weight {
```
shamil-gadelshin, line 732: Please, add weight-related comments to the function.

runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -766,11 +845,13 @@ impl<T: Trait> Module<T> {
 
     /// Perform voting period check, vote result tally, approved proposals
     /// grace period checks, and proposal execution.
-    fn process_proposals() {
+    fn process_proposals() -> Weight {
```
shamil-gadelshin, line 895: Please, add weight-related comments to the function.

shamil-gadelshin (changes requested):

runtime-modules/proposals/engine/Cargo.toml
conectado: No, because dependencies in "dev-dependencies" can't have `optional = true` and dependencies in the feature scope needs `optional = true`

runtime-modules/proposals/engine/src/lib.rs
conectado: I agree! moved most of the calculation to a separate function.

runtime-modules/proposals/engine/Cargo.toml
shamil-gadelshin: I was able to run successfully both `cargo test` and `cargo build --features runtime-benchmarks` with this config:

```
[package]
name = 'pallet-proposals-engine'
version = '4.0.1'
authors = ['Joystream contributors']
edition = '2018'

[dependencies]
serde = { version = "1.0.101", optional = true, features = ["derive"] }
codec = { package = 'parity-scale-codec', version = '1.3.4', default-features = false, features = ['derive'] }
sp-std = { package = 'sp-std', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
frame-support = { package = 'frame-support', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
frame-system = { package = 'frame-system', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
pallet-timestamp = { package = 'pallet-timestamp', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
sp-arithmetic = { package = 'sp-arithmetic', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
sp-runtime = { package = 'sp-runtime', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
membership = { package = 'pallet-membership', default-features = false, path = '../../membership'}
common = { package = 'pallet-common', default-features = false, path = '../../common'}
frame-benchmarking = { package = 'frame-benchmarking', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca', optional = true}

# Benchmark dependencies.
governance = { package = 'pallet-governance', default-features = false, path = '../../governance', optional = true}
recurringrewards = { package = 'pallet-recurring-reward', default-features = false, path = '../../recurring-reward', optional = true}
minting = { package = 'pallet-token-mint', default-features = false, path = '../../token-minting', optional = true}

[dev-dependencies]
sp-io = { package = 'sp-io', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
sp-core = { package = 'sp-core', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
governance = { package = 'pallet-governance', default-features = false, path = '../../governance'}
recurringrewards = { package = 'pallet-recurring-reward', default-features = false, path = '../../recurring-reward'}
minting = { package = 'pallet-token-mint', default-features = false, path = '../../token-minting'}

[features]
default = ['std']
runtime-benchmarks = [
    'frame-benchmarking',
	'governance/std',
	'recurringrewards/std',
	'minting/std'
]
std = [
	'serde',
	'codec/std',
	'sp-std/std',
	'frame-support/std',
	'frame-system/std',
	'pallet-timestamp/std',
	'sp-arithmetic/std',
	'sp-runtime/std',
	'balances/std',
    'membership/std',
    'common/std',
]
```

shamil-gadelshin (approved):
LGTM

PR #1749, opened 2020-11-17
enable cover video playback

### LINKED ISSUE #1627: Finalise Atlas header section
Design is found here

https://www.figma.com/file/rBjfm2tJ2Pr9M0UNTduU9v/Joystream-Atlas?node-id=2374%3A45800

Here is the description of how it should work

- The top area should first load an image, and then when a background video is loaded, and if the user still lingers in the top area, it should transition into a video playing in the background.
- There is explicit visual handling of loading of the initial top image.
- The top image/video will fill the full width of the browser, hence, since this varies, the height of the image/video must vary proportionally. Since the height may grow quite small when the browser becomes narrower, we should shift the black gradient to  start much lower.
- When the video is playing, there must be a mute button for the audio.
- The content we will feature is this video https://www.youtube.com/watch?v=kjtKrJarrTw. Playback will start at 00:44s
- The cover image is attached (ask internally).
- The brand (SCHISM logo) is attached (ask internally).
- Content brand is displayed above title.

kdembler:
Resolves #1627

Adds a playback to the cover featured video. It will start autoplaying muted once loaded and can be controlled with pause/mute buttons.


atlas/src/shared/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -5,31 +5,90 @@ import { useVideoJsPlayer, VideoJsConfig } from './videoJsPlayer'
...
+    }
+  }, [player, initialized, autoplay])
+
+  useEffect(() => {
+    if (!player) {
+      return
+    }
+
+    if (playing != null) {
+      if (playing) {
```
fcescob, line 80: I think this second if is not necessary given the one above

atlas/src/shared/components/VideoPlayer/videoJsPlayer.ts
```diff
@@ -76,5 +78,21 @@ export const useVideoJsPlayer: VideoJsPlayerHook = ({ fill, fluid, height, src,
...
+
+    player.muted(muted)
+  }, [player, muted])
+
+  useEffect(() => {
+    if (!player || !posterUrl) {
+      return
+    }
+
+    player.poster(posterUrl)
```
fcescob, line 94: Since this is setting a poster, I think a `useLayoutEffect` would be more effective here. wdyt?

atlas/src/shared/components/VideoPlayer/VideoPlayer.tsx
kdembler: First `if` is meant only to check if the value is `undefined` or `null`, the second one will check for `true` / `false`

atlas/src/shared/components/VideoPlayer/videoJsPlayer.ts
kdembler: I don't think it's needed. This would ensure that we can read DOM and synchronously re-render, but it works fine as it is

atlas/src/shared/components/VideoPlayer/VideoPlayer.tsx
fcescob: I see, lets leave it as is then.

atlas/src/shared/components/VideoPlayer/videoJsPlayer.ts
fcescob: I can't find any actual problems with rendering, so I'm fine to keep this as is for now.

fcescob (approved):

Issue #1681, opened 2020-11-11
Content directory smart contracts: Bridge contracts initial tests


Lezek123:
At the beginning we can use Ganache to run the tests.

This would mean we would use ethereum rpc interface, so we obviously aren't able to test integration with a substrate runtime, but we can do some basic tests to make sure that the main contract logic works as expected

Eventually we will want to have tests that use the substrate api to intract with the contracts and be also able to verify if the intreactions between substrate runtime and evm work as expected. At this point we can decide whether we should maintain two separate sets of tests (unit and integration) and whether unit tests should rely on eth-compatible rpc interface or use the substrate api.

## Bridge contracts

For initial tests we will use some predefined key that will act as a "runtime key".

We will also need to preserve some state of bridge contracts in order to be able to succesfully perform other tests (channels, videos, curator groups etc.), as they will rely on bridge contracts state to validate sender permissions.

### Initialization

**Batch insert** 
- success case (verify storage changed as expected)
- access denied case (wrong key)
- override case (verify batch insert does not override existing data)

**Finish initialization**
- access denied case
- already finished case
- success case (verify `batchInsert` can no longer be used)

### Hooks

**Memberships - addMemberControllerAddress**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- member address already set case (check if event is emitted)

**Memberships - updateMemberControllerAddress**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- member address not set case (check if event is emitted)

**ContentWorkingGroup - setLeadAddress** 
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- address already set case (check if event is emitted)

**ContentWorkingGroup - updateLeadAddress, removeLeadAddress** 
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- lead not set case (check if event is emitted)

**ContentWorkingGroup - addCuratorAddress**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- curator address already set case (check if event is emitted)

**ContentWorkingGroup - updateCuratorAddress, removeCuratorAddress**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- curator not set case (check if event is emitted)

**ContentWorkingGroup - disableLead**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
- already disabled case
- optionally: no lead set case

**ContentWorkingGroup - enableLead**
- success case (verify storage changed accordingly)
- access denied case (wrong key)
[...]

Issue #1520, opened 2020-10-12
Redux: Requirements for smart contract content directory


bedeho:
# Redux: Requirements for smart contract content directory

## Problem

Our current content directory has a bespoke, complex and arbitrary write access model that has multiple independent strands in order to model ownership, relationship constraints and ration resource utilisation. It is a very inflexible model in terms of what functionality it support, and increasing this scope requires compounding the already disorganised write access model.

## Proposal

This proposal for requirements for a smart contract based content directory rests on the analysis here

https://github.com/Joystream/joystream/issues/1518

and the assumption that the, yet to be conducted, benchmarking analysis finds that it is practical and safe to accept the need for migrations in rare cases.

### Note

This is not a specification, it is a draft for a set of requirements for a specification.

### Prior Approach

There was an exploratory attempt at sketching out an implementation for a smart contract based content directory here

https://github.com/Joystream/joystream/issues/368

The outcome was evaluated to be quite positive, however more recent lessons suggest that even this evaluation was too pessimistic, specifically

- **_MUCH_ smaller migrations:** Almost no data needs to be stored, now that we have a query node, at worst there are some hash commitments, but even those are of unknown utility, and there there are genuine business logic values that need to be read back later. All large fields, like human readable text blobs, URLs, structured metadata, is not stored. This means state to migrate is orders of magnitude smaller than before.
- **Long running migrations:** Migrations can run over more than a single block when state mutexed properly. This was always the case, it was just not clear in my mind. This means large migrations are possible, as long locking of system is acceptable.
- **Less frequent migrations:** Most changes to the write logic require no migration at all, only deploying new contracts at worst, because most changes will not change how they read or write from storage, or they will read objects in a backwards compatible way.

#### EVM

We are using the EVM module because it has the best supporting tooling and the broadest developer adoption at this time. One important limitation is worth keeping in mind when using the `EVM` FRAME module, which actually also applies for the `Contract` FRAME module. This limitation is that it is not possible to make any call, synchronous or asynchronous, from the smart contract environment and into the rest of the runtime. In the same vein it is not possible to read or write to external storage. However, it is possible to make the following synchronous calls into the smart contract environment from the runtime:

- `Call`: Call a method on a contract.
- `Deploy`: Deploy a new contract.
- `Deposit`: Deposit funds from a `balances` account to a smart contract account.
- `Withdraw`: Withdraw funds from a smart contract account into a `balances` account.

[...]

PR #1752, opened 2020-11-18
Content directory - initial smart contracts

Lezek123:
This PR includes the initial code of new content directory Solidity smart contracts.

The implementation is based on https://github.com/Joystream/joystream/issues/1520, the initial design was described in https://github.com/Joystream/joystream/issues/1520#issuecomment-724869225

I also icluded a few simple tests that can be run against a Ganache node:
```
cd smart-contracts
yarn
yarn start
yarn test
```

Of course there is also a possibility to run any other Truffle command, ie.:
```
yarn truffle deploy
```

Currently we rely on Ganache to run the tests, but once the evm pallet is enabled in the runtime it would be possible to run some test against the actual Joystream node (this was further described here: https://github.com/Joystream/joystream/issues/1681)

**Covered:**
- very simple bridge contracts allowing only the "runtime" (which in test scenarios is just a random address generated by the Ganache node) to set members / content working group data
- storage contracts for channels, videos and curator groups
- `ContentDirectory.sol` contract currently containing all the logic related to managing channels, curator groups and videos. It probably needs to be split into separate contracts / libraries as described in the _Things to consider_ section below
- a few example tests to run against a Ganache node

**Not yet covered:**
- storage contracts interfaces
- upgradability / migration / initialization logic (there is a `migrate` method in `ContentDirectory.sol` but it's not yet fully functional)
- full testing scenarios

**Things to consider:**
- the methods that can be executed both by a curator and member (ie. updateChannelMetadata) have kind of awkward "overloads", since we additionaly need `_curatorId` as context in case the method is executed by curator. There are a few alternatives, but each of them has some drawbacks:
  - make `_curatorId` arg part of the original methods like `updateChannelMetadata`, `removeVideo` - in this case this value could just be ignored on the smart-contract side if it's not needed (ie. sender has owner access to channel), but it would have to still be provided on the frontend (where it can be just set to `0` if  it's not needed). This saves us some code in the smart contract, but seems like a bit less "cleaner" approach.
  - add separate, reverse `curatorIdsByAddress` map in `ContentWorkingGroupBridge` - we could then derive the curator context from the `msg.sender` address, but it would require a loop and could potentially cost a lot of gas in case the address controlls many curators
  - add an argument similar to current content directory `Actor` to all content directory contract methods - this wouldn't look very good in Solidity due to lack of enums that can hold additional values depending on the variant. It also adds complexity on the frontend. 
  - add `Actor` argument to `evm.call` extrinsic in the runtime and use special `msg.sender` address format for members/curators - this would also allow us to get rid of bridge-contracts, but at the expense of not beeing able to verify whether a curator / member id is valid from insde of the evm environement (we would only know that the current `msg.sender` is valid curator/member)
- reducing the size of `ContentDirectory.sol` - currently it's a very big contract that costs a lot of gas to deploy (exceeds the default gas limit) and requires compiler optimalization specifically for deployment in order to prevent other issues. It needs to be split into libraries / other contracts or optimalized in some other way.
- non-metadata fields of channels/videos - we may want to store some of the `Channel` / `Video` fields like `handle` or  `contentId` to validate whether they are unique, prevent them from beeing updated etc. There is also an open question w.r.t. how to handle the metadata that is clearly invalid (and in which case, if any, it should be handled on the smart contracts side)


bedeho:
Here is my response to your notes

# Not yet covered

Don't do any of this except tests, the upgradebility can wait, and it requires a deeper review of alternatives.

# Things to consider

1. I think each approach had pretty nasty costs, what about just separating out different methods for each caller? so `FooAsOwner` and `FooAsCurator`, it's not as nice as what we can do in Rust, but it seems better than any of the alternatives listed.
2. reducing the size of `ContentDirectory.sol`: My impression is that, while there are many methods, it's all relatively simple stuff, and the state is very simple. If that is the case, we can just keep going forward and refactor later to find the best way to organize things, often this is only clear at a later stage.
3. non-metadata fields: Yes, if they genuinely are not just normal metadat fields, then they have to be lifted out. Unique handle would be a proper example. Its not obvious that contentid has to be one of those, and also the way we encode content source will change in the future when we update things on the storage side, so I would keep it as is. As a general rule, lets agree on a case by case basis before anything is lifted out.

smart-contracts/contracts/ContentDirectory.sol
```diff
@@ -0,0 +1,475 @@
...
+    }
+    Ownable oldStorage = Ownable(_oldStorageAddress);
+    oldStorage.transferOwnership(_newLogicAddress);
+  }
+
+  // CHANNELS
+  using ChannelOwnershipDecoder for ChannelOwnership;
+
+  // Access/validation utils:
+  function _validateOwnership(ChannelOwnership memory _ownership) internal view returns (bool) {
```
ondratra, line 175: This function is never used.

smart-contracts/contracts/storage/VideoStorage.sol
```diff
@@ -0,0 +1,57 @@
...
+  bool isActive;
+  bool isExisting;
+}
+
+contract VideoStorage is Ownable {
+  using SafeMath64 for uint64;
+
+  mapping(uint64 => Video) private videoById;
+  mapping(uint64 => uint64) public videoCountByChannelId;
+  uint64 nextVideoId = 1;
```
ondratra, line 20: Is there a reason why to use a 1-based index here? It seems to me that a regular 0-based index is used in the content directory pallet. 

This topic is relevant to `nextChannelId`, `nextGroupId`, and `nextVideoId`.

smart-contracts/contracts/storage/ChannelStorage.sol
```diff
@@ -0,0 +1,132 @@
...
+  ChannelOwnership ownership;
+  bool isActive;
+  uint32 videoLimit; // 0 = use default
+  bool isExisting;
+}
+
+// A helper library to parse ChannelOwnership.
+// New ownership types can be added if needed without the need for migration
+// (but changing/removing existing ones would still require migration to new storage)
+enum ChannelOwnerType {Address, Member, CuratorGroup}
```
ondratra, line 29: Not a problem on its own, but we should be careful when using enums in multiple contracts as there are non-intuitive issues while upgrading contracts. See https://hackernoon.com/beware-the-solidity-enums-9v1qa31b2 .

Either we should limit their existence to only one contract or write a `WARNING` commentary next to each such enum to prevent this kind of error as it might not be revealed by regular tests (as stated in the article).

smart-contracts/contracts/ContentDirectory.sol
```diff
@@ -0,0 +1,475 @@
...
+  // Limits change events
+  event ChannelInstancesLimitUpdated(uint64 _newLimit);
+  event VideosPerChannelDefaultLimitUpdated(uint32 _newLimit);
+  event VideosPerChannelMaxLimitUpdated(uint32 _newLimit);
+
+  // Migration event
+  event Migrated(address _logic, address _videoStorage, address _channelStorage, address _curatorGroupStorage);
+
+  // Channel-related events
+  event ChannelCreated(uint64 _id, ChannelOwnership _ownership, string[2][] _metadata);
```
ondratra, line 50: What `string[2][]` actually represents? I'm a little bit puzzled on what is the motivation behind the `[2][]` part (even number of characters?). Could we replace it with struct so we can use a more expressive data type name like `Metadata` through the code?

smart-contracts/contracts/ContentDirectory.sol
```diff
@@ -0,0 +1,475 @@
...
+    oldStorage.transferOwnership(_newLogicAddress);
+  }
+
+  // CHANNELS
+  using ChannelOwnershipDecoder for ChannelOwnership;
+
+  // Access/validation utils:
+  function _validateOwnership(ChannelOwnership memory _ownership) internal view returns (bool) {
+    require(_ownership.isValid(), "Invalid ownership data");
+    if (_ownership.isMember()) {
```
ondratra, line 177: Since `_ownership.isMember() != _ownership.isCuratorGroup()` is always true, we can return right after `require` check
```
if (_ownership.isMember()) {
    require(...);
    return;
}
```

smart-contracts/contracts/ContentDirectory.sol
```diff
@@ -0,0 +1,475 @@
...
+      require(membershipBridge.memberExists(_ownership.asMember()), "Member ownership - member does not exist!");
+    }
+    if (_ownership.isCuratorGroup()) {
+      require(
+        curatorGroupStorage.groupExists(_ownership.asCuratorGroup()),
+        "CuratorGroup ownership - group does not exist!"
+      );
+    }
```
ondratra, line 185: Is it ok that the check for `.isAddress()` is not here with the rest of the ownership checks?

smart-contracts/contracts/lib/SafeMath16.sol
```diff
@@ -0,0 +1,158 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.6.0;
+
+// uint16 version of OpenZeppelin's SafeMath
+library SafeMath16 {
```
ondratra, line 5: Instead of type-specific implementations of `SafeMath` (`SafeMath16`, `SafeMath32`, ...) not originating from OpenZeppelin, we should use `SafeMath`(256bit) in combination with OpenZeppelin's `SafeCast`. Maintaining multiple math libraries that are not supported by `OpenZeppelin` would create for us an unnecessary maintenance burden and security weak point, especially when (eventually) updating the Solidity version. See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1625#issuecomment-545978860

smart-contracts/contracts/ContentDirectory.sol
```diff
@@ -0,0 +1,475 @@
...
+  }
+
+  function _hasOwnerAccess(address _address, ChannelOwnership memory _ownership) internal view returns (bool) {
+    if (_ownership.isAddress()) {
```
ondratra, line 189: Just a code style note for your consideration. As the Object Calisthenics recommends, it is good to evade writing `else` and `else if` and instead of such branching style use *process of elimination* style instead.

```
if (_ownership.isAddress()) {
   return _address == _ownership.asAddress();
}

// easily commentable
if (_ownership.isMember()) {
  return membershipBridge.isMemberController(_address, _ownership.asMember());
}
...
```

ondratra (commented):
A very nice kickoff on the whole thing, Leszek! Some notes:

1) We should use `int256` and `uint256` instead of a smaller number types by default and use smaller types only when it brings some concrete advantage. In our use case, (only?) Rust/Solidity interface and events should share a smaller number type. In Ethereum, 256bit number is default *Word* and downscaling it costs extra gas - storing `uint8` costs more than storing `uint256`, the same goes for calculations. This is one of the (possibly significant) causes of too expensive contract deployment (~6871595 gas).

2) In general, it might be a good idea to split the smart contract into multiple contract definitions that each focus on individual aspects. It will make code inspection easier as the contract gets bigger. In our case, `ContentDirectory` contract could be split into one part focusing on council-controlled mechanics (`onlyCouncil` functions), second focusing on channels, third on videos, etc.; and then joined via `contract ContentDirectory is ContentDirectoryCouncil, ContentDirectoryChannels, ... {`. In a sense, this is mentioned in the original PR description.

3) Right now, events are emitted from the main `ContentDirectory` contract, and storage is saved in a changeable contract(s). One small con of this is the lack of event emits in the storage - we can't simply walkthrough/audit storage operations using Ethereum logs. It might also create confusion/problems when a new storage contract is set - events may not necessarily correlate to storage contents. Another thing we need to keep in mind when doing the next iteration on upgradability is that when new storage is introduced, it needs to somehow include the old storage content in its inner workings because copying the whole storage content to the new contract will be expensive. We should definitely create two versions of storage contracts and a test for an actual migration and afterward querying data before the release.

smart-contracts/contracts/ContentDirectory.sol
Lezek123: It's a dynamic-size array of arrays containing exactly 2 strings. Initially it was meant to represent cusomizable metadata with dynamic `key-value` pairs, ie.:
```
[
  [ 'title', 'Some video title' ],
  [ 'description', 'Some video description' ]
]
```
But since this is not very flexible, in https://github.com/Joystream/joystream/issues/1816 I explored an alternative possiblity to replace it with just a `string` containing `json` metadata representation (I think that's probably the best way to solve this if the contracts don't need to care about metadata validity and structure)

smart-contracts/contracts/storage/VideoStorage.sol
Lezek123: I'm used to 1-based index when it comes to "entity ids", since that's usually how ids are assigned in `MySQL`/`PostgresSQL` databases etc. (and since query node uses PostgresSQL database it seems to make sense).

Also it looks like currently we're using 1-based index in the runtime too: https://github.com/Joystream/joystream/blob/4865e9b3fc436cb0878778922bfc1cb3689d1acc/node/src/chain_spec/content_config.rs#L238

smart-contracts/contracts/ContentDirectory.sol
Lezek123: Nicely spotted! It was supposed to be used instead of `ownership.isValid` in `createChannel`/`updateChannelOwnership`. I'll fix it and add a related test case

smart-contracts/contracts/ContentDirectory.sol
Lezek123: In case of membership/group we can additionaly check if it exists, but with address I don't think we need any additional checks (except maybe disallowing zero-address)

smart-contracts/contracts/lib/SafeMath16.sol
Lezek123: I stumbled into this comment, but it seemed to add a lot of the burden to peform casting after every operation, but since this is kind of related to the next comment, I'll address it there.

Lezek123: I stumbled upon this link when trying to figure out the best solution, but it seemed to add a lot of burden to perform this casting after every operation. Since this is related to the general `u256` comment, I'll address it further below.

Lezek123:
> we should use int256 and uint256 instead of a smaller number types by default and use smaller types only when it brings some concrete advantage

After thinking it through I agree.

At the beginning I could see a few benefits of using smaller types, ie.:
- Smaller storage cost when used inside `Struct`s
- Possibility to easily limit some numbers to predictable size
- The types are more-or-less compatible with the current Rust runtime implementation

But considering the major drawbacks on the other side, ie. the need to either use custom `SafeMath` or "manually" cast integers with `SafeCast` after each operation + the additional costs when used outside `Stucts` (which as you said may be one of the causes of very high deployment cost) , probably it's better to just enforce min/max value when necessary and avoid using smaller `int`s/`uint`s

Issue #1355, opened 2020-09-15
Removing dangling InputValidationLengthConstraint from fourm


bedeho:
SSIA

iorveth:
Closed by https://github.com/Joystream/joystream/pull/1755

Issue #1314, opened 2020-09-13
Polls have no limit on number of alternatives


bedeho:
A hardcoded, module configurable parameter, low value should be introduced.

iorveth:
cc @bedeho  The limit for number of alternatives is currently ensured by `PollItemsConstraint` runtime storage value.
Did you mean, we should make it configurable on Trait level and only have lower bound?

iorveth:
Close by https://github.com/Joystream/joystream/pull/1755

PR #1753, opened 2020-11-18
Voting logic refactoring

iorveth:
Fix for https://github.com/Joystream/joystream/issues/1315

Based on https://github.com/Joystream/joystream/pull/1675

Issue #1346, opened 2020-09-14
Misleading mutation-safe marker in forum extrinsics


bedeho:
In quite a few extrinsics, we have this type of a situation

```Rust
//
// == MUTATION SAFE ==
//

/* eeeeeek! what about this space, seems safe to mutate, but is not! */

Self::some_call_with_sideffects_when_OK(...)?;
```

The line with the comment appears like a safe place to mutate, but it is not. The problem is that we have a call which both checks and mutates. This should be avoided when possible, and it almost always is.

Some examples are 
- `create_thread`
- `add_post`

possibly there are more.

iorveth:
Closed by https://github.com/Joystream/joystream/pull/1756

PR #1755, opened 2020-11-18
Feature/max poll alternatives number

iorveth:
Closes https://github.com/Joystream/joystream/issues/1355
Closes https://github.com/Joystream/joystream/issues/1314
Closes https://github.com/Joystream/joystream/issues/1758
Refactors `create_thread`, as suggested in https://github.com/Joystream/joystream/issues/1346
Based on https://github.com/Joystream/joystream/pull/1753


bedeho:
This has a merge conflict.

iorveth:
Fixed

runtime-modules/forum/src/lib.rs
```diff
@@ -1038,21 +1034,32 @@ decl_module! {
         }
 
         /// Edit post text
+        // TODO need a safer approach for frame_system call
```
bedeho: Why was this comment added here? I do not understand it, also the heading is "EDit post text" , which does not match the extrinsic.

bedeho (changes requested):

runtime-modules/forum/src/lib.rs
iorveth: Just moved it from the inner function, performing actual mutation. Seems, that that naming mismatch was here before, i can fix.

bedeho (approved):

Issue #1711, opened 2020-11-16
Carousel improvements


kdembler:
- The left most preview gets clipped when hovered<br>
![](https://user-images.githubusercontent.com/437292/98926719-f613e680-24d7-11eb-8085-61ecf3ad1ae8.png)

- The carousel should rotate out full row, not one item at a time

Issue #1708, opened 2020-11-16
Carousel buttons improvements 


kdembler:
There's a couple of issues surrounding the carousel next/previous buttons:

- at some point we've removed the feature that allowed setting the alignment of the buttons based on the carousel content height, we should bring that back so we can always vertically center the buttons according to the design
- the buttons are also misaligned horizontally - their outward vertical edge should align with the video previews edges
<br>![](https://user-images.githubusercontent.com/437292/98939204-5bbc9e80-24e9-11eb-9071-768c73858769.png)

PR #1757, opened 2020-11-19
Carousel Improvements 

### LINKED ISSUE #1707: Video preview carousel improvements at max columns
The resizing policy on videos still breaks when max item cap (6) is reached, as margins start growing

Seems we need to make sure video previews don't have a `max-width`

fcescob:
This PR fixes #1707 #1711 and the aligns the buttons to the middle of a Video Cover as specified in #1708

### DIFF
--- a/atlas/src/components/VideoGallery.tsx
+++ b/atlas/src/components/VideoGallery.tsx
@@ -15,7 +15,7 @@ type VideoGalleryProps = {
 
 const PLACEHOLDERS_COUNT = 12
 const CAROUSEL_ARROW_HEIGHT = 48
-const trackPadding = `${sizes.b2}px 0 0 0`
+const trackPadding = `${sizes.b2}px 0 0 ${sizes.b2}px`
 
 // This is needed since Gliderjs and the Grid have different resizing policies
 const breakpoints = breakpointsOfGrid({
@@ -27,6 +27,7 @@ const breakpoints = breakpointsOfGrid({
   breakpoint,
   settings: {
     slidesToShow: idx + 1,
+    slidesToScroll: idx + 1,
   },
 }))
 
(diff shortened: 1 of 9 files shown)


atlas/src/components/VideoGallery.tsx
```diff
@@ -16,8 +14,8 @@ type VideoGalleryProps = {
 }
 
 const PLACEHOLDERS_COUNT = 12
-
-const trackPadding = `${spacing.xs} 0 0 0`
+const CAROUSEL_ARROW_HEIGHT = 48
```
kdembler: This should be imported from wherever the button height is defined

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
```diff
@@ -74,6 +74,17 @@ const VideoPreview: React.FC<VideoPreviewProps> = ({
     },
   })
 
+  useLayoutEffect(() => {
```
kdembler: We can use resize observer with a height changed callback instead of that

atlas/src/shared/components/Carousel/Carousel.tsx
```diff
@@ -7,12 +8,23 @@ import { Container, GliderContainer, Arrow, Track, BackgroundGradient } from './
 type CarouselProps = {
   trackPadding?: string
   className?: string
+  arrowCss?: SerializedStyles
 } & GliderProps
 
+const trackPaddingToMargin = (padding: string) =>
```
kdembler: Let's just pass numerical values for top and left padding so we don't have to parse a string

kdembler (commented):
Ah forgot to post these after our call. Submitting the review now so these items don't get lost, will close them if already done

atlas/src/shared/components/Carousel/Carousel.tsx
kdembler: Done

atlas/src/shared/components/VideoPreview/VideoPreview.tsx
kdembler: Done

atlas/src/components/VideoGallery.tsx
kdembler: Done

kdembler (approved):
LGTM!

Issue #1487, opened 2020-09-29
Integrated budget constraints


bedeho:
# Background

Funds are minted and spent for lots of different purposes that needs to be constrained

a. funding proposals
b. council member rewards
c. worker rewards
d. working group discretionary spending
e. channel rewards (WIP)

Currently, there the spending constraints around these are not integrated.

# Proposal

There is _one_ single top level budget. This budget is updated whenever

1. council member rewards are paid: budfgt reduced
2. working group budgets are changed: budget is reduced if group budget is increaed, otherwise it is increased
3. funding proposals are paid out: budfgt reduced

This budget is increased by a given amount X (storage parameter) every certain blocks Y (hard coded constant) blocks.
Whenever actions 1-3 are attempted, they will fail if it requires more resources than the budget. If a working group budget is reduced, then the budget is actually increased.

The value X can be changed with a proposal, but only within certain hard coded bounds. Notice that when spending occurs within working groups (c,d above), that does nothing, because that is out of a separate budget.

We could also have a cutoff where budget cannot increase beyond a certain level L, but that is not important.

PR #1760, opened 2020-11-19
Council spending budget2

ondratra:
This PR adds elected member rewards budget to the council (part of #1487). A minor typo and compile bugs are also included.

Budget balance can be set from an external source and will be periodically paid to currently elected council members. In case reward(s) can't be paid in full, the remaining amount(s) will be remembered and paid next time. Any unpaid and at the moment, unpayable rewards will be discarded when the new council is established.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -121,21 +121,38 @@ pub struct Candidate<AccountId, Balance, Hash> {
 /// Council member representation.
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, PartialEq, Eq, Debug, Default, Clone)]
-pub struct CouncilMember<AccountId, MembershipId, Balance> {
+pub struct CouncilMember<AccountId, MembershipId, Balance, BlockNumber> {
     staking_account_id: AccountId,
     membership_id: MembershipId,
     stake: Balance,
+
+    last_payment_block: BlockNumber,
+    unpaid_reward: Balance,
 }
 
-impl<AccountId, MembershipId, Balance, Hash>
-    From<(Candidate<AccountId, Balance, Hash>, MembershipId)>
-    for CouncilMember<AccountId, MembershipId, Balance>
+impl<AccountId, MembershipId, Balance, Hash, BlockNumber>
+    From<(
+        Candidate<AccountId, Balance, Hash>,
+        MembershipId,
+        BlockNumber,
+        Balance,
+    )> for CouncilMember<AccountId, MembershipId, Balance, BlockNumber>
 {
-    fn from(candidate_and_user_id: (Candidate<AccountId, Balance, Hash>, MembershipId)) -> Self {
+    fn from(
+        from: (
+            Candidate<AccountId, Balance, Hash>,
+            MembershipId,
+            BlockNumber,
+            Balance,
+        ),
+    ) -> Self {
         Self {
-            staking_account_id: candidate_and_user_id.0.staking_account_id,
-            membership_id: candidate_and_user_id.1,
-            stake: candidate_and_user_id.0.stake,
+            staking_account_id: from.0.staking_account_id,
+            membership_id: from.1,
+            stake: from.0.stake,
+
+            last_payment_block: from.2,
+            unpaid_reward: from.3,
         }
     }
 }
@@ -153,8 +170,12 @@ pub type VotePowerOf<T> = <<T as Trait>::Referendum as ReferendumManager<
     <T as system::Trait>::Hash,
 >>::VotePower;
 
-pub type CouncilMemberOf<T> =
-    CouncilMember<<T as system::Trait>::AccountId, <T as Trait>::MembershipId, Balance<T>>;
+pub type CouncilMemberOf<T> = CouncilMember<
+    <T as system::Trait>::AccountId,
+    <T as Trait>::MembershipId,
+    Balance<T>,
+    <T as system::Trait>::BlockNumber,
+>;
 pub type CandidateOf<T> =
     Candidate<<T as system::Trait>::AccountId, Balance<T>, <T as system::Trait>::Hash>;
 pub type CouncilStageUpdateOf<T> = CouncilStageUpdate<<T as system::Trait>::BlockNumber>;
@@ -198,11 +219,21 @@ pub trait Trait: system::Trait {
     /// Duration of idle period
     type IdlePeriodDuration: Get<Self::BlockNumber>;
 
+    /// The value elected members will be awarded each block of their reign.
+    type ElectedMemberRewardPerBlock: Get<Balance<Self>>;
+    /// Interval for automatic reward payments.
+    type ElectedMemberRewardPeriod: Get<Self::BlockNumber>;
+
     /// Checks that the user account is indeed associated with the member.
     fn is_council_member_account(
         membership_id: &Self::MembershipId,
         account_id: &<Self as system::Trait>::AccountId,
     ) -> bool;
+
+    /// This function is needed because there is no way to enforce `Trait::BlockNumber: From<Balance<T>> + Into<Balance<T>>`
+    /// needed by `get_current_reward()`. That's because `where` clause is not expected inside of `decl_error`
+    /// (and there might be other problems).
+    fn blocks_to_balance(block_number: &Self::BlockNumber) -> Balance<Self>;
 }
 
 /// Trait with functions that MUST be called by the runtime with values received from the referendum module.
@@ -230,6 +261,12 @@ decl_storage! {
 
         /// Index of the current candidacy period. It is incremented everytime announcement period starts.
         pub AnnouncementPeriodNr get(fn announcement_period_nr) config(): u64;
+
+        /// Budget for the council's elected members rewards.
+        pub Budget get(fn budget): Balance<T>;
+
+        /// The next block in which the budget will be increased.
+        pub NextBudgetRefill get(fn next_budget_refill): T::BlockNumber;
     }
 }
 
@@ -238,6 +275,7 @@ decl_event! {
     where
         Balance = Balance::<T>,
         <T as Trait>::MembershipId,
+        <T as system::Trait>::AccountId,
     {
         /// New council was elected
         AnnouncingPeriodStarted(),
@@ -265,6 +303,12 @@ decl_event! {
 
         /// The candidate has set a new note for their candidacy
         CandidacyNoteSet(MembershipId, Vec<u8>),
+
+        /// The whole reward was paid to the council member.
+        RewardPayment(MembershipId, AccountId),
+
+        /// The reward was paid to the council member only partially.
+        RewardPartialPayment(MembershipId, AccountId),
     }
 }
 
@@ -348,6 +392,8 @@ decl_module! {
         fn on_finalize(now: T::BlockNumber) {
             Self::try_progress_stage(now);
         }
+
+        /////////////////// Election-related ///////////////////////////////////
 
         /// Subscribe candidate
         #[weight = 10_000_000]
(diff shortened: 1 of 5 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -455,6 +528,11 @@ impl<T: Trait> Module<T> {
             }
             _ => (),
         }
+
+        // budget reward progress
+        if now == NextBudgetRefill::<T>::get() {
```
bedeho, line 564: This should not live in `try_progress_stage`, because it actually unrelated to current stage or change in stage, it should just be in `on_finalize`, or some new method invoked from there.

runtime-modules/council/src/lib.rs
```diff
@@ -455,6 +528,11 @@ impl<T: Trait> Module<T> {
...
+            Self::pay_elected_member_rewards(now);
```
bedeho, line 570: Why are you paying people when the budget is to be refilled? These are totally independent events.

runtime-modules/council/src/lib.rs
```diff
@@ -527,6 +607,68 @@ impl<T: Trait> Module<T> {
...
+            }
+
+            // calculate withdrawable balance
+            let (available_balance, missing_balance) =
+                Calculations::<T>::payable_reward(&new_balance, &unpaid_reward);
+
+            // pay reward
+            Mutations::<T>::pay_reward(
+                member_index,
+                &council_member.staking_account_id,
```
bedeho: Lets introduce a separate reward account for council members, just like in the working group worker role.

runtime-modules/council/src/lib.rs
```diff
@@ -198,18 +220,29 @@ pub trait Trait: system::Trait {
...
         membership_id: &Self::MembershipId,
         account_id: &<Self as system::Trait>::AccountId,
     ) -> bool;
+
+    /// Converts block number type to balance type.
+    /// NOTE: This function is needed because there is no way to enforce type constraint
+    ///       `Trait::BlockNumber: From<Balance<T>> + Into<Balance<T>>` needed by `get_current_reward()`.
+    ///       That's because cascade of `where` clauses leads to `decl_error` macro that doesn't support
+    ///       type constraint via `where`.
+    fn blocks_to_balance(block_number: &Self::BlockNumber) -> Balance<Self>;
```
bedeho: I don't think should be required, have a look here

https://github.com/Joystream/joystream/blob/working_team_pallet/runtime-modules/working-team/src/lib.rs#L966

runtime-modules/council/src/lib.rs
```diff
@@ -706,6 +876,53 @@ impl<T: Trait> Mutations<T> {
     fn set_candidacy_note(membership_id: &T::MembershipId, note_hash: &T::Hash) {
         Candidates::<T>::mutate(membership_id, |value| value.note_hash = Some(*note_hash));
     }
+
+    /// Removes member's candidacy record.
+    fn clear_candidate(membership_id: &T::MembershipId) {
+        // clear candidate record
+        Candidates::<T>::remove(membership_id);
```
bedeho, line 948: It feels a bit risky to drop the candidates but leave the lock removal until later in the code path in `elect_new_council`. Perhaps its a bit tidier to move lock removal in here, and then that new method just applies the council lock?

runtime-modules/council/src/lib.rs
```diff
@@ -579,13 +717,49 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
+        budget_balance: &Balance<T>,
+        reward_amount: &Balance<T>,
+    ) -> (Balance<T>, Balance<T>) {
+        // check if reward has enough balance
+        if reward_amount <= budget_balance {
+            return (*reward_amount, 0.into());
+        }
+
+        // calculate missing balance
+        let missing_balance = *reward_amount - *budget_balance;
```
bedeho: Is this safe, I think perhaps saturated arithmetic is required, this applies to a few other places also. Please consult @shamil-gadelshin about this.

runtime-modules/council/src/lib.rs
```diff
@@ -579,13 +717,49 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
+
+impl<T: Trait> Calculations<T> {
+    /// Calculate current reward for the recipient.
+    fn get_current_reward(
+        council_member: &CouncilMemberOf<T>,
+        reward_per_block: Balance<T>,
+        now: T::BlockNumber,
+    ) -> Balance<T> {
+        council_member.unpaid_reward
+            + T::blocks_to_balance(&(now - council_member.last_payment_block)) * reward_per_block
```
bedeho: Same issue here about safety, please consult @shamil-gadelshin .

runtime-modules/council/src/lib.rs
```diff
@@ -527,6 +607,68 @@ impl<T: Trait> Module<T> {
...
+                &available_balance,
+                &missing_balance,
+                &now,
+            );
+
+            // remember new balance
+            new_balance -= available_balance;
+
+            // emit event
+            if missing_balance > 0.into() {
```
bedeho: If you were unable to pay out the current `council_member`, aren't all payouts now impossible because `new_balance == 0`? If this is true, then
a) lets make the conditional on `new_balance`.
b) I think the general iteration should stop, but we should still iterate over the remaining council members and emit events saying they did not get paid at all and what their current owed amount is. One way to do this is to just introduce a first `if` inside the loop which automatically sidesteps all the member specific calculations which are known to fail. Also see my comment on suggested event changes. Let me know if you have any cleaner suggestion here.

runtime-modules/council/src/lib.rs
```diff
@@ -266,6 +306,15 @@ decl_event! {
 
         /// The candidate has set a new note for their candidacy
         CandidacyNoteSet(MembershipId, Vec<u8>),
+
+        /// The whole reward was paid to the council member.
+        RewardPayment(MembershipId, AccountId),
```
bedeho: Can we include what was paid out and what new debt is for this member, that way we don't need a separate event for partial or missing payment.

runtime-modules/council/src/lib.rs
```diff
@@ -527,6 +607,68 @@ impl<T: Trait> Module<T> {
...
     }
 
+    /////////////////// Budget-related /////////////////////////////////////
+
+    /// pay rewards to elected council members
+    fn pay_elected_member_rewards(now: T::BlockNumber) {
+        let reward_per_block = T::ElectedMemberRewardPerBlock::get();
+
+        // intermediate balance
+        let mut new_balance = Budget::<T>::get();
```
bedeho: I think you can sidestep this mutable with a map+reduce type pattern here, but I think we can skip it for now, this feels relatively manageable.

bedeho (changes requested):
Almost there, I think the most important required change here which I have not looked carefully at is the lack of safer arithmetic and possibly comparisons, ask @shamil-gadelshin about his approach, we should have some consistency. He could also review this PR for good measure after you apply some fix to this.

runtime-modules/council/src/lib.rs
ondratra: Done

runtime-modules/council/src/lib.rs
ondratra: Name mismatch. `NextBudgetRefill` was treated as `NextRewardPayments` as it should, but had the wrong name. Name and comments were changed.

runtime-modules/council/src/lib.rs
ondratra: Yes, that's probably better. Done.

runtime-modules/council/src/lib.rs
ondratra: I usually try to get rid of side-effects, but the reason I loop here via `for` instead of reducing the vector was the easy use of circuit-shorting (`break` out of `for`) when budget is completely depleted.

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: Removed; problem solved itself after saturated arithmetic was introduced.

runtime-modules/council/src/lib.rs
ondratra: An event is now emitted for all council members. Only one event, `RewardPayment`, with info about how much was paid off and how much unpaid reward exists now.

runtime-modules/council/src/lib.rs
ondratra: In the end, `for` loop was removed in favor of `fold()`(reduce) due to rework mention here https://github.com/Joystream/joystream/pull/1760#discussion_r527630479 .

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs

runtime-modules/council/src/lib.rs
ondratra: Done. Just a note: `reward_account_id` must be selected when announcing candidacy.

bedeho: Yes correct.

runtime-modules/council/src/lib.rs
bedeho: But when does budget refill happen then? This needs to also happen at a separate interval, as the proposal says

`This budget is increased by a given amount X (storage parameter) every certain blocks Y (hard coded constant) blocks.`

bedeho (commented):
Looks good to me, and very nice work on the quick turnaround. I just made the one comment still about auto refill of budget being needed. I have also pinged Shamil for extra review.

bedeho:
Merge conflict after I merged PR #1767

runtime-modules/council/src/lib.rs
ondratra: Looking at the working group source code with `set_budget(...)` extrinsic made me think the periodical budget increases will be handled in the runtime and passed to the council module via `set_budget()` in the same way. There doesn't seem to be periodic refills in the working group module.

Is the periodic refill specific to the council only? Should I add this feature?

bedeho: Yes, there is nothing outside of the council that can make this determination. Voters cannot vote on updating this for example, voters just vote on who is on the council. Pre-launch we could use the SUDO module, but obviously that is not a solution for mainnet. Hence this must be a periodic automatic top up, and it should live in the council itself.

ondratra: Periodic budget refill has been added.

runtime-modules/council/src/lib.rs
```diff
@@ -524,7 +532,17 @@ impl<T: Trait> Module<T> {
     }
 
     /// Checkout elected council members reward payments.
-    fn try_reward_payments(now: T::BlockNumber) {
+    fn try_process_budget(now: T::BlockNumber) {
+        // budget autorefill
+        let next_refill = NextBudgetRefill::<T>::get();
+        if next_refill == 0.into() {
```
bedeho: Am I correct to interpret this to say that the council module insists that the first refill happens at block heights `BudgetRefillPeriod*k` for integers `k>0`?. If so then I don't think we should enforce this, it should be up to the runtime developer to chose the initial value of `NextBudgetRefill` in the context of a deployment, and if this is occuring as part of a runtime upgrade, then it has to be done anyway, as block `0` based setup is never triggered.

bedeho: If you make `NextBudgetRefill` configurable + have a module setter function that runtime upgrade migration code can invoke, then I think that should be good enough.

bedeho (commented):
Looks good, just one last fix!

runtime-modules/council/src/lib.rs
ondratra: Yes, you are correct; this was an imperfect initialization mechanism. `NextBudgetRefill` and others made configurable + new extrinsic `plan_budget_refill` was created.

bedeho (approved):
Great, looks good.

Issue #1751, opened 2020-11-18
Update proposals system to be able to calculate the weight for proposal execution in engine's on_initialize


conectado:
We need to move the logic for `on_finalize` of the `proposals/engine` to `on_initialize` as's recommended, since we need to know the weights of the proposal to-be executed in `on_initialize` and this is the only way to do it. 

Also, to be able to calculate the execution weight of the proposals that will be executed in `on_initialize` we need to refactor both codex and engine to store the weights of the functions that will be executed.

### Proposed steps
* [x] Move `on_finalize` to `on_initialize` in engine
* [x] Use `GetDispatchInfo` to calculate the weight for executed proposals in `on_initialize`

**Edit:** Changed the way to calculate the `on_initialize` weight using the encoded weight information.
**Edit:** Move the proposal benchmarking to its own ticket

shamil-gadelshin:
Done

PR #1764, opened 2020-11-19
Proposals update

conectado:
# PR to track #1751 

In this PR we refactored `proposal/engine` to execute the logic for proposal execution in `on_initialize` instead of `on_finalize` as it's recommended.

Furthermore, we take into account the weight for each executed proposal in the weight calculation for `on_initialize`

* Move `process_proposals` to `on_initialze`
* Update tests accordingly
* Add `GetDispatchInfo` as a required trait for `DispatchableCallCode`
* Use the weight info to calculate the executed proposal weights in `on_initialize`

### DIFF
--- a/runtime-modules/proposals/discussion/src/benchmarking.rs
+++ b/runtime-modules/proposals/discussion/src/benchmarking.rs
@@ -3,13 +3,12 @@ use super::*;
 use crate::Module as ProposalsDiscussion;
 use core::convert::TryInto;
 use frame_benchmarking::{account, benchmarks};
+use frame_system::EventRecord;
+use frame_system::Module as System;
+use frame_system::RawOrigin;
 use membership::Module as Membership;
 use sp_std::cmp::min;
 use sp_std::prelude::*;
-use system as frame_system;
-use system::EventRecord;
-use system::Module as System;
-use system::RawOrigin;
 
 const SEED: u32 = 0;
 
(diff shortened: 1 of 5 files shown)


shamil-gadelshin (changes requested):
A very nice finding of `GetDispatchInfo`! 

Please, consider making some changes:
- the base branch should be changed to olympia, since the proposal system update was merged to that branch. 
- update extrinsic comments to reflect recent changes (and module comment but it could be postponed)
- add additional weight calculation to the finalize_proposal. Weights calculation is a different (even orthogonal) aspect of the functions. So, it is hard to evaluate the full solution until the finalize_proposal is fully covered by weights estimation.

conectado:
>    add additional weight calculation to the finalize_proposal. Weights calculation is a different (even orthogonal) aspect of the functions. So, it is hard to evaluate the full solution until the finalize_proposal is fully covered by weights estimation.

This will be covered by the benchmark of `on_initialize`. When that is done and we include the weight function for `on_initialize` we change the `10_000_000`  in:

```
10_000_000 // TODO: adjust weight
  .saturating_add(Self::process_proposals())
```

To the maximum weight of all the paths in the `on_initialize` benchmarks and that will take into account the weight for `finalize_proposal` and the executed proposal if there is any.

conectado:
Moving this PR as part of #1731

PR #1766, opened 2020-11-19
Add text fadeIn and Make just cover clickable

### LINKED ISSUE #1712: Video preview improvements
- Make only the cover trigger the hover effect. Other than the cover, only the title should redirect to the video
- Make the appear transition (opacity) affect the text as well, not only the cover image

fcescob:
Fixes #1712

### DIFF
--- a/atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
+++ b/atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
@@ -1,11 +1,10 @@
 import React from 'react'
-import { keyframes } from '@emotion/core'
 import styled from '@emotion/styled'
 import { fluidRange } from 'polished'
 import { colors, spacing, typography } from '../../theme'
 import Avatar from '../Avatar'
 import Icon from '../Icon'
-import { HOVER_BORDER_SIZE } from './VideoPreviewBase.styles'
+import { HOVER_BORDER_SIZE, fadeInAnimation } from './VideoPreviewBase.styles'
 
 type MainProps = {
   main: boolean
@@ -92,16 +91,6 @@ export const StyledAvatar = styled(Avatar)<ChannelProps>`
   cursor: ${({ channelClickable }) => (channelClickable ? 'pointer' : 'auto')};
 `
 
-const fadeIn = keyframes`
-  0% {
-    opacity: 0;
-
-  }
-  100% {
-    opacity: 1;
-
-  }
-`
 export const TitleHeader = styled.h3<MainProps & ScalesWithCoverProps & ClickableProps>`
   margin: 0;
   font-weight: ${typography.weights.bold};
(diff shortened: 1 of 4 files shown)


atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
```diff
@@ -91,22 +92,36 @@ export const StyledAvatar = styled(Avatar)<ChannelProps>`
   cursor: ${({ channelClickable }) => (channelClickable ? 'pointer' : 'auto')};
 `
 
-export const TitleHeader = styled.h3<MainProps & ScalesWithCoverProps>`
+const fadeIn = keyframes`
```
kdembler: Couldn't we add that animation to the `InfoContainer` in the `VideoPreviewBase.style.tsx`? Then we can fade in the whole container, not the individual nodes

atlas/src/shared/components/VideoPreview/VideoPreviewBase.styles.tsx
```diff
@@ -31,16 +31,33 @@ export const CoverWrapper = styled.div`
...
   padding-top: 56.25%;
-
   transition-property: box-shadow, transform;
   transition-duration: 0.4s;
   transition-timing-function: cubic-bezier(0.165, 0.84, 0.44, 1);
   animation: ${fadeIn} 0.5s ease-in;
+  cursor: ${(props) => (props.clickable ? 'pointer' : 'auto')};
+  :hover {
+    ${(props) =>
+      props.clickable
```
kdembler: I found it a good practice to separate conditional styles to a separate `css` constant. That way it's much easier to read the styles. I know this is copied but perhaps we could do that anyway?

atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
fcescob: That would be better indeed, but doing so changes the color of the placeholders inside `InfoContainer` when loading, so I suggest we keep this as is for now

atlas/src/shared/components/VideoPreview/VideoPreviewBase.styles.tsx
fcescob: Good point, will change that

atlas/src/shared/components/VideoPreview/VideoPreview.styles.tsx
kdembler: Okay, we can skip that, although there is another problem. Currently the fade in animations aren't in sync, i.e. this uses different duration and easing than the cover. I suggest we define the keyframes and the animation in the base styles file and import it here so these parameters are always the same. Like so:

```js
const fadeIn = keyfreames...

export const fadeInAnimation = css`
  animation: ${fadeIn} 0.5s ease-in;
`
```

kdembler (approved):
LGTM

PR #1767, opened 2020-11-19
council - referendum start infallible

### LINKED ISSUE #1725: Council - statically prevent error on referendum start
Find a way how to statically ensure no error can occur when starting a referendum in the council. Basically, get rid of `unwrap()` from the following council's code inside of `end_announcement_period()`
```
        // TODO: try to find way how to get rid of unwrap here or staticly ensure it will not fail here
        // update state
        Mutations::<T>::finalize_announcing_period(&stage_data).unwrap(); // starting referendum should always start if implementation is valid - unwrap
```

ondratra:
This PR fixes #1725 by introducing a new infallible way of (re)starting a referendum.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -57,7 +57,7 @@ use serde::{Deserialize, Serialize};
 use sp_arithmetic::traits::BaseArithmetic;
 use sp_runtime::traits::{Hash, MaybeSerialize, Member};
 use std::marker::PhantomData;
-use system::{ensure_signed, RawOrigin};
+use system::ensure_signed;
 
 use referendum::{OptionResult, ReferendumManager};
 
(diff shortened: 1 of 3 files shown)


bedeho (commented):
This was an interesting approach.

So one thing I missed was the fact that the panickable code had to do with the integration of two modules. I thought it was just a random point. At the point where two modules are integrated, there is actually a genuine problem of lack of static guarantees in the module itself, its not an implementation imperfection per say. Normally I would have been OK with accepting the possibility of a silent & broken integration, and we have those other places also, as does other FRAME modules. I think the way to go would have been to indeed have `T::Referendum::start_referendum` be infallible, and then the runtime integration code just has to suppress any possible problem, since it is responsible for composing a full runtime where the parameters/calls passed to each module ensure that states are always compatible.

What I do wonder though is whether this forced start is entirely complete, is there no need to do any other sort of cleanup of the possible interrupted stage? like stake or votes or something? I can't really see anything obvious, but I am a bit nervous about it.

ondratra:
The referendum is quite a simple module, so we can go with this approach without any negative effects. All the votes cast in the incomplete referendum round will be ignored (since the round is never concluded), and relevant stakes will be available for unstaking. All checks voting/releasing/etc. checks in the referendum are concerned only about `ReferendumStage` and `CurrentCycleId`; both of them are updated when (re)starting the referendum, so everything is ok.

In the council use-case, the referendum will never even be prematurely ended.

PR #1778, opened 2020-11-20
Query node mappings - isCurated default value

Lezek123:
Fix the issue with `video.isCurated` defaulting to `true` when falsy in the query node mappings (making it impossible to set it to `false`)

### DIFF
--- a/query-node/mappings/content-directory/entity/create.ts
+++ b/query-node/mappings/content-directory/entity/create.ts
@@ -216,8 +216,7 @@ async function createVideo(
   video.description = p.description
   video.duration = p.duration
   video.hasMarketing = p.hasMarketing
-  // TODO: needs to be handled correctly, from runtime CurationStatus is coming
-  video.isCurated = p.isCurated || true
+  video.isCurated = p.isCurated || false
   video.isExplicit = p.isExplicit
   video.isPublic = p.isPublic
   video.publishedBeforeJoystream = p.publishedBeforeJoystream
--- a/query-node/mappings/content-directory/entity/create.ts
+++ b/query-node/mappings/content-directory/entity/create.ts
@@ -61,7 +61,7 @@ async function createChannel(
   channel.id = id
   channel.title = p.title
   channel.description = p.description
-  channel.isCurated = p.isCurated
+  channel.isCurated = !!p.isCurated
   channel.isPublic = p.isPublic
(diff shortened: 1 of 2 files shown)


query-node/mappings/content-directory/entity/create.ts
```diff
@@ -216,8 +216,7 @@ async function createVideo(
   video.description = p.description
   video.duration = p.duration
   video.hasMarketing = p.hasMarketing
-  // TODO: needs to be handled correctly, from runtime CurationStatus is coming
-  video.isCurated = p.isCurated || true
+  video.isCurated = p.isCurated || false
```
mnaamani: To convert truthy/falsy to boolean I find this syntax more comprehensible, and should convert falsy -> false and truthy -> true which is what is intended here?

```suggestion
  video.isCurated = !!p.isCurated
```

If the intent is to set video.isCurated to the value of p.isCurated if it is truthy, and false otherwise then your implementation is correct.

mnaamani: And if p.isCurated is boolean, why not just `video.isCurated = p.isCurated`

metmirr: Mokhtar when I start to implement the curation status it wasn't clear that if we would have a simple boolean field or some struct which would have more information. That's why I setting to a default value in case if there is no such field on the runtime. But later I have missed this code :(

metmirr:
Leszek you will need to make the fields nullable in the input schema (query-node/schema.graphql) as well

Lezek123:
Thanks @metmirr , I added a conversion from `undefined` to `false` through `!!` in `createVideo` and `createChannel`, hope this will be enough to solve this issue? (https://github.com/Joystream/joystream/pull/1778/commits/3970c5a934b83ea712346e5b86ebd11b989cadd2)

metmirr:
> Thanks @metmirr , I added a conversion from `undefined` to `false` through `!!` in `createVideo` and `createChannel`, hope this will be enough to solve this issue? ([3970c5a](https://github.com/Joystream/joystream/commit/3970c5a934b83ea712346e5b86ebd11b989cadd2))

Ahh, I miss the last commit, yeah It should solve there will be value in any case so tests should pass.

metmirr (approved):
Looks good!

PR #1798, opened 2020-11-23
Fix placeholders from jumping around

### LINKED ISSUE #1504: Prevent Layout From Jumping Around When Placeholders are replaced

fcescob:
Fix #1504


kdembler (approved):
LGTM

Issue #1732, opened 2020-11-17
Integrate new working-team pallet into the runtime.


shamil-gadelshin:
Additional tasks:
- We need to change `decrease_stake`, `increase_stake` for `working-team` pallet because of the changed `StakingHandler` interface. Replace absolute values with deltas.
- Remove old `working-group` pallet
- Rename `working-team` to `working-group`
- Fix `fill_opening` bug for zero possible applications.
- Consider removing old pallets: hiring, stake, mint, recurring-rewards

shamil-gadelshin:
Done

PR #1804, opened 2020-11-24
Working team fixes.

shamil-gadelshin:
Fixes:
- Change decrease_stake, increase_stake for working-team pallet because of the changed StakingHandler interface. Replace absolute values with deltas.
- Fix fill_opening bug for zero possible applications.

[Root issue](https://github.com/Joystream/joystream/issues/1732)


bedeho (approved):
Nice job:

1. How did you find the staking bug, why did we not find it before?
2. Can you double check that handbook is in synch with these changes, if not, make PR?

shamil-gadelshin:
1. Gabriel suspected the bug during his benchmarking work.
2. Handbook was initially correct, this how we found the need for the change during the last handbook changes 3 weeks ago.

Issue #1792, opened 2020-11-23
Implement log collecting


kdembler:
As Atlas heads into the phase where it will be finally used outside of the development team, we should integrate some log collecting solution into it. This ensures that if users run into some bugs/errors we don't encounter ourselves, we know of their existence and hopefully can reproduce&fix them.

PR #1809, opened 2020-11-24
Add Sentry Integration

fcescob:
Takes care of #1792 . Adds an environment variable called `SENTRY_DNS` which should be the Sentry DNS of our project. 

For now only basic error handling is added.


kdembler:
Moved to https://github.com/Joystream/atlas/pull/6

PR #1811, opened 2020-11-25
Content directory schemas - fix isCensored unique contraint bug

Lezek123:
Remove `unique` constraint from `isCensored` field (relic of `CurationStatus` beeing a separate entity before)


mnaamani:
Does this change require any updates in cli or query-node mappings? Otherwise LGTM

Lezek123:
I tested the CLI after the change and everything worked well.
The query node doesn't seem to enfore this uniquness, so the runtime schema fix should be enough.

PR #1815, opened 2020-11-25
council - winner vote stakes longer

### LINKED ISSUE #1768: Council - voting for a winner keeps stake locked longer
In case a voter votes for a council member candidate and the candidate wins the election, the voter's stake should be kept locked until the next election ends. Voters voting for a non-winning candidate might unstake right after the election ends (as it is right now).

ondratra:
Fixes #1768.

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -681,7 +681,7 @@ impl<T: Trait> Module<T> {
 
                 // depleted budget or no accumulated reward to be paid?
                 if balance == 0.into() || unpaid_reward == 0.into() {
-                    // no need to save anything because unpaid_reward will be recalculated next time rewards are paid
+                    // no need to update council member record here their; unpaid reward will be recalculated next time rewards are paid
 
                     // emit event
                     Self::deposit_event(RawEvent::RewardPayment(
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -584,7 +584,7 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
         // at this point vote.cycle_id == current_voting_cycle_id
 
         // ensure election has ended and voter haven't voted for winner
-        if voting_for_winner || !matches(Stage::<T>::get().stage, CouncilStage::Idle) {
+        if voting_for_winner || !matches!(Stage::<T>::get().stage, CouncilStage::Idle) {
             return Err(Error::CantReleaseStakeNow);
         }
 
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -249,10 +249,7 @@ pub trait ReferendumConnection<T: Trait> {
     fn recieve_referendum_results(winners: &[OptionResult<VotePowerOf<T>>]);
 
     /// Process referendum results. This function MUST be called in runtime's implementation of referendum's `can_release_voting_stake()`.
-    fn can_unlock_vote_stake(
-        vote: &CastVoteOf<T>,
-        current_voting_cycle_id: &u64,
-    ) -> Result<(), Error<T>>;
+    fn can_unlock_vote_stake(vote: &CastVoteOf<T>) -> Result<(), Error<T>>;
 
     /// Checks that user is indeed candidating. This function MUST be called in runtime's implementation of referendum's `is_valid_option_id()`.
     fn is_valid_candidate_id(membership_id: &T::MembershipId) -> bool;
@@ -756,12 +753,11 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
     }
 
     /// Check that it is a proper time to release stake.
-    fn can_unlock_vote_stake(
-        vote: &CastVoteOf<T>,
-        current_voting_cycle_id: &u64,
-    ) -> Result<(), Error<T>> {
+    fn can_unlock_vote_stake(vote: &CastVoteOf<T>) -> Result<(), Error<T>> {
+        let current_voting_cycle_id = AnnouncementPeriodNr::get();
+
         // allow release for very old votes
-        if *current_voting_cycle_id > vote.cycle_id + 1 {
+        if current_voting_cycle_id > vote.cycle_id + 1 {
             return Ok(());
         }
 
@@ -771,7 +767,7 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
             .any(|membership_id| vote.vote_for == Some(membership_id.into()));
 
         // allow release for vote from previous elections only when not voted for winner
-        if *current_voting_cycle_id == vote.cycle_id + 1 {
+        if current_voting_cycle_id == vote.cycle_id + 1 {
             // ensure vote was not cast for the one of winning candidates / council members
             if voting_for_winner {
                 return Err(Error::CantReleaseStakeNow);
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -59,7 +59,7 @@ use sp_runtime::traits::{Hash, MaybeSerialize, Member};
 use std::marker::PhantomData;
 use system::ensure_signed;
 
-use referendum::{OptionResult, ReferendumManager};
+use referendum::{CastVote, OptionResult, ReferendumManager};
 
 // declared modules
 mod mock;
@@ -152,6 +152,7 @@ pub type VotePowerOf<T> = <<T as Trait>::Referendum as ReferendumManager<
     <T as system::Trait>::AccountId,
     <T as system::Trait>::Hash,
 >>::VotePower;
+pub type CastVoteOf<T> = CastVote<<T as system::Trait>::Hash, Balance<T>>;
 
 pub type CouncilMemberOf<T> =
     CouncilMember<<T as system::Trait>::AccountId, <T as Trait>::MembershipId, Balance<T>>;
@@ -175,7 +176,8 @@ pub trait Trait: system::Trait {
         + Copy
         + MaybeSerialize
         + PartialEq
-        + From<u64>;
+        + From<u64>
+        + Into<u64>;
 
     /// Referendum used for council elections.
     type Referendum: ReferendumManager<Self::Origin, Self::AccountId, Self::Hash>;
@@ -212,7 +214,10 @@ pub trait ReferendumConnection<T: Trait> {
         -> Result<(), Error<T>>;
 
     /// Process referendum results. This function MUST be called in runtime's implementation of referendum's `can_release_voting_stake()`.
-    fn can_release_vote_stake() -> Result<(), Error<T>>;
+    fn can_release_vote_stake(
+        vote: &CastVoteOf<T>,
+        current_voting_cycle_id: &u64,
+    ) -> Result<(), Error<T>>;
 
     /// Checks that user is indeed candidating. This function MUST be called in runtime's implementation of referendum's `is_valid_option_id()`.
     fn is_valid_candidate_id(membership_id: &T::MembershipId) -> bool;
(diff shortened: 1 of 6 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -552,12 +557,35 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
     }
 
     /// Check that it is a proper time to release stake.
-    fn can_release_vote_stake() -> Result<(), Error<T>> {
```
bedeho, line 762: Isn't this old policy actually saying that you can never unlock your voting stake if there is an active election, even if it was 100 cycled ago? If so then this was actually a bug, right? not only limited in not allowing losing votes to unlock right away.

runtime-modules/council/src/lib.rs
```diff
@@ -552,12 +557,35 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
-        // ensure it's proper time to release stake
-        match Stage::<T>::get().stage {
-            CouncilStage::Idle => (),
-            _ => return Err(Error::CantReleaseStakeNow),
-        };
+    fn can_release_vote_stake(
```
bedeho: Can we rename this to `can_unlock_...`, release is slightly more ambiguous :D.

runtime-modules/council/src/lib.rs
```diff
@@ -552,12 +557,35 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
+        vote: &CastVoteOf<T>,
+        current_voting_cycle_id: &u64,
+    ) -> Result<(), Error<T>> {
+        // allow release for very old votes
+        if vote.cycle_id > current_voting_cycle_id + 1 {
```
bedeho: This test is supposed to unlock very old votes, in which case the test seems wrong, because old votes would have a a much _lower_ cycle ID than the current one, this test checks the opposite?

runtime-modules/council/src/lib.rs
```diff
@@ -552,12 +557,35 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
     /// Check that it is a proper time to release stake.
-    fn can_release_vote_stake() -> Result<(), Error<T>> {
-        // ensure it's proper time to release stake
-        match Stage::<T>::get().stage {
-            CouncilStage::Idle => (),
-            _ => return Err(Error::CantReleaseStakeNow),
-        };
+    fn can_release_vote_stake(
+        vote: &CastVoteOf<T>,
+        current_voting_cycle_id: &u64,
```
bedeho: Why are you passing in a cycle id? This is the announcement period nr, right? This is stored in `AnnouncementPeriodNr`?

bedeho: This comment made me look at referendum to see where this id is coming from, and it is from a separate

```
        /// Index of the current referendum cycle. It is incremented everytime referendum ends.
        pub CurrentCycleId get(fn current_cycle_id) config(): u64;
```

This is only updated in `reset_referendum`, which is called only from `conclude_referendum`, which is called only from `end_reveal_period`, which at the same time informs the rest of the runtime about an ended election with `T::process_results(&winners);`. So I wonder, is there a need for having a separate representation of cycle id here? Its not clear to me, please respond.

If its not required, then suggest a fix in a separate issue, solve that in a separate PR, have me review it, when its accepted, reconcile this PR with that one by fixing the signature for this routine and anything else.

runtime-modules/council/src/lib.rs
```diff
@@ -552,12 +557,35 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
...
+            if voting_for_winner {
+                return Err(Error::CantReleaseStakeNow);
+            }
+
+            return Ok(());
+        }
+
+        // at this point vote.cycle_id == current_voting_cycle_id
+
+        if voting_for_winner {
```
bedeho: I don't understand the case for accepting the unlocking if this test is false? What scenario is that, why is it `Ok` ?

bedeho (changes requested):

runtime-modules/council/src/lib.rs
ondratra: Yes, this was a bug - unnecessary limitation of stake release.

runtime-modules/council/src/lib.rs
ondratra: There was a part of condition missing.  `if voting_for_winner || !matches!(Stage::<T>::get().stage, CouncilStage::Idle) { return Err...` allows stake unlock only after election has ended and the voter haven't voted for a winner.

runtime-modules/council/src/lib.rs
ondratra: `AnnouncementPeriodNr` is a number of announcement phases, while the referendum's `CurrentCycleId` is the number of elections held. These two will differ as there can be more announcement phases than elections. 

Your note on `reset_referendum`, etc., made me notice a bug there. `CurrentCycleId` is now increased when the referendum concludes as well as when it is restarted using `force_start`. This didn't create a problem for a council but was a mistake anyway. Fixed

runtime-modules/council/src/lib.rs
ondratra: Yes, you are right. The two variables were mistakenly exchanged in this condition.

runtime-modules/council/src/lib.rs
ondratra: Done.

runtime-modules/council/src/lib.rs
bedeho: Ah, yes I agree, those are of course different, but then my question is: do we need to have the election id as separate? 

For the purpose of the referendum, we want to distinguish votes from different election rounds, hence couldn't the council just pass in the id to be used for the given election, and use it's announcement number as that id, and that way there is no need to maintain independent state in the referendum module from one cycle to the next? You can simply accept the unique id from the starting call of the caller module (election), and store it as part of the state of the current election, and then have it expire once the referendum goes to idle state. 

What do you think? Is this safer, or not? the caller can of course make a mistake and pass in non-unique values across calls also, this is currently not possible.

bedeho:
Merge conflict!

runtime-modules/council/src/lib.rs
bedeho: It took me a while but I think I understood the bug you found, however the result now is that you are effectively having a dummy value for the cycle Id `Inactive` stage. So I think its basically like this:

1. You start in Voting stage of a new referendum with id = `N`
2. When reveal stage is over the new id is `N+1`, however this value is not ever written back anywehre, no vote is cast with this value, nor is there any referendum so to speak.
3. A new voting stage is started with id = `N+2`

If this is correct then I think its quite akward, and I also think part of the reason it took me so long to figure out whether things were correct in `can_unlock_vote_stake` this time around, in terms of all of the id compairson checks, is that we have a cycle id set when there is not cycle going. 

My proposal above would implictly solve this problem, as there is only an id for an election when there is an election, when one is in the `Inactive` stage, then there is no cycle id to even compare votes against, clearing things up.

ondratra: This is a really good question. After giving it some thought, I think it is a viable solution, but it would introduce the potential to start a referendum with the bad index, as you mentioned (breaking ascending order of indices). Also, this externalization of the election counter creates a need to reimplement such a counter in each module that uses referendum, thus reducing reusability and easy-to-use (in the future as we are using it only in the council right now).

On the other hand, I don't think there are any security or terrible performance issues with keeping the index; thus, I am rather for keeping the index present in the referendum.

ondratra: Replying to your second post: 

The id doesn't change between voting and revealing stages:
1. voting stage starts with id = `N`
2. revealing stage still has id = `N`
3. when a new voting stage starts, it has id = `N+1`

bedeho: > The id doesn't change between voting and revealing stages:

I know, that is not what I wrote, I wrote `When reveal stage is over the new`

|  Stage  |  Id  |
|---|---|
| Voting  |  `N`   |
| Revealing  |  `N`   |
| Idle  |  `N+1`  (dummy value) |
| Voting  |  `N+2`   |
| Revealing  |  `N+2`   |
| Idle  |  `N+3`  (dummy value) |

> I don't think there are any security or terrible performance issues with keeping the index

I agree it's not about performance at all, it's about clarity. If we are going to keep this Id I don't know what semantics we give it that gives clarity to the reader. The fact that I even missed the latest fix you added in `897b63f` makes me think we do not have sufficient clarity.

> election counter creates a need to re-implement such a counter in each module that uses referendum

This is true, but remember that this voting system is inherently periodic, this is why we have the concept of votes hanging around from old elections while new elections get started. I think it will be very natural for any client module to itself reflect this periodicity in some way, as it has to provide the candidate pool and so on over time. Stated differently, I can't think of a user module of `referendum` which only ever invokes a single election.

ondratra: I think there was a dummy value as you described at one point, but it was later reworked, so `N` "is" increased at the referendum started, so there is no such dummy value.

Anyway, I removed the counter from the referendum and adapted the council.

runtime-modules/referendum/src/lib.rs
```diff
@@ -827,20 +828,25 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
...
+
         // allow release only for past cycles
-        if cycle_id == cast_vote.cycle_id {
+        if current_cycle_id == Some(cast_vote.cycle_id) {
             return Err(Error::UnstakingVoteInSameCycle);
         }
 
         // ask runtime if stake can be released
-        if !T::can_unlock_vote_stake(&cast_vote, &cycle_id) {
+        if !T::can_unlock_vote_stake(&cast_vote) {
```
bedeho, line 837: This seems out of synch with signature on council, it accepts `current_voting_cycle_id: &u64`.

runtime-modules/referendum/src/lib.rs
```diff
@@ -827,20 +828,25 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
...
     fn can_release_vote_stake(origin: T::Origin) -> Result<T::AccountId, Error<T, I>> {
-        let cycle_id = CurrentCycleId::<I>::get();
-
         // ensure superuser requested action
         let account_id = Self::ensure_regular_user(origin)?;
 
         let cast_vote = Self::ensure_vote_exists(&account_id)?;
 
+        // get current cycle id
+        let current_cycle_id = match Stage::<T, I>::get() {
```
bedeho: I guess the question is here: should we do this check here, or just leave it to the client to decide whether its sometimes ok to even unstake in the same election cycle? If it happens during voting, then there is no harm really, but if the client allows it during revealing, the the client could allow people to unstake regardless of what the relationship is between the vote and the outcome. I think that is fine in principle, in some voting systems it could be perfectly acceptable.

A benefit of keeping all the checks in the council is that you have one small place that holds all the rules, rather than splitting it across the two modules like this.

runtime-modules/referendum/src/lib.rs
bedeho: This should have caused the CI to complain if I am correct, but its actually complaining for a differnt reason, so perhaps you can fix that formatting issue also.

runtime-modules/council/src/lib.rs
```diff
@@ -682,18 +682,10 @@ impl<T: Trait> Module<T> {
...
-                        council_member.membership_id,
-                        council_member.reward_account_id.clone(),
-                        0.into(),
-                        0.into(),
-                    ));
-                    return balance;
-                }
+                // depleted budget or no accumulated reward to be paid?
+                if balance == 0.into() || unpaid_reward == 0.into() {
+                    // no need to save anything because unpaid_reward will be recalculated next time rewards are paid
```
bedeho: I think change is fine, but this comment threw me off, what are you trying to say here on L687?

runtime-modules/referendum/src/lib.rs
ondratra: Unnecessary cycle id has been removed from the council function; synced now.

runtime-modules/referendum/src/lib.rs
ondratra: The conditioning has been moved to council.

runtime-modules/council/src/lib.rs
ondratra: Comment updated. I just wanted to explain that there is no need to update the council member record even when he is not paid.

bedeho (approved):
Perfect!

PR #1818, opened 2020-11-26
CLI - handling CreativeCommons license attribution

Lezek123:
Add a way to handle CC licenses that require attribution field (ie. `CC_BY`)

In order to keep the schemas backward-compatible this is achieved by converting a `KnownLicense` to `UserDefinedLicense` in case attribution is required.

The resulting `UserDefinedLicense` then has a following tempate:
```
{ CODE } ({ NAME })

{ DESCRIPTION }

{ URL }

Attribution: { ATTRIBUTION }
```

For example:
```
CC_BY_NC_ND (Creative Commons Attribution-NonCommercial-NoDerivs License)

Some short description of CC_BY_NC_ND license (it's optional)

https://creativecommons.org/licenses/by-nc-nd/4.0

Attribution: John Doe
```


bwhm (approved):
AFAICT, `CC0 doesn't require attribution, but looks good apart from that.

It will make sorting by license a little more difficult, but that shouldn't be a problem, so it looks like a good solution!

bwhm:
I should mention that @blrhc has written descriptions for the licenses, but we can do that in another PR, as I will add more details for other `entityBatches` as well...

Lezek123:
Will be included in https://github.com/Joystream/joystream/pull/1821. Since handling it in `json` input would require adding even more boilerplate code, we decided to actually include `attribution` field in the `License` schema and implement necessary changes in Query node and Atlas

PR #1817, opened 2020-11-26
Council constants exposure

ondratra:
This PR exposes referendum and council constants to the RPC. 

To minimize merge conflict potential, this PR assumes #1760 and #1815 will be merged first.

PR #1820, opened 2020-11-27
Council olympia

ondratra:
This PR merges recent changes from the Olympia branch to the council branch.

To minimize merge conflict potential, this PR assumes #1817 will be merged first.


ondratra:
Issues with current StakingHandler implementation:
- `is_account_free_of_conflicting_stakes()` checks if the same lock type already exists, but I think it should actually expect Runtime to do this check and (dis)approve certain combinations
- `is_member_staking_account` is not actually implemented (returns `true`)
- `set_stake` unlocks and locks balance again in one certain case; this can be prevented by check for balance at the start

PR #1823, opened 2020-11-27
Introduce TransactionFailed event and adjust related core logic & test coverage

iorveth:
This PR aims to 
1. Introduce `TransactionFailed(Actor, FailedAtOperationIndex)` event, emitted only when an error during `operations` list execution occurs.
1. Adjust related test coverage to reflect this case


bedeho (approved):
Thanks for the suggestion and very quick implementation, hopefully will save us tons of pain in Hydra.

PR #1825, opened 2020-11-29
Query node transaction failed event

metmirr:
Add an event handler for `TransactionFailed` event, only successful operations are executed by the event handler.


query-node/mappings/content-directory/transaction.ts
```diff
@@ -81,33 +81,45 @@ async function getNextEntityId(db: DB): Promise<number> {
...
+  debug(`TransactionFailed event: ${JSON.stringify(event)}`)
+
+  const failedOperationIndex = event.params[1].value as number
+  const operations = decode.getOperations(event)
+
+  if (operations.length === 0 || operations.length === 1) return
+
+  const successfulOperations = operations.filter((op, index) => index < failedOperationIndex)
+
+  const {
```
mnaamani: You can factor out this common code in `contentDirectory_TransactionFailed` and `contentDirectory_TransactionCompleted` into a function `applyOperations(operations: IBatchOperation)`

query-node/mappings/content-directory/transaction.ts
```diff
@@ -81,33 +81,45 @@ async function getNextEntityId(db: DB): Promise<number> {
...
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export async function contentDirectory_TransactionFailed(db: DB, event: SubstrateEvent): Promise<void> {
+  debug(`TransactionFailed event: ${JSON.stringify(event)}`)
+
+  const failedOperationIndex = event.params[1].value as number
+  const operations = decode.getOperations(event)
+
+  if (operations.length === 0 || operations.length === 1) return
+
+  const successfulOperations = operations.filter((op, index) => index < failedOperationIndex)
```
mnaamani: I think you can combine these two statements into a single one using `operations.slice()`

mnaamani (changes requested):
Implementation looks correct, just left some refactoring change request.
Will test queries now and report back.

query-node/mappings/content-directory/transaction.ts
metmirr: Much better now :) done.

metmirr: Earlier I attempted to use `slice` but I was getting `TypeError: registry.createClass is not a function`, I couldn't recall calling `toArray` on polkadot codec.

query-node/mappings/content-directory/transaction.ts
metmirr: Done.

query-node/mappings/content-directory/transaction.ts
```diff
@@ -88,19 +89,10 @@ export async function contentDirectory_TransactionFailed(db: DB, event: Substrat
   const failedOperationIndex = event.params[1].value as number
   const operations = decode.getOperations(event)
 
-  if (operations.length === 0 || operations.length === 1) return
+  const successfulOperations = operations.toArray().slice(2, failedOperationIndex)
```
mnaamani: why are we always skipping first two operations?

metmirr: If `operations` length is `0` then there is no successful operation, if the length is `1` then there is only one operation which is the failed one (that's why the `TransactionFailed` event is triggered for). In both cases, there is nothing for me to process.

metmirr: Okay, that's wrong! I am updating now! I mixed the indices with the length

PR #1836, opened 2020-12-01
Proposals staking check fix.

shamil-gadelshin:


runtime-modules/proposals/engine/src/lib.rs
```diff
@@ -516,8 +522,17 @@ impl<T: Trait> Module<T> {
             creation_params.exact_execution_block,
         )?;
 
-        // checks passed
-        // mutation
+        if creation_params.proposal_parameters.required_stake.is_some() {
```
bedeho: Looks correct, but I think its slightly clearer if the check is not split across the ensure boundry

```
            // Staking account is provided if required
            ensure!(
                creation_params.proposal_parameters.required_stake.is_none() ||
                creation_params.staking_account_id.is_some(),
                Error::<T>::EmptyStake
            );

```

bedeho (changes requested):

runtime-modules/proposals/engine/src/lib.rs
shamil-gadelshin: I thought about such composition but it requires extra comment because it is less obvious: we need to parse ` (is_none() || is_some())` formula in the head.

shamil-gadelshin: Removed it at all. It was a duplicate of the part of `ensure_create_proposal_parameters_are_valid`. But I added a test for this error.

shamil-gadelshin:
I also added a check for the conflicting stakes (and a test).

bedeho (approved):

PR #1840, opened 2020-12-01
Query node - remove duplicate typedefs file

Lezek123:
Remove `query-node/typedefs.json` and replace the references in `.env` to `types/augment/all/defs.json`.
This should let us avoid the need to maintain 2 separate files with the same content (since `types/augment/all/defs.json` is already always updated by `yarn workspace @joystream/types generate:all`)


mnaamani (approved):
I can imagine how this could have been an issue with new types added recently not reflecting in query node. Nice fix.

metmirr (approved):
Thanks!

PR #1844, opened 2020-12-01
forum_config: fix data_migration_done encoded representation

iorveth:
This PR aims to fix `data_migration_done` encoded representation, setting it to "0x01" (true), when `empty(forum_sudo: AccountId)` funtion called.


shamil-gadelshin (approved):
The fix works. I tested it locally and the benchmarking bug disappeared.

PR #1846, opened 2020-12-01
Contitution pallet benchmarks and weights.

shamil-gadelshin:
SSIA


runtime-modules/constitution/src/benchmarking.rs
```diff
@@ -0,0 +1,53 @@
...
+    // compare to the last event record
+    let EventRecord { event, .. } = &events[events.len() - 1];
+    assert_eq!(event, &system_event);
+}
+
+benchmarks! {
+    _{ }
+
+    amend_constitution{
+        let i in 1 .. 50000;
```
conectado: I think it would be a good idea to move the upper-bound constant(50000) into a constant `MAX_BYTES` to be consistent with the benchmark for the other pallets

runtime-modules/constitution/src/lib.rs
```diff
@@ -8,18 +8,34 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #[cfg(test)]
-mod tests;
+pub(crate) mod tests;
+
+#[cfg(feature = "runtime-benchmarks")]
```
conectado: We don't need to use restrict this to `runtime-benchmarks` since `benchmarking.rs` already has ` #![cfg(feature = "runtime-benchmarks")]`

conectado (changes requested):
Seems quite good, just 2 small comments.

runtime-modules/constitution/src/benchmarking.rs
shamil-gadelshin: Ok.

runtime-modules/constitution/src/lib.rs
shamil-gadelshin: Ok.

conectado (approved):
LGTM

PR #1847, opened 2020-12-01
Query node add attribution field to License entity

metmirr:
Add `attribution` field to `LicenseEntity`. After this update, fetching `license` field for videos should be like this:

```graphql
query {
  videos {
    id
    license {
      attribution
      type {
        __typename
        ... on KnownLicense {
          code
          name
        }
        ... on UserDefinedLicense {
          content
        }
      }
    }
  }
}
```

Querying all the `License` entities:

```graphql
  licenseEntities(where: { type_json: { code_eq: "CC_BY_SA" } }) {
    id
    attribution
    type {
      __typename
      ... on KnownLicense {
        code
        name
      }
      ... on UserDefinedLicense {
        content
      }
    }
  }
```

### DIFF
--- a/query-node/mappings/content-directory/entity/create.ts
+++ b/query-node/mappings/content-directory/entity/create.ts
@@ -205,14 +205,12 @@ async function createVideoMedia(
     const { httpMediaLocation, joystreamMediaLocation } = m
     if (httpMediaLocation) {
       const mediaLoc = new HttpMediaLocation()
-      mediaLoc.isTypeOf = 'HttpMediaLocation'
       mediaLoc.port = httpMediaLocation.port
       mediaLoc.url = httpMediaLocation.url
       videoMedia.location = mediaLoc
     }
     if (joystreamMediaLocation) {
       const mediaLoc = new JoystreamMediaLocation()
-      mediaLoc.isTypeOf = 'JoystreamMediaLocation'
       mediaLoc.dataObjectId = joystreamMediaLocation.dataObjectId
       videoMedia.location = mediaLoc
     }
@@ -331,7 +329,7 @@ async function createLicense(
       nextEntityIdBeforeTransaction
     )
     const k = new KnownLicense()
-    k.isTypeOf = 'KnownLicense'
+    // k.isTypeOf = 'KnownLicense'
     k.code = kLicense.code
     k.description = kLicense.description
     k.name = kLicense.name
--- a/query-node/mappings/content-directory/entity/create.ts
+++ b/query-node/mappings/content-directory/entity/create.ts
@@ -329,7 +329,6 @@ async function createLicense(
       nextEntityIdBeforeTransaction
     )
     const k = new KnownLicense()
-    // k.isTypeOf = 'KnownLicense'
     k.code = kLicense.code
     k.description = kLicense.description
     k.name = kLicense.name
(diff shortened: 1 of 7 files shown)


query-node/mappings/content-directory/entity/remove.ts
```diff
@@ -48,49 +48,28 @@ async function removeLicense(db: DB, where: IWhereCond): Promise<void> {
...
-      },
-    })
-  }
-  if (userdefinedLicense) {
-    videos = await db.getMany(Video, {
-      where: { license: { isTypeOf: 'UserDefinedLicense', content: userdefinedLicense.content } },
-    })
+  if (record.videolicense) {
+    // Remove all the videos under this license
+    record.videolicense.map(async (v) => await removeVideo(db, { where: { id: v.id } }))
```
mnaamani: Does it make sense to await for all `removeVideo` operations to complete before removing the License Entity ? Could there be a foreign key violation if we do not?

```suggestion
    await Promise.all(record.videolicense.map(async (v) => await removeVideo(db, { where: { id: v.id } })))
```

query-node/mappings/content-directory/entity/create.ts
```diff
@@ -340,27 +321,42 @@ async function createLicense(
...
+
+  if (p.knownLicense) {
+    const kLicense = await getOrCreate.knownLicense(
       { db, block, id },
       classEntityMap,
-      knownLicense,
+      p.knownLicense,
       nextEntityIdBeforeTransaction
     )
+    const k = new KnownLicense()
```
mnaamani, line 331: Its a shame these classes generated from input schema don't have a constructor, passing these values in the constructor would be so much better than setting the fields individually.

query-node/mappings/content-directory/entity/create.ts
```diff
@@ -340,27 +321,42 @@ async function createLicense(
...
+    k.isTypeOf = 'KnownLicense'
```
mnaamani: I think this is the default value on construction and doesn't seem necessary to set again.

mnaamani (approved):
LGTM will test queries.
Just left a note about some async/await issue,

query-node/mappings/content-directory/entity/remove.ts
metmirr: Yes, otherwise it will not be possible to delete the license since videos have references. But this code (removing videos) should never get executed because runtime enforces reference counter to be zero so they are removed in an order that key violation would never happen. I was just keeping them around maybe would be better to remove them?

query-node/mappings/content-directory/entity/create.ts
metmirr: Agree with that, we should add a constructor to our template for these classes.

query-node/mappings/content-directory/entity/create.ts
metmirr: Done.

query-node/mappings/content-directory/entity/remove.ts
mnaamani: Okay thanks for clarifying. So when you say code should never get executed, that is generally an indication that you should not actually have any code there and as you suggested remove it (because you are trying to fix a mistake in the runtime) and instead have an assertion that `record.videolicense` is an empty array instead.

mnaamani: You could ofcourse just remove the code completely, and the processor would fail with an ugly message about key violation. Having the assertion with a good message will make it easier to debug the error if it ever happens in the future.

metmirr: Done.

mnaamani (approved):

PR #1850, opened 2020-12-02
joystream node docker image add some dependencies for rocksdb

mnaamani:
Based on https://github.com/facebook/rocksdb/blob/master/INSTALL.md seems some dependencies were missing in the docker image to successfully build joystream/node


mnaamani:
It seems the dependencies haven't actually changed recently so this fix isn't actually necessary.

Looking at recent docker hub builds it seems its due to a version change in librocks db. If you have a recent cache of cargo packages you might not run into the build issue. I tried locally without issue. But on clean builds it is a problem.

Screen Shot 2020-12-02 at 10 26 01 AM

mnaamani:
Replaced with https://github.com/Joystream/joystream/pull/1852

Issue #516, opened 2020-05-30
Drop speculative unused fields from membership


bedeho:
From each profile

```Rust
    /// Whether the member is suspended or not.
    pub suspended: bool,

    /// The type of subscription the member has purchased if any.
    pub subscription: Option<SubscriptionId>,
```

and also from storage

```Rust
pub PaidMembershipTermsById get(paid_membership_terms_by_id) build(|config: &GenesisConfig<T>| {
            // This method only gets called when initializing storage, and is
            // compiled as native code. (Will be called when building `raw` chainspec)
            // So it can't be relied upon to initialize storage for runtimes updates.
            // Initialization for updated runtime is done in run_migration()
            let terms = PaidMembershipTerms {
                fee:  config.default_paid_membership_fee,
                text: Vec::default(),
            };
            vec![(T::PaidTermId::from(DEFAULT_PAID_TERM_ID), terms)]
        }) : map T::PaidTermId => PaidMembershipTerms<BalanceOf<T>>;

        /// Active Paid membership terms
        pub ActivePaidMembershipTerms get(active_paid_membership_terms) : Vec<T::PaidTermId> = vec![T::PaidTermId::from(DEFAULT_PAID_TERM_ID)]

pub ScreeningAuthority get(screening_authority) : T::AccountId;
```

bedeho:
@shamil-gadelshin I forget now, but I am seeing that this issue and a bunch of other related membership module cleanup issues appear to not have been tackled in `Alexandria` release. What did we agree upon here, are we tabling this for later?

PR #1855, opened 2020-12-02
Drop obsolete membership code.

shamil-gadelshin:
### Related issues
- [Drop speculative unused fields from membership](https://github.com/Joystream/joystream/issues/516)
- [Drop InitialMembersBalance membership module configuration](https://github.com/Joystream/joystream/issues/516)

Comments:
- buy_membership extrinsic was changed from `paid_term_id` to get membership price to `MembershipFee` trait constant
- `default_membership_fee` was removed from the genesis config.

### DIFF
--- a/runtime-modules/membership/src/lib.rs
+++ b/runtime-modules/membership/src/lib.rs
@@ -14,13 +14,14 @@ use sp_runtime::traits::{MaybeSerialize, Member};
 use sp_std::borrow::ToOwned;
 use sp_std::vec::Vec;
 
-use common::currency::{BalanceOf, GovernanceCurrency};
+// Balance type alias
+type BalanceOf<T> = <T as balances::Trait>::Balance;
 
 //TODO: Convert errors to the Substrate decl_error! macro.
 /// Result with string error message. This exists for backward compatibility purpose.
 pub type DispatchResult = Result<(), &'static str>;
 
-pub trait Trait: frame_system::Trait + GovernanceCurrency + pallet_timestamp::Trait {
+pub trait Trait: frame_system::Trait + balances::Trait + pallet_timestamp::Trait {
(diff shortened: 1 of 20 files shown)


runtime-modules/membership/src/lib.rs
```diff
@@ -273,24 +209,23 @@ decl_module! {
             // make sure we are accepting new memberships
             ensure!(Self::new_memberships_allowed(), "new members not allowed");
 
-            // ensure paid_terms_id is active
-            let terms = Self::ensure_active_terms_id(paid_terms_id)?;
+            let fee = T::MembershipFee::get();
 
             // ensure enough free balance to cover terms fees
-            ensure!(T::Currency::can_slash(&who, terms.fee), "not enough balance to buy membership");
+            ensure!(T::Currency::can_slash(&who, fee), "not enough balance to buy membership");
```
bedeho: Are `can_slash` and `slash` correct to use? I know they were already there, but from the docs it says

`Deducts up to value from the combined balance of who, preferring to deduct from the free balance. This function cannot fail.`

So presumably `can_slash` will also count funds that are locked for some purpose, but we really don't want people to be able to use any locked funds for buying a membership. Only free balance should count, the handbook says

```
Free balance of the signer account exceeds `membership_price`. 
```

bedeho (commented):
So small, so nice, great job, just one question.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Fixed:
- changed `free_balance` check with `usable_balance` (We don't use other balance purposes such as reservation, so, for us, `usable_balance` is `free_balance` minus locked amount.)
- added a test for the fix

bedeho: Excellent, and thanks for the offline explainer.

PR #1858, opened 2020-12-03
Working group weight annotations

conectado:
This PR includes:

* Fix `working_group` benchmark in Olympia
* Add `working_group` pallet to benchmark
  * **Note:** I used `ForumWorkingGroup` but for what I've seen which instance we should use to benchmark is arbitrary
* Add weight annotations for `working_group` pallet

### DIFF
--- a/runtime-modules/working-group/src/lib.rs
+++ b/runtime-modules/working-group/src/lib.rs
@@ -278,6 +278,14 @@ decl_module! {
         /// Exports const -  max simultaneous active worker number.
         const MaxWorkerNumberLimit: u32 = T::MaxWorkerNumberLimit::get();
 
+        /// # <weight>
+        ///
+        /// ## Weight
+        /// `O (W)` where:
+        /// - `W` is the number of workers currently present in the `working_group`
+        /// - DB:
+        ///    - O(W)
+        /// # </weight>
         fn on_initialize() -> Weight {
             let leaving_workers = Self::get_workers_with_finished_unstaking_period();
             let mut biggest_number_of_processed_workers = leaving_workers.len();
@@ -303,6 +311,15 @@ decl_module! {
 
         /// Add a job opening for a regular worker/lead role.
         /// Require signed leader origin or the root (to add opening for the leader position).
+        ///
+        /// # <weight>
+        ///
+        /// ## Weight
+        /// `O (D)` where:
+        /// - `D` is the length of `description`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
         #[weight = WeightInfoWorkingGroup::<T, I>::add_opening(description.len().saturated_into())]
         pub fn add_opening(
             origin,
@@ -343,6 +360,15 @@ decl_module! {
         }
 
         /// Apply on a worker opening.
+        ///
+        /// # <weight>
+        ///
+        /// ## Weight
+        /// `O (D)` where:
+        /// - `D` is the length of `p.description`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
         #[weight = WeightInfoWorkingGroup::<T, I>::apply_on_opening(p.description.len().saturated_into())]
         pub fn apply_on_opening(origin, p : ApplyOnOpeningParameters<T>) {
             // Ensure the origin of a member with given id.
@@ -406,6 +432,14 @@ decl_module! {
 
         /// Fill opening for the regular/lead position.
         /// Require signed leader origin or the root (to fill opening for the leader position).
+        /// # <weight>
+        ///
+        /// ## Weight
+        /// `O (A)` where:
+        /// - `A` is the length of `successful_application_ids`
+        /// - DB:
+        ///    - O(A)
+        /// # </weight>
         #[weight =
             WeightInfoWorkingGroup::<T, I>::fill_opening_worker(successful_application_ids.len().saturated_into())
             .max(WeightInfoWorkingGroup::<T, I>::fill_opening_lead())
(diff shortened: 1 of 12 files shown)


iorveth:
Just one comment here. 
Wouldn`t it be safer to have separate benchmarks for all working group instances on node runtime level?
My primarily concern is that we can probably have individual constraint limits for working group instances and that can result in different worst-case scenarios. https://github.com/Joystream/joystream/blob/olympia/runtime/src/lib.rs#L620

conectado:
> Just one comment here.
> Wouldn`t it be safer to have separate benchmarks for all working group instances on node runtime level?
> My primarily concern is that we can probably have individual constraint limits for working group instances and that can result in different worst-case scenarios. https://github.com/Joystream/joystream/blob/olympia/runtime/src/lib.rs#L620

I agree. I don't know if it's possible however to run multiple benchmarks for the same pallet. I will try.

conectado:
> > Just one comment here.
> > Wouldn`t it be safer to have separate benchmarks for all working group instances on node runtime level?
> > My primarily concern is that we can probably have individual constraint limits for working group instances and that can result in different worst-case scenarios. https://github.com/Joystream/joystream/blob/olympia/runtime/src/lib.rs#L620
> 
> I agree. I don't know if it's possible however to run multiple benchmarks for the same pallet. I will try.

Can be done, will upload the changes ASAP.

bedeho:
Nice observation @iorveth .

conectado:
> > > Just one comment here.
> > > Wouldn`t it be safer to have separate benchmarks for all working group instances on node runtime level?
> > > My primarily concern is that we can probably have individual constraint limits for working group instances and that can result in different worst-case scenarios. https://github.com/Joystream/joystream/blob/olympia/runtime/src/lib.rs#L620
> > 
> > 
> > I agree. I don't know if it's possible however to run multiple benchmarks for the same pallet. I will try.
> 
> Can be done, will upload the changes ASAP.

After implementing the way I found to do it, I found a problem where the weight-generated files can't be used without manually modifying them I will research more.

conectado:
> > > > Just one comment here.
> > > > Wouldn`t it be safer to have separate benchmarks for all working group instances on node runtime level?
> > > > My primarily concern is that we can probably have individual constraint limits for working group instances and that can result in different worst-case scenarios. https://github.com/Joystream/joystream/blob/olympia/runtime/src/lib.rs#L620
> > > 
> > > 
> > > I agree. I don't know if it's possible however to run multiple benchmarks for the same pallet. I will try.
> > 
> > 
> > Can be done, will upload the changes ASAP.
> 
> After implementing the way I found to do it, I found a problem where the weight-generated files can't be used without manually modifying them I will research more.

Until [this PR](https://github.com/paritytech/substrate/pull/7669) which solves this problem lands in substrate and we migrate to a version that includes it, we will either use the worst-case instance for benchmarking and if there's no such obvious instance we will include a mock instance representing the worst-case.

For this PR the worst-case instance is `ContentDirectoryWorkingGroup`

runtime-modules/working-group/src/lib.rs
```diff
@@ -469,7 +563,23 @@ decl_module! {
...
-        #[weight = 10_000_000] // TODO: adjust weight
+        /// # <weight>
+        ///
+        /// ## Weight
+        /// `O (P)` where:
+        /// - `P` is the length of `penalty.slashing_text`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = WeightInfoWorkingGroup::<T, I>::terminate_role_lead(
```
shamil-gadelshin: Consider moving the weight-code to a separate function.

runtime-modules/working-group/src/lib.rs
```diff
@@ -767,7 +941,23 @@ decl_module! {
...
+        ///
+        /// ## Weight
+        /// `O (S)` where:
+        /// - `S` is the length of the contents of `status_text` when it is not none
+        ///
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = WeightInfoWorkingGroup::<T, I>::set_status_text(
+            if let Some(status_text) = status_text {
```
shamil-gadelshin: Consider changing this snippet to a functional style: using Option.map with unwrap_or_default().

runtime-modules/working-group/src/lib.rs
```diff
@@ -58,6 +59,36 @@ pub use checks::{ensure_origin_is_active_leader, ensure_worker_exists, ensure_wo
 use common::origin::ActorOriginValidator;
 use staking_handler::StakingHandler;
 
+type WeightInfoWorkingGroup<T, I> = <T as Trait<I>>::WeightInfo;
+
+/// working_group WeightInfo
```
shamil-gadelshin: Please, change this auto-generated comment to `Working group WeightInfo`.

runtime-modules/working-group/src/lib.rs
```diff
@@ -469,7 +563,23 @@ decl_module! {
...
+        ///
+        /// ## Weight
+        /// `O (P)` where:
+        /// - `P` is the length of `penalty.slashing_text`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = WeightInfoWorkingGroup::<T, I>::terminate_role_lead(
+                    penalty.as_ref().map(|penalty| penalty.slashing_text.len().saturated_into())
+                    .unwrap_or_else(|| 0)
```
shamil-gadelshin: Please, change to unwrap_or_default()

runtime-modules/working-group/src/lib.rs
```diff
@@ -469,7 +563,23 @@ decl_module! {
...
+                )
+                .max(WeightInfoWorkingGroup::<T, I>::terminate_role_worker(
+                        penalty.as_ref().map(|penalty| penalty.slashing_text.len().saturated_into())
+                            .unwrap_or_else(|| 0)
```
shamil-gadelshin: Please, change to unwrap_or_default()

shamil-gadelshin (changes requested):

shamil-gadelshin (approved):
LGTM

PR #1868, opened 2020-12-04
Fix benchmarks after membership changes

conectado:
Benchmarks broke in olympia after membership changes, where `add_screened_member` and `set_screening_authority` were removed.

This PR fixes it by using `buy_membership` instead.


shamil-gadelshin (approved):
LGTM

PR #1870, opened 2020-12-04
Drop obsolete pallets and fix tests.

shamil-gadelshin:
### Pallets dropped
- stake
- hiring

### Comments
- I also discovered and fixed corrupted tests (the result of the invalid merge)
- `token-minting` and `recurring-rewards` pallets will be removed after the new governance pallet integration


bedeho (approved):
LGTM.

PR #1872, opened 2020-12-04
Council mock events

### LINKED ISSUE #1853: Deposit event anomaly in council
For a yet unknown reason, the current council version doesn't emit events as expected. Events can't be retrieved once deposited. If we add `panic!` right after event deposition and print events, no events will be present. 
```
// emit event
Self::deposit_event(RawEvent::NewCandidate(membership_id, stake));
panic!("{:?}", <system::Module<T>>::events()); // outputs empty array `[]`
```
Strangely enough `<system::Module<T>>::event_count()` seems to be increasing correctly. The problem is exclusive to the council module with no obvious differences from other modules.

ondratra:
This PR fixes #1853 and adds two previously missing tests for council extrinsics.

### DIFF
--- a/runtime-modules/council/src/mock.rs
+++ b/runtime-modules/council/src/mock.rs
@@ -66,7 +66,7 @@ parameter_types! {
 impl Trait for Runtime {
     type Event = TestEvent;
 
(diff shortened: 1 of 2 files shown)


bedeho (approved):
Nice work, how did you get to the bottom of this?

Issue #1861, opened 2020-12-03
Add benchmarking for FRAME's pallets


conectado:
Currently, for FRAME's pallets, we are using the default weights. 

We should benchmark them ourselves to take the particularities of our runtime into account.

The goal of this issue is to add them to our benchmarks and update the runtime to point to our own weights.

shamil-gadelshin:
Done

PR #1879, opened 2020-12-07
Add frame benchmarks

conectado:
# Adressing #1861 

* Adds FRAME's pallet to benchmarks
* Fix an issue with working-group benchmarking and membership
* Simplifies `generate-weights` script


runtime/src/weights/mod.rs
```diff
@@ -15,15 +15,17 @@
 
 //! A list of the different weight modules for our runtime.
 
+// FRAME pallets
 pub mod frame_system;
 pub mod pallet_balances;
+pub mod pallet_constitution;
```
shamil-gadelshin: pallet_constitution is a Joystream pallet

runtime/src/runtime_api.rs
```diff
@@ -284,7 +294,19 @@ impl_runtime_apis! {
             let mut batches = Vec::<BenchmarkBatch>::new();
             let params = (&config, &whitelist);
 
-            add_benchmark!(params, batches, system, SystemBench::<Runtime>);
+            // Note: For benchmarking Stake and Balances we need to change ExistentialDeposit to
+            // a non-zero value.
+            // For now, due to the complexity grandpa and babe aren't benchmarked automatically
+            // we should use the default manually created weights.
+            // Finally, pallet_offences have no `WeightInfo` so there's no need to benchmark it
+            // the benchmark is only for illustrative pourpuses.
```
shamil-gadelshin, line 302: What did you mean by this? - "the benchmark is only for illustrative purposes."

runtime-modules/working-group/src/benchmarking.rs
```diff
@@ -183,20 +183,16 @@ fn member_funded_account<T: Trait<I>, I: Instance>(
...
-        account_id.clone(),
+    Membership::<T>::buy_membership(
+        RawOrigin::Signed(account_id.clone()).into(),
         Some(handle),
         None,
         None,
     )
     .unwrap();
 
+    let _ = Balances::<T>::make_free_balance_be(&account_id, BalanceOf::<T>::max_value());
```
shamil-gadelshin, line 194: why do you need to replenish the balance twice?

scripts/generate-weights.sh
```diff
@@ -2,22 +2,18 @@
...
+    --steps=50 \
+    --repeat=20 \
+    --execution=wasm \
+    --output=.  > /dev/null
 
-echo "Benchmarking pallet_constitution..."
-./target/release/joystream-node benchmark --pallet=pallet_constitution --extrinsic=* --chain=dev --steps=50 --repeat=20 --execution=wasm --output=. > /dev/null
-mv pallet_constitution.rs runtime/src/weights/
-echo "pallet_constitution benchmarked"
+mv $SCRIPT_DIR/../*.rs $SCRIPT_DIR/../runtime/src/weights/
```
shamil-gadelshin: Please, handle errors or change the final comment.

scripts/generate-weights.sh
```diff
@@ -2,22 +2,18 @@
...
-echo "proposals_discussion benchmarked"
+echo "Generating weights..."
 
-echo "Benchmarking proposals_engine..."
-./target/release/joystream-node benchmark --pallet=proposals_engine --extrinsic=* --chain=dev --steps=50 --repeat=20 --execution=wasm --output=. > /dev/null
-mv proposals_engine.rs runtime/src/weights/
-echo "proposals_engine benchmarked"
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+$SCRIPT_DIR/../target/release/joystream-node benchmark \
+    --pallet=* \
```
shamil-gadelshin: This command gets stuck with pallet=*. I waited for 40-50min. But it worked with single pallet in current build (I tried pallet_constitution).

shamil-gadelshin (changes requested):

runtime/src/runtime_api.rs
conectado: That since `pallet_offences` itself doesn't have a `WeightInfo` but it's the `OffenceHandler` that should report the weight, the benchmarks that are implemented inside that pallet are only to show how the benchmarking would be done.

Here's the answer I got from the Substrate team when I asked about the benchmarks for this pallet:

> `pallet_offences` itself does not have extrinsics thus does not need any weight info on its own. The benchmarks there are for informational purposes.
> The `OnOffenceHandler` returns a weight which is used for the `on_initialize` calculation, and that lives in each pallet which implements that trait.

runtime-modules/working-group/src/benchmarking.rs
conectado: The first time is to make sure that there is enough funding to buy the membership no matter how much it costs. The second case is to make sure the account is fully funded after the membership fee is discounted, which is needed for some border cases to get to the worst-case scenario.

scripts/generate-weights.sh
conectado: I think this is the fault of `pallet_im_online` with just 5 steps and 2 repeats it takes 44seconds in my computer. If you reduce in this script to 5 steps and 2 repeats it'll probably finish in around a minute.

And if we extrapolate this to 50 steps with 20 repeats(around 100 times more) it would take 440 seconds, meaning ~1 and a half hours. I'll leave it running today to confirm how much it'd take to finish.

runtime-modules/working-group/src/benchmarking.rs
shamil-gadelshin: Ok.

runtime/src/runtime_api.rs
shamil-gadelshin: Ok.

shamil-gadelshin (approved):
LGTM

PR #1880, opened 2020-12-08
Council staking handler

ondratra:
This PR integrates StakingHandler pallet into council pallet.

To minimize merge conflict potential, this PR assumes #1820 will be merged first.


ondratra:
Issues with current StakingHandler implementation:
- `is_account_free_of_conflicting_stakes()` checks if the same lock type already exists, but I think it should actually expect Runtime to do this check and (dis)approve certain combinations of locks
- `is_member_staking_account` is not actually implemented (returns `true`)
- `set_stake` unlocks and locks balance again in one certain case; this can be prevented by check for balance at the start

ondratra:
- `is_account_free_of_conflicting_stakes` - lock conflict checks moved to separate trait that is implemented by runtime now
- `is_member_staking_account` - will be fixed when the next version of the membership module is finished
- `set_stake` - re-locking of staking lock prevented

runtime-modules/staking-handler/src/lib.rs
```diff
@@ -11,66 +11,86 @@ use frame_support::traits::{Currency, Get, LockIdentifier, LockableCurrency, Wit
...
-pub type MemberId<T> = <T as membership::Trait>::MemberId;
-
-/// Balance alias for `balances` module.
-pub type BalanceOf<T> = <T as pallet_balances::Trait>::Balance;
+/// Trait for (dis)allowing certain stake locks combinations.
+pub trait LockComparator<Balance> {
+    /// Checks if stake lock that is about to be used is conflicting with existing locks.
+    fn are_locks_conflicting(
+        new_lock: &LockIdentifier,
+        existing_locks: &[BalanceLock<Balance>],
```
shamil-gadelshin: Why do you pass the whole BalanceLock instead of just the LockId ?

runtime/src/constants.rs
```diff
@@ -35,6 +38,40 @@ pub const DAYS: BlockNumber = HOURS * 24;
...
+parameter_types! {
+    pub const ProposalsLockId: LockIdentifier = [5; 8];
+    pub const StorageWorkingGroupLockId: LockIdentifier = [6; 8];
+    pub const ContentWorkingGroupLockId: LockIdentifier = [7; 8];
+    pub const ForumGroupLockId: LockIdentifier = [8; 8];
+}
+
+lazy_static! {
+    // pairs of allowed lock combinations
+    pub static ref ALLOWED_LOCK_COMBINATIONS: BTreeSet<(LockIdentifier, LockIdentifier)> = [
```
shamil-gadelshin, line 50: This is a good idea to maintain such a table! However, I think that we could get away with much less coding. I suggest maintaining pairs `(LockId, Vec<LockId>)`. This would quickly show what locks allow concurrency. And also, it is easy to set up - because it would be just lines from the 'staking reusability relationships' table of the handbook. What do you think?

shamil-gadelshin (changes requested):
Very nice handling the Staking dependencies! I suggested just a couple of improvements.

runtime/src/constants.rs
ondratra: That was also my initial idea, but I didn't like the necessity to iterate through vector on each check. Binary tree set lookup should be more performance effective. On the other hand, it is definitely easier to define/inspect the table with a vector, as you suggested. 

I will rewrite it so we can definite the table like `(LockId, Vec<LockId>)`, but it's immediately recalculated to `BTreeSet` for faster lookup.

runtime-modules/staking-handler/src/lib.rs
ondratra: I wanted to give Runtime a possibility to take into account the amount locked (and possibly even the `Reasons`), but maybe it is overengineering.

ondratra: I've replaced it with `LockIdentifier`. In case we ever need `BalanceLock` there, we can return it. Now it is not needed there.

runtime/src/lib.rs
```diff
@@ -586,6 +587,18 @@ impl forum::Trait for Runtime {
     }
 }
 
+impl LockComparator<<Runtime as pallet_balances::Trait>::Balance> for Runtime {
+    fn are_locks_conflicting(new_lock: &LockIdentifier, existing_locks: &[LockIdentifier]) -> bool {
+        for lock in existing_locks {
```
shamil-gadelshin: Could be simplified down to any() method.

ondratra: The section has been simplified. Also, there was a bug returning exactly the opposite of what `are_locks_conflicting` should return. Fixed.

shamil-gadelshin (approved):
LGTM

PR #1881, opened 2020-12-08
Integration tests refactoring

mnaamani:
Some basic preparation for larger refactoring of the testing framework:

Enabling `"@polkadot/api/augment": [ "../../types/augment-codec/augment-api.ts"]` in tsconfig and fixing issues found as a result. Simplification mostly.

Some small fixes for running test scripts.

### DIFF
--- a/query-node/package.json
+++ b/query-node/package.json
@@ -18,7 +18,8 @@
 		"db:indexer:migrate": "(cd ./generated/indexer && yarn db:migrate)",
 		"codegen:indexer": "yarn hydra-cli codegen --no-install --no-graphql && cp indexer-tsconfig.json generated/indexer/tsconfig.json",
 		"codegen:server": "yarn hydra-cli codegen --no-install --no-indexer",
-		"cd-classes": "ts-node scripts/get-class-id-and-name.ts"
+		"cd-classes": "ts-node scripts/get-class-id-and-name.ts",
+		"integration-tests": "./run-tests.sh"
 	},
 	"author": "",
 	"license": "ISC",
--- a/query-node/run-tests.sh
+++ b/query-node/run-tests.sh
@@ -29,9 +29,6 @@ function cleanup() {
 
 trap cleanup EXIT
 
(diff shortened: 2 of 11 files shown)


metmirr (approved):
Looks good! I successfully ran `query-node` and `storage-node` tests. Btw on my machine, I needed to increase the wait to `3(120000)` minutes for the storage node tests. I can make a separate PR for that.

Just a comment on query-node integration tests, would it be good if we add yarn script for running test `yarn workspace query-node-root integration-tests` (currently, we run `./query-node/run-tests.sh`)

mnaamani:
> Looks good! I successfully ran `query-node` and `storage-node` tests. Btw on my machine, I needed to increase the wait to `3(120000)` minutes for the storage node tests. I can make a separate PR for that.
> 
> Just a comment on query-node integration tests, would it be good if we add yarn script for running test `yarn workspace query-node-root integration-tests` (currently, we run `./query-node/run-tests.sh`)

Sure I can add that.

The issue with delay is really problematic. It really depends on network conditions at the time of running the test, because it depends on how quickly ipns name registration occurs. I'll bump it in the script to 3m. In the github workflow for the storage node I actually did a retry loop: https://github.com/Joystream/joystream/blob/babylon/.github/workflows/run-network-tests.yml#L212

Maybe in future I can add similar approach.

metmirr:
Ohh sorry, I didn't make it clear what I mean by storage node tests. I had problem with querying video from query node it wasn't related to storage-node directly. So I increase time here https://github.com/Joystream/joystream/blob/1ee71fe41117d317fa8f187fffe1df124f0737d9/tests/network-tests/src/flows/storageNode/getContentFromStorageNode.ts#L13

mnaamani:
Oh thanks for clarifying. I'll push another update, it also doesn't hurt to have bumped that other delay also :)

metmirr:
> Oh thanks for clarifying. I'll push another update, it also doesn't hurt to have bumped that other delay also :)

Thanks :)

Issue #1781, opened 2020-11-20
Add a working group for the membership module


shamil-gadelshin:
The description could be found [here](https://joystream.gitbook.io/joystream-handbook/subsystems/membership).

Additionally, migrate errors to the `decl_error` style.

shamil-gadelshin:
Done.

PR #1883, opened 2020-12-08
Add membership working group.

shamil-gadelshin:
### Changes
- membership working group was added with all dependencies
- membershp pallet error style was changed from the strings to the `decl_error` macro.
- runtime version was bumped
- chain-spec-builder version was bumped
- `MemberId` and `ActorId` were moved to the newly created `common::Trait` from the membership module in order to remove cyclic dependencies with the working group pallet. As a result, the dependency on the membership module is decreased among the runtime pallets.
- `working_group::Interface` trait prototype was added 

[Main PR issue](https://github.com/Joystream/joystream/issues/1781)

### Comments
- The usage of the working group is different in the pallets Storage, Content Directory, Forum (and Membership)
- `working_group::Interface` trait is the prototype for the [working-group interface unification](https://github.com/Joystream/joystream/issues/1884)

### DIFF
--- a/runtime-modules/common/src/working_group.rs
+++ b/runtime-modules/common/src/working_group.rs
@@ -12,6 +12,8 @@ pub enum WorkingGroup {
     Forum,
     /// Storage working group: working_group::Instance2.
     Storage,
-    /// Storage working group: working_group::Instance3.
+    /// Content directory working group: working_group::Instance3.
     Content,
+    /// Membership working group: working_group::Instance3.
+    Membership,
 }
--- a/runtime-modules/common/src/working_group.rs
+++ b/runtime-modules/common/src/working_group.rs
@@ -14,6 +14,15 @@ pub enum WorkingGroup {
     Storage,
     /// Content directory working group: working_group::Instance3.
     Content,
-    /// Membership working group: working_group::Instance3.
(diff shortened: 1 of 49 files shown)


runtime-modules/common/src/working_group.rs
```diff
@@ -12,6 +12,17 @@ pub enum WorkingGroup {
...
     Storage,
-    /// Storage working group: working_group::Instance3.
+    /// Content directory working group: working_group::Instance3.
     Content,
+    /// Membership working group: working_group::Instance4.
+    Membership,
+}
+
+/// Working group interface to use in the in the pallets with working groups.
+pub trait Interface<T: crate::Trait> {
```
bedeho: Can we give this a more semantically rich name? e.g. `WorkingGroupRoleChecker`?

bedeho (changes requested):
Phenomenal work, really nice.

runtime-modules/common/src/working_group.rs
shamil-gadelshin: Changed to WorkingGroupIntegration (WorkingGroup had a conflict with enum).

bedeho (approved):
LGTM.

Issue #1839, opened 2020-12-01
Drop input length validation in runtime upgrade proposal


bedeho:
ssia.

Issue #1833, opened 2020-12-01
Drop input length validation in signal proposal


bedeho:
ssia.

conectado:
I talked with @shamil-gadelshin about this issue, couldn't dropping this validation let us vulnerable to someone storing a huge value since the text will go into the `Engine` storage as part of the dispatchable code.

The same goes for #1839

shamil-gadelshin:
As a solution, we could replace the text with text hash for storage and for antispam protection use the transaction weight.

PR #1886, opened 2020-12-08
Remove length validations from Text and Runtime Upgrade proposals

conectado:
Tracking #1839 and #1833

### DIFF
--- a/runtime-modules/proposals/codex/src/lib.rs
+++ b/runtime-modules/proposals/codex/src/lib.rs
@@ -108,9 +108,6 @@ pub trait Trait:
     + governance::election::Trait
     + staking::Trait
 {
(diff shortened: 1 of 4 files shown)


shamil-gadelshin (approved):
LGTM

shamil-gadelshin:
Do we need to change text storing to hash storing for signal proposal in the scope of this task? @bedeho

bedeho:
> Do we need to change text storing to hash storing for signal proposal in the scope of this task? @bedeho

Good question, for now we do not have any systematic approach to this. I suggest we do not add any such storing of hashes until a later time, and then we do a full review. My hunch is that it is a bad idea to store it in all cases, even forum, because we are not ever reading it back. The thin client rationale is unlikely to materialize.

PR #1887, opened 2020-12-08
Babylon content dir changes

bwhm:


bwhm:
#### Changes

**Globally**
- `MaxNumberOfEntitiesPerClass` from 400 -> 5000
This makes it so we can have up to 5000 videos. Pretty arbitrary number, but 400 is way to low.
- `IndividualEntitiesCreationLimit` from 50 -> 500
This means any/all members are capped at 500 entities in each class. There are more restrictive limits for the individual classes.

**Class limits**
Channel:
-  `maximum_entities_count` from 400 -> 5000
In line with global limit
- `default_entity_creation_voucher_upper_bound` from 50 -> 25
A single member can "only" make 25 channels

Content Category:
- `maximum_entities_count` from 100 -> 500
- `default_entity_creation_voucher_upper_bound` from 50 -> 500
Only CC lead can create entities. 500 seems more than sufficient.
These numbers should be the same, although I'm not 100% sure if this means a new lead can be capped by the old lead(s).

Featured Video:
- `maximum_entities_count` from 400 -> 10
- `default_entity_creation_voucher_upper_bound` from 50 -> 10
Only CC lead can create entities. As the system works now, having too many featured videos makes no sense.

Known License:
- `maximum_entities_count` from 100 -> 500
- `default_entity_creation_voucher_upper_bound` from 50 -> 500
Only CC lead can create entities. 500 seems more than sufficient.

Language:
- `maximum_entities_count` from 100 -> 500
- `default_entity_creation_voucher_upper_bound` from 50 -> 500
Only CC lead can create entities. 500 seems more than sufficient.

HttpMediaLocation, JoystreamMediaLocation, License, UserDefinedLicense, Video, VideoMedia, MediaLocation
- `maximum_entities_count` from 400 -> 5000
- `default_entity_creation_voucher_upper_bound` from 50 -> 100
This makes it so that there can be 5000 videos in total. Each member can, by default, create 100 videos across their channels. The curator lead can at any point increase the limit for any member. It makes it a little more tricky to spam...

mnaamani:
> default_entity_creation_voucher_upper_bound from 50 -> 500
Only CC lead can create entities. 500 seems more than sufficient.
These numbers should be the same, although I'm not 100% sure if this means a new lead can be capped by the old lead(s).

As I'm interpreting the [implementation](https://github.com/Joystream/joystream/blob/b156248357f6b7c729968d810eee62c4aec8fec2/runtime-modules/content-directory/src/lib.rs#L2728), the check for the maximum entities comes before the "per_actor_constraint" check, and since you are making the voucher_upper_bound == maximum_entities_count == 500, then yes if new lead finds that the old lead created 500 categories, they will not be able to add more.

Otherwise LGTM

bwhm:
> > default_entity_creation_voucher_upper_bound from 50 -> 500
> > Only CC lead can create entities. 500 seems more than sufficient.
> > These numbers should be the same, although I'm not 100% sure if this means a new lead can be capped by the old lead(s).
> 
> As I'm interpreting the implementation, the check for the maximum entities comes before the "per leader" limit.
> Since you are making the voucher_upper_bound == maximum_entities_count , then yes if new lead finds that the old lead created 500 categories, they will not be able to add more.
> 
> Otherwise LGTM

What I haven't confirmed yet is whether a new lead is able to `removeEntities` created by the previous lead. For a non-member, it should be allowed, but haven't tested...

PR #1890, opened 2020-12-09
content-directory: set initial ids to 1 on runtime upgrade

mnaamani:
Next Id's are configured at genesis. 
Our applications and tools are hardcoded to expect initial values of the ids to start at 1.
With a runtime upgrade the initial values will not be configured and get an initial default value of zero. 
This corrects this problem by setting them to `one` on runtime upgrade.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2053,7 +2053,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node-runtime"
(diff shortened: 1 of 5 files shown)


iorveth (approved):
LGTM!

shamil-gadelshin (approved):
LGTM

bwhm (approved):

Issue #1841, opened 2020-12-01
DRY Proposal Creation


bedeho:
Every single proposal has its own creation function, and there is significant replication in the signatures due to the general proposal parameters, namely

```Rust
pub fn create_foo_proposal(
            origin,      
            member_id: MemberId<T>,
            title: Vec<u8>,
            description: Vec<u8>,
            staking_account_id: Option<T::AccountId>,
...)
```

This can be improved in two steps.

1. Introduce 

```Rust
struct GeneralProposalParameters {
            member_id: MemberId<T>,
            title: Vec<u8>,
            description: Vec<u8>,
            staking_account_id: Option<T::AccountId>,
}
```

now each function will instead by 

```Rust
pub fn create_foo_proposal(
            origin,      
            general_paramaters: GeneralProposalParameters
...)
```

Moreover, its likely to simplify some internal code also.

2. One could go one step further and just have a single creation extrinsic, which segments out the general and specific parameters, so that you dont need to repeate the substantial repetition in the body of each creator function, so 

```Rust

enum SpecificParameters {
  FundingRequestParameters(...),
  ...
}; 

pub fn create_proposal(origin, genera_parameters: GeneralParameters, specific_parameters: SpecificParameters) {

   let (proposal_parameters, proposal_code) = ensure_creation_checks(origin, genera_parameters, specific_parameters)?;

  let params = CreateProposalParameters{ ...proposal_parameters, proposal_code };

   /// Not clear that this standalone combined checker & mutator is even neded now, as its only called once, may
   /// may be cleaner to just inline here.
   Self::create_proposal(params)?;

}
```

bedeho:
It actually seems like `create_proposal` has multiple combined ensure+mutate calls in row, hence it can partially mutate storage, this is severe bug.

```Rust
 let initial_thread_mode = ThreadMode::Open;
        <proposals_discussion::Module<T>>::ensure_can_create_thread(&initial_thread_mode)?;

        let discussion_thread_id = <proposals_discussion::Module<T>>::create_thread(
            params.member_id,
            initial_thread_mode,
        )?;

        let proposal_creation_params = ProposalCreationParameters {
            account_id,
            proposer_id: params.member_id,
            proposal_parameters: params.proposal_parameters,
            title: params.title,
            description: params.description,
            staking_account_id: params.staking_account_id,
            encoded_dispatchable_call_code: params.proposal_code,
            exact_execution_block: params.exact_execution_block,
        };

[...]

PR #1891, opened 2020-12-09
DRY proposal creation

conectado:
# Addressing #1841 

This PR change `create_*_proposals` the following ways:
* Unify repeated parameters into `GeneralProposalParameters`
* Unify all extrinsics into `create_proposal`
  * Make use of already existing `ProposalDetals` to distinguish between specific proposals and pass specific parameters
    * **Note**: The issue originally suggested creating a new enum for these parameters, I took advantage of the already existing enum, let me know if it's preferable to do it the other way
  * Inline the old `create_proposal` function into the extrinsic
* Update tests to adapt to these changes

### DIFF
--- a/runtime-modules/proposals/codex/src/lib.rs
+++ b/runtime-modules/proposals/codex/src/lib.rs
@@ -52,12 +52,11 @@
 // Disable this lint warning because Substrate generates function without an alias for the ProposalDetailsOf type.
 #![allow(clippy::too_many_arguments)]
 
-mod proposal_types;
+mod types;
 
 #[cfg(test)]
 mod tests;
 
-use codec::{Decode, Encode};
 use frame_support::dispatch::DispatchResult;
 use frame_support::traits::{Currency, Get};
 use frame_support::{decl_error, decl_module, decl_storage, ensure, print};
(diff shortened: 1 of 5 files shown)


runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -85,20 +87,22 @@ const MAX_SPENDING_PROPOSAL_VALUE: u32 = 5_000_000_u32;
 // Max validator count for the 'set validator count' proposal
 const MAX_VALIDATOR_COUNT: u32 = 100;
 
-// Data container struct to fix linter warning 'too many arguments for the function' for the
-// create_proposal() function.
-struct CreateProposalParameters<T: Trait> {
-    pub origin: T::Origin,
-    pub member_id: MemberId<T>,
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Debug, Default, Clone, PartialEq, Eq)]
```
shamil-gadelshin: Please, move new structs to the `proposal_types` module and create comments for both structs and their fields. Also please, rename that module to just `types` and downgrade it to the file instead of the folder.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -735,45 +400,112 @@ decl_module! {
...
+                    Error::<T>::DecreasingStakeIsZero
+                );
+            }
+            ProposalDetails::SlashWorkingGroupLeaderStake(_, ref penalty, _) => {
+                ensure!(
+                    penalty.slashing_amount != Zero::zero(),
+                    Error::<T>::SlashingStakeIsZero
+                );
+            }
+            ProposalDetails::SetWorkingGroupLeaderReward(_, _, _) => {
```
shamil-gadelshin: Change all multiple blank entries (_, _) to (..) in enums.

shamil-gadelshin (changes requested):
I like this new pallet design and the work resulted in 200 lines reduction which is a good sign 😄 

Left just a couple of small comments.

shamil-gadelshin (approved):
LGTM

bedeho (approved):
Very nice work!

> The issue originally suggested creating a new enum for these parameters, I took advantage of the already existing enum, let me know if it's preferable to do it the other way

Nicely spotted, I saw them later also, but left as exercise for the reader, good work.

PR #1892, opened 2020-12-09
Query node entity remove fix

metmirr:
Replaces `entity not found` error with an informative debug message. There are cases like removing entities before adding schema support which causes the error. And some refactoring.

### DIFF
--- a/query-node/mappings/content-directory/entity/index.ts
+++ b/query-node/mappings/content-directory/entity/index.ts
@@ -91,7 +91,7 @@ async function contentDirectory_EntitySchemaSupportAdded(db: DB, event: Substrat
   const { blockNumber: block } = event
   const entityId = decode.stringIfyEntityId(event)
 
-  const knownClass = await getKnownClass(db, { where: { id: entityId } })
+  const [knownClass] = await getKnownClass(db, { where: { id: entityId } })
   if (!knownClass) return
 
   const arg: IDBBlockId = { db, block, id: entityId }
@@ -184,7 +184,7 @@ async function contentDirectory_EntityRemoved(db: DB, event: SubstrateEvent): Pr
   const entityId = decode.stringIfyEntityId(event)
   const where: IWhereCond = { where: { id: entityId } }
 
-  const knownClass = await getKnownClass(db, where)
+  const [knownClass, classEntity] = await getKnownClass(db, where)
   if (!knownClass) return
 
   switch (knownClass.name) {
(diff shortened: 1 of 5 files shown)


query-node/mappings/content-directory/entity/index.ts
```diff
@@ -194,19 +184,10 @@ async function contentDirectory_EntityRemoved(db: DB, event: SubstrateEvent): Pr
...
-    return
-  }
-
-  const cls = contentDirectoryClassNamesWithId.find((c) => c.classId === classEntity.classId)
-  if (cls === undefined) {
-    console.log('Unknown class')
-    return
-  }
+  const knownClass = await getKnownClass(db, where)
+  if (!knownClass) return
```
mnaamani, line 188: Isn't this just hiding the fact that the database is inconsistent with the state of the runtime?

query-node/mappings/content-directory/entity/index.ts
```diff
@@ -290,21 +273,17 @@ async function contentDirectory_EntityPropertyValuesUpdated(db: DB, event: Subst
...
-  const ce = await db.get(ClassEntity, { where: { id: entityId } })
-  if (ce === undefined) throw Error(`Class not found for the entity id: ${entityId}`)
-
-  const cls = contentDirectoryClassNamesWithId.find((c) => c.classId === ce.classId)
-  if (cls === undefined) throw Error(`Not known class id: ${ce.classId}`)
-
   const where: IWhereCond = { where: { id: entityId } }
 
+  const knownClass = await getKnownClass(db, where)
+  if (!knownClass) return
```
mnaamani, line 278: Here again, if we got an `EntityPropertyValuesUpdated` event how come we don't have the entity in the database?
We are effectively ignoring the event and our database is inconsistent with the runtime?

mnaamani: Had a call with @metmirr and it was pointed out that this return here is because the mappings isn't aware of arbitrary new classes which are created in the runtime and must be ignored here.

query-node/mappings/content-directory/entity/index.ts
mnaamani: Its only logical that mappings have to ignore unknown classes that may have been added arbitrarily.

query-node/mappings/content-directory/entity/index.ts
```diff
@@ -259,8 +240,10 @@ async function contentDirectory_EntityRemoved(db: DB, event: SubstrateEvent): Pr
       break
 
     default:
-      throw new Error(`Unknown class name: ${cls.name}`)
+      throw new Error(`Unknown class name: ${knownClass.name}`)
   }
+  const ce = await db.get(ClassEntity, where)
+  if (ce) await db.remove<ClassEntity>(ce)
```
mnaamani: Further to our call, given that code at this point is dealing with a known class, the entity must be in the `ClassEntity` table and `ce` should not be undefined. So this line should be turned into an assertion that `ce` is truthy followed by a call to remove it. Or get/remove operations can be changed to a single db operation to remove that fails or throws if the entity is not found?

PR #1897, opened 2020-12-10
Forum_v2 create_thread clean up: fix double lookup ensure_category_is_mutable

iorveth:
`ensure_category_is_mutable` check performed twice.


bedeho (approved):
Nice catch.

PR #1901, opened 2020-12-11
council - restriction on reward_account_id membership lifted

ondratra:
This PR removes `reward_account_id` membership requirement - rewards can be paid out to any account now. It also adds test for `staking_account_id` membership requirement and `reward_account_id` no requirement.


bedeho (approved):
LGTM.

PR #1909, opened 2020-12-14
Modify Membership module.

shamil-gadelshin:
#### Changes
- change `buy_membership` extrinsic:
  - add support for the referral cuts
  - add root and controller fields
  - refactor to use parameter struct
- modify update_profile extrinsics:
  - add `name` field
  - remove fields: registered_at_block, registered_at_time, entry
- add missing comments to the crate
- add update_accounts extrinsic instead of separate extrinsics.
- other changes:
  - remove fine grained separate extrinsics like `change_member_handle`
  - remove separate extrinsics for root and controller accounts update


runtime-modules/membership/src/genesis.rs
```diff
@@ -1,17 +1,19 @@
...
-pub struct Member<MemberId, AccountId, Moment> {
+pub struct Member<MemberId, AccountId> {
     pub member_id: MemberId,
     pub root_account: AccountId,
     pub controller_account: AccountId,
     pub handle: String,
     pub avatar_uri: String,
     pub about: String,
-    pub registered_at_time: Moment,
+    pub name: String,
```
bedeho, line 16: This is not needed to be stored.

runtime-modules/membership/src/genesis.rs
```diff
@@ -1,17 +1,19 @@
...
 use crate::{GenesisConfig, Trait};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Serialize, Deserialize)]
-pub struct Member<MemberId, AccountId, Moment> {
+pub struct Member<MemberId, AccountId> {
     pub member_id: MemberId,
     pub root_account: AccountId,
     pub controller_account: AccountId,
     pub handle: String,
```
bedeho, line 13: I think we can make this a hash actually, same for key in `MemberIdByHandle`. The actual value is irrelevant for any of the runtime processing, its only uniqueness that matters.

runtime-modules/membership/src/lib.rs
```diff
@@ -1,3 +1,43 @@
+//! Joystream membership module.
```
bedeho, line 1: Nice!

runtime-modules/membership/src/lib.rs
```diff
@@ -81,60 +114,78 @@ pub struct MembershipObject<BlockNumber, Moment, AccountId> {
...
 }
 
-#[derive(Encode, Decode, Debug, PartialEq)]
-pub enum EntryMethod {
-    Paid,
-    Genesis,
-}
+/// Parameters for the buy_membership extrinsic.
+#[derive(Encode, Decode, Default, Clone, PartialEq, Debug)]
+pub struct BuyMembershipParameters<AccountId, MemberId> {
```
bedeho, line 125: Why is this type needed?

runtime-modules/membership/src/lib.rs
```diff
@@ -34,35 +76,26 @@ const DEFAULT_MIN_HANDLE_LENGTH: u32 = 5;
...
->;
+/// Public membership profile alias.
+pub type Membership<T> = MembershipObject<<T as frame_system::Trait>::AccountId>;
 
 #[derive(Encode, Decode, Default)]
-/// Stored information about a registered user
-pub struct MembershipObject<BlockNumber, Moment, AccountId> {
-    /// The unique handle chosen by member
+/// Stored information about a registered user.
+pub struct MembershipObject<AccountId> {
```
bedeho, line 86: This name makes me sad, why not just `Membership`?

runtime-modules/membership/src/lib.rs
```diff
@@ -144,38 +195,49 @@ decl_storage! {
...
             T::AccountId => Vec<T::MemberId>;
 
-        /// Registered unique handles and their mapping to their owner
+        /// Registered unique handles and their mapping to their owner.
         pub MemberIdByHandle get(fn handles) : map hasher(blake2_128_concat)
             Vec<u8> => T::MemberId;
 
-        /// Is the platform is accepting new members or not
+        /// Is the platform is accepting new members or not.
         pub NewMembershipsAllowed get(fn new_memberships_allowed) : bool = true;
```
bedeho, line 215: Can you create a new issue to drop this? I see no real case for it.

runtime-modules/membership/src/lib.rs
```diff
@@ -144,38 +195,49 @@ decl_storage! {
...
 
-        // User Input Validation parameters - do these really need to be state variables
-        // I don't see a need to adjust these in future?
+        /// Minimum allowed handle length.
         pub MinHandleLength get(fn min_handle_length) : u32 = DEFAULT_MIN_HANDLE_LENGTH;
+
```
bedeho, line 219: All of these length constraints, we decided to not do them right? because value is not stored. Even for handle we dont need it because of my prior comment to store hash only, not full handle.

If correct, then please address in standalone PR, as this is not part of this PR scope.

runtime-modules/membership/src/lib.rs
```diff
@@ -221,154 +276,180 @@ decl_module! {
...
-            let user_info = Self::check_user_registration_info(handle, avatar_uri, about)?;
+            // Verify user parameters.
+            let user_info = Self::check_user_registration_info(
+                params.name,
+                params.handle,
+                params.avatar_uri,
+                params.about)
+            ?;
+
+            let mut referrer: Option<Membership<T>> = None;
```
bedeho: This can be done without mutable.

bedeho (approved):

runtime-modules/membership/src/lib.rs
shamil-gadelshin: The linter complains about too many arguments for a function.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Because `Membership` is a trait bound alias of the `MembershipObject` which is used in the system. It is similar to Balance and BalanceOf<T>.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: I can drop it in the next PR (current work).

runtime-modules/membership/src/lib.rs
shamil-gadelshin: I was unsure about it but turned out you are correct! A small 'func-fu' trick made it. The code snippet is much nicer now.

runtime-modules/membership/src/genesis.rs
shamil-gadelshin: Ok. Will do within https://github.com/Joystream/joystream/issues/1927

runtime-modules/membership/src/genesis.rs
shamil-gadelshin: Ok. Will do within https://github.com/Joystream/joystream/issues/1927

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Ok. Will do within https://github.com/Joystream/joystream/issues/1927

runtime-modules/membership/src/lib.rs
bedeho: LOL!

bedeho (approved):

PR #1910, opened 2020-12-14
Sudo init working group leads

mnaamani:
Modified the content directory schemas `initialize:dev` script to be configurable with two env variables to for sudo and lead key pairs so it can be re-used on production chain.

example:

```
SUDO_URI=//Alice LEAD_URI=//Alice yarn workspace @joystream/cd-schemas initialize:dev
```
I had previously also made a similar addition to the storage node cli commands to also initialize a lead storage provider on a production chain so I included it that work in this PR.

```
SUDO_URI=//Alice MEMBER_ID=1 MEMBER_URI=//Bob WORKER_URI=//Bob \
    yarn storage-cli sudo-create-sp
```


storage-node/packages/cli/src/commands/dev.ts
```diff
@@ -142,4 +142,101 @@ const init = async (api: RuntimeApi): Promise<any> => {
...
+
+  api.workers.devBeginStorageOpeningReview(openingId)
+
+  debug(`Filling storage opening.`)
+  const providerId = await api.workers.devFillStorageOpening(openingId, applicationId)
+
+  debug(`Assigned storage provider id: ${providerId}`)
+}
+
+function getLeadOpeningInfo() {
```
Lezek123: It should be possible to use `GenericJoyStreamRoleSchema` type from `@joystream/hiring/schemas/role.schema.typings.ts` here for typesafety and then convert to json with `JSON.strigify`.

storage-node/packages/cli/src/commands/dev.ts
```diff
@@ -142,4 +142,101 @@ const init = async (api: RuntimeApi): Promise<any> => {
   return check(api)
 }
 
-export { init, check, aliceKeyPair, roleKeyPair, developmentPort }
+// Using sudo create initial storage lead and worker with given keys taken from env variables.
+// Used to quickly setup a storage provider on a new chain before a council is ready.
+const makeMemberInitialLeadAndStorageProvider = async (api: RuntimeApi): Promise<any> => {
+  const sudoKey = getKeyFromAddressOrSuri(api, process.env.SUDO_URI)
```
Lezek123, line 170: When I don't provide the env variables I'm getting `Command Failed: TypeError: Cannot read property 'startsWith' of undefined` (It looks like it originates from `keyring.addFromUri` in `getKeyFromAddressOrSuri`)

storage-node/packages/cli/src/cli.ts
```diff
@@ -72,6 +72,9 @@ const commands = {
   'dev-check': async (api) => {
     return dev.check(api)
   },
+  'sudo-create-sp': async (api) => {
```
Lezek123, line 75: The `usage` const is not updated and still includes the old `vstore-init` command and no `sudo-create-sp`.

storage-node/packages/cli/src/commands/dev.ts
```diff
@@ -142,4 +142,101 @@ const init = async (api: RuntimeApi): Promise<any> => {
...
+  const memberId = parseInt(process.env.MEMBER_ID)
+  const memberController = getKeyFromAddressOrSuri(api, process.env.MEMBER_URI).address
+  const leadAccount = memberController
+  const workerAccount = process.env.WORKER_ACCOUNT
```
Lezek123: At the beginning I tried using `WORKER_URI` as provided in PR description and this resulted in some arbitrary(?) role key (`5C4hrf...yUpnhM`).

When using `WORKER_ACCOUNT=5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty` (Bob) the result is as expected.

Lezek123 (commented):
Mostly works as expected, I added a few small comments regarding the storage CLI command.

storage-node/packages/cli/src/commands/dev.ts
mnaamani: Thanks, will update to use this technique.

storage-node/packages/cli/src/commands/dev.ts
mnaamani: Looks like I didn't do good input validation here. Will fix.

storage-node/packages/cli/src/cli.ts
mnaamani: will update.

storage-node/packages/cli/src/commands/dev.ts
mnaamani: Yes I think it derived account from an empty string maybe. The description in the PR was incorrect. :)
I'll make some general improvements on input validation.

mnaamani:
Applied suggested fixes and improved input validation. 
Renamed some env variables for the `sudo-create-sp` command:

```
SUDO_URI
MEMBER_ID
MEMBER_CONTROLLER_URI
STORAGE_WORKER_ADDRESS
```

Lezek123 (approved):
Works as expected.
The only small issue is that it fails with `SudoFailed` if the lead is already initialized, perhaps a simple check if lead is already set can be added before sending the sudo tx?

mnaamani:
> Works as expected.
> The only small issue is that it fails with `SudoFailed` if the lead is already initialized, perhaps a simple check if lead is already set can be added before sending the sudo tx?

Yes I tested that as well, might as well add the check.

bwhm (approved):
lgtm!

Issue #1799, opened 2020-11-23
New commands to CLI - `content-directory`


bwhm:
## `content-directory:createEntity`
Although there _might_ be some use case for just creating an entity, I think we should batch it with `addSchemaSupportToEntity`.

I think most of this could be copypasta from `media:uploadVideo`/`media:createChannel`...

#### Use

`joystream-cli content-directory:createEntity --help`
```
Creates a new entity in the specified class (can be executed in Member, Curator or Lead context)

USAGE
   $ joystream-cli content-directory:createEntity CLASSNAME

ARGUMENTS
   CLASSNAME  Name or ID of the Class

OPTIONS
   --context=(Member|Curator|Lead)  Actor context to execute the command in (Member/Curator/Lead)
```

#### Details
(Note that it should follow the flow in similar CLI commands, this is just how I think it could look. A simplified version might be fine)
1. Check if the user has permission to create an entity in this class
2. Read the class, and fetch the properties info (from the highest indexed schema)
3. Ideally prompt the user for all the properties, similar to `media:uploadVideo`/`media:createChannel`, but less verbose.
   - if not required, default to `null`(?)
   - take text or numbers when possible
   - for single reference, include the "valid" class - don't think it's worth the effort to fetch all the possible entities and allow the user to select
   - since `EntityVec` is not possible (?), same as above
   - for `Bool`, it may be simple enough to use a list :)
4. Display the parsed input, and prompt for confirmation at the end

## `content-directory:updateEntityPropertyValues`

I think most of this could be copypasta from `media:updateVideo`/`media:updateChannel`.

#### Use

`joystream-cli content-directory:updateEntityPropertyValues --help`
```
Updates the property values of the specified entity (can be executed in Member, Curator or Lead context)

USAGE
   $ joystream-cli content-directory:updateEntityPropertyValues ID

ARGUMENTS
  ID  ID of the Entity

OPTIONS
   --context=(Member|Curator|Lead)  Actor context to execute the command in (Member/Curator/Lead)
```

#### Details
1. Check if the user has permission to update the entity
2. Fetch the entity and its properties
3. Prompt the user to (re-)populate all the values, similar to `media:updateVideo`/`media:updateChannel`, defaulting to current value.
4. Display the parsed input, and prompt for confirmation at the end.

PR #1912, opened 2020-12-14
createEntity and updateEntityPropertyValues CLI commands 

DzhideX:
This PR aims to implement issue #1799.

One thing I'd like to add is that a lot of code is shared between the two files. I didn't know what the best way to factor this out would be so I'm open for suggestions!

### DIFF
--- a/cli/src/commands/content-directory/createEntity.ts
+++ b/cli/src/commands/content-directory/createEntity.ts
@@ -1,6 +1,7 @@
 import ContentDirectoryCommandBase from '../../base/ContentDirectoryCommandBase'
 import inquirer from 'inquirer'
 import { InputParser } from '@joystream/cd-schemas'
+import ExitCodes from '../../ExitCodes'
 
 export default class CreateEntityCommand extends ContentDirectoryCommandBase {
   static description =
@@ -20,7 +21,11 @@ export default class CreateEntityCommand extends ContentDirectoryCommandBase {
 
   async run() {
     const { className } = this.parse(CreateEntityCommand).args
-    const { context } = this.parse(CreateEntityCommand).flags
+    let { context } = this.parse(CreateEntityCommand).flags
+
+    if (!context) {
+      context = await this.promptForContext()
+    }
 
     const currentAccount = await this.getRequiredSelectedAccount()
     await this.requestAccountDecoding(currentAccount)
@@ -28,9 +33,13 @@ export default class CreateEntityCommand extends ContentDirectoryCommandBase {
 
     const actor = await this.getActor(context, entityClass)
 
+    if (actor.isOfType('Member') && entityClass.class_permissions.any_member.isFalse) {
+      this.error('Choosen actor has no access to create an entity of this type', { exit: ExitCodes.AccessDenied })
+    }
+
     const answers: {
       [key: string]: string | number | null
-    } = await inquirer.prompt(this.getQuestionsFromClass(entityClass))
+    } = await inquirer.prompt(this.getQuestionsFromProperties(entityClass.properties.toArray()))
 
     this.jsonPrettyPrint(JSON.stringify(answers))
     await this.requireConfirmation('Do you confirm the provided input?')
--- a/cli/src/commands/content-directory/createEntity.ts
+++ b/cli/src/commands/content-directory/createEntity.ts
@@ -1,0 +1,170 @@
+import ContentDirectoryCommandBase from '../../base/ContentDirectoryCommandBase'
+import { flags } from '@oclif/command'
+import { Class as ContentDirectoryClass } from '@joystream/types/content-directory'
+import inquirer from 'inquirer'
+import { CLIError } from '@oclif/errors'
+import { InputParser } from '@joystream/cd-schemas'
+
+const CONTEXTS = ['Member', 'Curator', 'Lead'] as const
+
+type CreateEntityArgs = {
+  className: string
+}
+
+type CreateEntityFlags = {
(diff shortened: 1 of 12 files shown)


cli/src/commands/content-directory/createEntity.ts
```diff
@@ -0,0 +1,170 @@
+import ContentDirectoryCommandBase from '../../base/ContentDirectoryCommandBase'
+import { flags } from '@oclif/command'
+import { Class as ContentDirectoryClass } from '@joystream/types/content-directory'
+import inquirer from 'inquirer'
+import { CLIError } from '@oclif/errors'
+import { InputParser } from '@joystream/cd-schemas'
+
+const CONTEXTS = ['Member', 'Curator', 'Lead'] as const
```
Lezek123: I don't think there's a need to re-define `CONTEXT` const and `context` flag here, since they are already available in `ContentDirectoryCommandBase` (see how `content-directory:removeEntity` implements this).

I noticed you also re-define them (along with some methods like `getActor`, `getQuestionsFromClass`) in `updateEntityPropertyValues.ts`.

Anything that is reused by 2 or more commands can be put in `ContentDirectoryCommandBase` and either imported from there or accessed through `this`.

cli/src/commands/content-directory/createEntity.ts
```diff
@@ -0,0 +1,170 @@
...
+
+type CreateEntityArgs = {
+  className: string
+}
+
+type CreateEntityFlags = {
+  context: typeof CONTEXTS[number]
+}
+
+type Question = {
```
Lezek123: Can we instead use a type that comes directly from `inquirer` library like `DistinctQuestion`, ie.:
https://github.com/Joystream/joystream/blob/1ee71fe41117d317fa8f187fffe1df124f0737d9/cli/src/base/DefaultCommandBase.ts#L3

I suspect this type may cause some issues if we just use it as `DistinctQuestion[]`, since `inquirer` uses quite complex generic typings and I ran into some issues with that in the past too, but if we could use some type that comes directly from the library - that would be the best approach.

Otherwise we should only define a type like this in one place (ie. `cli/src/Types.ts`)

cli/src/commands/content-directory/createEntity.ts
```diff
@@ -0,0 +1,170 @@
...
+    await this.requestAccountDecoding(currentAccount)
+    const [, entityClass] = await this.classEntryByNameOrId(className)
+
+    const actor = await this.getActor(context, entityClass)
+
+    const answers: {
+      [key: string]: string | number | null
+    } = await inquirer.prompt(this.getQuestionsFromClass(entityClass))
+
+    await this.checkReferencesValidity(answers, entityClass)
```
Lezek123: I think if we use inquirer question `validate` property for this we can validate it more "dynamically" instead of waiting until all questions are answered (and in that case the user wouldn't need to provide the entire input again in case of an error, which seems like a nice improvement)

cli/src/commands/content-directory/updateEntityPropertyValues.ts
```diff
@@ -0,0 +1,174 @@
...
+  static flags = {
+    context: flags.string({
+      char: 'c',
+      description: 'Actor context to execute the command in (Member/Curator/Lead)',
+      required: true,
+      options: [...CONTEXTS],
+    }),
+  }
+
+  getQuestionsFromClass = (
```
Lezek123: As I mentioned in the first comment - there seems to be a lot of functionality copy-pasted from `createEntity.ts` here, which is always best to avoid. Can this be moved to `ContentDirectoryCommandBase` for example?

cli/src/commands/content-directory/updateEntityPropertyValues.ts
```diff
@@ -0,0 +1,174 @@
...
+
+type Question = {
+  name: string
+  message: string
+  type: 'list' | 'input'
+  choices?: string[]
+  filter?: (answer: string) => string | null | number | boolean
+}
+
+export default class UpdateEntityPropertyValues extends ContentDirectoryCommandBase {
```
Lezek123, line 6: As far as permissions are concerned, here we would need to be careful about class properites that have `is_locked_from_maintainer` and `is_locked_from_contoller` flags set.

If we just prompt for values for all class properties and provide them in the final extrinsic, we may get an unavoidable runtime error in case the actor has no privilages to set value for given property (ie. `Channel.isCensored`)

Lezek123 (changes requested):
Overall this seems to work quite nicely!
> One thing I'd like to add is that a lot of code is shared between the two files. I didn't know what the best way to factor this out would be so I'm open for suggestions!

I added a few suggestions about that in the review and also a comment about `updateEntityPropertyValues` permissions (currently this command will always fail in context of `Channel` class, for example)

DzhideX:
Done!

Lezek123 (changes requested):
Good job! I created a PR to your branch (https://github.com/DzhideX/joystream/pull/2) with some small fixes (mostly related to `is_locked_from_controller` / `is_locked_from_maintainer` permissions) and some code cleanup. I did a few tests after that and everything seems to work nicely! (the only problem I noticed is that we cannot "unset" `optional` property value if it has been previously set, but that's a minor issue)

DzhideX:
Done. Did some final checks, everything seems to work great!

Lezek123 (approved):

bwhm (approved):
Tested, and works well! 
Nice :)

PR #1924, opened 2020-12-15
storage-node: anonymous (replication only) mode

mnaamani:
Added an `--anonymous` argument as option when running colossus server.
In this mode the storage node can run without need to be a registered provider on the network. Its function is limited to "replication only". It will not act as a storage liason. The main purpose of this option is to be able to serve content only.
For babylon release we will setup two load balanced nodes in this mode to serve as the content providers for Atlas.


shamil-gadelshin (approved):
LGTM (tested locally)

PR #1926, opened 2020-12-15
Council runtime

ondratra:
This PR adds runtime for the new council and the referendum.

- values for council settings are not finally set. Do we have any document with relevant settings? (how often election happens, etc.)
- old council from `governance` pallet was left intact as some modules seems to rely on it (e.g. proposals)

### DIFF
--- a/runtime/src/constants.rs
+++ b/runtime/src/constants.rs
@@ -43,6 +43,9 @@ parameter_types! {
     pub const StorageWorkingGroupLockId: LockIdentifier = [6; 8];
     pub const ContentWorkingGroupLockId: LockIdentifier = [7; 8];
     pub const ForumGroupLockId: LockIdentifier = [8; 8];
+    pub const CandidacyLockId: LockIdentifier = [9; 8];
+    pub const ElectedMemberLockId: LockIdentifier = [10; 8];
+    pub const ReferendumLockId: LockIdentifier = [11; 8];
 }
 
 lazy_static! {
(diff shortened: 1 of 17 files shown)


runtime/src/constants.rs
```diff
@@ -43,6 +43,9 @@ parameter_types! {
     pub const StorageWorkingGroupLockId: LockIdentifier = [6; 8];
     pub const ContentWorkingGroupLockId: LockIdentifier = [7; 8];
     pub const ForumGroupLockId: LockIdentifier = [8; 8];
+    pub const CandidacyLockId: LockIdentifier = [9; 8];
```
shamil-gadelshin, line 46: LockIds differ from the handbook.

shamil-gadelshin (changes requested):
Nice work! Very clean and easy to read and understand pallets.

A couple of questions and comments:
1. Is there support for the council integration with other pallets?
I can see the current council getter (that is needed by proposals engine), but cannot find the "council elected" event hook. I'm not sure about other pallets' dependencies.
2. Please, add all TODOs like "TODO: implement when membership module is ready" for Olympia to the "First audit project" as a task (or tasks) and estimate it.
3. As a general comment - consider keeping all your comments within the rust recommended boundaries (less than 100 chars per line) - it is easier to read them.
4. Also, private methods are not benefitting from the doc comments, regular comments should be sufficient.
5. Cannot build the runtime with benchmarking feature enabled!

conectado:
Continuing in #1930 from my fork

Issue #1843, opened 2020-12-01
Drop reward policy in working group


bedeho:
This seems redundant now

```
/// Reward policy for the job opening.
#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
#[derive(Encode, Decode, Debug, Clone, Default, PartialEq, Eq)]
pub struct RewardPolicy<Balance> {
    /// Reward per block for the worker.
    pub reward_per_block: Balance,
}
```

PR #1928, opened 2020-12-16
Drop RewardPolicy from working group

conectado:
# Tackling #1843

This PR drops `RewardPolicy` struct and use `Balance` instead to represent the `reward_per_block` in `Opening`


shamil-gadelshin (approved):
LGTM

PR #1930, opened 2020-12-16
Council runtime

conectado:
Continuing #1926 from my fork

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2345,6 +2345,7 @@ dependencies = [
  "pallet-common",
  "pallet-constitution",
  "pallet-content-directory",
+ "pallet-council",
  "pallet-finality-tracker",
  "pallet-forum",
  "pallet-governance",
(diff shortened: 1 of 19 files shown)


conectado:
Answering to @shamil-gadelshin [This comment](https://github.com/Joystream/joystream/pull/1926#pullrequestreview-553518207)

>     1. Is there support for the council integration with other pallets?
>        I can see the current council getter (that is needed by proposals engine), but cannot find the "council elected" event hook. I'm not sure about other pallets' dependencies.

@ondratra please correct me if I'm wrong, but what used to be `council_elected` is replaced by `process_winners` which is a hook in `referendum_pallet` that in the current implementation calls `recieve_referendum_results` in `council_pallet`

>     2\. Please, add all TODOs like "TODO: implement when membership module is ready" for Olympia to the "First audit project" as a task (or tasks) and estimate it.

Done in #1933 (the other TODOs are regarding weights and those tasks are already created)

ondratra:
There is no hook like `council_elected` right now. The feature can be included in this PR by merging this https://github.com/conectado/joystream/pull/2. This will add the function `new_council_elected(...)` to the council trait that can be used for such purpose.

shamil-gadelshin (approved):
LGTM

Issue #1896, opened 2020-12-10
Redundant post author information in proposal post editing


bedeho:
This is the signature

```Rust
pub fn update_post(
            origin,
            post_author_id: MemberId<T>,
            thread_id: T::ThreadId,
            post_id : T::PostId,
            _text : Vec<u8>
        )
```

But obviously, the runtime already knows the `post_author_id` of this post, because it is set on `DiscussionPost::author_id`.

PR #1936, opened 2020-12-17
Drop author_id from update_post in proposals/discussion

conectado:
# Tracking #1896 

Removes the `author_id` parameter from `update_post` and updates tests and benchmarking accordingly.


shamil-gadelshin (approved):
LGTM

Issue #1736, opened 2020-11-17
Add staking accounts binding to the membership pallet.


shamil-gadelshin:
New extrinsics:
- confirm_staking_account (origin, member_id); // origin corresponds to the staking_account
- add_staking_account_candidate(origin, member_id, staking_account_id) // origin corresponds to membership
- remove_staking_account(origin, member_id, staking_account_id) // origin corresponds to membership

Also, we need to introduce pallet_utility in the runtime.

shamil-gadelshin:
Done.

PR #1953, opened 2020-12-21
Add staking accounts to the membership pallet.

shamil-gadelshin:
### Changes
- add `confirm_staking_account` extrinsic
- add `add_staking_account_candidate` extrinsic
- add `remove_staking_account` extrinsic
- add `StakingAccountValidator` trait to the common module
- implement `StakingAccountValidator` in the membership module
- refactor staking handler and working group

[Main issue](https://github.com/Joystream/joystream/issues/1736)

### DIFF
--- a/runtime-modules/membership/src/lib.rs
+++ b/runtime-modules/membership/src/lib.rs
@@ -369,6 +369,7 @@ decl_event! {
         LeaderInvitationQuotaUpdated(u32),
         InitialInvitationCountUpdated(u32),
         StakingAccountAdded(MemberId, AccountId),
+        StakingAccountRemoved(MemberId, AccountId),
     }
 }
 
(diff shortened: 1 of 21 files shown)


runtime-modules/membership/src/lib.rs
```diff
@@ -684,6 +741,88 @@ decl_module! {
...
+        pub fn confirm_staking_account(
+            origin,
+            member_id: T::MemberId,
+        ) {
+            let staking_account_id = ensure_signed(origin)?;
+
+            let membership = Self::ensure_membership(member_id)?;
+
+            ensure!(
+                membership.staking_account_exists(&staking_account_id),
```
bedeho: Should we not check that its not confirmed already?

bedeho (commented):
Currently a single staking account can be associated with an unlimited number of memberships. While in principle it _may_ be that this still is safe, as long as all staking on such accounts uses `is_account_free_of_conflicting_stakes` implementation that rejects reuse a lock type with itself, but I am still not 100% sure about that, and also its not the mental model we want to encourage. Having a storage map from equivalent to `account => member_idXstatus` would solve the problem, and also remove need for any staking account cap.

What do you think?

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Ok. I added this check.

shamil-gadelshin:
I changed staking account verification to the double map. I removed the limit for the staking account number and added "staking account could be used only once in the system" check.

shamil-gadelshin:
Review fixes:
- changed double map to map for the staking account registration
- changed the workflow (adding and removing account requires staking account origin, confirmation - member origin).

bedeho (approved):
Excellent work.

PR #1961, opened 2020-12-23
Council rework integration with other modules

conectado:
This PR integrates the council rework with other modules by doing the following:
* Get rid of the old governance module
* Update `ActorOriginValidator` implementation according to the logic of this new module
* Implements the hook `new_council_elected` in the runtime
* Fix tests and benchmarks


shamil-gadelshin (approved):
LGTM

PR #1962, opened 2020-12-24
Modify message for a proposal’s decoding.

shamil-gadelshin:
### Problem
The text proposal execution had failed on the testnet. Text proposal execution doesn't contain any logic except printing to console.
https://testnet.joystream.org/#/proposals/80

### Investigation
I confirmed that the text proposal could not be executed as well as runtime upgrade and working group proposals after the runtime upgrade. Council and election proposals work well after the runtime upgrade.

#### Facts
1. Council and election pallets have last changes committed before the alexandria release.

2. Text proposals and runtime upgrade belong to the codex pallet. Codex and working group pallets were changed during the babylon preparation.

3. Codex and working group proposals don't work after the runtime upgrade.

4. Council (spending) and election (set_election_parameters) proposals work after the runtime upgrade.

#### My theory
Proposals engine cannot deserialize extrinsic from the old code into new code.
I proved it by changing the error message for decoding the extrinsic (decode function from the Substrate returns an empty error).

#### Solution
I will change the error message for the Olympia release, but we should cancel manually affected proposals before the next runtime upgrade.

### Changes
- changed non-working Substrate decoding error message to the "Decoding error" 
- removed printing to the console in the text proposal and placed the comment instead

### DIFF
--- a/runtime-modules/proposals/codex/src/lib.rs
+++ b/runtime-modules/proposals/codex/src/lib.rs
@@ -64,7 +64,6 @@ use frame_support::{decl_error, decl_module, decl_storage, ensure, print};
 use frame_system::ensure_root;
 use sp_arithmetic::traits::Zero;
 use sp_std::clone::Clone;
-use sp_std::str::from_utf8;
 use sp_std::vec::Vec;
 
 pub use crate::types::{
--- a/runtime-modules/proposals/codex/src/lib.rs
+++ b/runtime-modules/proposals/codex/src/lib.rs
@@ -354,18 +354,16 @@ decl_module! {
 
 // *************** Extrinsic to execute
 
-        /// Text proposal extrinsic. Should be used as callable object to pass to the `engine` module.
+        /// Text proposal extrinsic.
+        /// Should be used as callable object to pass to the `engine` module.
         #[weight = 10_000_000] // TODO: adjust weight
         pub fn execute_text_proposal(
             origin,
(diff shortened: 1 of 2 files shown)


bwhm:
Just to be clear here, you are saying these proposals don't work if they were made before the runtime upgrade, but executed after? 

It's a bit ambiguous in the text above whether they also apply to all future proposals fitting this description, but I'm quite sure these have, and will, work.

Can't comment on the code...

shamil-gadelshin:
> Just to be clear here, you are saying these proposals don't work if they were made before the runtime upgrade, but executed after?

Yes, for proposals which code was altered in the new runtime.

mnaamani:
Just some general points.

The encoded dispatch call is made up of three components, the module, the function and the arguments data.
Its important to note that the module and function are referenced by an index and not by name.
So after a runtime upgrade (depending on what new modules and dispatchable methods have been added or removed) there is no guarantee that the indices will be the same. 

This can explain the behavior observed.

There are security implications to this. If two dispatchable methods have identical signatures and following a runtime upgrade the module:function call index ends up point to the wrong module, we can have un-intended consequences.

So I propose we fix/enhance the proposals engine. Simplest way is to kill proposals during a runtime upgrade.
A more complex solution could be to add migration code to update the encoded calls to point at the new correct module:function indices.

bedeho (approved):
LGTM.

Issue #1884, opened 2020-12-08
Implement single interface for the working groups.


shamil-gadelshin:
Currently, all pallets that depend on the working group implement a different version of the working group integration.
We need to implement a single interface for Storage, Content Directory, Forum, Membership pallets.

shamil-gadelshin:
Done

PR #1963, opened 2020-12-25
Refactoring of the working group integration with other pallets.

shamil-gadelshin:
### Affected pallets
- storage
- forum
- content-directory

#### Changes
- added `WorkingGroupIntegration` member to the pallets
- changed existing methods to use `WorkingGroupIntegration` trait.

[Main issue](https://github.com/Joystream/joystream/issues/1884)


iorveth (approved):
Awesome work, @shamil-gadelshin. 

I saw you only upgraded content-directory pallet version, don`t we need to do the same with the rest of the affected pallets?

shamil-gadelshin:
> I saw you only upgraded content-directory pallet version, don`t we need to do the same with the rest of the affected pallets?

I had upgraded other affected pallets in bulk to version 4.0.0 previously. After the benchmarking, we could upgrade them to 4.0.1.

iorveth:
Ok, great

bedeho:
My extra pair of eyes are not needed here, so just merging.

Issue #1819, opened 2020-10-21
Howto Improve Testing, Local Dev and Deployment Infrastructure


mnaamani:
Create a ranked list of biggest limitations and problems in our test, dev, deployment infra, including man hour estimates on what fixes would require.

PR #1970, opened 2020-12-29
Integration tests refactoring

mnaamani:
Partial implementation of general Improvements to the network testing framework described in https://github.com/Joystream/joystream/issues/1819

Primarily:
- General code cleanup
- Improved Logging
- Implementing concrete types for patterns used in scenario, flows and fixtures to make them easier to write and test and specifically to remove a lot of boilerplate.
- Introduced concept of a job which is a collection of flows. Scenarios are basically a list of jobs that can be defined. Jobs can be configured to depend on other jobs completing first. Making write scenarios much cleaner.
- Better handling of tx sending. Also an important change is that the sending of transaction only throws exception if it is not accepted by the node. Extrinsic Failure or Success are treated as successful and the fixtures and flows must interpret the result and events to determine success or failure of a test. This makes it easier to write code to test expected failure of some runtime dispatches. Also fixed a rare race condition by using a single key for the async lock for all tx submissions.
- Introduced resource locking mechanism which allows synchronizing flows which might be trying to modify the same runtime state


tests/network-tests/src/Api.ts
```diff
@@ -80,15 +81,40 @@ export class Api {
...
+    this.keyring = keyring
+    this.treasuryAccount = treasuryAccount
+    this.sender = new Sender(api, keyring, label)
+  }
+
+  public enableDebugTxLogs(): void {
+    this.sender.setLogLevel(LogLevel.Debug)
+  }
+
+  public enableVerboseTxLogs(): void {
```
iorveth, line 115: I hope, them should be different.

`this.sender.setLogLevel(LogLevel.Verbose)`

iorveth (changes requested):
The idea of making testing framework architecture similar to the github actions is fantastic. Amazing how much more flexible and cleaner it looks now! 
Awesome work on introducing `InvertedPromise` and event search cleanup.

tests/network-tests/src/Api.ts
mnaamani: Good catch.
Fixed in 72fad7e2f436167e43e6612d5d68dab4c0f12b6c

mnaamani:
> The idea of making testing framework architecture similar to the github actions is fantastic. Amazing how much more flexible and cleaner it looks now!
> Awesome work on introducing `InvertedPromise` and event search cleanup.

Glad you liked the changes. Hopefully going forward we will be able to add a lot more test cases :)

iorveth (approved):
LGTM!

Issue #1864, opened 2020-12-03
New proposals


bedeho:
Codex should have the proposals reflected here, and not any other

https://joystream.gitbook.io/joystream-handbook/governance/proposals#proposals

shamil-gadelshin:
Currently, I see these new proposals:
- Set Membership Price
- Set Referral Cut
- Set Initial Invitation Count
- Set Initial Invitation Balance
- Set Membership Lead Invitation Quota
- Cancel Working Group Leader Opening.

bedeho:
I added two new proposals

- https://joystream.gitbook.io/joystream-handbook/governance/proposals#set-council-budget-increment
- https://joystream.gitbook.io/joystream-handbook/governance/proposals#set-councilor-reward

Issue #1748, opened 2020-11-17
Weights + benchmarking for proposals codex module.


shamil-gadelshin:
SSIA

shamil-gadelshin:
Done

PR #1977, opened 2021-01-04
Implement, test, and benchmark new proposals

conectado:
# This PR addresses #1748 and #1864

In this PR I did the following things:
* For existing proposals updated the names to closely match those in the handbook
  * Text -> Signal
  * Spending -> Funding Request
  * Set Validator Count -> Set Max Validator Count
  * Add Working Group Leader Opening -> Create Working Group Lead Opening
  * Fill Working Group Leader Opening -> Fill Working Group Lead Opening
  * Set Working Group Budget Capacity ->  Update Working Group Budget
  * Decrease Working Group Leader Stake -> Decrease Working Group Lead Stake
  * Slash Working Group Leader Stake -> Slash Working Group Lead
  * Set Working Group Leader Reward -> Set Working Group Lead Reward
  * Terminate Working Group Leader Role -> Terminate Working Group Lead
* Removed old proposals no longer in the handbook
* Added new proposals:
  * Cancel Working Group Lead Opening
  * Set Membership Price
  * Set Council Budget Increment
  * Set Councilor Reward
  * Set Initial Invitation Balance
  * Set Initial Invitation Count
  * Set Membership Lead Invitation Quota
  * Set Referral Cut
  * **Note:** _Update Budget Request_ and _Funding Request_ changed significantly
* Needed to change the `Balance` type in `Council` to the `Balance` in the `balances` pallet to be able to update the budget of `WorkingGroup` with the budget from `Council`
* Updated `councilor_reward` and `budget_incrment` to be part of the storage as to be able to be updated with `Set Council Budget Increment` and `Set Councilor Reward`

### Also, I have a couple of open questions:
* In [`Set Max Validator Count`](https://joystream.gitbook.io/joystream-handbook/governance/proposals#set-max-validator-count) we use `<staking::Module<T>>::minimum_validator_count()` instead of `MIN_VALIDATOR_COUNT`. Is this okay?

### [Handbook companion PR](https://github.com/Joystream/handbook/pull/27)

### DIFF
--- a/runtime-modules/proposals/codex/src/benchmarking.rs
+++ b/runtime-modules/proposals/codex/src/benchmarking.rs
@@ -18,8 +18,6 @@ use sp_std::prelude::*;
 const SEED: u32 = 0;
 const MAX_BYTES: u32 = 16384;
 
-use crate::call_wg;
-
 // Note: We use proposals_engine::Trait::Event here because crate::Trait
 // doesn't implement Event and we only use this function to assert events
 // from the proposals_engine pallet
@@ -177,13 +175,7 @@ fn create_proposal_verify<T: Trait>(
 fn set_wg_and_council_budget<T: Trait>(budget: u32, group: WorkingGroup) {
     Council::<T>::set_budget(RawOrigin::Root.into(), BalanceOf::<T>::from(budget)).unwrap();
 
-    call_wg!(
-        group<T>,
-        set_budget,
-        RawOrigin::Root.into(),
-        BalanceOf::<T>::from(budget)
-    )
-    .unwrap();
+    T::set_working_group_budget(group, BalanceOf::<T>::from(budget));
 
     assert_eq!(
         Council::<T>::budget(),
@@ -192,7 +184,7 @@ fn set_wg_and_council_budget<T: Trait>(budget: u32, group: WorkingGroup) {
     );
 
     assert_eq!(
-        call_wg!(group<T>, budget),
+        T::get_working_group_budget(group),
         BalanceOf::<T>::from(budget),
         "Working Group budget not updated"
     );
@@ -210,7 +202,7 @@ fn assert_new_budgets<T: Trait>(
     );
 
     assert_eq!(
-        call_wg!(group<T>, budget),
+        T::get_working_group_budget(group),
         BalanceOf::<T>::from(new_budget_working_group),
         "Working Group budget not updated"
     );
--- a/runtime-modules/proposals/codex/src/benchmarking.rs
+++ b/runtime-modules/proposals/codex/src/benchmarking.rs
@@ -566,37 +566,6 @@ benchmarks! {
     verify {
         assert_new_budgets::<T>(101, 99, WorkingGroup::Membership);
     }
-
-    funding_request {
-        let council_budget = BalanceOf::<T>::from(100);
-        let transfer_balance = BalanceOf::<T>::from(60);
-        let recieving_account = account::<T::AccountId>("reciever", 0, SEED);
-        Council::<T>::set_budget(RawOrigin::Root.into(), council_budget).unwrap();
-        assert_eq!(
-            Council::<T>::budget(),
-            council_budget,
-            "Council budget not updated"
-        );
-
-        assert_eq!(
-            Balances::<T>::free_balance(&recieving_account),
-            Zero::zero(),
-            "Recieving account has funds",
-        );
-    }: _ (RawOrigin::Root, transfer_balance, recieving_account.clone())
-    verify {
-        assert_eq!(
-            Council::<T>::budget(),
-            council_budget - transfer_balance,
-            "Council didn't discount transference",
-        );
-
-        assert_eq!(
-            Balances::<T>::free_balance(recieving_account),
-            transfer_balance,
-            "Recieving account didn't recieve amount",
-        );
-    }
 }
 
 #[cfg(test)]
--- a/runtime-modules/proposals/codex/src/benchmarking.rs
+++ b/runtime-modules/proposals/codex/src/benchmarking.rs
@@ -1,0 +1,560 @@
+#![cfg(feature = "runtime-benchmarks")]
+use super::*;
+use crate::Module as Codex;
+use balances::Module as Balances;
+use frame_benchmarking::{account, benchmarks};
+use frame_support::sp_runtime::traits::Bounded;
+use frame_support::traits::Currency;
+use frame_system::EventRecord;
+use frame_system::Module as System;
+use frame_system::RawOrigin;
+use membership::Module as Membership;
+use proposals_discussion::Module as Discussion;
+use proposals_engine::Module as Engine;
+use sp_runtime::traits::One;
+use sp_std::convert::TryInto;
+use sp_std::prelude::*;
+
+const SEED: u32 = 0;
+const MAX_BYTES: u32 = 16384;
+
+// Note: We use proposals_engine::Trait::Event here because crate::Trait
+// doesn't implement Event and we only use this function to assert events
+// from the proposals_engine pallet
+fn assert_last_event<T: Trait>(generic_event: <T as proposals_engine::Trait>::Event) {
+    let events = System::<T>::events();
+    let system_event: <T as frame_system::Trait>::Event = generic_event.into();
+    assert!(
+        events.len() > 0,
+        "If you are checking for last event there must be at least 1 event"
+    );
+    let EventRecord { event, .. } = &events[events.len() - 1];
+    assert_eq!(event, &system_event);
+}
+
+fn get_byte(num: u32, byte_number: u8) -> u8 {
+    ((num & (0xff << (8 * byte_number))) >> 8 * byte_number) as u8
+}
+
+// Method to generate a distintic valid handle
+// for a membership. For each index.
+fn handle_from_id(id: u32) -> Vec<u8> {
+    let mut handle = vec![];
+
+    for i in 0..4 {
+        handle.push(get_byte(id, i));
+    }
+
+    handle
+}
+
+fn member_funded_account<T: Trait + membership::Trait>(
+    name: &'static str,
+    id: u32,
+) -> (T::AccountId, T::MemberId) {
+    let account_id = account::<T::AccountId>(name, id, SEED);
+    let handle = handle_from_id(id);
+
+    // Give balance for buying membership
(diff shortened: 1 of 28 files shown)


runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -58,40 +58,92 @@ mod types;
...
+
 /// 'Proposals codex' substrate module Trait
 pub trait Trait:
     frame_system::Trait
     + proposals_engine::Trait
     + proposals_discussion::Trait
     + common::Trait
+    + council::Trait
     + staking::Trait
+    + working_group::Trait<ForumWorkingGroupInstance>
```
shamil-gadelshin: This will require the codex update for each new working group created.

runtime/src/proposals_configuration/sample_proposal_parameters.json
```diff
@@ -1,24 +1,25 @@
...
-  "fill_working_group_opening_proposal": {},
-  "set_working_group_budget_capacity_proposal": {},
-  "decrease_working_group_leader_stake_proposal": {},
-  "slash_working_group_leader_stake_proposal": {},
-  "set_working_group_leader_reward_proposal": {},
-  "terminate_working_group_leader_role_proposal": {},
-  "amend_constitution_proposal": {}
-}
\ No newline at end of file
+    "_comment": "This is a sample 'Proposal parameters' JSON file to enable conditional runtime compilation.",
```
shamil-gadelshin, line 2: Please, modify this file to reflect your changes - it is needed by the integration tests.

shamil-gadelshin: Make sure that this file works, you can use this `proposal_parameters_are_initialized` test for each proposal locally (but don't create tests for each proposal - we won't enable it by default because of the conditional compilation).

runtime-modules/working-group/src/lib.rs
```diff
@@ -63,6 +63,18 @@ use staking_handler::StakingHandler;
 
 type WeightInfoWorkingGroup<T, I> = <T as Trait<I>>::WeightInfo;
 
+/// The forum working group instance alias.
```
shamil-gadelshin: You made the working group - instance-dependent which downgrades its flexibility. For each new working group - you should update it.

runtime-modules/proposals/codex/src/types.rs
```diff
@@ -96,43 +136,43 @@ pub struct GeneralProposalParams<MemberId, AccountId, BlockNumber> {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, Clone, PartialEq, Debug)]
 pub struct TerminateRoleParameters<WorkerId, Balance> {
-    /// Leader worker id to fire.
+    /// Worker identifier.
     pub worker_id: WorkerId,
 
-    /// Terminate role slash penalty.
-    pub penalty: Option<Penalty<Balance>>,
+    /// Optional amount to be slashed.
```
shamil-gadelshin, line 142: Why did you remove the penalty here? My guess because the gitbook doesn't contain it. But termination slash is similar to the regular slash. And it contains the slashing rationale. You could confirm it with Bedeho.

runtime-modules/proposals/codex/src/benchmarking.rs
```diff
@@ -0,0 +1,811 @@
...
+        let (account_id, member_id, general_proposal_paramters) = create_proposal_parameters::<T>(t, d);
+
+        let proposal_details = ProposalDetails::SetReferralCut(One::one());
+    }: create_proposal(RawOrigin::Signed(account_id.clone()), general_proposal_paramters, proposal_details.clone())
+    verify {
+        create_proposal_verify::<T>(account_id, member_id, proposal_details);
+    }
+
+    update_working_group_budget_positive_forum {
+        set_wg_and_council_budget::<T>(100, WorkingGroup::Forum);
```
shamil-gadelshin, line 506: Is it possible to introduce some parameter x=0..1 which could define the balance sign? This could reduce the benchmarks count and simplify the code.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -372,31 +538,87 @@ decl_module! {
...
+                    ensure!(amount<=current_budget, Error::<T>::InsufficientFundsForFundingRequest);
+
+                    call_wg!(working_group<T>, set_budget, origin.clone(), wg_budget.saturating_add(amount))?;
+                    Council::<T>::set_budget(origin, current_budget - amount)?;
+                },
+                BalanceKind::Negative => {
+                    ensure!(amount <= wg_budget, Error::<T>::InsufficientFundsForFundingRequest);
+
+                    call_wg!(working_group<T>, set_budget, origin.clone(), wg_budget - amount)?;
+                    Council::<T>::set_budget(origin, current_budget.saturating_add(amount))?;
```
shamil-gadelshin, line 543: Why does it set the council budget rather than working group budget?

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -372,31 +538,87 @@ decl_module! {
...
+                }
+            }
+        }
+
+        /// Funding request proposal
+        #[weight = WeightInfoCodex::<T>::funding_request()] // TODO: adjust weight
+        pub fn funding_request(
```
shamil-gadelshin: You don't need this extrinsic. Because you could just run the extrinsic from the council module.

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -372,31 +538,87 @@ decl_module! {
...
+            ProposalDetails::FundingRequest(ref amount, _) => {
+                ensure!(
+                    *amount != BalanceOf::<T>::zero(),
+                    Error::<T>::InvalidFundingRequestProposalBalance
+                );
+
                 ensure!(
-                    *balance != BalanceOf::<T>::zero(),
-                    Error::<T>::InvalidSpendingProposalBalance
+                    *amount <= <BalanceOf<T>>::from(MAX_SPENDING_PROPOSAL_VALUE),
```
shamil-gadelshin, line 568: We don't check for the storage of other modules when create the proposal. We check only proposal parameters. Ask Bedeho for details.

shamil-gadelshin (changes requested):
Great work, but there are major issues:
- working group dependency on its own instances
- codex dependency on the working group instances
- unnecessary extrinsic proxies

Also, please, update the codex pallet comment.

runtime-modules/proposals/codex/src/types.rs
shamil-gadelshin: I think that the gitbook should be updated rather than changing this parameter.

runtime-modules/proposals/codex/src/lib.rs
conectado: I could fix this by adding a hook to the runtime that allows us to query and set the working group budget. Also, this would fix having the working group instances definitions in the working group since this way I would only need them in the runtime and I can move them back there. Do you think this is an acceptable way of handling this?

runtime-modules/proposals/codex/src/benchmarking.rs
conectado: We could do that, but I am a little reticent to do it since the parameter would imply an order from best to worst-case scenario and the benchmarks will try to do a linear regression over it and might create and spurious linear dependency.

runtime-modules/proposals/codex/src/lib.rs
conectado: Since the balance is non-negative in the line before this I reduce the working group budget by `amount` and in this line I add that `amount` to the council budget. That's the behavior I understand is expected [from the handbook](https://joystream.gitbook.io/joystream-handbook/governance/proposals#update-working-group-budget).

runtime-modules/proposals/codex/src/types.rs
conectado: I think this parameter is missing in the gitbook because the generic description for the proposal already doubles as a rationale. I remember bedeho saying he dropped the rationale from slashing lead for this reason.

runtime-modules/proposals/codex/src/lib.rs
conectado: I see, in the [handbook](https://joystream.gitbook.io/joystream-handbook/governance/proposals#creation-conditions-3) it says that it should check for council balance, so I will ask for clarification

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: Yes, I also thought of the runtime hook. Let's create a prototype to see how it goes.

runtime-modules/proposals/codex/src/types.rs
shamil-gadelshin: Ok. Make a comment please when you pass the empty text on two occasions: slashing and termination.

runtime-modules/proposals/codex/src/benchmarking.rs
shamil-gadelshin: Fair enough. Let it be.

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: Sorry, I missed the call_wg part.

runtime/src/proposals_configuration/tests.rs
```diff
@@ -1,10 +1,371 @@
 use crate::ProposalParameters;
 
 // Enable during the conditional compilation tests.
+#[test]
+#[ignore]
+fn proposal_parameters_are_initialized_max_validators() {
+    let actual_params = super::SetMaxValidatorCountProposalParameters::get();
+    let expected_params = ProposalParameters {
```
shamil-gadelshin: Please, consider creating "default_proposal_parameters" function.

runtime-modules/working-group/src/lib.rs
```diff
@@ -1029,6 +1029,16 @@ decl_module! {
 }
 
 impl<T: Trait<I>, I: Instance> Module<T, I> {
+    // Set working group budget
+    pub fn put_budget(new_budget: BalanceOf<T>) {
```
shamil-gadelshin: Please, remove these duplicated methods.

shamil-gadelshin (changes requested):
There are still a couple of minor issues.

shamil-gadelshin (approved):
LGTM

Issue #1933, opened 2020-12-16
Implement is_council_member in pallet_council for Runtime


conectado:
This implementation depends on membership being ready

shamil-gadelshin:
Also, please address this request from the original PR:
> values for council settings are not finally set. Do we have any document with relevant settings? (how often election happens, etc.)

ondratra:
Council settings task was moved to separate issue https://github.com/Joystream/joystream/issues/1929

shamil-gadelshin:
Fixed in  #1981

Issue #1898, opened 2020-12-10
Refactor common::ActorOriginValidator trait and implementations.


shamil-gadelshin:
#### Changes
 - `ActorOriginValidator` should be migrated to the DispatchError instead of the string
 - runtime  `MembershipOriginValidator` should be implemented in the membership module
 - runtime `CouncilOriginValidator` should be implemented in the new council module

These changes will allow to simplify the runtime code and reduce code duplication. The task could be done after the new governance module [integration](https://github.com/Joystream/joystream/issues/1734) into the olympia release branch.

shamil-gadelshin:
Done.

PR #1981, opened 2021-01-05
Refactor actor origin validators.

shamil-gadelshin:
### Changes
- refactor CouncilOriginValidator - move implementation to the council pallet
- refactor MemberOriginValidator - move implementation to the membership pallet
- migrate forum, content-directory, council to the new MemberOriginValidator implementation
- remove obsolete code from the membership pallet
   - remove MemberIdsByRootAccountId storage
   - remove MemberIdsByControllerAccountId storage
   - remove is_member_account method

[Main issue](https://github.com/Joystream/joystream/issues/1898)
[Related issue](https://github.com/Joystream/joystream/issues/1933)


runtime-modules/forum/src/lib.rs
```diff
@@ -24,16 +25,11 @@ mod tests;
 /// Moderator ID alias for the actor of the system.
 pub type ModeratorId<T> = common::ActorId<T>;
 
+/// Forum user ID alias for the member of the system.
+pub type ForumUserId<T> = common::MemberId<T>;
+
 pub trait Trait: frame_system::Trait + pallet_timestamp::Trait + common::Trait {
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
-    type ForumUserId: Parameter
```
bedeho, line 65: This change here, switching this out with `common::MemberId<T>`, I suspect is emblematic for the sort of change we also discussed in the referendum? I think it is unfortunate to do this, as it makes modules, like the forum, coupled to unrelated code as a software artifact. Someone wanting to just use our forum module should not be dealing with this.

I do not know what problem this solves, but I worry now that this new `common` concept has become a locus of coupling for the entire runtime code base.

bedeho (commented):
Herculean work, well done, I only have one comment on this topic of `common`.

iorveth (approved):
.

iorveth (changes requested):
LGTM, but i see some merge conflicts here.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: Yes, `common::Trait` is a dependency for multiple modules. But we get this dependency only for modules that need membership binding and/or working group. Here, the forum user cannot be a non-member, because we check their rights to act in the forum using the working group. @iorveth  could correct me if I'm wrong.

We could refactor `common::Trait` and split it into two separate traits - MembershipBound (with MemberId defined) and ActorBound (with ActorId defined) to separate the need for membership from the need of the working group for pallets.

Also, we could see MemberId as a similar concept with the AccountId from the frame_system which is an unavoidable dependency. This is why this Trait was placed in the 'common' pallet. 

The alternative would be to implement a conversion code for each module that uses MemberId to/from the local identifier. It would also make the entire system not statically type-safe.

bedeho: I guess the trait separation seems like the best tradeoff, but lets not do it now, instead make an issue with tech-debt label, so then we can re-evaluate later.

shamil-gadelshin: Here it is https://github.com/Joystream/joystream/issues/1996

conectado (approved):
LGTM

Issue #1867, opened 2020-12-04
Include a weight to fee function in pallet_transaction_payment


conectado:
For weights to impact in the fees we need to modify the runtime implementation of `pallet_transaction_payment` we are using.

Currently `type WeightToFee = integration::transactions::NoWeights;` which means that there's no cost to weights.
Furthermore, we also need to decide if to include `FeeMultiplierUpdate` which also impacts fee when including weights.

conectado:
The `WeightToFee` defines a polynomial function to transform a weight into balance.
In most cases, it seems to make sense to use a linear function. Inside Substrate there is a function readily prepared called [`IdentityFee`](https://substrate.dev/rustdocs/v2.0.0/frame_support/weights/struct.IdentityFee.html). This functions defines 1 weight = 1 balance.

However, this is not the function that Polkadot uses.

Polkadot uses the [following](https://github.com/paritytech/polkadot/blob/19b88fffbc93b8c47233af4e5c27d1bb95e94e76/runtime/polkadot/src/constants.rs#L71-L85):

```rs
impl WeightToFeePolynomial for WeightToFee {
		type Balance = Balance;
		fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
			// in Polkadot, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
			let p = super::currency::CENTS;
			let q = 10 * Balance::from(ExtrinsicBaseWeight::get());
			smallvec![WeightToFeeCoefficient {
				degree: 1,
				negative: false,
				coeff_frac: Perbill::from_rational_approximation(p % q, q),
				coeff_integer: p / q,
			}]
		}
```

This defines a scale between the `ExtrinsicBaseWeight` (The minimum weight for an extrinsic no matter the computation time, length or storage time) and the minimum balance an extrinsic takes and `MaximumBlockWeight` which is the maximum weight for a block. 

In the case of Polkadot it uses [`ExtrinsicBaseWeight = 1/10th Cent`](https://github.com/paritytech/polkadot/blob/19b88fffbc93b8c47233af4e5c27d1bb95e94e76/runtime/polkadot/src/constants.rs#L104-L112) and [`MaximumBlockWeight = 16 Dollars`](https://github.com/paritytech/polkadot/blob/19b88fffbc93b8c47233af4e5c27d1bb95e94e76/runtime/polkadot/src/constants.rs#L94-L102).

[...]

PR #1983, opened 2021-01-06
Weight to fee function

conectado:
# Adressing #1867

This PR adds the parameters for `pallet_transaction_payment`

This includes: 
* `TransactionByteFee`: is 1, since it's the minimum non-zero we can use.
* `WeightToFee`: The value here is also the same as in [Polkadot](https://github.com/paritytech/polkadot/blob/v0.8.26/runtime/polkadot/src/constants.rs#L70)
* `FeeMultiplierUpdate`: The value here is also the same as in [Polkadot](https://github.com/paritytech/polkadot/blob/v0.8.26/runtime/common/src/lib.rs#L85)

These values for the multiplier update function are based on [this document](https://w3f-research.readthedocs.io/en/latest/polkadot/economics/1-token-economics.html#relay-chain-transaction-fees-and-per-block-transaction-limits).

The value for `JOYS` is set to the total number of tokens in the runtime.

To decide on the constants for `WeightToFee` we set the target value of a `transfer` on 20. But to have more rounded values we set DOLLARS on `JOYS / 12500`. This gives us a `transfer` value of 30, having the advantage of keeping the same `WeightToFee` function as Polkadot, this doesn't matter since thanks to `FeeMultiplierUpdate` the value will evolve to have the optimal value to have %25 average block saturation. 

Also, the value for `TransactionByteFee` makes the cost for creating a Runtime Upgrade proposal at least 3M JOYS if we assume an average of 3MB for the length of a Runtime Upgrade and the cost for executing it 6M JOYS(3M for the length and 3M for the block weight), this might seem high, but we will settle with this until we can upgrade the runtime to have more decimal places for the JOYS which would allow us to have a byte fee less than 1.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -261,15 +261,19 @@ impl pallet_balances::Trait for Runtime {
 }
 
 parameter_types! {
-    pub const TransactionByteFee: Balance = 0; // TODO: adjust fee
+    pub const TransactionByteFee: Balance = 10 * constants::currency::MILLICENTS;
 }
 
 impl pallet_transaction_payment::Trait for Runtime {
     type Currency = Balances;
+    // TODO: Implement a function that sends %80 of fee to treasury and %20 to author
+    // and %100 of the tip to author.
+    // See: https://w3f-research.readthedocs.io/en/latest/polkadot/economics/1-token-economics.html#setting-transaction-fees
+    // and https://w3f-research.readthedocs.io/en/latest/polkadot/economics/1-token-economics.html#-2.-slow-adjusting-mechanism
     type OnTransactionPayment = ();
     type TransactionByteFee = TransactionByteFee;
-    type WeightToFee = integration::transactions::NoWeights; // TODO: adjust weight
-    type FeeMultiplierUpdate = (); // TODO: adjust fee
+    type WeightToFee = constants::fees::WeightToFee;
+    type FeeMultiplierUpdate = constants::fees::SlowAdjustingFeeUpdate<Self>;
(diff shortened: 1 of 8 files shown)


runtime/src/lib.rs
```diff
@@ -261,15 +263,39 @@ impl pallet_balances::Trait for Runtime {
...
 }
 
 impl pallet_transaction_payment::Trait for Runtime {
     type Currency = Balances;
-    type OnTransactionPayment = ();
+    type OnTransactionPayment = DealWithFees;
     type TransactionByteFee = TransactionByteFee;
-    type WeightToFee = integration::transactions::NoWeights; // TODO: adjust weight
-    type FeeMultiplierUpdate = (); // TODO: adjust fee
+    type WeightToFee = constants::fees::WeightToFee;
```
shamil-gadelshin, line 297: Please, remove the unused NoWeights struct.

runtime/src/tests/fee_tests.rs
```diff
@@ -0,0 +1,28 @@
+use super::initial_test_ext;
+use crate::constants::currency::{CENTS, DOLLARS};
+use crate::MaximumBlockWeight;
+use crate::Runtime;
+use frame_support::weights::WeightToFeePolynomial;
```
shamil-gadelshin: There are warnings on a test build.

shamil-gadelshin (changes requested):
Looks good, there is a couple of minor issues.

shamil-gadelshin (approved):
LGTM

PR #1989, opened 2021-01-07
Fix referendum tests compilation

conectado:
After the latest merges in olympia the referendum tests compilation broke, this PR fixes it


shamil-gadelshin:
LGTM

Issue #1934, opened 2020-12-16
Move on_finalize to on_initialize in pallet_council


conectado:
This is needed to be able to correctly benchmark the function

shamil-gadelshin:
Done

PR #2000, opened 2021-01-11
Council and referendum move finalize to initialize and add weights

conectado:
# Tracking #1934 

* Moves `on_finalize` to `on_initialize` to be able to use weights in the function.
* Fixes testing for these changes.
* Add weights to the `on_intialize` on both pallets

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -460,12 +460,16 @@ decl_module! {
         /////////////////// Lifetime ///////////////////////////////////////////
 
         // No origin so this is a priviledged call
-        fn on_finalize(now: T::BlockNumber) {
+        fn on_initialize() -> Weight {
+            let now = frame_system::Module::<T>::block_number();
+
             // council stage progress
             Self::try_progress_stage(now);
 
             // budget reward payment + budget refill
             Self::try_process_budget(now);
+
+            1_000_000 // TODO: replace
         }
 
         /////////////////// Election-related ///////////////////////////////////
@@ -669,14 +673,12 @@ impl<T: Trait> Module<T> {
         // election progress
         match Stage::<T>::get().stage {
             CouncilStage::Announcing(stage_data) => {
-                if now
-                    == Stage::<T>::get().changed_at + T::AnnouncingPeriodDuration::get() - 1.into()
-                {
+                if now == Stage::<T>::get().changed_at + T::AnnouncingPeriodDuration::get() {
                     Self::end_announcement_period(stage_data);
                 }
             }
             CouncilStage::Idle => {
-                if now == Stage::<T>::get().changed_at + T::IdlePeriodDuration::get() - 1.into() {
+                if now == Stage::<T>::get().changed_at + T::IdlePeriodDuration::get() {
                     Self::end_idle_period();
                 }
             }
@@ -1020,9 +1022,7 @@ impl<T: Trait> Mutations<T> {
         // set stage
         Stage::<T>::put(CouncilStageUpdate {
             stage: CouncilStage::Announcing(stage_data),
-            // set next block as the start of next phase (this function is invoke on block
-            // finalization)
-            changed_at: block_number + 1.into(),
+            changed_at: block_number,
         });
 
         // increase anouncement cycle id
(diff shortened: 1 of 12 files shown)


shamil-gadelshin (approved):
LGTM

runtime-modules/council/src/tests.rs
```diff
@@ -1108,15 +1108,24 @@ fn council_budget_refill_can_be_planned() {
 
         Mocks::plan_budget_refill(origin.clone(), next_refill, Ok(()));
 
+        assert_eq!(frame_system::Module::<Runtime>::block_number(), 1);
+
         // forward to one block before refill
-        MockUtils::increase_block_number(next_refill - 1);
+        MockUtils::increase_block_number(next_refill - 2);
```
ondratra: Let's evade using magic constant `2` and rather rewrite this part into:

```
let current_block = frame_system::Module::<Runtime>::block_number();

assert_eq!(current_block, 1);

// forward to one block before refill
MockUtils::increase_block_number(next_refill - current_block - 1);
```

runtime-modules/council/src/tests.rs
```diff
@@ -1188,19 +1197,16 @@ fn council_missed_rewards_are_paid_later() {
         MockUtils::increase_block_number(<Runtime as Trait>::ElectedMemberRewardPeriod::get());
 
         let last_payment_block = council_settings.cycle_duration
-            + (reward_period
-                - (council_settings.idle_stage_duration
-                    % <Runtime as Trait>::ElectedMemberRewardPeriod::get()))
-            - 1; // -1 because current block is not finalized yet
+            - (council_settings.idle_stage_duration
+                % <Runtime as Trait>::ElectedMemberRewardPeriod::get());
```
ondratra, line 1203: Just curious here - I was expecting there will only the ending `-1` to be removed. Why is there such a big change in the calculation? Was there a bug before?

runtime-modules/referendum/src/lib.rs
```diff
@@ -381,8 +381,11 @@ decl_module! {
         /////////////////// Lifetime ///////////////////////////////////////////
 
         // No origin so this is a priviledged call
-        fn on_finalize(now: T::BlockNumber) {
-            Self::try_progress_stage(now);
+        fn on_initialize() -> Weight{
```
ondratra: A small formatting fix -> `Weight {`.

runtime-modules/council/src/lib.rs
```diff
@@ -460,12 +460,33 @@ decl_module! {
         /////////////////// Lifetime ///////////////////////////////////////////
 
         // No origin so this is a priviledged call
-        fn on_finalize(now: T::BlockNumber) {
+        fn on_initialize() -> Weight {
```
ondratra, line 463: It would be great to keep `on_initialize()` small and compact so that everybody seeing the function will have an immediate idea of what happens there. The amount of weight calculation code slightly obscures the sequence of actions.

I suggest
a) moving `try_progress_stage` call to separate line
b) renaming `try_progress_weight` to `weight` since `try_*` prefix is used for functions there
c) adding a comment before `T::WeightInfo::try_progress_stage_idle()` line
```
            // minimum weight calculation
            let weight = T::WeightInfo::try_progress_stage_idle().max(
                T::WeightInfo::try_progress_stage_announcing_restart()
            );

            // council stage progress
            let mb_candidate_count = Self::try_progress_stage(now);

            // calculate weight increase
            let weight = if let Some(candidate_count) = mb_candidate_count {
                ...
            }

```

It would be great if the weight calculations could be moved into a separate function, but I assume this can't be done because functions like `try_progress_stage_idle()` are reading from the state/storage and their order in `on_initialize` matters.

ondratra (approved):
Good job. I only suggested some cosmetic changes. 

During a review, I discovered a bug in the elections, but it seems it existed in some form in the older code version (with `on_finalize`) as well. So I've created a separate issue just for that #2012.

runtime-modules/council/src/tests.rs
conectado: I believe it was an artifact because the `run_full_cycle` actually ran 4 blocks more than a full cycle.

So with the current params(idle period 27 blocks and reward period 10) the last time a reward is paid is in the 20th block since the council was elected(the idle period started) but if we run the period for 4 extra blocks(until the 31st block after the idle period started) the last time a reward will be paid is the 30th block since the idle period started.

ondratra:
LGTM

Issue #1932, opened 2020-12-16
Add locked balance for the invited member.


shamil-gadelshin:
Add support for these requirements from the [handbook](https://joystream.gitbook.io/joystream-handbook/subsystems/membership#invite-a-member):

> A new membership is created with the provided information, and initial invites set to zero, and the root account is credited withinvited_initial_balance.
Working group budget is deducted by invited_initial_balance.
controller_account is encumbered with lock with ID INVITE_LOCK_ID and of size invited_initial_balance  that only allows transaction fees, and credited with the same amount of tokens.

shamil-gadelshin:
Done.

PR #2004, opened 2021-01-11
Add the default balance for an invited member.

shamil-gadelshin:
#### Changes
- added invitation lock id
- subtracted the default invitation balance from the working group budget (if it is sufficient)
- transferred and locked (with exception of the transaction payments) default invitation balance on the member controller account
- added some tests

[Main issue](https://github.com/Joystream/joystream/issues/1932)


runtime-modules/common/src/working_group.rs
```diff
@@ -19,8 +19,8 @@ pub enum WorkingGroup {
     Membership,
 }
 
-/// Working group interface to use in the in the pallets with working groups.
-pub trait WorkingGroupIntegration<T: crate::Trait> {
+/// Working group interface to work with its members - workers and leaders.
+pub trait WorkingGroupParticipation<T: crate::Trait> {
```
bedeho: This name, and in fact the last, is not ideal, can we find a proper noun here? What does this actually represent?

runtime-modules/membership/src/lib.rs
```diff
@@ -544,6 +565,23 @@ decl_module! {
                 membership.invites = membership.invites.saturating_sub(1);
             });
 
+            // Decrease the working group balance.
+            let new_wg_budget = current_wg_budget - default_invitation_balance;
```
bedeho: This is perfectly safe? I have not looked at the safemath stuff, so just alerting you to the arithmetic going on.

bedeho (commented):
Looks good, just some minor comments, will merge when you tell me.

runtime-modules/membership/src/lib.rs
shamil-gadelshin: Changed to saturating_sub.

runtime-modules/common/src/working_group.rs
shamil-gadelshin: I renamed it to `WorkingGroupAuthenticator`.

bedeho (approved):
Great!

Issue #1935, opened 2020-12-16
Limit MaxWinners/IntermediateWinners in referendum


conectado:
This is needed to be able to calculate the weight of `reveal_vote` without accessing the store and without having the current number of intermediate winners as a parameter.

shamil-gadelshin:
Done

PR #2006, opened 2021-01-12
Limit max winners

conectado:
# Tracking #1935 

This PR limit the maximum number of target winners a referendum instance can have so the worst-case scenario for the extrinsics can be predicted.

Note that since `force_start` is used in `on_initialize` we can't fail there. So we fall back to using the maximum value if the passed value exceeds this.


runtime-modules/referendum/src/lib.rs
```diff
@@ -882,6 +893,14 @@ impl<T: Trait<I>, I: Instance> EnsureChecks<T, I> {
         Ok((current_cycle_id, account_id))
     }
 
+    fn allowed_winner_target_count(count: u64) -> Result<(), ()> {
```
shamil-gadelshin: Why do we need empty Err and Ok? Could we get away with boolean? And also, it looks like ensure! macro could be used otherwise.

shamil-gadelshin (changes requested):
Look good. I have just a couple of minor comments. 

Do you mind bringing referendum/lib.rs closer to rust 99 chars per line?

shamil-gadelshin (approved):
LGTM

Issue #1984, opened 2021-01-06
Change referendum pallet staking from LockableCurrency to the StakingHandler.


shamil-gadelshin:
Currently, we use LockableCurrency trait to manage the stakes in the referendum pallet rather than StakingHandler trait like all other pallets. We need to change the referendum to make it consistent with other pallets.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703952879841/1201704784810713) by [Unito](https://www.unito.io)

PR #2010, opened 2021-01-12
Replace LockableCurrency with StakingHandler in referendum

conectado:
# Tracking #1984 

In referendum use `Balance` from `pallet_balances` instead of `LockableCurrency` and use `StakingHandler` to lock the currency

### DIFF
--- a/runtime-modules/council/Cargo.toml
+++ b/runtime-modules/council/Cargo.toml
@@ -16,13 +16,13 @@ sp-std = { package = 'sp-std', default-features = false, git = 'https://github.c
 common = { package = 'pallet-common', default-features = false, path = '../common'}
 referendum = { package = 'pallet-referendum', default-features = false, path = '../referendum'}
 staking-handler = { package = 'pallet-staking-handler', default-features = false, path = '../staking-handler'}
+balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
 
 #Benchmark dependencies
 frame-benchmarking = { package = 'frame-benchmarking', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca', optional = true}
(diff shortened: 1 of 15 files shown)


runtime-modules/council/Cargo.toml
```diff
@@ -16,13 +16,13 @@ sp-std = { package = 'sp-std', default-features = false, git = 'https://github.c
 common = { package = 'pallet-common', default-features = false, path = '../common'}
 referendum = { package = 'pallet-referendum', default-features = false, path = '../referendum'}
 staking-handler = { package = 'pallet-staking-handler', default-features = false, path = '../staking-handler'}
+balances = { package = 'pallet-balances', default-features = false, git = 'https://github.com/paritytech/substrate.git', rev = 'a200cdb93c6af5763b9c7bf313fa708764ac88ca'}
```
shamil-gadelshin, line 19: Add this pallet to the std feature list, please. As well as staking-handler.

runtime-modules/referendum/src/lib.rs
```diff
@@ -646,16 +645,11 @@ impl<T: Trait<I>, I: Instance> Mutations<T, I> {
...
     ) -> Result<(), Error<T, I>> {
-        // lock stake amount
-        T::Currency::set_lock(
-            T::LockId::get(),
-            account_id,
-            *stake,
-            WithdrawReason::Transfer.into(),
-        );
+        // Should call after `can_vote`
+        T::StakingHandler::lock(account_id, *stake);
```
shamil-gadelshin: WithdrawReasons are incorrect. The latest version of the staking handler should contain suitable method to set a proper withdraw reason.

shamil-gadelshin (changes requested):
Please, have a look at the feature list in both `Cargo.toml` files for the referendum and council pallets. We need to keep them updated with dependencies. Otherwise, we could get subtle hard to find bugs during the release build.

shamil-gadelshin (changes requested):
Looks good. Just a small issue: please remove the staking handler from the dev section in the referendum Cargo.toml. It is redundant now.

shamil-gadelshin (approved):
LGTM

PR #2011, opened 2021-01-12
council - membership test reintroduction

ondratra:
Part of the test that's checking membership/account association was temporarily commented out during StakingHandler development. This PR reintroduces the proper checks and also removes some obsolete code from `mock.rs`.


mnaamani:
@ondratra is this still relevant ?

ondratra:
The changes in this PR seem to already be incorporated.

PR #2013, opened 2021-01-13
Council many iterations test

ondratra:
A new test for the council elections that showcases a bug described in #2012.

This PR assumes #2000 will be merged first.

PR #2016, opened 2021-01-13
Council many iterations test

### LINKED ISSUE #2012: Council - ensure rewards paid as expected over long run
I created a new test that simply runs 1000 elections in a row, and it discovered a bug - after some number of cycles (13), something unexpected happens, and a cycle simulation fails. A test can be seen in a PR below.

The problem is probably not related to the recent move of business logic from `on_finalize()` to `on_initialize()`. Even old code failed after a couple of cycles, but it failed earlier (only 5 cycles).

Find a source of this bug. Also, ensure in the test that all rewards are paid for the elected council members properly (after 1000 elections).

conectado:
Fixes #2012 

It continues work from #2013

### DIFF
--- a/runtime-modules/council/src/tests.rs
+++ b/runtime-modules/council/src/tests.rs
@@ -6,6 +6,7 @@ use super::{
 };
 use crate::mock::*;
 use common::origin::CouncilOriginValidator;
+use frame_support::traits::Currency;
 use frame_support::StorageValue;
 use frame_system::RawOrigin;
 use staking_handler::StakingHandler;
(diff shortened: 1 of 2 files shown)


runtime-modules/council/src/tests.rs
```diff
@@ -1389,13 +1389,15 @@ fn council_many_cycle_rewards() {
         let council_settings = CouncilSettings::<Runtime>::extract_settings();
 
         // quite high number of election cycles that will uncover any reward payment iregularities
-        let num_iterations = 1000;
+        let num_iterations = 100;
         //let num_iterations = 12; // working
         //let num_iterations = 13; // not working
```
ondratra: Please remove these 2 obsolete comments.

ondratra (approved):
LGTM.

ondratra (approved):

PR #2017, opened 2021-01-14
github workflows: build joystream-node with benchmarks

mnaamani:
Enable building joystream-node with features benchmarking, followed by running benchmarks.
Running of git diff at the end to show generated benchmarking code as a way to check if it has been checked in correctly.

Minor tweaks to setup.sh to work on github workflows. (docker installation and volta) although they are not strictly required for running the build or benchmarks but I wanted to have a similar experience in github as with local development.


shamil-gadelshin (changes requested):
What do you think about the renaming of the check? It looks like `joystream-node-checks` verifies only benchmarking now.

mnaamani:
> What do you think about the renaming of the check? It looks like `joystream-node-checks` verifies only benchmarking now.

I guess it makes sense to give it a more descriptive name, like joystream-node-benchmarks

shamil-gadelshin (approved):
LGTM

Issue #1992, opened 2021-01-07
Update the CurrencyToVotesHandler from the common pallet to the latest version.


shamil-gadelshin:
The current Substrate version contains a newer implementation of that function.

shamil-gadelshin:
Done.

PR #2022, opened 2021-01-14
Update CurrencyToVoteHandler.

shamil-gadelshin:
We need to update CurrencyToVoteHandler to the latest version from the Substrate framework.

[Main issue](https://github.com/Joystream/joystream/issues/1992)


mnaamani (approved):
This is a good improvement on current implementation which was too specific for joystream and our selection of very small total issuance as a % of a u64. The implementation in this PR factors this into the calculation so it should have the same result we need.

PR #2024, opened 2021-01-14
Membership benchmarking

iorveth:
1. Benchmarks for membership extrinsics.
2. Generated weight functions.
3. Extrinsics weight annotations.
4. Newly introduced `MembershipWorkingGroupHelper` trait and respective logic to be able to set up lead for membership benchmarking purposes.
5. Upgraded mock runtimes for pallets dependent on membership in dev enviroment.


runtime-modules/membership/src/lib.rs
```diff
@@ -313,7 +346,28 @@ decl_module! {
...
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (W)` where:
+        /// - `W` is the handle length
+        /// - DB:
+        ///    - O(W)
+        /// # </weight>
+        #[weight = if params.referrer_id.is_some() {
```
conectado: I think it'd be better to move the weight calculation to a separate function

runtime-modules/membership/src/benchmarking.rs
```diff
@@ -0,0 +1,775 @@
...
+        let second_member_id = member_id + T::MemberId::one();
+
+        assert_eq!(MemberIdByHandleHash::<T>::get(second_handle_hash), second_member_id);
+
+        assert_eq!(MembershipById::<T>::get(second_member_id), membership);
+
+        assert_last_event::<T>(RawEvent::MemberRegistered(second_member_id).into());
+    }
+
+    update_profile{
```
conectado, line 218: Please test for all parameters possible length(name, handle, avatar_uri, about)

runtime-modules/membership/src/lib.rs
```diff
@@ -412,7 +479,22 @@ decl_module! {
...
+        /// <weight>
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = match (new_root_account.is_some(), new_controller_account.is_some()) {
```
conectado: Please move weight calculation to a separate function

runtime-modules/membership/src/benchmarking.rs
```diff
@@ -0,0 +1,775 @@
...
+
+        assert_eq!(MembershipById::<T>::get(first_member_id), first_membership);
+
+        assert_eq!(MembershipById::<T>::get(second_member_id), second_membership);
+
+        assert_last_event::<T>(RawEvent::InvitesTransferred(first_member_id, second_member_id, number_of_invites).into());
+    }
+
+    invite_member {
+        let member_id = 0;
```
conectado, line 423: Please benchmark with all different parameters length in `InviteMembershipParameters`

conectado (changes requested):
Great work!

Just a couple of comments. 

But more importantly, we could do without introducing the trait in general, we can create a trait only within the benchmarking features. [See this example(the tests aren't fixed here)](https://github.com/conectado/joystream/commit/271a6b7df4c37b472fffe5283f7e1d556286fb6b).

runtime-modules/membership/src/benchmarking.rs
iorveth: It seems, that they do not affect benchmarking weight at all.
`	// WARNING! Some components were not used: ["j"]`
Only `handle` possible lengths affects benchmarking. 
We do not perform any computations/db operations with any of those parameters.

runtime-modules/membership/src/benchmarking.rs
iorveth: It seems, that they do not affect benchmarking weight at all.
`// WARNING! Some components were not used: ["j"]`

Only `handle` possible lengths affects benchmarking.
We do not perform any computations/db operations with any of those parameters.

runtime-modules/membership/src/lib.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2024/commits/187fa9e4ba17abd5e1c97dd2fc8db253f282dd7f

runtime-modules/membership/src/lib.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2024/commits/187fa9e4ba17abd5e1c97dd2fc8db253f282dd7f

iorveth:
`But more importantly, we could do without introducing the trait in general, we can create a trait only within the benchmarking features. See this example(the tests aren't fixed here).
`

Thank you for the helpful suggestion!

runtime-modules/membership/src/benchmarking.rs
conectado: I think it's better if we leave the benchmarking for those parameters on and we manually modify the trait(this won't be neccesary with the next Substrate update). It's how we have handled it in the other pallets and sometimes unexpectedly these parameters have some footprint in processing time.

runtime-modules/membership/src/benchmarking.rs
conectado: I think it's better if we leave the benchmarking for those parameters on and we manually modify the trait(this won't be neccesary with the next Substrate update). It's how we have handle it in the other pallets and sometimes unexpectedly these parameters have some footprint in processing time.

conectado (changes requested):
I just recommend that we would benchmark unused parameters even if they don't affect the results.

PR #2029, opened 2021-01-15
Fix ignored tests in council 

conectado:
I removed the ignored tests since it was already tested by a commented test in `council_membership_checks`. So I uncommented it and fixed and then erased the old ignored test.


shamil-gadelshin (approved):
LGTM

PR #2053, opened 2021-01-21
Bump crates versions for the node, runtime, pallets.

shamil-gadelshin:
SSIA


mnaamani (approved):
Since there is no real network deployed with the current version of Olympia, bumping the runtime version isn't really necessary, but its okay.

Issue #2048, opened 2021-01-21
Define Metadata Create and Update messages


mnaamani:
Define all the metadata 'blobs'/messages for the new content directory.

Add protocol message definitions into a consumable package for clients (CLI, Query-Node, Atlas)

mnaamani:
Implemented in #2096

Issue #2032, opened 2021-01-16
Conceptual: Atomic uploads and content directory transactions


bedeho:
# Background

One of the goals in the `Sumer` release is to allow single transactions to do multiasset uploads and content directory transactions. Changes will be made in the storage system to faciliatate atomic multiasset uploads (https://github.com/Joystream/joystream/issues/2030), and we are also moving to a content directory metaprotocol, so this issue tries to bring these together in a conceptual integrated example of how this all works together.

# Example Scenario

Lets say that a member with identifier `member_id` wishes to make a change to a published video with identifier `567` involving

1. editing title to "My great vacation"
2. change preview image to `blob_1`.
3. change video file to `blob_2`.

## 1. Issuing metaprotocol transaction

The metaprotocol has some message format

```
message UpdateVideo {
 required video_id int64 = 1;
 optional new_title string = 2;
 optional new_description string = 3;
 optional new_preview_image bool = 4;
 optional new_preview_gif bool = 5
 optional new_video bool = 6;
}
```

The publishing app (or some library used by the app realistically) would in this case construct a message instance doing something like this

```
let update_video_message = new UpdateVideo();

update_video_message.set_video_id(567);
update_video_message.set_new_title("My great vacation");
update_video_message.set_new_preview_image(true);
update_video_message.set_new_video(true);
```

Then using new metaprotocol module `metaprotocol`(https://github.com/Joystream/joystream/issues/2028), the extrinsic `atomic_postfixed_message` where
- `origin` is signed with member controller key
- `author` is `MessageAuthor::Member(member_id)`
- `call` is a an invocation of `multi_add_content` in the updated datadirectory (https://github.com/Joystream/joystream/issues/2030) where
   - `origin` is as above.
   - `member_id` is `member_id`.
   - `content` is a vector of  two `ContentParameters`, one for each of the two assets _in the same order of the update fields of `UpdateVideo`, so first the preview image and then the video.

## 2. Query Node Metatransaction Processing

The query node must process the `AtomicPostfixedMessage(MessageAuthor, call: Vec<u8>, serialized_message: Vec<u8>)` event in the `metaprotocol` module, and apply roughly the following validation checks in order.
 
1. `serialized_message` parameter is a valid encoding of the `UpdateVideo`, if so decode into message instance `m`, otherwise stop.
[...]

PR #2096, opened 2021-01-25
protobuf content-metadata

mnaamani:
Implements https://github.com/Joystream/joystream/issues/2048

Introduces a new npm package `@joystream/content-metadata-protobuf` for constructing, serializing and de-serializing the metadata "blob" component in runtime dispatch calls to create and update various content elements:
- Channels,
- Video,
- Playlist,
- Series,
- Persons 

The package also documents and defines the "opaque" types for things such as Language, and Media/codec types, as well as Joystream specific types such as the License.

Remember this is a low level library so there is very little hand-holding or wrapper classes around the protobuf message constructors, to do things like validation etc. This is responsibility of another library specifically.

There is an important implication of the messages on the input schema for the query-node, which is that all fields must be "nullable".


content-metadata-protobuf/compiled/proto/Channel_pb.d.ts
```diff
@@ -0,0 +1,45 @@
+// package:
```
bedeho, line 1: Should we include the compiled outputs in the repo, seems inadvisable?

bedeho (changes requested):
1. I see that all fields are optional, and there are no distinct metadata types for initially setting vs updating, but I think we do want to have some non-optional fields? we may even want constraints not directly captured in the types, such as max min lengths and so on. At least having the alternative to express such constraints, what do you think?

2. One thing that strikes me is that we have put the storage assets into the runtime rules and not in the metadat. At this point, it appears this is largely to make the content directory trigger the uploads atomically with a content directory change. This is an important property, but it also means that any change to what storage assets are needed implies a runtime upgrade (I do not think it requires a migration currently). However, we could still generate this atomicity in another way which does not need runtime upgrades. Namely, a sort of `atomic_storage_upload_operation_and(Vec<ContentParameters>, call)` type extrinsic, where `call` could be any content directory extrinsic, such as editing channels, videos etc. the `call` would need a metadata encoding which explains what index in the vector should be what field in the content metadta. So very similar ot this
https://github.com/Joystream/joystream/issues/2032

I guess this is more complex, but unlocks freedom to change assets without any runtime change. It does however mean we permanently lose the ability for content directory to manage storage, like doing cleanup, or blocking deletion, etc.

What do you think?

content-metadata-protobuf/test/channel.ts
```diff
@@ -0,0 +1,27 @@
+import { ChannelMetadata } from '../src'
```
bedeho, line 1: Where do these tests come from? are they autogenerated? if not, does it make sense to write tests for autogenerate code?

content-metadata-protobuf/proto/Video.proto
```diff
@@ -0,0 +1,69 @@
+syntax = "proto2";
+
+// Publication status before joystream
+message PublishedBeforeJoystream {
+    // Was video published before joystream platform
+    optional bool is_published = 1;
```
bedeho, line 6: is this really needed? coudlnt the whole `PublishedBeforeJoystream` actually just be an optional timestamp? that covers every state.

content-metadata-protobuf/proto/Video.proto
```diff
@@ -0,0 +1,69 @@
...
+
+message VideoMetadata {
+    // Video Title
+    optional string title = 1;
+
+    // Video Description
+    optional string description = 2;
+
+    // Lengths of video in seconds
+    optional int32 duration = 3;
```
bedeho: Next three fields should be unsigned.

bedeho:
I just wanted to amend my prior review and point out a simpler approach to get the same benefit.

For each extrinsic that has something to do with metadata, we simply amend it to include the content directory changes, so

- `create_channel(atomic_upload_content: Vec<ContentParameters>, metadata)`
- `create_video(atomic_upload_content: Vec<ContentParameters>, metadata)`
- `create_playlist(atomic_upload_content: Vec<ContentParameters>, metadata)`
- etc.

It now works exactly as before, where the extrinsic checks whether the uploads can AND adding the content can happen, then it does both. The simplicitiy of this approach over the `atomic_storage_upload_operation_and` approach is that its just a bit too abstract, and is a lot trickier to deal with on the query node side.

content-metadata-protobuf/compiled/proto/Channel_pb.d.ts
mnaamani: I'm making an exception here, because this is a consumable package and it seems a burden to expect the consumer to also have the protocol buffer compiler to use the package.

content-metadata-protobuf/test/channel.ts
mnaamani: I wrote them as a way to demo how to use the messages, and also as a spec to detect breaking changes when the message format is updated.

content-metadata-protobuf/proto/Video.proto
mnaamani: will fix

content-metadata-protobuf/proto/Video.proto
mnaamani: Not quite. So it being an `optional`  field in the protobuf message is sufficient to inform the consumer that we want to change or set the publishedbeforejoystream attribute of the metadata, but the the value is itself is an optional to have assocaited with a video.  ie. we need `Option<Option<timestamp>>` in rust syntax.

I could have chosen a simpler representation of just `optional u32 timestamp` and application would have to interpret the value 0 for example to mean that a video was not published before joystream. But that doesn't work if a video came before the unix timestamp start date..

mnaamani:
>     1. I see that all fields are optional, and there are no distinct metadata types for initially setting vs updating, but I think we do want to have some non-optional fields? we may even want constraints not directly captured in the types, such as max min lengths and so on. At least having the alternative to express such constraints, what do you think?

So I made a general point about this in the README but seems I didn't explain is clearly: https://github.com/Joystream/joystream/pull/2096/files#diff-a832c60a464076077a639635ca39fe6fa92a45e6d9fdb81ad9846f8665d2547dR9

 So about the non-optional fields. I agree it doesn't make for most of the fields to not have real value and to be required. But this validation cannot be done at the protobuf layer. Its really just validation that has to be done by the applications. Infact the code generated to construct the messages doesn't put any constraints other than the value passed for a field can fit in it. Even for `required` fields you can get away with constructing them without setting any of them explicitly. Its just about how the message will be encoded (wether a field value will be in the message or not).

mnaamani:
> I just wanted to amend my prior review and point out a simpler approach to get the same benefit.
> 
> For each extrinsic that has something to do with metadata, we simply amend it to include the content directory changes, so
> 
>     * `create_channel(atomic_upload_content: Vec<ContentParameters>, metadata)`
> 
>     * `create_video(atomic_upload_content: Vec<ContentParameters>, metadata)`
> 
>     * `create_playlist(atomic_upload_content: Vec<ContentParameters>, metadata)`
> 
>     * etc.
> 
> 
> It now works exactly as before, where the extrinsic checks whether the uploads can AND adding the content can happen, then it does both. The simplicitiy of this approach over the `atomic_storage_upload_operation_and` approach is that its just a bit too abstract, and is a lot trickier to deal with on the query node side.

From the two approaches I definitely prefer calling into storage module from the content module. But I still don't see why we "loose ability for content directory to manage storage, like doing cleanup, or blocking deletion, etc." in the first approach.

With this approach:
```
* `create_channel(atomic_upload_content: Vec<ContentParameters>, metadata)`
```

Are you suggesting to have the asset information in metadata blob as well?

mnaamani:
So going with the less complex approach for now as agreed so the content directory will call into the storage module to do the uploads. I also went with the index into the vec for both creating and updating messages, because we need index when updating (the string url in metadata or implicit element in vector only works for creating, when updating you have to be explicit about which asset you are changing and then be able to determine which form the asset is being provided as).
So with this approach there is a consistent way to deal with create and upload events with respect to parsing the event data { Vec, metadata }

```rust
// How new assets are to be added on creating and updating
// Channels,Videos,Series and Person
enum NewAsset {
  Upload(ContentUploadParameters),
  Uri(Vec<u8>),
}

struct VideoCreationParameters {
  in_category: VideoCategoryId,
  meta: Vec<u8>, // binary serialized VideoMetadata protobuf message
}

struct VideoUpdateParameters {
  new_in_category: Option<VideoCategoryId>,
  new_meta: Option<Vec<u8>>, // binary serialized VideoMetadata protobuf message
}

fn create_video(origin, owner: ChannelOwner, assets: Vec<NewAsset>, params: VideoCreationParameters) {}
fn update_video(origin, owner: ChannelOwner, assets: Vec<NewAsset>, params: VideoUpdateParameters) {}

// Events associated with creating and updating
VideoCreated(VideoId, Vec<NewAsset>, VideoCreationParameters)
VideoUpdated(VideoId, Vec<NewAsset>, VideoUpdateParameters)

// How the video metadata references the "Out of Band" array

assets = [
    NewAsset::Uri("https://mydomain.net/thumbnail.png"),
    NewAsset::Upload({
       content_id,
       ipfs_hash,
       size,
       ...
    }),
];

metat = VideoMetadata {
    ...
    video: 1,
    thumbnail_photo: 0,
    ...
};
```

mnaamani:
Just tagging @metmirr because information here is relevant for your work on query-node

mnaamani:
> Just tagging @metmirr because information here is relevant for your work on query-node

and @ondratra of course

content-metadata-protobuf/doc/index.md
```diff
@@ -0,0 +1,326 @@
...
+
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| title | [string](#string) | optional | Channel Title |
+| description | [string](#string) | optional | Channel Description |
+| is_public | [bool](#bool) | optional | Wether to display channel to the public |
+| language | [string](#string) | optional | ISO_639-1 Language [Code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) |
+| cover_photo | [uint32](#uint32) | optional | index into OOB array |
```
bedeho: This term `OOB` array is used as a description for many fields, but it's not unpacked anywhere I can see what that means. This is needed.

content-metadata-protobuf/doc/index.md
```diff
@@ -0,0 +1,326 @@
...
+<a name=".SeasonMetadata"></a>
+
+### SeasonMetadata
+
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| title | [string](#string) | optional |  |
+| description | [string](#string) | optional |  |
```
bedeho, line 165: In the current Atlas design a season actually also has a cover photo.

content-metadata-protobuf/doc/index.md
```diff
@@ -0,0 +1,326 @@
...
+<a name=".PublishedBeforeJoystream"></a>
+
+### PublishedBeforeJoystream
+Publication status before joystream
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| is_published | [bool](#bool) | optional | Was video published before joystream platform |
+| timestamp | [uint32](#uint32) | optional | Unix timestamp in milli-seconds |
```
bedeho: Won't this time format prevent accurate dates for anything prior to epoch in 1970? The a major use case for this public domain stuff, which by def. is far back, so this makes that hard.

bedeho (changes requested):
Nice job, only very minor comments.

content-metadata-protobuf/doc/index.md
mnaamani: I'll try to add a description into the generated docs and a link from each field description where it is referenced.

content-metadata-protobuf/doc/index.md
mnaamani: will add a `cover_photo` field to `SeasonMetadata`

content-metadata-protobuf/doc/index.md
mnaamani: I originally went with the timestamp like we have for current content dir schema, but you make a very good point actually. 
We can go with a simple string encoding like: "YYYY-MM-DD" which would be sufficient.
https://www.iso.org/iso-8601-date-and-time-format.html

mnaamani:
I addressed all comments in ee49f91d3e89321697c4912cbc2b22bc8afd16a8

content-metadata-protobuf/doc/index.md
mnaamani: done in https://github.com/Joystream/joystream/commit/ee49f91d3e89321697c4912cbc2b22bc8afd16a8

content-metadata-protobuf/doc/index.md
mnaamani: done in https://github.com/Joystream/joystream/commit/ee49f91d3e89321697c4912cbc2b22bc8afd16a8

content-metadata-protobuf/doc/index.md
mnaamani: done in https://github.com/Joystream/joystream/commit/ee49f91d3e89321697c4912cbc2b22bc8afd16a8

bedeho (approved):
Great job.

Issue #2044, opened 2021-01-19
Storage for all


bedeho:
Currently storage objects are directly owned by members, this has at least two problems

1. when you transfer ownership of a channel to someone else, the runtime will need to enumerate and update storage object ownership to that new member, if its even a member... see next point.
2. non-member ownership is required, such as a particular curator group, or a DAO in the future.

Meanwhile, we still want members to also directly own storage assets, so basically we get.

```Rust

enum WorkinGroupType {
 Builders,
 Marketers,
 Membership,
 ...
}

enum AbstractStorageObjectOwner {
 Channel(ChannelId), // acts through content directory module, where again DAOs can own channels for example
 DAO(DAOId), // acts through upcoming `content_finance` module
 Council, // acts through proposal system
 WorkingGroup(WorkinGroupType) // acts through new extrinsic in working group
 }

// new owner type for storage object struct
enum StorageObjectOwner {
 Member(MemberId), 
 AbstractStorageObjectOwner(AbstractStorageObjectOwner)
}
```

and the extrinsics would be

```Rust

// auth using origin using normal checks
fn upload_as_member(origin, member_id: MemberId, content: ...);

// auth using Root origin check
fn upload(origin,  abstract_owner:  AbstractStorageObjectOwner , content: ...)
```

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2099

PR #2099, opened 2021-01-26
Storage for all

iorveth:
https://github.com/Joystream/joystream/issues/2044

### DIFF
--- a/runtime-modules/storage/src/data_directory.rs
+++ b/runtime-modules/storage/src/data_directory.rs
@@ -98,14 +98,14 @@ decl_error! {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Clone, Encode, Decode, PartialEq, Eq, Debug)]
 pub enum WorkinGroupType {
+    ContentDirectory,
     Builders,
-    Marketers,
-    Membership,
+    StorageProviders,
 }
 
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Clone, Encode, Decode, PartialEq, Eq, Debug)]
-pub enum AbstractStorageObjectOwner {
+pub enum AbstractStorageObjectOwner<ChannelId, DAOId> {
     Channel(ChannelId), // acts through content directory module, where again DAOs can own channels for example
     DAO(DAOId),         // acts through upcoming `content_finance` module
     Council,            // acts through proposal system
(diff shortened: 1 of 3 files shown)


runtime-modules/storage/src/data_directory.rs
```diff
@@ -89,6 +95,31 @@ decl_error! {
     }
 }
 
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, PartialEq, Eq, Debug)]
+pub enum WorkinGroupType {
+    Builders,
```
bedeho, line 102: I think for this release lets limit it to `ContentDirectory`, `Builders`, `StorageProviders`

runtime-modules/storage/src/data_directory.rs
```diff
@@ -40,6 +40,12 @@ use crate::data_object_type_registry;
 use crate::data_object_type_registry::IsActiveDataObjectType;
 use crate::{MemberId, StorageProviderId, StorageWorkingGroup, StorageWorkingGroupInstance};
 
+// Temporary representation.
+type ChannelId = u64;
```
bedeho, line 44: Both of these types need to be configured, because they have to match types used in other modules.

bedeho (changes requested):
Solid work, only minor comments.

runtime-modules/storage/src/data_directory.rs
iorveth: Agree, they should be generic in data structures. Though we don`t have DAOs and new content directory pallets yet to be able to refer them.

iorveth: https://github.com/Joystream/joystream/pull/2099/commits/6f7481efd0170f63797ed930f5bf74e251773ac5

runtime-modules/storage/src/data_directory.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2099/commits/6f7481efd0170f63797ed930f5bf74e251773ac5

runtime-modules/storage/src/data_directory.rs
bedeho: I think you misunderstood what I meant here, I meant the type should come from the runtime via the `Trait`.

iorveth: Nevertheless, i assume that they would come via trait inheritance mechanism from other runtime pallet `Trait`s.
Similar to 

```
// Alias for the member id.
pub(crate) type MemberId<T> = <T as membership::Trait>::MemberId;

/// Storage provider is a worker from the working group module.
pub type StorageProviderId<T> = working_group::WorkerId<T>;
```

```
pub trait Trait:
 ....
    + membership::Trait
    + working_group::Trait<StorageWorkingGroupInstance>
{
```

Either way, i suppose, that  there will be some integration work right after new content directory and daos pallets would be complete.

iorveth: Does it make sense?

bedeho: ok.

Issue #2030, opened 2021-01-16
Atomic multi upload


bedeho:
A atomic version of `add_content` in the `DataDirectory` pallet, so something like

```Rust
struct ContentParameters {
            content_id: T::ContentId,
            type_id: <T as data_object_type_registry::Trait>::DataObjectTypeId,
            size: u64,
            ipfs_content_id: Vec<u8>
}

pub fn multi_add_content(
    origin,
    member_id: MemberId<T>,
    content: Vec<ContentParameters>
)       
```

which only has side-effects if adding everything in `content` will suceed, and if so emits a new event `ContentMultiAdded(ContentId, MemberId, Vec<ContentParameters>)`.

This is unfortunately temporary, as we will redo the storage system in a later release, however having this now allows other parts of the infra+apps to behave as they ultimately should.

mnaamani:
This looks reasonable.
The event will more likely be: `ContentMultiAdded(StorageObjectOwner, Vec<ContentParameters>)` to reflect the ownership as per https://github.com/Joystream/joystream/issues/2044

iorveth:
Is there any reason to have both versions similar to https://github.com/Joystream/joystream/issues/2044

// auth using origin using normal checks
```
fn multi_add_content_as_member(
    origin,
    member_id: MemberId<T>,
    content: Vec<ContentParameters>
);
```

// auth using Root origin check
```
fn multi_add_content(
  origin,  
  abstract_owner:  AbstractStorageObjectOwner , 
  content:: Vec<ContentParameters>
)
```

iorveth:
cc @bedeho @mnaamani

bedeho:
Yes I think so, because these groups should be able to just directly upload to the storage system for purposes outside of teh content directory.

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2103

Issue #2070, opened 2021-01-25
Simple Content Directory (revised)


mnaamani:
## Simple Content Directory for Sumer runtime

Based on sketch and ideas in https://github.com/Joystream/joystream/issues/2037

- updated Feb 1 2021
  - metadata for categories
  - revenue field for channel
- updated: Jan 29 2021
  - removed any state about assets
  - no more blocking of deleting storage objects

## Features Overview

1. Channels are owned by members, curator groups or DAOs.
1. Channel ownership exchange with payment
1. Channels must have a JOY treasury which receives revenues from content later.
1. Update representations to use opaque metadata blobs not stored in storage, or validated for parts of the metadata that can be validated statelesslly.
1. There must be mutually exclusive storage locations for assets, in old schemas assets could simultaneously live on web and Joystream infra.
1. Atomic content directory and storage system interaction
1. DAO & curator ability to upload as these categories: this requires minor adjustment in storage system
1. Video removal
1. Channel removal,
1. Users can manually delete assets when they are getting low on free space.
1. Channel categories
1. Playlists
1. Series: add or update with one transaction and uploading multiple assets
1. Featured videos are locked.
1. Persons

## Ownership
Everything under a channel (including data objects for assets) are owned by the channel owner. When a new video is created a new dataobject must be created for its assets. A data object can only be attributed to a single asset.

## Data Model
- Channel
  - Has an owner
  - Is in a ChannelCategory

- Video
  - Belongs to a channel
  - Is in a VideoCategory
  - Can be marked as featured
  - Can be marked as curated

- Playlist
  - Is in a channel
  - Is a collection of Videos from same channel
  - Belongs to a Channel

- Featured Videos
   - Videos marked as featured are in the "collection of featured videos"

- Series
  - Is a collection of Seasons
  - Belongs to a channel
- Season
  - Season belongs to a Series
  - A Season is a collection of ordered episodes
- Episode
   - An episode is a Video from same channel

- Person
  - Can be created by Members or Curators (PersonActor)
  - Can be added by Channel owners to a Video they own

## Pseudo Code

Please review actual [implementation](https://github.com/Joystream/joystream/pull/2123) as some types have been slightly modified 
### Types

```Rust

// To be implemented by current storage data_directory runtime module.
// Defined in 'common' package
Trait StorageSystem {

  // Should hook into call on storage system,
  // but requires rich error (with reasons)  types.
  // caller already knows the `ContentId`s as they are part of
  // the ContentUploadParameters
[...]

PR #2103, opened 2021-01-27
Atomic multi upload

iorveth:
Closes https://github.com/Joystream/joystream/issues/2030

Based on https://github.com/Joystream/joystream/pull/2099

### DIFF
--- a/runtime-modules/common/src/lib.rs
+++ b/runtime-modules/common/src/lib.rs
@@ -4,10 +4,10 @@
 pub mod constraints;
 pub mod currency;
 pub mod origin;
-pub mod working_group;
 pub mod storage;
+pub mod working_group;
 
-use codec::{Decode, Codec, Encode};
+use codec::{Codec, Decode, Encode};
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
 
@@ -16,13 +16,13 @@ use sp_arithmetic::traits::BaseArithmetic;
 use sp_runtime::traits::{MaybeSerialize, Member};
 
 /// Member id type alias
-pub type MemberId<T> = <T as Trait>::MemberId;
+pub type MemberId<T> = <T as MembershipTypes>::MemberId;
 
 /// Actor id type alias
-pub type ActorId<T> = <T as Trait>::ActorId;
+pub type ActorId<T> = <T as MembershipTypes>::ActorId;
 
 /// Generic trait for membership dependent pallets.
-pub trait Trait: system::Trait {
+pub trait MembershipTypes: system::Trait {
     /// Describes the common type for the members.
     type MemberId: Parameter
         + Member
@@ -44,8 +44,11 @@ pub trait Trait: system::Trait {
         + MaybeSerialize
         + Ord
         + PartialEq;
+}
 
-    /// Channel id representation
+/// Generic trait for strorage ownership dependent pallets.
+pub trait StorageOwnership {
+    /// Channel id representation.
     type ChannelId: Parameter
         + Member
         + BaseArithmetic
@@ -56,7 +59,7 @@ pub trait Trait: system::Trait {
         + Ord
         + PartialEq;
 
-    /// DAO id.
+    /// DAO id representation.
     type DAOId: Parameter
         + Member
         + BaseArithmetic
@@ -66,8 +69,8 @@ pub trait Trait: system::Trait {
         + MaybeSerialize
         + Ord
         + PartialEq;
-        
-    /// Content id.
+
+    /// Content id representation.
     type ContentId: Parameter
         + Member
         + BaseArithmetic
@@ -78,7 +81,7 @@ pub trait Trait: system::Trait {
         + Ord
         + PartialEq;
 
-    /// DAO object type id.
+    /// Data object type id.
     type DataObjectTypeId: Parameter
         + Member
         + BaseArithmetic
--- a/runtime-modules/common/src/lib.rs
+++ b/runtime-modules/common/src/lib.rs
@@ -71,15 +71,7 @@ pub trait StorageOwnership {
         + PartialEq;
 
     /// Content id representation.
-    type ContentId: Parameter
-        + Member
-        + BaseArithmetic
-        + Codec
-        + Default
-        + Copy
-        + MaybeSerialize
-        + Ord
-        + PartialEq;
+    type ContentId: Parameter + Member + Codec + Default + Copy + MaybeSerialize + Ord + PartialEq;
 
     /// Data object type id.
     type DataObjectTypeId: Parameter
--- a/runtime-modules/common/src/lib.rs
+++ b/runtime-modules/common/src/lib.rs
@@ -5,10 +5,90 @@ pub mod constraints;
 pub mod currency;
 pub mod origin;
 pub mod working_group;
+pub mod storage;
 
-use codec::{Decode, Encode};
+use codec::{Decode, Codec, Encode};
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
+
+use frame_support::Parameter;
+use sp_arithmetic::traits::BaseArithmetic;
+use sp_runtime::traits::{MaybeSerialize, Member};
+
+/// Member id type alias
+pub type MemberId<T> = <T as Trait>::MemberId;
+
+/// Actor id type alias
+pub type ActorId<T> = <T as Trait>::ActorId;
+
+/// Generic trait for membership dependent pallets.
+pub trait Trait: system::Trait {
+    /// Describes the common type for the members.
+    type MemberId: Parameter
+        + Member
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + Ord
+        + PartialEq;
+
+    /// Describes the common type for the working group members (workers).
+    type ActorId: Parameter
+        + Member
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + Ord
+        + PartialEq;
+
+    /// Channel id representation
+    type ChannelId: Parameter
+        + Member
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + Ord
+        + PartialEq;
+
+    /// DAO id.
+    type DAOId: Parameter
+        + Member
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + Ord
+        + PartialEq;
+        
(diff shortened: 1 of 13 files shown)


mnaamani:
can you rebase on sumer or merge sumer to see the work specific to this PR

mnaamani:
> can you rebase on sumer or merge sumer to see the work specific to this PR

oh sorry no need

runtime-modules/storage/src/data_directory.rs
```diff
@@ -204,14 +212,21 @@ decl_event! {
...
         /// - Id of the relationship.
-        /// - Id of the member.
+        /// - StorageObjectOwner enum.
         ContentAdded(ContentId, StorageObjectOwner),
 
+        /// Emits on atomic adding of the content.
+        /// Params:
+        /// - StorageObjectOwner enum.
+        /// - MultiContent parameters representation.
+        ContentMultiAdded(StorageObjectOwner, MultiContent),
```
mnaamani: For consistency with `ContentAdded` event perhaps `ContentMultiAdded(Vec<ContentId>, StorageObjectOwner)` instead?

Or combine both events into a single `ContentAdded(Vec<ContentParameters>, StorageObjectOwner)` I don't think we need a distinct event to distinguish between multi and single uploads

runtime-modules/storage/src/data_directory.rs
```diff
@@ -238,45 +253,31 @@ decl_module! {
         /// Maximum objects allowed per inject_data_objects() transaction
         const MaxObjectsPerInjection: u32 = T::MaxObjectsPerInjection::get();
 
-        /// Adds the content to the system. Member id should match its origin. The created DataObject
+        /// Adds the content to the system. Requires root privileges. The created DataObject
         /// awaits liaison to accept or reject it.
         #[weight = 10_000_000] // TODO: adjust weight
         pub fn add_content(
```
mnaamani: add_content can call into multi_add_content since its just doing the same thing for a single upload.
Also I wonder if we need two separate calls, since you can upload a single item with `multi_add_content()` by passing a Vec of size 1.

mnaamani (changes requested):
The add_content and add_multi_content methods need to be an `impl` of a `StorageSystsem` Trait so the new content directory can be configured to use data_directory module as its storage backend at runtime. see: psuedo code section here: https://github.com/Joystream/joystream/issues/2070

I think we can define the Trait in the common package?

runtime-modules/storage/src/data_directory.rs
mnaamani: the add content calls need some good Error types to return in DispatchResult for the various ways they can fail. When you add quota limits there will be more ways they can fail.

mnaamani:
> The add_content and add_multi_content methods need to be an `impl` of a `StorageSystsem` Trait so the new content directory can be configured to use data_directory module as its storage backend at runtime. see: psuedo code section here: #2070

I'm seeing the pattern we are now following is in-fact to implement this "integration" in the runtime crate's integration module.

runtime-modules/storage/src/data_directory.rs
iorveth: > add_content can call into multi_add_content since its just doing the same thing for a single upload.
> Also I wonder if we need two separate calls, since you can upload a single item with `multi_add_content()` by passing a Vec of size 1.

https://github.com/Joystream/joystream/pull/2103/commits/ae0e77f2a511ab24bd7cc07e689d20b8bfa87d74 Done here

iorveth: > the add content calls need some good Error types to return in DispatchResult for the various ways they can fail. When you add quota limits there will be more ways they can fail.

Don't sure if i understand what you mean under  `good Error types`. I see, there are 2 different checks in `ensure_content_is_valid` that return different kind of errors.

iorveth: `When you add quota limits there will be more ways they can fail.`

I`ll add an additional type(s) of errors for quota limits feature

runtime-modules/storage/src/data_directory.rs
iorveth: `Or combine both events into a single ContentAdded(Vec<ContentParameters>, StorageObjectOwner) I don't think we need a distinct event to distinguish between multi and single uploads`

I prefer to go with this approach. Done here https://github.com/Joystream/joystream/commit/ae0e77f2a511ab24bd7cc07e689d20b8bfa87d74

bedeho (changes requested):
Nice work.

1. Do we need an upper bound on the length of `content`? 
2. Can you add to methods `can_add_content`and `can_add_content_as_member` which just returns either an rich type safe error code, or returns rich success information. These will be used by the content directory and other systems to check if they will be successful in ding the upload.

iorveth:
```
Nice work.
1. Do we need an upper bound on the length of content?
```
Thank you. I hope it is reasonable to have this restriction. Should i make a separate issue for that?

```
Can you add to methods can_add_contentand can_add_content_as_member which just returns either an rich type safe error code, or returns rich success information. 
These will be used by the content directory and other systems to check if they will be successful in ding the upload.
```

I can, but what is the point to have a separate check here? 
I mean, currently we make check, which returns a rich error and perform upload if all checks succeeded simultaneously, can you explain a use case for that, please?

iorveth:
```
The add_content and add_multi_content methods need to be an impl of a StorageSystsem Trait so the new content directory can be configured to use data_directory module as its storage backend at runtime. see: psuedo code section here: #2070

I think we can define the Trait in the common package?
```

Good point, https://github.com/Joystream/joystream/pull/2103/commits/f045fa5c86e7a0260c6a37e5a33f8e73fb3c7c77
I also moved types to the common crate to make them reusable among the pallets (content-directory, storage)

runtime-modules/storage/src/lib.rs
```diff
@@ -14,7 +14,19 @@ pub type StorageWorkingGroupInstance = working_group::Instance2;
...
 
 // Alias for the member id.
-pub(crate) type MemberId<T> = <T as membership::Trait>::MemberId;
+pub(crate) type MemberId<T> = <T as common::Trait>::MemberId;
+
+// Alias for the content id.
+pub(crate) type ContentId<T> = <T as common::Trait>::ContentId;
+
+/// Channel id representation
+pub(crate) type ChannelId<T> = <T as common::Trait>::ChannelId;
```
mnaamani: This is not the correct place actually where this type should be defined. It will be available in the runtime primitives configuration and should be passed to the storage Trait as a configurable type.

mnaamani: see here: https://github.com/Joystream/joystream/pull/2123/files#diff-00cafb309cb91b1faac60e76701a3ffc8a5c26aee603b17d81514383615a0c3fR45

runtime-modules/storage/src/tests/mock.rs
```diff
@@ -130,6 +131,15 @@ impl pallet_timestamp::Trait for Test {
     type MinimumPeriod = MinimumPeriod;
 }
 
+impl common::Trait for Test {
```
mnaamani: The content directory and the storage system share these same types so it is a nice idea to have a Trait combining these shared types, but I think we need a better name for the trait type itself.

runtime-modules/common/src/lib.rs
```diff
@@ -5,11 +5,91 @@ pub mod constraints;
...
+        + BaseArithmetic
+        + Codec
+        + Default
+        + Copy
+        + MaybeSerialize
+        + Ord
+        + PartialEq;
+        
+    /// Content id.
+    type ContentId: Parameter
```
mnaamani: These types don't really fit given the description of the trait being "for membership dependent pallets."
This is related to my other comment on this `common::Trait`

runtime-modules/storage/src/lib.rs
iorveth: I made them configurable through the common trait for the purpose of reusability between new content directory and storage.

runtime-modules/common/src/lib.rs
iorveth: Done here https://github.com/Joystream/joystream/pull/2103/commits/a59562e0991801a84f1a16d91f6326129cca744c

runtime-modules/storage/src/tests/mock.rs
iorveth: https://github.com/Joystream/joystream/pull/2103/commits/a59562e0991801a84f1a16d91f6326129cca744c

mnaamani:
> ```
> Nice work.
> 1. Do we need an upper bound on the length of content?
> ```
> 
> Thank you. I hope it is reasonable to have this restriction. Should i make a separate issue for that?
> 
> ```
> Can you add to methods can_add_contentand can_add_content_as_member which just returns either an rich type safe error code, or returns rich success information. 
> These will be used by the content directory and other systems to check if they will be successful in ding the upload.
> ```
> 
> I can, but what is the point to have a separate check here?
> I mean, currently we make check, which returns a rich error and perform upload if all checks succeeded simultaneously, can you explain a use case for that, please?

I think here the the purpose is to allow the content directory module to do its own check before calling add_content on the storage module, so it can provide its own error types.

mnaamani:
The `can_add_content()` check is fine, but just returning a bool seems limited. I think it should return a `Result<(), Error>` and the same set of errors that could come as a result of calling `add_content()`. So you can in-fact re-use it from `add_content()`.

Last point but this can be done later when you work on working groups and proposals. You introduced a new `enum WorkinGroupType` (which is actually misspelled) in `common/src/storage.rs` which seems un-necessary. There is a `enum WorkingGroup` defined in `common/src/working_group.rs` which we can re-use. Unless we want to limit which working groups can actually directly own data objects in storage.

iorveth:
> The `can_add_content()` check is fine, but just returning a bool seems limited. I think it should return a `Result<(), Error>` and the same set of errors that could come as a result of calling `add_content()`. So you can in-fact re-use it from `add_content()`.
>
Good point, https://github.com/Joystream/joystream/pull/2103/commits/31f6bbc4e686c6204db79bbadf2a852d31c486ed

> Last point but this can be done later when you work on working groups and proposals. You introduced a new `enum WorkinGroupType` (which is actually misspelled) in `common/src/storage.rs` which seems un-necessary. There is a `enum WorkingGroup` defined in `common/src/working_group.rs` which we can re-use. Unless we want to limit which working groups can actually directly own data objects in storage.
>
Thank you for the suggestion, i overlooked that they are the same. https://github.com/Joystream/joystream/pull/2103/commits/f4170d4c446bc611598250fd30322f70a2d55df0

mnaamani (approved):
LGTM

bedeho (approved):
LGTM.

PR #2107, opened 2021-01-29
Add weight-to-fee analysis notebook

conectado:
This commit adds a notebook to analyze a "price to demand" function, it should be flexible enough to be very simple to change to a different "price to demand". 

There is a dashboard to see the different values of different extrinsic given the demand function and other parameters.

The instructions to run the notebook are in `analyses/fee-analysis/README.md`


bedeho (commented):
This seems extremely well done, but unfortunately I have zero experience with Jupyter, and it took me a long time to just get this to run I afraid. I am seeing some `FileNotFoundError` error message in here, not sure if that is because I need some weight function to run this? At any rate, nothing happened when I pressed "Run".

conectado:
> This seems extremely well done, but unfortunately I have zero experience with Jupyter, and it took me a long time to just get this to run I afraid. I am seeing some `FileNotFoundError` error message in here, not sure if that is because I need some weight function to run this? At any rate, nothing happened when I pressed "Run".

Sorry, forgot to push the latest changes fixing this.

It should be working now.

bedeho:
I see this now, not sure what the significance is

Screenshot 2021-01-31 at 20 49 13

conectado:
> I see this now, not sure what the significance is
> 
> Screenshot 2021-01-31 at 20 49 13

This should only happen when in `params_avg` you add an extrinsic but some of the variables that extrinsic is using are not defined. This never happened to me because there was some undefined behavior since I didn't take into account that some extrinsics have repeated names across different pallets, so when I added `add_post` to `params_avg` I never realized that it could refer to any of the `add_post` in the different pallets.

To solve this I introduced these changes:
* Now extrinsics are prefixed with their pallet name.
* Hardened the calculation of the total weight so that if this case ever repeated only a warning is shown.

I hope it works fine now!

conectado:
Added a CLI tool that can output files with the weight calculation for each extrinsic in a CSV and optionally include the fee analysis considering the parameters given in `analysis_cli/config.json`.

bedeho (approved):
Great job.

Issue #2072, opened 2021-01-25
quota limits utilisation tracking for all actors


iorveth:
Introduce quota limits utilisation tracking for all actors: same fixed number for everyone to begin with, lead can adjust.

Each actor instance can have a differnet limit. So member A has different limit from member B, and so on

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2122

PR #2122, opened 2021-02-01
Quota limits

iorveth:
Closes https://github.com/Joystream/joystream/issues/2072


runtime-modules/storage/src/data_directory.rs
```diff
@@ -164,6 +208,11 @@ decl_storage! {
         /// Maps data objects by their content id.
         pub DataObjectByContentId get(fn data_object_by_content_id) config():
             map hasher(blake2_128_concat) T::ContentId => Option<DataObject<T>>;
+
+        pub Quotas get(fn quotas) config():
```
bedeho, line 252: Can you add a comment here which describes when a quota is created?

runtime-modules/storage/src/data_directory.rs
```diff
@@ -153,6 +162,41 @@ pub struct DataObjectInternal<
     pub ipfs_content_id: Vec<u8>,
 }
 
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, PartialEq, Eq, Debug, Default)]
+pub struct Quota {
```
bedeho, line 186: I should have been more clear here that the limit is in terms of size, not in terms of number of objects sorry.

However, I think it actually could make sense to have both, so lets have both kinds of limits+usage in this struct: total size and number of objects.

runtime-modules/storage/src/data_directory.rs
```diff
@@ -342,7 +427,28 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T> {
+    // Ensures quota limit constraint satisfied, returns delta.
+    fn ensure_quota_limit_constraint_satisfied(
```
bedeho: Very elegant :)

bedeho (changes requested):
Looks like a good start.

- I think we probably also want two global limits, because people can make many channels or memberships and spam. So lets have a global size limit and global object counter limit, which goes across all all owners. These limits should also be adjustable in storage through proposals as before.
- Can we also introduce a blocker variable which blocks all new uploads, sort of like a an emergency switch, can be updated by proposal or lead.

runtime-modules/storage/src/data_directory.rs
iorveth: Done here https://github.com/Joystream/joystream/pull/2122/commits/ff989c45bcd0b849082ec16ae25b9adbab5cd13a#diff-5f5bdbc581be08dfecff6870d8fcf9b767904a351d254a70cf35267e5465589bR185

runtime-modules/storage/src/data_directory.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2122/commits/ff989c45bcd0b849082ec16ae25b9adbab5cd13a#diff-5f5bdbc581be08dfecff6870d8fcf9b767904a351d254a70cf35267e5465589bR241

runtime-modules/storage/src/data_directory.rs
iorveth: Thank you, it was hard to come up with :) https://github.com/Joystream/joystream/pull/2122/commits/ff989c45bcd0b849082ec16ae25b9adbab5cd13a#diff-5f5bdbc581be08dfecff6870d8fcf9b767904a351d254a70cf35267e5465589bR526

iorveth:
```
I think we probably also want two global limits, because people can make many channels or memberships and spam. So lets have a global size limit and global object counter limit, which goes across all all owners. 
These limits should also be adjustable in storage through proposals as before.
```
Great idea, implemented this here https://github.com/Joystream/joystream/pull/2122/commits/ff989c45bcd0b849082ec16ae25b9adbab5cd13a#diff-5f5bdbc581be08dfecff6870d8fcf9b767904a351d254a70cf35267e5465589bR252

iorveth:
`Can we also introduce a blocker variable which blocks all new uploads, sort of like a an emergency switch, can be updated by proposal or lead.`

Sure, though i feel like it`s better to create a separate issue for that feature.

bedeho:
> Sure, though i feel like it`s better to create a separate issue for that feature.

Agreed.

runtime-modules/storage/src/data_directory.rs
```diff
@@ -162,38 +174,55 @@ pub struct DataObjectInternal<
     pub ipfs_content_id: Vec<u8>,
 }
 
+#[derive(Clone, Copy)]
+pub struct Delta {
```
bedeho: Can we change this name to `Voucher`, I think quota is not actually fully accurate.

bedeho (changes requested):
Nice fixes, one last improvement, sorry for not pointing out sooner.

runtime-modules/storage/src/data_directory.rs
iorveth: https://github.com/Joystream/joystream/pull/2122/commits/66937035d6a16bbb5985e533c1064139fb8b41c9

bedeho (approved):
Excellent.

Issue #2084, opened 2021-01-25
Content Directory: Integration


mnaamani:
Integrate content directory into runtime (replacing old pallet)

Integrate with storage data_directory module.

Runtime-update hook to initialize content directory state.

mnaamani:
implemented in #2123

Issue #2078, opened 2021-01-25
Content Directory: Curator Groups


mnaamani:
Expanded from https://github.com/Joystream/joystream/issues/2074

Implement same curator groups feature from babylon content directory

mnaamani:
implemented in #2123

Issue #2075, opened 2021-01-25
Update @joystream/types Library: Content Directory types


mnaamani:

mnaamani:
implemented in #2123

Issue #2051, opened 2021-01-21
Implement Core Types and Events in runtime module


mnaamani:
Implement runtime, storage and event types for new content directory.

Update joystream/types package at the same time

mnaamani:
Implemented in implemented in #2123

PR #2123, opened 2021-02-01
new content directory: initial setup

mnaamani:
Initial types, events and extrinsics skeleton implementation, and cleanup of types package:

closes https://github.com/Joystream/joystream/issues/2051
closes https://github.com/Joystream/joystream/issues/2075
closes https://github.com/Joystream/joystream/issues/2078
closes https://github.com/Joystream/joystream/issues/2084

Details:

- [x] Dropping old content-directory runtime module
- [x] Dropping old content-directory module types in `@joystream/types` package
- [x] Renamed `media` namespace in types library to `storage` in `@joystream/types` package
- [x] "Copied" CuratorGroup functionality from old module into new content module
- [x] Adding new content module types to `@joystream/types`
- [x] Implementing new content runtime module
  - [x] storage
  - [x] events
  - [x] place holder dispatchables in decl_module


mnaamani:
@bedeho I'm still adding some rust docs but ready for initial review.

mnaamani:
cc: @metmirr @ondratra FYI final implementation of types and events

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    + Clone
+    + StorageOwnership
+    + MembershipTypes
+    + GovernanceCurrency
+{
+    /// The overarching event type.
+    type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+
+    /// Channel Transfer Payments Escrow Account seed for ModuleId to compute deterministic AccountId
+    type ChannelOwnershipPaymentEscrowId: Get<[u8; 8]>;
```
bedeho, line 79: What is this and `ChannelRevenueTreasuryId`?

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    Uri(Vec<u8>),
+}
+
+/// The owner of a channel, is the authorized "actor" that can update
+/// or delete or transfer a channel and its contents.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum ChannelOwner<MemberId, CuratorGroupId, DAOId> {
+    /// Do not use - Default value representing empty value
+    Nobody,
```
bedeho: Drop this, we we _must_ have a default value, then just pick one of the other ones, this is less bad than having a fully invalid state bleed into all of our business logic and state.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    // as this is the only field it is not an Option
+    /// Metadata update for the category.
+    new_meta: Vec<u8>,
+}
+
+/// Type representing an owned channel which videos, playlists, and series can belong to.
+/// If a channel is deleted, all videos, playlists and series will also be deleted.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct ChannelInternal<MemberId, CuratorGroupId, ChannelCategoryId, DAOId, Balance> {
```
bedeho: I think shamil calls these _records_, so can we call it `ChannelRecord`?

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    /// The category the channel belongs to
+    in_category: ChannelCategoryId,
+    /// The number of videos under this channel
+    number_of_videos: u32,
+    /// The number of playlists under this channel
+    number_of_playlists: u32,
+    /// The number of series under this channel
+    number_of_series: u32,
+    /// If curators have curated this channel or not
+    is_curated: bool,
```
bedeho: I think we are changing the name to something else, because `curated` is just so vague, I think `censored` is great and direct, but I think @kdembler may have some input here also.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    /// Earned revenue yet to be withdrawn by channel owner
+    revenue: Balance,
+    // TODO: I think we need to add these instead of the counters!
```
bedeho: Why?

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    /// The number of videos under this channel
+    number_of_videos: u32,
+    /// The number of playlists under this channel
+    number_of_playlists: u32,
+    /// The number of series under this channel
+    number_of_series: u32,
+    /// If curators have curated this channel or not
+    is_curated: bool,
+    /// Earned revenue yet to be withdrawn by channel owner
+    revenue: Balance,
```
bedeho: I guess this some how connects to the treasury things I saw up top. Not sure where this is going because I have not seen anything written about it.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    <T as ContentActorAuthenticator>::CuratorGroupId,
+    <T as Trait>::ChannelCategoryId,
+    <T as StorageOwnership>::DAOId,
+    BalanceOf<T>,
+>;
+
+/// A request to buy a channel by a new ChannelOwner.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct ChannelOwnershipTransferRequestType<ChannelId, MemberId, CuratorGroupId, DAOId, Balance>
```
bedeho: Can this just be `ChannelOwnershipTransferRequest`?

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    Member(MemberId),
+    Curator(CuratorId),
+}
+
+/// The authorized actor that may update or delete a Person.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum PersonController<MemberId> {
+    /// Do not use - Default value representing empty value
+    Nobody,
```
bedeho: Same here.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    /// If set, metadata update for the video.
+    new_meta: Option<Vec<u8>>,
+}
+
+/// A video which belongs to a channel. A video may be part of a series or playlist.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct Video<ChannelId, SeriesId, PlaylistId> {
+    in_channel: ChannelId,
+    // keep track of which seasons and playlists which reference the video
```
bedeho: # `in_playlists`

This reminded me that videos should be placeable into playlists owned by third parties, and this should not block the deletion of those videos by the owner. Hence, since a playlist is in some sense totally valid even if including dangling references, then the entire playlist can just have video inclusion information as metadata.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+    // - prevent removing a video if it is in a season (because order is important)
+    // - remove from playlist on deletion
+    in_series: Vec<SeriesId>,
```
bedeho: I think the assumption about videos has to be that they are in at most one _season_, so we just need `in_season: Option<SeasonId>`, and this season lives in a series which has same owner as video. If we want to have simple one step deletion of series or seasons, then we must constrain the number of seasons in a series, and the number of videos in a season, to do it safely in one transaction. We also need to have a list of seasons and videos as inline vectors in series and seasons respectively, as I think you may have even suggested above.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+}
+
+/// A Person represents a real person that may be associated with a video.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct Person<MemberId> {
+    /// Who can update or delete this person.
+    controlled_by: PersonController<MemberId>,
+    /// The number of videos this person appears in.
+    number_of_videos_person_involed_in: u32,
```
bedeho: Where is it stored what specific videos/seasons/series a person appears in? How this should be done depends on who manages creation and destruction of such relationships. One way is to allow content owners to add person to something. That scales well, and it prevents someone from spamming a person onto tons of content, like adding "Jesus" to everything. At the same time, do we then allow a person to be deleted from the directory if third parties added them to their content? I think that is OK. In that case, content can refer to dangling persons. In this case, presence of persons on content is just part of their metadata, and no reference counting is needed.

What do you think?

bedeho (changes requested):
Great start!

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+        ) -> DispatchResult {
+
+            // Ensure given origin is lead
+            ensure_is_lead::<T>(origin)?;
+
+            // Ensure CuratorGroup under given curator_group_id exists
+            let curator_group = Self::ensure_curator_group_exists(&curator_group_id)?;
+
+            // We should previously ensure that curator_group  maintains no classes to be able to remove it
+            curator_group.ensure_curator_group_maintains_no_classes()?;
```
bedeho: Seems like you need to fix all these old curator concepts, there are no classes.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+                curator_group.get_curators_mut().remove(&curator_id);
+            });
+
+            // Trigger event
+            Self::deposit_event(RawEvent::CuratorRemoved(curator_group_id, curator_id));
+            Ok(())
+        }
+
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn create_channel(
```
bedeho, line 642: Its not clear why we need to have these creation parameter types which only cover a subset of the parameter. Like, why isn't `assets` part of parameters class? it seems either we do not need those classes (which is my first thought given that we are within the parameter count bound of Rust), or all non `origin` params should be there. Same goes for update stuff.

bedeho (changes requested):
Forgot some stuff!

runtime-modules/content/src/lib.rs
kdembler: It would be great if we could use something other than `curated`, that really sounds like something hand-picked to be displayed more prominently. `censored` is straightforward and sounds much better to me

runtime-modules/content/src/lib.rs
mnaamani: ok

runtime-modules/content/src/lib.rs
mnaamani: sounds like a good convention :+1:

runtime-modules/content/src/lib.rs
mnaamani: Okat that also gives completely different meaning to what this field represents. I interpreted curated to mean something that was checked by curators and found to be of good quality and worth promoting over other non curated content. So clearly getting rid of this vague/ambiguous field makes sense.

And I agree censored is obviously quite clear on what it means.

runtime-modules/content/src/lib.rs
mnaamani: Yes so my thoughts here were that when the content finance or what ever mechanism decides to give a channel revenue, it wouldn't go directly to an account specified in the channel, but to the treasury account, and we keep track of the earned amount here. The channel owner can at a later time request to transfer their allocation to any account.

There are obviously pros and cons. I found the main advantage was that it easier to on the user to not have to keep their channel updated with an account id. Also depending on the owner the the account for revenue destination may not be known ahead of time?

For example also when making transfers of ownership requests, a destination account for revenue doesn't need to provided provided upfront. Of course whomever is transfering their channel should remember to claim any revenue before the transfer. 

Perhaps this is overly complicated?

runtime-modules/content/src/lib.rs
mnaamani: Its the seed to generate a deterministic AccountId without a corresponding (known private key) which will be used to store minted funds allocated to the revenue earned by channels. commented on approach [here](https://github.com/Joystream/joystream/pull/2123#discussion_r570898211)

runtime-modules/content/src/lib.rs
mnaamani: This is another case of the type simplification so I will rename it to `ChannelOwnershipTransferRequestRecord`

runtime-modules/content/src/lib.rs
mnaamani: ok

runtime-modules/content/src/lib.rs
mnaamani: Okay that makes sense. I was thinking of playlists really as a very simple series of videos. A series with a single season basically.  (interestingly a series can be though of as a playlist of playlists and perhaps even a series could be fully described in metadata and doesn't really need a lot of runtime state, but that would mean series would also not be constrained to only being composed of channel videos)

So the playlist you are describing is really a "curated list" of videos. In such a case does it make sense to allow a playlist to live "outside" of a channel?

runtime-modules/content/src/lib.rs
mnaamani: Your assumption is correct, and yes will update to do the tracking that way.

Regarding: suggestion of `Option<SeasonId>` as per current design we don't actually have a separate `SeasonId`, we only store  a `SeriesId => Series` map where

```
struct Season = {
  espisodes: Vec<VideoId>
}

struct Series = {
  seasons: Vec<Season>,
}
```

mnaamani: I was infact meaning to discuss this lack of SeasonId with query-node team to ask how the metadata for a season would be stored. So we might need to infact introduce a SeasonId => Season map and have

```
struct Season = {
  espisodes: Vec<VideoId>
}

struct Series = {
  seasons: Vec<SeasonId>,
}
```

runtime-modules/content/src/lib.rs
mnaamani: Obviously I hadn't expanded on the implementation of `add_person_to_video` and `remove_person_from_video()` but that is indeed the permission model I was planning to go with.

With respect to deletion, yes I don't see a cleaner way to deal with constraint and resulting dangling references. Putting it in metadata both reduces state burden and code to keep track of references, and moves the burden on UI apps to deal with missing persons (similarly with missing videos from a playlist)

runtime-modules/content/src/lib.rs
mnaamani: hehe I did rename classes to channels in most places seems i missed a couple :) :+1:

runtime-modules/content/src/lib.rs
mnaamani: I think at one point the asset information was in the parameters class, and yes there is no strict reason that it should be separated. I will favor putting all "metadata" sutff in the params struct, because over time that may grow to encompass other things and its better to have named fields rather than a longer list of args.

runtime-modules/content/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+
+/// Specifies how a new asset will be provided on creating and updating
+/// Channels, Videos, Series and Person
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum NewAsset<ContentParameters> {
+    /// Upload to the storage system
+    Upload(ContentParameters),
+    /// A url string pointing at an asset
+    Uri(Vec<u8>),
```
mnaamani: From chat: "we probably want to allow a vector of URIs, not just a single one, as a backup." - @bedeho 

Will update that in upcoming PR where I'm implementing the extrinsics

mnaamani: Atlas could even be creative about it and do some nice amimated transitions between the multiple urls that gives some dynamic life to a view. (depending on the type of asset of course)

runtime-modules/content/src/lib.rs
mnaamani: done in dd0ee6ce15e0d7a924b9620ffcdf258f0e585b57

runtime-modules/content/src/lib.rs
mnaamani: done in e8fcf5c537ccc8456d4b71ec3bfd89b66fdc794c

runtime-modules/content/src/lib.rs
mnaamani: done in ebdcd7f326692296d27c4d0baab3731a5e6e9bfc

runtime-modules/content/src/lib.rs
mnaamani: done in e8fcf5c537ccc8456d4b71ec3bfd89b66fdc794c

runtime-modules/content/src/lib.rs
mnaamani: done in dd0ee6ce15e0d7a924b9620ffcdf258f0e585b57

runtime-modules/content/src/lib.rs
mnaamani: fixed in 1a0052319299a38b85a42a17059dd900fe8dbde6

runtime-modules/content/src/lib.rs
mnaamani: Two additional thoughts come to mind:

1 - With the `Person` struct now being reduced to having a single field (PersonController), does it make sense to keep it as a struct an maintain the map:  `PersonById: PersonId => Person` or might as well just drop it and have a map: `PersonControlledBy: PersonId => PersonController`

2 - We have a similar situation with ChannelCategory and VideoCategory. Which are maintaining counters to prevent deletion of the categories if videos or channels are in them.. should I apply same treatment and let applications deal with dangling reference to category ids?

```
pub struct VideoCategory {
    /// The number of videos in this category.
    number_of_videos_in_category: u32,
}

pub struct ChannelCategory {
    /// The number of channels in this category.
    number_of_channels_in: u32,
}
```

runtime-modules/content/src/lib.rs
bedeho: This does feel overly complicated from my point of view, we already require miners, council members and all workers to provide reward accounts, it seems to work just fine. When it comes to transfer of accounts then you are quite right that one would have to set this during the transfer, but this is trivial really.

runtime-modules/content/src/lib.rs
bedeho: It could in theory, but Atlas is not designed for that, so would rather not try to go down that path now.

But you will need to update the metadata then for this?

mnaamani: Yes now that you merged the protobuf PR I can update the metadata message in this PR.

runtime-modules/content/src/lib.rs
bedeho: 1. If we limit the number of season to be a low number, and they are not big objects, then we could possibly just do what you have already done? It is safer and simpler, as long as the size does not become an issue.

2. Thinking more about this, I actually don't think it makes sense for normal videos to be part of series. This means that a video could be included in a random season of one series and another season of another. So if you just go staight to that video, you could not even get informed that it is part of some other canonical whole. So I think we need to have a distinct SeasonEpisode, which can only be in exactly one season and one series. I am not sure if it needs different metadata than videos, but possibly. I think we should opt to store an entire series as one state object, with severe limits on season lenght and season count per series.

runtime-modules/content/src/lib.rs
mnaamani: done in 59a7b3ca3fee7a7f73aac423d54a48183fd23772

runtime-modules/content/src/lib.rs
bedeho: I would keep them, even if just with one field. Its semantically cleaner and we will possibly find reasons to expand later.

runtime-modules/content/src/lib.rs
mnaamani: done in e19e1f677e2f3fd588ff7dd3d2446ca8b5538fc0

runtime-modules/content/src/lib.rs
mnaamani: So that we may know which playlists, videos, and series to delete when the channel is deleted.

runtime-modules/content/src/lib.rs
bedeho: 2. Following up on this point.

I guess that is ok, otherwise people can block deletion of categories by having their content in there. This does however mean that people will have invalid category information, which I guess is OK, at least for now.

mnaamani: Updated Person representation in runtime and added to video metadata in e19e1f677e2f3fd588ff7dd3d2446ca8b5538fc0
Added persons fields to Series and Season metadata in 8974e224017107a1fcb14217fcbbac9ea907c8a9

mnaamani: removed counters in Category types in 35c00c0dc428d8c97d02e9f9123f1d002da846fe

runtime-modules/content/src/lib.rs
mnaamani: done in e0a27bee7a23893463bec459e7a14e4bc9896789

runtime-modules/content/src/lib.rs
mnaamani: category assigned in metadata instead in a903a493d8f612c13c88dcfd482ae1301e315fe3

runtime-modules/content/src/lib.rs
mnaamani: Simple reward account implemented in 6cc09fd3b880a0ed08fb9e5ce225b7dae39a49c6

runtime-modules/content/src/lib.rs
mnaamani: done in 41af8a4ff0a47e9171763841be202aac4e4f9067 and 2a03591a491e1e1de3e1e898f192bc3dadfedced

runtime-modules/content/src/lib.rs
mnaamani: updated transfer request to include new reward account in 25d8174b60aabc09b3220b7964f9569407129e10

runtime-modules/content/src/lib.rs
mnaamani: done in cb6cb17e02a5c442b6173fb06c4c5434bcd85580

runtime-modules/content/src/lib.rs
mnaamani: The Series is stored a one state object currently.
Now I don't this a distinct `SeasonEpisode` is strictly required if we have for a Video struct a `in_series: Option<SeriesId>` (which I changed from `in_series: Vec<SeriesId>`) the runtime can apply the constraint of video can only appear in a single series. and in Atlas that will be sufficient to link to the series, if you land on that video.

Perhaps I can dig deeper into the series in a future release as we will not support it in atlas for sumer release?

mnaamani:
I addressed all the issues, and partial change for video in series issue https://github.com/Joystream/joystream/pull/2123#discussion_r571975076 and I've linked to this discussion in https://github.com/Joystream/joystream/issues/2083 so we don't forget about it.

bedeho (approved):
LGTM.

Issue #1998, opened 2021-01-09
Crowdfunded Bounties


bedeho:
# Background

Currently, the only way to fund the production of goods that create benefits to a broad set of platform participants is through a financing proposal or discretionary spending by a working group lead out of the group budget. These processes incur the transaction costs of beneficiaries having to convince a number of external decision-makers, such as a council financing quorum, that this is a good idea. For smaller initiatives that ideally should start and finish sooner, or where they depend on knowledge or insight that is not as broadly shared, these processes become too costly.

# Proposal

Crowd funded bounties which allows a member, or the council, to crowd fund work on projects with a public benefit. The funding mechanism for the bounties attempts to facliate two forms of crowd funding:

- **Assurance Contract**: It only triggers if some minimal quantity is raised, otherwise all funds are returned. Described here https://en.wikipedia.org/wiki/Assurance_contract.
- **Dominant Assurance Contract**: The proposer is the only person who can submit a bounty solution - and thus claim the raised funds, however, the proposer must put up a pool of funds which will be distributed among all third parties that fund the pool in the event that the minimum quantity is not raised. Described further here http://mason.gmu.edu/~atabarro/PrivateProvision.pdf.

In either case, an oracle is required to judge how much of the collected funds should be paid to any given contributor of work on the bounty, and this will either be a pre-specified member, or the council iself.

## Bounty Creation

Any member or the council can create a bounty by providing the following information.

- **Metadata:** A standardised structure document describing user facing information, for example a title, amount requested, deliverable, discovery metadata, link to forum etc. Is not stored in storage, chain only sees raw extrinsic payload blob, like rationales before.
- **Oracle:** Origin that will select winner(s), is either a given member or the council.
- **Cherry**: An mount of funding, possibly 0, provided by the creator which will be split among all other contributors should the min funding bound not be reached. If reached, cherry is returned to the creator. When council is creating bounty, this comes out of their budget, when a member does it, it comes from an account.
- **Screened Entrants:** The set of members who are allowed to submit their work, if not set, then it is open. Main use case for this is to model dominant assurance contract where member sets contribution cherry and him/herself sa only elidable worker.
- **Minimum Amount:** The minimum total quantity of funds, possibly 0, required for the bounty to become available for people to work on.
- **Max Amount:** Maximumm funding accepted, if this limit is reached, funding automatically is over.
- **Entrant Stake:** Amount of stake required, possibly 0, to enter bounty as entrant.
[...]

PR #2129, opened 2021-02-02
Create bounties for the runtime.

shamil-gadelshin:
#### Comments
- new `pallet_bounty` was created and added to the runtime
- bounty creation, cancellation, veto, funding are supported
- bounty creation is possible for both a member and a council
- work submission, judgment, funds withdrawal are not supported yet
- tests and benchmarks are ready for an existing code
- council budget manager (helper) was created

#### Extrinsics
- create_bounty
- cancel_bounty
- veto_bounty
- fund_bounty


runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+            judging_period: One::one(),
+            cherry,
+            ..Default::default()
+        };
+
+    }: create_bounty (RawOrigin::Root, params.clone(), metadata)
+    verify {
+        let bounty_id: T::BountyId = 1u32.into();
+
+        assert!(Bounties::<T>::contains_key(bounty_id));
```
conectado, line 104: We could add an assertion that `BountyCount` also mutated as expected.

The same for `create_bounty_by_,member`

shamil-gadelshin:
@bedeho 

The Substrate `treasury` pallet has [an implementation](https://github.com/paritytech/substrate/blob/c7f52d184727b11171eea458c4f277bc6b37df02/frame/treasury/src/lib.rs#L351) of your idea of bounty accounts.
I consider using it in the next PR of pallet_bounty.

conectado (commented):
Only left one comment and also this:

I think we should add an assertion after each bounty is created that the bounty exists as we expect in the storage just because it's easier to read from the assertions the pre-conditions for benchmarking.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+/// Defines who can submit the work.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum AssuranceContractType<MemberId> {
+    /// Anyone can submit the work.
+    Open,
+
+    /// Only specific members can submit the work.
+    Closed(Vec<MemberId>),
```
bedeho, line 105: Closed was originally intended to mean that only bounty creator can submit, and thus also get any cherry back if funding fails. In this version where any set of third parties can make entry, I assume creator still gets cherry back if funding fails?

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct BountyParameters<Balance, BlockNumber, MemberId> {
+    /// Origin that will select winner(s), is either a given member or a council.
+    pub oracle: OracleType<MemberId>,
+
+    /// Contract type defines who can submit the work.
+    pub contract_type: AssuranceContractType<MemberId>,
+
+    /// Bounty creator member ID, should be None if created by a council.
+    pub creator_member_id: Option<MemberId>,
```
bedeho: Can we make this slightly more explicit so something like using a type

```Rust
enum BountyCreator {
Council,
Member(MemberId)
}
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+pub type Bounty<T> = BountyRecord<
+    BalanceOf<T>,
+    <T as frame_system::Trait>::BlockNumber,
+    <T as common::Trait>::MemberId,
+>;
+
+/// Crowdfunded bounty record.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct BountyRecord<Balance, BlockNumber, MemberId> {
```
bedeho, line 223: Is there a reason we cannot just call this `Bounty` directly?

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// Bounty creation parameters.
+    pub creation_params: BountyParameters<Balance, BlockNumber, MemberId>,
```
bedeho, line 225: Are you 100% sure that every single value in `BountyParameters` is indeed required to be stored?

As the bounty progresses through the different stages an increasing number of the values become redundant, but its not that big of a problem. Addressing that would require lifting out smaller and smaller subsets into the different stage variants.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+    /// The current bounty stage.
+    pub stage: BountyStage<BlockNumber>,
+
+    /// The current unspent "cherry" balance.
+    pub cherry_pot: Balance,
```
bedeho: What does this do and why does it live here? What is meant by "unspent".

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+    /// The current unspent funding balance.
+    pub current_funding: Balance,
```
bedeho: What does this do and why does it live here? What is meant by "unspent".

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+        /// Double map for bounty funding. It stores member funding for bounties.
+        pub Funding get(fn funding_by_bounty_by_member): double_map hasher(blake2_128_concat)
+            T::BountyId, hasher(blake2_128_concat) MemberId<T> => BalanceOf<T>;
+
+        /// Count of all bounties that have been created.
+        pub BountyCount get(fn bounty_count): u32;
+    }
+}
+
+decl_event! {
```
bedeho, line 253: Can we make all of these have exhaustive parameters, so full extrinsic argument list.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+/// Crowdfunded bounty record.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct BountyRecord<Balance, BlockNumber, MemberId> {
+    /// Bounty creation parameters.
+    pub creation_params: BountyParameters<Balance, BlockNumber, MemberId>,
+
+    /// The current bounty stage.
+    pub stage: BountyStage<BlockNumber>,
```
bedeho: I don't think its possible to represent the stage in state like this in combination with

1. No iteration over bounties
2. Block height deadlines.

I have not looked through details, but something must fail here I think. I think the mildest possible "failure" would be that the stored `stage` value is not in synch with reality until some extrinsic occurs where its resynched. The right approach here should be that the stage itself is not stored, but just computable from a combination of when bounty was created, period deadlines and later milestone block heights.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// Funds provided by a bounty creator.
+    pub creator_funding: Balance,
+}
+
+/// Defines current bounty stage.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub enum BountyStage<BlockNumber> {
+    /// Bounty founding stage with starting block number.
+    Funding(BlockNumber),
```
bedeho: What do you mean by "starting block", are you saying a bounty can start accepting funding in a future block, but not right away? That was not part of the spec I think. Can it still be cancelled here?

Also, are you not losing this information now in later stages? that would seem to make it impossible to compute stage without advancing it through `on_finalized` computation.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+    /// A bounty was canceled.
+    Canceled,
```
bedeho, line 183: I do not think  cancel, veto are really distinct stages as far as the runtime is concerned, the only thing that matters when these actions are initiated is whether to go to a perpetual withdraw stage where funders can get their money back, or just to drop the bounty from the system right away if there are no such contributions that need to be recuperated.

bedeho (changes requested):
Promising start, but I think you need to think more deeply about how to organise the state of bounties to support our goals of no iteration and also no archival state. Here it seems you have come up with a state model with only very first stage in mind, and I do not think it works.

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: Ok

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Done.

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Removed.

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Renamed.

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Done.

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Refactored.

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Removed.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// Bounty creation parameters.
+    pub creation_params: BountyParameters<Balance, BlockNumber, MemberId>,
+
+    /// Total funding balance reached so far.
+    /// Includes initial funding by a creator and other members funding.
+    pub total_funding: Balance,
+
+    /// Bounty current state. It represents fact known about the bounty, eg.:
+    /// it was canceled or max funding amount was reached.
+    pub state: BountyMilestone<BlockNumber>,
```
bedeho: Rename field to `milestone`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+        /// # </weight>
+        #[weight = WeightInfoBounty::<T>::cancel_bounty_by_member()
+              .max(WeightInfoBounty::<T>::cancel_bounty_by_council())]
+        pub fn cancel_bounty(origin, creator: BountyCreator<MemberId<T>>, bounty_id: T::BountyId) {
+            let bounty_creator_manager = BountyCreatorManager::<T>::get_bounty_creator(
+                origin,
+                creator.clone(),
+            )?;
+
+            ensure!(
```
bedeho: 1. we will need to safely recover bounties in multiple places, so lets refactor this check+get into `ensure_bounty_exists`.
2. Let us avoid this mutable, this seems avoidable at little cost.

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+/// Balance alias for `balances` module.
+pub type BalanceOf<T> = <T as balances::Trait>::Balance;
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Bounty {
+        /// Bounty storage
+        pub Bounties get(fn bounties) : map hasher(blake2_128_concat) T::BountyId => Bounty<T>;
+
+        /// Double map for bounty funding. It stores member funding for bounties.
+        pub Funding get(fn funding_by_bounty_by_member): double_map hasher(blake2_128_concat)
```
bedeho: Can we rename `Funding` to `BountyContributions`

runtime-modules/bounty/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+        ///    - `O(1)` doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = WeightInfoBounty::<T>::fund_bounty()]
+        pub fn fund_bounty(
+            origin,
+            member_id: MemberId<T>,
+            bounty_id: T::BountyId,
+            amount: BalanceOf<T>
+        ) {
+            let account_id = T::MemberOriginValidator::ensure_member_controller_account_origin(
```
bedeho: We only allow contributing from controller account? I am fine with that, just confirming that this is what we wanted.

bedeho (changes requested):
Excellent work, you nailed fixing the state model here.

Not sure what we do about the bounty contribution cleanup in the success case, which presumably is the majority of cases. We may just have to accept it.

One alternative is to take k*`withdraw_fee` away from the initial contribution, and return it in exchange for cleanup, for some integer k>1, as an incentive to cleanup.

conectado (approved):
LGTM

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Done

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Done

runtime-modules/bounty/src/lib.rs
shamil-gadelshin: Yes. Renamed to `controller_account_id` for clarity.

bedeho (approved):
Excellent.

Issue #2133, opened 2021-02-03
Unify AbstractStorageObjectOwner & StorageObjectOwner


iorveth:
SSIA

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2134

PR #2134, opened 2021-02-03
Storage owner unification

iorveth:
Closes https://github.com/Joystream/joystream/issues/2133

Based on https://github.com/Joystream/joystream/pull/2122

### DIFF
--- a/runtime-modules/storage/src/data_directory.rs
+++ b/runtime-modules/storage/src/data_directory.rs
@@ -324,14 +324,9 @@ decl_module! {
             owner: StorageObjectOwner<MemberId<T>, ChannelId<T>, DAOId<T>>,
             content: Vec<ContentParameters<ContentId<T>, DataObjectTypeId<T>>>
         ) {
-            if let StorageObjectOwner::Member(member_id) = owner {
-                T::MemberOriginValidator::ensure_actor_origin(
-                    origin,
-                    member_id,
-                )?;
-            } else {
-                ensure_root(origin)?;
-            };
+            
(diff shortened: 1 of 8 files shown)


runtime-modules/storage/src/data_directory.rs
```diff
@@ -316,58 +316,25 @@ decl_module! {
...
         /// awaits liaison to accept or reject it.
         #[weight = 10_000_000] // TODO: adjust weight
         pub fn add_content(
             origin,
-            abstract_owner: AbstractStorageObjectOwner<ChannelId<T>, DAOId<T>>,
+            owner: StorageObjectOwner<MemberId<T>, ChannelId<T>, DAOId<T>>,
             content: Vec<ContentParameters<ContentId<T>, DataObjectTypeId<T>>>
         ) {
-            ensure_root(origin)?;
+            if let StorageObjectOwner::Member(member_id) = owner {
```
bedeho: I think this will be needed in many places, right? so we may as well factor it out into a `ensure_is_storage_object_owner(origin, owner)`

bedeho (changes requested):
Looks good, just a minor fix.

runtime-modules/storage/src/data_directory.rs
iorveth: Good point, it can be reused among storage uploading and further storage deletion extrinsics.  https://github.com/Joystream/joystream/pull/2134/commits/84ce5b9d251b730709d04f999f7d7444c1c2c18f

bedeho (approved):
Excellent.

Issue #2145, opened 2021-02-06
Cleanup extrinsic for setting leader invite quota


bedeho:
1. There is a needless validity check on membership returned from `T::WorkingGroup::get_leader_member_id`
1. There is a second matching again in the mutation section.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703952879841/1201704788362262) by [Unito](https://www.unito.io)

Issue #2138, opened 2021-02-04
Increase max category depth of forum by one


bedeho:
SSIA.

Issue #2108, opened 2021-01-29
Separate event for inviting member


bedeho:
Currently `invite_member` is just using same event, but a distinct one is needed to reflect this is a different effect, and it should have full information required to compute side-effect outside system.

In same spirit, lets rename `MemberRegistered` to `MembershipBought` to really clarify the different origins.

PR #2151, opened 2021-02-08
Minor fixes

conectado:
# Fixing #2108 #2138 #2145

* Separate events for membership invite
* Increase max category depth for forum by one
* Remove redundant check in setting leader invite quota


shamil-gadelshin (approved):
LGTM

PR #2157, opened 2021-02-09
Storage deletion hook

iorveth:
SSIA


mnaamani:
Does this supersede #2156 ?

iorveth:
@mnaamani Yep, it basically depends on https://github.com/Joystream/joystream/pull/2156. @bedeho asked me to make them separate, though we can close https://github.com/Joystream/joystream/pull/2156, if it would be easier for you to review both of them simultaneously.

runtime-modules/common/src/storage.rs
```diff
@@ -41,4 +41,19 @@ pub trait StorageSystem<T: crate::StorageOwnership + crate::MembershipTypes> {
         owner: StorageObjectOwner<T::MemberId, T::ChannelId, T::DAOId>,
         content_parameters: Vec<ContentParameters<T::ContentId, T::DataObjectTypeId>>,
     ) -> DispatchResult;
+
+    // Should hook into call on storage system,
+    // but requires rich error (with reasons)  types.
+    // caller already knows the `ContentId`s as they are part of
+    // the ContentUploadParameters
+    fn atomically_remove_content(
```
mnaamani, line 49: We are not currently keeping track of content ids in the content directory so when we call into this method we can't provide an `&[T::ContentId]` argument. I had left some [notes in the content directory PR](https://github.com/Joystream/joystream/pull/2158/files#diff-530b6b5839302012b1dd920e330e8aed807c41206038f85b19bdf263ace7f349R811)

It would be useful to have a signature like this:

```rust
/// Removes all content ids owned by storage object owner
fn atomically_remove_content_owned_by(
    owner: &StorageObjectOwner<T::MemberId, T::ChannelId, T::DAOId>,
) -> DispatchResult;
```
However this implies the `data_directory` needs to iterate over the `DataObjectById` map.. Is this a problem? I think even @bedeho suggested that there is no more `IterableStorageMap` trait for storage maps in the runtime.
If that is the case then I will need to update the content directory to take this into account.

runtime-modules/storage/src/data_directory.rs
```diff
@@ -358,6 +368,30 @@ decl_module! {
...
+        /// Remove the content from the system.
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn remove_content(
+            origin,
+            owner: StorageObjectOwner<MemberId<T>, ChannelId<T>, DAOId<T>>,
+            content_ids: Vec<ContentId<T>>
+        ) {
+
+            // Ensure given origin can perform operation under specific storage object owner
+            Self::ensure_storage_object_owner_origin(origin, &owner)?;
```
mnaamani, line 380: I'm not seeing an implementation of this function `ensure_storage_object_owner_origin()`. I assume this going to come from a trait to be implemented in the runtime module that can validate origin vs owner ?

iorveth: It`s here. https://github.com/Joystream/joystream/blob/sumer/runtime-modules/storage/src/data_directory.rs#L463

runtime-modules/common/src/storage.rs
iorveth: No, i did not look at sketch, it's my mistake. We should use `IterableStorageMap` for `DataObjectById`  representation, isn`t it available for storage maps in runtime anymore  https://substrate.dev/rustdocs/v2.0.1/frame_support/storage/trait.IterableStorageMap.html ?

mnaamani: My fault really, I removed it from the sketch by mistake at one point.

Okay, but I still think it is dangerous to iterate over a potentially huge map, so I will have to opt to track the content ids in the content directory and use the call as implemented here.

iorveth: > My fault really, I removed it from the sketch by mistake at one point.
> 
> Okay, but I still think it is dangerous to iterate over a potentially huge map, so I will have to opt to track the content ids in the content directory and use the call as implemented here.

It seems, that we already restricted the number of object per owner & their total size with `Quota`, is it dangerous still?

mnaamani: Yes I think so, because there is no restriction on number of channels or members, so the map will still be quite large despite the quota limits.

iorveth: Hm.. But, why can't we switch `DataObjectById` mapping representation to the double map (k1, k2) - > value, where `storage_owner` is k1, `object_id` -> k2?

Though afterwards we also need to introduce `ContentId -> SrorageOwner` relation to be compatible with `reject_content` & `accept_content` extrinsics

mnaamani: had a good discussion with @shamil-gadelshin and @conectado about this. Perhaps you can give us a brief summary of your recommendation.

mnaamani: Yes that is what was recommended :)

iorveth: Cool, i`ll create an issue shortly.

shamil-gadelshin: Summary for our discussion:
- sometimes DoubleMap could be helpful
- on_initialize and on_finalize should be avoided. If it is not possible, we should choose on_initialize because it can be benchmarked.
- if the storage iteration is unavoidable - bounded iteration could limit possible problems
- sometimes iterations could be avoided using additional extrinsic as Parity did with validator payments.

mnaamani: I think it will be simpler for now to have content directory tack content ids, to ensure they are removed when a channel is deleted. If we find a real need for this feature we can add it in future.

One important point raised however by @conectado is that we should probably constrain the number of content ids that can be deleted per transaction. Since we have a quota on the number of objects that can be created by an owner, that in itself will be sufficient I think.

mnaamani (approved):
LGTM

runtime-modules/common/src/storage.rs
mnaamani: Thank you @shamil-gadelshin for the valuable input :+1:

mnaamani: In the end I realized tracking the content ids was getting a bit out of hand, and actually not-necessary. While a channel exists, the channel owner is incentivised to [delete unused assets](https://github.com/Joystream/joystream/pull/2158/files#diff-530b6b5839302012b1dd920e330e8aed807c41206038f85b19bdf263ace7f349R835) because of quota limit. Its really only the edge case of when a channel is deleted that I wanted to have a way to delete all assets by owner. 

I implemented a [helper extrinsic](https://github.com/Joystream/joystream/pull/2158/files#diff-530b6b5839302012b1dd920e330e8aed807c41206038f85b19bdf263ace7f349R862) which can be used by content lead (I should probably also allow curators) to call to delete left-over content after a channel is deleted. This because a non-issue as bedeho pointed out if we choose not to support deleting of channels.

Ultimately the only cost of not cleaning up these objects is wasted capacity on storage providers and potentially having them sync uselessly sync some dataobjects. Eventually storage nodes coul depend on a query-node to determine what content needs to be synced and can intelligently ignore some data objects.

Issue #2079, opened 2021-01-25
Content directory: Channels


mnaamani:
Expansion from https://github.com/Joystream/joystream/issues/2074

Implement Channels feature described in https://github.com/Joystream/joystream/issues/2070

mnaamani:
Implemented in https://github.com/Joystream/joystream/pull/2158

PR #2158, opened 2021-02-10
Implement content dir Channels

mnaamani:
Implement extrinsics for content directory channel management:

- [x] Create Channels
- [x] Update Channels
- [x] Delete Channels
- [x] Censore/Uncensor channels
- [x] Create/Update/Delete Channel Categories
- [x] Channel owner can delete storage
- [x] Lead can delete storage of deleted channels - this feature will depend on lead being able fetch content ids of assets associated with deleted channels.
- [x] updated @joystream/types
- [x] Small bug fix in curator group management. Ensure id being added to a group is in-fact a working-group worker id.

Minimal unit-tests. Will add more in separate PR.

Just some notes:

There are no limits on number of channels an actor can create, and it doesn't cost anything either. Should we introduce some limits.. otherwise the quota limits in storage are not enough to limit total storage that can be taken up by the content directory.

I haven't explicitly implemented support for DAO actor yet.

closes https://github.com/Joystream/joystream/issues/2079


runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+    owner: &ChannelOwner<T::MemberId, T::CuratorGroupId, T::DAOId>,
+) -> DispatchResult {
+    // Only owner of a channel can update and delete it.
+    // Should we allow lead to also update and delete curator group owned channels,
+    // to avoid need for them to add themselves into the group?
+    match actor {
+        ContentActor::Lead => {
+            let sender = ensure_signed(origin)?;
+            ensure_lead_auth_success::<T>(&sender)?;
+            Ok(())
```
iorveth: There is no need to return `Ok(()) ` for each branch, make `ensure_lead_auth_success` return `DispatchResult` instead of `Result<(), Error<T>>` and you can just leave  `ensure_lead_auth_success::<T>(&sender)` at the end of this branch without any additional conversion.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+            // Authorize curator, performing all checks to ensure curator can act
+            CuratorGroup::<T>::perform_curator_in_group_auth(
+                curator_id,
+                curator_group_id,
+                &sender,
+            )?;
+
+            // Curators cannot censor curator group channels
+            match owner {
+                ChannelOwner::CuratorGroup(_curator_group_id) => {
```
iorveth: It is much clearer to use ` if let` statement, when you only would like to match against one case.

```
// Curators cannot censor curator group channels
            if let ChannelOwner::CuratorGroup(_) = owner {
                 Err(Error::<T>::CannotCensoreCuratorGroupOwnedChannels.into())
            }
```

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+
+            Ok(())
+        },
+        ContentActor::Member(_member_id) => {
+            // Members cannot censore channels!
+            ensure!(
+                false,
+                Error::<T>::ActorNotAuthorized
+            );
+            Ok(())
```
iorveth: just return an error here
 `Err(Error::<T>::ActorNotAuthorized.into())`

iorveth: if we do not plan to use `_member_id` somewhere in the future, just leave this parameter blank `ContentActor::Member(_)`, please

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+    origin: T::Origin,
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+    owner: &ChannelOwner<T::MemberId, T::CuratorGroupId, T::DAOId>,
+) -> DispatchResult {
+    // Only owner of a channel can update and delete it.
+    // Should we allow lead to also update and delete curator group owned channels,
+    // to avoid need for them to add themselves into the group?
+    match actor {
+        ContentActor::Lead => {
+            ensure!(
```
iorveth: Just return an error, please `Err(Error::<T>::ActorNotAuthorized.into())`

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+                &sender,
+            )?;
+
+            // Ensure curator group is the channel owner.
+            ensure!(
+                *owner == ChannelOwner::CuratorGroup(*curator_group_id),
+                Error::<T>::ActorNotAuthorized
+            );
+
+            Ok(())
```
iorveth, line 164: There is no need to return` Ok(())` at the end of each branch, just make it one at the end of the function.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
 }
 
+pub fn ensure_actor_authorized_to_create_channel<T: Trait>(
+    origin: T::Origin,
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+) -> DispatchResult {
+    match actor {
+        // Lead should use their member or curator role to create or update channels.
+        ContentActor::Lead => {
+            ensure!(
```
iorveth: `Just return an error, please Err(Error::<T>::ActorCannotOwnChannel.into())
`

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+            let sender = ensure_signed(origin)?;
+
+            // Authorize curator, performing all checks to ensure curator can act
+            CuratorGroup::<T>::perform_curator_in_group_auth(
+                curator_id,
+                curator_group_id,
+                &sender,
+            )?;
+
+            Ok(())
```
iorveth: There is no need to return` Ok(())`, rework `CuratorGroup::<T>::perform_curator_in_group_auth` and all related functions to return  `DispatchResult` instead of `Result<(), Error<T>>`

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+        }
+        ContentActor::Member(member_id) => {
+            let sender = ensure_signed(origin)?;
+
+            ensure_member_auth_success::<T>(member_id, &sender)?;
```
iorveth: Please, rework ` ensure_member_auth_success::<T>(member_id, &sender)?; `and all related functions to return DispatchResult instead of Result<(), Error<T>>

use 
`ensure_member_auth_success::<T>(member_id, &sender)`
instead of 
```
ensure_member_auth_success::<T>(member_id, &sender)?;

            Ok(()
```

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+                curator_id,
+                curator_group_id,
+                &sender,
+            )?;
+
+            Ok(())
+        },
+        ContentActor::Member(_member_id) => {
+            // Members cannot censore channels!
+            ensure!(
```
iorveth: just return an error here

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+) -> DispatchResult {
+    // Only owner of a channel can update and delete it.
+    // Should we allow lead to also update and delete curator group owned channels,
+    // to avoid need for them to add themselves into the group?
+    match actor {
+        ContentActor::Lead => {
+            let sender = ensure_signed(origin)?;
+            ensure_lead_auth_success::<T>(&sender)?;
+            Ok(())
```
iorveth: There is no need to return Ok(()) for each branch, make ensure_lead_auth_success return DispatchResult instead of Result<(), Error<T>> and you can just leave ensure_lead_auth_success::<T>(&sender) at the end of this branch without any additional conversion.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+        },
+        ContentActor::Curator(curator_group_id, curator_id) => {
+            let sender = ensure_signed(origin)?;
+
+            // Authorize curator, performing all checks to ensure curator can act
+            CuratorGroup::<T>::perform_curator_in_group_auth(
```
iorveth, line 236: Same here

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -87,24 +99,217 @@ pub fn ensure_is_lead<T: Trait>(origin: T::Origin) -> DispatchResult {
...
+        // ContentActor::Dao(_daoId) => ...,
+    }
+}
+
+pub fn ensure_actor_authorized_to_delete_stale_assets<T: Trait>(
+    origin: T::Origin,
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+) -> DispatchResult {
+    // Only Lead and (sudo) can delete assets no longer associated with a channel or person.
+    match actor {
```
iorveth:
```rust
if let ContentActor::Lead =  actor {
            let sender = ensure_signed(origin)?;
            ensure_lead_auth_success::<T>(&sender)
        } else  {
            ensure_root(origin)?;
            Ok(())
        }
    }
```

runtime-modules/content/src/lib.rs
```diff
@@ -635,198 +658,484 @@ decl_module! {
...
+
+            // Ensure channel does not exist
+            ensure!(
+                !ChannelById::<T>::contains_key(channel_id),
+                Error::<T>::ChannelMustNotExist
+            );
+
+            let object_owner = StorageObjectOwner::<T>::Channel(channel_id);
+
+            T::StorageSystem::atomically_remove_content(&object_owner, &assets)?;
```
iorveth, line 881: Missing mutation safe marker and event dispatch?

runtime-modules/content/src/lib.rs
```diff
@@ -635,198 +658,484 @@ decl_module! {
...
+            // check that channel exists
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+
+            ensure_actor_authorized_to_update_or_delete_channel::<T>(
+                origin,
+                &actor,
+                &channel.owner,
+            )?;
+
+            let object_owner = StorageObjectOwner::<T>::Channel(channel_id);
```
iorveth, line 840: Missing mutation safe marker and event dispatch?

runtime-modules/content/src/lib.rs
```diff
@@ -635,198 +658,484 @@ decl_module! {
...
+        ) {
+            // check that channel exists
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+
+            ensure_actor_authorized_to_update_or_delete_channel::<T>(
+                origin,
+                &actor,
+                &channel.owner,
+            )?;
+
```
iorveth, line 791: Mutation safe marker?

runtime-modules/content/src/lib.rs
```diff
@@ -635,198 +658,484 @@ decl_module! {
...
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // TODO: Burn funds from sender account as fee for creating a channel
+
+            // add assets to storage
+            // This should not fail because of prior can_add_content() check!
+            T::StorageSystem::atomically_add_content(
```
iorveth, line 691: Don't see the purpose of making the same check twice in a row, to be honest. Wouldn't it be better to remove the additional check inside of a `T::StorageSystem::atomically_add_content` method implementation?

iorveth (changes requested):
Overall looks good, just leaved a couple of minor suggestions.

runtime-modules/content/src/permissions/mod.rs
mnaamani: Thanks for the tip here, I did initially try by directly returning the Error type but struggled with the type inference. I see now I was missing the `.into()` :+1: will update

runtime-modules/content/src/lib.rs
mnaamani: yes I see in this extrinsic the can_add_content() check is really redundant because there is nothing between it and the call to add_content(). It would have made sense when I had some additional checks there.

But no definitely keep the check inside atomically_add_content() it is safer. I'll remove the can_add_content check in this dispatch instead.

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/permissions/mod.rs
mnaamani: done in d64896858ac14ab2ff3b5cc65bdbdd0413779577

runtime-modules/content/src/lib.rs
mnaamani: done in a40b58fd87d2c645e95ffebc69c81c8501a09bb2

runtime-modules/content/src/lib.rs
mnaamani: done in a40b58fd87d2c645e95ffebc69c81c8501a09bb2

runtime-modules/content/src/lib.rs
mnaamani: done in a40b58fd87d2c645e95ffebc69c81c8501a09bb2

runtime-modules/content/src/lib.rs
mnaamani: done in a40b58fd87d2c645e95ffebc69c81c8501a09bb2

mnaamani:
Addressed all your points @iorveth 
Also updated the augment types

mnaamani:
Fyi changes to type names in events in 6c4822dc1d1897e37ab42ba9ad24e45701f081d2
Are just so we don't need to introduce new type names unnecessarily to the `@joystream/types` library.

iorveth (approved):
LGTM!

mnaamani:
Just pushed a last minor change to the permissions to:
- allow Lead to update and delete curator group owned channels
- allow Lead to censor a curator group owned channel

Final thought on deleting channels.

Its not clear that we really want to allow channels to be deleted, if the owner doesn't want a their channel to be visible they can choose to make it non-public and if they wish they can delete assets. But for now we need to allow deletion of channels if we must allow curator groups to be deleted (since curator groups can own a channel and lead cannot delete a group if it owns one).

---
I can't see a good reason why a lead would need to delete curator groups. So if we drop this requirement, then we can also drop deleting of channels.

bedeho (approved):
Overall this looks good and correct.

I want to warn that all of the deletion stuff, both for channels and channel assets, is not really safe. The number of assets and objects could be huge here, and even if we put a limit, it would not be cost effective weight, because of so many database reads/writes, so this approach will need to be changed, probably by inlining channel objects, not in separate maps. Same goes for storage assets, they will be managed differently in the future.

PR #2105, opened 2021-01-28
Query node membership mappings

metmirr:

Issue #2033, opened 2021-01-16
Updated media asset location representation


bedeho:
For assets living on native Joystream infrastructure, we must capture the fact that

- when first being uploaded, there is no asset to display for the content until at least the liason confirms the upload was successful, something which may itself be perpetually delayed if the upload first fails.
- when editing the asset, the old asset should keep getting used until the new asset has been confirmed by the liason. Notice that this resets on the assumption that the old asset will not be deleted in the interim, which is true, as deletion is not even possible in the current storage system, and the future it will be impossible if the asset is use. Notice that an edit can interrupt another in flight edit, for example if the first edit failed to complete.

The current representation looks like this

```graphql
  "Id of the data object in the Joystream runtime dataDirectory module"
  dataObjectId: String! @unique

  happenedIn: Block!
```

and it needs to be replaced by something in this spirit

```graphql

enum UploadStage {
  IncompleteFirstUpload
  IncompleteLaterUpload
  ConfirmedUploaded
}

...

   "Stage of an upload."
   upload_stage: UploadStage!

  "Id of last initiated upload of data object."
  lastDataObjectAttemptedUploaded: String!
  
   "Id of last initiated upload of data object."
  lastDataObjectConfirmedUploaded: String

  "Block where last upload attempt was initiated on-chain."
  lastUploadInitiatedInBlock: Block!
  
```

This is just a sketch of something capturing the intended functionality, it is not confirmed to work well in the broader content directory data model refactoring for `Sumer`.

bedeho:
Or with variants somethign like this

```Rust 
enum StorageAssetStatus {
  PendingFirstUpload(DataObjectId),
  CompletedUpload(DataObjectId),
  WaitingForReplacementUpload(/* old */ DataObjectId, /* new */ DataObjectId )
}
```

bedeho:
I am working on schemas for Pioneer that also need some sort of representation of assets being uploaded, so I gave it a crack, ended up with something like this

```graphql


# Below, the term upload life-cycle basically refers to
# semantics of the following enum from the chain
# pub enum LiaisonJudgement {
#    /// Content awaits for a judgment.
#    Pending,
#
#   /// Content accepted.
#    Accepted,
#
#    /// Content rejected.
#    Rejected,
# }

# Asset is missing
type NeverProvided @variant {
}

# Asset was deleted
type Deleted @variant {
}

# Status of an upload
type UploadStatus @variant {

  "Data object in upload life-cycle.
[...]

PR #2159, opened 2021-02-10
Query node input schema

ondratra:
Input schema for query node reflecting the state of the new content directory.


bedeho:
Am I supposed to review this? seems a bit premature, the same problems are still here? @mnaamani ?

mnaamani:
> Am I supposed to review this? seems a bit premature, the same problems are still here? @mnaamani ?

Not yet, It hasn't been updated yet following my discussion with Andy. I just tagged you in it to stay updated with changes.

mnaamani:
@ondratra I would avoid making any changes to code in the runtime in this PR (even though as far as I can see they are just code comments). It would be better to collect any changes specific for runtime separately.

ondratra:
Only input schema related changes are left in this PR. Other changes previously included in this PR were removed and will be part of other PRs that will come in the future.

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
   "Short text chosen by member to share information about themselves"
-  about: String
-
-  "Blocknumber when member was registered"
-  registeredAtBlock: Int!
+  about: String!
 
   "Member's controller account id"
-  controllerAccount: Bytes!
+  controllerAccount: String!
```
mnaamani, line 41: Why the change from Bytes to String? (I assume Bytes would have represented the 32Byte raw account id and the string will be the encoded address?)

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
   "MemberId: runtime identifier for a user"
   id: ID!
 
   "The unique handle chosen by member"
-  handle: String @unique @fulltext(query: "membersByHandle")
+  handle: String! @unique @fulltext(query: "membersByHandle")
 
   "A Url to member's Avatar image"
-  avatarUri: String
+  avatarUri: String!
```
mnaamani: avatar should be nullable because the runtime allows that.

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
 
   "Short text chosen by member to share information about themselves"
-  about: String
-
-  "Blocknumber when member was registered"
-  registeredAtBlock: Int!
+  about: String!
```
mnaamani: The about field should also be nullable because the runtime allows it. Can we still have an empty string if we use `String!` ?

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
+  entry: MembershipEntryMethod!
+
+  "The type of subscription the member has purchased if any."
+  subscription: BigInt
+}
+
+# TODO: add description to fields
+# TODO: revise property requirement for the new and existing fields derived from content-metadata-protobuf
+
+type Account @entity {
```
mnaamani: This type isn't used anywhere. What is it for?

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
+
+  firstName: String
+
+  middleName: String
+
+  lastName: String
+
+  about: String
+
+  coverPhotoUrl: String
```
mnaamani: This is an asset which may be a url or content in the storage system identified by its `ContentId`.
So both the name of the field and the type is not correct.

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
+
+  avatarPhotoUrl: String
```
mnaamani: Same as `coverPhoto`

query-node/schema.graphql
```diff
@@ -12,32 +12,127 @@ type Block @entity {
...
+
+  people: [Person]
+
+  seasons: [Season]
+}
+
+type Asset @entity {
+  id: ID!
+
+  url: String!
```
mnaamani: An asset needs to represent the possible types, either a Url or content in the storage system. And in the latter case the upload status. see: https://github.com/Joystream/joystream/issues/2033

query-node/schema.graphql
```diff
@@ -98,9 +195,6 @@ type Channel @entity {
   "Flag signaling whether a channel is public."
   isPublic: Boolean!
 
-  "Flag signaling whether a channel is curated/verified."
-  isCurated: Boolean!
```
mnaamani, line 102: replaced with `is_censored`

query-node/schema.graphql
```diff
@@ -83,6 +178,8 @@ type Channel @entity {
   # or a member. We are not handling events related to curator group so we will not set this field
   # owner: Member!
 
+  category: ChannelCategory
+
```
mnaamani, line 212: handle is renamed to `title` and is not unique

query-node/schema.graphql
```diff
@@ -203,6 +297,24 @@ type HttpMediaLocationEntity @entity {
   happenedIn: Block!
 }
 
+type MediaType @entity {
```
mnaamani: I don't think we need to populate the query node with media types or languages

query-node/schema.graphql
```diff
@@ -226,6 +338,8 @@ type VideoMedia @entity {
 
   locationEntity: MediaLocationEntity
 
+  publishedBeforeJoystream: PublishedBeforeJoystream
+
```
mnaamani: I think the concept of two separate types `VideoMedia` and `Video` was relic of old content directory design. There is only one concept of a `Video` in the content directory. Is there reason to have these two types still?

mnaamani (changes requested):
I'm not super familiar with the syntax here but there seems to be lots of gaps and in some cases some misunderstanding of some of the key ideas that need to be captured in this schema model. Specifically around assets and storage.

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: I copied the membership from https://github.com/Joystream/joystream/pull/2105 where this change was present. I guess the point was to have it in some readable format for easy conversion to JSON, but I will ask Metin to double-check.

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: I forgot to remove this one. Initially, I thought an entity for an account would be handy there, but since it has no properties, it is redundant. Done.

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: Asset upgraded. Check if it's ok, please.

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: I removed the media type. Should the language be completely removed as well? Has it been removed from metadata protobuf recently? Just double-checking.

query-node/schema.graphql
ondratra: Great, that simplifies the video structure in my eyes. Please check that the new structure is ok (and that assets are used correctly).

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: Done. The exclamation mark means non-nullable - it doesn't prevent empty strings. I'll discuss it with Atlas team if it is desirable to ensure non-empty strings. If that's the case, we should probably be consistent about it everywhere a `String` is used.

query-node/schema.graphql
ondratra: The String is really used just for convenient work with the data in sandboxes, while building apps, etc.

query-node/schema.graphql
ondratra: Confirmed from Atlas team that non-nullable string is the desired state and empty strings are ok there.

bedeho (commented):
Overall this looks good and correct.

I want to warn that all of the deletion stuff, both for channels and channel assets, is not really safe. The number of assets and objects could be huge here, and even if we put a limit, it would not be cost effective weight, because of so many database reads/writes, so this approach will need to be changed, probably by inlining channel objects, not in separate maps. Same goes for storage assets, they will be managed differently in the future.

mnaamani:
@bedeho 
> I want to warn that all of the deletion stuff, both for channels and channel assets, is not really safe.

Are you referring to just the query node or in runtime as well. I'm inclined to remove support for deleting channels in runtime after some discussion with @bwhm https://github.com/Joystream/joystream/issues/2165#issuecomment-782098356 https://github.com/Joystream/joystream/pull/2158#issuecomment-781960074

However deleting assets just still be supported otherwise when a channel reaches its quota limit in storage and owner wants to remove some old videos lets say to make room for new content they will not be able to.

metmirr:
I have two suggestions:

1. I suggest we use `String` for accounts instead of `Bytes` which is Hex string in the graphql api. I can tell that accounts are human readable on all substrate explorers. Example: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY' is will look like `0x307864343335393363373135666464333163363131343161626430346139396664363832326338353538383534636364653339613536383465376135366461323764`
2. If I am not mistaken Atlas team requested to send `DateTime` for `publishedBeforeJoystream` field and Hydra support `DateTime` scalar type. Substrate's timestamp(`Moment`) is compatible with javascript. So mapping author can create `Date` objects without any problem.

bedeho:
> Overall this looks good and correct.
> 
> I want to warn that all of the deletion stuff, both for channels and channel assets, is not really safe. The number of assets and objects could be huge here, and even if we put a limit, it would not be cost effective weight, because of so many database reads/writes, so this approach will need to be changed, probably by inlining channel objects, not in separate maps. Same goes for storage assets, they will be managed differently in the future.

I have no clue how this happened, probably me getting confused, but this reply was meant for this PR: https://github.com/Joystream/joystream/pull/2158

Terribly sorry @mnaamani @ondratra .

query-node/schema.graphql
```diff
@@ -12,30 +12,142 @@ type Block @entity {
...
+# TODO: add description to fields
+# TODO: revise property requirement for the new and existing fields derived from content-metadata-protobuf
+
+type ChannelCategory @entity {
+  id: ID!
+
+  name: String
+}
+
+type Person @entity {
```
mnaamani: Lets not include Person, Series, Season.. anything that will not be part of Sumer release

query-node/schema.graphql
```diff
@@ -12,30 +12,142 @@ type Block @entity {
...
+  description: String
+
+  coverPhoto: String
+
+  people: [Person!]
+
+  seasons: [Season!]
+}
+
+union Asset = AssetUrl | AssetStorage
```
mnaamani: An Asset may also not be provided at all so it can be null

query-node/schema.graphql
```diff
@@ -12,30 +12,142 @@ type Block @entity {
...
+
+type AssetUrl @variant {
+  id: ID!
+
+  "The http url pointing to the media"
+  url: String!
+
+  "The port to use when connecting to the http url (defaults to 80)"
+  port: Int
```
mnaamani: I don't see a need to separate url and port, just use a single string as will be provided by the metadata

query-node/schema.graphql
```diff
@@ -12,30 +12,142 @@ type Block @entity {
...
+}
+
+enum AssetStorageUploadStatus {
+  INCOMPLETE_FIRST_UPLOAD
+  INCOMPLETE_LATER_UPLOAD
+  CONFIRMED_UPLOADED
```
mnaamani: The discussion #2033 had other recommendations that were an improvement on this initial design.

query-node/schema.graphql
```diff
@@ -83,23 +195,25 @@ type Channel @entity {
...
   "The title of the Channel"
-  handle: String! @fulltext(query: "search")
+  title: String! @fulltext(query: "search")
 
   "The description of a Channel"
   description: String!
 
   "Url for Channel's cover (background) photo. Recommended ratio: 16:9."
-  coverPhotoUrl: String
+  coverPhotoUrl: Asset
```
mnaamani: Either the `Asset` should be `union Asset = AssetUrl | AssetStorage | Null` or this filed should be nullable.
the Url should be removed from the field name because it may not be a url.

mnaamani: All the fields that come from protobuf metadata should be nullable because they are optional

query-node/schema.graphql
```diff
@@ -269,37 +322,40 @@ type Video @entity {
...
+  "Flag signaling whether a video is censored."
+  isCensored: Boolean!
 
   "Whether the Video contains explicit material."
   isExplicit: Boolean!
 
   license: LicenseEntity!
 
+  "Encoding of the video media object"
+  encoding: VideoMediaEncoding!
```
mnaamani: VideoMediaEncoding is not correct, its the old content directory schema entity. In new design it comes from the protobuf metadata. And I don't think we will have 'pre-populated' encoding entities in the query-node database. Or should we?

query-node/schema.graphql
ondratra: Person, Series, Season, and Playlist were removed.

query-node/schema.graphql
ondratra: Asset is nullable already. It can't be seen in the asset definition itself, but everywhere it is used, it is nullable (`myProperty: Asset` - without exclamation mark that marks non-nullable properties).

query-node/schema.graphql
ondratra: Port removed.

query-node/schema.graphql
ondratra: As stated above: Asset is nullable already. It can't be seen in the asset definition itself, but everywhere it is used, it is nullable (`myProperty: Asset` - without exclamation mark that marks non-nullable properties).

ondratra: I revised all content coming from protobuf and made it nullable in a couple of places where it was previously non-nullable.

query-node/schema.graphql
ondratra: Language stays and they will be identified via ISO_639-1 string.

query-node/schema.graphql
ondratra: VideoMediaEncoding entity has been removed and relevant properties from protobuf have been integrated into Video entity.

ondratra:
The changes requested in the last review have been reflected in the input schema.

Other changes that have been discussed outside of this GitHub issue:
- `PublishedBeforeJoystream` integrated into Video
- language representation as ISO string
- timestamps converted to `DateTime`
- storage asset upload status improved

query-node/schema.graphql
```diff
@@ -8,36 +13,149 @@ type Block @entity {
...
+
+  "Storage provider as liaison judgment"
+  liaisonJudgement: LiaisonJudgement!
+
+  "IPFS content id"
+  ipfsContentId: String!
+}
+
+"Owner type for storage object"
+enum AssetOwner {
```
mnaamani: Can enum variants enclose values? Because variant `Member` is not enough to capture which member id is the owner for example

query-node/schema.graphql
```diff
@@ -83,23 +201,25 @@ type Channel @entity {
...
-  description: String!
+  description: String
 
   "Url for Channel's cover (background) photo. Recommended ratio: 16:9."
-  coverPhotoUrl: String
+  coverPhotoUrl: Asset
 
   "Channel's avatar photo."
-  avatarPhotoUrl: String
+  avatarPhotoUrl: Asset
```
mnaamani: the field name should not have `Url` in it

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
   channel: Channel!
 
   "Reference to a video category"
-  category: Category!
+  category: Category
```
mnaamani: rename to `VideoCategory`

`ChannelCategory` is very simple:
```
"Category of media channel"
type ChannelCategory @entity {
  id: ID!

  name: String
}
```

and the `Category` you are using here as the VideoCategory has name and description, but profobuf message only has `name`. Can you make both types more consistent.

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
...
   "The description of the Video"
-  description: String!
+  description: String
 
   "Video duration in seconds"
-  duration: Int!
+  duration: Int
 
   "Video's skippable intro duration in seconds"
   skippableIntroDuration: Int
```
mnaamani: I think `skippableIntroDuration` is missing from protobuf definition. Is this required in Atlas?

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
...
 
   "Video thumbnail url (recommended ratio: 16:9)"
-  thumbnailUrl: String!
+  thumbnailUrl: String
```
mnaamani: This is an Asset not a string url

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
...
 
   "Video's main langauge"
   language: Language
 
-  "Reference to VideoMedia"
-  media: VideoMedia!
```
mnaamani, line 261: We need an Asset in its place for the actual video

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
...
-  "Video curation status set by the Curator"
-  isCurated: Boolean!
+  "Flag signaling whether a video is censored."
+  isCensored: Boolean!
 
   "Whether the Video contains explicit material."
-  isExplicit: Boolean!
+  isExplicit: Boolean
+
+  license: LicenseEntity
```
mnaamani: revise the LicenseEntity type it doesn't match with our design anymore: https://github.com/Joystream/joystream/blob/sumer/content-metadata-protobuf/doc/index.md#license

mnaamani (changes requested):
Left some more comments

query-node/schema.graphql
kdembler: No, it's not being used currently. I think it was intended as a future feature

query-node/schema.graphql
```diff
@@ -237,45 +281,60 @@ type Video @entity {
...
+
+  "Encoding of the video media object"
+  codecName: String
```
kdembler: Any specific reason to drop the VideoMedia and add the fields directly here? I personally liked that separation of video metadata and video file metadata

query-node/schema.graphql
```diff
@@ -83,23 +201,25 @@ type Channel @entity {
...
-  handle: String! @fulltext(query: "search")
+  title: String @fulltext(query: "search")
 
   "The description of a Channel"
-  description: String!
+  description: String
 
   "Url for Channel's cover (background) photo. Recommended ratio: 16:9."
-  coverPhotoUrl: String
+  coverPhotoUrl: Asset
```
kdembler: We will need to use one common naming for the assets, `Url` suffix will not work anymore. I guess it could be just `coverPhoto`?

query-node/schema.graphql
```diff
@@ -8,36 +13,149 @@ type Block @entity {
...
+  size: BigInt!
+
+  "Storage provider id of the liaison"
+  liaisonId: BigInt!
+
+  "Storage provider as liaison judgment"
+  liaisonJudgement: LiaisonJudgement!
+
+  "IPFS content id"
+  ipfsContentId: String!
```
kdembler, line 152: Just to make sure, this is the old `dataObjectId`?

query-node/schema.graphql
mnaamani: Yes, I made the same point on another field also (thumbnailUrl -> thumbnail)

query-node/schema.graphql
ondratra: `Url` suffix removed

query-node/schema.graphql
ondratra: Types structure unified (`ChannelCategory` now looks like `VideoCategory`).

query-node/schema.graphql
ondratra: The field has been removed.

query-node/schema.graphql
ondratra: `Url` suffix removed

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: `License` has been reworked according to metadata.

query-node/schema.graphql
ondratra: Done

query-node/schema.graphql
ondratra: I assume so since it's only `id` in `storage`'s `DataObjectInternal`.

query-node/schema.graphql
ondratra: I think they can't. Rework `AssetOwner` into a union of variants that can hold relevant ids.

query-node/schema.graphql
ondratra: We've discussed this with Mokhtar on the call, and if I understood it correctly, the reason was simply not to have 1:1 relations anymore, but maybe I'm mistaken. What do you think @mnaamani?

mnaamani: My bad, I had asked Andy to drop the separate entities because I assumed this model was chosen to match explicitly the separate entities in the old content directory. In new content directory there isn't this explicit separation so we can revert if it makes more sense. - sorry @ondratra

query-node/schema.graphql
mnaamani: IPFS content id is the IPFS hash of the content. That is different than the data object id (also referred to as ContentId)

mnaamani: Maybe rename it to avoid this confusion ?

bedeho:
I only looked representation of `DataObject`, but I don't think this is quite right yet? It has some of the elements from this discussions 

https://github.com/Joystream/joystream/issues/2033

but it also has lots of stuff which is redundant? Perhaps I have not kept up with some discussion on this?

query-node/schema.graphql
kdembler: If this isn't the `dataObjectId`, then isn't it missing from `AssetDataObject`?

bedeho:
> I only looked representation of `DataObject`, but I don't think this is quite right yet? It has some of the elements from this discussions
> 
> #2033
> 
> but it also has lots of stuff which is redundant? Perhaps I have not kept up with some discussion on this?

Disregard this comment, was plain wrong.

query-node/schema.graphql
kdembler: From Atlas side, it doesn't make that much of a difference, I think it's just better DX separated. If we don't care, let's keep the old version

query-node/schema.graphql
ondratra: The discussed id was really missing. It is now called `joystreamContentId` to distinguish it from IPFS id.

query-node/schema.graphql
ondratra: `VideoMedia` was reintroduced, but under the name `VideoMetadata`.

kdembler (approved):
LGTM, as far as I can tell seems to suit all the Atlas needs :+1:

query-node/schema.graphql
kdembler: Just a thought: maybe we should call this `VideoMediaMetadata` or something else? Generally, I'd expect the title for example to be `VideoMetadata` as well. I think the current name may not properly indicate that we're talking about the media

mnaamani (changes requested):
We are almost there so final points:

- There are some old types that can be removed:
```
ClassEntity
NextEntityId
KnownLicenseEntity
UserDefinedLicenseEntity
```
- The Membership type is still using `Bytes` for account and I think we decided a String would be best.
- The Channel type is missing the optional rewardAccount field


There seems to be perhaps some typeo somewhere in the file because syntax highlighting appears to be broken when viewing the whole file.

ondratra:
Old entities have been removed. Accounts are represented as strings now, and the channel now has a reward account. Multiple categories for videos and channels are now supported (reflects changes in https://github.com/Joystream/joystream/pull/2214).

Syntax highlighting in GitHub seems to be broken for graphql. I've tried to make some changes in a testing branch, but the block quotes highlight doesn't work properly and brokes the rest of the file.

query-node/schema.graphql
kdembler: @ondratra ping

ondratra: Sorry, I've missed the last reply. The rename does make sense since other video-related properties can also be considered metadata, and this change better distinguishes them. The name has been changed, and the entity is called `VideoMediaMetadata` now.

query-node/schema.graphql
ondratra: The changes were made accordingly to the mentioned discussion.

mnaamani:
> Old entities have been removed. Accounts are represented as strings now, and the channel now has a reward account. Multiple categories for videos and channels are now supported (reflects changes in #2214).

Well I did say "If we merge" :)  .. so please revert changes for multiple categories.

Otherwise this looks good to merge now.

ondratra:
Multiple categories reverted.

mnaamani (approved):

PR #2160, opened 2021-02-10
Withdraw funds from bounty.

shamil-gadelshin:
### Changes
- added `withdraw_member_funding` extrinsic
- added `withdraw_creator_funding` extrinsic
- introduced bounty accounts.
- introduced bounty removal in some cases

### Implementation comments
- when the cherry can't be distributed fully among the funders - the remainder gets burned
- bounty state could be removed during either `withdraw_member_funding` or `withdraw_creator_funding`extrinsics. The actual removal occurs when all possible withdrawals are completed.

### Open questions
- Should we migrate a council budget to the account to eliminate slashes/deposit_creations? 
- How do we prevent bad bounty creation (eg.: funding_period = 0, creator_funding = 0)? Should we limit active bounties number for a member or we rely on the extrinsic fee? 

### DIFF
--- a/runtime-modules/bounty/src/benchmarking.rs
+++ b/runtime-modules/bounty/src/benchmarking.rs
@@ -231,7 +231,7 @@ benchmarks! {
     }: _ (RawOrigin::Signed(account_id.clone()), member_id, bounty_id, amount)
     verify {
         assert_eq!(Balances::<T>::usable_balance(&account_id), T::Balance::max_value() - amount);
-        assert_last_event::<T>(Event::<T>::MaxFundingReached(bounty_id).into());
+        assert_last_event::<T>(Event::<T>::BountyMaxFundingReached(bounty_id).into());
     }
 }
 
--- a/runtime-modules/bounty/src/benchmarking.rs
+++ b/runtime-modules/bounty/src/benchmarking.rs
@@ -149,6 +149,7 @@ benchmarks! {
 
     cancel_bounty_by_council {
         let cherry: BalanceOf<T> = 100.into();
+        let max_amount: BalanceOf<T> = 1000.into();
 
         T::CouncilBudgetManager::set_budget(cherry);
 
@@ -158,6 +159,7 @@ benchmarks! {
             judging_period: One::one(),
             cherry,
             creator: creator.clone(),
+            max_amount,
             ..Default::default()
         };
 
@@ -176,6 +178,7 @@ benchmarks! {
 
     cancel_bounty_by_member {
         let cherry: BalanceOf<T> = 100.into();
+        let max_amount: BalanceOf<T> = 1000.into();
         let (account_id, member_id) = member_funded_account::<T>("member1", 0);
 
         T::CouncilBudgetManager::set_budget(cherry);
@@ -187,6 +190,7 @@ benchmarks! {
             judging_period: One::one(),
             cherry,
             creator: creator.clone(),
+            max_amount,
             ..Default::default()
         };
 
@@ -208,9 +212,12 @@ benchmarks! {
     }
 
     veto_bounty {
+        let max_amount: BalanceOf<T> = 1000.into();
+
         let params = BountyCreationParameters::<T>{
             work_period: One::one(),
             judging_period: One::one(),
+            max_amount,
             ..Default::default()
         };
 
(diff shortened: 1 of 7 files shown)


bedeho:
> Should we migrate a council budget to the account to eliminate slashes/deposit_creations?

I did not understand this question, could you restate?

> How do we prevent bad bounty creation (eg.: funding_period = 0, creator_funding = 0)? Should we limit active bounties number for a member or we rely on the extrinsic fee?

If we have no caps on active bounties overall, then I see no need for such a limit.

runtime-modules/bounty/src/lib.rs
```diff
@@ -194,19 +217,35 @@ pub enum BountyStage {
...
-    MaxFundingReached(BlockNumber),
+    /// Creator funds (initial funding and/or cherry) were withdrawn.
+    CreatorFundsWithdrawn,
+
+    /// A bounty funding was successful and it exceeded max funding amount.
+    BountyMaxFundingReached {
+        ///  A bounty funding was successful on the provided block.
+        max_funding_reached_at: BlockNumber,
+        /// A bounty reached its maximum allowed contribution on creation.
+        reached_on_creation: bool,
```
bedeho, line 239: Why is `reached_on_creation` needed? I see it only read here

`cherry_needs_withdrawal: reached_on_creation && cherry_is_not_zero,`

but it's also not clear this field is needed, or why it depends on anything except `cherry_is_not_zero`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -194,19 +217,35 @@ pub enum BountyStage {
...
+    },
 
     /// A bounty was canceled.
     Canceled,
 
-    /// A bounty funding was successful on the provided block.
-    /// The stage is set when the funding exceeded max funding amount.
-    MaxFundingReached(BlockNumber),
+    /// Creator funds (initial funding and/or cherry) were withdrawn.
+    CreatorFundsWithdrawn,
```
bedeho, line 232: Why is there a distinct extrinsic and thus milestone for withdrawing funding? All there is is bounty cancellation? 

See design doc
> The creator can cancel at any time before this, something that is of particular importance of in perpetual funding periods. If no one has contributed, any cherry is returned and the bounty is instantl removed, but if at least one contributor has funded, then the cherry will be split amont all contributors, and the bounty goes to the withdraw period.

So cancelling a bounty covers possible recovery of bounty contributed, if relevant. Right now you have two separate extrinsics and milestones. I am not actually sure you even need two `Canceled` and `CreatorFundsWithdrawn` even if we wanted two separate extrinsics, as the runtime should do the exact same thing going forward in both cases?

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -172,66 +190,181 @@ benchmarks! {
...
+            ..Default::default()
+        };
+
+        Bounty::<T>::create_bounty(RawOrigin::Root.into(), params, Vec::new()).unwrap();
+
+        let bounty_id: T::BountyId = Bounty::<T>::bounty_count().into();
+
+        assert!(Bounties::<T>::contains_key(bounty_id));
+
+        Bounty::<T>::cancel_bounty(RawOrigin::Root.into(), creator.clone(), bounty_id).unwrap();
```
conectado, line 322: Can we add an assertion after this that the stage is `BountyStage::Withdrawal` with `cherry_needs_withdrawal == true`? (The same for  the other benchmark for `withdraw_creator_funding`.)

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -172,66 +190,181 @@ benchmarks! {
...
     verify {
-        let bounty = <crate::Bounties<T>>::get(bounty_id);
+        let bounty = Bounty::<T>::bounties(bounty_id);
 
         assert!(matches!(bounty.milestone, BountyMilestone::Canceled));
         assert_last_event::<T>(Event::<T>::BountyVetoed(bounty_id).into());
     }
 
     fund_bounty {
+        let max_amount: BalanceOf<T> = 100.into();
```
conectado, line 238: Shouldn't we have a benchmark for each branch of `get_bounty_milestone_on_funding`?

As I understand currently we are only benchmarking the case `maximum_funding_reached` is true. Am I right?

conectado (changes requested):
Overall benchmarking looks great, just left one minor comment and one important question.

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: Ok

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: This is the worst-case scenario because it fires an additional event. Other branches are similar.

conectado (approved):
LGTM

PR #2167, opened 2021-02-11
Query node: indexer, processor workflow update

metmirr:
This PR removes code generation for `indexer` and `hydra-processor`. There is no need to generate indexer and processor code since docker image for indexer is compatible with joystream chain and hydra-processor is a standalone cli tool anymore.

P.S I didn't commit yarn.lock file because of a weird error `An unexpected error occurred: "could not find a copy of ts-node to link in /Users/metin/dev/joystream/node_modules/api-scripts/node_modules/@joystream/types/node_modules"`. As a workaround, I install packages directly to the root workspace i.e `yarn add -D -W @dzlzv/hydra-typegen`.


mnaamani:
> P.S I didn't commit yarn.lock file because of a weird error `An unexpected error occurred: "could not find a copy of ts-node to link in /Users/metin/dev/joystream/node_modules/api-scripts/node_modules/@joystream/types/node_modules"`. As a workaround, I install packages directly to the root workspace i.e `yarn add -D -W @dzlzv/hydra-typegen`.

This is most likely due to so many different versions of `ts-node` used in various packages and yarn gave up trying to find a version that works for all. I can try to fix it later.

mnaamani:
Tried the following inside `query-node/` dir:

```bash
yarn clean
yarn codegen:server
yarn # to install any packages needed by generated/graphql-server package
# warning Resolution field "@polkadot/api@1.26.1" is incompatible with requested version "@polkadot/api@^2.10.1"
yarn db:up
yarn db:migrate
```
but it is failing with:
```
yarn run v1.22.4
$ ./db-migrate.sh
$ (cd ./generated/graphql-server && yarn db:create && yarn db:sync && yarn db:migrate)
$ warthog db:create
password authentication failed for user "postgres"
$ SYNC=true WARTHOG_DB_SYNCHRONIZE=true ts-node --type-check src/index.ts
error: password authentication failed for user "postgres"
```

Will try to do some more testing..

should I be using `docker:db:migrate` script instead of `db:migrate` ? If there are any more scripts which are no longer correct maybe we can remove them.

query-node/manifest.yml
```diff
@@ -0,0 +1,55 @@
+version: '0.1'
+dataSource:
+  kind: substrate 
+  chain: joystream
+  indexerVersion: '0.1.6'
+metadata:
```
mnaamani, line 10: This manifest file looks like it was made for live network? does it need to be changed to work for local development testing?

query-node/mappings/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "lib/mappings/index.js",
+  "license": "MIT",
+  "scripts": {
+    "build": "rm -rf lib && tsc --build tsconfig.json",
+    "lint": "echo \"Skippinng\"",
+    "clean": "rm -rf lib"
+  },
+  "dependencies": {
+    "@dzlzv/hydra-common": "0.0.3",
+    "@polkadot/types": "~2.10.2-7"
```
mnaamani: Has this version of @polkadot/types been tested against the babylon runtime?
The monorepo yarn workspace will anyway override this to `v1.26.1`. Will that be a problem ?

mnaamani:
Can't get this to work. Also we might need a similar or re-target this work to master branch, as I'm having issues using latest hydra-cli to update only graphql-server for the live network.

ondratra (commented):
The changed code itself looks good and is similar to what I changed when working on the query node environment with the new hydra. My only tip for improvement right now is to add new lines to the end files where they are missing `query-node/mappings/tsconfig.json`, `query-node/mappings/package.json`, `query-node/manifest.yml`.

The branch can't be compiled now because there are old mappings (and maybe other things) that are not compatible with the new environment. I guess the full-scale review can be done when we combine all the new hydra, query node, and mapping-related pieces together. I will build mappings on top of this branch, so maybe I resolve some issues along the way.

metmirr:
@mnaamani @ondratra I update the workflow for hydra-cli v2.

Steps to try it out:

```
yarn build
yarn db:prepare
yarn db:migrate

# Build mappings
yarn mappings:build

# Run processor, it should fail due to the missing types
yarn processor:start
```

I expect to hydra processor to fail and that should be fixed once we have types generated by hydra typegen. And also I add mappings functions with an empty body to make sure that everything is set as expected. After you guys confirm then I can make another commit to remove them.

query-node/mappings/package.json
metmirr: New mappings will depend on this version because of the typegen stuff but not sure how we can manage that.

query-node/manifest.yml
metmirr: The most important part is the metadata section which has `source` and `blockHash`, for local testing one should update them.

ondratra:
When running `yarn build` a lot of missing bash variable errors occurred to me. I had to solve it by creating a new file that imports env variables and run `yarn build` afterward to overcome that. Then run this file instead of plain `yarn build`.

`query-node/tmp.sh`:
```
set -a
. ../.env
set +a

yarn build
```

metmirr:
> When running `yarn build` a lot of missing bash variable errors occurred to me. I had to solve it by creating a new file that imports env variables and run `yarn build` afterward to overcome that. Then run this file instead of plain `yarn build`.
> 
> `query-node/tmp.sh`:
> 
> ```
> set -a
> . ../.env
> set +a
> 
> yarn build
> ```

ohh I forgot to update `package.json`'s build script, it should be `"build": "./build.sh"`

query-node/mappings/package.json
mnaamani: We will have to test but I'm concerned the newer version will break the other apps. In that case we may need to move query-node out of the top level yarn workspace.

query-node/build.sh
```diff
@@ -11,19 +11,7 @@ set +a
...
-
-# Build indexer customizing DB name
-DB_NAME=${INDEXER_DB_NAME} yarn codegen:indexer
-
-# Build graphql-server customizing DB name
-DB_NAME=${PROCESSOR_DB_NAME} yarn codegen:server
-
-# We run yarn again to ensure processor and indexer dependencies are installed
-# and are inline with root workspace resolutions
-yarn
```
mnaamani, line 25: We still need another `yarn` step after codegen to properly install the dependencies after the generated/graphql-server pacakge is generated.

mnaamani: perhaps:
```
yarn codegen:noinstall
yarn
```
There is also some problem coming from installing dependencies in particular the monorepo's husky configuration is being overridden by the husky configuration of generated/graphql-server. Is there some option to skip running the husky bootstrapping somehow.

metmirr: `generated/graphql-server/package.json` has this   

```JSON
"husky": {
    "hooks": {
      "pre-commit": "yarn run config:dev && lint-staged && tsc -p ./tsconfig.json && yarn test"
    }
  },
```

is it this what you want to get rid of? if yes, it is possible and would require a new hydra-cli release.

query-node/package.json
```diff
@@ -1,42 +1,54 @@
...
-		"debug": "^4.2.0",
-		"dotenvi": "^0.9.1",
-		"tslib": "^2.0.0"
-	},
-	"volta": {
+  "name": "query-node-root",
+  "private": "true",
+  "version": "0.0.0",
+  "description": "GraphQL server and mappings. Generated with \u2665 by Hydra-CLI",
+  "workspaces": [
```
mnaamani: Does this "workspaces" key belong here? We already have it in top package.json (except for mappings which needs to be added)

I don't think yarn 1.2x properly supports nested workspaces.

With this config, the dependencies are being installed under the query-node/node_modules rather than the root. Not sure if this is a good or bad thing. But I suspect if mappings package start trying to use other packages in the root workspace (joystream/types and the metadata protobuf package) they will not be found.

query-node/build.sh
mnaamani: Well that is what is interfering yes, but I don't want to get rid of it per say, I'm sure it is useful for projects that are checking in the graphql-server into their repo. I think maybe if we fix this https://github.com/Joystream/joystream/pull/2167#discussion_r586852488 it might no longer be an issue?

query-node/package.json
metmirr: Done! (Just as information hydra-cli v2 generate `package.json` with the workspace above.)

mnaamani:
Just needs a merge conflict fix now

mnaamani:
Thanks for updating PR metin.
There appears to be some checks failing because of missing dependency:
```
error https://github.com/metmirr/warthog/releases/download/v2.23.0/warthog-v2.23.0.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, open '/home/runner/.cache/yarn/v6/.tmp/a37e5733c888f0d4d92bd22435676c2c/dist/core/server.test.js'"
```

I'm re-running some of the jobs to see if it resolves, I double checked that version of warthog is available now.


Lastly the issue I mentioned before with polkadot types version issue we see this in yarn logs:

```
warning Resolution field "@polkadot/types@1.26.1" is incompatible with requested version "@polkadot/types@~2.10.2-7"
warning Resolution field "@polkadot/api@1.26.1" is incompatible with requested version "@polkadot/api@^2.10.1"
```

So we will have to see what impact that has later.

mnaamani:
The problem with downloading the warthog package seems to be only on the MacOS ci checks.. :(

query-node/manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+dataSource:
+  kind: substrate 
+  chain: joystream
+  indexerVersion: '0.1.6'
+entities:
+  - mappings/lib/generated/**/*.model.js
+metadata:
+  # TODO: update source and blockHash for sumer release
+  source: wss://babylon-sub-1.joystream.app/staging/rpc
+  blockHash: '0x206a6cca31333cb19ee120cde7455e13eea368ec3f6176d0cd48698228410489'
```
dzhelezov, line 13: I don't think we should really need to specify `blockHash` here. By default, it just takes the latest block which should be fine

query-node/manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+  - members.buyMembership
+  - members.addScreenedMember
+outDir: ./mappings/generated/types
+customTypes:
+  lib: '@joystream/types/augment/all/types'
+  typedefsLoc: '../types/augment/all/defs.json'
+mappings:
+  hydraCommonVersion: '0.0.3'
+    # process only blocks with height >= 1M 
+  blockInterval: '[1000000,]'
```
dzhelezov, line 31: `blockInterval` is not needed here. In this example it will process blocks with height > 1M, which definitely not what we want.

query-node/typegen.yml
```diff
@@ -0,0 +1,22 @@
+metadata:
```
dzhelezov, line 1: if this info is duplicated in `manifest.yml`, a separate `typegen.yml` file is not needed

Issue #2171, opened 2021-02-11
Init data_rectory on runtime upgrade


mnaamani:
Similar to https://github.com/Joystream/joystream/pull/2170 the new storage values for the data_directory:

```
Quotas
QuotaSizeLimitUpperBound
QuotaObjectsLimitUpperBound
GlobalQuota
UploadingBlocked
```
need to be initialized on runtime upgrade.

The `DataObjectByContentId` map also needs to be migrated or all its elements deleted since the `struct DataObject` has been augmented (`owner` field has been changed from a `MemberId` to a `StorageObjectOwner`).

`DataObjectByContentId`  ->  `T::ContentId => Option<DataObject<T>>`

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2170

PR #2173, opened 2021-02-12
Init data_directory on runtime upgrade

iorveth:
Closes https://github.com/Joystream/joystream/issues/2171

### DIFF
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -527,8 +527,8 @@ export default class Api {
     return exists ? await this._api.query.contentDirectory.entityById<Entity>(id) : null
   }
 
-  async DataByContentId(contentId: ContentId): Promise<DataObject | null> {
-    const dataObject = await this._api.query.dataDirectory.DataByContentId<Option<DataObject>>(contentId)
+  async dataByContentId(contentId: ContentId): Promise<DataObject | null> {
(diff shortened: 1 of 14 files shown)


cli/src/Api.ts
```diff
@@ -527,8 +527,8 @@ export default class Api {
     return exists ? await this._api.query.contentDirectory.entityById<Entity>(id) : null
   }
 
-  async dataObjectByContentId(contentId: ContentId): Promise<DataObject | null> {
```
mnaamani, line 530: I think you in-inadvertently auto-replaced occurrences in client code. The queries will start with a lower case.
Perhaps revert these and leave it for another PR.

mnaamani (changes requested):
Implementation looks good.
Just small issue, with name change in client packages.

cli/src/Api.ts
iorveth: Thank you for pointing that. I hope it wouldn`t be hard to make them lower case again

iorveth: Done https://github.com/Joystream/joystream/pull/2173/commits/7f1cd93229fb8188bc4055d5579ea9b2c37c7fcc

mnaamani (approved):
The `types/augment-*` files need to be generated separately but that can be fixed later

Issue #2147, opened 2021-02-08
Replicate extrinsic parameters in all events


bedeho:
In all modules. Makes life easier for query node mapping authors.

PR #2174, opened 2021-02-15
Add extrinsic parameters to all events

conectado:
# Tracking #2147 

For all existing events emitted in all extrinsics in all pallets this PR ensures that every parameter for the extrinsic is present on the event.

### DIFF
--- a/runtime-modules/content-directory/src/lib.rs
+++ b/runtime-modules/content-directory/src/lib.rs
@@ -1590,6 +1590,8 @@ decl_module! {
             });
 
             // Trigger event
+            // Note: The first `nonce + T::Nonce::one()` is the new value for the nonce while
+            // the second `nonce` is the input paramer for the extrinsic
             Self::deposit_event(
                 RawEvent::RemovedAtVectorIndex(
                     actor, entity_id, in_class_schema_property_id, index_in_property_vector,
--- a/runtime-modules/content-directory/src/lib.rs
+++ b/runtime-modules/content-directory/src/lib.rs
@@ -534,7 +534,7 @@ decl_module! {
                 <EntityCreationVouchers<T>>::insert(class_id, controller.clone(), entity_creation_voucher.clone());
 
                 // Trigger event
-                Self::deposit_event(RawEvent::EntityCreationVoucherUpdated(controller, entity_creation_voucher))
+                Self::deposit_event(RawEvent::EntityCreationVoucherUpdated(controller, entity_creation_voucher, class_id))
             } else {
                 // Create new EntityCreationVoucher instance with provided maximum_entities_count
                 let entity_creation_voucher = EntityCreationVoucher::new(maximum_entities_count);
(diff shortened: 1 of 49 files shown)


runtime-modules/content-directory/src/lib.rs
```diff
@@ -534,7 +534,7 @@ decl_module! {
                 <EntityCreationVouchers<T>>::insert(class_id, controller.clone(), entity_creation_voucher.clone());
 
                 // Trigger event
-                Self::deposit_event(RawEvent::EntityCreationVoucherUpdated(controller, entity_creation_voucher))
+                Self::deposit_event(RawEvent::EntityCreationVoucherUpdated(controller, entity_creation_voucher, class_id))
```
shamil-gadelshin, line 537: Should we add `maximum_entities_count` here?

runtime-modules/content-directory/src/lib.rs
```diff
@@ -1543,7 +1593,7 @@ decl_module! {
             Self::deposit_event(
                 RawEvent::RemovedAtVectorIndex(
                     actor, entity_id, in_class_schema_property_id, index_in_property_vector,
-                    nonce + T::Nonce::one(), involved_entity_and_side_effect
+                    nonce + T::Nonce::one(), involved_entity_and_side_effect, nonce,
```
shamil-gadelshin, line 1598: Why do we have two nonces here?

runtime-modules/forum/src/lib.rs
```diff
@@ -415,54 +415,56 @@ decl_event!(
...
+        PostAdded(PostId, ForumUserId, CategoryId, ThreadId, Vec<u8>),
 
         /// Post with givne id was moderated.
-        PostModerated(PostId, Vec<u8>),
+        PostModerated(PostId, Vec<u8>, PrivilegedActor, CategoryId, ThreadId),
 
         /// Post with given id had its text updated.
         /// The second argument reflects the number of total edits when the text update occurs.
-        PostTextUpdated(PostId),
+        PostTextUpdated(PostId, ForumUserId, CategoryId, ThreadId, PostId, Vec<u8>),
```
shamil-gadelshin: Why do we have the second PostId? Also, please fix the comment.

shamil-gadelshin (changes requested):
Titanic work! Left a couple of comments and it looks like these events are missing:
#### service_provider
- set_default_lifetime
- set_bootstrap_endpoints
#### data_directory
- remove_known_content_id
- inject_data_objects

runtime-modules/content-directory/src/lib.rs
conectado: I didn't do it because it's already in `entity_creation_voucher`

runtime-modules/content-directory/src/lib.rs
conectado: I thought that keeping the nonce that was signaling before was a good idea and adding the nonce that was actually the input parameter.

runtime-modules/forum/src/lib.rs
conectado: It was a slip

runtime-modules/content-directory/src/lib.rs
shamil-gadelshin: Makes sense, please describe it in the comment.

runtime-modules/content-directory/src/lib.rs
shamil-gadelshin: Ok

shamil-gadelshin (approved):
LGTM

Issue #2086, opened 2021-01-25
Add raw storage field to working group roles


iorveth:
Add raw storage field to working group roles that they can update as they please.

iorveth:
cc @bedeho, should i introduce a length constraint for the raw storage field?

bedeho:
Yes, sorry for the late reply. Only upper bound is strictly needed, but its fine either way.

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2180

PR #2180, opened 2021-02-16
Working groups raw storage field

iorveth:
Closes https://github.com/Joystream/joystream/issues/2086

### DIFF
--- a/runtime-modules/service-discovery/src/mock.rs
+++ b/runtime-modules/service-discovery/src/mock.rs
@@ -167,6 +167,7 @@ pub(crate) fn hire_storage_provider() -> (u64, u64) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
+        storage: Vec::new(),
     };
 
     <working_group::WorkerById<Test, StorageWorkingGroupInstance>>::insert(
--- a/runtime-modules/service-discovery/src/mock.rs
+++ b/runtime-modules/service-discovery/src/mock.rs
@@ -167,7 +167,6 @@ pub(crate) fn hire_storage_provider() -> (u64, u64) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
-        storage: Vec::new(),
     };
 
     <working_group::WorkerById<Test, StorageWorkingGroupInstance>>::insert(
--- a/runtime-modules/storage/src/tests/mock.rs
+++ b/runtime-modules/storage/src/tests/mock.rs
@@ -63,6 +63,7 @@ impl SetLeadFixture {
             role_account_id: DEFAULT_LEADER_ACCOUNT_ID,
             reward_relationship: None,
             role_stake_profile: None,
+            storage: Vec::new(),
         };
 
         // Create the worker.
@@ -401,6 +402,7 @@ pub(crate) fn hire_storage_provider() -> (u64, u32) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
+        storage: Vec::new(),
     };
 
     <working_group::WorkerById<Test, StorageWorkingGroupInstance>>::insert(
--- a/runtime-modules/storage/src/tests/mock.rs
+++ b/runtime-modules/storage/src/tests/mock.rs
@@ -63,7 +63,6 @@ impl SetLeadFixture {
             role_account_id: DEFAULT_LEADER_ACCOUNT_ID,
             reward_relationship: None,
             role_stake_profile: None,
-            storage: Vec::new(),
         };
 
         // Create the worker.
@@ -402,7 +401,6 @@ pub(crate) fn hire_storage_provider() -> (u64, u32) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
-        storage: Vec::new(),
     };
 
     <working_group::WorkerById<Test, StorageWorkingGroupInstance>>::insert(
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -85,10 +85,7 @@ decl_error! {
         WorkerExitRationaleTextTooShort,
 
         /// Worker storage text is too long.
-        WorkerStorageTextTooLong,
-
-        /// Worker storage text is too short.
-        WorkerStorageTextTooShort,
+        WorkerStorageValueTooLong,
 
         /// Signer is not worker role account.
         SignerIsNotWorkerRoleAccount,
--- a/runtime-modules/working-group/src/errors.rs
+++ b/runtime-modules/working-group/src/errors.rs
@@ -84,6 +84,12 @@ decl_error! {
         /// Worker exit rationale text is too short.
         WorkerExitRationaleTextTooShort,
 
+        /// Worker storage text is too long.
+        WorkerStorageTextTooLong,
+
+        /// Worker storage text is too short.
+        WorkerStorageTextTooShort,
+
         /// Signer is not worker role account.
         SignerIsNotWorkerRoleAccount,
 
--- a/runtime-modules/working-group/src/lib.rs
+++ b/runtime-modules/working-group/src/lib.rs
@@ -318,6 +318,10 @@ decl_storage! {
         pub WorkerById get(fn worker_by_id) : map hasher(blake2_128_concat)
             WorkerId<T> => WorkerOf<T>;
 
+        /// Maps identifier to corresponding worker storage.
+        pub WorkerStorage get(fn worker_storage): map hasher(blake2_128_concat)
+            WorkerId<T> => Vec<u8>;
+
         /// Count of active workers.
         pub ActiveWorkerCount get(fn active_worker_count): u32;
 
@@ -404,7 +408,7 @@ decl_module! {
         ) {
 
             // Ensure there is a signer which matches role account of worker corresponding to provided id.
-            let mut worker = Self::ensure_worker_signed(origin, &worker_id)?;
+            Self::ensure_worker_signed(origin, &worker_id)?;
 
             Self::ensure_worker_role_storage_text_is_valid(&storage)?;
 
@@ -413,8 +417,7 @@ decl_module! {
             //
 
             // Complete the role storage update
-            worker.storage = storage.clone();
-            WorkerById::<T, I>::insert(worker_id, worker);
+            WorkerStorage::<T, I>::insert(worker_id, storage.clone());
 
             // Trigger event
             Self::deposit_event(RawEvent::WorkerStorageUpdated(worker_id, storage));
--- a/runtime-modules/working-group/src/lib.rs
+++ b/runtime-modules/working-group/src/lib.rs
@@ -327,6 +327,9 @@ decl_storage! {
         /// Worker exit rationale text length limits.
         pub WorkerExitRationaleText get(fn worker_exit_rationale_text) : InputValidationLengthConstraint;
 
+        /// Worker storage text length limits.
+        pub WorkerStorageText get(fn worker_storage_text) : InputValidationLengthConstraint;
+
         /// Map member id by hiring application id.
         /// Required by StakingEventsHandler callback call to refund the balance on unstaking.
         pub MemberIdByHiringApplicationId get(fn member_id_by_hiring_application_id):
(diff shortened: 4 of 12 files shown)


runtime-modules/working-group/src/lib.rs
```diff
@@ -384,6 +390,34 @@ decl_module! {
...
+            origin,
+            worker_id: WorkerId<T>,
+            storage: Vec<u8>
+        ) {
+            // Ensuring worker actually exists
+            let worker = Self::ensure_worker_exists(&worker_id)?;
+
+            // Ensure that origin is signed by member with given id.
+            ensure_on_wrapped_error!(
+                membership::Module::<T>::ensure_member_controller_account_signed(origin, &worker.member_id)
```
mnaamani: I think for this extrinsic it makes more sense to authenticate origin is signed by the `worker.role_account_id`

mnaamani:
This PR could have been made independent of the storage PRs since de632a2 doesn't touch the storage pallet

runtime-modules/working-group/src/lib.rs
mnaamani: with:
```rust
// Ensure there is a signer which matches role account of worker corresponding to provided id.
let worker = Self::ensure_worker_signed(origin, &worker_id)?;
```

mnaamani (changes requested):
Minor change request. And if work on working group pallet can be cherry-picked and based of sumer, it will be better so we can merge this without waiting on other storage pallet work.

runtime-modules/working-group/src/lib.rs
iorveth: Done here https://github.com/Joystream/joystream/pull/2180/commits/7385ebfc75b1f10b59861054ae84d4c6a430cca1

iorveth:
Just introduced a length constraint for the raw storage field https://github.com/Joystream/joystream/pull/2180/commits/f93b965c5dbd310351fd4c219c36717ff5b19c6a

runtime-modules/working-group/src/lib.rs
```diff
@@ -1488,6 +1534,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
         opening_human_readable_text_constraint: InputValidationLengthConstraint,
         worker_application_human_readable_text_constraint: InputValidationLengthConstraint,
         worker_exit_rationale_text_constraint: InputValidationLengthConstraint,
+        worker_storage_text_constraint: InputValidationLengthConstraint,
```
mnaamani: I think this "raw" storage field should be versatile and allow potentially having arbitrary binary data in it, not just plain text. The worker should also be able to set it to an empty value. For this reason I think we should allow the minimum length to be zero. Does `InputValidationLengthConstraint` and validation allow for min length to be 0?

Also `worker_storage_text_constraint` can be re-named to reflect this to maybe: `worker_storage_size_constraint`

What do you think?

cc: @bedeho

runtime-modules/working-group/src/lib.rs
```diff
@@ -1329,11 +1365,21 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+                Error::<T, I>::WorkerStorageTextTooLong.into(),
+            )
+            .map_err(|e| DispatchError::Other(e))
+    }
 }
 
 /// Creates default text constraint.
 pub fn default_text_constraint() -> InputValidationLengthConstraint {
-    InputValidationLengthConstraint::new(1, 1024)
+    InputValidationLengthConstraint::new(1, 2048)
```
mnaamani: I think we should keep this default, and have a separate "default" for the storage field if we want to allow 0 length.

mnaamani (commented):
I think the storage field should allow 0 size Vec<u8>.

runtime-modules/working-group/src/lib.rs
iorveth: Good idea, https://github.com/Joystream/joystream/pull/2180/commits/9f1d0ce17addd360318b008408e4f5516656d6ec

runtime-modules/working-group/src/lib.rs
iorveth: Nice catch. Yep, `InputValidationLengthConstraint` and validation allows for min length to be 0.

iorveth: `Also worker_storage_text_constraint can be re-named to reflect this to maybe: worker_storage_size_constraint`
Done https://github.com/Joystream/joystream/pull/2180/commits/9f1d0ce17addd360318b008408e4f5516656d6ec

runtime-modules/working-group/src/tests/mod.rs
```diff
@@ -1165,6 +1166,131 @@ fn update_worker_role_account_fails_with_invalid_origin() {
...
+            .call_and_assert(Err(DispatchError::Other("WorkerStorageTextTooLong")));
+    });
+}
+
+#[test]
+fn update_worker_storage_fails_with_too_short_text() {
+    build_test_externalities().execute_with(|| {
+        let worker_id = fill_default_worker_position();
+
+        let storage_field = vec![0u8].repeat(default_text_constraint().min as usize - 1);
```
mnaamani: If we want to always allow 0 size it wouldn't make sense to have this test.

runtime-modules/working-group/src/lib.rs
```diff
@@ -1329,13 +1365,28 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+            .map_err(|e| DispatchError::Other(e))
+    }
 }
 
 /// Creates default text constraint.
 pub fn default_text_constraint() -> InputValidationLengthConstraint {
     InputValidationLengthConstraint::new(1, 1024)
 }
 
+/// Creates default storage size constraint.
```
mnaamani, line 1386: Thinking though this again, setting default to allow 0 min is okay, but it means in future the lead can set different constraint with non-zero min. If that is a policy then okay, but if we want to always allow 0 size
we can drop `InputValidationLengthConstraint` and just have the upper bound configurable.

runtime-modules/working-group/src/tests/mock.rs
```diff
@@ -169,6 +169,10 @@ pub fn build_test_externalities() -> sp_io::TestExternalities {
             WORKING_GROUP_CONSTRAINT_MIN,
             WORKING_GROUP_CONSTRAINT_DIFF,
         ),
+        worker_storage_size_constraint: InputValidationLengthConstraint::new(
+            WORKING_GROUP_CONSTRAINT_MIN,
```
mnaamani, line 173: We probably want a test with this min being 0 (This const is currently 1)

runtime-modules/working-group/src/errors.rs
```diff
@@ -84,6 +84,12 @@ decl_error! {
         /// Worker exit rationale text is too short.
         WorkerExitRationaleTextTooShort,
 
+        /// Worker storage text is too long.
+        WorkerStorageTextTooLong,
```
mnaamani: Maybe change these two error names to:
```
WorkerStorageValueTooLong,
WorkerStorageValueTooShort,
```

runtime-modules/working-group/src/types.rs
```diff
@@ -196,6 +197,9 @@ pub struct Worker<AccountId, RewardRelationshipId, StakeId, BlockNumber, MemberI
 
     /// When set, describes role stake of the worker/lead.
     pub role_stake_profile: Option<RoleStakeProfile<StakeId, BlockNumber>>,
+
+    /// Raw storage field
+    pub storage: Vec<u8>,
```
mnaamani: One issue to address is that with the addition of this field, we need to migrate the existing workers during runtime upgrade.

One approach is to implement a custom `Decode` trait for `struct Worker` that can deal with missing field.
Alternatively we can store this `storage` value outside of this struct in a separate state storage map:

```rust
decl_storage! {
   ...
   WorkerStorage map WorkerId => Vec<u8>;
}
```

mnaamani (changes requested):
Left a few suggestions and requirement to migrate `Worker`s during runtime upgrade

runtime-modules/working-group/src/errors.rs
iorveth: Good, point i also removed an error for the lower bound limit https://github.com/Joystream/joystream/pull/2180/commits/33f53545ed1b9cff1ceb7adc3a0b4a705766ce68

runtime-modules/working-group/src/lib.rs
iorveth: done https://github.com/Joystream/joystream/pull/2180/commits/33f53545ed1b9cff1ceb7adc3a0b4a705766ce68

runtime-modules/working-group/src/tests/mod.rs
iorveth: Fixed it here https://github.com/Joystream/joystream/pull/2180/commits/1340c1f031aadbcdf0ca95a5409bdaff25ae6f04

runtime-modules/working-group/src/types.rs
iorveth: I chose to introduce an additional storage map :) https://github.com/Joystream/joystream/pull/2180/commits/956934ab2be4f0f3aa3fbdaac23813db3159cfd5

runtime-modules/service-discovery/src/mock.rs
```diff
@@ -167,6 +167,7 @@ pub(crate) fn hire_storage_provider() -> (u64, u64) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
+        storage: Vec::new(),
```
mnaamani: I think you forgot to remove this field

runtime-modules/storage/src/tests/mock.rs
```diff
@@ -63,6 +63,7 @@ impl SetLeadFixture {
             role_account_id: DEFAULT_LEADER_ACCOUNT_ID,
             reward_relationship: None,
             role_stake_profile: None,
+            storage: Vec::new(),
```
mnaamani: This field can also now be removed.

runtime-modules/storage/src/tests/mock.rs
```diff
@@ -401,6 +402,7 @@ pub(crate) fn hire_storage_provider() -> (u64, u32) {
         role_account_id,
         reward_relationship: None,
         role_stake_profile: None,
+        storage: Vec::new(),
```
mnaamani: one more to remove..

mnaamani (changes requested):
Final changes look good you just missed removing the 'storage' field in the Worker in some of the test mocks. That is why the build is failing:

```
error[E0560]: struct `pallet_working_group::Worker<u64, _, _, _, u64>` has no field named `storage`
  --> runtime-modules/storage/src/tests/mock.rs:66:13
   |
66 |             storage: Vec::new(),
   |             ^^^^^^^ `pallet_working_group::Worker<_, _, _, _, _>` does not have this field
   |
   = note: available fields are: `member_id`, `role_account_id`, `reward_relationship`, `role_stake_profile`

error[E0560]: struct `pallet_working_group::Worker<{integer}, _, _, _, {integer}>` has no field named `storage`
   --> runtime-modules/storage/src/tests/mock.rs:405:9
    |
405 |         storage: Vec::new(),
    |         ^^^^^^^ `pallet_working_group::Worker<_, _, _, _, _>` does not have this field
    |
    = note: available fields are: `member_id`, `role_account_id`, `reward_relationship`, `role_stake_profile`

error: aborting due to 2 previous errors
```

runtime-modules/storage/src/tests/mock.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2180/commits/dc39181954b99319c9fb98b0ff9cecb3ec475fe5

runtime-modules/storage/src/tests/mock.rs
iorveth: https://github.com/Joystream/joystream/pull/2180/commits/dc39181954b99319c9fb98b0ff9cecb3ec475fe5

runtime-modules/service-discovery/src/mock.rs
iorveth: https://github.com/Joystream/joystream/pull/2180/commits/dc39181954b99319c9fb98b0ff9cecb3ec475fe5

mnaamani (approved):

PR #2182, opened 2021-02-17
CLI: Add missing axios dependency, fix api:setUri connection error

Lezek123:
CLI bug fixes:
- Adding missing `axios` dependency
- Fixing `api:setUri` connection issue (users were unable to change api uri in case connection via the current one was unsuccesful)

### DIFF
--- a/cli/src/commands/media/uploadVideo.ts
+++ b/cli/src/commands/media/uploadVideo.ts
@@ -204,6 +204,7 @@ export default class UploadVideoCommand extends MediaCommandBase {
           'Content-Length': fileSize.toString(),
         },
         maxContentLength: MAX_FILE_SIZE,
+        maxBodyLength: MAX_FILE_SIZE,
       }
       await axios.put(uploadUrl, fileStream, config)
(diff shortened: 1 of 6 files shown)


cli/src/commands/media/uploadVideo.ts
```diff
@@ -204,6 +204,7 @@ export default class UploadVideoCommand extends MediaCommandBase {
           'Content-Length': fileSize.toString(),
         },
         maxContentLength: MAX_FILE_SIZE,
+        maxBodyLength: MAX_FILE_SIZE,
```
mnaamani, line 207: Adding reference to https://github.com/axios/axios/pull/2781
So we really only need to set the `maxBodyLength` here to MAX_FILE_SIZE
`maxContentLength` is the max length of the response which should never be this large after uploading.

mnaamani (approved):
I'm surprised missing axios dependency didn't break the cli. Perhaps it was a dependency of another package?

Tested to be working correctly. I was able to `setUri` successfully.

PR #2187, opened 2021-02-18
Olympia audit fixes.

shamil-gadelshin:
### Fixed issues
- [`Referral cut` was changed to the percentage of the membership fee](https://github.com/Joystream/audits/issues/2)
- [Missing check for the invitation lock.](https://github.com/Joystream/audits/issues/3)
- [Duplicate IDs in the `set_stickied_threads` extrinsic](https://github.com/Joystream/audits/issues/6)
- [Invalid lock reason for voting in the referendum](https://github.com/Joystream/audits/issues/8)
- [Zero existential deposit](https://github.com/Joystream/audits/issues/9)
- Obsolete runtime parameters (TransferFee, CreationFee)
- [Double voting in forum polls](https://github.com/Joystream/audits/issues/10)


bedeho:
Reassigning to @conectado here, I think he will be more effective than me at this point.

runtime-modules/forum/src/benchmarking.rs
```diff
@@ -923,6 +923,15 @@ benchmarks! {
             text.clone(), text.clone(), poll
         );
 
+        // Add poll voting.
+        Module::<T>::vote_on_poll(
```
conectado: Shouldn't we have multiple votes to capture the complexity of ` <PollVotes<T>>::remove_prefix(thread_id);`?

runtime-modules/forum/src/benchmarking.rs
```diff
@@ -985,6 +995,15 @@ benchmarks! {
 
         let moderator_id = ModeratorId::<T>::from(forum_user_id.try_into().unwrap());
 
+        // Add poll voting.
+        Module::<T>::vote_on_poll(
```
conectado: Shouldn't we have multiple votes to capture the complexity of ` <PollVotes<T>>::remove_prefix(thread_id);`?

runtime-modules/forum/src/lib.rs
```diff
@@ -1955,9 +1973,19 @@ impl<T: Trait> Module<T> {
...
 
+        // Helps to prevent thread ID duplicates.
+        let mut unique_stickied_ids = BTreeSet::<T::ThreadId>::new();
+
         // Ensure all thread id valid and is under the category
-        for item in stickied_ids {
-            Self::ensure_thread_exists(&category_id, item)?;
+        for thread_id in stickied_ids {
+            // Check for ID duplicates.
+            if unique_stickied_ids.contains(thread_id) {
```
conectado, line 2165: Do we need to check for duplicates? Can't we just insert them and ignore it?

conectado (changes requested):
I just a couple of small comments. 

Also,  maybe since we have an existencial deposit, we could enable with this PR the benchmarking for pallet_staking and pallet_balances since we were missing that. Otherwise we should create an issue to do that once this is merged.

runtime-modules/forum/src/lib.rs
shamil-gadelshin: It's a fix for https://github.com/Joystream/audits/issues/6

conectado: My bad, I missed that.

runtime-modules/forum/src/benchmarking.rs
shamil-gadelshin: Fixed [here](https://github.com/Joystream/joystream/pull/2187/commits/71e852f22aacd9a8d883974088e33d096438a56d)

runtime-modules/forum/src/benchmarking.rs
shamil-gadelshin: Fixed [here](https://github.com/Joystream/joystream/pull/2187/commits/71e852f22aacd9a8d883974088e33d096438a56d)

shamil-gadelshin:
I included benchmarking for the staking and balances pallets but disabled both for different reasons: 
- staking benchmarks were disabled because of a too long time to complete
- balances benchmarks affect the fee model (because of changed `transfer` extrinsic weight) and they should be run on the reference machine.

conectado (approved):
LGTM

Issue #2076, opened 2021-01-25
Update Colossus for Sumer runtime changes


mnaamani:
Changes in the runtime data_directory require updating several packages related to storage.
This Issue will focus with primarily Colossus (storage-node) and storage-cli (developer utils for the storage node)

Runtime change to deal with:
- Removal of `KnownContentIds` from data_directory state
- Renaming of map `DataObjectByContentId` to `DataByContentId`
- Change of field type owner from MemberId to `StorageObjectOwner`
- ... ?

PR #2195, opened 2021-02-23
Sumer colossus update

mnaamani:
Update colossus in line with latest runtime changes to the data_directory (they were done in multiple PRs https://github.com/Joystream/joystream/pulls?q=is%3Apr+author%3Aiorveth+base%3Asumer+storage)

Implements https://github.com/Joystream/joystream/issues/2076

1. updated `@joystream/types`
2. Minor changes to the runtime
  - Increased the default initial voucher limits for basic upload tests to work
  - Better error return when no storage provider available. So Atlas can better interpret DispatchError (There are a lot more of this type of fixes needed related to membership/authrorization which I will do later https://github.com/Joystream/joystream/issues/2200

3. Some bug fixes and improvements in handling PUT and GET requests to upload and download content:

  - Since HTTP API connections are not authenticated, the node will only `Reject` content uploads if the upload completes, ipfs hash is computed and matches the data object, and the detected MIME type does not pass the filter (video/audio/image).  This was not correctly handled before, and would allow anyone to cause content to be rejected by the liaison but uploading invalid content.
  - The node will abort handling an upload if it detects file types that don't pass the filter early on, but not reject the content (on chain). This means object owner will need to manually delete the data object or liaison judgement will remain in `Pending` state. This is done to avoid wasting a node's bandwidth.

  - The storage node will not serve content unless it was `Accepted` by liaisons.
  - New content will only get 'committed' to IPFS node after it passes checks.
  - Cleanup temporary files after upload is processed.

4. Minor fixes/cleanup of `utils/api-scripts`

5. Updated CI checks
6. Update and fix storage-node unit tests


iorveth (approved):
LGTM!

storage-node/packages/colossus/paths/asset/v0/{id}.js
```diff
@@ -54,24 +69,36 @@ module.exports = function (storage, runtime, ipfsHttpGatewayUrl, anonymous) {
...
+        return
+      }
+
+      // Ensure content_length from request equals size in data object.
+      if (!dataObject.size_in_bytes.eq(filterResult.content_length)) {
+        errorHandler(res, new Error('Content Length does not match expected size of content'), 403)
+        return
+      }
+
+      const sufficientBalance = await runtime.providerHasMinimumBalance(PROCESS_UPLOAD_BALANCE)
```
shamil-gadelshin: `providerHasMinimumBalance` checks for the free balance rather than the usable balance. Free balance doesn't consider locks. Do we want such behavior?

storage-node/packages/colossus/paths/asset/v0/{id}.js
```diff
@@ -21,15 +21,30 @@
...
 const filter = require('@joystream/storage-node-backend/filter')
 const ipfsProxy = require('../../../lib/middleware/ipfs_proxy')
+const assert = require('assert')
 
 function errorHandler(response, err, code) {
   debug(err)
   response.status(err.code || code || 500).send({ message: err.toString() })
 }
 
+const PROCESS_UPLOAD_BALANCE = 3
```
shamil-gadelshin: Do you mind documenting this constant? Also, why do we choose this exact value? FYI: we are implementing existential deposit in the Olympia - it could be higher than this value.

storage-node/packages/cli/src/commands/consts.ts
```diff
@@ -0,0 +1,2 @@
+// Defines maximum content length for the assets (files)
```
shamil-gadelshin: Did you consider placing this constant in the existing base.ts file? Both usage of this constant use base.ts.

shamil-gadelshin (changes requested):
Overall looks good. Left a few questions.

storage-node/packages/colossus/paths/asset/v0/{id}.js
mnaamani: Yes indeed, because node will send out transactions to perform its job. Will update to look at "usable" balance.

storage-node/packages/colossus/paths/asset/v0/{id}.js
mnaamani: This is the expected fees for all the 3 transactions that will need to be sent after processing the upload. Will document.
I don't think existential deposit is actually relevant. In-fact in Olympia we need to figure out how clients applications can compute fees ahead of time for transactions, but that is discussion for another time :)

storage-node/packages/cli/src/commands/consts.ts
mnaamani: Yes that does look like a better place for it :+1:

storage-node/packages/colossus/paths/asset/v0/{id}.js
mnaamani: Looks like our own internal method name is badly chosen but it does in-fact get the spendable/usable balance which is derived from freeBalance - lockedBalance: [ref](https://github.com/polkadot-js/api/blob/23f8eb9713c7edc1118bbab59043238fc84dcb74/packages/api-derive/src/balances/all.ts#L57)

I will change our method name to properly reflect this.

mnaamani: done in 42eb0d191e4d43b03f44f93218eec826c1811788

storage-node/packages/colossus/paths/asset/v0/{id}.js
mnaamani: done in 8d08ccc78a1b6ad919f137d988f4fc9c29228baa

storage-node/packages/cli/src/commands/consts.ts
mnaamani: done in 543f46c4a98eda323091f2a0053ddbc7bc62db00

shamil-gadelshin (approved):
LGTM

PR #2199, opened 2021-02-24
storage-node: fix broken proxying to ipfs gateway

mnaamani:
After merging and updating sumer branch with https://github.com/Joystream/joystream/pull/2182 I observed an issue with http-proxy-middleware failing with `next` is not defined error.

So just back porting fix to master, just in-case although I'm not able re-produce the error on master branch.


mnaamani:
this fix made it into master from sumer

PR #2209, opened 2021-02-26
change parameters, and fix tests

bwhm:
After the new inflow of users, we have discussed making some smaller changes to runtime parameters with the [community](https://testnet.joystream.org/#/proposals/162).

### Proposals
Perhaps the most impactful change is the suggestion to increase the maximum active proposals allowed from 5 to 12. There has been a _lot_ of proposals coming in, and it's frequently a the case that there are no "open" slots.

There is an argument against this - it can "overload" the Council, but as we're adding more proposal types and the community grows, it seems preferable to add more work to paid Councilers than to turn away community members.

The staking requirements for the arguably three most critical proposals, `Runtime Upgrade`, `Set Election Parameters` and`Set Validator Count` is increased 5x. These takes a lot more time for the Council to consider, and in order to actually use the `Runtime Upgrade` proposal as intended, I propose reducing the `grace_period` from 5 to 2 days.

#### Set Election Parameters
This is to allow the community to propose a larger Council.

#### Set Validator Count
As Polkadot is running with 297 max, it seems safe enough to bump this, but as can be gathered from the discussion, there may be better ways to incentivize running nodes.

### Unbonding
Somewhat related to the Validator count, I'm suggesting to increase the unbonding period from ~1 day to ~1 week. We changed this from ~28 days previously, in an attempt to (1) reduce the frequent slashing problem, and (2) let users deploy resources better.

1. It seems upgrading substrate improved the node connectivity, and there are far less issues with nodes going down now. In fact, the unbonding period is now effectively making it close to impossible to get slashed, even if one is trying to (I have on staging networks), and last time I checked, there had been zero slashes on the current network.
2. The uneven distribution may actually make it "better" if users are "forced" in to locking up their tokens for a longer time.


bwhm:
This is just a draft at this point, so mainly looking for input on parameters (any code corrections can come later :D)

@mnaamani @bedeho

bedeho:
# Proposals
1. I agree, lets increase count to 12 or much more even. The reason for the limit is not to limit council load, it is for safety, and this number could be much larger, so I se no reason to cap this at 12. Make it 30 even, should be fine.
2. 5x Stake required: I am not so sure this is needed, and it would discourage participation, but I am fine with it.
3. Set Election Parameters: This is both risky and no reason is provided for why we need this now. I am certainly not convinced w can do this for mainnet, so I would for that reason alone be hesitant to add it.
4. Set Validator Count: I dont know what is being asked here. Is the question whether we can increase the target or cap to some new value? Would be nice to state the new value in that case.

# Unbonding

Ok.

bwhm:
>     1. I agree, lets increase count to 12 or much more even. The reason for the limit is not to limit council load, it is for safety, and this number could be much larger, so I se no reason to cap this at 12. Make it 30 even, should be fine.

My reasoning for doing it somewhat slowly is to avoid the Council getting overwhelmed. I'd be happy to bump this number further later though.
 
>     2. 5x Stake required: I am not so sure this is needed, and it would discourage participation, but I am fine with it.

After initially thinking all proposals, it's only for Runtime Upgrade, Set Election Parameters and Set Validator Count.
I don't think we want too many of these really.

>     3. Set Election Parameters: This is both risky and no reason is provided for why we need this now. I am certainly not convinced w can do this for mainnet, so I would for that reason alone be hesitant to add it.

Not sure we're on the same page here. Specifically, the new limits for what _can_ be proposed to the Council would be:
`COUNCIL_SIZE_MAX` 20->40
`CANDIDACY_LIMIT_MIN` 50->100

This only means a Council of 40 people _can_ be proposed. There will be 12 CMs in the Council after this election (was 6 before).

I'm happy to scrap this one though!

>     4. Set Validator Count: I dont know what is being asked here. Is the question whether we can increase the target or cap to some new value? Would be nice to state the new value in that case.

The suggestions is that a proposal _can_ be made for there to be 300 validators.
Currently, there are (max) 90 validators, and the proposal is capped at 100.

mnaamani:
I'm happy with the suggestion for increasing the limit for how large the validator count proposal can be.

Yes concurrent proposals can definitely be increased, I don't have a magic number really. But 15 looks good, and with grace period lowered it would be a good improvement on what we have now. Better to go up gradually.

Increasing stake for those sensitive proposals is smart. The council should have ability to potentially slash large amounts for someone making proposals that risks compromising stability of the network.

With respect to council size I feel that "lean and mean" is potentially better so keeping the `COUNCIL_SIZE_MAX` at 20 is better I think. No objection to raising candidacy limit.

bwhm:
@mnaamani !

Issue #2196, opened 2021-02-23
Pallets where there is no deposit charged for filling up the storage


conectado:
## Problem

It was found during the audit that any forum user can use the forum's extrinsic `create_thread` to fill up the blockchain storage and exhaust the number of threads of a given category for a small fee since this requires no deposit.

This problem exists in two forms across the runtime: Either we allow with no deposit filling up the storage of a given module with no bound or with bounds.

In the bounded case, this would allow a denial of service attack where other users won't be able to use the storage. 

In the unbounded case, this'd allow to grow the storage maliciously. 

There are different permission levels that'd allow a user to abuse these extrinsics, either a **lead** or a **worker**/**member**. 

When a **lead** can fill up a **bounded** storage is not really a problem because there is no other user that it can be denied of service but I leave those cases in here just to keep them in mind.

## Pallets

- `forum`
  - `create_thread`: **Any forum user** can create a thread, the thread storage for a given category is **bounded**.
  - `add_post`: **Any forum user** can create a post, the post storage for a given thread is **bounded**.
  - `create_category`: **The lead** can create categories with no cost, the category storage is **bounded**. (Same goes for sub-categories)

- `working_group`
  - `add_opening`: **A lead** can use this extrinsic to create openings **unbounded**.
  - `apply_on_opening`: **A working group member** can apply multiple times, filling blockchain storage(when no stakes) **unbounded**.
  - `add_opening` + `apply_on_opening` + `fill_opening`: **A lead** can add an opening with no stakes, apply on it multiple times and fill it, and occupy completely the worker storage.

- `membership`
  - `add_staking_account_candidate`: **A user** could use multiple accounts to fill up the blockchain storage, **unbounded**. 

- `proposals_discussion`
  - `add_post`: **A member** can add posts to any thread, **unbounded**.

- `storage/data_directory`
  - `add_content`: **A member** can use it to fill up the storage **unbounded**.

- `content_directory`
  - `add_curator_group`: **A lead** can add groups to fill up the storage **unbounded**.
  - `add_curator_to_group`: **A lead** can add curators to a group to fill it up **bounded**.

- `service_discovery`:
  - `set_ipns_id`: **Any worker** can use it to fill up storage **unbounded**.

- `data_object_type_registry`
   - `register_data_object_type`: **A lead** can use it to fill up storage **unbounded**.

shamil-gadelshin:
Working group leaders have their stake locked. And it could be slashed any time by a council.

[...]

PR #2210, opened 2021-02-26
Charge for thread and post creation

conectado:
# Fix `forum_pallet` for #2196 

In #2196 we found that thread/post creation has a low cost and can be used to fill up the storage.

To prevent this we add a cost to creating a thread or post and then pay the sum of those costs to the deleted of the thread to incentivize deletion.

Furthermore, we move the posts from their own storage to within the thread storage as part of the `Thread` struct to improve deletion times.

This also fixes Joystream/audits#5

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -7,6 +7,7 @@ pub use serde::{Deserialize, Serialize};
 
 use codec::{Codec, Decode, Encode};
 pub use frame_support::dispatch::DispatchResult;
+use frame_support::traits::{Currency, ExistenceRequirement};
 use frame_support::weights::Weight;
 use frame_support::{
     decl_error, decl_event, decl_module, decl_storage, ensure, traits::Get, Parameter,
@@ -14,8 +15,10 @@ use frame_support::{
 use frame_system::ensure_signed;
 use sp_arithmetic::traits::{BaseArithmetic, One};
 pub use sp_io::storage::clear_prefix;
-use sp_runtime::traits::{MaybeSerialize, Member};
-use sp_runtime::SaturatedConversion;
+use sp_runtime::traits::{AccountIdConversion, MaybeSerialize, Member, Saturating};
+use sp_runtime::{ModuleId, SaturatedConversion};
+use sp_std::collections::btree_map::BTreeMap;
+use sp_std::fmt::Debug;
 use sp_std::prelude::*;
 
 use common::origin::MemberOriginValidator;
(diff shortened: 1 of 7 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -210,8 +242,14 @@ pub struct Thread<ForumUserId, CategoryId, Moment, Hash> {
     /// poll description.
     pub poll: Option<Poll<Moment, Hash>>,
 
-    // Number of posts in thread, needed for map limit checks
+    /// Number of posts in thread, needed for map limit checks
     pub num_direct_posts: u32,
```
bedeho: is `Post::num_direct_posts` needed any longer? `posts` probably exposes a coutner for the number of mappings.

runtime-modules/forum/src/lib.rs
```diff
@@ -110,9 +129,22 @@ pub trait Trait: frame_system::Trait + pallet_timestamp::Trait + common::Trait {
         + From<u64>
         + Into<u64>;
 
+    /// Base deposit for any thread (note: thread creation also needs a `PostDeposit` since
+    /// creating a thread means also creating a post)
+    type StartingCleanupPayOff: Get<Self::Balance>;
```
bedeho: Please find a more descriptive name.

runtime-modules/forum/src/lib.rs
```diff
@@ -738,14 +777,22 @@ decl_module! {
...
+                cleanup_pay_off,
             };
 
             // Store thread
             <ThreadById<T>>::mutate(category_id, new_thread_id, |value| {
                 *value = new_thread.clone()
             });
 
+            // Reserve cleanup pay off in the thread account
+            Self::transfer_to_thread_account(cleanup_pay_off, new_thread_id, &account_id);
```
bedeho: Can we call this account which holds funds associated with state cleanup bounties something like `state cleanup treasury account`.

bedeho: Why do you not need to actually check that there is sufficient balance for this call in the pre-mutation stage?

shamil-gadelshin (changes requested):
Adding to upcoming Bedeho's review:
- Is it the case that we could have a non-zero thread account balance if we delete a thread without previously deleting posts?
- You didn't change the AccountId type for mocks from the u64 to u128 as in Parity's recommendation. Are you sure that you get different account ids for different threads in the test environment?

runtime-modules/forum/src/lib.rs
bedeho: Ah. found this in `ensure_can_create_thread` now.

bedeho (changes requested):
First, off excellent work, looks correct.

- `Post` no longer needs `thread_id` when inlined.
- Could you create an issue about dropping archiving threads functionality, it seems now that we can effectively drop threads from storage that its not worth having, and it can confuse people about when and why to do cleanups.
- Could you do a handbook PR to reconcile?

bedeho:
Forgot to ask, is there any reason to couple this to the `balances` module?  could you not just use the `Currency` trait or something similar to keep the coupling loose and tests without sideffects?

conectado:
>     * Could you create an issue about dropping archiving threads functionality, it seems now that we can effectively drop threads from storage that its not worth having, and it can confuse people about when and why to do cleanups.

Created in #2222

conectado:
>     * Is it the case that we could have a non-zero thread account balance if we delete a thread without previously deleting posts?

Could you re-phrase this? I don't understand. However, as it currently stand deleting post doesn't change the cleanup thread pay off.

>     * You didn't change the AccountId type for mocks from the u64 to u128 as in Parity's recommendation. Are you sure that you get different account ids for different threads in the test environment?

Added a test to check for this and inded I needed to change to u128.

conectado:
>     Could you do a handbook PR to reconcile?

In Joystream/handbook#34

bedeho (approved):
Good job.

PR #2211, opened 2021-02-26
Remove cd-schemas package

mnaamani:
dropped "@joystream/cd-schemas", preserved the script to initialize a content lead in api-scripts.
So the old way 
`yarn workspace @joystream/cd-schemas initialize:lead`
is now 
`yarn workspace api-scripts initialize-content-lead`

Removed the package from dependencies section of joystream-cli, network-tests

@ondratra @iorveth if you were planning on writing any "scripts" to play around with the content directory you should put them in `utils/api-scripts/src` for now. Obviously we will ultimately have more mature code in the network integration tests.

I'm not sure if the json-schema related code in the old cd-schemas package is useful for the mass upload scripts or even the joystream-cli to parse valid json, but in any case the code can still be referenced from master branch and it makes sense to start with a clean package.


bwhm (approved):
This looks reasonable to me.

Issue #2165, opened 2021-02-11
Customise Curator groups permission model for sumer


mnaamani:
## CuratorGroups permission model

The current CuratorGroups permissions model in the new content directory is as follows:

- The content working group lead can create and delete curator groups.
- The lead can add workers from the content working group into one or more curator groups.
- Once a worker is in a group they have permission to censor and uncensor Channels and Videos.
- A CuratorGroup can own a channel. Any curator from that group can create a new channel for the group, and subsequently update it.
- A curator in one group cannot censor channels or videos owned by other curator groups.
- Lead can manage curatorgroup owned channels. (need to double check this one!)

This is decent but a bit limited. The grouping is really only for the purpose of managing ownership of channels. 

## Proposal
Give the lead more flexibility in organizing their workers responsibilities.

#### Each curator group will have a set of explicit permission which can only be updated by the lead:

- can_create_channel_categories
- can_create_video_categories
- can_censor_in_channel_categories: [ ... array or category ids ...]
- can_censor_in_any_channel (except other curatorgroup owned channels)
- can_move_channels_between_categories
- can_delete_channel_categories? - or restrict to lead
- can_delete_video_categories? - or restrict to lead
- can_create_group_channel

bwhm:
Some questions here:
- what does `can_censor_in_channel_categories` and `can_censor_in_any_channel` mean?
- can the lead (and/or other curators) gain the permission to delete channels and videos not owned by the group?
- can he lead (and/or other curators) gain the permission to delete or update assets?
- my two cents on `can_delete_video_categories` and `can_delete_channel_categories` is that it should probably be restricted to lead, but wouldn't be a big problem if it's not.

I guess I could check the implementation for these, but:
- is channel/video category required? If so, what happens if that category gets deleted?
- can a channel/video have more than one category?
- will the curator pass their group id in the extrinsic, or does the runtime check all the groups for their `curatorId`?

In general, I like the idea that curators have no permissions by default. It would be _nicer_ if it could be set by `fill_opening`, but not important enough to make the change.

mnaamani:
> Some questions here:
> 
>     * what does `can_censor_in_channel_categories` and `can_censor_in_any_channel` mean?
`can_censor_in_any_channel: boolean` if set to true would allow curators to censor any channel in what-ever ChannelCategory it is in (and all videos under that channel individually)

`can_censor_in_channel_categories` if latter permission is set to false, censoring would be restricted to channels under this list of categories.

[...]

PR #2214, opened 2021-03-01
content-metadata: channels and videos can be in multiple categories

mnaamani:
Allow metadata for Channels and Videos to support multiple categories, as discussed in https://github.com/Joystream/joystream/issues/2165#issuecomment-782026174


mnaamani:
cc: @ondratra If we merge this you'll need to update the query-node input schema so videos and categories can be in multiple categories.

bedeho:
We decided to table this.

PR #2215, opened 2021-03-01
Content directory/no deletion

mnaamani:
Remove support to delete curator groups and channels in content-directory - ref: https://github.com/Joystream/joystream/issues/2165#issuecomment-787788908

### DIFF
--- a/runtime-modules/content/src/errors.rs
+++ b/runtime-modules/content/src/errors.rs
@@ -58,9 +58,6 @@ decl_error! {
         /// Channel does not exist
         ChannelDoesNotExist,
 
-        /// Channel must have been deleted
-        ChannelMustNotExist,
-
         /// Video does not exist
         VideoDoesNotExist,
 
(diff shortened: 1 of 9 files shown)


iorveth (approved):
LGTM!

bedeho (commented):
I actually think removing curator groups is not really needed, that is unlikely to change in the future, but perhaps there was some other reason for this?

mnaamani:
> I actually think removing curator groups is not really needed, that is unlikely to change in the future, but perhaps there was some other reason for this?

The  reason for removing support for "removing curator groups" was originally just a by-product of removing ability to delete channels. It became no longer necessary to actually keep tracked of how many channels were owned by a group, and more importantly deleting a group that owned a channel would have left the channel with dangling ref to an owner. Of course we could have added an extrinsic to give Lead option to change the owner of curator group owned channel to another group but that just seemed more overhead.

In the process of thinking through that I also realized that practically a Lead will never really find a need delete groups since they have flexibility to move curators between the groups, and I could only think of good reasons for them needed to a create new groups.

PR #2201, opened 2021-02-24
Bounty work submission.

shamil-gadelshin:
#### Changes
- limits for minimal values of cherry and funding.
- removed initial creator funding
- allow the council to fund the bounty
- new extrinsics:
  - announce_work_entry
  - withdraw_work_entry
  - submit_work
- closed assurance contract checks

[Main issue](https://github.com/Joystream/joystream/issues/1998)

PR #2218, opened 2021-03-02
Add bounty judgments.

shamil-gadelshin:
### Changes
- limits for minimal values of cherry and funding.
- removed initial creator funding
- allow the council to fund the bounty
- new extrinsics:
  - announce_work_entry
  - withdraw_work_entry
  - submit_work
  - submit_oracle_judgment
- closed assurance contract checks
- major rework of the bounty stage calculation

Supersedes this [PR](https://github.com/Joystream/joystream/pull/2201)


runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -367,17 +503,302 @@ benchmarks! {
...
         assert!(!Bounties::<T>::contains_key(bounty_id));
         assert_last_event::<T>(Event::<T>::BountyRemoved(bounty_id).into());
     }
+
+    announce_work_entry {
+        let cherry: BalanceOf<T> = 100.into();
+        let funding_amount: BalanceOf<T> = 100.into();
+        let stake: BalanceOf<T> = 100.into();
+
+        let member_ids = (0.. T::MaxWorkEntryLimit::get())
```
conectado, line 517: Perhaps we could use a variable `let i in 0 .. T::MaxWorkEntryLimit::get() + 1` and then here do `let members_ids = (0..i + 1)` to fine tune a little the benchmark and take the linear relation into account?

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -367,17 +503,302 @@ benchmarks! {
...
+            cherry,
+            max_amount: funding_amount,
+            ..Default::default()
+        };
+
+        let bounty_id = create_funded_bounty::<T>(params);
+
+        let (account_id, member_id) = member_funded_account::<T>("member1", 1);
+
+        Bounty::<T>::announce_work_entry(
```
conectado, line 602: Shouldn't the bounty have a `funding_period` so that when `get_bounty_milestone_on_work_submitting` gets called the worst branch is called? Also, can we add an assertion about the `BountyMilestone` before calling the extrinsic to be clear that we are hitting the worst-case scenario.

conectado (changes requested):
Benchmarks looks good, just left 2 comments

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: I implemented just the worst case. We cannot introduce "let i 0..MaxEntryLimit" here because the extrinsic parameters don't contain any dependency on the underlying state (WorkEntries storage size). We could get this using the special weight function, but it contradicts Parity's recommendation  ("never access the state in the weight function").

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: Fixed in this [PR](https://github.com/Joystream/joystream/pull/2231)

shamil-gadelshin:
Superseded by this [PR](https://github.com/Joystream/joystream/pull/2231)

runtime-modules/bounty/src/benchmarking.rs
conectado: I think it's still better if we use the 'i in 0..MaxEntryLimit' and then when calculating the weight we pass the `T::MaxWorkEntryLimit::get()` as a parameter like we do [here](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/referendum/src/lib.rs#L452-L454).

This leaves the door open to weight refund if it's ever needed.

shamil-gadelshin: Ok. Fixed in this [PR](https://github.com/Joystream/joystream/pull/2231)

Issue #2166, opened 2021-02-11
Implement faucet feature in runtime


mnaamani:
Runtime side implementation of faucet to support member onboarding in atlas

mnaamani:
done in https://github.com/Joystream/joystream/pull/2219

PR #2219, opened 2021-03-03
Locked balance for screened members

mnaamani:
Feature: Extend membership:add_screened_member() dispatchable to take one additional argument, the amount of funds to endow the new member account with.

The new account is endowed with a balance that is not transferable, but spendable on fees, tips, slashing, to allow member to start interacting with the chain. This is achieved using balance locks.

A runtime constant limits the maximum that the authority can use at their discretion.

The new account for the member cannot be an existing account that has previously been used on chain (This is to avoid applying locks on arbitrary accounts).

The add_screened_member()  dispatch call is expected to be called by a backend service which Atlas will communicate with. It would be ideal but not necessary if Atlas can send a signature along with the HTTP request to the backend service to demonstrate ownership of the account being selected for the member.


mnaamani:
Implements https://github.com/Joystream/joystream/issues/2166

bedeho (approved):
LGTM.

PR #2221, opened 2021-03-03
Working group: make apply opening stake mandatory and store relation between application and opening

### LINKED ISSUE #2196: Pallets where there is no deposit charged for filling up the storage

### LINKED ISSUE #2216: There is no stored relation between an opening and an application in the working group
## The problem
When a member applies for a given opening the [application doesn't store any information about the opening originally applied to](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/working-group/src/lib.rs#L454-L460).

This means, that later on, when [filling the opening](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/working-group/src/lib.rs#L489-L493) there is no way to check if a given application is related to the opening that is being filled. **This also means that an opening can be filled by someone with less or no stakes than the required for the opening**

## Possible solutions
* Store in `Opening` the current applicants. This would make it easier(`O(number_of_applicants_for_opening)`) to check that a member isn't applying twice to the same opening as expected [here](https://github.com/Joystream/joystream/issues/2196#issuecomment-784189158). However, it would also be harder (`O(number_of_applicants_for_opening)`) to check if a set of applications are valid for a given opening.

* Store in the `Application` the corresponding opening. This would make it easier(`O(size_of_set)`) to check if a set of applications are valid for a given opening. But harder(`O(number_of_applications_in_storage)`) to check that a member isn't applying to the same opening twice.

conectado:
# Fix #2196 for `working_group` and  fix #2216 

In #2196 we decided to make the application for an opening require stake as an anti-spam mechanism for applications.

This PR includes:
* Make stake required for application in `working_group`
  * Fix tests and benchmarks
* Small fix for fee analysis to correctly calculate the minimum stake
* Store opening within application
  * When filling an opening verify the applications are consistent with the opening
  * Add tests for incorrect mapping between opening and application

**Note:** the corresponding handbook PR is Joystream/handbook#35

This also fixes Joystream/audits#14

### DIFF
--- a/runtime-modules/working-group/src/benchmarking.rs
+++ b/runtime-modules/working-group/src/benchmarking.rs
@@ -18,11 +18,6 @@ use membership::Module as Membership;
 const SEED: u32 = 0;
 const MAX_BYTES: u32 = 16384;
 
-pub enum StakingRole {
-    WithStakes,
-    WithoutStakes,
-}
-
 fn assert_last_event<T: Trait<I>, I: Instance>(generic_event: <T as Trait<I>>::Event) {
     let events = System::<T>::events();
     let system_event: <T as frame_system::Trait>::Event = generic_event.into();
@@ -38,15 +33,11 @@ fn get_byte(num: u32, byte_number: u8) -> u8 {
 fn add_opening_helper<T: Trait<I>, I: Instance>(
     id: u32,
     add_opening_origin: &T::Origin,
-    staking_role: &StakingRole,
     job_opening_type: &OpeningType,
 ) -> OpeningId {
-    let staking_policy = match staking_role {
-        StakingRole::WithStakes => Some(StakePolicy {
-            stake_amount: One::one(),
-            leaving_unstaking_period: T::MinUnstakingPeriodLimit::get() + One::one(),
-        }),
-        StakingRole::WithoutStakes => None,
+    let staking_policy = StakePolicy {
+        stake_amount: T::MinimumStakeForOpening::get(),
+        leaving_unstaking_period: T::MinUnstakingPeriodLimit::get() + One::one(),
     };
 
     WorkingGroup::<T, _>::add_opening(
@@ -70,18 +61,14 @@ fn add_opening_helper<T: Trait<I>, I: Instance>(
 
 fn apply_on_opening_helper<T: Trait<I>, I: Instance>(
     id: u32,
-    staking_role: &StakingRole,
     applicant_id: &T::AccountId,
     member_id: &T::MemberId,
     opening_id: &OpeningId,
 ) -> ApplicationId {
-    let stake_parameters = match staking_role {
-        StakingRole::WithStakes => Some(StakeParameters {
-            // Due to mock implementation of StakingHandler we can't go over 1000
-            stake: min(BalanceOf::<T>::max_value(), BalanceOf::<T>::from(1000)),
-            staking_account_id: applicant_id.clone(),
-        }),
-        StakingRole::WithoutStakes => None,
+    let stake_parameters = StakeParameters {
+        // Due to mock implementation of StakingHandler we can't go over 1000
+        stake: min(BalanceOf::<T>::max_value(), BalanceOf::<T>::from(1000)),
+        staking_account_id: applicant_id.clone(),
     };
 
     WorkingGroup::<T, _>::apply_on_opening(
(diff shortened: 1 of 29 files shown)


runtime-modules/working-group/src/lib.rs
```diff
@@ -492,28 +496,42 @@ decl_module! {
...
+                    !<CurrentLead<T,I>>::exists(),
+                    Error::<T, I>::CannotHireLeaderWhenLeaderExists
+                );
             }
 
-            let checked_applications_info = checks::ensure_succesful_applications_exist::<T, I>(&successful_application_ids)?;
+            let checked_applications_info =
+                checks::ensure_succesful_applications_exist::<T, I>(&successful_application_ids)?;
+
+            // Check that all applications are for the intended opening
```
shamil-gadelshin, line 525: Consider using `ensure!` macro here.

runtime-modules/working-group/src/lib.rs
```diff
@@ -414,48 +416,48 @@ decl_module! {
...
+                      &p.stake_parameters.staking_account_id
+                  ),
+                  Error::<T, I>::InvalidStakingAccountForMember
+            );
 
-                ensure!(
-                    T::StakingHandler::is_account_free_of_conflicting_stakes(&sp.staking_account_id),
-                    Error::<T, I>::ConflictStakesOnAccount
-                );
+              ensure!(
```
shamil-gadelshin: Incorrect code indentation here.

shamil-gadelshin: Also, line length greater than 99 (official rust recommendation)

runtime/src/lib.rs
```diff
@@ -697,6 +697,10 @@ parameter_types! {
     pub const StorageWorkingGroupRewardPeriod: u32 = 14400 + 20;
     pub const ContentWorkingGroupRewardPeriod: u32 = 14400 + 30;
     pub const MembershipRewardPeriod: u32 = 14400 + 40;
+    // This should be more costly than `add_opening` fee with the current configuration
+    // the base cost of `add_opening` in tokens is 193. And has a very slight slope
+    // with the lenght with the length of rationale, with 2000 stake we are probably safe.
+    pub const MinimumStakeForOpening: u32 = 2000;
```
shamil-gadelshin: Incorrect type. Should be `Balance`.

runtime-modules/working-group/src/benchmarking.rs
```diff
@@ -951,46 +899,16 @@ benchmarks_instance! {
                 1,
                 description,
                 OpeningType::Regular,
-                Some(stake_policy),
+                stake_policy,
                 Some(BalanceOf::<T>::max_value())
             ).into()
         );
     }
```
shamil-gadelshin, line 907: Why did you remove this branch? Also, in "leave_role" benchmark you could use "_" syntax extrinsic call.

shamil-gadelshin (changes requested):
Looks good. Left several cosmetics comments.

runtime-modules/working-group/src/benchmarking.rs
conectado: Because since the stake is now required there is no way to leave the role immediatly, which also means I should remove that branch from the extrinsic(Since no working group should have `MinimumStakeForOpening == 0` since that would prevent the stake from being an anti-spam mechanism and we require the leaving period for any application to be exclusively greater than `MinUnstakingPeriod` which means that no application can have an unstaking period of 0)

bedeho:
Removing myself from review here.

shamil-gadelshin (approved):
LGTM

PR #2225, opened 2021-03-04
Query node mappings

ondratra:
Query node mappings for new content directory.

This PR is based on work in #2167 and assumes it will be merged before this PR.


ondratra:
At the moment, only rough implementation of mappings is done. Mappings haven't been compiled (= there are probably compilation bugs) yet due to Hydra errors that are being fixed at the moment. Only videos, channels, and their respective categories are handled now. 

Related entities as Language, MediaType, etc., are yet to be handled. There might be some extra mappings for events that no longer exist or missing mappings for some events due to recent changes to content directory implementation that are not reflected in the mappings yet.

Relevant mappings are located at `query-node/mappings/mappingsContent.ts`.

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+// TODO: handling of Language, MediaType, etc.
+
+import { SubstrateEvent } from '@dzlzv/hydra-common'
+import { DatabaseManager } from '@dzlzv/hydra-db-utils'
+
+// protobuf definitions
+import {
+  ChannelMetadata,
+  ChannelCategoryMetadata
+} from '../../content-metadata-protobuf/compiled/proto/Channel_pb'
```
mnaamani: This is not the best way to import from the workspace package.
You should add the package as a dependency and 
`import { ChannelMetadata, ChannelCategoryMetadata } from '@joystream/content-metadata-protobuf'`

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+import {
+  PublishedBeforeJoystream as PublishedBeforeJoystreamMetadata,
+  License as LicenseMetadata,
+  MediaType as MediaTypeMetadata,
+  VideoMetadata,
+  VideoCategoryMetadata,
+} from '../../content-metadata-protobuf/compiled/proto/Video_pb'
+
+// enums
+import { Network } from '../generated/graphql-server/src/modules/enums/enums'
```
mnaamani: the graphql-server pacakge should also now be a workspace npm package and should be added as a dependency in the mappings package, and imported not through a relative link to the ../generated/graphql-server/src directory

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+  */
+
+  const protobufContent = readProtobuf(ProtobufEntity.Channel, (event.params[3] as any).meta) // TODO: get rid of `any` typecast
+
+  const channel = new Channel({
+    id: event.params[0].toString(), // ChannelId
+    isCensored: false,
+    videos: [],
+    happenedIn: convertblockNumberToBlock(event.blockNumber),
+    ...protobufContent.toObject()
```
mnaamani: This should work as long as all the field names and javascript object from the protobuf message match the data model

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+  ChannelUpdateParameters<ContentParameters, AccountId>,
+  */
+
+  const channelId = event.params[1].toString()
+  const channel = await db.get(Channel, { where: { id: channelId } })
+
+  const protobufContent = readProtobuf(ProtobufEntity.Channel, (event.params[3] as any).meta) // TODO: get rid of `any` typecast
+
+  const updatedChannel = new Channel({
+    // TODO: check that this kind of new updated channel construction is valid or it should rather be edited as `channel.myProperty = something`
```
mnaamani: You are correct,
This object expansion might have worked for creation, but not updating.
All the fields are optional and must be checked individually to see if they have been set. And only update those properties of the Channel. see https://developers.google.com/protocol-buffers/docs/reference/javascript-generated?hl=zh-CN

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+    ...channel,
+    ...protobufContent.toObject()
+  })
+
+  await db.save<Channel>(updatedChannel)
+}
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export async function content_ChannelDeleted(
```
mnaamani: ChannelDeleted event has been dropped, I think you already mentioned that some events are now removed in latest runtime.

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+  ContentActor,
+  ChannelId,
+  VideoId,
+  VideoCreationParameters<ContentParameters>,
+  */
+
+  const protobufContent = readProtobuf(ProtobufEntity.Video, (event.params[3] as any).meta) // TODO: get rid of `any` typecast
+
+  const channel = new Video({
+    id: event.params[1].toString(), // ChannelId
```
mnaamani: wrong id

query-node/mappings/mappingsContent.ts
```diff
@@ -0,0 +1,464 @@
...
+
+  for (let videoId in videoIds) {
+    const video = await db.get(Video, { where: { id: videoId } })
+
+    video.isFeatured = true;
+
+    await db.save<Video>(video)
+  }
+
+  // TODO: remove featured flag from previously featured videos
```
mnaamani: Yes. Try to do the update in the most efficient way by reducing number of db updates, by getting array of current featured videos, and intersecting with array of new set of featured videos to get two array of videos to unset and videos to set..

mnaamani (changes requested):
Left a few important points about dealing with updates.

dzhelezov:
The `bootstrap` module is an ancient artifact used to pre-load the data (e.g. seeded in the genesis block and thus not available via event data). I think we don't need it anymore and it may be safely deleted

query-node/manifest.yml
```diff
@@ -2,37 +2,41 @@ version: '0.1'
...
+    - members.setControllerAccount
+  outDir: ./generated/types
+  customTypes:
+    lib: '@joystream/types/augment/all/types'
+    typedefsLoc: '../types/augment/all/defs.json'
 mappings:
   hydraCommonVersion: '0.0.3'
-    # process only blocks with height >= 1M 
+    # process only blocks with height >= 1M
   blockInterval: '[1000000,]'
```
dzhelezov: I think `blockInterval` is not need here as we're going to run all the blocks from zero height?

query-node/manifest.yml
```diff
@@ -2,37 +2,41 @@ version: '0.1'
 description: Joystream query-node manifest file for sumer
 repository: https://github.com/Joystream/joystream
 dataSource:
-  kind: substrate 
+  kind: substrate
   chain: joystream
   indexerVersion: '0.1.6'
```
dzhelezov, line 8: `data.indexerVersion` and `mappings.hydraCommonVersion` should be removed and replaced with a single top-level `hydraVersion: "2"`

metmirr:
Apparently, it escaped, would be better to use camel case https://github.com/ondratra/joystream/blob/baff0058429225cde08d392f6bdc60ecb527e04b/query-node/schema.graphql#L349. In the generated files you will notice the property is camel case already.

query-node/mappings/mappingsContent.ts
ondratra: Imports were improved

query-node/mappings/mappingsContent.ts
ondratra: Imports were improved.

query-node/mappings/mappingsContent.ts
ondratra: Db object updates has been fixed.

query-node/mappings/mappingsContent.ts
ondratra: This approach is now used only when creating new object - the approach is valid there.

query-node/mappings/mappingsContent.ts
ondratra: Event handler has been removed.

query-node/mappings/mappingsContent.ts
ondratra: Fixed

query-node/mappings/mappingsContent.ts
ondratra: Diff sets are now calculated, and only changed videos in the featured video list will get updated.

query-node/manifest.yml
ondratra: `blockInterval` has been removed

query-node/manifest.yml
ondratra: Manifest has been updated to work with the most recent Hydra version.

ondratra:
All previously raised points were resolved. Currently, the mappings are almost complete (see code commentaries below) and compilable. They are yet to be tested in integration tests. 

To compile mappings from this branch, follow the instructions at the bottom of the `query-node/README.md`.

query-node/mappings/src/common.ts
```diff
@@ -0,0 +1,86 @@
...
+    owner,
+    addedAt: blockNumber,
+    typeId: contentParameters.type_id.toNumber(),
+    // `size` is masked by `size` special name in `Struct` so there needs to be `.get('size') as unknown as u64` to retrieve proper value
+    size: (contentParameters.get('size') as unknown as u64).toBn(),
+
+    // TODO: liason & judgement information can't be retrieved now - it's neither event parameter or extrinsic parameter
+    liaisonId: new BN(0),
+    liaisonJudgement: LiaisonJudgement.PENDING,
+
```
ondratra, line 60: Liaison and liaison judgment can't be retrieved right now from runtime module `content` extrinsic (for example, `create_channel`) because it is not part of extrinsic or event parameters. The same goes for `add_content` extrinsic of `storage` module.

This will need a little runtime change, am I right?

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,382 @@
+// TODO: check all `db.get()` and similar calls recieve a proper type argument (aka add `.toString()`, etc. to those calls)
+// TODO: can we rely on db having "foreign keys"? When item is deleted will automaticly be all relations to it unset?
+//       Similarly, will saving item also save all its related items no-yet-saved in db, or do they need to saved individually?
+//       Also, is the assumption that `db.save(MyType, {myProperty: undefined})` unsets value in db correct?
```
ondratra: Can we rely on these assumptions, or we need to handle db save/delete for related items manually?

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,382 @@
...
+
+    return owner
+  }
+
+  if (actor.isLead) {
+    // TODO: who will be owner?
+  }
+
+  if (actor.isCurator) {
+    // TODO: who will be owner?
```
ondratra: When a curator changes or creates a new channel/video, who will become the owner of newly uploaded assets?

query-node/mappings/src/content/video.ts
```diff
@@ -0,0 +1,355 @@
...
+      video[key] = value
+    }
+
+    // license has changed - delete old license
+    if (originalLicense && video.license != originalLicense) {
+      await db.remove<License>(originalLicense)
+    }
+  }
+
+  // TODO: handle situation when only assets changed
```
ondratra: When video is updated this event parameter is provided:
```
export interface VideoUpdateParameters extends Struct {
  readonly assets: Option<Vec<NewAsset>>;
  readonly new_meta: Option<Bytes>;
}
```

This seems like only assets (without meta) can be updated. But what would that mean? Metadata usually contains asset indices for assets that should be associated with, let's say cover photo, etc.

query-node/mappings/src/membership.ts
```diff
@@ -0,0 +1,153 @@
...
+    id: memberId.toString(),
+    rootAccount: accountId.toString(),
+    controllerAccount: accountId.toString(),
+    handle: convertBytesToString(handle.unwrapOr(null)),
+    about: convertBytesToString(about.unwrapOr(null)),
+    avatarUri: convertBytesToString(avatarUri.unwrapOr(null)),
+    registeredAtBlock: event_.blockNumber,
+    registeredAtTime: new Date(event_.blockTimestamp.toNumber()),
+    // TODO: in the runtime there is currently no way to distinguish distinguish `buy_membership`(method `Paid`) and `add_screened_member`(`Screening`)
+    entry: MembershipEntryMethod.PAID,
```
ondratra: There seems to be a need to provide an entry method in event parameters, am I right? Event is declared as `MemberRegistered(MemberId, AccountId)` in the runtime.

query-node/schema.graphql
```diff
@@ -117,38 +117,48 @@ union DataObjectOwner = DataObjectOwnerMember
...
+  dummy: Int
 }
 
 "Asset owned by a WorkingGroup"
 type DataObjectOwnerWorkingGroup @variant {
-  "Working group identifier"
-  workingGroupId: BigInt!
+  # TODO: working group doesn't seem to have id but rather `isStorage`, `isContent`, ... shoudl we care about that?
+  #"Working group identifier"
+  #workingGroup: BigInt!
```
ondratra: Should we somehow handle working groups, or simply stating that "some" working group is owning the asset?

iorveth: Why can't we treat an working group identifier here like it's respective enum index? `workingGroup.index`

query-node/mappings/src/storage.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  if (objectOwner.isDao) {
+    const owner = new DataObjectOwnerDao()
+    owner.dao = objectOwner.asDao.toBn()
+
+    return owner
+  }
+
+  if (objectOwner.isWorkingGroup) {
```
iorveth: I guess, you made a typo here and the check here should look like `(objectOwner.isCouncil)`

query-node/mappings/src/common.ts
iorveth: During the `add_content` execution `LiaisonJudgement` seems to always be in the `Pending` stage. It waits until `Storage provider` would execute `accept_content` or 'reject_content' to update the judgement accordingly. I saw, you already have a mappings for both `accept_content` or 'reject_content'.

With regards to the liaisonId, you are right, we should add it to the `ContentAdded` event.

Why do we introduce new `liaisonId` naming instead of using already existing `storage_provider_id`, btw?

query-node/mappings/src/content/utils.ts
iorveth: `                let object_owner = StorageObjectOwner::<T>::Channel(channel_id);`

I guess, that storage object owner of channel related assets would always be a channel. `Video` would always correspond to the specific channel.

query-node/mappings/src/membership.ts
iorveth: Good point, i guess it would be easier to make a separate event for each extrinsic in runtime.

query-node/mappings/src/content/video.ts
iorveth: Seems, that there is no way for runtime to ensure the correctness between metadata and links to the respective assets. 

I guess, that means, that metadata and it's indices would remain the same as you have now, but links to the respective assets would be changed. 
Same applies to `Channels`.

iorveth (changes requested):

query-node/mappings/src/common.ts
ondratra: There was a change to liaison and judgment very recently that affects this code part. The default state truly is `Pending`, and liaison is not set before content is accepted/rejected by any storage provider.

Great point about the naming! I'm not sure why there are two names for something that seems to be one thing (or is it?).

query-node/mappings/src/content/utils.ts
ondratra: Mappings regarding the content owner have been improved. Now all the video/channel content is owned by the channel. 

Note: there still might be a different owner when the content is created via `runtime-modules/storage`'s `add_content` extrinsic.

query-node/mappings/src/content/video.ts
ondratra: Ok. So if I understand this correctly, there is no need to handle this situation in mappings? (assets are set while metadata are not)

query-node/mappings/src/membership.ts
ondratra: Ok. I created PR for this issue https://github.com/Joystream/joystream/pull/2332

query-node/schema.graphql
ondratra: You are absolutely right. These "isStorage", etc. properties in types confused me and I forgot it's enum. Fixed

query-node/mappings/src/storage.ts
ondratra: Yes, that was a typo. Fixed

metmirr:
I couldn't see that you override `createdAt` field for `Video` and `Channel` entities. if you don't then hydra insert timestamp which is not the correct time for the entity creation.

ondratra:
`createdAt`, `updatedAt` and `createdInBlock` are now properly set.

query-node/mappings/package.json
```diff
@@ -1,17 +1,21 @@
...
   },
   "dependencies": {
-    "@dzlzv/hydra-common": "0.0.3",
-    "@polkadot/types": "~2.10.2-7",
+    "@dzlzv/hydra-common": "2.1.0-beta.7",
+    "@dzlzv/hydra-db-utils": "2.1.0-beta.7",
+    "@joystream/content-metadata-protobuf": "^1.0.0",
+    "@joystream/types": "^0.16.0",
+    "iso-639-1": "^2.1.8",
+    "query-node": "^0.1.0",
```
mnaamani, line 19: the package by this name is the generated graphql-server `query-node/generated/graphql-server` and its version is 0.0.0
So what package is this supposed to be?

The build is failing because of this:
```
error Couldn't find package "query-node@^0.1.0" required by "query-node-mappings@0.1.0" on the "npm" registry.
```

mnaamani: The problem is also that this package is not available until after the code is generated.

query-node/mappings/src/common.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  DataObjectOwner,
+} from 'query-node/src/modules/variants/variants.model'
+import {
+  DataObject,
+  LiaisonJudgement,
+} from 'query-node/src/modules/data-object/data-object.model'
+import {
+  ContentParameters
+} from '@joystream/types/augment'
```
mnaamani, line 14: Why do you not import from `@joystream/types/storage/ContentParameters` ?

mnaamani: If it is to give your IDE better type information, as far as I understand the proper way to use the generated augment types are in the `tsconfig.json` like so: https://github.com/Joystream/joystream/blob/master/tests/network-tests/tsconfig.json#L13

@Lezek123 Is there anything wrong to use the import this way?

query-node/mappings/src/common.ts
```diff
@@ -0,0 +1,82 @@
...
+  contentParameters: ContentParameters,
+  blockNumber: number,
+  owner: typeof DataObjectOwner,
+): Promise<DataObject> {
+  const dataObject = new DataObject({
+    owner,
+    createdInBlock: blockNumber,
+    typeId: contentParameters.type_id.toNumber(),
+    // `size` is masked by `size` special name in `Struct` so there needs to be `.get('size') as unknown as u64` to retrieve proper value
+    size: (contentParameters.get('size') as unknown as u64).toBn(),
```
mnaamani: you can simplify to use: `contentParameters.size_in_bytes.toBn()`

query-node/mappings/src/content/utils.ts
ondratra: The first TODO was resolved. The second one can be rearticulated into:
```
// assuming there is 1:N relation between MySecondaryItem and MyMainItem
const mySecondaryItem = new MySecondaryItem({someData})
const myMainItem = new MyMainItem({mySecondaryItem})

await db.save<MyMainItem>(myMainItem)
// is `mySecondaryItem` also saved in db now?
await db.remove<MyMainItem>(myMainItem)
// is `mySecondaryItem` also removed from db now? 
```

types/augment/all/defs.json
```diff
@@ -828,5 +828,8 @@
         }
     },
     "MaxNumber": "u32",
-    "IsCensored": "bool"
+    "IsCensored": "bool",
+    "CompactAssignments": "CompactAssignmentsTo257",
+    "RewardDestination": "RewardDestinationTo257"
+}
```
mnaamani, line 834: This extra `}` is misplaced.

Technically this file shouldn't be edited manually.

mnaamani (changes requested):

query-node/build.sh
```diff
@@ -16,7 +16,6 @@ set +a
 
 yarn query-node:build
 yarn mappings:build
-ln ../types/augment/all/defs.json mappings/lib/generated/types/typedefs.json
```
mnaamani: The defs.json is mounted in the volume in docker-compose, but maybe I overlooked the need to actually also have this file copied in-case the processor is run outside of a docker container?

query-node/mappings/src/content/channel.ts
```diff
@@ -0,0 +1,313 @@
...
+      blockNumber: event.blockNumber,
+      assets: channelCreationParameters.assets,
+      contentOwner: convertContentActorToOwner(contentActor, channelId.toBn()),
+    }
+  )
+
+  // create entity
+  const channel = new Channel({
+    // main data
+    id: channelId.toString(),
```
mnaamani, line 51: Why are we using string representation for a number?
Shouldn't this be channelId.toNumber()?

query-node/mappings/src/content/channel.ts
```diff
@@ -0,0 +1,313 @@
...
+  // load channel
+  const channel = await db.get(Channel, { where: { id: channelId.toString() } as FindConditions<Channel> })
+
+  // ensure channel exists
+  if (!channel) {
+    return inconsistentState('Non-existing channel update requested', channelId)
+  }
+
+  // prepare changed metadata
+  const newMetadata = channelUpdateParameters.new_meta.isSome && channelUpdateParameters.new_meta.unwrapOr(null)
```
mnaamani: the `.isSome` check should be enough? why extra unwrapOr() test ?

mnaamani: oh I see its not a check but newMetadata will get the value. In that case the second part would be enough.

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+  type: T,
+  parameters: IReadProtobufArgumentsWithAssets,
+): Promise<Partial<T>> {
+  // true option here is crucial, it indicates that we want just the underlying bytes (by default it will also include bytes encoding the length)
+  const metaU8a = parameters.metadata.toU8a(true);
+
+  // process channel
+  if (type instanceof Channel) {
+    const meta = ChannelMetadata.deserializeBinary(metaU8a)
+    const metaAsObject = meta.toObject()
```
mnaamani: The approach you are taking here by converting to a JSON object is loosing important information about whether the field was set or not.
It would be better to use the `isSet` helper on the original meta object

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+}
+
+function handlePublishedBeforeJoystream(video: Partial<Video>, publishedAtString?: string) {
+  // published elsewhere before Joystream
+  if (publishedAtString) {
+    video.publishedBeforeJoystream = new Date(publishedAtString)
+  }
+
+  // unset publish info
+  video.publishedBeforeJoystream = undefined // plan deletion (will have effect when saved to db)
```
mnaamani, line 408: did you miss a return statement in previous if block? , otherwise the value is always being set to undefined on the following line

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+
+    // prepare language if needed
+    if (metaAsObject.language) {
+      result.language = await prepareLanguage(metaAsObject.language, parameters.db, parameters.blockNumber)
+    }
+
+    // prepare information about media published somewhere else before Joystream if needed.
+    if (metaAsObject.publishedBeforeJoystream) {
+      // this will change the `channel`!
+      handlePublishedBeforeJoystream(result, metaAsObject.publishedBeforeJoystream.date)
```
mnaamani, line 318: we should rely on the `publishedBeforeJoystream.isPublished` boolean to determine if the caller intended to set the date instead of relying on wether the date field is a string or not.

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+  return videoSize
+}
+
+async function prepareLanguage(languageIso: string, db: DatabaseManager, blockNumber: number): Promise<Language> {
+  // validate language string
+  const isValidIso = ISO6391.validate(languageIso);
+
+  // ensure language string is valid
+  if (!isValidIso) {
+    return inconsistentState(`Invalid language ISO-639-1 provided`, languageIso)
```
mnaamani: I don't think it makes sense to terminate here. There is no way to recover from this. The metadata is not going to change. Instead we should fall back to some default language code or set it to null

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+
+async function extractVideoSize(assets: NewAsset[], assetIndex: number | undefined): Promise<BN | undefined> {
+  // escape if no asset is required
+  if (assetIndex === undefined) {
+    return undefined
+  }
+
+  // ensure asset index is valid
+  if (assetIndex > assets.length) {
+    return inconsistentState(`Non-existing asset video size extraction requested`, {assetsProvided: assets.length, assetIndex})
```
mnaamani: Here again the processor shouldn't halt when user makes a mistake in the metadata.

query-node/mappings/src/content/utils.ts
```diff
@@ -0,0 +1,484 @@
...
+  return videoMeta
+}
+
+async function prepareVideoCategory(categoryId: number, db: DatabaseManager): Promise<VideoCategory> {
+  // load video category
+  const category = await db.get(VideoCategory, { where: { id: categoryId.toString() } as FindConditions<VideoCategory> })
+
+  // ensure video category exists
+  if (!category) {
+    return inconsistentState('Non-existing video category association with video requested', categoryId)
```
mnaamani: Mistake in metadata, ie user selecting no-existant category should not halt the processor.

query-node/mappings/src/common.ts
```diff
@@ -0,0 +1,83 @@
...
+    liaisonJudgement: LiaisonJudgement.PENDING, // judgement is pending at start; liaison id is set when content is accepted/rejected
+    ipfsContentId: contentParameters.ipfs_content_id.toHex(),
```
mnaamani: We expect ipfsContentId to be a Text, so don't convert to hex, instead convert to string.

types/augment/all/defs.json
ondratra: The extra `}` has been removed. 

The changes made will not be needed when Antioch is merged into Sumer (a newer Substrate version). So it's just a temporary solution that will be removed the next time someone runs `generate:all`.

query-node/mappings/src/content/channel.ts
ondratra: Yes, `isSome` check was redudant. Fixed

query-node/mappings/src/content/utils.ts
ondratra: `return` was trully missing. Fixed

query-node/mappings/src/content/utils.ts
ondratra: Check for `isPublished` added.

query-node/mappings/src/common.ts
ondratra: Conversion to text added via `.toString()`

query-node/mappings/src/content/utils.ts
ondratra: Invalid state doesn't throw error anymore.

query-node/mappings/src/content/utils.ts
ondratra: Invalid state doesn't throw error anymore.

query-node/mappings/src/content/utils.ts
ondratra: Invalid state doesn't throw error anymore.

query-node/mappings/src/content/utils.ts
ondratra: This approach has been reworked to check if value was trully set (using `has...` prefixed functions like `hasCoverPhoto()`).

query-node/build.sh
ondratra: Copy of definitions file has been added as part of building mappings. That way, types should be available both in docker and locally run processor.

query-node/mappings/src/common.ts
ondratra: Type `ContentParameters` from `@joystream/types/augment` corresponds to the actual object received from processor. It needs to be cast into custom Joystream object like this:
```
import {
  ContentParameters,
} from '@joystream/types/augment'

import { ContentParameters as Joystream_ContentParameters } from '@joystream/types/storage'
import { registry } from '@joystream/types'

...
function doIt(contentParameters: ContentParameters) {
    const joystreamContentParameters = new Joystream_ContentParameters(registry, contentParameters.toJSON() as any)
    ...
}
```
Otherwise, getters like `size_in_bytes` and similar will not be available.

query-node/mappings/src/common.ts
ondratra: The expression has been simplified. Relates to https://github.com/Joystream/joystream/pull/2225#discussion_r618459108

query-node/mappings/src/content/channel.ts
ondratra: Schema contains this definition of id: 
```
type Channel @entity {
  "Runtime entity identifier (EntityId)"
  id: ID!
...
}
```
The `ID` translates to `string` field in database. That's why there is a conversion to `string`. `toNumber()` would cause compilation error.

mnaamani (approved):
Great work getting through this.
You Still need to review locations where you call `inconsistentState()` and make sure to  only use it when there is  a runtime inconsistency. (Not protobuf metadata)

ondratra:
`inconsistentState()` has been replaced by a new `invalidMetadata()` that informs about invalid metadata, but doesn't use `console.error`. In case a bloated log becomes an issue for us, it can be silenced and kept only for development purposes (easy debug).

Merge conflict has been resolved.

mnaamani:
> `inconsistentState()` has been replaced by a new `invalidMetadata()` that informs about invalid metadata, but doesn't use `console.error`. In case a bloated log becomes an issue for us, it can be silenced and kept only for development purposes (easy debug).

You are handling Actors and Curators as metadata? Those are validated by runtime and should error with inconsistent state. Membership inconsistency will only occur if we don't bootstrap the database with all members created before the runtime upgrade block where are going to start the query-node.

PR #2230, opened 2021-03-05
Verify that staking account is valid for member when creating a proposal

### LINKED ISSUE #2005: create_proposal missing check for member staking account
In call to create_proposal extrinsic in proposals engine module, it seems we are missing one important check in 

https://github.com/Joystream/joystream/blob/2536ab0a92947524cf13044f96e07c9e1cf401f2/runtime-modules/proposals/engine/src/lib.rs#L423

The first argument `account_id` is not checked to ensure that it is a member staking account. In the same way that we check in working group module when caller passes in a staking account id.

conectado:
# Fix #2005 

This PR:
* Ensures that a valid member id is associated with the staking account when calling `create_proposal`
* Update tests accordingly and add tests for verification


shamil-gadelshin (approved):
LGTM

PR #2231, opened 2021-03-05
Complete bounty pallet. Stages finished - "work submission", "judgment", "withdrawal".

shamil-gadelshin:
This PR completes the bounty pallet creation work.

### Changes
- limits for minimal values of cherry and funding.
- removed initial creator funding
- allow the council to fund the bounty
- new extrinsics:
  - announce_work_entry
  - withdraw_work_entry
  - submit_work
  - submit_oracle_judgment
  - withdraw_work_entrant_funds
- closed assurance contract checks
- major rework of the bounty stage calculation
- rewards for winners
- unstaking for winners and "legitimate participants"
- fine-grained "invalid stage" errors
- changed crate structure - new modules
    - `actors.rs` for bounty actor management 
    - `stages.rs` for bounty stage calculation

Supersedes this [PR](https://github.com/Joystream/joystream/pull/2218)

###  Open questions
_1. if we just store winners and to_be_slashed, and we do no iteration even on
these sets, but instead require active action from everyone, then we may perhaps not even need to
limit the set size? We could store thousands of ids in either of those values without impacting
read io speed, as we have seen, so could be fine._

_2. Should we have exactly 100% reward distribution for the winners?
 We could allow less than 100% but what to do with the rest of the bounty pool?
 It could be a problem because the remaining sum could be very low to be an incentive to withdraw for funders.
 Should we introduce a remaining sum threshold that affects the "burning decision"?_


shamil-gadelshin:
#### Answering "Open questions"
1. We need to iterate through the "winners" to calculate and check the total winners' reward. We also have a limited "closed contract members" number.  So, "judgment" collection size should be limited as well.

2. We have several options:
   - demand the 100% funds distribution on the judgment,
   - burn undistributed funds (**current implementation**),
   - allow non-100% funds distribution on the judgment and allow funders to withdraw the rest. This option assumes that we don't worry about the state bloat by non-removed bounties left by funders (because of possible too low remaining funds).

conectado (approved):
Benchmarks LGTM

bedeho (changes requested):
Excellent work wrapping this, up, I think I significantly underestimated the complexity of this scope, but you have managed it well here.

Leaving a list of points here, most of which we have discussed offline.

# Conceptual

1. Make staking > 0 for all entries, and thus providing a staking account is not optional.

2. Judgements only cover entries that are to be paid (winners) or to be slashed (malicious), and there is a hardcap on the total number. Use map, not set! Judgement does cleanup for these and creator in successful bounty, so no action is required by these users. Everyone else, i.e. honest losers, are not referenced directly by judgements.

3. Remove generalized cherry reclaiming by creator, replace it by automated cleanup in 2 + a new _stage_ called `FundingExpired` which is when non-perpetual funding period expired AND no funder contributed anything. Expand `cancel_bounty` to also work in this stage, so that creator can reclaim cherry. Remove explicit `withdraw_creator_cherry`.

4. Split up generalized withdrawl stage into withdrawl stages for failed or successful bounty.

5. Do not allow over funding, if funding exceeds max limit, then only use required amount and leave funder the difference.

For 2,3,4, the following state chart summarizes functional organization of stages

bounty_module


# Practical

6. Flesh out event payloads.

7.  Reject min, max = 0 funding limits, and enf

8.  flatten WorkEntries storage map to just be single key map.

9. drop `WorkEntryRecord::last_submitted_work`.

10. drop `WorkEntryRecord::reward_account_id` and just use controller account of member

11. rename `WorkEntry` to just `Entry`.

12. there are some mutables floating around, are they truly worth it? revist.

13. Vague comment: The `BountyActorManager` type appears to have vague meaning and role, although its clearly useful in some way to do some repetitive tasks. Is there some way to sharpen this? Even the idea of a bounty actor concept is a broad, becuase it does not cover the worker role.

bedeho:
# Updated State Chart

bounty_module(1)

shamil-gadelshin:
### Review comments fixes summary

#### Major changes
- make a stake for work entries mandatory
- remove `withdraw_creator_cherry` extrinsic
- auto-transfer cherry on successful bounty judgment, or bounty canceling (veto)
- remove a bounty on cancellation or veto
- split `Withdrawal` stage into separate `SuccessfulBountyWithdrawal` and `FailedBountyWithdrawal`
- introduce exact funding for a maximum amount of limited funding and the target value for a perpetual one
- change "burn the rest" algorithm to the "demand 100% distribution" on bounty judgment
- remove work entries number limit
#### Minor changes
- introduced a limit for a minimum stake
- improve the event for `create_bounty` extrinsic
- introduce a special `FundingType` with separate parameters for limited and perpetual fundings.
- introduce a minimum funding limit.
- change `Entries` storage from a double map to a simple map
- change `EntryRecord::last_submitted_work` to work_submitted flag
- remove `EntryRecord::reward_account_id` 
- rename `WorkEntry` to `Entry`
- add "non-zero work entries number" check for the `submit_oracle_judgment` extrinsic
- add "entry has work submission" check for the `submit_oracle_judgment` extrinsic
- changes for benchmarks and tests.

bedeho (changes requested):
Phenomenal work here, just oustanding, only minor fixes

1. ` Amount of stake required, possibly 0, to enter bounty as entrant.`
2. Drop `OracleWorkEntryJudgment::Legit´, and instead make `EntryRecord::oracle_judgment_result` into an `Option`, this avoids judgements specifying legit outcomes, as well as having to set oracle judgement results that are not true.

shamil-gadelshin:
#### Next changeset
- removed `Legit` oracle judgment type for a work entry
- changed obsolete comment for `BountyCreationParameters::entrant_stake`
- changed obsolete comment for `BountyCreationParameters::cherry`

bedeho (approved):
Great!

I will leave benchmarking review to @conectado.

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -80,51 +93,150 @@ fn member_funded_account<T: Trait + membership::Trait>(
...
     create_bounty_by_council {
         let i in 1 .. MAX_BYTES;
         let metadata = vec![0u8].repeat(i as usize);
         let cherry: BalanceOf<T> = 100.into();
+        let entrant_stake: BalanceOf<T> = T::MinWorkEntrantStake::get();
+        let max_amount: BalanceOf<T> = 1000.into();
 
         T::CouncilBudgetManager::set_budget(cherry);
 
         let params = BountyCreationParameters::<T>{
```
conectado, line 222: I think this should also have a `Closed` contract type since it's the worst case scenario for `ensure_create_bounty_parameters_valid`.

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -134,32 +246,37 @@ benchmarks! {
             work_period: One::one(),
             judging_period: One::one(),
             cherry,
-            creator: BountyCreator::Member(member_id),
+            entrant_stake,
+            creator: BountyActor::Member(member_id),
+            funding_type: FundingType::Perpetual{ target: max_amount },
             ..Default::default()
         };
```
conectado, line 267: I think this should also have a `Closed` contract type since it's the worst case scenario for `ensure_create_bounty_parameters_valid`.

conectado (changes requested):
Benchmarks look mostly good, just 2 very small comments.

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: OK.

runtime-modules/bounty/src/benchmarking.rs
shamil-gadelshin: OK.

runtime-modules/bounty/src/benchmarking.rs
conectado: I think it'd be a good idea to have a `let j in 1 .. T::ClosedContractSizeLimit::get()` for the number of members, we can use that to calculate the weight in the extrinsic.

conectado (commented):
Just 1 small thing

conectado (approved):

PR #2240, opened 2021-03-10
Multiple forum fixes

conectado:
This PR contains the following fixes for the forum:
* Use `usable_balance` instead of `free_balance` to check for available funds for deposit
  * Depend on `pallet_balances` instead of `Currency` to make this possible
* Let `transfer` errors surface in case it happens.
* Use `AllowDeath` instead of `KeepAlive` to transfer fund from account

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -787,10 +787,20 @@ decl_module! {
             });
 
             // Reserve cleanup pay off in the thread account
-            Self::transfer_to_state_cleanup_treasury_account(cleanup_pay_off, new_thread_id, &account_id);
+            Self::transfer_to_state_cleanup_treasury_account(
+                cleanup_pay_off,
+                new_thread_id,
+                &account_id
+            )?;
 
             // Add inital post to thread
-            let _ = Self::add_new_post(&account_id, new_thread_id, category_id, &text, forum_user_id);
+            let _ = Self::add_new_post(
+                &account_id,
+                new_thread_id,
+                category_id,
+                &text,
+                forum_user_id
+            )?;
 
             // Update next thread id
             <NextThreadId<T>>::mutate(|n| *n += One::one());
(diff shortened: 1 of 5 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -787,10 +783,20 @@ decl_module! {
...
-            Self::transfer_to_state_cleanup_treasury_account(cleanup_pay_off, new_thread_id, &account_id);
+            Self::transfer_to_state_cleanup_treasury_account(
+                cleanup_pay_off,
+                new_thread_id,
+                &account_id
+            )?;
 
             // Add inital post to thread
-            let _ = Self::add_new_post(&account_id, new_thread_id, category_id, &text, forum_user_id);
+            let _ = Self::add_new_post(
```
shamil-gadelshin, line 791: In case of error here - we don't have "mutation safety".

shamil-gadelshin: Consider removing "token transfer" from the `add_new_post`: 
- in the case of a thread creation we will have a single transfer of the sum of two deposit values
- `add_post` will get a "transfer" right after the "mutation safe" mark.

runtime-modules/forum/src/lib.rs
```diff
@@ -787,10 +783,20 @@ decl_module! {
             });
 
             // Reserve cleanup pay off in the thread account
-            Self::transfer_to_state_cleanup_treasury_account(cleanup_pay_off, new_thread_id, &account_id);
+            Self::transfer_to_state_cleanup_treasury_account(
```
shamil-gadelshin: In case of error here - we don't have "mutation safety". Please, consider moving it before the infallible operations (like `<ThreadById<T>>::mutate`)

shamil-gadelshin (changes requested):
All forum balance issues were resolved but I found some "mutation safe" mark violations.

shamil-gadelshin (approved):
LGTM

PR #2241, opened 2021-03-11
Fetch cached joystream/node

mnaamani:
build docker images script can fetch cached image from docker hub

### DIFF
--- a/build-docker-images.sh
+++ b/build-docker-images.sh
@@ -8,6 +8,26 @@ then
   exit 0
 fi
 
+# Build or fetch cached joystream/node docker image
+if [[ "$SKIP_JOYSTREAM_NODE" = 1 || "$SKIP_JOYSTREAM_NODE" = "true" ]]; then
+  echo "Skipping build of joystream/node docker image."
+else
+  # Fetch a cached joystream/node image if one is found matching code shasum instead of building
+  CODE_HASH=`scripts/runtime-code-shasum.sh`
+  IMAGE=joystream/node:${CODE_HASH}
+  echo "Trying to fetch cashed ${IMAGE} image"
+  docker pull ${IMAGE} || :
+
+  if ! docker inspect ${IMAGE} > /dev/null;
+  then
+    echo "Fetch failed, building image locally"
+    docker-compose build joystream-node
+  else
+    echo "Tagging cached image as 'latest'"
+    docker image tag ${IMAGE} joystream/node:latest
+  fi
+fi
+
 # Build joystream/apps docker image
 echo "Building 'joystream/apps' docker image..."
 docker-compose build colossus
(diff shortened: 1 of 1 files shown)


build-docker-images.sh
```diff
@@ -8,18 +8,30 @@ then
   exit 0
 fi
 
+# Build or fetch cached joystream/node docker image
+if [[ "$SKIP_JOYSTREAM_NODE" = 1 || "$SKIP_JOYSTREAM_NODE" = "true" ]]; then
+  echo "Skipping build of joystream/node docker image."
+else
+  # Fetch a cached joystream/node image if one is found matching code shasum instead of building
+  CODE_HASH=`scripts/runtime-code-shasum.sh`
+  IMAGE=joystream/node:${CODE_HASH}
```
kdembler, line 17: I take it we use the same `scripts/runtime-code-shasum.sh` to tag the image once it's created?

kdembler (commented):
Is it possible that the build of `joystream/node` gets triggered somewhere later anyway? I run the build before that change and manually did beforehand:
```
docker pull joystream/node:sumer
docker image tag joystream/node:sumer joystream/node:latest
```
I also picked `no` to the question of whether to build the node but I think it got built anyway. Just curious, I'm not familiar with our build process so maybe I mistaken something else for the node building

build-docker-images.sh
mnaamani: correct that is done in our github actions here: https://github.com/Joystream/joystream/blob/master/.github/workflows/joystream-node-docker.yml#L18

mnaamani:
> Is it possible that the build of `joystream/node` gets triggered somewhere later anyway? I run the build before that change and manually did beforehand:

The build will get triggered if no image is found and you try to bring up the joystream-node service with `docker-compose up joystream-node`

mnaamani:
> I also picked `no` to the question of whether to build the node but I think it got built anyway. Just curious, I'm not familiar with our build process so maybe I mistaken something else for the node building

I think you were on the wrong branch when you experienced this if I remember correctly from our chat?

kdembler:
I think I then tried again on the correct one and I think it built anyway, but it's possible I'm mistaking those occasions

build-docker-images.sh
kdembler: :+1:

ondratra:
Works for me without any problem. Is there any reason why the same cache fetch is not done for other docker images (joystream/apps, pioneer, etc.)?

ondratra (approved):

mnaamani:
> Works for me without any problem. Is there any reason why the same cache fetch is not done for other docker images (joystream/apps, pioneer, etc.)?

Primarily because compiling the joystream node build takes a very long time and without any local build caching.

PR #2242, opened 2021-03-11
Add default schema to all newly created entities

metmirr:
This PR adds default schema to entities before the schema support added to them. In order to make it work, we need to seed the database with default schemas with id `0`, ie. we create an initial channel like this:

```ts
const channel = new Channel({
  handle: `Channel(0) - ${nanoid()}`,
  description: `Channel 0`,
  isPublic: false,
  isCurated: false,
})
```

**To seed the database**

`yarn workspace query-node-root db:init`

P.S there are so many codes that unused, if everything works as expected I will do a cleanup PR later.


mnaamani:
@metmirr I fixed the integration tests for this branch in https://github.com/metmirr/joystream/pull/4

PR #2246, opened 2021-03-15
Runtime: Fix Balance type alias for the council pallet.

shamil-gadelshin:
#### Problem
`decl_storage` macro creates different types for two identical rust type aliases during the metadata generation for the `council` pallet storage.

From the rust perspective, `Balance<T>` and `Balance::<T>` are identical types. However, Substrate generates different metadata for them.

The solution is to change all `Balance::<T>` entries to `Balance<T>`


mnaamani (approved):
LGTM

PR #2247, opened 2021-03-15
update protobuf package.json

mnaamani:
Make package publishable and make sure to build before publishing.


kdembler (approved):
`package.json` looks good to me, not sure what the `@suppress` is about

PR #2249, opened 2021-03-15
CLI - initial olympia-related changes

Lezek123:
This PR makes CLI work with the latest Olympia branch, with the exception of Content Directory related functionality, which is currently beeing reworked for Sumer release, so I think there is no point in updating it on this branch now, as it would only cause unnecessary conflicts. I temporarly excluded `content-directory` and `media` commands from both TypeScript and ESLint checks.

**More specifically the PR includes:**
- `@joystream/types` updates for compatibility with latest `olympia` runtime
- required changes for the CLI to work with `olympia`, mostly `Api` class and `working-groups` commands changes (some of them have also been improved in terms of scriptability, taking all the required input in form of args / flags)
- introduction of new working group commands: `apply` and `cancelOpening`
- removal of `council:info` command and some other outdated functionality / types
- rework of `account:` functionality and the way accounts are handled by the CLI (described below)
- better typesafety (added api augmentations)
- some improvements w.r.t. extrinsics (fee estimations, parameters logging etc.)

#### Accounts rework
I decided to rework the way accounts/keys are handled by the CLI.
This code (mostly residing in `AccountsCommandBase`) was the first code ever written for the CLI and it started to become quite messy as more and more code was beeing introduced "on top of it". The most notable changes are:
- There is now no need to select an account - the CLI will determine which key to use based on the context (alternatively prompt for more context if there are multiple possibilities)
- Accounts can be imported from mnemonic phrases, seeds and suris as well as backup json files. They can also be either `sr25519` or `ed25519` accounts.
- Commands that require account/address as input can use a special prompt that will allow choosing one of the existing CLI accounts, specifying a custom key or creating a new account "on fly" (for example see how `working-groups:apply` works)
- Staking account can also be created "on fly" while applying for an opening (this includes token transfer and setup via `members.addStakingAccountCandidate` and `members.confirmStakingAccount` extrinsics). This functionality can be reused for other commands that will require a staking account in the future.
- Development accounts like `Alice`, `Alice Stash`, `Bob` etc. are by default available when connecting to dev chain.

#### String fields (ie. member handles)
Since the string fields have been changed to hashes in the runtime to prevent it from storing unrelated data, I changed member handles to hashes in the CLI until it's possible to fetch this data from query node.


mnaamani:
superceded by #2298

PR #2251, opened 2021-03-15
Sumer update from master

mnaamani:
Updated constants and parameters for proposals that will be in live testnet when proposal is approved.
Bringing in same changes to sumer.


bwhm (approved):
I approve, but I think we should hold on merging until the current proposal gets approved by the council!

PR #2253, opened 2021-03-16
Use single metadata field for membership

### LINKED ISSUE #2244: Consolidate membership metadata
# Background

Memberships currently have distinct fields for different metadata types, specifically

```Rust
    /// New member user name.
    pub name: Option<Vec<u8>>,

    /// New member avatar URI.
    pub avatar_uri: Option<Vec<u8>>,

    /// New member 'about' text.
    pub about: Option<Vec<u8>>,
```

There is really not need for this. The runtime uses precious storage space that is never read, and it means that it requires a full runtime upgrade to add or change anything.

PS: Note that `handle` was excluded, because it is not metadata in the sense that the runtime actually does validate it in order to ensure uniqueness.

# Proposal

Lets consolidate all of this into a singular opaque metadata field, similar to the content directory for `Sumer`, meaning that its not stored in storage, and not validated by the runtime in any way.

**Also update handbook!**

conectado:
# Fix #2244

This PR unifies all metadata fields into `metadata`. (They already weren't stored so no change was needed for that)

Companion handbook PR: Joystream/handbook#36

### DIFF
--- a/runtime-modules/membership/src/lib.rs
+++ b/runtime-modules/membership/src/lib.rs
@@ -173,8 +173,8 @@ pub struct BuyMembershipParameters<AccountId, MemberId> {
     /// New member handle.
     pub handle: Option<Vec<u8>>,
 
-    /// Meta data concerning new member.
-    pub meta_data: Vec<u8>,
+    /// Metadata concerning new member.
+    pub metadata: Vec<u8>,
 
     /// Referrer member id.
     pub referrer_id: Option<MemberId>,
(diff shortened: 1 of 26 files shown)


runtime-modules/membership/src/lib.rs
```diff
@@ -198,17 +192,11 @@ pub struct InviteMembershipParameters<AccountId, MemberId> {
...
-
     /// New member handle.
     pub handle: Option<Vec<u8>>,
 
-    /// New member avatar URI.
-    pub avatar_uri: Option<Vec<u8>>,
-
-    /// New member 'about' text.
-    pub about: Option<Vec<u8>>,
+    /// Meta data concerning new member.
```
shamil-gadelshin, line 198: Please, make an extended comment on what is the possible content of the field (name, family name, etc).

shamil-gadelshin (changes requested):
Looks good. Just a single cosmetic request:
- please, rename "meta data", "meta", "meta_data" to "metadata" in both code and comments.

shamil-gadelshin (approved):
LGTM

PR #2255, opened 2021-03-17
Add helper script to make sudo the screening authority

mnaamani:
Useful script for local dev setup to have sudo be the screening authority, so default configuration of the membership faucet works without need for configuration. https://github.com/Joystream/membership-faucet/pull/1

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -30,6 +30,9 @@ docker-compose up -d colossus
 # Create a new content directory lead
 yarn workspace api-scripts initialize-content-lead
 
+# Set sudo as the membership screening authority
+yarn workspace api-scripts set-sudo-as-screening-auth
+
 ## Query Node Infrastructure
 # Initialize a new database for the query node infrastructure
 docker-compose up -d db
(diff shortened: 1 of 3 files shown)


kdembler (commented):
I was wondering if this is a safe approach to this? Assuming we use the screening authority in the faucet, that means we need to put sudo private key in the server running faucet, which is publicly accessible. Wouldn't it be safer to have a separate screening authority account so that if somehow the faucet server would get compromised, the damage to the system would be minimized?

mnaamani:
> I was wondering if this is a safe approach to this? Assuming we use the screening authority in the faucet, that means we need to put sudo private key in the server running faucet, which is publicly accessible. Wouldn't it be safer to have a separate screening authority account so that if somehow the faucet server would get compromised, the damage to the system would be minimized?

You are 100% correct. But this script is only meant to be used for local development. We will indeed not re-use the sudo account for the screening authority.

kdembler:
Ahh, my bad then

kdembler (approved):
Looks good

PR #2258, opened 2021-03-17
Content cli reworkings

iorveth:
Closes https://github.com/Joystream/joystream/issues/2237

PR #2259, opened 2021-03-18
Cli channels

iorveth:
This PR aims to introduce following content CLI commands:
- `joystream-cli content:channel ChannelId`
```
Shows Channel details by id.

USAGE
  $ joystream-cli content:channel ChannelId

ARGUMENTS
  ChannelId  Name or ID of the Channel
```
- `joystream-cli content:channels`
```
Lists existing content channels.

USAGE
  $ joystream-cli content:channels

```
- `joystream-cli content:createChannel`
```
Create channel inside content module. 

USAGE
  $ joystream-cli content:createChannel

OPTIONS
  -i, --input=input    Path to JSON file to use as input (if not specified - the input can be provided interactively)

  -o, --output=output  Path to the directory where the output JSON file should be placed (the output file can be then
                       reused as input)
```

- `joystream-cli content:updateChannel`
```
Update channel. 

USAGE
  $ joystream-cli content:updateChannel

OPTIONS
  -i, --input=input    Path to JSON file to use as input (if not specified - the input can be provided interactively)

  -o, --output=output  Path to the directory where the output JSON file should be placed (the output file can be then
                       reused as input)
ARGUMENTS
  ChannelId  Name or ID of the Channel
```

Closes https://github.com/Joystream/joystream/issues/2204


cli/src/commands/content/channel.ts
```diff
@@ -0,0 +1,41 @@
...
+    {
+      name: 'channelId',
+      required: true,
+      description: 'Name or ID of the Channel',
+    },
+  ]
+
+  async run() {
+    const { channelId } = this.parse(ChannelCommand).args
+    const [id, aChannel] = await this.channelEntryById(channelId)
```
Lezek123: I think `channel` here would work, previously I was using `aClass` variable name because `class` is a reserved keyword

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+      channelMetadata.setCoverPhoto(channelCreationParametersInput.meta.coverPhoto!)
+      channelMetadata.setAvatarPhoto(channelCreationParametersInput.meta.avatarPhoto!)
+      channelMetadata.setCategory(channelCreationParametersInput.meta.category!)
+
+      const serialized = channelMetadata.serializeBinary();
+
+      const api = await this.getOriginalApi()
+      
+      const metaRaw = api.createType('Raw', serialized)
+      const meta = new Bytes(api.registry, metaRaw)
```
Lezek123: I would suggest using `this.createType('Bytes', '0x' + Buffer.from(serialized).toString('hex'))` (or at least sticking with `createType` approach instead of mixed `createType` / `new Bytes`)

This also applies to `updateChannel`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+        assets: channelCreationParametersInput.assets,
+        meta,
+        reward_account: channelCreationParametersInput.reward_account,
+      }
+
+      this.jsonPrettyPrint(JSON.stringify(channelCreationParameters))
+      const confirmed = await this.simplePrompt({ type: 'confirm', message: 'Do you confirm the provided input?' })
+
+      if (confirmed && channelCreationParametersInput)  {
+        saveOutputJson(output, `${channelCreationParametersInput.meta.title}Channel.json`, channelCreationParametersInput)
```
Lezek123: Since the input can only be provided in form of a file this is probably unnecessary.
In this case I would suggest not even using the `IOFlags`, but specify a custom flag/arg instead (flag descriptions etc. don't really match in this case)

This also applies to `updateChannel`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.log('Sending the extrinsic...')
+
+        await this.sendAndFollowNamedTx(currentAccount, 'content', 'createChannel', [actor, channelCreationParameters])
+
+      }
+    } else {
+      this.log('Input invalid or was not provided...')
```
Lezek123: For this I would use `this.error`, but actually in this case you could just make input arg/flag required

This also applies to `updateChannel`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+      let channelCreationParameters: ChannelCreationParameters = {
+        assets: channelCreationParametersInput.assets,
+        meta,
+        reward_account: channelCreationParametersInput.reward_account,
+      }
+
+      this.jsonPrettyPrint(JSON.stringify(channelCreationParameters))
+      const confirmed = await this.simplePrompt({ type: 'confirm', message: 'Do you confirm the provided input?' })
+
+      if (confirmed && channelCreationParametersInput)  {
```
Lezek123: `channelCreationParametersInput` should always be truthy at this point, because earlier you make an assumption that some properties exist (ie. `meta.title` etc.), so an error would be thrown in case they didn't

This also applies to `updateChannel`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+      context = await this.promptForContext()
+    }
+
+    const currentAccount = await this.getRequiredSelectedAccount()
+    await this.requestAccountDecoding(currentAccount)
+
+    const actor = await this.getActor(context)
+
+    if (input) {
+      let channelCreationParametersInput = await getInputJson<ChannelCreationParametersInput>(input)
```
Lezek123: It may be useful to have some validation here, ie. checking if the language is a correct ISO code, title is a non-empty string etc.
Previously I was using json-schemas for that, because they could potentially be reused between query-node and the CLI and I also built the JsonSchemaPrompter so the users would be able to provide input interactively based on the schema.
The schemas were probably a little overcomplicated before because of all the relations, nested objects etc., but in this case it probably wouldn't be that much of a problem to add them (`getInputJson` previously supported schema validation, so you could use it as a reference if you decide to do that)

This also applies to `updateChannel`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+    let { context, input, output } = this.parse(CreateChannelCommand).flags
+
+    if (!context) {
+      context = await this.promptForContext()
+    }
+
+    const currentAccount = await this.getRequiredSelectedAccount()
+    await this.requestAccountDecoding(currentAccount)
+
+    const actor = await this.getActor(context)
```
Lezek123, line 38: I noticed I cannot own a channel as lead, so it would make sense to limit choices to `CuratorGroup`/`Member`

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,78 @@
...
+      const metaRaw = api.createType('Raw', serialized)
+      const meta = new Bytes(api.registry, metaRaw)
+
+      let channelCreationParameters: ChannelCreationParameters = {
+        assets: channelCreationParametersInput.assets,
+        meta,
+        reward_account: channelCreationParametersInput.reward_account,
+      }
+
+      this.jsonPrettyPrint(JSON.stringify(channelCreationParameters))
```
Lezek123: Perhaps also additionally print metadata as json object (instead of just the hex)

This also applies to `updateChannel`

cli/src/commands/content/updateChannel.ts
```diff
@@ -0,0 +1,89 @@
...
+    const { channelId } = this.parse(UpdateChannelCommand).args
+
+    if (!context) {
+      context = await this.promptForContext()
+    }
+
+    const currentAccount = await this.getRequiredSelectedAccount()
+    await this.requestAccountDecoding(currentAccount)
+
+    const actor = await this.getActor(context)
```
Lezek123, line 48: Here an actor could be derived based on the channel owner (since currently no other actor can perform channel update)

Lezek123 (changes requested):
Besides the things I mentioned there are some formatting/linter issues to fix (similarly to https://github.com/Joystream/joystream/pull/2258)

cli/src/commands/content/channel.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2259/commits/94df1b71f1e7131e12f33eaaa08d7765b4b34f42

cli/src/commands/content/createChannel.ts
iorveth: ty, done here https://github.com/Joystream/joystream/pull/2259/commits/f1aa7e5cb1dbaf84e0f2dbe704e7eb92e5c27992

cli/src/commands/content/createChannel.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2259/commits/f1aa7e5cb1dbaf84e0f2dbe704e7eb92e5c27992

cli/src/commands/content/createChannel.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2259/commits/94287f276dba0924488a71dfeff303afd5f8d20e

cli/src/commands/content/createChannel.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2259/commits/dd6cef9ab6b043bdbe75d9c66268e41ad456ecb9

cli/src/commands/content/createChannel.ts
iorveth: Good point, done here https://github.com/Joystream/joystream/pull/2259/commits/301adf882880416cd4d34d43b4c671d9fccbf3f1

cli/src/commands/content/createChannel.ts
iorveth: https://github.com/Joystream/joystream/pull/2259/commits/2c0ee3c1c438b5aca82cdeb5471b710dd60e67d9 done

cli/src/commands/content/createChannel.ts
iorveth: Good point, but it would be much harder to accomplish the same with Videos and protobuf seems to be fairly flexible, so there would probably be lots of adjustments in the future. Let's go as is now if you don't mind.

cli/src/commands/content/updateChannel.ts
iorveth: Good point, did not thought about that before, but it would not possible to convert owner into actor, because channel owner `Curator` representation does not hold `curator_id` to perform authorization. https://github.com/Joystream/joystream/blob/sumer/runtime-modules/content/src/permissions/mod.rs#L153

mnaamani:
merge into https://github.com/Joystream/joystream/pull/2334

PR #2266, opened 2021-03-19
Move away from IPNS name resolution

mnaamani:
IPNS name resolution in the storage infrastructure (mapping from storage provider id to the http endpoint where the nodes provide the API endpoint) has always been something we wanted to get rid of for two primary reasons:

1 - latency in publishing and resolution
2 - Impractical to rely on IPFS node by clients resulted in introducing a webservice hosted by storage nodes which adds a layer of complexity and caching.

Solution
With addition of an arbitrary storage value for each worker, storage node (Colossus) will now use this storage value on chain to store their endpoint. This will greatly improve the user experience both in CLI and Atlas when uploading content.


shamil-gadelshin (approved):
LGTM

iorveth (approved):
LGTM!

Issue #2267, opened 2021-03-19
Sudo can update storage limits


mnaamani:
Address issue raised in https://github.com/Joystream/joystream/pull/2250#issuecomment-801891715

mnaamani:
done in https://github.com/Joystream/joystream/pull/2272

PR #2272, opened 2021-03-21
Sudo update storage limits

mnaamani:
Implements https://github.com/Joystream/joystream/issues/2267
Added support for sudo to update storage limits.
In the process fixed some bugs and some missing checks.


runtime-modules/storage/src/data_directory.rs
```diff
@@ -369,14 +443,23 @@ decl_module! {
             // Ensure global voucher constraints satisfied.
             Self::ensure_global_voucher_constraints_satisfied(upload_voucher_delta)?;
 
+            // fill vouchers - overflow cannot happen because of prior constraint checks
+            let new_owner_voucher = owner_voucher.fill_voucher::<T>(upload_voucher_delta)?;
+            let new_global_voucher = Self::global_voucher().fill_voucher::<T>(upload_voucher_delta)?;
```
iorveth: This checks have been already performed in `ensure_global_voucher_constraints_satisfied` function.
I understand the purpose of having separate functions for both size and objects limit, though i guess it would be better to separate mutation from checks here and reuse the separate checks inside the `ensure_global_voucher_constraints_satisfied`.

iorveth (changes requested):
Looks good to me, just one suggestion.

runtime-modules/storage/src/data_directory.rs
mnaamani: Yes I see the redundant check here which I did highlight with the comment. I'll try to refactor.

mnaamani: Fixed in 8126da5

iorveth (approved):
LGTM!

PR #2277, opened 2021-03-22
Input schema for channel owner 

ondratra:
Input schema definitions for channel's owner.

### DIFF
--- a/query-node/schema.graphql
+++ b/query-node/schema.graphql
@@ -252,11 +252,14 @@ type ChannelOwnerDao @variant {
 }
 
 type CuratorGroup @entity {
+  "Runtime entity identifier (EntityId)"
+  id: ID!
+
   "Curators belonging to this group"
   curatorIds: [BigInt!]
 
   "Is group active or not"
-  isActive: Boolean
+  isActive: Boolean!
 }
 
 type VideoCategory @entity {
--- a/query-node/schema.graphql
+++ b/query-node/schema.graphql
@@ -248,7 +248,7 @@ type CuratorGroup @entity {
   "Is group active or not"
   isActive: Boolean!
 
-  videos: [Video!] @derivedFrom(field: "ownerCuratorGroup")
+  channels: [Channel!] @derivedFrom(field: "ownerCuratorGroup")
 }
 
 type VideoCategory @entity {
--- a/query-node/schema.graphql
+++ b/query-node/schema.graphql
@@ -50,7 +50,7 @@ type Membership @entity {
   "The type of subscription the member has purchased if any."
   subscription: BigInt
 
-  videos: [Video!] @derivedFrom(field: "ownerMember")
+  channels: [Channel!] @derivedFrom(field: "ownerMember")
 }
 
 "Category of media channel"
@@ -201,10 +201,10 @@ type Channel @entity {
   "Runtime entity identifier (EntityId)"
   id: ID!
 
-  "Member owning the video (if any)"
+  "Member owning the channel (if any)"
   ownerMember: Membership
 
-  "Curator group owning the video (if any)"
+  "Curator group owning the channel (if any)"
   ownerCuratorGroup: CuratorGroup
 
   category: ChannelCategory
--- a/query-node/schema.graphql
+++ b/query-node/schema.graphql
@@ -199,7 +199,7 @@ type Channel @entity {
   "Runtime entity identifier (EntityId)"
   id: ID!
 
-  #"Owner of the channel"
+  "Owner of the channel"
   owner: ChannelOwner!
 
   category: ChannelCategory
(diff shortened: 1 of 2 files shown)


query-node/schema.graphql
```diff
@@ -204,9 +201,11 @@ type Channel @entity {
   "Runtime entity identifier (EntityId)"
   id: ID!
 
-  # "Owner of the channel" Commenting out this field: 'owner' can be curator_group, lead
-  # or a member. We are not handling events related to curator group so we will not set this field
-  # owner: Member!
+  "Member owning the video (if any)"
```
kdembler: `owning the channel` I presume?

query-node/schema.graphql
```diff
@@ -54,6 +49,8 @@ type Membership @entity {
 
   "The type of subscription the member has purchased if any."
   subscription: BigInt
+
+  videos: [Video!] @derivedFrom(field: "ownerMember")
```
kdembler: Shouldn't this be `channels`? I don't see a use case in which we would want videos of a given member, we always fetch videos in context of a channel

query-node/schema.graphql
ondratra: Mistake indeed. Fixed

query-node/schema.graphql
ondratra: Yes, it should. Fixed

query-node/schema.graphql
```diff
@@ -239,6 +238,19 @@ type Channel @entity {
...
+  "Runtime entity identifier (EntityId)"
+  id: ID!
+
+  "Curators belonging to this group"
+  curatorIds: [BigInt!]
+
+  "Is group active or not"
+  isActive: Boolean!
+
+  videos: [Video!] @derivedFrom(field: "ownerCuratorGroup")
```
kdembler: Shouldn't this be `channels` as well? It seems to be referring to the same thing as the field in `Membership`

query-node/schema.graphql
```diff
@@ -54,6 +49,8 @@ type Membership @entity {
 
   "The type of subscription the member has purchased if any."
   subscription: BigInt
+
+  channels: [Channel!] @derivedFrom(field: "ownerMember")
```
kdembler: Any reason this couldn't be `[Channel!]!`?

query-node/schema.graphql
```diff
@@ -204,9 +201,11 @@ type Channel @entity {
   "Runtime entity identifier (EntityId)"
   id: ID!
 
-  # "Owner of the channel" Commenting out this field: 'owner' can be curator_group, lead
-  # or a member. We are not handling events related to curator group so we will not set this field
-  # owner: Member!
+  "Member owning the channel (if any)"
```
bedeho, line 204: Really, the representation here should be a variant, rather than two separate fields for members and curator group, because you are either owned by a member or by a curator group (or DAO as well actually, bu they do not exist yet). However, if we use that representation, you cannot do reverse lookups into the variant, as you are doing now. This is a missing feature:

https://github.com/Joystream/hydra/issues/187.

So we are probably better off just keeping this for now, but be aware of this.

query-node/schema.graphql
bedeho: Correct, curators do not directly operate channels.

bedeho (changes requested):
Only minor fix required.

query-node/schema.graphql
ondratra: If I remember correctly, we have discussed it (quite some time ago) and decided there will be a check for an array's existence and non-emptiness (aka `member.channels && member.channels.length`) in Atlas. It can be changed easily, and an array can be required (`[Channel!]!`) on all places in the input schema. What do you think?

query-node/schema.graphql
kdembler: I think they _do_ operate channels, didn't you mean they don't operate on `videos` directly?

query-node/schema.graphql
kdembler: Well, from my point of view it's more convenient if arrays are always defined but just empty, less code for us. It still has the same semantic meaning I'd say - `no items == no array == no relationships`

query-node/schema.graphql
ondratra: Yes, that was a mistake. Fixed

query-node/schema.graphql
ondratra: Ok, all arrays are now required to exist.

query-node/schema.graphql
ondratra: I'm aware of that. This approach was selected to overcome the missing feature you mentioned.

kdembler (approved):
LGTM!

bedeho (approved):
Looks good!

Issue #2269, opened 2021-03-19
Remove Constraint of uploading to a single Liason Storage Provider


mnaamani:
We currently have a constraint in the data_directory that when a data object is created a storage provider is selected at random and assigned the responsibility of processing an upload. Only that provider can 'Accept' the content.

I propose to remove this constraint and any active provider should be able to process the upload, and the client uploading can at its own discretion select the "best" provider to upload to and will be able to re-try other providers on failure. Removing this constraint also means the client will never have to create a new data object when the liaison is unavailable for a long period of time, and when/if the liaison is fired, and not need to call into the content directory to perform channel or video Update operations just to re-attempt uploading assets and delete old assets.
This greatly improves the UX both from joystream-cli and Atlas viewpoint.

I've discussed this both with @bwhm and @kdembler and got positive feedback. The implementation changes in the runtime are pretty straight forward. This leaves the question of the impact on Atlas design changes, if any, will be required. As I understand from Klaudiusz the designers are currently in the process of working on this aspect of Atlas and now would be the best time to decide whether we should go ahead with this proposal.

cc: @bedeho

bedeho:
I really like this, but it almost seems too good to be true. I wonder if we are overlooking something obvious.

Would there still be an on-chain confirmation step for a valid upload?

I think this is a good idea, if we are really sure, and it should be incorporated into v2

https://github.com/Joystream/joystream/issues/2224

This impacts some rather nuanced product related design considerations currently being worked on https://github.com/Joystream/atlas/issues/381. It may perhaps simplify them, but the work has already started along this path.

So in conclusion, **ACK** from me, but some some double checking and coordination with designers is needed.

bwhm:
As implied by @mnaamani, also thinks this seems great for the testnet.

> I really like this, but it almost seems too good to be true. I wonder if we are overlooking something obvious.

If you're only looking at this for the testnet, I don't think so. On mainnet, there needs to be some way of reliably determining whether some SPs are not performing, but from previous conversations about the design of a new storage system, it seems like we are moving towards lots of off-chain monitoring`*` anyway.

`*` keeping tabs of the various actors (storage/bandwidth providers, gateway/query node operators) interacting in the "content system" are performing.

mnaamani:
> Would there still be an on-chain confirmation step for a valid upload?

[...]

PR #2278, opened 2021-03-23
Uploads can be done to any storage provider

mnaamani:
Implements https://github.com/Joystream/joystream/issues/2269

Building on https://github.com/Joystream/joystream/pull/2266 and https://github.com/Joystream/joystream/pull/2272

Any storage provider can accept content upload. Upload can be repeated at any time, which also means that if content is ever lost by all nodes, anyone can re-upload the content.

Dropped 'Rejected' enum variant of LiaisonJudgement


storage-node/packages/cli/src/commands/download.ts
```diff
@@ -5,38 +5,27 @@ import { BaseCommand } from './base'
...
   }
 
   // Shows command usage. Overrides the abstract method from the base class.
   protected showUsage() {
     console.log(
       chalk.yellow(`
-        Usage:   storage-cli download colossusURL contentID filePath
-        Example: storage-cli download http://localhost:3001 0x7a6ba7e9157e5fba190dc146fe1baa8180e29728a5c76779ed99655500cff795 ./movie.mp4
+        Usage:   storage-cli download contentID filePath
+        Example: storage-cli download 0x7a6ba7e9157e5fba190dc146fe1baa8180e29728a5c76779ed99655500cff795 ./movie.mp4
```
shamil-gadelshin: Please, change an example for download to different CID format (eg.: "5GFwJXfTf19eUAgoHLsDwxd3YZ5ADdHs5mSzPZVU3Aw6T5Gf")

runtime-modules/storage/src/data_directory.rs
```diff
@@ -818,7 +798,8 @@ impl<T: Trait> Module<T> {
                 size: content.size,
                 added_at: common::current_block_time::<T>(),
                 owner: owner.clone(),
-                liaison,
+                // It is irrelevant what liaison is set when judgement is pending
```
shamil-gadelshin: It is confusing. Do you mind changing it to the Option?

shamil-gadelshin (changes requested):
Looks good. I tested "upload" and "download" commands. Left just a couple of small issues.

storage-node/packages/cli/src/commands/download.ts
mnaamani: makes sense, will update

runtime-modules/storage/src/data_directory.rs
mnaamani: I agree, an Option will be clearer.

storage-node/packages/cli/src/commands/download.ts
mnaamani: done in https://github.com/Joystream/joystream/pull/2278/commits/edf7e2671e3e897505158fdedefb7892c24fcbb1

runtime-modules/storage/src/data_directory.rs
mnaamani: done in https://github.com/Joystream/joystream/pull/2278/commits/fad04b7e2caf78333f360ba04741c0b92d9f5725

shamil-gadelshin (approved):
LGTM

Issue #2263, opened 2021-03-18
CLI: content censoring support


iorveth:
Add censoring support for channels and videos to the CLI.

PR #2280, opened 2021-03-23
Content censorship cli support

iorveth:
Closes https://github.com/Joystream/joystream/issues/2263


cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -57,58 +37,53 @@ export default abstract class ContentDirectoryCommandBase extends RolesCommandBa
...
-    return createType('Actor', { Curator: [groupId, curator.workerId.toNumber()] })
+    return createType('ContentActor', { Curator: [groupId, curator.workerId.toNumber()] })
   }
 
-  async promptForClass(message = 'Select a class'): Promise<Class> {
-    const classes = await this.getApi().availableClasses()
-    const choices = classes.map(([, c]) => ({ name: c.name.toString(), value: c }))
+  async promptForChannel(message = 'Select a channel'): Promise<Channel> {
+    const channels = await this.getApi().availableChannels()
+    const choices = channels.map(([id, c]) => ({ id: id.toString(), value: c }))
```
Lezek123: The values in the prompt don't say much, I think without query-node integration it would be best to just display channel ids as choices, ie.:
```
>  Channel 1
   Channel 2
   Channel 3
```
You can use choice `name` property for that.
(on the second thought it may make sense to just get rid of the prompt completely and require the channel id to be provided as arg/flag)

cli/src/commands/content/updateChannelCensorshipStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+      description: 'ID of the Channel',
+    },
+    {
+      name: 'status',
+      required: false,
+      description: 'New status of the channel (1 - active, 0 - inactive)',
+    },
+    {
+      name: 'rationale',
+      required: false,
```
Lezek123: This is required in `updateVideoCensorshipStatus`, I think it should be required in both?

cli/src/commands/content/createChannel.ts
```diff
@@ -0,0 +1,55 @@
...
+  static flags = {
+    context: ContentDirectoryCommandBase.contextFlag,
+    ...IOFlags,
+  }
+
+  async run() {
+    let { context, input, output } = this.parse(CreateChannelCommand).flags
+
+    if (!context) {
+      context = await this.promptForContext()
```
Lezek123: Here only Curator / Lead context is valid, so I would limit the choice to that (perhaps derive from selected account if possible).

cli/src/commands/content/updateChannelCensorshipStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (id === undefined) {
+      id = await this.promptForChannel()
+    }
+
+    if (status === undefined) {
+      status = await this.simplePrompt({
+        type: 'list',
+        message: 'Select new status',
+        choices: [
+          { name: 'Active', value: true },
```
Lezek123: Perhaps `Censored` / `Not censored` or something like that would be more appropriate.
With this it can be a little confusing (whether the censorship is `Active` or the channel itself).
Also the flag description says `1` means that channel/video is `active`, while in fact it would be the opposite in this case

cli/src/commands/content/updateChannelCensorshipStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+      context = await this.promptForContext()
+    }
+
+    const currentAccount = await this.getRequiredSelectedAccount()
+    await this.requestAccountDecoding(currentAccount)
+
+    const actor = await this.getActor(context)
+
+    if (id === undefined) {
+      id = await this.promptForChannel()
```
Lezek123: `promptForChannel` return `Channel` struct (not an id), which causes this command to fail.
I encourage you to strongly-type the return value (especially when it's `any` by default) to avoid mistakes like that

cli/src/commands/content/updateChannelCensorshipStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      if (status !== '0' && status !== '1') {
+        this.error('Invalid status provided. Use "1" for Active and "0" for Inactive.', {
+          exit: ExitCodes.InvalidInput,
+        })
+      }
+      status = !!parseInt(status)
+    }
+
+    await this.sendAndFollowNamedTx(currentAccount, 'content', 'updateChannelCensorshipStatus', [
```
Lezek123, line 61: Another reason why this command fails is that there is no `IsCensored` type in `@joystream/types` (after adding new type to the library you need to run a the current node locally and then run `yarn workspace @joystream/types generate:all` to update the api augmentations and a few other auto-generated files)

Lezek123 (changes requested):

cli/src/base/ContentDirectoryCommandBase.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2280/commits/c2e27791c688f13158d745920ce869fbaa6ef556

cli/src/commands/content/updateChannelCensorshipStatus.ts
iorveth: Done, thank you for the suggestion. https://github.com/Joystream/joystream/pull/2280/commits/c2e27791c688f13158d745920ce869fbaa6ef556

cli/src/commands/content/updateChannelCensorshipStatus.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2280/commits/eec63ee52532003fda9943b015babae86e0058f9

cli/src/commands/content/createChannel.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2280/commits/10ecab1598e8cd0ed8a535eb6ac9b4a38b9488a6

cli/src/commands/content/updateChannelCensorshipStatus.ts
iorveth: Done here https://github.com/Joystream/joystream/pull/2280/commits/fa486b2d95db3817776525438ff8aeded0bc63d7

cli/src/commands/content/updateChannelCensorshipStatus.ts
iorveth: Done https://github.com/Joystream/joystream/pull/2280/commits/8635bea118d08e85f862bc00685b44d3c241db15, thank you so much for the explanation.

mnaamani:
merge into https://github.com/Joystream/joystream/pull/2334

PR #2283, opened 2021-03-25
Query node input schema assets

ondratra:
Simplification of Asset representation + new redundant properties enabling queries important for Joystream.

### DIFF
--- a/query-node/schema.graphql
+++ b/query-node/schema.graphql
@@ -1,20 +1,7 @@
-# temporary type used before `DateTime` type is working
-type DateTime @entity {
-  timestamp: BigInt!
-}
-
 enum Network {
   BABYLON
   ALEXANDRIA
   ROME
-}
-
-type Block @entity {
-  "Block number as a string"
-  id: ID!
-  block: Int!
-  executedAt: DateTime!
-  network: Network!
 }
 
 enum MembershipEntryMethod {
@@ -44,7 +31,7 @@ type Membership @entity {
   rootAccount: String!
 
   "Blocknumber when member was registered"
-  registeredAtBlock: Block!
+  registeredAtBlock: BigInt!
 
   "Timestamp when member was registered"
   registeredAtTime: DateTime!
@@ -65,7 +52,7 @@ type ChannelCategory @entity {
 
   channels: [Channel!] @derivedFrom(field: "category")
 
-  happenedIn: Block!
+  happenedIn: BigInt!
 }
 
 "Asset availability representation"
@@ -87,7 +74,7 @@ type AssetUrl @variant {
 
 "Asset was deleted and is no longer available."
 type AssetDeleted @variant {
-  happenedIn: Block!
+  happenedIn: BigInt!
 }
 
 "Storage asset"
@@ -98,7 +85,7 @@ type AssetStorage @variant {
   """
   dataObject: AssetDataObject!
 
-  happenedIn: Block!
+  happenedIn: BigInt!
 }
 
 "The decision of the storage provider when it acts as liaison"
@@ -119,7 +106,7 @@ type AssetDataObject @entity {
   owner: AssetOwner!
 
   "Content added at"
-  addedAt: Block!
+  addedAt: BigInt!
 
   "Content type id"
   typeId: Int!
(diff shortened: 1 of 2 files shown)


query-node/schema.graphql
```diff
@@ -65,55 +52,40 @@ type ChannelCategory @entity {
...
+"Asset availability representation"
+enum AssetAvailability {
+  ACCEPTED, # asset is available in storage
+  PENDING, # asset is being uploaded to storage
+  URL, # asset is referencing URL
+  INVALID, # invalid storage (meta)data used
+}
+
+"Asset representation"
+union Asset = AssetUrl | AssetStorage | AssetDeleted
```
bedeho: I thought we were dropping `AssetDeleted`?

bedeho (changes requested):

query-node/schema.graphql
```diff
@@ -65,55 +52,40 @@ type ChannelCategory @entity {
...
 "Asset was deleted and is no longer available."
 type AssetDeleted @variant {
-  happenedIn: Block!
+  happenedIn: BigInt!
 }
 
-"Status of an asset upload"
-type AssetUploadStatus @variant {
+"Storage asset"
+type AssetStorage @variant {
```
kdembler: Do we actually need this now? Couldn't we link directly to `AssetDataObject`? `AssetStorage` has an additional `happenedIn` but `AssetDataObject` has `addedAt` so I don't think we're losing any data?

query-node/schema.graphql
```diff
@@ -300,12 +275,17 @@ type Video @entity {
...
+  happenedIn: BigInt!
 
   "Is video featured or not"
   isFeatured: Boolean!
 
   featured: FeaturedVideo @derivedFrom(field: "video")
+
+  # helper properties - redudant info enabling important queries for Joystream use-cases
+
+  is_thumbnail_available: AssetAvailability!
```
kdembler: 1. Shouldn't we use `camelCase` here, like in other fields?
2. `is_..._available` suggests a boolean, available or not. Maybe `thumbnailAvailability` is better?

kdembler (changes requested):
We're also missing the availability fields for channel assets

query-node/schema.graphql
ondratra: Yes, it was mistakenly kept there. Now the whole structure is simplified and flatten.

query-node/schema.graphql
ondratra: No, we don't.  Now the whole structure is simplified and flatten.

query-node/schema.graphql
ondratra: 1) Yes, that was a mistake.
2) Naming has been changed accordingly.

query-node/schema.graphql
```diff
@@ -65,55 +52,34 @@ type ChannelCategory @entity {
...
-}
+  "Asset is being uploaded to storage"
+  PENDING,
 
-"Asset was deleted and is no longer available."
-type AssetDeleted @variant {
-  happenedIn: Block!
-}
+  "Asset is referencing URL"
+  URL,
```
kdembler: I think we will be better off with removing `URL` as explicit state and use `ACCEPTED` for that? We will want to query entities with assets either accepted by storage or provided with URL and Hydra won't allow to express that (as there's no `OR` predicate) with this setup

query-node/schema.graphql
```diff
@@ -57,35 +57,29 @@ type ChannelCategory @entity {
...
 
-"Storage asset"
-type AssetStorage @variant {
-  """
-  Data object in upload life-cycle.
-  If this is deleted, then set oldDataObject in its place if it is set and not rejected, otherwise union goes to Deleted.
-  """
-  dataObject: AssetDataObject!
+"Asset representation"
+type Asset @entity {
```
bedeho: I don't think this is needed any longer.

bedeho (changes requested):
Just one minor fix.

query-node/schema.graphql
ondratra: `URL` option has been removed.

query-node/schema.graphql
ondratra: `Asset` has been removed.

kdembler (approved):

bedeho (approved):

PR #2288, opened 2021-03-26
Upgrade node to the Substrate v2.0.0

shamil-gadelshin:
SSIA

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -13,6 +13,16 @@
 #[cfg(feature = "std")]
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
+#[cfg(feature = "std")]
+/// Wasm binary unwrapped. If built with `BUILD_DUMMY_WASM_BINARY`, the function panics.
+pub fn wasm_binary_unwrap() -> &'static [u8] {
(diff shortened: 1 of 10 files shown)


conectado (approved):

runtime/src/lib.rs
```diff
@@ -13,6 +13,16 @@
 #[cfg(feature = "std")]
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
+#[cfg(feature = "std")]
+/// Wasm binary unwrapped. If built with `BUILD_DUMMY_WASM_BINARY`, the function panics.
```
mnaamani, line 17: Do we need to remove this line from our README now ?
https://github.com/Joystream/joystream/blob/master/README.md#rls-extension-in-vscode-or-atom-editors

mnaamani (approved):

runtime/src/lib.rs
shamil-gadelshin: `cargo clippy` worked in my CLion without the `BUILD_DUMMY_WASM_BINARY`. I will try again on building the v2.0.1

mnaamani: yeah we still need it in VSCode ..

PR #2292, opened 2021-03-27
Antioch integration test fixes

mnaamani:
Testing the integration tests using @Lezek123 's work to update polkadot-js/api to v4.2.1 and pioneer.

Fixes so far:
as result of change in enum serialization: https://github.com/Joystream/joystream/pull/2292/commits/0501f86df59432bd1e5c6a2ac9bb963e835f1ead

- [x] Fix api connect retry loop (get rid of temp `sleep 10`  hack in scripts)
- [x] Global logger
- [x] Disable runtime upgrade job in run-network-tests workflow since Antioch will be a new chain
- [x] Back-port tweak to build.sh to fetch cached joystream/node image from docker hub

Will leave these to @metmirr 
~- Fix Query-node hydra dependencies~
~- Fix Query-node integration tests~


tests/network-tests/src/Api.ts
```diff
@@ -53,7 +54,7 @@ export class ApiFactory {
     treasuryAccountUri: string,
     sudoAccountUri: string
   ): Promise<ApiFactory> {
-    const debug = Debugger('api-factory')
+    const debug = Debugger('integration-tests:api-factory')
```
Lezek123: I think it may be a good practice to create a global `integration-tests` debug instance and then use `.extend()` (https://www.npmjs.com/package/debug#extend)

tests/network-tests/src/Api.ts
```diff
@@ -851,30 +852,31 @@ export class Api {
...
-            record.event.method &&
-            record.event.method.toString() === 'ProposalStatusUpdated' &&
-            record.event.data[0].eq(id) &&
-            record.event.data[1].toString().includes('ExecutionFailed')
-          ) {
-            unsubscribe()
-            resolve([false, events])
-          }
-        })
+  public async waitForProposalToFinalize(id: ProposalId): Promise<InvertedPromise<[boolean, EventRecord[]]>> {
```
Lezek123: The function name seems no longer adequate

tests/network-tests/src/Api.ts
```diff
@@ -888,18 +890,20 @@ export class Api {
...
-      const unsubscribe = await this.api.query.system.events<Vec<EventRecord>>((events) => {
-        events.forEach((record) => {
-          if (record.event.method && record.event.method.toString() === eventName) {
-            unsubscribe()
-            resolve(record.event)
-          }
-        })
+  // Use this when the event we are expecting is not particular to a specific extrinsic,
+  // Normally events emitted from on_initialize() or on_finalize() in a call
+  public async waitForSystemEvent(eventName: string): Promise<InvertedPromise<Event>> {
```
Lezek123: This name also may now be a little misleading

tests/network-tests/src/flows/proposals/electionParametersProposal.ts
```diff
@@ -7,17 +7,18 @@ import { Resource } from '../../Resources'
 
 // Election parameters proposal scenario
 export default async function electionParametersProposal({ api, lock }: FlowProps): Promise<void> {
-  const debug = Debugger('flow:electionParametersProposal')
+  const debug = Debugger('integration-tests:flow:electionParametersProposal')
   debug('Started')
   await lock(Resource.Proposals)
-
+  debug('lock acquired contining flow')
```
Lezek123: Not sure if we still need those?

Lezek123 (commented):
Added a few initial suggestions.
There also seems to be an issue with `utils/api-scripts/src/dev-set-runtime-code.ts` (probably related to `patched-ws-provider`, no longer compatible with the new api):
https://github.com/Joystream/joystream/pull/2292/checks?check_run_id=2213124302

tests/network-tests/src/Api.ts
mnaamani: Good suggestion.
I had considered to pass around a 'global' debugger instance at one point, but changed my mind because I wasn't aware of that feature.

tests/network-tests/src/Api.ts
mnaamani: agree

tests/network-tests/src/Api.ts
mnaamani: agree

tests/network-tests/src/flows/proposals/electionParametersProposal.ts
mnaamani: I missed a couple of those while I was doing "console.log" debugging :)

tests/network-tests/src/Api.ts
mnaamani: will drop this, it is not used anywhere

types/README.md
```diff
@@ -26,7 +26,7 @@ async function main() {
   const provider = new WsProvider('ws://127.0.0.1:9944')
 
   // Create the API and wait until ready
-  const api = await ApiPromise.create({ provider, types })
+  const api = new ApiPromise({ provider, types })
```
Lezek123, line 29: I think `await ApiPromise.create({ provider, types })` is just a shortcut for:
```
const api = new ApiPromise({ provider, types })
await api.isReady
```
The only difference is that if we use the 2nd approach, `await api.isReadyOrError` can be used instead.

build.sh
```diff
@@ -14,42 +14,37 @@ yarn workspace storage-node build
...
-   [yY]* ) yarn cargo-checks
-           yarn cargo-build
-           break;;
-
-   [nN]* ) break;;
-
-   * )     break;;
-  esac
-done
+if [[ "$SKIP_JOYSTREAM_NODE" = 1 || "$SKIP_JOYSTREAM_NODE" = "true" ]]; then
```
Lezek123, line 45: It would be useful to skip `joystream/apps` build or local node build too (but still fetch docker image if available), but I'm not sure if it makes sense to make it one configurable script or just a few smaller ones that can be composed into a bigger one (and whether it's something that should be tackled in this or a separate PR)

tests/network-tests/src/Job.ts
```diff
@@ -30,7 +30,7 @@ export class Job {
     this._flows = flows
     this._outcome = new InvertedPromise<JobOutcome>()
     this._manager.on('run', this.run.bind(this))
-    this.debug = Debugger(`integration-tests:job:${this._label}`)
+    this.debug = extendDebug(`${this._label}`)
```
Lezek123: Was the `job:` prefix dropped on purpose?

build.sh
```diff
@@ -14,42 +14,37 @@ yarn workspace storage-node build
...
-  # if one is found matching code shasum instead of building
-  while true
-  do
-    read -p "Rebuild joystream/node docker image? (y/N): " answer2
+  # Build or fetch cached joystream/node docker image
+  if [[ "$SKIP_JOYSTREAM_NODE" = 1 || "$SKIP_JOYSTREAM_NODE" = "true" ]]; then
+    echo "Skipping build of joystream/node docker image."
+  else
+    # Fetch a cached joystream/node image if one is found matching code shasum instead of building
+    CODE_HASH=`scripts/runtime-code-shasum.sh`
```
Lezek123, line 22: I noticed that the hash that is calculated locally in my case differs from the one in the CI (locally it was `34e163be8e43c5631d8b92e9c43ab0bf0fa62b9c`)

When I ran just ran `./scripts/runtime-code-shasum.sh` from the command line I got a bunch of errors about unexisting files/directories (`Cargo.lock`, `runtime-modules`, `joystream-node.Dockerfile` etc.).

I traced it to:
```
export WORKSPACE_ROOT=`cargo metadata --offline --no-deps --format-version 1 | jq .workspace_root -r`
```
Which in my case returned `Command 'jq' not found`.

After installng `jq` the errors are no longer showing, but when I run `./scripts/runtime-code-shasum.sh` the hash is still different (`d63feb0531a22ffa2db1c04e63c7ff4fef3d3a97` compared to `f3f77e114254389a84eed86e6d1d97eb5d73e741` in the CI)

build.sh
mnaamani: I agree, I did a better approach in sumer branch where I split building npm packages, docker images and native cargo build into separate scripts. I was going to bring them into antioch branch but didn't want to go over board in this PR.

tests/network-tests/src/Job.ts
mnaamani: well spotted :), mistake

build.sh
mnaamani: the difference may come if your version of `tar` is not gnu tar. The key thing for the hash to be deterministic is for the timestamps, ordering, ownership info of files to be calculated in the same way. What version of `tar` do you have?

types/README.md
mnaamani: Good point. And we only really need the `isReadyOrError` to trap the error.

build.sh
mnaamani: After some troubleshooting we came to realize the file mode can differ between systems so we added the `--mode` argument to the mix in https://github.com/Joystream/joystream/pull/2292/commits/f68b14f25b1dd48320b8f5671cd508a2018b3d51

tests/network-tests/src/Job.ts
mnaamani: fixed in https://github.com/Joystream/joystream/pull/2292/commits/4a0bab088945c9448fd648437fba0b47885b182b

build.sh
```diff
@@ -14,42 +14,37 @@ yarn workspace storage-node build
...
+    IMAGE=joystream/node:${CODE_HASH}
+    echo "Trying to fetch cashed ${IMAGE} image"
```
Lezek123, line 24: Noticed a small typo (ca**c**hed)

Lezek123 (approved):
Looks good, the tests pass and the shasum matches the one from the PR (the image is not available, but I belive this is expected?)

Just 2 small comments:
- I noticed in some READMEs we still have `WASM_BUILD_TOOLCHAIN=nightly-2020-05-23`, but perhaps that's to be addressed in another PR?
- in `types/README` there is still the change to `new ApiPromise` (nothing wrong with it really, but just pointing out this was only changed in one of the 2 `README`s)

mnaamani:
> Looks good, the tests pass and the shasum matches the one from the PR (the image is not available, but I belive this is expected?)

Correct because the image is only pushed to docker-hub during a `push` so after the PR is merged to the branch (because secrets are only available to the repo)

mnaamani:
Will polish last few changes in another PR

PR #2298, opened 2021-03-29
CLI - initial update for olympia + query-node integration

Lezek123:
This PR introduces:
- changes required for the CLI to be compatible with Olympia runtime along with some account-management refactorization (previously part of https://github.com/Joystream/joystream/pull/2345) 
- CLI -> query-node integration for the purpose of displaying membership handles (currently) and most likely some more metadata that would be hard to retrieve directly from chain in the future (ie. openings/applications metadata)


mnaamani:
I'm guessing this old PR is obsolete now @Lezek123 ?

PR #2307, opened 2021-03-31
Antioch params

bwhm:


.github/workflows/run-network-tests.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       - name: Ensure tests are runnable
         run: yarn workspace network-tests build
       - name: Execute network tests
-        run: RUNTIME=babylon tests/network-tests/run-tests.sh full
+        run: RUNTIME=antioch tests/network-tests/run-tests.sh full
```
mnaamani, line 106: If you update from antioch branch you will find this job has actually been disabled. It is for testing runtime upgrade. So `RUNTIME=name` is the starting runtime which will be upgraded. So this change isn't really necessary.

node/src/chain_spec/mod.rs
```diff
@@ -294,13 +294,13 @@ pub fn testnet_genesis(
...
+                announcing_period: 4 * DAYS,
                 voting_period: 1 * DAYS,
                 revealing_period: 1 * DAYS,
-                council_size: 6,
-                candidacy_limit: 25,
+                council_size: 16,
+                candidacy_limit: 50,
                 min_council_stake: 1_000,
-                new_term_duration: 10 * DAYS,
+                new_term_duration: 1 * DAYS,
```
mnaamani, line 303: Not sure why it is so low. No proposals will have time to be voted on and executed in time. Is that the intention?

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -83,25 +83,25 @@ pub use proposal_types::{ProposalDetails, ProposalDetailsOf, ProposalEncoder};
...
 // candidacy_limit max value for the 'set election parameters' proposal
-const ELECTION_PARAMETERS_CANDIDACY_LIMIT_MAX_VALUE: u32 = 100;
+const ELECTION_PARAMETERS_CANDIDACY_LIMIT_MAX_VALUE: u32 = 200;
 // min_voting_stake min value for the 'set election parameters' proposal
 const ELECTION_PARAMETERS_MIN_STAKE_MIN_VALUE: u32 = 1;
 // min_voting_stake max value for the 'set election parameters' proposal
 const ELECTION_PARAMETERS_MIN_STAKE_MAX_VALUE: u32 = 100_000_u32;
 // new_term_duration min value for the 'set election parameters' proposal
-const ELECTION_PARAMETERS_NEW_TERM_DURATION_MIN_VALUE: u32 = 14400;
+const ELECTION_PARAMETERS_NEW_TERM_DURATION_MIN_VALUE: u32 = 1;
```
mnaamani, line 102: 1 block ?

.github/workflows/run-network-tests.yml
mnaamani: There is no harm in keeping this change, it will be required for the next runtime upgrade integration tests for sumer

node/src/chain_spec/mod.rs
bwhm: They can still vote throughout the announcing and voting period, so it shouldn't have any effect on that.
The purpose of this is to have very little downtime where you can't apply, nor vote for a CM.

An "actual" term will now last one week instead of two, and the incentives will be skewed heavily towards a CM's work instead of recurring rewards.

runtime-modules/proposals/codex/src/lib.rs
bwhm: Yeah, I considered doing this low value as the starting default.

.github/workflows/run-network-tests.yml
bwhm: I just copied from the last PR, where you corrected this...

If there are any benefit in reverting, I'll do that?

node/src/chain_spec/mod.rs
mnaamani: Okay that is very interesting approach :+1:

runtime-modules/proposals/codex/src/lib.rs
mnaamani: Makes sense now given the explanation here https://github.com/Joystream/joystream/pull/2307#discussion_r605058902

shamil-gadelshin (approved):
LGTM

PR #2309, opened 2021-03-31
Remove the ‘authority-discovery’ feature from the chain.

shamil-gadelshin:
#### Major changes
- AuthorityDiscovery was removed from the runtime configuration (as well as corresponding RPC)
- 'authority discovery' key was removed from the SessionKeys 
- 'authority discovery' worker was removed from the node configuration 

#### Affected crates
- runtime
- node 
- chain-spec-builder

#### Comments
- `chain-spec-builder` generates three keys now for the keystore
- consensus test for nodes works
- local "two-validators setup" run works
- no "QuorumFailed" messages in the log

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2347,7 +2347,6 @@ dependencies = [
  "pallet-transaction-payment",
  "pallet-transaction-payment-rpc",
  "parity-scale-codec",
- "sc-authority-discovery",
  "sc-basic-authorship",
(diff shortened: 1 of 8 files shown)


mnaamani (approved):
LGTM

mnaamani:
After some network testing it seems we do need this module, as we had issues with grandpa finalization.
We can revisit in future if required.

PR #2311, opened 2021-03-31
single instance of ApiPromise

mnaamani:
We only need a single instance of ApiPromise when retrying connections.


mnaamani:
This is wrong.. I should get some sleep

PR #2318, opened 2021-04-02
Fixes db benchmarkings

conectado:
When having the DB benchmarkings in the root-level workspace compilation failed due to version incompatibility between that module and  the DB benchmarking crate.

This PR separates the DB Benchmarking crate compilation from the rest of the workspace.


shamil-gadelshin (approved):
LGTM

PR #2319, opened 2021-04-02
Update benchmarks CI to use less steps to save time

conectado:
SSIA

### DIFF
--- a/scripts/generate-weights.sh
+++ b/scripts/generate-weights.sh
@@ -4,6 +4,9 @@
 
 SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 
+STEPS=${1:-50}
+REPEAT=${2:-20}
+
 benchmark() {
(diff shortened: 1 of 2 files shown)


scripts/generate-weights.sh
```diff
@@ -62,6 +65,5 @@ benchmark council
 benchmark referendum
 benchmark forum
 benchmark membership
-benchmark bounty
```
shamil-gadelshin: Why did you remove the bounty pallet?

shamil-gadelshin (changes requested):
Looks good. But it looks like you removed the bounty pallet from benchmarking.

scripts/generate-weights.sh
conectado: Sorry, a mess up I made with git

shamil-gadelshin (approved):

PR #2320, opened 2021-04-03
Antioch stack overflow fix

mnaamani:
Fix for the stack overflow encountered while testing, resulting from infinite recursion in impl of `core::fmt::Debug` trait for some types in content-directory https://github.com/Joystream/joystream/pull/2320/commits/8d0aed31399d3183d1ee8a300222f233de9281cc, and bumped runtime impl version.

Misc improvements:
Linter fixes to build with latest rust nightly.
Bumping cargo crates.
Update build scripts to use latest rust `nightly-2021-03-24`

Some thoughts,,

There are three types in the content directory that implement the Debug trait with infinite recursion.
One example:
https://github.com/Joystream/joystream/blob/656fa3412da81029650672051b28d25980a08d3f/runtime-modules/content-directory/src/operations.rs#L71

```rust
use sp_std::prelude::*; // where write! macro comes from

// This trait implementation is required by substrate because OperationType is a paramter type
// in a dispatchable call 'transaction' which we use to batch Operations.
impl<T: Trait> core::fmt::Debug for OperationType<T> {
    fn fmt(&self, formatter: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        write!(formatter, "OperationType {:?}", self)
    }
}
```

node log on crash when a block being imported had some transactions from content directory:
```
Apr 01 18:44:55 localhost joystream-node[2530]: thread 'tokio-runtime-worker' has overflowed its stack
Apr 01 18:44:55 localhost joystream-node[2530]: fatal runtime error: stack overflow
Apr 01 18:44:55 localhost systemd[1]: joystream-node.service: Main process exited, code=killed, status=6/ABRT
```

Output from valgrind 
```
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
==17173==
==17173== Process terminating with default action of signal 6 (SIGABRT)
==17173==    at 0x5DE6FB7: raise (raise.c:51)
==17173==    by 0x5DE8920: abort (abort.c:79)
==17173==    by 0x1D41099: std::sys::unix::abort_internal (mod.rs:237)
==17173==    by 0x1D3AA14: std::sys_common::util::abort (util.rs:19)
==17173==    by 0x1D404BF: std::sys::unix::stack_overflow::imp::signal_handler (stack_overflow.rs:106)
==17173==    by 0x55F997F: ??? (in /lib/x86_64-linux-gnu/libpthread-2.27.so)
==17173==    by 0x1D5AC04: core::fmt::write (mod.rs:1075)
```
The `mod.rs` is in the rust std lib which implements the `write!` macro.

What was unusual about this bug which was hard to track down is that it manifests only under the following conditions.
1 - Node is using native execution for block syncing (and the node should be syncing of course)
when running node with `--execution-syncing Wasm` the stack doesn't overflow or the code path is not executed.
2 - When node log level is set to info, debug or trace
when node was run with `--log warn` the stack doesn't overflow or the code path is not executed.

Its not clear why our previous network with older version of substrate we didn't suffer from this dormant bug. Why did this code path never get executed?

fmt::Debug implementation as far as I can tell is primarily used to "serialize" the object to a string for the purpose of a comparison/assertions in unit tests. So they are not adequately implemented for this purpose. I didn't have the stamina to do it, and I don't think the implementation has any affect on stored state. Please confirm this @shamil-gadelshin ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -950,7 +950,7 @@ dependencies = [
  "cfg-if 1.0.0",
  "crossbeam-utils 0.8.3",
  "lazy_static",
- "memoffset 0.6.2",
+ "memoffset 0.6.3",
  "scopeguard 1.1.0",
 ]
 
@@ -1668,7 +1668,7 @@ dependencies = [
  "lazy_static",
  "log",
  "parking_lot 0.9.0",
- "pin-project 0.4.27",
+ "pin-project 0.4.28",
  "serde",
  "serde_json",
 ]
@@ -1788,7 +1788,7 @@ dependencies = [
  "bytes 0.5.6",
  "futures 0.3.13",
  "memchr",
- "pin-project 0.4.27",
+ "pin-project 0.4.28",
 ]
 
 [[package]]
@@ -2463,9 +2463,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.49"
+version = "0.3.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc15e39392125075f60c95ba416f5381ff6c3a948ff02ab12464715adf56c821"
+checksum = "2d99f9e3e84b8f67f846ef5b4cbbc3b1c29f6c759fcbce6f01aa0e73d932a24c"
 dependencies = [
  "wasm-bindgen",
 ]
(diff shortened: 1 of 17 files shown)


shamil-gadelshin (approved):
Tested locally - the bug was not reproduced. LGTM

PR #2323, opened 2021-04-05
Enable hydra-typegen

metmirr:
It enables hydra-typegen and removes generated type assets. Hydra-typegen fetch chain metadata from `wss://staging-3.joystream.app/rpc`.


mnaamani (approved):
Great. Its working as expected. I do think however that having the build depend on an external endpoint is brittle. I think we need to do something similar to typegen files in the `@joystream/types` package, where we check-in the typedefs.json file into repo and make typegen step manual.

PR #2324, opened 2021-04-05
Update proposal stake requirements in pioneer to match runtime

mnaamani:
Required stake for creating proposals was update in https://github.com/Joystream/joystream/pull/2209
But we forgot to update pioneer to reflect the changes.


bwhm (approved):
Tested, and works well :)

PR #2327, opened 2021-04-06
pioneer: reduce max batch size in staking payouts from 40 to 30

mnaamani:
to fit into max block weigth limit


bwhm (approved):
Solved the issue!

PR #2336, opened 2021-04-08
Fix blog state bloat

conectado:
# Fix blog pallet for #2196

Fixes the state bloat problem for the blog in a similar manner to what we did for [`proposal_discussions` and `forum`](https://github.com/Joystream/joystream/issues/2196#issuecomment-802111952).

The way it works in the case of the blog is the following:
* Replies to post can be editable/non-editable.
* Editable replies take space in storage and a fee, this is not the case for non-editable replies.
* Editable replies can be deleted only by the author until the `ReplyLifetime` has gone by.
* If a reply has been editable for more than `ReplyLifetime` it can be erased by anyone.
* A reply to a reply only requires that the parent reply has been created at some point not that it's present in storage.

Companion handbook PR: Joystream/handbook#41

### DIFF
--- a/runtime-modules/blog/src/lib.rs
+++ b/runtime-modules/blog/src/lib.rs
@@ -50,6 +50,7 @@ use sp_runtime::{
     traits::{AccountIdConversion, Hash, MaybeSerialize, Member, Saturating},
     ModuleId,
 };
+use sp_std::collections::btree_map::BTreeMap;
 use sp_std::prelude::*;
 
 mod benchmarking;
@@ -656,7 +657,7 @@ decl_module! {
         /// `O (R)` where
         /// - R is the number of replies to be deleted
         /// - DB:
-        ///    - O(1) doesn't depend on the state or parameters
+        ///    - O(R)
         /// # </weight>
         #[weight = BlogWeightInfo::<T, I>::delete_replies(replies.len().saturated_into())]
         pub fn delete_replies(
@@ -667,6 +668,7 @@ decl_module! {
             let account_id = Self::ensure_valid_participant(origin, participant_id)?;
 
             let mut erase_replies = Vec::new();
+            let mut pay_off_map = BTreeMap::new();
             for ReplyToDelete { post_id, reply_id, hide } in replies {
                 // Ensure post with given id exists
                 let post = Self::ensure_post_exists(post_id)?;
@@ -688,7 +690,18 @@ decl_module! {
                     ensure!(!hide, Error::<T, I>::ReplyOwnershipError);
                 }
 
+                *pay_off_map.entry(post_id).or_default() += reply.cleanup_pay_off;
                 erase_replies.push((post_id, reply_id, reply.cleanup_pay_off, hide));
+            }
+
+            for (post_id, post_deposit) in pay_off_map.into_iter() {
+                ensure!(
+                    Balances::<T>::usable_balance(
+                        &Self::get_treasury_account(post_id)
+                    ) >= post_deposit,
+                    Error::<T, I>::InsufficientBalanceInPostAccount
+
+                );
             }
 
             //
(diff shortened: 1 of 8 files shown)


runtime-modules/blog/src/lib.rs
```diff
@@ -578,36 +637,117 @@ decl_module! {
...
+
+                erase_replies.push((post_id, reply_id, reply.cleanup_pay_off, hide));
+            }
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            for (post_id, reply_id, cleanup_pay_off, hide) in erase_replies {
+                Self::pay_off(post_id, cleanup_pay_off, &account_id)?;
```
shamil-gadelshin, line 712: It looks like it's possible to rewrite it as a single call (sum pay-off). Each separate `pay_off` call in the loop could issue an error and the state would be corrupted. However, a single call would leave consistent data even in the case of an error.

runtime-modules/blog/src/lib.rs
```diff
@@ -578,36 +637,117 @@ decl_module! {
...
+                let post = Self::ensure_post_exists(post_id)?;
+
+                // Ensure post unlocked, so mutations can be performed
+                Self::ensure_post_unlocked(&post)?;
+
+                // Ensure reply with given id exists
+                let reply = Self::ensure_reply_exists(post_id, reply_id)?;
+
+                // Ensure reply -> owner relation exists if post lifetime hasn't ran out
+                if (frame_system::Module::<T>::block_number() - reply.last_edited) <
```
shamil-gadelshin: Please use the saturating semantics here.

runtime-modules/blog/src/lib.rs
```diff
@@ -578,36 +637,117 @@ decl_module! {
...
+        /// ## Weight
+        /// `O (1)` doesn't depends on the state or parameters
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = BlogWeightInfo::<T, I>::delete_reply()]
+        pub fn delete_replies(
+            origin,
+            participant_id: ParticipantId<T>,
+            replies: Vec<(PostId, T::ReplyId, bool)>,
```
shamil-gadelshin: Could we use here some struct with named and commented fields instead of the tuple?

runtime-modules/blog/src/lib.rs
```diff
@@ -578,36 +637,117 @@ decl_module! {
             Ok(())
         }
 
+        /// Remove reply from storage
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)` doesn't depends on the state or parameters
```
shamil-gadelshin: Is the weight comment correct here?

runtime-modules/blog/src/mock.rs
```diff
@@ -222,17 +222,22 @@ impl membership::WeightInfo for Weights {
 parameter_types! {
     pub const PostsMaxNumber: u64 = 20;
     pub const RepliesMaxNumber: u64 = 100;
+    pub const ReplyDeposit: u64 = 500;
+    pub const BlogModuleId: ModuleId = ModuleId(*b"m00:blog"); // module : blog
+    pub const ReplyLifetime: <Runtime as frame_system::Trait>::BlockNumber = 10;
 }
```
shamil-gadelshin, line 229: Don't we need to change the AccountId type to the u128?

shamil-gadelshin (changes requested):
Looks ok, however, there are several issues to fix.

runtime-modules/blog/src/lib.rs
conectado: The problem with that is that the pay offs might be in different post accounts, we could however do first a loop doing the pay off for each deleted reply and then this loop but that would have a performance cost and if it errors in any of the pay off after the first it would still be left in an inconsistent state.

shamil-gadelshin: Adding to our conversation: it makes sense then to check separately that all post accounts have enough funds to make pay-offs (prior to any state mutation).

runtime-modules/blog/src/lib.rs
conectado: It was correct given the weight function used but it seemed weird so I re-run the benchmarks and now it depends on the number of replies.

shamil-gadelshin (changes requested):
Almost done. There is a couple of issues:
- please, implement a check for an available sum in accounts before the mutation in the delete_replies()
- Why the benchmarks DB-operations don't depend on the reply count? You mentioned "DB: O(1)" in the weight comment of the extrinsic and the weight file confirms that. Shouldn't we increase the maximum reply number in the benchmarks?

conectado:
>     * please, implement a check for an available sum in accounts before the mutation in the delete_replies()

Do we want such a check? If a reply is in storage its deposit must exists in the post account. If there's no funds for the reply deletion we are already in an inconsistent state.

shamil-gadelshin (approved):
LGTM

PR #2342, opened 2021-04-10
multi-arch joystream/node build with buildx

mnaamani:
Attempting to build multi-arch docker image with `docker buildx build ...`
The build for the paltform happens concurrently and that seems to be too much for the host machine as it is running out of space:
```
#21 4680.7 error: failed to write /joystream/target/release/deps/rmetaW6qF1Z/lib.rmeta: No space left on device (os error 28)
```

Will need to find a way to do the builds serially and make sure storage space is recovered between builds..? Or take the more manual approach to achieve this using `docker manifest ...`

Using https://github.com/easimon/maximize-build-space action to increase freespace for build to complete.


mnaamani:
Trying solution from https://github.community/t/worker-running-out-of-disk-space/18245
Still running out of space..

mnaamani:
Seems the `/mnt` location actually has less space than expected. But there is only 20GB available to work with on `/` and it doesn't seem sufficient for three parallel builds...

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   64G   20G  77% /
devtmpfs        3.4G     0  3.4G   0% /dev
tmpfs           3.4G   12K  3.4G   1% /dev/shm
tmpfs           696M  1.1M  695M   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.4G     0  3.4G   0% /sys/fs/cgroup
/dev/loop0       56M   56M     0 100% /snap/core18/1997
/dev/loop1       71M   71M     0 100% /snap/lxd/19647
/dev/loop2       33M   33M     0 100% /snap/snapd/11588
/dev/sda15      105M  7.8M   97M   8% /boot/efi
/dev/sdb1        14G  4.1G  9.0G  32% /mnt
```

mnaamani:
Made progress by switching to use official docker `rust` image which supports the architectures we are building for.
During build of `arm/v7` though there seems to be issues downloading from `crates.io`

```
 #22 [linux/arm/v7 builder 3/3] RUN BUILD_DUMMY_WASM_BINARY=1 cargo clippy --release --all -- -D warnings &&     cargo test --release --all &&     cargo build --release
#22 sha256:c33546a4717450b9bb6f152292a81c7bf15b11d4239caace2a49c6f2670752d4
#22 10.91     Updating crates.io index
#22 11.31 warning: spurious network error (2 tries remaining): could not read directory '/usr/local/cargo/registry/index/github.com-1285ae84e5963aae/.git//refs': Value too large for defined data type; class=Os (2)
```
Will try to re-start jobs (perhaps it is just spurious network issue..) If it continues to fail will just drop that architecture.

mnaamani:
Came across this nice article: https://dev.to/jdrouet/building-multi-arch-docker-images-without-using-qemu-322c

Which suggests an alternative approach .. building on remote machines for each architecture. 
cc: @ahhda

PR #2343, opened 2021-04-12
Pioneer - add Operations group

Lezek123:
Adds `Operations` working group support in `joy-roles`

### DIFF
--- a/pioneer/packages/joy-roles/src/tabs/WorkingGroup.controller.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/WorkingGroup.controller.tsx
@@ -38,6 +38,7 @@ export class WorkingGroupsController extends Controller<State, ITransport> {
     );
 
     this.setState(newState);
+    this.dispatch();
   }
 }
 
--- a/pioneer/packages/joy-roles/src/tabs/WorkingGroup.controller.tsx
+++ b/pioneer/packages/joy-roles/src/tabs/WorkingGroup.controller.tsx
@@ -4,45 +4,40 @@ import { Controller } from '@polkadot/joy-utils/react/helpers';
 import { View } from '@polkadot/joy-utils/react/hocs';
 
 import { ITransport } from '../transport';
+import { AvailableGroups } from '../working_groups';
 
-import { ContentCurators,
-  WorkingGroupMembership,
-  StorageProviders } from './WorkingGroup';
+import { WorkingGroupMembership,
+  ContentCurators,
+  StorageProviders,
+  OperationsGroup } from './WorkingGroup';
 
 import styled from 'styled-components';
-import { normalizeError } from '@polkadot/joy-utils/functions/misc';
 
 type State = {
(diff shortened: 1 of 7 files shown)


pioneer/packages/joy-roles/src/tabs/WorkingGroup.controller.tsx
```diff
@@ -4,45 +4,40 @@ import { Controller } from '@polkadot/joy-utils/react/helpers';
...
+    const newState: Partial<State> = {};
 
-  getStorageGroup () {
-    this.transport.storageGroup()
-      .then((value: WorkingGroupMembership) => {
-        this.setState({ storageProviders: value });
-        this.dispatch();
+    await Promise.all(
+      AvailableGroups.map(async (group) => {
+        newState[group] = await this.transport.groupOverview(group);
```
yourheropaul, line 36: I know I'm not invited to this party, so do disregard this if it's inappropriate! I happen to be the original author of this controller pattern for this project and I saw this PR by chance.

Should this not be a `Promise.all()` sort of setup? Otherwise this loop will fetch the group overviews in series, not in parallel, which I presume is still quite slow.

Lezek123: Hello Paul, thanks for the comment, but I'm not sure what you mean. This is a `Promise.all` setup and the queries should be executed in parallel, since I'm using `await Promise.all` on an array of promises created from `AvailableGroups`.

yourheropaul: Maybe I've misunderstood what's going on here (it's been a while since I looked at this codebase) but isn't the mapping of `AvailableGroups` assigning values in the variable `newState` but not returning a Promise for `Promise.all` to execute?

In other words, don't you want

```
const groups = await Promise.all( AvailableGroups.map(group => this.transport.groupOverview(group)));
// ... add each group to newState
```

? Otherwise, isn't `AvailableGroups.map` waiting for the promise to resolve on each iteration?

Lezek123: I'm passing an async function to `map`. An async function returns a promise (by definition) and all the promises returned by `map` are awaited on with `Promise.all`, so they are all executed in parallel.

You can try it out:
```
const a = () => new Promise(
  (resolve, reject) => {
    setTimeout(() => resolve(true), 1000)
  }
)

async function b () {
  const start = Date.now()
  await Promise.all(
    [1,2,3].map(async () => {
      await a()
    })
  )
  console.log(Date.now() - start)
}

await b()
```

yourheropaul: I did not know about that feature! Thanks, and sorry to bother you.

mnaamani:
Looks good. Only thing is now the "Lead" shows up twice.

Screen Shot 2021-04-16 at 1 20 37 PM

PR #2344, opened 2021-04-12
@joystream/types - createType optimalization

Lezek123:
While working on Sumer CLI I noticed that using the `createType` function from `@joystream/types` (which allows creating types in a typesafe way) slows down the IDE typechecking quite noticably.

Using trial-and-error I was able to come up with an approach that's slightly less processor-intensive.


mnaamani:
I can't say that I was able to confirm an improvement, but you certainly spend more time in typescript code so I will take your word for it. As far as I can tell the refactor is correct and even a bit clearer for us mere mortals to parse :)

Issue #2303, opened 2021-03-30
Olympia membership input schema


bedeho:
# Background

The purpose of this issue is to serve as a discussion thread for the state of the current `Olympia` input schemas.

Here is a fully rendition of the current schema, lifted from 

https://github.com/Lezek123/substrate-runtime-joystream/blob/olympia-integration-tests/query-node/schema.graphql

```graphql
enum Network {
  BABYLON
  ALEXANDRIA
  ROME
  OLYMPIA
}

type Block @entity {
  "Block number as a string"
  id: ID!
  block: Int!
  executedAt: DateTime!
  network: Network!
}

enum MembershipEntryMethod {
  PAID
  INVITED
  GENESIS
}

"Stored information about a registered user"
type Membership @entity {
  "MemberId: runtime identifier for a user"
  id: ID!

  "The unique handle chosen by member"
  handle: String! @unique @fulltext(query: "membersByHandle")

  "Member's name"
  name: String

  "A Url to member's Avatar image"
  avatarUri: String

  "Short text chosen by member to share information about themselves"
  about: String

  "Member's controller account id"
  controllerAccount: String!

  "Member's root account id"
  rootAccount: String!

  "Blocknumber when member was registered"
  registeredAtBlock: Block!

  "Timestamp when member was registered"
  registeredAtTime: DateTime!

  "How the member was registered"
  entry: MembershipEntryMethod!

  "Whether member has been verified by membership working group."
  isVerified: Boolean!

  "Staking accounts bounded to membership."
  boundAccounts: [String!]

  "Current count of invites left to send."
  inviteCount: Int!

  "All members invited by this member."
  invitees: [Membership!] @derivedFrom(field: "invitedBy")

  "A member that invited this member (if any)"
  invitedBy: Membership

  "All members referred by this member"
  referredMembers: [Membership!] @derivedFrom(field: "referredBy")

  "A member that referred this member (if any)"
  referredBy: Membership
}

type MembershipSystem @entity {
  "Initial invitation count of a new member."
  defaultInviteCount: Int!

  "Current price to buy a membership."
  membershipPrice: BigInt!

  "Amount of tokens diverted to invitor."
  referralCut: BigInt!

  "The initial, locked, balance credited to controller account of invitee."
  invitedInitialBalance: BigInt!
}
```
# Comments

[...]

PR #2345, opened 2021-04-14
Olympia - working groups schemas, mappings and tests

Lezek123:
Based on https://github.com/Joystream/joystream/pull/2276 in order to include the current membership schemas.

The input schema to review can be found here: https://github.com/Lezek123/substrate-runtime-joystream/blob/working-groups-schemas/query-node/schema.graphql

The changes compared to https://github.com/Joystream/joystream/pull/2276 are in commits:
- https://github.com/Joystream/joystream/commit/9c0f53c59ec04924dc9cbbdbb3656607b34cbca9
- https://github.com/Joystream/joystream/pull/2345/commits/faa59096c66bfabb778f4fa6a3d628170f8a5f3b

#### Working group status

Currently in the runtime we have `set_status_text` extrinsic and `StatusTextChanged` event, but it can also be used to set any other working group metadata (including group description etc.)

The name of it could be therefore a little misleading, so my proposition would be to rename it to `set_group_metadata`.
The associated protobuff message would be:
```
message WorkingGroupMetadata {
  optional string description = 1; // Working group description (md-formatted)
  optional string about = 2; // Working group about text (md-formatted)
  optional string status = 3; // Current working group status (expected to be 1-3 words)
  optional string status_message = 4; // Short status message
}
```

#### Missing events

- We have `WorkerLeft` event declared in the runtime, but it's never actually emitted. I think it would be benefitial to emit it, so that we could have an event associated with both `WorkerStatusTerminated` and `WorkerStatusLeft` (see the schema) and also for general event listing purposes.
- Currently there are no events in the runtime associated with rewarding workers, so I created a dummy `WorkerPayoutEvent` in the schema. This will most likely change in the near future.

#### Examples of aggregating data

I skipped some aggreated data fields (like worker total reward), because I wasn't sure what's the approach we should to take for those.

Here are a few examples of how those values can be calculated:

- **Total JOY currently staked by a member:** sum of `Membership.roles[i].stake`
- **Total JOY earned in the past x days by member:** sum of `Membership.roles[i].payouts.recieved` where `payout.time <= x` (this it just an example, it will probably change soon, as payout events are not currently supported)
- **Total workers ever hired in group:** count of `WorkingGroup.workers`
- **Total workers ever fired in group:** count of `WorkingGroup.workers` where `worker.status` is `WorkerStatusTerminated`
- **Working group avg stake:** sum of `WorkingGroup.workers[].stake` / count of `WorkingGroup.workers`

Of course in some cases it may turn out that an additional aggregation field may be benefitial for reducing the load on the database while querying larger datasets. In this case we can consider adding it.

#### Immutable opening / application data

Currently from the runtime perspective both the opening and the application data is immutable (in a sense that it cannot be edited), so I didn't include a separate "snapshot" of this data events like `OpeningAdded` / `AppliedOnOpening` (only the `opening` / `application` relation is included) to avoid redundancy. In case we consider making openings / applications editable this should probably be changed.

#### Functionality not supported by the runtime

There are a few concepts in the designs that are not currently supported by the runtime, therefore are also not part of the input schema:
- Working group debt
- Upcoming openings
- Opening applicants limit - perhaps it could be part of the opening metadata, but the runtime will not enforce it anyway


bedeho (commented):
This review is only about the input schema.

# General

1. First off, this looks really good, nice job! 🚀 

2. This is starting to get really big, so it would be great to split up the schema, something we now support. In particular segmenting the events from the other core entities would be great, but even beyond this I think we need to have things split. Would one entity per file work? Could one entity/enum/interface/union per class even work? Let me know what you think.

3.  Events are getting segmented into top level topics using interfaces which are largely empty, like `WorkingGroupEvents`,  `MembershipEvent`. This has two problems. First, we may plausibly like to query across these boundries in the future, but this is no longer possible in single queries with built in filtering, ordering, pagination. Second problem is that the categories are not even obvious boundries. Say I get fired, this is happening both to a worker and to a member, so is it a working group or membership event? Semantically its not clear. Most importantly perhaps is that those categories are not doing any real work, so there is no upside either. I suggest we drop them and just derive all from one root  `Event` entity. When a given use case wants to only consider a subset of event types, it can filter on the `type` field value.

4. Should `Event.id` also include network identifier in case we want to cache events from old networks? This seems plausible if we want the query node to hold history across distinct genesis blocks.

5. I see you are using `AtBlock` and `AtTime` as a way to capture when somethign happened. This may have even been something we agreed upon at some point? One issue with this is that it does not capture networks for the block case. The other issue is that having a distinct relationship with a `Block` entity would avoid having three separate fields in each case, and with the introduction of filtering through relationships, we can now do all the filtering we want on the original entity still. The only downside is that when setting up these inside mappings, you will have ot have extra logic for correctly creating exactly one instance of a block, even though it may have multiple such events in it across different mappings.

Please share your thoughts here, what should we do?

6. `WorkingGroup::name` seems like it should be replaced with an enum for working groups, do we have this?

7. You have created types for the structure of the application form, but do we really need this? The main benefit is that the query system becomes aware of it and allows us to key off it, however, I can't imagine that use case. The downside is that if the format changes, then the query nodes have to be updated, which is disruptive and complex. If this is the fully cost-benefit, then I say we dont need it. What do you think?

# Examples of aggregating data

I think in all of these cases, this can be done by doing aggregate queries (scheduled here https://github.com/Joystream/hydra/issues/195) across events

See events section of https://github.com/Joystream/joystream/issues/2303.

This is a much more powerful approach, as it allows the controlling independent parameters like block ranges or subsets of members. 

The downside is that you could not directly filter or order on those values when qerying members or workers, but that seems of limited importance. This information seems important to me only in cases where you are dealing with one or very few entities, so you can manually fire off multiple queries.

# Immutable opening / application data

I agree. Even if it was mutable, its not obvious that having all snapshots available is a super high priority anyway, the last one would be sufficient for now. 

# Functionality not supported by the runtime

There are a few concepts in the designs that are not currently supported by the runtime, therefore are also not part of the input schema:

- Working group debt: This is implicit, you just add up all the owed reward to all workers, that is the debt.
- Upcoming openings: this should be set as a message over the working group status channel `set_group_metadata`. So this would just be a distinct message one could pass. I guess you would have possibly want separate message for creating, editing and droppings an upcoming opening.
- Opening applicants limit: where is this from? this would be a bad thing to have, so we wont introduce this.

Lezek123:
**Ad.2.**

I didn't realize this is already supported, that's great to know!
We usually split code by runtime modules (ie. `memberships`, `workingGroups`, `contentDirectory` etc.), so I was thinking about taking a similar approach, perhaps also separating events from other entities, so we could have:
- `common.graphql` - ie. network enum, event enum, common entities / interfaces etc.
- `storage.graphql` - ie. `DataObject` and related enums / entities / variants / unions
- `membership.graphql` - membership base entities (`Membership`, `MembershipMetadata`...)
- `membershipEvents.graphql` - membership events
- `workingGroups.graphql` - working groups base entities (`Opening`, `Application`, `Worker`...)
- `workingGroupsEvents.graphql` - working groups events

Having each entity / interface etc. in a separate file seems like could be a bit too much, especially since subdirectories don't seem to be supported yet, so we would end up with a gigantic list of files in one directory.

**Ad.3.**

That's true, This was a bit of an experiment, I thought it may make it easier to query a specific group of events to display ie. in the `Last events` box on the `working groups` page in Pioneer, but since the client needs to be aware of how to parse each individual event anyway this probably isn't that much helpful. I'll make a single interface for all events.

**Ad.4.**

I agree, it seems like it could be useful.

**Ad.5.**

I think we agreed for it to be a temporary solution in https://github.com/Joystream/joystream/issues/2303, of course I can change it. Making sure a block is unique should not be a problem, I already use similiar logic for creating a single `MembershipSnaphot` per block.

**Ad.6.**

Currently I was using api module name as `WorkingGroup.name` which allowed taking some shortcuts. I was also using a relation to a working group entity anywhere I needed to associate some other entity with a given group.

An alternative could be to use an enum like `WorkingGroupName` for `WorkingGroup.name` property, perhaps to make schema a bit more strict and descriptive, but it seems like using a relation to a working group where we can instead of enum value would still be better for building more complex queries.

**Ad.7.**

I thought it could be a good addition and make it much easier for the frontend to process, connect and query data, for example,  we could use multiple different queries like:
- all application question-answer pairs for given application
- all answers (and related members) to question X in a given opening
- answers to similar questions across different openings (ie. questions with a keyword)
- answers to all application questions a member has ever aswered together with related openings
- etc.

All that without any additional client-side data processing required.

Changing metadata format for membership already requires query nodes to updates so I assumed the same for other types of metadata.

Should we just store a binary then and force the frontend apps to decode and process this part of the metadata by themselves?

Lezek123:
### A short summary of the current state of the mappings:

At this moment the mappings cover all existing working groups runtime events.
The tests cover everything except worker payouts and  `WorkerExited` event, since testing those requires a custom runtime build with adjusted `MinUnstakingPeriodLimit` and `RewardPeriod`. I tested those locally though and didn't encounter any issues.

**Not yet supported: Variant relationships and cross-fitering**

Hydra v3 has not yet been released, so these features are currently unavailable (related PRs: https://github.com/Joystream/hydra/pull/360 and https://github.com/Joystream/hydra/pull/365).

After Hydra v3 is released, all the variants like:
```
type OpeningStatusCancelled @variant {
  # TODO: Variant relationships
  openingCancelledEventId: ID!
}
```
Will be adequately changed to:
```
type OpeningStatusCancelled @variant {
  openingCancelledEvent: OpeningCancelledEvent!
}
```

Also some obvious queries like querying events by `event.inBlock.blockNumber` should finally become possible.

**Not yet supported: `Event` interface**

Due to a bug in Hydra (https://github.com/Joystream/hydra/issues/359) the `Event` interface currently cannot be used, so it's impossible to make a generic `events` query like:
```
  events {
    __typename 
    ... on WorkingGroupOpeningCreatedEvent {
      ...
    }
    ... on WorkingGroupOpeningCancelledEvent {
      ...
    }
  }
```

This is expected to be supported after the issue is resolved

**Protobufjs and always optional metadata**

After a short discussion with @mnaamani I decided to switch to `protobufjs` for compiling the protobuf messages, as it seems much easier to work with, ie.:
- It includes validation for plain json object
- It allows encoding plain json objects and has correct typings for `null`/`undefined` value cases
- It can be a standard `package.json` dependancy (so we're not forcing the user to separately install `protoc`)

I also decided to make all fields in protobuf messages optional, as using `required` fields is considered a bad practice. Although both `optional` and `required` fields are encoded exactly in the same way - adding a `required` field to a message would prevent us from using new compiled messages to decode previously valid messages. It also makes the messages less reusable in case we only care about some, but not all fields in the message (like in a typical atomic update scenario).

I was considering then adding some default values for fields I made required in the schema itself (for example: `WorkingGroupOpeningMetadata.description`), but this seemed often counter-intuitive and problematic (it wasn't clear whether the field value was the default or was explicitly set by the user. It also wasn't clear if we should set everything to default in case any "required" field was not provided or just override this single field), so I decided to just made those fields optional in the schema too. I think the frontend should be able to handle this accordingly (for example, by displaying a simple message like `Description not set` or not showing some parts of the UI).

I also added conversion of falsy values like `''` or `0` to `null` in the query node entities in case those values are not considered valid (ie. `hiringLimit` of `0` is not valid, therefore it's converted to `null`). This seems like a cleaner approach as it also allows clear "unsetting" of optional values like group status or member `name`/`about` without any major changes in the messages.

**The problem with `LeaderSet` event**

Due to the fact that `LeaderSet` event is emitted before `OpeningFilled`, so at the time the worker is not yet available in the database - the `worker` field had to be made be optional, but it should be the case that it is always set anyway, because `OpeningFilled` and `LeaderSet` events are emitted in the same extrinsic.

Lezek123:
### Hydra v3 migration

I performed an initial migration of the mappings to Hydra v3 locally and I noticed a few issues regarding the graphql server generation process and the execution of certain queries.

**Optional relationships**
See: https://github.com/Joystream/hydra/issues/391

**Nested relationship filtering**

Because all the events currently share a specific relationship schema like:
`SpecificEventData (ie. MembershipBoughtEvent) -> GenericEventData (Event) -> BlockData (Block)` 

Without nested relationships filtering we cannot filter specific events by block data, for example:

```
query {
	membershipBoughtEvents(where: { event: { inBlock: { number_gt: 10 } } } ) {
		event {
			inBlock {
				number
			}
		}
	}
}
```

Currently a query like this will result in:
```
 "errors": [
    {
      "message": "column event.undefined does not exist",
    # ...
```

This is just one example, but I think having the ability to filter by nested relationship fields would be very convinient in general, otherwise we may need to modify the input schema to faciliate some of the queries.

**Ordering by relations**

Currently in Hydra v3 there is no way to use a related entity field for ordering.

This could also be useful when querying events, because usually we'd like to order them by block number and then index in block DESC. We can use `createdAt` instead of block number, because the value is based on a block timestamp (so the ordering should be the same), but we cannot access `index` for ordering, as it is part of a related entity (for example: `membershipBoughtEvent.event.indexInBlock`).

I'm not sure how feasible it is to implement this in Hydra, so as an alternative we could either have a field in each specific event that is dedicated just for ordering purposes (ie. `blockNumber * 1000000 + indexInBlock`) or perhaps remove the `Event` relationship entirely and specify all the generic event fileds (`indexInBlock`, `inBlock`, `inExtrinsic`) as part of the specific event entities instead (this could also help with the previous issue, but it would mean a lot of redundant columns in SQL tables and will also prevent creating relationships to events where the event type is not limited, like `block.events`)


**Filtering by variant type (ie. `worker.status`)**

In the current schema it became sort of a standard to have an entity status represented by algebraic type, for example:

```
type WorkerStatusActive @variant {
  # No additional information needed
  _phantom: Int
}

type WorkerStatusLeft @variant {
  "Related event emitted on leaving initialization"
  workerStartedLeavingEvent: WorkerStartedLeavingEvent!

  "Related event emitted (and set) when the unstaking period is finished"
  workerExitedEvent: WorkerExitedEvent
}

type WorkerStatusTerminated @variant {
  "Related event emitted on worker termination"
  terminatedWorkerEvent: TerminatedWorkerEvent!
}

union WorkerStatus = WorkerStatusActive | WorkerStatusLeft | WorkerStatusTerminated

type Worker @entity {
  "Current worker status"
  status: WorkerStatus!
  # ...
}
```

This way it's possible to, for example, include additional event(s) information based on status (ie. here in case the `Worker` was terminated, the frontend can additionally fetch `reason`, `blockNumber` and `blockTimestamp` from the related `TerminatedWorkerEvent`).

Unfortunately, currently in this case it doesn't seem to be possible to create a query that would filter workers by a specific status, for example:

```
query {
	workers(where: { status_json: { __typename_eq: "WorkerStatusActive" } }) {
		status {
			__typename
		}
	}
}
```

Moreover, it also doesn't seem to be possible to filter by other variant fields, even though it's mentioned in the documentation (https://dzhelezov.gitbook.io/hydra/docs/schema-spec/variant-types).

For example the following query:
```
query {
	workingGroupApplications {
		status {
			...on ApplicationStatusAccepted {
				openingFilledEvent {
					id
					createdAt
				}
			}
		}
	}
}
```
Returns:
```
{
  "data": {
    "workingGroupApplications": [
      {
        "status": {
          "openingFilledEvent": {
            "id": "-69-UePHPb",
            "createdAt": "2021-05-11T10:41:18.000Z"
          }
        }
      },
      # ...
```
But with filters like:
```
  workingGroupApplications(where: { status_json: { openingFilledEvent: { createdAt_gt: "2021-05-10T00:00:00.000Z" } } }) {
```
Or:
```
  workingGroupApplications(where: { status_json: { openingFilledEvent: { id_eq: "-69-UePHPb" } } }) {
```
The query returns no results.

**Deeply nested variant types**

There seems to be an issue with querying deeply nested varaint types, let's assume we have a following schema:
```
type ApplicationStatusPending @variant {
  # No additional information needed
  _phantom: Int
}

type ApplicationStatusAccepted @variant {
  "Related OpeningFilled event"
  openingFilledEvent: OpeningFilledEvent!
}

type ApplicationStatusRejected @variant {
  "Related OpeningFilled event"
  openingFilledEvent: OpeningFilledEvent!
}

type ApplicationStatusCancelled @variant {
  "Related OpeningCanceled event"
  openingCanceledEvent: OpeningCanceledEvent!
}

type ApplicationStatusWithdrawn @variant {
  "Related ApplicationWithdrawn event"
  applicationWithdrawnEvent: ApplicationWithdrawnEvent!
}

union WorkingGroupApplicationStatus =
    ApplicationStatusPending
  | ApplicationStatusAccepted
  | ApplicationStatusRejected
  | ApplicationStatusWithdrawn
  | ApplicationStatusCancelled

type WorkingGroupApplication @entity {
  "Application id ({workingGroupName}-{applicationId})"
  id: ID!

  "Related working group opening"
  opening: WorkingGroupOpening!

  "Current application status"
  status: WorkingGroupApplicationStatus!
  # ...

type WorkingGroupOpening @entity {
  "Opening id ({workingGroupName}-{openingId})"
  id: ID!

  "List of opening applications"
  applications: [WorkingGroupApplication!] @derivedFrom(field: "opening")

  # ...
}
```

When I'm executing the following query:
```
query {
	workingGroupApplications {
		id
		opening {
			id
		}
		status {
			__typename
			...on ApplicationStatusAccepted {
				openingFilledEvent {
				 id
				}
			}
		}
	}
}
```

I'm getting some results like:
```
{
  "data": {
    "workingGroupApplications": [
      {
        "id": "contentDirectoryWorkingGroup-0",
        "opening": {
          "id": "contentDirectoryWorkingGroup-0"
        },
        "status": {
          "__typename": "ApplicationStatusAccepted",
          "openingFilledEvent": {
            "id": "-69-UePHPb"
          }
        }
      },
     # ...
```

But if I add one more level of nesting, by querying applications in a specific opening:
```
query {
	workingGroupOpeningByUniqueInput(where: { id: "contentDirectoryWorkingGroup-0" }) {
		applications {
			id
			opening {
				id
			}
			status {
				__typename
				...on ApplicationStatusAccepted {
					openingFilledEvent {
						id
					}
				}
			}
		}
	}
}
```

I get:
```
{
  "data": {
    "workingGroupOpeningByUniqueInput": {
      "applications": [
        {
          "id": "contentDirectoryWorkingGroup-0",
          "opening": {
            "id": "contentDirectoryWorkingGroup-0"
          },
          "status": {
            "__typename": "ApplicationStatusAccepted",
            "openingFilledEvent": null
          }
        }
      ]
    }
  }
}
```
_Notice the `openingFilledEvent` is `null` in this case_

dzhelezov:
> ### Hydra v3 migration
> **Optional relationships**
> See: [Joystream/hydra#391](https://github.com/Joystream/hydra/issues/391)

This is certainly a bug, needs a fix (though it seems it requires a bit of work to make things done properly)

> **Nested relationship filtering**
> **Ordering by relations**


Both options are indeed nice to have.  I have the following reservations though:

- Performance may be poor for a moderately large database unless extra care is taken. Even for million blocks and 10M events joins might be expensive compared to a flat structure. 
- The implementation complexity is significant 

As you said, an alternative would to flatten the schema at the expense of data duplication (which I personally think is a reasonable trade-off).

> 
> **Filtering by variant type (ie. `worker.status`)**
> 

This is fairly easy to implement. For the time being, one can manually add a `type` field and filter by it.

> 
> Moreover, it also doesn't seem to be possible to filter by other variant fields, even though it's mentioned in the documentation (https://dzhelezov.gitbook.io/hydra/docs/schema-spec/variant-types).
> 
> 

This looks like a bug will look closer into this. Does it work for any variant field?


> **Deeply nested variant types**
> 
> There seems to be an issue with querying deeply nested varaint types, let's assume we have a following schema:

I _think_ this is easy to implement using the latest version of warthog where there's better support for typed JSON fields.

bedeho:
# Nested relationship filtering

I totally agree with Dmitrii here, one level of indirection on the filtering is OK, we have that now, but doing multiple levels is just too complex, and doing flattening on specific things is the best tradeoff.

# Ordering by relations

I thought this was planned for one level of indirection here

https://github.com/Joystream/hydra/issues/197

but if we have no use case for this, and we only rely on flattening like the prior point, then I guess its not very urgent at the moment.

# Filtering by variant type & Deeply nested variant types

@dzhelezov I thought this came out of the box by using the JSON filtering powers of PostGres or whatever it was? Anyway, this seems sufficiently important that it should be there. Good that its not that hard to add.

mnaamani:
### Test results

After much tweaking and fiddling with scripts, the tests ran but I saw some occasional `2021-06-03 17:42:13        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Transaction is outdated` which resulted in some scenarios failing.

I ran again and all passed. So I think we are still suffering from some issues with non deterministic scenarios especially when optimizing to run parallel jobs and flows. Should we attempt to review the framework critically?

### Docker issues - on Mac
Getting to the point of being able to run the tests was very awkward and took some trial and error.

First issue I encountered is that my docker-compose doesn't seem to handle services processor-mnt and graphql-server-mnt properly.

```
  processor-mnt:
    <<: *processor
    image: node:12
    build: . 
```

```
Mokhtars-MBP:olympia mokhtar$ docker compose up -d processor-mnt
[+] Building 0.1s (1/2)                                                                                                           
 => [internal] load build definition from Dockerfile                                                                         0.0s
 => => transferring dockerfile: 2B                                                                                           0.0s
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount862589388/Dockerfile: no such file or directory
```

So I proceeded to used the processor and graphql-server without the mounted volume (I changed the start.sh script). However I had to build them manually with:

```
docker compose build processor --build-arg TYPEGEN_WS_URI=ws://host.docker.internal:9944
docker compose build graphql-server --build-arg TYPEGEN_WS_URI=ws://host.docker.internal:9944
```
I could have also added it to .env file, (after building query-node) and disabling `yarn clean` from build.sh. because the same TYPEGEN_WS_URI value doesn't make sense on local build vs when building the image.

On linux it seems `ws://localhost:9944` does reach the host machine in all the scenarios which is very different to behavior on the mac. 

Switching start.sh script back to use `-mnt` services (had to remove the `build: .` from compose file for those services to run). When running the tests the graphql-server is not reachable:

```
ApolloError: request to http://127.0.0.1:8081/graphql failed, reason: connect ECONNREFUSED 127.0.0.1:8081
```
... logs from the service shows:
```
error Command "yarn" not found.
```
tried pulling newer version of `node:12` image didn't help.

In short, the experience on mac is very different than linux, and difficult to work with.
I like the mounted volume approach to eliminate building in docker image, it can simplify the configurations of the services and it works nicely during development,  but we do loose  ability to publish re-usable image of the processor and graphql-server for remote deployment for example. So it would be good to get both working properly make it easier to start and run the tests.

mnaamani:
> Switching start.sh script back to use `-mnt` services (had to remove the `build: .` from compose file for those services to run). When running the tests the graphql-server is not reachable:

That was actually my mistake. I had originally tried to build the `processor-mnt`, but adding the `build: .` line allowed me to run the services successfully.

--

Okay so I feel really silly now. All the problems I encountered were caused by my misguided attempt to `build` an image for the `*-mnt` services, and when failing to do that, build the image of joystream/apps to be used for the processor and graphql-server,  its now crystal clear why you introduced the new approach

I need a vacation

query-node/mappings/common.ts
```diff
@@ -0,0 +1,58 @@
...
+    // We use `toObject()` to get rid of .prototype defaults for optional fields
+    return metadataType.toObject(metadataType.decode(metadataBytes.toU8a(true))) as T
+  } catch (e) {
+    console.error(`Cannot deserialize ${metadataType.name}! Provided bytes: (${metadataBytes.toHex()})`)
+    return null
+  }
+}
+
+export function bytesToString(b: Bytes): string {
+  return Buffer.from(b.toU8a(true)).toString()
```
mnaamani: If `Bytes` is storing a single null byte `0x00` the conversion to utf-8 string here works, but the utf-8 null cannot be stored by postgres database as we have experienced when testing in sumer. So we need to deal with it, or on inserting/updating the resulting `string` into database query will result in an error similar to this:
```
name: QueryFailedError, message: invalid byte sequence for encoding "UTF8": 0x00, stack: QueryFailedError: invalid byte sequence for encoding "UTF8": 0x00
    at new QueryFailedError (/joystream/node_modules/typeorm/error/QueryFailedError.js:12:28)
    at PostgresQueryRunner.<anonymous> (/joystream/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:248:31)
    at step (/joystream/node_modules/typeorm/node_modules/tslib/tslib.js:143:27)
    at Object.throw (/joystream/node_modules/typeorm/node_modules/tslib/tslib.js:124:57)
    at rejected (/joystream/node_modules/typeorm/node_modules/tslib/tslib.js:115:69)
```
Background:
https://stackoverflow.com/questions/1347646/postgres-error-on-insert-error-invalid-byte-sequence-for-encoding-utf8-0x0

We have simple solution in sumer branch:
https://github.com/Joystream/joystream/blob/ce563b4981f6840397131882fc7a966b36357c2b/query-node/mappings/src/membership.ts#L216

tests/integration-tests/src/QueryNodeApi.ts
```diff
@@ -0,0 +1,679 @@
...
+  private readonly tryDebug: Debugger.Debugger
+
+  constructor(queryNodeProvider: ApolloClient<NormalizedCacheObject>) {
+    this.queryNodeProvider = queryNodeProvider
+    this.debug = Debugger('query-node-api')
+    this.queryDebug = this.debug.extend('query')
+    this.tryDebug = this.debug.extend('try')
+  }
+
+  public tryQueryWithTimeout<QueryResultT>(
```
mnaamani: The implementation is correct. However it is a bit painful to parse and maintain. I think combination of using `Promise.any()` and an async function while a while(true) in it would make the code a bit cleaner.

`Promise.any(P1, P2)` where P1 would be promise waiting for the timeout and the P2 being promise doing the query and retries.

psuedo code
```js
async function tryQuery() {
  while(true) {
    try {
       const result = await query()
       // check result.. 
       return result
     } catch {
        //...
     }

     if (retries++ < N) {
       await delay()
     } else {
       throw
     }
  }
}

P2 = tryQuery()
```

tests/integration-tests/src/scenarios/full.ts
```diff
@@ -0,0 +1,2 @@
+import './memberships'
+import './workingGroups'
```
mnaamani, line 2: As long as the scenarios do not interfere I guess it is safe to run them in parallel.

tests/integration-tests/src/fixtures/membership/AddStakingAccountsHappyCaseFixture.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        assert.isNotEmpty(qMember.boundAccounts)
+        assert.includeMembers(qMember.boundAccounts, accounts)
+      }
+    )
+
+    // Check events
+    const qAddedEvents = await this.query.getStakingAccountAddedEvents(memberContext.memberId)
+    const qConfirmedEvents = await this.query.getStakingAccountConfirmedEvents(memberContext.memberId)
+    accounts.forEach(async (account, i) => {
```
mnaamani:
```suggestion
    accounts.forEach((account, i) => {
```

async does not seem necessary here

tests/integration-tests/query-node-schema.json
```diff
@@ -0,0 +1,4988 @@
+{
```
mnaamani: how was this file generated or copied into here? is it the job of the `generate:graphql-types` script?

mnaamani (commented):
Great job on this. I focused mostly on the high level, I have to re-read the runtime code to be able to confidently even look for any "incorrect" tests.
The tests are running well although I did comment earlier about it failing sometimes. Maybe the running of the two scenarios concurrently has anything to do with it?

tests/integration-tests/query-node-schema.json
Lezek123: It think it was a relict of my old approach, removed in https://github.com/Joystream/joystream/pull/2345/commits/f3df57de62d5531745295ccad1ef6409f75205ab

tests/integration-tests/src/fixtures/membership/AddStakingAccountsHappyCaseFixture.ts
Lezek123: Thanks for noticing, fixed in https://github.com/Joystream/joystream/pull/2345/commits/f3df57de62d5531745295ccad1ef6409f75205ab

tests/integration-tests/src/scenarios/full.ts
Lezek123: True, I had to move away from this approach later on (ie. forum mappings).

tests/integration-tests/src/QueryNodeApi.ts
Lezek123: Thank you for the suggestion, implemented in https://github.com/Joystream/joystream/pull/2345/commits/f3df57de62d5531745295ccad1ef6409f75205ab

query-node/mappings/common.ts
Lezek123: I already implemented and tested a fix for this on forum mappings branch, added it to this PR also to make sure it's not lost: https://github.com/Joystream/joystream/pull/2345/commits/f3df57de62d5531745295ccad1ef6409f75205ab

PR #2347, opened 2021-04-15
Add missing events for working-group pallet.

shamil-gadelshin:
#### Changes
- add `RewardPaid` event
- add `WorkerStartedLeaving` event
- fix `leaving period` bug

### DIFF
--- a/runtime-modules/working-group/src/lib.rs
+++ b/runtime-modules/working-group/src/lib.rs
@@ -256,10 +256,16 @@ decl_event!(
 
         /// Emits on budget from the working group being spent
         /// Params:
-        /// - Reciever Account Id.
+        /// - Receiver Account Id.
         /// - Balance spent.
         /// - Rationale.
         BudgetSpending(AccountId, Balance, Option<Vec<u8>>),
+
+        /// Emits on paying the reward.
+        /// Params:
+        /// - Receiver Account Id.
+        /// - Reward
+        RewardPaid(AccountId, Balance),
     }
 );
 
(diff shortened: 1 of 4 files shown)


bedeho (changes requested):
LGTM. But don't we also need events for failed payments? This is currently a silent side-effect.

Lezek123:
> LGTM. But don't we also need events for failed payments? This is currently a silent side-effect.

Right, it would be nice to have this information as part of the event/events too, as well as some information about whether the reward was paid as part of standard worker payout or was paid to "make up" for the previously missed reward.

shamil-gadelshin:
Added new events:
- MissedRewardPaid 
- NewMissedRewardLevelReached

runtime-modules/working-group/src/lib.rs
```diff
@@ -1376,7 +1432,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
         WorkerById::<T, I>::iter()
             .filter_map(|(worker_id, worker)| {
                 if let Some(started_leaving_at) = worker.started_leaving_at {
-                    if started_leaving_at + worker.job_unstaking_period >= Self::current_block() {
+                    if started_leaving_at + worker.job_unstaking_period <= Self::current_block() {
```
conectado, line 1435: I don't understand this change, isn't the unstaking period finished after `job_unstaking_period`?

conectado (commented):
I just left one question that I didn't understand a change.

runtime-modules/working-group/src/lib.rs
shamil-gadelshin: It's easier to reason about if we rewrite the condition: `if current_block >= last_worker_date`

conectado: But this condition was correct before? It seems that before it was `if current_block <= worker_leaving_block` or there was some other change related to this?

shamil-gadelshin: Added an additional test to verify correct work of the `job_unstaking_period` parameter.

conectado (approved):
LGTM

PR #2352, opened 2021-04-16
Drop archival status from forum thread

### LINKED ISSUE #2222: Drop archival status functionality from threads
Since in #2210 we introduced an incentive to delete threads there is no value to keep the archival in threads.

This would also require to update the handbook according to these changes.

conectado:
Fix #2222 

Companion handbook PR: Joystream/handbook#43


shamil-gadelshin (approved):
LGTM

Issue #2224, opened 2021-03-03
v2 Storage and Distribution System Specification


bedeho:
# Background

The current Joystream network, as of the `Sumer` release, has an extremely limited system for storing and distributing data, both in terms of functionality and ability of the system to accommodate any kind of even limited scale of utilisation. This specification is intended to substantially improve upon these limitation by settling the organisation and function of the system at a level suitable for mainnet purposes. Importantly, this specification should be read in the context of the Gateway specification https://github.com/Joystream/joystream/issues/2089, which is complementary in that it outlines an incentive model for the vast majority of expected load on distributors.

# Major Changes

The overall design philosophy of the system remains the same as before in the following respects

- Permissioned entry of of staked service providers.
- Discretionary slashing, not bound by cryptographic evidence.
- Chain holds index of data, service provider obligations and rations utilisation of the system.
- Publishers are not charged for the distribution cost of data.
- Single chain selected upload host for each data object.

However the following major changes are introduced

- Distinct roles for storage and distributing data.
- Storage system with redundancy and only partial replication in nodes.
- Distribution system with flexible policy space, allowing for CDN like organisation.
- Efficient deletion and ownership transfer of groups of data objects.
- On-chain host resolution metadata.
- Distributors are incentivised in the Gateway system: https://github.com/Joystream/joystream/issues/2089

# Architecture

## Working Groups and Roles

There are two working groups, the _storage working group_ and the _distribution working groups_, each with its own set of workers and two separate leads, called the _storage lead_ and _distribution lead_ respectively. The workers in the storage working group are called _storage providers_, and operate dedicated nodes for this purpose, called _storage nodes_. Likewise for the distribution group there is _distribution providers_ and _distribution nodes_.

## Activity

There are two major service provider activities: storing data and distributing data. By storing we mean long term archiving, and by implication accepting uploads of new data over time. Storing data also implies distributing data to others in order for them to persist it, for example as new data or providers enter the system. The primary operational requirement of this activity is the ability to securely persist and recover very large datasets. By distributing data we mean streaming transmission of a small subset of of all available data, with low latency and high throughput, to a large number of simultaneous users. Initial data copies are obtained from storage providers. All of communication uses HTTP, specifically

1. user data uploads,
2. coordination and data transmission between storage nodes,
3. data transmission between storage nodes and distribution nodes, and
4. user downloads.

When new data is added to the system, the chain initially determines how it should be stored and distributed, but later manual intervention can update this.

[...]

PR #2357, opened 2021-04-21
Storage-v2 runtime pallet. Upload workflow.

shamil-gadelshin:
This is the initial PR for the storage version 2 runtime pallet it is based on the Antioch release.

The PR aims to prepare extrinsics required for the simplest "upload data objects" workflow, complete new storage system specification could be found [here](https://github.com/Joystream/joystream/issues/2224).

##### Uploading
- `upload` - creating the data object from the Atlas.
- `accept_pending_objects` - accept just uploaded object by the Storage Provider.

##### Storage provider registration (via CLI):
- `set_storage_operator_metadata` - set storage provider URL.
- `accept_storage_bucket_invitation` - approved connection of the Storage Provider to the bucket.

#####  Bucket management by the Storage WG leader (probably via CLI):
- `create_storage_bucket` - creates a bucket with the optional storage provider invitation.
- `update_storage_buckets_for_static_bags` - manages a connection between storage buckets and static bags.

### Limitations
- no benchmarking (until the Olympia release)
- no dynamic bags
- no working group static bags (council bag only)
- no distribution-related code and data structures
- `authentication_key` is not used on `upload` (see question 2-c below). 
### Comments
- Some code is already outdated compared to the Olympia release and will be refactored after the merging with Olympia (ensure_working_group_leader_origin, ensure_worker_origin)
- `deletion_prize_source_account_id` from the `upload` extrinsic should be equal to the member controller account (see question 2-a below).
- There is a useful abstraction created for the 'local module treasury pattern' (`ModuleAccount` trait). The bounty and forum pallets have to be refactored after the merging with Olympia.
- `StorageBucketById` map is implemented as a Substrate map instead of a "flat" rust map from the specification.
### Open questions
1. Constant values:
   - a) MaxStorageBucketNumber
   - b) MaxNumberOfDataObjectsPerBag
   - c) DataObjectDeletionPrize
2. `upload` extrinsic
   - a) `deletion_prize_source_account_id` - in order to have such a dedicated account - we need to verify it first (like staking accounts). Could we just use the signing account? 
   - b) Should we check for the duplicated content (ipfs_content_id)?
   - c) What is the purpose of the `authentication_key`? Should we copy it to each data object and demand the storage provider to have it on each 'accepting data object' request?  
   - d) Should we copy the data from `DataObjectCreationParameters` to the `DataObject` on actual object creation like `size` and `ipfs_content_id`?  
   - e) Do we need `can_upload` extrinsic?
3. `create_storage_bucket` extrinsic
   - a) Should we validate the `voucher` parameter? What are the limits? 
   - b) I set `StorageBucket.accepting_new_bags` to the `accepting_new_data_objects` parameter from the extrinsic. Is it correct?
4. `update_storage_buckets_for_static_bags` extrinsic
   - a) Is the Storage WG leader the correct user for this extrinsic?
   - b) Parameters validation requires O(NxM) storage reads to verify storage buckets' existence ("bags" x "storage buckets"). Do we have any limits on buckets-per-bag number? We could also allow changing single bag only in this extrinsic.
5. `set_storage_operator_metadata` extrinsic
   - a) We don't validate the metadata, correct?
6. `accept_pending_data_objects` extrinsic
   - a) Should I use a bucket voucher here to validate the extrinsic call?
   - b) How do we validate that objects are accepted by a correct storage provider that was invited for the storage bucket. Should we introduce an additional `storage_bucket_id` parameter to the extrinsic?


bedeho (changes requested):
I think this is a great start, but this was actually super hard to review with the amount of TODO notes and background implicit changes we are going to make. I think it would actually be best for the next PR for review to incorporate answers to basically all outstanding questions on the storage side. So yes, a full implementation of the storage side of everything else (obv without benchmarks). I know I encourage small PRs, but perhaps its easier when those PRs are internally complete change sets, this middle ground was difficult. 

Here are some superficial notes

- Perhaps we should allow a single upload to act across bags, currently we cannot due to `bag_id` being outside `DataObjectCreationParameters`. I can imagine cases where this flexibility is needed.
- Lets put metadata for storage operators into that type, not a seprate map.
- Some cases we end up with mutables in scope of extrinsic, probably because constructors of `BTreeMap` and `BTreeSet`, in these cases perhaps we could at least hide these temporary mutables inside scope of dedicated helper methods that just return a new copy of whatever object we need.

shamil-gadelshin:
Superceded by [this PR](https://github.com/Joystream/joystream/pull/2407)

Issue #2252, opened 2021-03-16
Storage provider host metadata for Sumer


bedeho:
# Background

In `Sumer` the working group workers have a raw storage fields for putting metadata

https://github.com/Joystream/joystream/blob/sumer/runtime-modules/working-group/src/lib.rs#L322

using new extrinsic `update_role_storage`

https://github.com/Joystream/joystream/blob/sumer/runtime-modules/working-group/src/lib.rs#L404

This will be used for two purposes

1. Storage provider workers will hold the URL for their storage nodes.
2. Gateway operators will hold the URL for their Atlas instance.

We want Atlas to be able to use 1 in order to lookup storage providers to interact with, rather than the current extremely complex IPNS based solution, however, Atlas current is not well equipped to directly read storage state, hence they need the query node to support this. 

This is a temporary requirement, as the new storage system will surely require revising this.

The gateway operator information is not relevant to the proposal below.

# Proposal

Introduce a simple standalone entity for storage providers, something like this in the input schema

```graphql
@entity StorageProvider {
 id: ID!
 metadata: String!
 }
 ```
 
 Mappings create and destroy instance by looking at workers getting hired & fired from the storage working group, and `metadata` field is updated by looking at `WorkerStorageUpdated`.

ondratra:
Watching for metadata update as well as a user entering/leaving storage working group has been implemented. `DataObject` entity, which is basically an asset representation in query node, now has a relation with a new `StorageProvider` entity.

A question remains, what should happen when a storage provider is leaving the storage working group (either voluntarily or is kicked out) with `DataObject`s that are associated with a provider. Should all data objects provided be deleted and an asset availability set to `INVALID`, or is there an algorithm for selecting the next storage provider hosting the content (if there exists one)? Or something entirely else should happen?

kdembler:
I think that other storage providers should have the data object repliacated so I certainly wouldn't go as far as deleting the data object. I guess the query node could just keep the status as `ACCEPTED` but not provide a storage provider relationship. In that case Atlas can just try some random provider

ondratra:
I suspect there might be an edge-case situation when a storage provider accepts content, but he leaves the storage working group before it can be fully replicated to others. That is likely the only situation when complete invalidation of `DataObject`s might be needed.

bedeho:
[...]

PR #2361, opened 2021-04-26
Query node storage provider

ondratra:
Adds storage provider metadata to query node. Solves https://github.com/Joystream/joystream/issues/2252


mnaamani:
I don't see anything new in this PR @ondratra ? If you can update from sumer it will be clearer,,

query-node/mappings/src/storage.ts
```diff
@@ -87,8 +88,16 @@ export async function dataDirectory_ContentAccepted(db: DatabaseManager, event:
     return inconsistentState('Non-existing content acceptation requested', encodedContentId)
   }
 
+  // load storage provider
+  const storageProvider = await db.get(StorageProvider, { where: { id: storageProviderId.toString() } as FindConditions<StorageProvider>})
+
+  // ensure object exists
+  if (!storageProvider) {
+    return inconsistentState('Non-existing content acceptation requested', storageProviderId)
```
mnaamani:
```suggestion
    return inconsistentState('Missing Storage Provider Id', storageProviderId)
```

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+export async function workingGroup_OpeningFilled(
+  db: DatabaseManager,
+  storageProviderType: StorageProviderType,
+  applicationIdToWorkerIdMap: ApplicationIdToWorkerIdMap
+): Promise<void> {
+  const storageProviderIds = [...applicationIdToWorkerIdMap.values()]
+
+  for (const storageProviderId of storageProviderIds) {
+    // load storage provider
+    const storageProvider = await db.get(StorageProvider, { where: { id: storageProviderId.toString() } as FindConditions<StorageProvider> })
```
mnaamani: This check is not necessary, the same worker id will never get re-used.
As a matter of fact if you find an existing worker id you have inconsistent state.

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+      id: storageProviderId.toString(),
+      type: storageProviderType,
+      isActive: true,
+    })
+
+    await db.save<StorageProvider>(newStorageProvider)
+  }
+
+  // emit log event
+  logger.info("Storage provider has beed created", {ids: storageProviderIds.map(item => item.toString())})
```
mnaamani:
```suggestion
  logger.info("Worker has been created", {ids: storageProviderIds.map(item => item.toString())})
```

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+  // do removal logic
+  await deactivateStorageProvider(db, storageProviderId)
+
+  // emit log event
+  logger.info("Storage provider has beed removed (worker exited)", {id: storageProviderId})
+}
+
+/////////////////// Helpers ////////////////////////////////////////////////////
+
+async function reactivateStorageProvider(db: DatabaseManager, storageProvider: StorageProvider) {
```
mnaamani: this method is not necessary.

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+  storageProvider.isActive = true
+
+  await db.save<StorageProvider>(storageProvider)
+}
+
+async function deactivateStorageProvider(db: DatabaseManager, storageProviderId: WorkerId) {
```
mnaamani: rename this method to `deactivateWorker()`

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+  logger.info("Storage provider has beed created", {ids: storageProviderIds.map(item => item.toString())})
+}
+
+export async function workingGroup_WorkerStorageUpdated(db: DatabaseManager, storageProviderId: WorkerId, newMetadata: Bytes): Promise<void> {
+  // load storage provider
+  const storageProvider = await db.get(StorageProvider, { where: { id: storageProviderId.toString() } as FindConditions<StorageProvider> })
+
+  // ensure storageProvider exists
+  if (!storageProvider) {
+    return inconsistentState('Non-existing storage provider update requested', storageProviderId)
```
mnaamani: message can be improved and generalized to a worker not storage provider

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,180 @@
...
+  await db.save<StorageProvider>(storageProvider)
+}
+
+async function deactivateStorageProvider(db: DatabaseManager, storageProviderId: WorkerId) {
+  // load storage provider
+  const storageProvider = await db.get(StorageProvider, { where: { id: storageProviderId.toString() } as FindConditions<StorageProvider> })
+
+  // ensure storageProvider exists
+  if (!storageProvider) {
+    return inconsistentState('Non-existing storage provider deletion requested', storageProviderId)
```
mnaamani: message can be improved and generalized to a worker not storage provider

query-node/schema.graphql
```diff
@@ -386,3 +386,22 @@ type FeaturedVideo @entity {
   "Reference to a video"
   video: Video!
 }
+
+enum StorageProviderType {
```
mnaamani:
```suggestion
enum WorkerType {
```

or 

```suggestion
enum WorkeringGroup {
```

query-node/schema.graphql
```diff
@@ -386,3 +386,22 @@ type FeaturedVideo @entity {
...
+  GATEWAY
+  STORAGE
+}
+
+type StorageProvider @entity {
```
mnaamani:
```suggestion
type Worker @entity {
```

query-node/schema.graphql
```diff
@@ -386,3 +386,22 @@ type FeaturedVideo @entity {
...
+  "Runtime entity identifier (EntityId)"
```
mnaamani: What is an entity and EntityId ?

mnaamani: Do you mean runtime `WorkerId` type

query-node/schema.graphql
```diff
@@ -386,3 +386,22 @@ type FeaturedVideo @entity {
...
+  id: ID!
```
mnaamani, line 397: This will not work.. You need separate `@entity` for each group. because worker Id  is NOT unique across the various groups. so you can have a storage provider with id 10, as well as gateway worker with the same id 10.

mnaamani (changes requested):
You cannot put all workers under the same `@entity`.
A few other minor points as well.

query-node/mappings/src/storage.ts
ondratra: Messages has been improved.

query-node/mappings/src/workingGroup.ts
ondratra: The check has been removed.

query-node/mappings/src/workingGroup.ts
ondratra: Messages has been improved.

query-node/mappings/src/workingGroup.ts
ondratra: The function has been removed.

query-node/mappings/src/workingGroup.ts
ondratra: The change is part of storage provider rename to worker.

query-node/mappings/src/workingGroup.ts
ondratra: Messages has been improved.

query-node/mappings/src/workingGroup.ts
ondratra: Messages has been improved.

query-node/schema.graphql
ondratra: Name has been changed to `WorkerType`.

query-node/schema.graphql
ondratra: The commentaries have been improved for various IDs. Now they state `Runtime identifier` for ids that correspond to runtime ids, or `Unique identifier` where they don't correspond.

query-node/schema.graphql
ondratra: Name has been changed.

query-node/schema.graphql
ondratra: All workers are still saved into one `@entity`, but they save runtime id into a new field `workerId`. `workerId` in combination with `workerType` can be used to distinguish workers from different working groups.

ondratra: Example GraphQL query: 
```
query {
   workers(where: {workerId_eq: "1", type_eq: STORAGE }) {id, workerId type}
}
```

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,165 @@
...
+  // call generic processing
+  await workingGroup_OpeningFilled(db, WorkerType.STORAGE, applicationIdToWorkerIdMap)
+}
+
+export async function storageWorkingGroup_WorkerStorageUpdated(db: DatabaseManager, event: SubstrateEvent): Promise<void> {
+  // read event data
+  const {workerId, bytes: newMetadata} = new StorageWorkingGroup.WorkerStorageUpdatedEvent(event).data
+
+  // call generic processing
+  await workingGroup_WorkerStorageUpdated(db, workerId, newMetadata)
```
mnaamani: you are passing worker id as the unique id, you should be passing the worker type as well or derive the unique ID here for the worker type and pass it.

query-node/mappings/src/workingGroup.ts
```diff
@@ -0,0 +1,165 @@
...
+    await db.save<Worker>(newWorker)
+  }
+
+  // emit log event
+  logger.info("Worker has been created", {ids: workerIds.map(item => item.toString())})
+}
+
+export async function workingGroup_WorkerStorageUpdated(db: DatabaseManager, workerId: WorkerId, newMetadata: Bytes): Promise<void> {
+  // load worker
+  const worker = await db.get(Worker, { where: { id: workerId.toString() } as FindConditions<Worker> })
```
mnaamani: I don't think this is correct. you are using worker id in place of the unique id

query-node/mappings/src/workingGroup.ts
mnaamani: Same problem in the following functions event handlers`_TerminatedWorker`, `_WorkerExited`, `_WorkerStorageUpdated`

ondratra: You're right, part of implementation was missing there. Fixed

query-node/mappings/src/workingGroup.ts
ondratra: Fixed

mnaamani (approved):

Issue #2366, opened 2021-04-27
Necessary steps to make the Sumer runtime a Parachain on Roccoco 


iorveth:
1. Upgrade substrate dependencies to `rococo-v1` branch.
2. Upgrade Joystream runtime to be compatible with `rococo-v1`.
3. Merge Joystream runtime into cumulus framework.
4. Introduce feature approach to have a single snapshot codebase for both parachain and standalone chain, similar to 
https://github.com/AcalaNetwork/Acala/blob/eeeba69b223ec28c64592ad58fe34dce837ec45c/runtime/mandala/src/lib.rs#L1746
5. Apply for the rococo v1 parachain registration https://wiki.polkadot.network/docs/ru/build-parachains-rococo#rococo-v1-parachain-requirements
6. Register as rococo parachain https://wiki.polkadot.network/docs/en/build-parachains-rococo

iorveth:
Outdated, look into PR itself https://github.com/Joystream/joystream/issues/2366

PR #2370, opened 2021-04-28
Joystream parachain

iorveth:
- [x] Upgrade substrate dependencies to be fetched from `polkadot-v0.9.6` branch.
- [x] Upgrade Joystream runtime to be compatible with `polkadot-v0.9.6`.
- [x] Merge Joystream runtime into cumulus framework.
- [x] Introduce feature approach to have a single snapshot codebase for both parachain and standalone chain, similar to
https://github.com/AcalaNetwork/Acala/blob/eeeba69b223ec28c64592ad58fe34dce837ec45c/runtime/mandala/src/lib.rs#L1746

To build Joystream runtime in standalone chain mode, run
`cargo +nightly-2021-05-11 build --features standalone`

Closes https://github.com/Joystream/joystream/issues/2366, https://github.com/Joystream/joystream/issues/2470


shamil-gadelshin (changes requested):
### Rococo-migration PR comments

#### Important

- We need to change the merging branch from sumer to dedicated rococo branch.

- How did we choose PARA_ID?

- Invalid SS58Prefix (default) in the frame_system settings in runtime.

- I get the `Error: Input("Error parsing spec file: missing field `relay_chain` at line 190 column 1")` message on trying to connect to the rococo relay chain. Am I missing something (I checked the polkadot version and generated the spec json using the cumulus-workshop example)?

- runtime/proposals-codex - why is standalone an empty feature? Should it include the 'std' instead? Acala network has empty features set as well for the standalone feature. What is the rationale behind it?


#### Regular

- Change comments on this entry `type AccountId = u128; // u64 is not enough to hold bytes used to generate bounty account`

- Node commands parameters were erased with the substrate defaults

- Change Module to Pallet to satisfy new Parity convention.

- Do we need acala license notification in the runtime?

- We need to upgrade the Frame-syntax to a different macro style (Frame v2 - pallet::*). Please, create a dedicated issue and label it with 'tech-debt'.

- Fix warning: use of deprecated associated function `sp_runtime::Perbill::from_rational_approximation`: Use `PerThing::from_rational` 

- Restore substrate pallet weights (for system pallet, staking, etch..)

- Restore affected typescript files! (like augment-api-tx.ts)

- Rewrite chain-spec code snippet - fix duplicates:

```
pallet_balances: BalancesConfig {
            balances: endowed_accounts
                .iter()
                .cloned()
                .map(|k| (k, ENDOWMENT))
                // TODO Fix dulicates in initial_authorities
                // Additional duplicates check in new balances pallet crashes runtime
                //.chain(initial_authorities.iter().map(|x| (x.0.clone(), STASH)))
                .chain(
                    initial_balances
                        .iter()
                        .map(|(account, balance)| (account.clone(), *balance)),
                )
                .collect(),
```

- Node dependencies - change to the single-line format.

mnaamani:
On first attempt to build I get some errors with:

```
active toolchain
----------------

nightly-x86_64-apple-darwin (default)
rustc 1.54.0-nightly (b663c0f4f 2021-05-29)
```

My nightly version:
nightly-x86_64-apple-darwin unchanged - rustc 1.54.0-nightly (b663c0f4f 2021-05-29)


```
     Compiling environmental v1.1.2
  error[E0557]: feature has been removed
    --> /Users/mokhtar/.cargo/registry/src/github.com-1ecc6299db9ec823/environmental-1.1.2/src/lib.rs:42:43
     |
  42 | #![cfg_attr(not(feature = "std"), feature(const_fn))]
     |                                           ^^^^^^^^ feature has been removed
     |
     = note: split into finer-grained feature gates

  For more information about this error, try `rustc --explain E0557`.
  error: could not compile `environmental`

```

mnaamani:
I did not notice that I had to use specific version of `nightly-2021-05-11` as was noted in https://github.com/Joystream/joystream/pull/2370#issue-625292212

mnaamani:
after a successful build, just tried to run `joystream-node --version` and I got :)

```
./target/release/joystream-node --version
joystream-node SUBSTRATE_CLI_IMPL_VERSION
```

mnaamani:
Build is failing with errors coming for dependency `jsonrpsee-http-client`
Are you seeing this too @iorveth

mnaamani:
I'm doing a `cargo update` to pull latest changes from substrate polkadot-v0.9.4 branch and to fix the json package breaking with:

```
cargo update
cargo update -p jsonrpsee-types -p --precise 0.2.0-alpha.6
```

Does it make sense to do that and checking in the changes to Cargo.lock @iorveth ?

mnaamani:
And for testing on the local relay chain I'm using polkadot repo branch `release-v0.9.4` and doing similar cargo update

iorveth:
> I'm doing a `cargo update` to pull latest changes from substrate polkadot-v0.9.4 branch and to fix the json package breaking with:
> 
> ```
> cargo update
> cargo update -p jsonrpsee-types -p --precise 0.2.0-alpha.6
> ```
> 
> Does it make sense to do that and checking in the changes to Cargo.lock @iorveth ?

I did not try it yet, but in theory it should work. Also on ubuntu i occasionally get a build error due to fingerprints mismatch, in that case, i suggest to just run `cargo clean` and rebuild

Also hope that issue wouldn't bother us at all after upgrading to  `polkadot-v0.9.5` https://github.com/paritytech/substrate/pull/9036

node/src/chain_spec.rs
```diff
@@ -0,0 +1,491 @@
...
+            changes_trie_config: Default::default(),
+        },
+        pallet_balances: BalancesConfig {
+            balances: endowed_accounts
+                .iter()
+                .cloned()
+                .map(|k| (k, ENDOWMENT))
+                // TODO Fix dulicates in initial_authorities
+                // Additional duplicates check in new balances pallet crashes runtime
+                //.chain(initial_authorities.iter().map(|x| (x.0.clone(), STASH)))
```
mnaamani: Could you explain here why you are not giving the initial authorities a starting balance?

mnaamani (approved):

node/src/chain_spec.rs
mnaamani: I think its fine when generating the raw chainspec for it to fail if we did indeed have duplicate accounts in initial balances and endowed accounts.

mnaamani:
> after a successful build, just tried to run `joystream-node --version` and I got :)
> 
> ```
> ./target/release/joystream-node --version
> joystream-node SUBSTRATE_CLI_IMPL_VERSION
> ```

Can we fix this @iorveth so it has proper name, version etc.. like for example:

```
joystream-node 5.5.0-f7bc4429a-aarch64-macos
```

mnaamani (changes requested):
Just left two small issues to fix

node/src/chain_spec.rs
iorveth: Agree, but previously we had the same accounts for both `initial_authorities` and `endowed_accounts` set via `development_config` https://github.com/iorveth/joystream/blob/sumer_parachain/node/src/chain_spec.rs#L144

Made a little filtering to ensure no duplicates were provided. https://github.com/Joystream/joystream/pull/2370/commits/8832b12be3f4a44350bd2f0ee6066b5e4f2c8528

iorveth:
`runtime/proposals-codex - why is standalone an empty feature? Should it include the 'std' instead? Acala network has empty features set as well for the standalone feature. What is the rationale behind it?`

Features in rust are strictly additive. If default feature includes 'std', standalone would also include it, unless you provide `--no-default-features` flag

mnaamani:
Here is a quick guide to how I'm testing the parachain https://github.com/Joystream/joystream/issues/2484

I'm trying to test basic balance transfer on the parachain of JOY tokens, from the endowed account to the sudo account.

```
1010: Invalid Transaction: Inability to pay some fees , e.g. account balance too low
```

I can't figure out what is wrong. The endowed account has 100M tokens. I think we need to check the SignedExtensions to make sure we didn't miss something in the runtime code. Specifically the implementation of `ChargeTransactionPayment`

PR #2371, opened 2021-04-29
Query node video metadata

ondratra:
This PR adds proper support for editing video media metadata.

Also, it fixes invalid actor reporting `invalidMetadata()` warning instead of proper `inconsistantState()` warning.

Ideally, this PR should be tested in integration test(s) before merging.


query-node/mappings/src/content/utils.ts
```diff
@@ -108,18 +108,37 @@ export interface IReadProtobufArgumentsWithAssets extends IReadProtobufArguments
...
 
+  /*
+    Determines property change from the given object property.
+  */
+  static fromObjectProperty<
+    T,
+    Key extends string,
+    ChangedObject extends {[key in Key]?: T}
+  >(object: ChangedObject, key: Key): PropertyChange<T> {
+    if (!(key in object)) {
```
mnaamani, line 131: This works, but you are having to write this code because of converting protobuf message to a JSON object.
On the original message type there is a method for each field to check if a field is set or not. Will it simplify to use that?

ondratra: Protobuf's `has*` checks are useful to some extent, but `PropertyChange` has been created because our logic behind updating/unsetting/keeping unchanged is more complex, and also, some fields we might be updating are not coming from metadata. 

The class was originally created to help with handling redundant content info like `avatarPhotoDataObject`, `avatarPhotoUrls`, `avatarPhotoAvailability` that are derived from metadata and runtime event parameters, then it got useful in few more places.

mnaamani (approved):

PR #2373, opened 2021-04-30
Olympia forum mappings & integration tests

Lezek123:
Draft PR currently for input schema review purposes.
Based on https://github.com/Joystream/joystream/pull/2345 to include other required, existing schemas.

The current commit of interest is: https://github.com/Joystream/joystream/commit/719ceffc0ef12176f2d8695c2e0f88e03897ec37

**Functionality not supported by runtime:**

Some functionality presented in the Wireframes is not supported by the current runtime implementation:

- Updating category `title` / `description` - no extrinsics exist for this purpose (only category archival status can be updated)
- Editing thread text (we have `edit_thread_title`, but no `edit_thread_text` extrinsic)
- Setting thread archival status - a category can currently be archived, but a single thread cannot. Pioneer wireframes indicate it should be possible.
- Adding threads to watchlist - there doesn't seem to be any extrinsic that could allow this
- Post / thread assets - current runtime functionality will not support uploading assets to storage (images etc. can still be added by providing urls to external resources inside md-formatted content)
- Thread reactions - no extrinsic to support this (only post reactions are supported)
- Creating threads that are not editable to begin with
	_Although this isn't something missing w.r.t. to Pioneer design wireframes, it seems like something that should be possible_

**Metadata standards:**

The current runtime implementation is quite explicit when it comes to what actions beeing performed with each extrinsic and what is the data structure of the entities (like threads and posts), so excluding the missing functionality mentioned above, the only place I could currently find some use-case for a metadata standard would be post text, since a post can be a reply to other post and this is not explicitly supported by the runtime.

A post metadata message could therefore look like this:
```
message ForumPostMetadata {
  optional string text = 1; // The post text
  optional uint32 repliesTo = 2; // Id of the post beeing replied to
}
```

In this current schema I assume the `repliesTo` field is immutable, so the standard only appiles when creating a new post.

**Fulltext searches:**
Fulltext search queries have been provided to faciliate the Pioneer search functionality:
- Thread title - `threadsByTitle`
- Thread text - `threadsByText`
- Post text - `postsByText`

**ForumModerator:**

Besides having a generic representation of `Worker`, it seems useful to have additional entities for workers beeing part of a specific group (like `ForumModerator` for forum workers).

Those entites could be created along with Worker entities during OpeningFilled and include a reference to the Worker entity.

This would be similar to each specific event (ie. `OpeningFilledEvent`) having a reference to generic event data entity (`Event`), as is currently the case.

Current example:

```
type ForumModerator @entity {
  "Moderator worker"
  worker: Worker!

  "Categories managed by the moderator"
  categories: [ForumCategory!] @derivedFrom(field: "moderators")
}
```

**Deleting database entities (`Category`, `Thread`, `Post`)**
It wasn't clear to me whether actions like deleting or moderating a post / thread / category should actually cause those query node entites to be removed, considering the relationships they may have, for exmple, with the events.

I think there are 3 potential approaches here:

1. Not deleting the entities, but setting a `Removed` / `Moderated` status instead

2. Events having both an optional relationship (like `post`) and an id field (like `postId`) - in case the post is removed the relationship is cleared, but the id remains. (this doesn't seem to be too benefitial, since post/thread contents will still remain in the database as part of the events like `PostAdded`, `PostTextUpdated` etc.)

3. Removing a post equals removing all associated events (perhaps except the removal event itself?) - this could be potentially confusing for the user though. As an api user I probably wouldn't expect some of the events to "disapear" (it seems natural that the event log is immutable, regardless of the current state), it would also make it impossible to track the stakes associated with some of the actions.

Having those in mind I decided to go for 1. The obvious drawback of this solution is that it requires persisting much more data in the database and exposes some data that we may not want the api to expose (like moderated posts), but since this data would still be available on the blockchain anyway, I think that's probably not a critical issue.

**Constants**
Constants like `POST_DEPOSIT`, `THREAD_DEPOSIT`, `MAX_POSTS_IN_THREAD` etc. (see full list: https://joystream.gitbook.io/joystream-handbook/subsystems/forum) are assumed to be known by the frontend.

They are immutable in the runtime (unless a runtime upgrade is performed), but are not exposed through the runtime api, so currently it seems like they need to be hardcoded (similarly to some other constants like minimum role unstaking period, opening creation stake etc.)  

**Archival status propagation**
Notice that `Active` threads and `Post`s that belong to archived category are still not editable by the user.
The propagation is currently implicit, so the `Thread`, `Post` and subcategory status does not change once the parent
category is archived.


bedeho:
Here are my reply to your PR text, will add review soon.


# Missing features

Can you add the following labelled issues for misssing forum features that someone can implement

- update category title.
- update category description.
- edit thread text.
- thread reactions - no extrinsic to support this (only post reactions are supported)
- creating threads that are not editable to begin with

# Notes

- `Adding threads to watchlist`: This is a purely client side feature.

- `current runtime functionality will not support uploading assets to storage`: Yes this is fine, nothing in Pioneer v2 currently has been designed to support it. Current thinking is to just power through and do the first version without this, but we may reevaluate later.

# Metadata standards

I agree, nicely spotted.

# ForumModerator

I did not understand the problem being tackled by the suggestion.

# Deleting database entities

I think not deleting from query node, but adding clear semantic tags, is a good policy, at least until we run into some problem. Remember that we explicitly want all versions of a post to be easiy recoverable from the query node, including if it was deleted. There is a great edit history UI for this in Pioneerv2 even.

I think its most important that all historical posts and threads are recoverable, categories less so, but if we want to avoid dangling references to deleted categories from deleted threads, then we will also have to keep them around.

# Constants

I did not understand why they are not exposed, isnt there a way to export module trait configurations? If it is because these constants are just local scope constants in the runtime module, then we should make them configurable through the runtime module trait. Please make an additional issue for this.

# Archival status propagation

Yes that is fine.

query-node/schemas/forum.graphql
```diff
@@ -0,0 +1,227 @@
+type ForumModerator @entity {
```
bedeho: I am not sure I see the need for this entity. A forum moderator is just a worker in the corresponding working group.

query-node/schemas/forum.graphql
```diff
@@ -0,0 +1,227 @@
...
+  poll: ForumPoll!
+
+  "The alternative text"
+  text: String!
+
+  "List of all associated vote events"
+  votes: [VoteOnPollEvent!] @derivedFrom(field: "pollAlternative")
+}
+
+enum PostReaction {
```
bedeho, line 137: Perhaps I missed it, but do we have some standard for how to encode these in extrinsics?

bedeho (commented):
Phenomenal work 💯

Lezek123:
> # Constants
> I did not understand why they are not exposed, isnt there a way to export module trait configurations? If it is because these constants are just local scope constants in the runtime module, then we should make them configurable through the runtime module trait. Please make an additional issue for this.

It is possible to make those constants accessible through `api.consts`, but it looks like in order for that, they need to be explicitly listed inside `decl_module` macro in the runtime, like for example: https://github.com/Joystream/joystream/blob/master/runtime-modules/working-group/src/lib.rs#L360

> # ForumModerator
> I did not understand the problem being tackled by the suggestion.

Perhaps it's not strictly necessary, but it seems useful to have additional entities for specific worker types.
Using the forum example - we could use it to query all moderators along with the list of categories that they manage.
This way we also enforce a specific worker type for relationships on the database level (ie. a Storage worker cannot be a category moderator).

I can skip it for now though and reconsider later if this turns out necessary.

query-node/schemas/forum.graphql
Lezek123: There is a `ReactionId` field in extrinsic, so we only need to map it to an enum variant.
Since GraphQL enum values are always strings, I can add a protobuf enum to map integers (`ReactionId`s) to specific reactions in a standarized way.

Lezek123:
> Can you add the following labelled issues for misssing forum features that someone can implement

Issues created:
- https://github.com/Joystream/joystream/issues/2389
- https://github.com/Joystream/joystream/issues/2390
- https://github.com/Joystream/joystream/issues/2391
- https://github.com/Joystream/joystream/issues/2392
- https://github.com/Joystream/joystream/issues/2393 (here I also mentioned some other issues with editable/non-editable posts/threads I noticed)

Also created an issue regarding non-exposed runtime constants: https://github.com/Joystream/joystream/issues/2395

Lezek123:
Everything from this branch is already included in https://github.com/Joystream/joystream/pull/2547

PR #2379, opened 2021-05-04
runtime: proposals: Fix insufficient balance bug.

### LINKED ISSUE #2377: Bug: User able to submit a prop. without enough tokens, thread created, no proposal created
A user in discord had issues with submitting a `spending proposal`, the error message returned was `system.ExtrinsicFailed: Other`. It turns out the user only had 10k tokens and not 25k as required, despite this the extrinsics were half successful and ended up creating several empty proposal threads via `proposal.Discussion.ThreadCreated`. No proposals were successfully created until sufficient tokens were added to the account.

Screenshot of submission:
image

Screenshot of threads being created:
image

shamil-gadelshin:
#### Bug description
In case of the insufficient balance for proposal stake, we get an error on the proposal creation, but the runtime creates a proposal discussion thread.

This happens because the runtime does not validate account balance before the thread or proposal creation (in the `ensure_create_proposal_parameters_are_valid`). The first check (and error) happen on creating the proposal stake.

#### Solution
Add additional check to the `ensure_create_proposal_parameters_are_valid`: ensure account balance is enough for stake.

#### Comments
 - `Olympia` release branch has a different staking approach and such a check is already implemented.

[Original issue](https://github.com/Joystream/joystream/issues/2377)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,7 +1993,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
-version = "3.7.2"
+version = "3.7.3"
 dependencies = [
  "frame-benchmarking",
  "frame-benchmarking-cli",
@@ -2053,7 +2053,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node-runtime"
-version = "7.16.0"
+version = "7.17.0"
 dependencies = [
  "frame-benchmarking",
  "frame-executive",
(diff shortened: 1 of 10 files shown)


mnaamani:
LGTM

mnaamani (approved):

bwhm (approved):
I can't of course really review the code, but looks correct to me!

PR #2384, opened 2021-05-06
query node - error throws on inconsistent states

ondratra:
Inconsistent states encountered in query node now throws JS error.


ondratra:
Bootstrapping of workers (storage providers) and members was added. There are no data prepared for bootstrapping in the repo. To add data, change content of `query-node/mappings/bootstrap/data/members.json` and `query-node/mappings/bootstrap/data/workers.json` by manually creating records or using export tool from https://github.com/Joystream/runtime-state-export-tools/pull/6 like this `WS_URL=wss://sumer-dev-2.joystream.app/rpc node ./lib/export_members.js`.

query-node/mappings/bootstrap/workers.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function bootWorkersInGroup(db: DatabaseManager, workers: IBootstrapWorker[], workerType: WorkerType): Promise<void> {
+  if (!workers) {
+    return
+  }
+
+  for (const rawWorker of workers) {
+    // create new membership
+    const worker = new Worker({
+      // main data
+      workerId: rawWorker.role_account_id,
```
mnaamani: This should be the `rawWorker.id` field (WorkerId) NOT the role account (AccounId).
See comment on the export script https://github.com/Joystream/runtime-state-export-tools/pull/6#discussion_r628383185

mnaamani (changes requested):

query-node/mappings/bootstrap/workers.ts
ondratra: Done

PR #2386, opened 2021-05-07
Fix cli upload of large files

mnaamani:
`maxBodyLength` was missing in request header to allow axios to upload larger content.


bwhm (approved):
Tested, and works!

Lezek123 (approved):
I must have accidently removed this part during refactorization. Works as expected now. I managed to upload a 4 GB file. With larger files (like 9 GB) I'm running into the runtime voucher limit.

mnaamani:
> With larger files (like 9 GB) I'm running into the runtime voucher limit.

Was it specifically runtime voucher issue? Does the transaction to add_content fail? Or did you see that problem during upload?

PR #2394, opened 2021-05-11
cli - chalk color change

ondratra:
CLI used `white` color for some command line outputs. White text is barely readable in consoles with white background. This PR changes it to `magentaBright` that was previously not used anywhere.


mnaamani:
I think you need to run `yarn format` in the `cli` folder to fix the prettier formatting issue

PR #2399, opened 2021-05-13
query node - working group leader support

ondratra:
This PR adds watching for working group-leader-related events to query node.

### DIFF
--- a/query-node/mappings/src/workingGroup.ts
+++ b/query-node/mappings/src/workingGroup.ts
@@ -57,6 +57,22 @@ export async function storageWorkingGroup_WorkerExited(db: DatabaseManager, even
   await workingGroup_WorkerExited(db, WorkerType.STORAGE, workerId)
 }
 
+export async function storageWorkingGroup_LeaderSet(db: DatabaseManager, event: SubstrateEvent): Promise<void> {
+  // read event data
+  const {workerId} = new StorageWorkingGroup.LeaderSetEvent(event).data
+
+  // call generic processing
+  await workingGroup_LeaderSet(db, WorkerType.STORAGE, workerId)
+}
+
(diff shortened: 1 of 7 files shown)


query-node/mappings/src/workingGroup.ts
```diff
@@ -134,27 +159,54 @@ export async function workingGroup_WorkerStorageUpdated(db: DatabaseManager, wor
...
 export async function workingGroup_WorkerExited(db: DatabaseManager, workerType: WorkerType, workerId: WorkerId): Promise<void> {
   // do removal logic
   await deactivateWorker(db, workerType, workerId)
 
   // emit log event
-  logger.info("Storage provider has beed removed (worker exited)", {workerId, workerType})
+  logger.info("Worker has been removed (worker exited)", {workerId, workerType})
+}
+
+export async function workingGroup_LeaderSet(db: DatabaseManager, workerType: WorkerType, workerId: WorkerId): Promise<void> {
```
mnaamani: We should not create a worker when handling LeaderSet event. Because the OpeningFilled event will also be triggered later for the same worker. So there will be two Workers created for leads. I suggest we either drop handling of LeaderSet event all together, or in the OpeningFilled event we look for same workerId already existing and not create a new worker. With latter approach we can risk hiding incorrect handling of when workers are removed however.

mnaamani (changes requested):
I think we should drop handling of LeadSet evet, see comment as to why.

query-node/mappings/src/workingGroup.ts
ondratra: I overlooked that. LeaderSet event was removed.

mnaamani (approved):

PR #2400, opened 2021-05-13
query node - video license update

ondratra:
This PR fixes a video's license update. Only changes truly requested will happen.


mnaamani:
Took me a while to understand the logic here. I think things could have been much simpler if the License was not a separate table/entity. The logic looks correct.

Issue #2359, opened 2021-04-25
Storage & Distribution System Utilisation Model


bedeho:
# Model

https://www.dropbox.com/s/db1ie15nflc80o8/utlization_model.numbers?dl=0

# Background

The model attempts to generate some constraints for the runtime, the node infrastructure and the costs involved. The primary goal was the runtime side, in particular the size of certain storage objects, but other aspects were also covered because it was practical. Hopefully this base can be extended to cover more cases.

Critically, this analysis only covers storage of video assets, and it also only covers how to effectively distribute all content with a baseline level of capacity. It is pivotal to be mindful of this point, because it understates that there are other cases which are likely to generate equal, if not, greater demands. In particular, we know that not streaming will be evenly distributed across all videos. In fact, some much smaller subset of channels will have videos that consistently have many orders of magnitudes greater streaming demand, likewise, some subset of videos will have the same, and to which particular videos this turns out to be the case will be very unpredictable. Lastly, there may be substantial geographic bias in the popularity of such channels and videos. None of these facs are accommodated well in the model above, and they would have to be accommodated by having lots of extra storage representations, in terms of distribution bucket families and distribution buckets, as well as lots of extra on-chain bandwidth in order for the lead to manage these configurations.

# Result

Using relatively ambitious utilisation assumptions (500K channels, 10 vids/channel, 1080p videos), it does seem in principle possible to get all the numbers to make sense, with the only exception being some bandwidth cost numbers, which we will not focus on here. However, the model also shows how bloated our representations are, and that this may easily start bumping into some limits quickly, as well as making transactions pricey. For example

- `storage_buckets_by_id` will take over 18Kb
- `distribution_bucket_families_by_id` will take over 21Kb

These must be read as singular state objects, and they are very large. We will need to increase the read/write cost coefficients in the weight functions which interact with these, this will make all transactions involved with these, such as uploading and confirming uploads, more expensive. How much more they must be increased is covered in this issue by @connectado

https://github.com/Joystream/joystream/issues/2234

It seems that, an increase to those sizes will require to increase the read coefficient by 20-30%

https://user-images.githubusercontent.com/3310803/110998882-4738c780-835e-11eb-8d4d-de050789294b.png

and write coefficient by up to 50%

https://user-images.githubusercontent.com/3310803/110998897-4e5fd580-835e-11eb-8335-46d117631737.png

but we should confirm my interpretation here.

# Analysis

[...]

PR #2407, opened 2021-05-18
Storage subsystem version 2

shamil-gadelshin:
This PR contains rewritten from scratch version 2 of the Storage subsystem. The previous version (pallets `storage` and `service discovery`) was removed.  The data objects migration for the `content directory` was removed as well.

## Initial spec links
- [spec](https://github.com/Joystream/joystream/issues/2224)
- [utilization model](https://github.com/Joystream/joystream/issues/2359)

## Notable abstractions
- DataObjectStorage - a trait for the pallet integration. Exposes pallet public methods
- ModuleAccount - a trait for the pallet specific accounts
- BagManager - a facade over dynamic and static bag operations. Provides a common interface.
- StorageBucketPicker - implements random (and complementary sequential) algorithm for picking storage buckets on dynamic bag creation

## Pallet functionality could be split into three distinct groups
- extrinsics for the working group leader
- extrinsics for the storage provider
- public methods for the pallet integration

#### Working group leader extrinsics
- create_storage_bucket
- update_storage_buckets_for_bag 
- delete_storage_bucket
- invite_storage_bucket_operator
- cancel_storage_bucket_operator_invite
- remove_storage_bucket_operator
- update_uploading_blocked_status
- update_data_size_fee
- update_storage_buckets_per_bag_limit
- update_storage_buckets_voucher_max_limits
- update_number_of_storage_buckets_in_dynamic_bag_creation_policy
- update_blacklist

#### Storage provider extrinsics
- accept_storage_bucket_invitation
- set_storage_operator_metadata
- set_storage_bucket_voucher_limits
- update_storage_bucket_status
- accept_pending_data_objects

#### Public integration methods are exposed via the DataObjectStorage.
- can_upload_data_objects
- upload_data_objects
- can_move_data_objects
- move_data_objects
- can_delete_data_objects
- delete_data_objects
- can_delete_dynamic_bag
- delete_dynamic_bag
- can_create_dynamic_bag
- create_dynamic_bag

#### Pallet constants
- MaxStorageBucketNumber
- MaxNumberOfDataObjectsPerBag
- DataObjectDeletionPrize
- BlacklistSizeLimit
- StorageBucketsPerBagValueConstraint
- DefaultMemberDynamicBagCreationPolicy
- DefaultChannelDynamicBagCreationPolicy

## Next iteration tasks
1. adding authentication_key.
2. adjust storage pallet-related constants in runtime.
3. consider adding config() to enable future storage migration.
4. make public methods as root extrinsics to enable storage-node dev mode.
5. consider adding a counter for the total objects size for a bag.

## Far future [changes](https://github.com/Joystream/joystream/issues/2409) (after the Olympia release): 
1. Benchmarking.
2. Refactoring the working group usage.
3. Moving the `ModuleAccount` to the common pallet.

Supercedes [this PR](https://github.com/Joystream/joystream/pull/2357).


runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+        /// - bag ID
+        /// - storage buckets to add ID collection
+        /// - storage buckets to remove ID collection
+        StorageBucketsUpdatedForBag(BagId, BTreeSet<StorageBucketId>, BTreeSet<StorageBucketId>),
+
+        /// Emits on uploading data objects.
+        /// Params
+        /// - data objects IDs
+        /// - initial uploading parameters
+        DataObjectdUploaded(Vec<DataObjectId>, UploadParameters),
```
```suggestion
        DataObjectsUploaded(Vec<DataObjectId>, UploadParameters),
```

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+            operation_type,
+        );
+
+        Self::change_storage_bucket_vouchers_for_bag(
+            &params.bag_id,
+            &bag_change.voucher_update,
+            operation_type,
+        );
+
+        Self::deposit_event(RawEvent::DataObjectdUploaded(
```
```suggestion
        Self::deposit_event(RawEvent::DataObjectsUploaded(
```

runtime-modules/storage/src/tests/mod.rs
```diff
@@ -1,6 +1,3551 @@
...
+        assert_eq!(
+            Balances::usable_balance(&DEFAULT_MEMBER_ACCOUNT_ID),
+            initial_balance - DataObjectDeletionPrize::get()
+        );
+        assert_eq!(
+            Balances::usable_balance(&<StorageTreasury<Test>>::module_account_id()),
+            DataObjectDeletionPrize::get()
+        );
+
+        EventFixture::assert_last_crate_event(RawEvent::DataObjectdUploaded(
```
Lezek123:
```suggestion
        EventFixture::assert_last_crate_event(RawEvent::DataObjectsUploaded(
```

Lezek123 (changes requested):
Just noticed a small typo

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+        pub fn invite_storage_bucket_operator(
+            origin,
+            storage_bucket_id: T::StorageBucketId,
+            operator_id: WorkerId<T>,
+        ){
+            T::ensure_working_group_leader_origin(origin)?;
+
+            let bucket = Self::ensure_storage_bucket_exists(&storage_bucket_id)?;
+
+            Self::ensure_bucket_missing_invitation_status(&bucket)?;
```
bedeho, line 1343: Should we _require_ that there is not even a pending invitation? Perhaps a bit overly strict for no clear benefit.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+        //
+
+        let data = Self::create_data_objects(params.object_creation_list.clone());
+
+        <StorageTreasury<T>>::deposit(
+            &params.deletion_prize_source_account_id,
+            bag_change.total_deletion_prize,
+        )?;
+
+        Self::slash_data_size_fee(
```
bedeho, line 1573: Given that the lead can unilaterally set the fee to any value they want, this could allow them to steal/burn an arbitrary quantity from a user who is unwittingly holding large quantities of funds in the account they use to fund these funds with. This can even be done after a user has sent a transaction if timed well. In particular if we are going to divert part of the fee to the storage provider confirming the upload, then it could become very attractive.

Should we remedy this by advising all users of this extrinsic to commit to the fee they are offering in the extrinsic that ultimately calls this method `upload_data_objects` here?

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+    /// _Storage_ substrate module.
+    pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+        /// Default deposit_event() handler
+        fn deposit_event() = default;
+
+        /// Predefined errors.
+        type Error = Error<T>;
+
+        /// Exports const - max allowed storage bucket number.
+        const MaxStorageBucketNumber: u64 = T::MaxStorageBucketNumber::get();
```
bedeho: I wonder if we need to cap this? buckets are just stored in a standalone map, so it's not clear if there is a big upside here, also, the number may need to get really big, in which case having a hard constant could end up being a bad problem.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+
+                Self::change_storage_buckets_vouchers(
+                    &remove_buckets,
+                    &voucher_update,
+                    OperationType::Decrease
+                );
+            }
+
+            Self::deposit_event(
+                RawEvent::StorageBucketsUpdatedForBag(bag_id, add_buckets, remove_buckets)
```
bedeho, line 1306: I think it would perhaps be better to only have a single event per extrinsic. I know we do not do this consistently prior, but it seems more clean to have one singular event carrying all side-effects of a given extrinsic. Currently, there will be up to two `VoucherChanged` events in this extrinsic prior to this final one.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+            });
+
+            Self::deposit_event(
+                RawEvent::StorageOperatorMetadataSet(storage_bucket_id, worker_id, metadata)
+            );
+        }
+
+        /// Sets storage bucket voucher limits.
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn set_storage_bucket_voucher_limits(
```
bedeho, line 1413: Only lead can set this.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+            }
+
+            Self::deposit_event(
+                    RawEvent::PendingDataObjectsAccepted(storage_bucket_id, worker_id, bag_id, data_objects)
+            );
+        }
+
+        /// Update whether new bags are being accepted for storage.
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn update_storage_bucket_status(
```
bedeho: Only lead can set this.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+        let new_objects_number: u64 = params.object_creation_list.len().saturated_into();
+        let total_possible_data_objects_number: u64 = new_objects_number + bag_objects_number;
+
+        // Check bag capacity.
+        ensure!(
+            total_possible_data_objects_number <= T::MaxNumberOfDataObjectsPerBag::get(),
+            Error::<T>::DataObjectsPerBagLimitExceeded
+        );
+
+        let mut bag_change = BagChangeInfo::default();
```
bedeho: Can we do a pure version of this, no `mut`, should be easy.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+        // Check global uploading block.
+        ensure!(!Self::uploading_blocked(), Error::<T>::UploadingBlocked);
+
+        // Check object creation list validity.
+        ensure!(
+            !params.object_creation_list.is_empty(),
+            Error::<T>::NoObjectsOnUpload
+        );
+
+        BagManager::<T>::ensure_bag_exists(&params.bag_id)?;
```
bedeho: This looks a bit fishy to me, as you are not returning an actual bag. This must mean that in many code paths you are going to be looking up that bag again, possibly with a fallible lookup, or a blind mutation, where the latter could risk updating an invalid (default constructed) instance.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+
+    // Calculates total objects number for dynamic bag.
+    pub(crate) fn objects_number(&self) -> u64 {
+        self.objects.len().saturated_into()
+    }
+}
+
+// Helper struct for the dynamic bag changing.
+#[derive(Clone, PartialEq, Eq, Debug, Copy, Default)]
+struct BagChangeInfo<Balance> {
```
bedeho: Can we call this `BagSideffects` or something, `Info` kills me 🗡️

runtime-modules/storage/src/lib.rs
bedeho: I meant to mention this earlier, but I think its probably prudent to crediting this fee to the eventual provider that correctly confirms the upload, in order to provide an incentive for this to be done correctly. This will require inclusion of some auth step in the upload process, so that only the genuine uploader can confirm. But lets do this in a later PR.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+    pub objects: BTreeMap<DataObjectId, DataObject<Balance>>,
+
+    /// Associated storage buckets.
+    pub stored_by: BTreeSet<StorageBucketId>,
+
+    /// Associated distribution buckets.
+    pub distributed_by: BTreeSet<DistributionBucketId>,
+
+    /// Dynamic bag deletion prize.
+    pub deletion_prize: Balance,
```
bedeho: What was the reason for having a separate prize here that also incorporates the size of `objects`, when each `DataObject` has its own prize to begin with? I think there was a reason, but I don't recall now.

runtime-modules/storage/src/lib.rs
```diff
@@ -1,20 +1,2373 @@
...
+                    accepting_new_bags
+                )
+            );
+        }
+    }
+}
+
+// Public methods
+impl<T: Trait> DataObjectStorage<T> for Module<T> {
+    fn can_upload_data_objects(params: &UploadParameters<T>) -> DispatchResult {
```
bedeho, line 1553: Would it be possible to make all these action extrinsics infallible provided the user provides a `bag_change` variable which they received from a positive call to  `can_*`.

bedeho (changes requested):
Awesome job! I did a decent level review, not as rigorous as I could, have, but still useful I think.

Reading this made me question whether we really should think of the distribution bucket operator needs a new id space. Ultimately, for each combination of worker & distribution bucket you have at most one relationship at any given time, as the same worker cannot operate multiple times on the same bucket at the same time. Off-chain one may want to identify distinct occasions where the same worker has a role in the same bucket over time, for example if they are added and removed multiple times, but the runtime does not need to care about this distinction.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Will fix it. Thank for finding.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Will fix it. Thank for finding.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: There are at least two possible states: 1. the same worker was invited, 2. another worker was invited.

Do we want to have the same outcome of the extrinsic with these states?

Should we silently replace the current invite (or maybe with a different event or even with a different event depending on the state)? 

Requiring to cancel the previous invite is a straightforward approach. However, this also increases the number of extrinsic calls (+1 for canceling). If you prefer to simplify usability - I recommend default invite replacement with the same event. Please, let me know if you still think we need to replace the invite.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok. We don't have any iterations with `StorageBucketById` so it is safe to delete now.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: It spares us from iteration through all data objects (could be expensive).

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok. BagManager has a state now (bag_id), and I combined its construction with an existence check.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Renamed to `BagUpdate`

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Yes, it is possible. However, such an interface has a couple of downsides:
1. It would disclose several internal helper abstractions like BagChange, VoucherUpdate, BagManager, etc. It would complicate the pallet API with no obvious benefit.

2. There will be a way to call the methods with manually constructed data bypassing the actual checks.

3. It arguably looks like a contradiction with your other comment about 
`BagManager::<T>::ensure_bag_exists(&params.bag_id)?;`

```
.... 
This must mean that in many code paths you are going to ...<use this abstraction again> ...
...possibly with a fallible lookup, or a blind mutation, where the latter could risk updating an invalid (default constructed) instance.
```

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Don't workers control their storage capacity (set_storage_bucket_voucher_limits), and a leader should only control the maximum possible capacity (update_storage_buckets_voucher_max_limits), do they? Workers seem to know more about their system's capabilities.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: I send events inside small methods closer to the storage changing code. It is very easy to forget about such events in upstream code. I started to do like that after you requested to reveal side effects in the bounty pallet. There is another such event in the storage pallet - `DeletionPrizeChanged` and there can be more than two or three such events (there is a loop inside the voucher changing function). Please, let me know if you think that forgetting about side effects is a minor issue and I will remove such hidden events in favor of more complicated upper-level events.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok. I return `DataSizeFeeChanged` error on fee change now.

runtime-modules/storage/src/lib.rs
bedeho: > Please, let me know if you still think we need to replace the invite.

We are deleting the bucket here, so its not a matter of replacing, its just a matter of entirely dropping the bucket, which by implication drops storage state which reflects the invite.

runtime-modules/storage/src/lib.rs
bedeho: Fixed in https://github.com/Joystream/joystream/pull/2407/commits/65a81779c274383c08bc5900622e39e0a0f90ea1

runtime-modules/storage/src/lib.rs
bedeho: I understand what you mean, it's a bit hard for me to judge tbh. I see the possible problems of both approaches. I guess we can just continue as is for now, and see what ends up happening upstream.

runtime-modules/storage/src/lib.rs
bedeho: No the capacity is in fact what they are paid for, not just actual utilization, so its not up to their own discretion.

runtime-modules/storage/src/lib.rs
bedeho: Does this really fix the problem? If you return a `BagManager`, you have lost the static guarantee that the bag exists, which was the main issue to begin with. Seems to me that if the way a manager is used is that its always kept around after a bag has been shown to exist, then you need to keep a copy in the manager, however, you are mostly not using these manager instances in a statefule way, so I am not sure that really makes sense. Perhaps the semantics of what this manager represents i a bit unclear, let me know what you think, and just make the decision you think is best.

runtime-modules/storage/src/lib.rs
bedeho: 1) Does that mean it is redundant, it is the sum of all the data objects?
2) When the objects are inlined, then its just iterating in memory, that is going to be super fast, and there is even going to be a length cap, so I don't think that is a real issue.
3) I believe we are still going for non-inlined dataobjects, at least last we spoke I argued pattionatley for this 🙏 📿 , in which case deletion of non-empty bags is not going to be a thing anyway.

runtime-modules/storage/src/lib.rs
bedeho: I agree.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Your comment relates to the `invite_storage_bucket_operator` extrinsic judging by line number 1357. But it looks you mean the `delete_storage_bucket` extrinsic which has the same `Self::ensure_bucket_missing_invitation_status(&bucket)?;` check. Yes, it is safe to delete the check here - will do.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: The `BagManager` is an abstraction over the static and dynamic bags storage and types. 

It handles different structs for `DynamicBag` and `StaticBag` and different storage maps for static and dynamic bags. Currently, there are only two types of storage for bags. One combined storage map for the static bags (council and working groups) and one for all dynamic bags(Member and Channel). But we could fall back to the initial storage design with one storage map per bag type without changing the code that uses the BagManager.

For example, instead of 

`BagManager::<T>::ensure_bag_exists(&bag_id)?;`

we could write

```
match bag_id{
    Static(static_bag_id) => ensure_static_bag_exists(static_bag_id)
    Dynamic(dynamic_bag_id) => ensure_dynamic_bag_exists(dynamic_bag_id)
}?;
```

The code for getting the bag, manipulating with its content and saving back will be the same for all storages. So, I absracted it with the `Bagmanager`.

However, this all goes away if we drop the `deletion prize` field in the dynamic bag struct (your another comment). Because, it makes the `StaticBag` struct identical with the `DynamicBag` struct. And I can even create a single storage map `BagId => Bag` and remove a huge part of `BagManager` or drop it completely.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: 1. 'Redundant' in the meaning of 'cached' or 'denormalized'.
2. I'm not sure 'super fast' would be still the case for large bags (starting with hundreds of thousands data objects). However, a large bag deletion looks like a very rare operation. And we're committing a primordial sin of preliminary optimization. So, let's try to drop the field until stress tests show us the light of wisdom. There are additional implications - see another comment about the `BagManager`.
3. Please, remind me: didn't we agree to prevent deleting non-empty bags forcing the leader to manually delete possibly thousands of objects considering it as a rare operation, and drop the inlined data objects map in favor of a double map `BagId => DataObjectId => DataObject`. If we don't prevent non-empty bag deletion - we need to iterate through the double-map prefix by BagId. The only iteration would be also on deleting the dynamic bag.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Fixed in https://github.com/Joystream/joystream/pull/2407/commits/1ee491524e7a22cfbb8d2eac2fa4ecfce002b44e

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Fixed in https://github.com/Joystream/joystream/pull/2407/commits/1ee491524e7a22cfbb8d2eac2fa4ecfce002b44e

runtime-modules/storage/src/lib.rs
bedeho: 1. I reviewed this again, and I think the rationale for having this is distinct from the data objects. Remember, even if you have 0 videos in your channel, for example, we still need to give you an incentive to throw out the channel itself. That is what this prize does. This means it is needed, even if we have non-inlined objects. However, I do wonder, should we have the calling code manage this incentive itself, thereby dropping the distinction between dynamic and static bags in this module? In this alternative approach, a channel would have its own storage prize in the content directory, and that prize would be calibrated by that code to give sufficient incentive to cleanup both that channel there, and the dynamic bag in the storage system. Let me know what you think.
3. Yes I think this is what we agreed, this is what  I was trying to say. Note that the storage lead is not the deleter, it has to be the owner of the bag.

runtime-modules/storage/src/lib.rs
bedeho: 1. I understand, that is indeed practical, I think my point in that particular case would be to return some sum type across the two bag types, so that the ensure abstraction is similar to how we typically do it. But I dont know if that causes other problems later.
2. I suggested a way to approach having just a single bag type here https://github.com/Joystream/joystream/pull/2407#discussion_r676404403, if that sounds advisable, then lets do that. However, let us not put static bags into the dynamic map, as that makes the representations much less safe, as static bags can be deleted and so on. Let just use the same approach as now, but with only one bag type.

shamil-gadelshin:
Changes summary:
- replaced DynamicBag and StaticBag structs for a single Bag struct. Added optional deletion prize for a bag
- removed BagManager abstraction
- introduced deletion prize withdrawal and deposit on dynamic bag creation/deletion.
- replaced inlined data objects map with storage double map BagId => DataObjectId => DataObject. I chose the double map over the simple map to avoid introducing an additional map with bag-to-object relations - the double map has it by design.
- introduced total objects size and number counters for bags instead of ad-hoc calculation using inlined data objects map and their changing event.

bedeho (changes requested):
Excellent, only thing I would have us change is to not use the name `BagObject` and `BagDataObjectsById` , and just use `DataObject`.

bedeho (approved):

PR #2416, opened 2021-05-19
query node - data object removal fix

ondratra:
This PR makes sure all connections to `dataObject` are removed first before `dataObject` deletion. That prevents db constraint error.

### DIFF
--- a/query-node/mappings/src/storage.ts
+++ b/query-node/mappings/src/storage.ts
@@ -69,7 +69,11 @@ export async function dataDirectory_ContentRemoved(db: DatabaseManager, event: S
 
   // remove assets from database
   for (let item of dataObjects) {
-      await db.remove<DataObject>(item)
+    // ensure dataObject is nowhere used to prevent db constraint error
+    await disconnectDataObjectRelations(db, item)
+
+    // remove data object
+    await db.remove<DataObject>(item)
   }
 
   // emit log event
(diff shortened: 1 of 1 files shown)


mnaamani:
Looks correct. Its a shame there isn't a simpler way to get the Video asset or Channel asset more directly from the dataObject, but this is no time to fiddle with the schema :)

query-node/mappings/src/storage.ts
```diff
@@ -161,6 +165,46 @@ async function updateSingleConnectedAsset<T extends Channel | Video>(db: Databas
...
+    if (type instanceof Channel) {
+      await db.save<Channel>(item)
+
+      // emit log event
+      logger.info("Content has been disconnected from Channel", {channelId: item.id.toString(), joystreamContentId: dataObject.joystreamContentId})
+    } else {
+      await db.save<Video>(item)
+
+      // emit log event
+      logger.info("Content has been disconnected from Channel", {videoId: item.id.toString(), joystreamContentId: dataObject.joystreamContentId})
```
mnaamani:
```suggestion
      logger.info("Content has been disconnected from Video", {videoId: item.id.toString(), joystreamContentId: dataObject.joystreamContentId})
```

query-node/mappings/src/storage.ts
```diff
@@ -161,6 +165,46 @@ async function updateSingleConnectedAsset<T extends Channel | Video>(db: Databas
...
+  const condition = {
+    where: {
+      [propertyName + 'DataObject']: dataObject
+    }
+  } // as FindConditions<T>
+
+  // in therory the following condition(s) can be generalized `... db.get(type, ...` but in practice it doesn't work :-\
+  const items = type instanceof Channel
+    ? await db.getMany(Channel, condition)
+    : await db.getMany(Video, condition)
```
mnaamani: getMany shouldn't be necessary because a dataobject cannot be re-used for more than one asset, so you could get rid of the iteration in next step and simplify.

query-node/mappings/src/storage.ts
```diff
@@ -161,6 +165,46 @@ async function updateSingleConnectedAsset<T extends Channel | Video>(db: Databas
...
+
+  for (const item of items) {
+    item[propertyName + 'Availability'] = AssetAvailability.INVALID
```
mnaamani: How does Atlas handle INVALID case?
Should we probably add a new variant `AssetAvailability.Deleted` ?

cc: @kdembler

mnaamani (changes requested):
Small typo fix and suggestion for simplification.
I don't know if we can/should deal with new variant for AssetAvailability

query-node/mappings/src/storage.ts
mnaamani: cc: @bwhm

query-node/mappings/src/storage.ts
ondratra: Done

query-node/mappings/src/storage.ts
ondratra: Unnecessary iteration has been removed, and a note about why it is not needed has been added.

query-node/mappings/src/storage.ts
ondratra: It seems that `Deleted` flag is currently not needed by Atlas.

kdembler: Just to reiterate, Atlas treats `Invalid` as "not provided" or "missing". I don't think we need a separate status for deleted

mnaamani (approved):

Issue #2414, opened 2021-05-19
Storage Vouchers for sumer


bwhm:
There are some issues with the `dataDirectory.vouchers` for sumer.

Currently, the default limit for a channel is:
```
  size_limit: 5,400,000,000,
  objects_limit: 1,000,
```
-> you have about 5.4MB per asset.
As about half of these would be videos, that number doesn't really make sense.

To add to that, the maximum allowed size of a single asset in the storage node is currently 10GB. That doesn't make sense for most users, but perhaps it's still worth keeping as the vouchers can be changed instantly (if needed), whereas I assume you have to change the storage nodes to increase the 10GB limit (perhaps even runtime?).

Another issue that isn't worth fixing now, but has to be revisited at some point, is that the vouchers are applied on the channel level. AFAICT, there's no limitations to how many channels a member can make. At the moment, it's of course also "free" to create memberships, so in that sense it makes no difference. Regardless, I think we need to take a look at this. Ideally, fees should be the final arbiter for most things, but I'm not sure if the plan is to have fees depend on asset sizes at all?

My two cents:
- We increase the default `size_limit` or decrease the `objects_limit` for sumer.
- We leave the 10GB as is
- We leave the potential sybil attack of channel/membership for now, but come up with a long term solution to this.

Assigning you for now @bedeho, as I assume you, if anyone, have a long term plan for this.

mnaamani:
The 10GB limit is purely a storage node policy. (Reverse proxies if used in-front of storage node also need to be configured correctly to allow these large uploads, the defaults for nginx and caddy are usually 10MB)

The runtime doesn't put any constraint on an individual object size. I think having a runtime limit might be a good idea though, to avoid situation where runtime allows a large dataobject but storage-nodes don't accept the large upload.

PR #2417, opened 2021-05-19
More generous initial data directory voucher limits

mnaamani:
Addresses some issues raised in https://github.com/Joystream/joystream/issues/2414

Increased initial voucher limits both size and objects.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2332,7 +2332,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
-version = "5.4.0"
+version = "5.5.0"
 dependencies = [
  "frame-benchmarking",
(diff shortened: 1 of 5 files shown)


bwhm (commented):
I think the size limits makes more sense, but I'm not sure about the number of objects assuming they are meant to be "in line".

A user can upload an enormous amount of small objects, and I was of the impression this could be a problem for (joystream) nodes?

mnaamani:
Yes I'm not sure what the right balance of size to number of objects should be really. Certainly a large number of objects will bloat the state. But that is solving we have to solve separately I think and should be addressed in storage v2?

bwhm (approved):
LGTM!

Issue #2419, opened 2021-05-20
WIP: Sumer feedback for Pioneer


bwhm:
Although small, these should be fixed before launch:
1. Remove "support" for gateways in proposals (all working group proposals, really)
2. In /#/working-groups  - the `operations` group lead appear twice.
![Screen Shot 2021-05-20 at 15 52 16](https://user-images.githubusercontent.com/35237943/118990959-9c5b1000-b983-11eb-9eaf-1d1c125fbb62.png)
Go [here](https://pioneer-1.joystream.app/#/working-groups), with endpoint wss://reckless-1.joystream.app/staging/rpc to compare with the others.

@DzhideX 
This is not as urgent, but would be nice if you could add the operations group to the tokenomics page when you have the time!

PR #2421, opened 2021-05-21
Pioneer - Sumer glitch fixes

Lezek123:
Fixes point `1` and `2` from issue: https://github.com/Joystream/joystream/issues/2419

The "lead appearing twice" bug was caused by incorrect handling of `lead.workerId === 0` in some parts of the Pioneer code (which resulted in lead beeing treated simultaneously as both lead and worker)


bwhm (approved):
Fixes the issues :)

PR #2423, opened 2021-05-22
Add Operations group to Tokenomics

DzhideX:
This is the implementation of the Operations group into the Tokenomics page. Relevant issue:  https://github.com/Joystream/joystream/issues/2419

Changes:
- Add operations group to tokenomics transport
- Update `spendingAndStakeDistribution` table and charts to accommodate the new group
- Update necessary types

### DIFF
--- a/pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
+++ b/pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
@@ -123,7 +123,8 @@ type TokenomicsGroup =
   'validators' |
   'council' |
   'storageProviders' |
-  'contentCurators'
+  'contentCurators' |
+  'operations'
 
 const SpendingAndStakeDistributionTable: React.FC<{data?: TokenomicsData; statusData?: StatusServerData | null}> = ({ data, statusData }) => {
   const { width } = useWindowDimensions();
@@ -139,6 +140,8 @@ const SpendingAndStakeDistributionTable: React.FC<{data?: TokenomicsData; status
         : (data && statusData) && `${(data[group][dataType] * Number(statusData.price)).toFixed(2)}`;
     }
   };
+
+  console.log(data);
 
   return (
     <StyledTable divideColumnsAt={[3, 6, 9]} celled>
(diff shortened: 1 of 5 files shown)


pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
```diff
@@ -140,6 +141,8 @@ const SpendingAndStakeDistributionTable: React.FC<{data?: TokenomicsData; status
     }
   };
 
+  console.log(data);
```
Lezek123: Dangling `console.log`

pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
```diff
@@ -229,6 +232,30 @@ const SpendingAndStakeDistributionTable: React.FC<{data?: TokenomicsData; status
...
+          role='Operations Lead'
+          helpContent='Current Operations Lead, and their projected reward and stake.'
+          numberOfActors={data && `${data.operations.lead.number}`}
+          groupEarning={data && `${Math.round(data.operations.lead.rewardsPerWeek)}`}
+          groupEarningDollar={displayStatusData('operations', 'rewardsPerWeek', true)}
+          earningShare={data && `${round(data.operations.lead.rewardsShare * 100)}`}
+          groupStake={data && `${data.operations.lead.totalStake}`}
+          groupStakeDollar={displayStatusData('operations', 'totalStake', true)}
+          stakeShare={data && `${round(data.operations.lead.stakeShare * 100)}`}
+          color='rgb(30, 132, 189)'
```
Lezek123: I noticed we're using the same rgb values here and in `TokenomicsCharts.tsx`.
It's not a priority, but it would be nice to have those colors represented by shared (exported) constants like `OPERATIONS_CHART_COLOR`, `OPERATIONS_LEAD_CHART_COLOR`, `CURATORS_CHART_COLOR` etc. (or maybe `CHART_COLORS.OPERATIONS`, `CHART_COLORS.OPERATIONS_LEAD` etc.) to avoid "magic values" which may lead to potential mismatches.

Lezek123: Some colors are now becoming hard to differentiate and also the color palette doesn't seem well balanced in general, My suggestion would be to use a tool like https://www.materialpalette.com/ in order to generate a palette of colors that would look well together and be easy to distinguish (although this is also quite minor issue)

Lezek123 (approved):
Great job, no changes are really _required_, but added just a few minor improvement suggestions in the comments which you can consider (shouldn't take long if you decide to go for it)

DzhideX:
I've made the change you requested as I agree completely with every comment! I've used the website to find the new colors and I tried to have it be as close to the previous ones as possible (to keep some consistency). Let me know if the colors are ok!

Lezek123 (changes requested):
Great! Actually thanks to that I now noticed we're missing `Operations` group in the `Stake` pie chart (it was only added in `Spending`).

Aside from than that everything looks good.

DzhideX:
I don't how I've missed that, thank you! Should be fixed now.

bwhm (approved):
Works great!

Issue #2411, opened 2021-05-19
Query node - reproducible IDs


ondratra:
At the moment, there are some entities in query node mappings that are created without explicitly defining ID. This is a problem because each processor instance creates a different ID for the same record, and it might cause issues to all depending services. 

For example, video `License` has no representation in runtime since it comes from not-validated metadata, thus creating a predictable ID from license content is required.

```
new Language({
    // proposed solution:
    // id: hash(JSON.stringify(licenseData) + block.hash)

    iso: languageIso,
    createdInBlock: blockNumber,
})
```

Check all entities in mappings and make sure they have a predictable ID generated.

ondratra:
It seems this issue is relevant to Olympia mappings as well.

PR #2433, opened 2021-05-25
Query node predictable ids

ondratra:
This PR adds predictable IDs to entities that previously hadn't predictable IDs. Partially solves #2411 - problem solved for Sumer release, Olympia not covered in this PR.

### DIFF
--- a/query-node/mappings/src/common.ts
+++ b/query-node/mappings/src/common.ts
@@ -45,9 +45,9 @@ export function invalidMetadata(extraInfo: string, data?: unknown): void {
 /*
   Creates a predictable and unique ID for the given content.
 */
-export async function createPredictableId(db: DatabaseManager): Promise<string> {
+export async function getNextId(db: DatabaseManager): Promise<string> {
   // load or create record
-  const existingRecord = await db.get(NextEntityId, {}) || new NextEntityId({id: '0', nextId: 0})
+  const existingRecord = await db.get(NextEntityId, {}) || new NextEntityId({id: '0', nextId: 1})
 
   // remember id
   const entityId = existingRecord.nextId
@@ -74,7 +74,7 @@ export async function prepareDataObject(
   const customContentParameters = new Custom_ContentParameters(registry, contentParameters.toJSON() as any)
 
   const dataObject = new DataObject({
-    id: await createPredictableId(db),
+    id: await getNextId(db),
     owner,
     createdInBlock: event.blockNumber,
     typeId: contentParameters.type_id.toNumber(),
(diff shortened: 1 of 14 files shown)


query-node/mappings/src/common.ts
```diff
@@ -40,20 +41,36 @@ export function invalidMetadata(extraInfo: string, data?: unknown): void {
...
 
+/*
+  Creates a predictable and unique ID for the given content.
+*/
+export function createPredictableId(event: SubstrateEvent, content: string | Object): string {
+  const contentType = typeof content == 'string'
+    ? content
+    : JSON.stringify(content)
+
+  const id = `${event.blockNumber}_${event.index}_${contentType}`
```
mnaamani: If the event index is relative to the transaction and not the block, this is not sufficiently deterministic if there happen to be two extrinsics in the same block that trigger an event at the same index with the same content.
If we go with this id implementation, we should probably add the transaction index as well.

ondratra: Id generation mechanism has been changed and is using simple number that is incremented each time (stored in db as `NextEntityId` entity).

mnaamani:
I think there is some left over code from previous approach of trying to make deterministic IDs, where you are passing the entire event instead of just just event.blockNumber as argument. It doesn't seem necessary as part of this PR, is it useful to keep?

query-node/mappings/src/common.ts
```diff
@@ -41,20 +42,41 @@ export function invalidMetadata(extraInfo: string, data?: unknown): void {
   logger.info(errorMessage, data)
 }
 
+/*
+  Creates a predictable and unique ID for the given content.
+*/
+export async function createPredictableId(db: DatabaseManager): Promise<string> {
```
mnaamani: The term predictable id is not the best name, why not simply.

```suggestion
export async function getNextId(db: DatabaseManager): Promise<string> {
```

query-node/mappings/src/content/video.ts
```diff
@@ -465,6 +467,14 @@ function integrateVideoMediaMetadata(
   // connect encoding to media metadata object
   mediaMetadata.encoding = encoding
 
+  // ensure predictable ids
+  if (!mediaMetadata.encoding.id) {
```
mnaamani, line 471: Does this test really work? What if the id is `0`

ondratra: Yes this could cause a trouble in a rare case. It has been prevented by setting mining `getNextId` result to `1`.

query-node/mappings/src/common.ts
ondratra: The function has been renamed to `getNextId`.

ondratra:
Yes, there was a leftover code. I removed it for this PR, but a similar change (reading from `event`) was also introduced by other PRs. So if you see a similar code in other PR reviews, don't be alarmed.

Issue #2415, opened 2021-05-19
Query node - proper createdAt and updatedAt date


ondratra:
Some entities in mappings don't have `createdAt` and `updatedAt` explicitly set, creating unexpected values and differences between processor instances. `dataObject`, `worker`,  `license`, `videoMetadata` and possibly other entities are affected.

It is likely that these values will be set properly by Hydra (v3) itself, but that's not the case for now and may not be available in v2 that Sumer is using.

The second part of the issue is that the `createdAt` and `updatedAt` values should be provided for bootstrapped workers. In case these values are made only optional for the bootstrap, prevent inconsistency between processor instances.

PR #2437, opened 2021-05-26
Query node create update time

ondratra:
Solves #2415 . Expects #2433 to be merged first.


ondratra:
Predictable `createdAt`/`updatedAt` times for bootstrapped workers is yet to be implemented.

PR #2438, opened 2021-05-26
CLI content commands - output created entity ids

Lezek123:
Affected commands:
- `content:createVideo`
- `content:createChannel`
- `content:createChannelCategory`
- `content:createVideoCategory`

### DIFF
--- a/cli/src/base/ApiCommandBase.ts
+++ b/cli/src/base/ApiCommandBase.ts
@@ -6,7 +6,7 @@ import { getTypeDef, Option, Tuple, TypeRegistry } from '@polkadot/types'
 import { Registry, Codec, CodecArg, TypeDef, TypeDefInfo } from '@polkadot/types/types'
 
 import { Vec, Struct, Enum } from '@polkadot/types/codec'
-import { ApiPromise, WsProvider } from '@polkadot/api'
+import { ApiPromise, SubmittableResult, WsProvider } from '@polkadot/api'
 import { KeyringPair } from '@polkadot/keyring/types'
 import chalk from 'chalk'
 import { InterfaceTypes } from '@polkadot/types/types/registry'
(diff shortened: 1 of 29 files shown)


bwhm (approved):
Haven't reviewed the code, but can confirm that all the four commands works as intended!

cli/src/base/ApiCommandBase.ts
```diff
@@ -424,12 +425,16 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
...
     params: CodecArg[],
     warnOnly = false
-  ): Promise<boolean> {
+  ): Promise<SubmittableResult | false> {
     this.log(chalk.magentaBright(`\nSending ${module}.${method} extrinsic...`))
     const tx = await this.getOriginalApi().tx[module][method](...params)
     return await this.sendAndFollowTx(account, tx, warnOnly)
   }
 
+  public findEvent(result: SubmittableResult, section: string, method: string): Event | undefined {
```
mnaamani, line 439: There is actually a `.findRecord(section, method)` method on the `SubmittableResult` type that i've used here for example https://github.com/Joystream/joystream/blob/master/storage-node/packages/runtime-api/index.js#L336

cli/src/base/ApiCommandBase.ts
```diff
@@ -401,11 +402,11 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
     account: KeyringPair,
     tx: SubmittableExtrinsic<'promise'>,
     warnOnly = false // If specified - only warning will be displayed in case of failure (instead of error beeing thrown)
-  ): Promise<boolean> {
+  ): Promise<SubmittableResult | false> {
```
mnaamani, line 405: I probably would have gone with `<SubmittableResult | null>` which would make a bit more sense but this works too.

mnaamani (approved):
Looks good, only minor suggestion but not critical.

PR #2441, opened 2021-05-26
Query node sudo calls members

ondratra:
This PR ensures proper parameters are extracted from sudo calls. Also, membership events that may be emitted from 2 different extrinsics are supported (for example `MemberUpdatedAboutText` is emitted in both `change_member_about_text` and `update_membership`).


query-node/mappings/src/membership.ts
```diff
@@ -210,7 +210,14 @@ function convertBytesToString(b: Bytes | null): string {
     return ''
   }
 
-  return Buffer.from(b.toU8a(true)).toString()
+  const result = Buffer.from(b.toU8a(true)).toString()
+
+  // prevent utf-8 null character
+  if (result.match(/^\0$/)) {
```
mnaamani, line 216: This is only matching a single null character right, but what if I pass `0x0000` ?

PR #2452, opened 2021-06-01
Fix Mac arm64 build

### LINKED ISSUE #2453: librocksdb-sys build failure on M1
Building joystream-node on Apple M1 (aarch64/arm64) fails because librocksdb-sys crate dependency.

```
thread 'main' panicked at 'Unable to find libclang:
the `libclang` shared library at /usr/local/Cellar/llvm/11.1.0/lib/libclang.dylib could not be opened: 
dlopen(/usr/local/Cellar/llvm/11.1.0/lib/libclang.dylib, 5): no suitable image found.  
Did find:\n\t/usr/local/Cellar/llvm/11.1.0/lib/libclang.dylib: mach-o, but wrong architecture:
stat() failed with errno=20
```

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/2453

Only issue is it requires updating rocksdb so an existing database will not be compatible with the new node and would need to be purged.

Its also important to link to correct libc library from llvm installed by brew. 
If llvm was installed in keg-only mode you maybe need to follow instructions provided by `brew info llvm`

```
llvm: stable 12.0.0 (bottled), HEAD [keg-only]
Next-gen compiler infrastructure
https://llvm.org/
/opt/homebrew/Cellar/llvm/12.0.0_1 (9,789 files, 1.5GB)
  Poured from bottle on 2021-06-01 at 21:19:04
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/llvm.rb
License: Apache-2.0 with LLVM-exception
==> Dependencies
Build: cmake ✔, swig ✔
Required: python@3.9 ✔
==> Options
--HEAD
	Install HEAD version
==> Caveats
To use the bundled libc++ please add the following LDFLAGS:
  LDFLAGS="-L/opt/homebrew/opt/llvm/lib -Wl,-rpath,/opt/homebrew/opt/llvm/lib"

llvm is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have llvm first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/llvm/bin:$PATH"' >> /Users/mokhtar/.bash_profile

For compilers to find llvm you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/llvm/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/llvm/include"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,7 +307,7 @@ dependencies = [
  "async-io",
  "async-lock",
  "async-process",
- "crossbeam-utils 0.8.3",
+ "crossbeam-utils 0.8.5",
  "futures-channel",
  "futures-core",
  "futures-io",
@@ -896,7 +896,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dca26ee1f8d361640700bde38b2c37d8c22b3ce2d360e1fc1c74ea4b0aa7d775"
 dependencies = [
  "cfg-if 1.0.0",
- "crossbeam-utils 0.8.3",
+ "crossbeam-utils 0.8.5",
 ]
 
 [[package]]
@@ -917,8 +917,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "94af6efb46fef72616855b036a624cf27ba656ffc9be1b9a3c931cfc7749a9a9"
 dependencies = [
  "cfg-if 1.0.0",
- "crossbeam-epoch 0.9.3",
- "crossbeam-utils 0.8.3",
+ "crossbeam-epoch 0.9.5",
+ "crossbeam-utils 0.8.5",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


mnaamani:
To manually patch edit the worspace root `Cargo.toml` file and add

```
[patch.crates-io]
librocksdb-sys = { git = "https://github.com/rust-rocksdb/rust-rocksdb", rev = "476cab4f4ed378db5c61ce109be27325d70a29ce" }
```

then update the dependencies in `Cargo.lock` file with:

```
cargo update -p librocksdb-sys
```

mnaamani:
Patching crates-io is not needed. All that is required is to update librocksdb-sys package:

`cargo update -p librocksdb-sys`

Remember the binary format is different so the generated joystream-node cannot use a chain database produced by a node with older version of librocksdb.

Issue #2406, opened 2021-05-15
Sideproject: Devops - Automated network deployment


mnaamani:
# Background

## Deployment tooling

We currently do not have the tooling to automate deployment of a new "joystream" network or even deployment of individual components to an existing network. The process is manual, intricate and time consuming.

The Development team is always relying on the "operations" team to run and deploy such networks and than leads to some delays to start testing new releases. During early release development before all new components are completed some basic local docker-compose environment are used by the development teams to test their changes.This is useful to some extent, and typically only near the end of a release where the various teams have completed features of their products, will a new test network be deployed.

With recent new use of M1 processor based laptops even local docker containers for joystream-node has become unusable.

The lack of automated deployment tools is reflected in the need to write length and detailed guides for the community on how to deploy [validator nodes](https://github.com/Joystream/helpdesk/tree/master/roles/validators) and various other services such as a [storage node](https://github.com/Joystream/helpdesk/tree/master/roles/storage-providers). This documentation needs to be updated often with every release and is also time consuming.

## Integration testing
Integration tests are executed for new pull-requests in a [github workflows](https://github.com/Joystream/joystream/blob/sumer/.github/workflows/run-network-tests.yml) with a very basic network setup (similar to what a local developer would test against) This is not sufficient and ideally the tests should be performed on a more realistic deployment.

# Project Goal

The goal of this project is to develop the tooling for deploying new networks in a configurable and composable fashion for production deployment and testing purposes on both remote and local Kubernetes clusters.

## Requirements

The tooling should facilitate:

- Single manifest file as network configuration with all required configuration parameters for network deployment
- Dynamically create and configure new cluster - geographically dispersed (platform independent, support AWS or Linode, or local cluster (linux/mac docker desktop)) 
- Host machines in cluster to compose multiple architectures if possible.
- Build docker images from source (potentially from multiple git repositories)
- Initialize network genesis state chainspec
- Deploy initial validator joystream-nodes
- Deploy additional nodes and turn into a validator
- Dispatch extrinsics to configure the network with additional state (workers, accounts, members etc)
- Handle domains and ssl certificates required by web services

## Joystream network components
- 3 initial validators (joystream-node)
- 1 RPC (joystream-node)
- 1 Storage Node (joystream-node, colossus, ipfs node)
- 1 Query Node + (joystream-node, indexer and indexer gateway, processor, redis, postgedb, graphql server)
- 1 Pioneer instance (static website)
- 1 Atlas instance (static website)
[...]

PR #2458, opened 2021-06-03
DevOps - Pioneer Build Automation

ahhda:
Adds automatic deployment of Pioneer app using CloudFront and S3

One step closer to #2406

<details>

Screenshot 2021-06-03 at 6 43 12 PM

Screenshot 2021-06-03 at 6 44 06 PM

</details>

### DIFF
--- a/devops/infrastructure/chain-spec-configuration.yml
+++ b/devops/infrastructure/chain-spec-configuration.yml
@@ -3,7 +3,6 @@
 
 - name: Create and copy the chain-spec file
   hosts: all
-  gather_facts: no
 
   tasks:
     - name: Generate chain-spec file and data keys either on localhost or admin server
@@ -16,12 +15,23 @@
 
 - name: Copy secret, auth and start joystream-node service for validators
   hosts: validators
+  gather_facts: no
 
   roles:
     - validators
 
 - name: Configure RPC service and start it
   hosts: rpc
+  gather_facts: no
 
   roles:
     - rpc
+
+- name: Build Pioneer and copy artifacts to S3
+  hosts: build
+  gather_facts: no
+
+  tasks:
+    - include_role:
+        name: admin
+        tasks_from: deploy-pioneer
--- a/devops/infrastructure/main.yml
+++ b/devops/infrastructure/main.yml
@@ -161,6 +161,44 @@ Resources:
       Timeout: '600'
       Count: 4
 
+  S3Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: PublicRead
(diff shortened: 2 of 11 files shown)


devops/infrastructure/chain-spec-configuration.yml
```diff
@@ -16,12 +15,23 @@
...
   roles:
     - validators
 
 - name: Configure RPC service and start it
   hosts: rpc
+  gather_facts: no
 
   roles:
     - rpc
+
```
mnaamani, line 29: The scope of work of this playbook is now more than what it's name suggests. So it should probably get a more "general" name to reflect the tasks it is performing.

devops/infrastructure/main.yml
```diff
@@ -161,6 +161,44 @@ Resources:
...
+            Resource: !Sub "arn:aws:s3:::${S3Bucket}/*"
+      Bucket: !Ref S3Bucket
+
+  CloudFrontDistribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        Origins:
+        - DomainName: !Select [1, !Split ["//", !GetAtt S3Bucket.WebsiteURL]]
+          Id: pioneer-origin-s3
```
mnaamani, line 191: was this something we created manually in AWS console website ?

devops/infrastructure/main.yml
```diff
@@ -161,6 +161,44 @@ Resources:
...
+          CustomOriginConfig:
+            OriginProtocolPolicy: http-only
```
mnaamani, line 193: https only would be better, since most browsers now have a default https only?

ahhda: `OriginProtocolPolicy` specifies the protocol (HTTP or HTTPS) that CloudFront uses to connect to the origin (S3 bucket). Since, S3 only supports HTTP for websites and since this is an internal communication, we use HTTP.

The client side will use HTTPS (set by `ViewerProtocolPolicy`).

devops/infrastructure/main.yml
ahhda: This is just for CloudFront distribution reference and not created manually.

devops/infrastructure/chain-spec-configuration.yml
ahhda: Changed to `chain-spec-pioneer.yml`

mnaamani:
You changed the name of the playbook to `chain-spec-pioneer.yml` but forgot to update it in `deploy-infra.sh`

Stack deletion script works as expected. :+1:

PR #2459, opened 2021-06-04
DevOps - Deployment improvements

ahhda:
* Add auto scaling groups to define variable number of Validator instances
* Change name of CloudFormation file from `main.yml` to `infrastructure.yml`
* Build only on build instance and copy binary to other nodes
* Ability to specify different instance type for build instance, rpc and node instances

One step closer to #2406 

<details>
<summary>Manual Testing:</summary>

Screenshot 2021-06-07 at 1 53 42 PM

Screenshot 2021-06-07 at 1 53 35 PM

Screenshot 2021-06-07 at 1 53 32 PM

</details>


devops/infrastructure/bash-config.sample.cfg
```diff
@@ -5,7 +5,8 @@ REGION=us-east-1
 CLI_PROFILE=joystream-user
 KEY_PATH="/Users/joystream/Joystream/joystream-key.pem"
 AWS_KEY_PAIR_NAME="joystream-key"
-EC2_INSTANCE_TYPE=t2.micro
+NODE_EC2_INSTANCE_TYPE=t2.micro
```
mnaamani: I see this is the type used for the "InstanceLaunchTemplate", and RPCInstance doesn't specify the instance type so its taken from the template. Therefore maybe a better name to reflect this, like `DEFAULT_EC2_INSTANCE_TYPE`

devops/infrastructure/bash-config.sample.cfg
```diff
@@ -5,7 +5,8 @@ REGION=us-east-1
...
+BUILD_EC2_INSTANCE_TYPE=t2.xlarge
```
mnaamani, line 10: We should have a variable for each "role" so in addition to BUILD_EC2_INSTANCE_TYPE:
- VALIDATOR_EC2_INSTANCE_TYPE
- RPC_EC2_INSTANCE_TYPE

and if a value is not set, the colud formation template can fall back to the default.

devops/infrastructure/bash-config.sample.cfg
ahhda: Added DEFAULT_EC2_INSTANCE_TYPE

devops/infrastructure/bash-config.sample.cfg
ahhda: Added option to set RPC, Validator and Build instance types

PR #2460, opened 2021-06-04
Update Weekly Top Ups Value

DzhideX:
Weekly top ups value is changed to reflect the time between replenishments changing from biweekly to weekly.


bwhm (approved):
LGTM!

PR #2473, opened 2021-06-11
DevOps - Deploy a new joystream-node to an existing network

ahhda:
Automated deployment of an EC2 instance with joystream-node and chain spec file

Related to #2406 

To run this:

* Edit `bash-config.sample.cfg` and update the parameters `CHAIN_SPEC_FILE` and `BINARY_FILE` with either the URL or local file path
* Run `./deploy-single-node.sh ~/Joystream/bash-config.sample.cfg`

### DIFF
--- a/devops/infrastructure/bash-config.sample.cfg
+++ b/devops/infrastructure/bash-config.sample.cfg
@@ -11,7 +11,7 @@ BUILD_EC2_INSTANCE_TYPE=t2.xlarge
 RPC_EC2_INSTANCE_TYPE=t2.micro
 
 # Set a prebuilt AMI if required
-EC2_AMI_ID="ami-064f99551235fb1ac"
+EC2_AMI_ID="ami-08ffec5991ca99db9"
 
 ACCOUNT_ID=$(aws sts get-caller-identity --profile $CLI_PROFILE --query Account --output text)
 
--- a/devops/infrastructure/bash-config.sample.cfg
+++ b/devops/infrastructure/bash-config.sample.cfg
@@ -23,6 +23,14 @@ INVENTORY_PATH="$DATA_PATH/inventory"
 
 NUMBER_OF_VALIDATORS=2
 
+## Used for Deploying a new node
+DATE_TIME=$(date +"%d-%b-%Y-%H-%M-%S")
+
+SINGLE_NODE_STACK_NAME="new-node-$DATE_TIME"
+
+BINARY_FILE="https://github.com/Joystream/joystream/releases/download/v9.3.0/joystream-node-5.1.0-9d9e77751-x86_64-linux-gnu.tar.gz"
+CHAIN_SPEC_FILE="https://github.com/Joystream/joystream/releases/download/v9.3.0/joy-testnet-5.json"
+
 #### PARAMETERS USED BY ANSIBLE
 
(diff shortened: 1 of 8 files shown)


mnaamani:
I was trying to do a deploy of new network before deploying the new single instance but got this:

```
Starting galaxy role install process
[WARNING]: - caddy_ansible.caddy_ansible (v3.0.4) is already installed - use --force to change version to unspecified
Starting galaxy collection install process
Nothing to do. All requested collections are already installed. If you want to reinstall them, consider using `--force`.
./deploy-infra.sh: line 47: get_aws_export: command not found
```

devops/infrastructure/deploy-single-node.sh
```diff
@@ -0,0 +1,50 @@
...
+    EC2InstanceType=$DEFAULT_EC2_INSTANCE_TYPE \
+    KeyName=$AWS_KEY_PAIR_NAME \
+    EC2AMI=$EC2_AMI_ID
+
+# If the deploy succeeded, get the IP and configure the created instance
+if [ $? -eq 0 ]; then
+  # Install additional Ansible roles from requirements
+  ansible-galaxy install -r requirements.yml
+
+  SERVER_IP=$(get_aws_export "PublicIp")
```
mnaamani: I think the get_aws_export function is missing

devops/infrastructure/bash-config.sample.cfg
```diff
@@ -33,3 +41,11 @@ BRANCH_NAME=sumer
 
 # If true will build LOCAL_CODE_PATH otherwise will pull from GIT_REPO:BRANCH_NAME
 BUILD_LOCAL_CODE=false
+
+get_aws_export () {
```
mnaamani: oh its here :)
I think it would be better to leave this in a common file other than the cfg, and maybe have the NEW_STACK_NAME and $1 be function arguments instead

devops/infrastructure/deploy-single-node.sh
ahhda: Moved functions to `common.sh`

devops/infrastructure/bash-config.sample.cfg
ahhda: Stack name and export name are parameters now

devops/infrastructure/delete-stack.sh
```diff
@@ -11,15 +13,7 @@ else
...
-get_aws_export () {
-  RESULT=$(aws cloudformation list-exports \
-    --profile $CLI_PROFILE \
-    --query "Exports[?starts_with(Name,'${NEW_STACK_NAME}$1')].Value" \
-    --output text | sed 's/\t\t*/\n/g')
-  echo -e $RESULT | tr " " "\n"
-}
-
-BUCKET_NAME=$(get_aws_export "S3BucketName")
+BUCKET_NAME=$(get_aws_export $NEW_STACK_NAME "S3BucketName")
```
mnaamani, line 16: I'm getting an empty bucket name here. So when trying to empty the bucket:

```
=========== Emptying bucket  ===========
fatal error: Parameter validation failed:
Invalid bucket name ""
```

mnaamani: seems this was an intermittent network issue communicating with aws

PR #2474, opened 2021-06-11
Working groups mappings - worker status and missingRewardAmount fix

Lezek123:
- Split `WorkerStatusLeft` into `WorkerStatusLeaving` and `WorkerStatusLeft` for easier parsing
- Clear `worker.missingRewardAmount` on `WorkerExited` event

### DIFF
--- a/query-node/mappings/workingGroups.ts
+++ b/query-node/mappings/workingGroups.ts
@@ -71,6 +71,7 @@ import {
   BudgetSetEvent,
   BudgetSpendingEvent,
   LeaderSetEvent,
+  WorkerStatusLeaving,
 } from 'query-node/dist/model'
 import { createType } from '@joystream/types'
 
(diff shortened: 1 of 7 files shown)


query-node/mappings/workingGroups.ts
```diff
@@ -800,9 +801,15 @@ export async function workingGroups_WorkerExited({ store, event }: EventContext
...
-  ;(worker.status as WorkerStatusLeft).workerExitedEventId = workerExitedEvent.id
+
+  const newStatus = new WorkerStatusLeft()
+  newStatus.workerStartedLeavingEventId = (worker.status as WorkerStatusLeaving).workerStartedLeavingEventId
+  newStatus.workerExitedEventId = workerExitedEvent.id
+
+  worker.status = newStatus
   worker.stake = new BN(0)
   worker.rewardPerBlock = new BN(0)
+  worker.missingRewardAmount = undefined
```
jodator, line 812: Just to be sure – this means that, if the worker left the owed reward is disregarded?

jodator: The scenario I'm thinking of is:

1.  A worker has missed reward due to an insufficient WG budget.
2.  After the unstaking period there's still: a) missing reward and b) insufficient budget

I think that there will not be payment if the worker is not in the working group, so I think that I have my answer :)

jodator (approved):
I think that I understand the change. It will be easier to differentiate _leaving_ and _left_ status for sure 👍 

The other thing that I see is that any worker that initiated leaving is already considered out of the Working Group (should not be displayed as "active").

query-node/mappings/workingGroups.ts
Lezek123: Judging by the code in https://github.com/Joystream/joystream/blob/olympia/runtime-modules/working-group/src/lib.rs#L1262, the runtime will try to pay the missed reward just before removing the worker. Once the worker is removed, any remaining missed reward is "gone" (since the worker is no loger part of the `WorkerById` map so the runtime effectively doesn't recognize it anymore), therefore it should be no longer part of the debt.

mnaamani (approved):

Issue #2306, opened 2021-03-31
Colossus: Doesn't gracefully handle committing file to IPFS when IPFS node is down


mnaamani:
Colossus exits if tries to commit a new upload when no IPFS daemon is running. 

Simplest fix catch the error, return `5xx` http response code, uploader will need to re-attempt upload.

Alternate fix, return 5xx http response code, but keep the temporary file in cache, re-try to commit to ipfs after some delay. Uploader will not need to re-try but instead wait to see asset 'Accepted' state.

First approach fits better with current Atlas UI/UX design.

```
  joystream:storage:storage Error committing stream Error: connect ECONNREFUSED 127.0.0.1:5001
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16) {
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5001
} +8ms
  joystream:colossus:api:asset Error: connect ECONNREFUSED 127.0.0.1:5001
  joystream:colossus:api:asset     at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16) +0ms
RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: ECONNREFUSED
    at ServerResponse.writeHead (_http_server.js:248:11)
    at ServerResponse._implicitHeader (_http_server.js:239:8)
    at write_ (_http_outgoing.js:642:9)
    at ServerResponse.end (_http_outgoing.js:753:5)
    at ServerResponse.send (/Users/mokhtar/joystream/joystream/node_modules/express/lib/response.js:221:10)
    at ServerResponse.expressOpenAPISend [as send] (/Users/mokhtar/joystream/joystream/storage-node/packages/colossus/lib/middleware/validate_responses.js:34:21)
    at ServerResponse.json (/Users/mokhtar/joystream/joystream/node_modules/express/lib/response.js:267:15)
    at ServerResponse.send (/Users/mokhtar/joystream/joystream/node_modules/express/lib/response.js:158:21)
    at ServerResponse.expressOpenAPISend [as send] (/Users/mokhtar/joystream/joystream/storage-node/packages/colossus/lib/middleware/validate_responses.js:54:21)
    at errorHandler (/Users/mokhtar/joystream/joystream/storage-node/packages/colossus/paths/asset/v0/{id}.js:28:44)
[...]

PR #2476, opened 2021-06-11
storage-node: ensure write stream flushed correctly

mnaamani:
Bug Fixes:
- On heavy load and large file uploads Colossus was sometimes not matching IPFS hash of file correctly. This was due to the hash being calculated early, before the entire temporary stream was flushed to disk. Fixed by calling callbacks after writes and flushes were complete.
- Added error handling for write errors to temporary stream, and proper handling of backpressure
- Fixed early file type detection check during upload - buffer length was not correctly checked to know if it was time to test
- Fix for  https://github.com/Joystream/joystream/issues/2306 - adjust errorHandler to deal with error objects that may not contain a proper http numeric status code

### DIFF
--- a/.github/workflows/run-network-tests.yml
+++ b/.github/workflows/run-network-tests.yml
@@ -193,7 +193,7 @@ jobs:
           docker-compose up -d joystream-node
       - name: Configure and start development storage node
         run: |
-          DEBUG=* yarn storage-cli dev-init
+          DEBUG=joystream:* yarn storage-cli dev-init
           docker-compose up -d colossus
       - name: Test uploading
         run: |
--- a/storage-node/packages/colossus/paths/asset/v0/{id}.js
+++ b/storage-node/packages/colossus/paths/asset/v0/{id}.js
@@ -25,7 +25,9 @@ const assert = require('assert')
 
 function errorHandler(response, err, code) {
   debug(err)
-  response.status(err.code || code || 500).send({ message: err.toString() })
+  // Some err types don't have a valid http status code such as one that come from ipfs node for example
+  const statusCode = typeof err.code === 'number' ? err.code : code
+  response.status(statusCode || 500).send({ message: err.toString() })
   response.end()
 }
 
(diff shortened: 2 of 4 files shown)


kdembler (commented):
My review was requested here but I don't really know much about our storage node and node.js streams. I don't see anything jarring but it's possible that I couldn't tell if there was. I think we should probably launch some instance of storage node with this change and test it out?

mnaamani:
> My review was requested here but I don't really know much about our storage node and node.js streams. I don't see anything jarring but it's possible that I couldn't tell if there was. I think we should probably launch some instance of storage node with this change and test it out?

I agree. I think @shamil-gadelshin will do a code review and then ask @bwhm to deploy on staging network where it can be tested interactively.

Some additional tests I can suggest for Shamil to see if my fixes do indeed resolve the problem:

1- Storage subsystem unit test. I re-enabled a unit test which was always problematic (because of incorrect handling of stream events)

You can run it with:

```
docker-compose up -d ipfs
cd storage-node/packages/storage
yarn test
```

You can see how it used to fail if you checkout master branch and run same test.

Also un-comment the test code https://github.com/Joystream/joystream/blob/20d73e12f6e610fb0eb4abc94f08a67f2c386e4f/storage-node/packages/storage/test/storage.js#L93 
To see more tests failing.

2 - You can run a basic bash script like this to send a bunch of concurrent uploads to dome some stress testing also and try to make the storage node fail.

```
#!/usr/bin/env bash

FILE=~/Downloads/large-video.mp4
# Allow time for tx to enter mempool to avoid nonce re-use
SLEEP=2

for i in {1..20}
do
    yarn storage-cli upload $FILE 1 0 &
    sleep $SLEEP
done
```

After starting dev node:
```
docker-compose up -d joystream-node
docker-compose up -d ipfs
yarn storage-cli dev-init
DEBUG=joystream:* yarn colossus --dev
```

shamil-gadelshin (approved):
LGTM

I tested the differences between old and new versions. Both unit tests and provided bash script work correctly in the new version and have some issues in the old one.

PR #2461, opened 2021-06-04
query node - GraphQL Playground query templates

ondratra:
This PR overloads dependency on GraphQL Playground to the forked version that features custom query templates. The actual Playground changes are in PR https://github.com/ondratra/graphql-playground/pull/2

PR #2449, opened 2021-05-31
Add QueryNode linting

mnaamani:
Replacing https://github.com/Joystream/joystream/pull/2404 - just includes adding the linting checks, but without checking in any of the code changes. Will leave that to @ondratra 

```
# to lint and auto-fix
yarn workspace query-node-mappings lint --fix

# apply prettier formatting
yarn workspace query-node-mappings format
```

PR #2485, opened 2021-06-18
Query node sumer features lint

ondratra:
This PR collects works from PRs #2433, #2437, #2461, and #2449 ~(they should be merged first for easy reviews)~. On top of it, it contains linting of existing query node source code.

**Edit:** now also includes work from #2538, #2495.
**Edit:** ~see a comment below about the need to update Atlas when deploying code with this PR~ (Atlas and query node is already running in both production and dev servers)


ondratra:
Looking at the current state of PRs related to this PR, I think it will be best to merge all at once via this PR rather than merging each of them separately.

query-node/mappings/src/common.ts
```diff
@@ -188,3 +208,21 @@ class Logger {
...
+*/
+export function convertBytesToString(b: Bytes | null): string {
+  if (!b) {
+    return ''
+  }
+
+  const result = Buffer.from(b.toU8a(true)).toString()
+
+  // prevent utf-8 null character
+  if (result.match(/^\0$/)) {
```
Lezek123: I don't think this will work for multiple null characters, for example: `0x000000`

image

ondratra:
Deploying the code including this PR to servers should be done together with deploying Atlas version including this patch https://github.com/Joystream/atlas/pull/1037

query-node/mappings/src/common.ts
ondratra: I feel like we fixed this issue couple of times before, but here we are again... Thx for spotting. Fixed

query-node/mappings/src/common.ts
```diff
@@ -220,7 +220,7 @@ export function convertBytesToString(b: Bytes | null): string {
   const result = Buffer.from(b.toU8a(true)).toString()
 
   // prevent utf-8 null character
-  if (result.match(/^\0$/)) {
+  if (result.match(/^\0+$/)) {
```
Lezek123: This is still a problem for strings that contain other characters.
For example I tried sending `createType('Bytes', 'abcabcabc\0\0\0\0\0')` as `handle` arg for `buyMembership` and it still crashed the query-node.

I would really recommend the well-tested `replace(/\u0000/g, '')` solution I used for olympia (https://github.com/Joystream/joystream/blob/olympia_dev/query-node/mappings/common.ts#L242)

Lezek123 (changes requested):

query-node/mappings/src/common.ts
ondratra: Thx for double-checking. `replace(/\u0000/g, '')` has been set there - fixed.

ondratra:
Atlas and query nodes are already running in both production and dev servers.

PR #2486, opened 2021-06-18
DevOps - Storage node deployment on EKS with Kubernetes and Pulumi

ahhda:
This PR will have the infrastructure and deployment code for setting up a storage node on AWS EKS using Kubernetes and Pulumi.

Related to #2406 

Steps to run:
```bash
pulumi login --local
pulumi up -y
```


devops/infrastructure/kubernetes/index.ts
```diff
@@ -0,0 +1,128 @@
...
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig
+
+// Create a repository
+const repo = new awsx.ecr.Repository('my-repo')
+
+// Build an image and publish it to our ECR repository.
+
+export const colossusImage = repo.buildAndPushImage({
```
mnaamani: I think you mentioned it will be possible to specify an image that could be fetched from a remote repo?
Building locally is fine for now, but it would be worth trying to optimize the size of the built image.

devops/infrastructure/kubernetes/index.ts
```diff
@@ -0,0 +1,128 @@
...
+      selector: { matchLabels: appLabels },
+      template: {
+        metadata: {
+          labels: appLabels,
+        },
+        spec: {
+          containers: [
+            {
+              name: 'ipfs',
+              image: 'ipfs/go-ipfs:latest',
```
mnaamani: We need to be able to configure the ipfs daemon as we do in the docker-compose file: https://github.com/Joystream/joystream/blob/65e98e921e04ae26f685bf229a20f80801508d85/docker-compose.yml#L33

devops/infrastructure/kubernetes/index.ts
```diff
@@ -0,0 +1,128 @@
...
+              image: colossusImage,
+              env: [
+                {
+                  name: 'WS_PROVIDER_ENDPOINT_URI',
+                  // example 'wss://18.209.241.63.nip.io/'
+                  value: process.env.WS_PROVIDER_ENDPOINT_URI,
+                },
+                {
+                  name: 'DEBUG',
+                  value: '*',
```
mnaamani: polkadot-js api debug logs are very verbose. I suggest only the logs specific to colossus:

```suggestion
                  value: 'joystream:*',
```

devops/infrastructure/kubernetes/index.ts
```diff
@@ -0,0 +1,128 @@
...
+                },
+              ],
+              command: [
+                'yarn',
+                'colossus',
+                '--dev',
```
mnaamani:
```suggestion
                '--anonymous',
```

Works better for test setup. This will just run a replicator node.

But ideally we want to be able to provide arguments:
```
    --provider-id ID, -i ID     StorageProviderId assigned to you in working group.
    --key-file FILE             JSON key export file to use as the storage provider (role account).
    --public-url=URL, -u URL    API Public URL to announce.
```

We either need to find a way securely mount a volume to provide the keyfile, or modify colossus to take the raw key from an env variable.

The public url will also either need to be known ahead of time or we can use `nip.io`

devops/infrastructure/kubernetes/index.ts
```diff
@@ -0,0 +1,128 @@
...
+    metadata: {
+      labels: appLabels,
+      namespace: namespaceName,
+    },
+    spec: {
+      type: 'LoadBalancer',
+      ports: [
+        { name: 'port-1', port: 5001 },
+        { name: 'port-2', port: 8080 },
+        { name: 'port-3', port: 3001 },
```
mnaamani: nginx only needs to publicly expose the port 3001. the ipfs ports 5001 and 8080 should only be reachable by colossus

mnaamani (changes requested):

devops/infrastructure/kubernetes/index.ts
ahhda: Added the config commands

devops/infrastructure/kubernetes/index.ts
ahhda: Done

devops/infrastructure/kubernetes/index.ts
ahhda: Done

devops/infrastructure/kubernetes/index.ts
ahhda: As discussed, we will be building locally and pushing. Will try to optimize the image.

devops/infrastructure/kubernetes/index.ts
ahhda: Added the optional parameters

devops/infrastructure/storage-node/index.ts
```diff
@@ -0,0 +1,250 @@
...
+
+// Export the Service name and public LoadBalancer Endpoint
+export const serviceName = service.metadata.name
+// When "done", this will print the hostname
+export let serviceHostname: pulumi.Output<string>
+serviceHostname = service.status.loadBalancer.ingress[0].hostname
+
+export let appLink: pulumi.Output<string>
+
+const publicUrlInput: pulumi.Input<string> = pulumi.interpolate`http://${serviceHostname}/`
```
mnaamani:
```suggestion
const publicUrlInput: pulumi.Input<string> = pulumi.interpolate`https://${serviceHostname}/`
```

PR #2487, opened 2021-06-20
Update election salt function to use proven crypto implementation

DzhideX:
This PR aims to change the current uuid implementation for the salt used in voting and revealing votes into a cryptographically secure random number generator. Issue further explained here: https://github.com/Joystream/audits/issues/4

After doing some research, I found the best/easiest option would be to use the browsers crypto api for this but upon further research I've found that polkadot in their own `util-crypto` random functions uses this same api. 

Due to this I've decided to go with `randomAsHex()`. Without any parameters, the size of the sequence will be 32 bytes which is coincidentally exactly the limit the current olympia runtime currently has for it. I've tested it with [this code](https://github.com/Lezek123/substrate-runtime-joystream/blob/olympia-proposals-mappings/tests/integration-tests/src/fixtures/council/ElectCouncilFixture.tshttps://github.com/Lezek123/substrate-runtime-joystream/blob/olympia-proposals-mappings/tests/integration-tests/src/fixtures/council/ElectCouncilFixture.ts) to make sure that it indeed works!

### DIFF
--- a/pioneer/packages/joy-election/src/VoteForm.tsx
+++ b/pioneer/packages/joy-election/src/VoteForm.tsx
@@ -1,5 +1,5 @@
 import BN from 'bn.js';
-import uuid from 'uuid/v4';
+import { randomAsHex } from '@polkadot/util-crypto';
 
 import React from 'react';
 import { Message, Table } from 'semantic-ui-react';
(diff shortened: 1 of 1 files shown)


shamil-gadelshin (approved):
LGTM

PR #2495, opened 2021-06-22
query node - set featured video fix

### LINKED ISSUE #2494: Sumer query node featured videos update bug
There is a bug in Sumer's query node that will likely prevent unsetting featured videos and changing featured videos (old featured videos will remain set). The root of cause is comparing `videoId.toString()` (used when creating video) to `videoId.toHex()`.

ondratra:
Fixes #2494.

### DIFF
--- a/query-node/mappings/src/content/video.ts
+++ b/query-node/mappings/src/content/video.ts
@@ -367,13 +367,13 @@ export async function content_FeaturedVideosSet(
   // calculate diff sets
   const toRemove = existingFeaturedVideos.filter(existingFV =>
     !videoIds
-      .map(item => item.toHex())
+      .map(item => item.toString())
       .some(isSame(existingFV.id))
   )
   const toAdd = videoIds.filter(video =>
     !existingFeaturedVideos
       .map(item => item.id)
-      .some(isSame(video.toHex()))
+      .some(isSame(video.toString()))
   )
 
   // mark previously featured videos as not-featured
--- a/query-node/mappings/src/content/video.ts
+++ b/query-node/mappings/src/content/video.ts
@@ -376,20 +376,28 @@ export async function content_FeaturedVideosSet(
       .some(isSame(video.toString()))
   )
 
+  // escape if no featured video needs to be added or removed
+  if (!toRemove.length && !toAdd.length) {
+    // emit log event
+    logger.info('Featured videos unchanged')
+
+    return
+  }
+
   // mark previously featured videos as not-featured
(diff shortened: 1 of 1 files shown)


Lezek123:
I tried to reproduce the potential bug from https://github.com/Joystream/joystream/issues/2494 using `master` branch first and it seemed like the bug is not reallly occuring.

Then I found out that there's something different going on:

```
  const toRemove = existingFeaturedVideos.filter(existingFV =>
    !videoIds
      .map(item => item.toHex())
      .some(isSame(existingFV.id))
  )
```

Because no existing video id matched any of the new featured video ids (`videoIds`) converted to hex, `toRemove` always contained all currently featured videos

```
  const toAdd = videoIds.filter(video =>
    !existingFeaturedVideos
      .map(item => item.id)
      .some(isSame(video.toHex()))
  )
```

`toAdd` was on the other hand always populated with all new featured videos ids (`videoIds`), since the filter funcion here was always returning `true`

So it actually was working "correctly" before, in the sense that one could unset currently featured videos and set the new ones. It was just slightly less optimized  (since the videos that were already featured were updated with `isFeatured = true` etc.) and some code was not doing what it was supposed to.

This part of code could be even further optimized with `Promise.all` instead of `await` inside a loop (see https://github.com/Joystream/joystream/pull/2497/commits/57d99c7ecd755bf9d8eb1fdf0f14482a9f89867f#diff-dadee0ee7bb0a5e1d5df4e4d0ee2d03c7b303b48ec039747e1df88c9d0da93baR222), but that's probably not a high priority.

Other than that, everthing is good now, tested a few combinations and didn't encounter any issues :+1:

Lezek123:
Just an additional note: https://github.com/Joystream/joystream/blob/master/query-node/mappings/src/content/video.ts#L390 - the code inside this `if` will never be executed, because `toAdd` is an array

Lezek123 (approved):

mnaamani:
Should this be targeting master branch? @ondratra

mnaamani:
> Just an additional note: https://github.com/Joystream/joystream/blob/master/query-node/mappings/src/content/video.ts#L390 - the code inside this `if` will never be executed, because `toAdd` is an array

Do you want to remove the redundant code block @ondratra ?

ondratra:
Bad condition `if (toAdd) {` has been fixed. Concurrent promises have been utiliezed.

Lezek123 (approved):
LGTM
Tested locally, works as expected :+1:

PR #2468, opened 2021-06-10
Channel payouts

dobertRowneySr:
I have added `update_max_reward` and `update_min_cashout` in order to keep track of the thresholds for the maximum total reward and minimum cashout allowed for a channel.
I have also updated the extrinsic to update the channel commitment over `F` (set containing the possible payments outstanding) in such a way that it is verified that the origin is the root origin.
Finally I wrote the extrinsic to update the channel reward (what it's actually missing is the code for transferring the amount of `JOY`s to the reward account for the Channel). 
Unit tests are in the merkle.rs file

PR #2498, opened 2021-06-23
Channel payouts 2

dobertRowneySr:
I erroneously merged `master` into `channel_payouts`. This is the same as #2468 without the `master` branch

### DIFF
--- a/runtime-modules/content/src/errors.rs
+++ b/runtime-modules/content/src/errors.rs
@@ -58,7 +58,7 @@ decl_error! {
         /// Video does not exist
         VideoDoesNotExist,
 
-        /// Video in season can`t be removed (because order is important)
+        /// Vfdeo in season can`t be removed (because order is important)
         VideoInSeason,
 
         /// Curators can only censor non-curator group owned channels
--- a/runtime-modules/content/src/errors.rs
+++ b/runtime-modules/content/src/errors.rs
@@ -62,6 +62,9 @@ decl_error! {
         VideoInSeason,
 
         /// Curators can only censor non-curator group owned channels
-        CannotCensoreCuratorGroupOwnedChannels
+        CannotCensoreCuratorGroupOwnedChannels,
+
+        /// Actor is not A Member
+        ActorNotAMember
     }
 }
--- a/runtime-modules/content/src/lemma.rs
+++ b/runtime-modules/content/src/lemma.rs
@@ -1,0 +1,60 @@
+/// lemma structure for the merkle proof path:
+/// `(data, path)` with
+/// `path = [ item1 item2 ... itemN ]` and `item = (hash, side)`
+/// `data` is the initial value whose proof membership is to be established
+/// item contains the hash value required for the proof together with the side, that is the provided
+/// hash is to the left or to the right to the current computed hash during verification.
+use codec::{Decode, Encode};
+use core::fmt::Debug;
+
+#[cfg(feature = "std")]
+pub use serde::{Deserialize, Serialize};
+
+use sp_runtime::traits::Hash;
+use sp_std::vec::Vec;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Side {
+    Left,
+    Right,
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct LemmaItem<HashOutput> {
+    pub hash: HashOutput,
+    pub side: Side,
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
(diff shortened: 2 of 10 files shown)


runtime-modules/content/src/errors.rs
```diff
@@ -58,7 +58,7 @@ decl_error! {
         /// Video does not exist
         VideoDoesNotExist,
 
-        /// Video in season can`t be removed (because order is important)
+        /// Vfdeo in season can`t be removed (because order is important)
```
bedeho, line 61: Don't make changes unrelated to the core purpose of the PR.

runtime-modules/content/src/lemma.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct LemmaItem<HashOutput> {
+    pub hash: HashOutput,
+    pub side: Side,
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct Proof<Algorithm: Hash, Value> {
+    pub data: Value,
```
bedeho: ´Proof`=> `MerkleProof`
`data` => `leaf`

runtime-modules/content/src/lib.rs
```diff
@@ -456,6 +476,23 @@ pub struct Person<MemberId> {
     controlled_by: PersonController<MemberId>,
 }
 
+/// Payment claim by a channel
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, Debug)]
+pub struct PullPaymentElementRecord<ChannelId, Balance, HashType> {
```
bedeho: `Record` is a postfix only used for types that are used as storage values, this is not.

runtime-modules/content/src/lib.rs
```diff
@@ -1259,8 +1307,72 @@ decl_module! {
...
+        ) {
+        ensure_root(origin)?;
+            let old_commitment = <Commitment<T>>::get();
+            if old_commitment == new_commitment { return Ok(()) }
+            <Commitment<T>>::put(new_commitment);
+            Self::deposit_event(RawEvent::CommitmentUpdated(new_commitment));
+        }
+
+    #[weight = 10_000_000]
+    pub fn update_channel_reward(
```
bedeho: Check the structure of other extrinsics, there should be a clear distinction between checking and mutation, and we introduce a `mutation safe` marker at the top lexical scope of each extrinsic to show this.

runtime-modules/content/src/lib.rs
```diff
@@ -1259,8 +1307,72 @@ decl_module! {
...
+            origin,
+            proof: PullPaymentProof<T>,
+    ) {
+
+        let signing_acc = ensure_signed(origin)?;
```
bedeho: Drop all mutable variables, they are not needed here.

runtime-modules/content/src/lib.rs
```diff
@@ -1259,8 +1307,72 @@ decl_module! {
...
+        let elem = &proof.data;
+
+        let mut channel = ChannelById::<T>::get(elem.channel_id);
```
bedeho: This will not work, this will always return a channel, even if there if there is no pair in the map, because it returns a default constructed channel. 

1. introduce a test that fails as a result of this bug
2. check how channel checking is done in other extrinsics, e.g. `update_channel`.
3. confirm new test now passes

runtime-modules/content/src/lib.rs
```diff
@@ -1259,8 +1307,72 @@ decl_module! {
...
+
+        if let Some(channel_acc) = channel.reward_account.clone() {
+
+            let cashout = elem
+        .amount_due
+        .checked_sub(&channel.cumulative_reward)
+        .ok_or("uinteger underflow")?;
+
+            ensure!(channel_acc == signing_acc, "Origin isn't channel's reward account");
```
bedeho: You do not sign with reward account of channel, you sign with owner account of channel, check comment about extrinsic signature.

runtime-modules/content/src/lib.rs
bedeho: 1) Lets rename this to `claim_channel_reward`.
2) The origin for this has to be the owner of the channel, and since we should only allow channels owned by members, vs. those owned by curators, the owner has to be a member. So signature needs a member id. Other extrinsics where the caller can be either a member or a curator, use `actor: ContentActor`, but this is not appropriate here.

bedeho (changes requested):
Looks promising, only minor fixes required.

dobertRowneySr:
I made basically one commit for every correction you asked.
I added a function to `permission/mod.rs` in order to verify that:
- `origin` is signed by a Member `AccountId` 
- such member coincides with the channel's owner.

When updating the channel I used the extrinsic `update_channel` as template model.

mnaamani:
I think it makes sense to rebase this work on master branch and target master branch as well instead of sumer branch.

dobertRowneySr:
Closing in favor of Olympia work

PR #2502, opened 2021-06-24
maser -> olympia-staging merge adjustments

Lezek123:
Updates and adjustments after `master` -> `olympia-staging` merge for:
- bash/devops scripts 
- query node schema & mappings
- CLI
- integration tests
- protobuf library


mnaamani:
Great stuff with this merge. I managed to go through it all, it was a bit difficult to keep track because it included some new work for olympia as well as the merge conflict. 

I managed to test query-node, the changes in some of the scripts, protobuf. Only issued I had with integration tests is the proposals flow seems to always fail, this seems to be happening in CI as well so I'm not sure if they are expected to be working yet?

Locally I had to tweak my joystream-node docker file to build on mac m1 laptop. The "fix" is in master now so we could probably update.

There are a couple files that don't seem to belong anymore in the root:
```
build-packages.sh  (replaced with build-npm-packages.sh)
rome-testnet.json
```

The root npm script: 
"build": "./build-npm-packages.sh && ./build-docker-images.sh",

Order should probably be reversed if query-node build depends on joystream-node starting up from the build-docker-images.sh script. But I do have an issue with the build actually resulting in a docker container running. Its probably better to remove this script and have clear documentation in README about build steps requirements.

Building the joystream/apps docker image was problematic again on mac because of the typegen step in query-node build needing to talk to a running joystream-node, I think again due to difference in network behaviour between linux and mac docker desktop.

I'm tempted to just drop the whole apps.Dockerfile and just depend on the mounted approach in docker-compose, and just have one simple colossus.Dockerfile we will soon have on master branch from here: https://github.com/Joystream/joystream/pull/2486/files

mnaamani:
Also a last point the github actions cache name used for the `run-integration-tests` workflow should probably be different than the one used for the `joystream-node-docker` flow because the they are built with different runtimes but the code hash is being calculated without taking that into account. I think we discussed this issue at one point already.

Lezek123:
> Only issued I had with integration tests is the proposals flow seems to always fail, this seems to be happening in CI as well so I'm not sure if they are expected to be working yet?

They are expected to work, but in order to test them locally, the node needs to be built with:
```
            docker build .\
              --file joystream-node.Dockerfile\
              --tag joystream/node\
              --build-arg TEST_NODE=true
```
(in order to include custom proposal parameters)

The check is failing because of `RangeError: Maximum call stack size exceeded` which wasn't happening during my local tests, I'll try to investigate that.

> There are a couple files that don't seem to belong anymore in the root

I'll remove those

> The root npm script:
> "build": "./build-npm-packages.sh && ./build-docker-images.sh",
> Order should probably be reversed if query-node build depends on joystream-node starting up from the build-docker- images.sh script. But I do have an issue with the build actually resulting in a docker container running.

The container can be removed after `hydra-typegen` fetches the metadata. Alternatively we could commit the chain metadata as `json` into the repository and keep it up to date (not sure if `hydra-typegen` supports this at this moment though) or use the staging endpoint `wss://olympia-dev.joystream.app/rpc` the way it's already done on `master` (but since olympia is still actively developed, that could cause some overhead w.r.t. restarting the staging network)

> Its probably better to remove this script and have clear documentation in README about build steps requirements.

Which one exactly? Just the query node build step from `build-npm-packages.sh`?

> I'm tempted to just drop the whole apps.Dockerfile and just depend on the mounted approach in docker-compose, and just have one simple colossus.Dockerfile we will soon have on master branch from here: https://github.com/Joystream/joystream/pull/2486/files

Sounds good to me

> Also a last point the github actions cache name used for the run-integration-tests workflow should probably be different than the one used for the joystream-node-docker flow because the they are built with different runtimes but the code hash is being calculated without taking that into account. I think we discussed this issue at one point already.

This is taken into account during hash calculation (see: https://github.com/Joystream/joystream/pull/2436/files#diff-4deed07d4de9616206bcbe54d5d120f93c8c6fd95f0baddbc57642854b83f4e8).
During `run-integration-tests` the runtime hash is `db820eca8311117ac0d08afe093575ac8d436b4f` (https://github.com/Joystream/joystream/pull/2502/checks?check_run_id=2930946290), which means `TEST_NODE=true` env was correctly applied:
```
$ ./scripts/runtime-code-shasum.sh 
7a53afc20797ea01630d8bf578d6550e0b0776cf
$ TEST_NODE=true ./scripts/runtime-code-shasum.sh 
db820eca8311117ac0d08afe093575ac8d436b4f
```

Lezek123:
The stack overflow issue during `query-node` build seems to have been specific to the `typescript` version that we were using (`^3.9.7`).

I tried different stack sizes and `tsc` versions locally and it seems like while `^3.9.7` is failing with `stack-size` as high as `500`, both earlier versions (ie. `3.8.*`) and the latest version of `tsc` (`4.3.4`) are compiling `query-node` succesfully even with `stack-size` as low as `200`

The issue could be therefore solved with either the downgrade to `typescript`/`tsc` v `3.8.*` or upgrade to `>4`, but in the second case we'd need to at least also upgrade `eslint`-related packages, so `3.8.*` seemed like a safer choice for now.

mnaamani:
> > Its probably better to remove this script and have clear documentation in README about build steps requirements.
> 
> Which one exactly? Just the query node build step from `build-npm-packages.sh`?

Here I was referring to the npm script in root package.json "build". So we don't force docker build unnecessarily and since it doesn't really work for mac. So mac users may need to build native joystream-node and run it then just manually run `build:packages`.

mnaamani:
> This is taken into account during hash calculation (see: https://github.com/Joystream/joystream/pull/2436/files#diff-4deed07d4de9616206bcbe54d5d120f93c8c6fd95f0baddbc57642854b83f4e8).
> During `run-integration-tests` the runtime hash is `db820eca8311117ac0d08afe093575ac8d436b4f` (https://github.com/Joystream/joystream/pull/2502/checks?check_run_id=2930946290), which means `TEST_NODE=true` env was correctly applied:

Ah I didn't spot setting the env var:

```
      - id: compute_shasum
        name: Compute runtime code shasum
        env:
          TEST_NODE: true
        run: |
```
:+1:

mnaamani:
Rebuilt with TEST_NODE=true and got the query-node tests passing now

mnaamani:
> The container can be removed after `hydra-typegen` fetches the metadata. Alternatively we could commit the chain metadata as `json` into the repository and keep it up to date (not sure if `hydra-typegen` supports this at this moment though) or use the staging endpoint `wss://olympia-dev.joystream.app/rpc` the way it's already done on `master` (but since olympia is still actively developed, that could cause some overhead w.r.t. restarting the staging network)

On master I think we committed generated code and typegen as well but I think that has been problematic. Lets stick with the current approach on olympia for now.

mnaamani (approved):

Issue #2492, opened 2021-06-21
Comments to videos


dobertRowneySr:
# Proposal
### T.O.C.

1.  [Goal](#org9251648)
2.  [Post structure](#orgb3b03a1)
    1.  [Post](#orgbc6c8d2)
    2.  [Replies / Comments](#org681f854)
3.  [Storage Elements](#org6d95e2f)
    1.  [Posts](#orgce6142b)
    2.  [Replies](#org00bfbc6)
4.  [Extrinsics](#orgfa3c3ab)
        1.  [Create Post](#org2899cba)
        2.  [Create Reply](#org110c126)
        3.  [Edit Post](#org02c809a)
        4.  [Edit Reply](#orgf52b304)
        5.  [Delete Post](#org0a108f4)
        6.  [Delete Reply](#org6a11afc)
        7.  [React to Post](#org1a6d6b2)
5.  [Questions & precisations](#org6c2952c)

**Target Branch**: `new_channel_features`

**Module**: `content`

<a id="org9251648"></a>

# Goal

Enrich each video with features for express a rating and comments, something similar to a
Youtube/Vimeo page.
`Basic idea` A video should start a <del>thread</del> post (since in a thread all the posted 
elements tends to be threated as simple posts/replies). Instead what we want
is a post that starts with a video and thereafter there are only replies (to 
the post or to other replies).
 Effectively it should be something along 
the lines of:

<a id="orgb3b03a1"></a>

# Post structure

<span class="underline">Basic idea</span>: A video that lives in a channel should start a  post. Once the post is set up only <span class="underline">text</span> replies are allowed by post owner and other users
We also want the video to be in the channel owned by the post owner

<a id="orgbc6c8d2"></a>

## Post

We want the video to be in the channel owner by the post owner

      pub struct Post<ChannelOwner,
    		  Hash,
    		  Balance,
    		  BlockNumber,
    		  ReplyId, 
    		  VideoCategoryId
    		  > {
          /// Id of thread to which this post corresponds.
    n
          /// Hash of current text
          pub title_hash: Hash,
    
          /// Author of post.
          pub author: ChannelOwner,
    
          /// Cleanup pay off
          pub cleanup_pay_off: Balance,
    
          /// When it was created or last edited
          pub last_edited: BlockNumber,
    
          /// Overall replies counter
          pub replies_count: ReplyId,
    
          /// video associated to the post (instead of the body hash as in the blog module)
          pub video: VideoId,
    
          /// category for this video
[...]

Issue #1990, opened 2021-01-07
Reference Explanation: Metaprotocols


bedeho:
# Background
Multiple upcoming technical ideas will likely involve the idea of metaprotocols in Joystream, and since the concept is not familiar to everyone I will describe it here, for reference in other issues or specs

# Metaprotocol
The fundamental service of a any consensus system is to maintain and build a transaction ordering, whether it is a blockchain or anything else. Once transactions are ordered, one can build consensus about any higher level abstraction, such as a token, exchange, registry or anything else. In all blockchains, the transactions carry some minimal amount of semantics that are also validated by the validator enforced consensus rules, for example rules for creation, transfer, locking and destruction of the native asset (BTC, ETH, .etc), as this is needed to generate the game theoretic incentives for validators/miners to participate at all. Beyond this, some systems have very rich validator in state machines, like the EVM, while others do nothing else, like Bitcoin (more or less). In both systems however, one can build arbitrarily complex consensus state machines on top by simply overloading raw data fields in the native transactions, such as `calldata` in the EVM, or `OP_RETURN` in Bitcoin. There are many examples of this, such as

* **Counterparty:** a Defi protocol on top of Bitcoin which actually implemented the [EVM on top of Bitcoin before Ethereum was aout](https://counterparty.io/news/counterpartys-evm-port-moves-forward/).
* **Omni:** an asset protocol on top of Bitcoin which was the foundation of the Tether system for a long time.
* **Blockstack:** a namespace protocol on top of Bitcoin.
* **Colored Coin:** another very old asset protocol on top of Bitcoin.

here is a more complete list: https://www.reddit.com/r/Bitcoin/comments/3q0ine/meta_protocols_list/

These systems all work by having a separate group of indexing nodes that can identify the embedded transactions in the blockchain, and understand their semantics, so as to allow these nodes to build and maintain a separate state machine.

These embedded consensus systems are often referred to as _metaprotocols_.

The computational load of the metaprotocol computation load on those nodes can even be charged using the normal transactional fee system of the blockchain, by having the metaprotocol require extra tip to be added above and beyond the base fee of the on-chain transaction itself in order to process the transaction and incurr the computational cost. Those nodes would however not receive those fees themselves, but it would deter denial of service attacks.

# Limitations
The main limitation of metaprotocols is that this state cannot impact the ownership of the built in state machine, and thus most importantly, they cannot impact the ownership of the native asset. Here are examples of parts of the current Joystream system which to which this could clearly apply:

[...]

PR #2510, opened 2021-07-01
Video comments

dobertRowneySr:
Specs are from #2492.


runtime-modules/content/src/lib.rs
```diff
@@ -456,6 +467,84 @@ pub struct Person<MemberId> {
...
+    pub cleanup_pay_off: Balance,
+
+    /// When it was created or last edited
+    pub last_edited: BlockNumber,
+
+    /// Overall replies counter
+    pub replies_count: ReplyId,
+
+    /// video associated to the post (instead of the body hash as in the blog module)
+    pub video: VideoId,
```
bedeho: I don't think we need both a `Post` and a `Reply` representation, correct me if I am wrong.

All we need is one, say `Post`, where its `parent` field is a sum type over a video, e.g. like `ParentId`, if its a post directly to a video itself, or to a prior post on the same video.

runtime-modules/content/src/lib.rs
```diff
@@ -456,6 +467,84 @@ pub struct Person<MemberId> {
...
+    parent_id: ParentId<ReplyId, PostId>,
+
+    /// Pay off by deleting post
+    cleanup_pay_off: Balance,
+
+    /// Last time reply was edited
+    last_edited: BlockNumber,
+
+    /// actual text (hash) of the reply
+    text: Hash,
```
bedeho: Drop all hash commitments, we are not using this any longer.

runtime-modules/content/src/lib.rs
```diff
@@ -456,6 +467,84 @@ pub struct Person<MemberId> {
...
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct Post_<MemberId, CuratorGroupId, DAOId, Balance, BlockNumber, ReplyId, VideoId> {
+    /// Author of post.
+    pub author: ChannelOwner<MemberId, CuratorGroupId, DAOId>,
+
+    /// Cleanup pay off
+    pub cleanup_pay_off: Balance,
+
+    /// When it was created or last edited
+    pub last_edited: BlockNumber,
```
bedeho: Check https://github.com/Joystream/joystream/issues/2545, we no longer use this timeout mechanism.

bedeho (changes requested):
Looks like a great start, here are some changes needed

- allow channel owner to turn comment section off or on, and while its off, no comments or reactions on comments are possible. during video creation, the creator can choose to have comment on or off to begin with.
- introduce reactions to the video itself, as distinct from the posts/replies.
- follow this updated scheme that is being introduced in the forum 
https://github.com/Joystream/joystream/issues/2545
- video deletion should include deletion of all posts/replies. This should work in the same way as thread deletion, using witnesses to size of object tree, and unrolling over multiple blocks. Again, we only allow one simultaneous unrolling. In order to traverse the entire tree of posts/replies for a video, an iterable double map with key space `VideoIdxPostId`may be needed to organise the posts/replies.
- Both the channel owner and the content curators can act as moderators. The channel owner can also add a finite number (which must have some hardcoded upper bound) of arbitrary members to the channel which can moderate both video comments and the channel subreddit.
- Can you calculate what the total state footprint, in bytes, of all comments will be as a functino of the number of videos and the average number of comments per video? We need this as a final sanity check on the cost of this feature on the system. If its too expensive, we will need to possible select an off-chain version of this using a metaprotocol version https://github.com/Joystream/joystream/issues/1990. The forum is likely fine as it is, as it won't grow nearly as much as the video comments possibly will.

dobertRowneySr:
The post data structure now looks like this.

``` rust
pub struct Post_<
    MemberId: Default + Clone + Copy,
    CuratorGroupId: Default + Clone + Copy,
    CuratorId: Default + Clone + Copy,
    Balance,
    PostId,
    VideoId,
> {
    /// Author of post.
    pub author: ContentActor<CuratorGroupId, CuratorId, MemberId>,

    /// Cleanup pay off
    pub bloat_bond: Balance,

    /// Overall replies counter
    pub replies_count: PostId,

    /// video associated to the post (instead of the body hash as in the blog module)
    pub post_type: PostType,

    /// parent comment if any
    pub parent_id: Option<PostId>,

    /// video reference
    pub video_reference: VideoId,
}

pub enum PostType {
    VideoPost,
    Comment,
}

```
The `video_reference` field is needed in order to have a quick access to video id the comment is referring to, In order to be inserted into the double map `VideoIdxPostId`. 
The text field works as a description for a `PostType::VideoPost` and a text comment for a `PostType::Comment`. For the moment I allowed only comments from `ContentActor::MemberId(_)`, I don't know If comments by curators should be allowed. Any suggestions @bedeho ?

dobertRowneySr:
## State Footprint computation & estimation
Running `std::mem::size_of` over post in a unit testing scenario (where most of ..Id types are u64 or u32) gives size = 72 bytes.
O(`sizeof(Post_)` x `Averege comments per video` x `Number Of Videos`) = O(`72 x avg_comment_per_video x number_of_video`)

### Estimation using Youtube as a reference
- Comments engagement (= comments/views) for a successful YT channel is at least .5% (source: https://tubularlabs.com/blog/3-metrics-youtube-success/)
- A youtube video (as of now) has on average 8,332 views (source: Google search)
- This means that on average you get 8,332 x .005 = 41.66 < 42 comments per video
- Multiple estimates suggest that there are from 135 to 145 billion videos on youtube (https://www.productmanagementexercises.com/1637/estimate-the-total-number-of-videos-on-youtube)
=> about `72 x 42 x 140,000,000,000 = 4.2336e+14` bytes (as of 2021)

@bedeho does this answer to:
>Can you calculate what the total state footprint, in bytes, of all comments will be as a function of the number of videos and the average number of comments per video?

dobertRowneySr:
As for now I left `cleanup_cost = 1`.
I have the following problem 
- posts are stored in a `double_map`: `VideoId x PostId`
- A `remove_prefix` invocation in a storage double map `N x M` **should** (and maybe I am wrong) have complexity O(log(N) + M) ( [source]( https://www.shawntabrizi.com/substrate/substrate-storage-deep-dive/), [source](https://eth.wiki/fundamentals/patricia-tree))
- cleanup cost for `post` should therefore be sized around `log (next_video_id) + log(post.replies_count)`. This quantity is effectively available at deletion time and not at post creation time, hence I cannot ask for the author to deposit `max( size_of::<Post>() x byte_price, cleanup_cost + margin)` upfront.

So here what I would do:
##### during post creation time
- `bloat_bond = size_of::<Post>() x byte_price` 
- bloat_bond is deposited from post author to post treasury account
##### during cleanup time 
- `bloat_bond = max( size_of::<Post>() x byte_price, cleanup_cost + margin)`
and this quantity (capped at `MaxBloatBond`) it is refunded if actor is the post author, burned otherwise

bedeho:
> A remove_prefix invocation in a storage double map N x M should (and maybe I am wrong) have complexity O(log(N) + M) 

- I could not find where you are getting this from those links? can you be more specific?
- My guess would be that each deletion of a post takes `log(N)` steps, and since you are doing this `M` times, it would be  `Mlog(N)`?
- Your general point here is that database access costs are not constant, but depend on the total size of the current database. This is an interesting point, but surely this should already be an issue confronting pricing normal single database reads, not only these sequences of deletions. Given that the benchmarking approach for Substrate appears to just have a fixed price for reads and writes in all contexts, I wonder if they just choose to pretend that some constant is good enough since `log` grows very slowly eventually. But you should def. dig deeper here, perhaps on the Substrate technical chat.

dobertRowneySr:
>I could not find where you are getting this from those links? can you be more specific?

I was thinking about O(M log N) too, but then I asked on the substrate technical chat and they said it was O(log N + M). I wasn't  fully convinced that is why I expressed caution in the () brackets. Maybe O(M log N) is the cost of M access to a single map and they 
use some optimization for a double map...

dobertRowneySr:
closing in favor of newest Olympia work

PR #2514, opened 2021-07-05
Storage node v2

shamil-gadelshin:
### Implemented features
- [OAS based API](https://swagger.io/specification/)
- API endpoints:
   - file upload
   - get auth token
   - files serving 
   - HEAD endpoint for files
 - [oclif based CLI](https://blog.heroku.com/open-cli-framework)
 - CLI commands:
   - create-bucket
   - accept-bucket invitation
   - update-bag
   - accept_pending_object
   - set alice as WG leader (dev only)
   - upload(dev only)
   - server (starts the WEB api)
   - multihash (dev only - creates a multihash for the file)
   - verify-bag-id (dev only - verifies bagId string)
   - set-bucket-limits
   - set-metadata
   - update-bucket-status
   - cancel-invite
   - delete-bucket
   - remove-operator
   - set-uploading-block
   - update-bag-limit
   - update-blackllist
   - update-data-fee
   - update-voucher-limit
   - update-dynamic-bag-policy
- joystream type library updated
- auth schema:
   - token request is verified by client signature and memberID
   - token is signed using server signature
   - token has expiration time
   - token has nonce that guarantees single usage (having in memory cache for valid tokens)
- runtime integration with the storage pallet 
- blake3 hashing support
- IPFS multihash support
- Winston logger 
- transaction nonce in-process protection (await-lock)

### Testing
- `dev:init` CLI command initializes Alice as storage WG leader in the development chain
- `./scripts/init-dev-bucket.sh` script creates storage bucket for future uploading
- `./scripts/create-auth-request-signature.ts` script creates JSON for the auth token API
- `./scripts/run-all-commands.sh` runs all leader and operator commands (no dev commands except `dev:init`)
- there is an experimental mocha test for testing incorrect command arguments

### TODO
- no p2p syncing
- resumable upload
- integration testing
- cluster support
- add file hash check on authentication
- extrinsic timeouts
- etc......


storage-node-v2/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
Lezek123: I think we should rely on the monorepo root `.editorconfig`

storage-node-v2/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
...
+  parserOptions: {
+    project: './tsconfig.json'
+  },
+  extends: ['@joystream/eslint-config'],
+  rules: {
+    'no-console': 'warn', // use dedicated logger
+    'no-unused-vars': 'off', // Required by the typescript rule below
+    '@typescript-eslint/no-unused-vars': ['error'],
+    '@typescript-eslint/no-floating-promises': 'error',
+    'prettier/prettier': 'off', // prettier-eslint conflicts inherited from @joystream/eslint-config
```
Lezek123: I'm not sure I understand the issue with `prettier/prettier` here

storage-node-v2/scripts/create-auth-request-signature.ts
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env ts-node
+
+import { createApi } from '../src/services/runtime/api'
+import { getAlicePair } from '../src/services/runtime/accounts'
+import { UploadTokenRequestBody, signTokenBody, UploadTokenRequest } from '../src/services/helpers/auth'
+import { exit } from 'process'
+
+//Wasm init
+createApi('ws://localhost:9944').then(() => {
```
Lezek123:
```suggestion
import { cryptoWaitReady } from '@polkadot/util-crypto'
import { getAlicePair } from '../src/services/runtime/accounts'
import { UploadTokenRequestBody, signTokenBody, UploadTokenRequest } from '../src/services/helpers/auth'
import { exit } from 'process'

//Wasm init
cryptoWaitReady().then(() => {
```

storage-node-v2/src/command-base/ApiCommandBase.ts
```diff
@@ -0,0 +1,146 @@
...
+   * Read the chain name from the runtime and throws an error if it's
+   * not 'Development' chain.
+   */
+  async ensureDevelopmentChain(): Promise<void> {
+    const api = await this.getApi()
+
+    const developmentChainName = 'Development'
+    const runningChainName = await api.rpc.system.chain()
+
+    if (runningChainName.toString() !== developmentChainName) {
```
Lezek123:
```suggestion
    const chainType = await api.rpc.system.chainType()

    if (!chainType.isDevelopment && !chainType.isLocal) {
```

storage-node-v2/scripts/create-several-buckets.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+yarn storage-node dev:init
+
+curl http://localhost:3333/test &
+curl http://localhost:3333/test &
+curl http://localhost:3333/test &
+curl http://localhost:3333/test &
+curl http://localhost:3333/test &
```
Lezek123: The endpoint no loger exists(?) so I think this file can be safely removed

storage-node-v2/src/api-spec/openapi.yaml
```diff
@@ -0,0 +1,207 @@
+openapi: 3.0.3
+info:
+  title: Storage node API
+  description: Storage node API
+  contact:
+    email: info@joystream.org
+  license:
+    name: MIT
+    url: https://opensource.org/licenses/MIT
```
Lezek123: The previous node was `GPL-3.0-only`, shouldn't we use the same license?

storage-node-v2/src/services/webApi/app.ts
```diff
@@ -0,0 +1,168 @@
...
+      err: HttpError,
+      req: express.Request,
+      res: express.Response,
+      next: express.NextFunction
+    ) => {
+      res.status(err.status).json({
+        type: 'request_validation',
+        message: err.message,
+        errors: err.errors,
+      })
```
Lezek123, line 74: As a generic error handler it can potentially catch other exceptions as well (not related to request validation).
Currently it probably wouldn't be the case, because all errors are handled by `try-catch` inside the route, but it seems like something worth considering.

Also I noticed that errors like failed upload (`410`) aren't really logged anywhere (only the request is logged with `info` level). It would make sense to call `next()` during such errors and then use an error handler like `expressWinston.errorLogger` to log them with `warn` or `error` level.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -0,0 +1,405 @@
...
+    res.setHeader('Content-Disposition', 'inline')
+    res.setHeader('Content-Type', fileInfo.mimeType)
+    res.setHeader('Content-Length', fileStats.size)
+
+    res.status(200).send()
+  } catch (err) {
+    if (isNofileError(err)) {
+      res.status(404).send()
+    } else {
+      res.status(410).send()
```
Lezek123: Since we don't know what exactly is the cause of the error, shouldn't `500` be used instead of  `410` (which indicates client-side error and also gets cached by default)?

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -0,0 +1,405 @@
...
+      uploadRequest.storageBucketId,
+      [uploadRequest.dataObjectId]
+    )
+    res.status(201).json({
+      id: hash,
+    })
+  } catch (err) {
+    await cleanupFileOnError(cleanupFileName, err.toString())
+
+    res.status(410).json({
```
Lezek123: Same as above, I think `500` would be more suitable here

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -0,0 +1,405 @@
...
+      validUntil: getTokenExpirationTime(),
+      ...tokenRequest.data,
+    }
+    const signedToken = createUploadToken(tokenBody, account)
+
+    res.status(201).json({
+      token: signedToken,
+    })
+  } catch (err) {
+    res.status(410).json({
```
Lezek123: Same as above, I think `500` would be more suitable here (any anywhere where we don't know exactly what caused the error)

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -0,0 +1,405 @@
...
+ * This is a helper function. It parses the request object for a variable and
+ * throws an error on failier.
+ */
+function getFileObject(req: express.Request): Express.Multer.File {
+  if (req.file) {
+    return req.file
+  }
+
+  const files = req.files as Express.Multer.File[]
+  if (files && files.length > 0) {
```
Lezek123, line 176: Is it possible that `files.length` will ever be `> 1`? In that case additional cleanup may be required

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -0,0 +1,405 @@
...
+ * @remarks
+ * This is a helper function. It parses the response object for a variable and
+ * throws an error on failier.
+ */
+function getWorkerId(res: express.Response): number {
+  if (res.locals.workerId || res.locals.workerId === 0) {
+    return res.locals.workerId
+  }
+
+  throw new Error('No Joystream worker ID loaded.')
```
Lezek123: This is just one example of an error that doesn't fit well into `410` category

storage-node-v2/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "importHelpers": true,
+    "module": "commonjs",
+    "outDir": "lib",
+    "rootDir": "src",
+    "strict": true,
+    "target": "es2017",
+    "skipLibCheck": true,
+    "baseUrl": ".",
+    "paths": {
+      "@polkadot/types/augment": ["../types/augment-codec/augment-types.ts"],
```
Lezek123, line 13: Is there a specific reason not to use `@polkadot/api/augment` too? I think it could be very benefitial in terms of typesafety.

storage-node-v2/src/commands/dev/multihash.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  async run(): Promise<void> {
+    const { flags } = this.parse(DevMultihash)
+
+    logger.info(`Hashing ${flags.file} ....`)
+
+    const multi = await hashFile(flags.file)
+
+    logger.info(`Hash: ${multi}`)
```
Lezek123, line 33: This is one of the commands that would be useful for bash scripts, so it would be nice if it would output just the hash to `stdout` and all other logs to `stderr` (perhaps a similar practice can be followed for other commands too).

Currently I had to use:
```
HASH=$(yarn storage-node dev:multihash -f "$FILE_PATH" | tr '\n' '\r' | sed 's|.*Hash: \([A-Za-z0-9]\+\).*$|\1|g')
```
To extract the result in my script

storage-node-v2/src/commands/dev/upload.ts
```diff
@@ -0,0 +1,52 @@
...
+    const objectSize = flags.size ?? 0
+    const objectCid = flags.cid
+
+    logger.info('Uploading data objects...')
+
+    const api = await this.getApi()
+
+    // Must be number.
+    const dataFee =
+      (await api.query.storage.dataObjectPerMegabyteFee()) as unknown as BN
```
Lezek123: This is one of the cases where adding `@polkadot/api/augment` may be helpful (it would automatically decorate all the methods with correct types)

storage-node-v2/src/commands/leader/cancel-invite.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  async run(): Promise<void> {
+    const { flags } = this.parse(LeaderCancelInvite)
+
+    const storageBucketId = flags.bucketId
+
+    logger.info('Canceling storage bucket operator invite...')
+    if (flags.dev) {
+      await this.ensureDevelopmentChain()
+    }
```
Lezek123, line 34: Perhaps this should be part of the base command's `init` function to avoid the need for adding it to each command separately.

storage-node-v2/src/commands/leader/set-uploading-block.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * CLI command:
+ * Manages global uploading block in the runtime.
+ *
+ * @remarks
+ * Storage working group leader command. Requires storage WG leader priviliges.
+ * Shell command: "leader:set-uploading-block"
+ */
+export default class LeaderSetUploadingBlock extends ApiCommandBase {
```
Lezek123: I found the command name a little confusing (it seems like it's referring to a _block_ as in _a block in a blockchain_), perhaps `GlobalUploadingStatus` would be more descriptive?

storage-node-v2/src/commands/leader/set-uploading-block.ts
```diff
@@ -0,0 +1,47 @@
...
+  static description = `Set global uploading block. Requires storage working group leader permissions.`
+
+  static flags = {
+    enable: flags.boolean({
+      char: 'e',
+      description: 'Enables global uploading block (default).',
+    }),
+    disable: flags.boolean({
+      char: 'd',
+      description: 'Disables global uploading block.',
```
Lezek123: Should be `exclusive: ['enable']` (and vice-versa). Or maybe just one flag `enabled` which is `flags.enum` with `['y', 'n']` options?

storage-node-v2/src/commands/leader/update-bag-limit.ts
```diff
@@ -0,0 +1,43 @@
...
+  async run(): Promise<void> {
+    const { flags } = this.parse(LeaderUpdateBagLimit)
+
+    logger.info('Update "Storage buckets per bag" number limit....')
+    if (flags.dev) {
+      await this.ensureDevelopmentChain()
+    }
+
+    const account = this.getAccount(flags)
+    const limit = flags.limit ?? 0
```
Lezek123: If the flag is `required` or has a `default` specified, the value will never be `undefined`/`null` (which is already inferred by TS compiler), therefore there seems to be no need to use `??` for most (all?) of the commands.

storage-node-v2/src/commands/leader/update-bag.ts
```diff
@@ -0,0 +1,73 @@
...
+  static flags = {
+    bucket: flags.integer({
+      char: 'b',
+      required: true,
+      description: 'Storage bucket ID',
+    }),
+    remove: flags.boolean({
+      char: 'r',
+      description: 'Remove a bucket from the bag',
+    }),
```
Lezek123: You could use `add` and `remove` flags instead with `multiple: true` to allow specifying multiple buckets to add/remove with just one runtime call (see [Distributor node's `update:bag` command](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/src/commands/leader/update-bag.ts))

storage-node-v2/src/commands/leader/update-blacklist.ts
```diff
@@ -0,0 +1,47 @@
...
+  static flags = {
+    cid: flags.string({
+      char: 'c',
+      required: true,
+      description: 'Content ID',
+    }),
+    remove: flags.boolean({
+      char: 'r',
+      description: 'Remove a content ID from the blaclist',
+    }),
```
Lezek123: Here you could also use `add` and `remove` flags with `multiple: true`

storage-node-v2/src/commands/leader/update-bucket-status.ts
```diff
@@ -0,0 +1,58 @@
...
+      description: 'Storage bucket ID',
+    }),
+    enable: flags.boolean({
+      char: 'e',
+      description: 'Enables accepting new bags (default).',
+    }),
+    disable: flags.boolean({
+      char: 'd',
+      description: 'Disables accepting new bags.',
+    }),
```
Lezek123, line 30: I think either using `exclusive` option or one flag with `y / n` values would also apply here.

storage-node-v2/src/commands/leader/update-dynamic-bag-policy.ts
```diff
@@ -0,0 +1,65 @@
...
+      description: 'New storage buckets number',
+    }),
+    member: flags.boolean({
+      char: 'e',
+      description: 'Member dynamic bag type (default)',
+    }),
+    channel: flags.boolean({
+      char: 'c',
+      description: 'Channel dynamic bag type',
+    }),
```
Lezek123: Perhaps use `enum` flag with `Member` and `Channel` as options? (this would also save the need for additional parsing inside `run()`)

storage-node-v2/src/commands/operator/set-metadata.ts
```diff
@@ -0,0 +1,61 @@
...
+    }),
+    operatorId: flags.integer({
+      char: 'w',
+      required: true,
+      description: 'Storage bucket operator ID (storage group worker ID)',
+    }),
+    metadata: flags.string({
+      char: 'm',
+      description: 'Storage bucket operator metadata',
+    }),
```
Lezek123, line 32: Just pointing out that this will probably be a [json input](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/src/commands/operator/set-metadata.ts) in the future, as the metadata will need to be encoded using the protobuf message class

storage-node-v2/src/services/helpers/fileInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export async function getFileInfo(fullPath: string): Promise<FileInfo> {
+  // Default file info if nothing could be detected.
+  const DEFAULT_FILE_INFO = {
+    mimeType: 'application/octet-stream',
+    ext: 'bin',
+  }
+
+  const buffer = readChunk.sync(fullPath, 0, MINIMUM_FILE_CHUNK)
+  const fileType = await FileType.fromBuffer(buffer)
```
Lezek123, line 40: I'm not sure how well using just a small chunk of data will work for different file types, it should be safer to rely on `FileType.fromFile` instead. Perhaps also keep some state (even if it's just in memory) about already detected mime types, to avoid re-running the detection script on each request.

Because the distributor node does proxy some requests to storage-node I think it's unavoidable that both need to be aware of the mimeType (and ideally would use the same mechanism to detect it)

storage-node-v2/src/services/runtime/extrinsics.ts
```diff
@@ -0,0 +1,584 @@
...
+ * @param api - runtime API promise
+ * @param account - KeyringPair instance
+ * @param invitedWorker - defines whether the storage provider should be
+ * invited on the bucket creation (optional)
+ * @param allowedNewBags - bucket allows new bag assignments
+ * @param sizeLimit - size limit in bytes for the new bucket (default 0)
+ * @param objectsLimit - object number limit for the new bucket (default 0)
+ * @returns promise with a success flag.
+ */
+export async function createStorageBucket(
```
Lezek123, line 23: With `@polkadot/api/augment` in `tsconfig` `paths` I think there won't be a need for most of those functions (which purpose as I understand is mostly just to provide some typesafety for the extrinsic calls). Api augmentation would also make it possible to retrieve events from extrinsic results in a typesafe way. You can check [this](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/tsconfig.json), [this](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/src/services/networking/runtime/api.ts) and [this](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/src/command-base/api.ts) for reference if you find it useful

storage-node-v2/src/api-spec/openapi.yaml
```diff
@@ -0,0 +1,207 @@
...
+          required: true
+          in: path
+          description: Content ID
+          schema:
+            type: string
+      responses:
+        200:
+          description: Ok
+          content:
+            application/octet-stream:
```
Lezek123, line 51: Since this api supports many different media types as responses (any type `FileType` will detect), shouldn't a `*/*` placeholder be used instead?
Or perhaps `image/*`, `video/*`, `audio/*`, `application/octet-stream` (since those are all currently supported ones)

I noticed that was changed in https://github.com/Joystream/joystream/pull/2514/commits/17a3d4d8fdbc0e49a32e652ad1f4505541f65df9

storage-node-v2/test/commands/leader/create-bucket.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      console.log("Initialized.")
+      done()
+    });
+  });
+
+  const cmd =  new CommandBuilder().cmd('leader:create-bucket').args('').final()
+  describe(cmd, () => {
+    it("fails without keyfile", function(done) {
+      child.exec(cmd, {timeout: 5000}, (error: any, stdout: string, stderr: string) => {
+          assert(stderr.includes('Key'), 'Should contain an error.')
```
Lezek123: Probably better to rely on exit codes if we decide to use `sterr` for logging.

storage-node-v2/test/commands/leader/create-bucket.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+          done()
+      });
+    });
+  })  
+  
+  const cmd2 =  new CommandBuilder().cmd('leader:create-bucket').args('--dev').final()
+  describe(cmd2, () => {
+
+    it("fails without keyfile", function(done) {
```
Lezek123: Test case description is wrong

storage-node-v2/scripts/run-all-commands.sh
```diff
@@ -0,0 +1,38 @@
...
+yarn storage-node dev:init # leader workerId = 0
+
+# Update limits and constants
+yarn storage-node leader:update-bag-limit -l 7 --dev
+yarn storage-node leader:update-voucher-limits -o 100 -s 10000000 --dev
+yarn storage-node leader:update-data-fee -f 10000 --dev
+yarn storage-node leader:update-dynamic-bag-policy -n 10 -m
+
+# Create and configure a bucket.
+yarn storage-node leader:create-bucket -i=0 --dev # bucketId = 0
```
Lezek123: If you decide to use `stderr` for logging and `stdout` for output, this script could then make no assumptions about ids, making it possible to run multiple times on the same chain (see: [this](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/scripts/test-commands.sh))

storage-node-v2/scripts/run-all-commands.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
Lezek123, line 1: The script runs all commands _except_ `update-blacklist`, which gives me:
```
Api Error: Error: Cannot send the extrinsic: 1002: Verification Error: Execution: Could not convert parameter `tx` between node and runtime: Error decoding field Call :: Storage.0: RuntimeApi, Execution: Could not convert parameter `tx` between node and runtime: Error decoding field Call :: Storage.0
```

When I try to run it manually

types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/types",
-  "version": "0.15.0",
+  "version": "0.17.0",
   "description": "Types for Joystream Substrate Runtime - Antioch release",
```
Lezek123:
```suggestion
  "version": "0.17.0",
  "description": "Types for Joystream Substrate Runtime - Giza release",
```

types/src/storage.ts
```diff
@@ -0,0 +1,262 @@
...
+export class DataObjectIdSet extends JoyBTreeSet(DataObjectId) {}
+export class DistributionBucketId extends u64 {}
+export class DistributionBucketFamilyId extends u64 {}
+export class StorageBucketIdSet extends JoyBTreeSet(StorageBucketId) {}
+export class DistributionBucketSet extends JoyBTreeSet(DistributionBucketId) {}
+
+export type DynamicBagDeletionPrizeTypeDef = {
+  account_id: AccountId
+  prize: BalanceOf
+}
```
Lezek123: Types like this are usually interfaces names `ITypeName`, so I think we should stick to this pattern:
```suggestion
interface IDynamicBagDeletionPrize {
  account_id: AccountId
  prize: BalanceOf
}
```

`TypeNameDef` is usually used for enum definition consts

types/src/storage.ts
```diff
@@ -0,0 +1,262 @@
...
+export const StaticBagIdDef = {
+  Council: Null,
+  WorkingGroup: WorkingGroup,
+} as const
+export type StaticBagIdKey = keyof typeof StaticBagIdDef
+export class StaticBagId extends JoyEnum(StaticBagIdDef) {}
+
+export class Static extends StaticBagId {}
+
+export class ChannelId extends u64 {}
```
Lezek123, line 97: This one should be imported from `content-directory`/`common` types once the new content directory is merged.

types/src/storage.ts
```diff
@@ -0,0 +1,262 @@
...
+  BTreeSet,
+  BTreeMap,
+  Option,
+} from '@polkadot/types'
+import { RegistryTypes } from '@polkadot/types/types'
+import { JoyBTreeSet, JoyEnum, JoyStructDecorated, WorkingGroup } from './common'
+import { MemberId } from './members'
+import { WorkerId } from './working-group'
+
+export class BalanceOf extends u128 {}
```
Lezek123: I think this one belongs to `common` types

types/src/storage.ts
```diff
@@ -0,0 +1,262 @@
...
+  StorageWorker: WorkerId,
+} as const
+export type StorageBucketOperatorStatusKey = keyof typeof StorageBucketOperatorStatusDef
+export class StorageBucketOperatorStatus extends JoyEnum(StorageBucketOperatorStatusDef) {}
+
+export type StorageBucketType = {
+  operator_status: StorageBucketOperatorStatus
+  accepting_new_bags: bool
+  voucher: Voucher
+  metadata: Text
```
Lezek123: I would suggest using `Bytes` which is a standard for metadata (it's also a `Vec<u8>`)

types/src/storage.ts
```diff
@@ -0,0 +1,262 @@
...
+  authenticationKey: Text
+  bagId: BagId
+  objectCreationList: Vec<DataObjectCreationParameters>
+  deletionPrizeSourceAccountId: AccountId
+  expectedDataSizeFee: BalanceOf
+}
+
+export class UploadParameters
+  extends JoyStructDecorated({
+    authenticationKey: Text,
```
Lezek123: Generally I think `Bytes` is a safer choice, because there are some issues when decoding a `Text` with some specific bytes (especially `0x90`)

storage-node-v2/src/services/logger.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @returns  Express-Winston logger handler
+ *
+ */
+export function httpLogger(): Handler {
+  const opts: expressWinston.LoggerOptions = {
+    transports: [new winston.transports.Console()],
+    format: winston.format.combine(
+      winston.format.colorize(),
+      winston.format.json()
```
Lezek123: This produces an output which is not colorized (in the CLI), but instead json values are broken (ie. `"level":"\u001b[32minfo\u001b[39m"`)

Lezek123 (changes requested):

storage-node-v2/.editorconfig
shamil-gadelshin: Ok.

storage-node-v2/scripts/create-several-buckets.sh
shamil-gadelshin: Sure.

storage-node-v2/src/api-spec/openapi.yaml
shamil-gadelshin: Our root repository shows the GPLv3. So, you are correct.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: I'm not sure there is a way to forbid it. Anyway, auto cleanup is a part of the sync process.

storage-node-v2/src/commands/dev/upload.ts
shamil-gadelshin: Makes sense.

storage-node-v2/tsconfig.json
shamil-gadelshin: Ok.

types/src/storage.ts
shamil-gadelshin: Changed to Bytes.

types/src/storage.ts
shamil-gadelshin: Changed to Bytes.

types/src/storage.ts
shamil-gadelshin: Moved there.

types/src/storage.ts
shamil-gadelshin: Made a note.

storage-node-v2/src/commands/leader/cancel-invite.ts
shamil-gadelshin: Most but not every command has the same code snippet.

storage-node-v2/src/commands/leader/set-uploading-block.ts
shamil-gadelshin: Ok, sure.

storage-node-v2/src/commands/leader/set-uploading-block.ts
shamil-gadelshin: Changed to --set (on/off)

storage-node-v2/src/commands/leader/update-bag-limit.ts
shamil-gadelshin: Makes sense. Fixed.

storage-node-v2/src/commands/leader/update-bag.ts
shamil-gadelshin: Good idea, thanks.

storage-node-v2/src/commands/leader/update-blacklist.ts
shamil-gadelshin: Good idea!

storage-node-v2/src/commands/leader/update-bucket-status.ts
shamil-gadelshin: Fixed.

storage-node-v2/src/commands/leader/update-dynamic-bag-policy.ts
shamil-gadelshin: Makes sense.

storage-node-v2/scripts/run-all-commands.sh
shamil-gadelshin: Fixed.

storage-node-v2/src/services/runtime/extrinsics.ts
shamil-gadelshin: I added `@polkadot/api/augment` but I'd like to keep these functions for now to hide runtime-dependent code.

storage-node-v2/.eslintrc.js
shamil-gadelshin: Removed for now (there are no conflicts after latest changes)

storage-node-v2/src/commands/operator/set-metadata.ts
shamil-gadelshin: Added to the working tasks queue. Will do it in the next PR.

storage-node-v2/src/services/logger.ts
shamil-gadelshin: Removed incorrect setting.

storage-node-v2/test/commands/leader/create-bucket.test.ts
shamil-gadelshin: Removed test example in favor of the future integration tests.

storage-node-v2/test/commands/leader/create-bucket.test.ts
shamil-gadelshin: Removed test example in favor of the future integration tests.

storage-node-v2/src/services/helpers/fileInfo.ts
shamil-gadelshin: To be changed in the future PRs.

storage-node-v2/scripts/run-all-commands.sh
shamil-gadelshin: Added event parsing for bucket id and introduced a dedicated logger for the standard output.

storage-node-v2/src/commands/dev/multihash.ts
shamil-gadelshin: Implemented your suggestion.

types/src/storage.ts
shamil-gadelshin: Fixed for all entries.

storage-node-v2/src/api-spec/openapi.yaml
shamil-gadelshin: It has no benefit from the library's response checking but we could add it to add some hints on the possible content.

storage-node-v2/src/services/webApi/app.ts
shamil-gadelshin: Added `expressWinston.errorLogger`, changed the handler's behavior.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced 500 in the default error handler.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced 500 in the default error handler.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced 500 in the default error handler.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced 500 in the default error handler.

storage-node-v2/src/commands/dev/upload.ts
```diff
@@ -34,16 +34,15 @@ export default class DevUpload extends ApiCommandBase {
...
 
     logger.info('Uploading data objects...')
 
     const api = await this.getApi()
 
-    // Must be number.
-    const dataFee =
-      (await api.query.storage.dataObjectPerMegabyteFee()) as unknown as BN
+    // Must be a number.
+    const dataFee = (await api.query.storage.dataObjectPerMegabyteFee()) as BN
```
Lezek123: I don't think there's a need to use `as BN`, since `Balance` / `BalanceOf` extends `u128` which extends `BN`

storage-node-v2/src/commands/leader/update-dynamic-bag-policy.ts
```diff
@@ -14,51 +14,45 @@ import { parseDynamicBagType } from '../../services/helpers/bagTypes'
...
     const account = this.getAccount(flags)
-    const newNumber = flags.number ?? 0
+    const newNumber = flags.number
 
-    let dynamicBagTypeString: 'Member' | 'Channel' = 'Member' // Default
-    if (flags.channel) {
-      dynamicBagTypeString = 'Channel'
-    }
+    // Verified by enum argument parser.
+    const dynamicBagTypeString = flags.bagType as 'Member' | 'Channel'
```
Lezek123: A safer approach is to provide the type during `flags.enum` call, ie.:
```
flags.enum<'Member' | 'Channel'>({ ... })
```
or
```
flags.enum<DynamicBagTypeKey>({ ... })
```

It will then be both enforced on `options` attribute and inferred for `flags.bagType` after `this.parse` (so no assertion needed)

storage-node-v2/src/services/runtime/api.ts
```diff
@@ -151,30 +134,43 @@ function formatDispatchError(api: ApiPromise, error: DispatchError): string {
...
-  params: CodecArg[],
-  sudoCall = false
-): Promise<void> {
-  logger.debug(`Sending ${module}.${method} extrinsic...`)
-  let tx = api.tx[module][method](...params)
+  tx: SubmittableExtrinsic<'promise'>,
+  sudoCall = false,
+  eventParser: ((result: ISubmittableResult) => T) | null = null
+): Promise<T | void> {
+  const description = tx.toHuman() as {
```
Lezek123: Why not just access `tx.method.method` and `tx.method.section` directly? (without converting `.toHuman()` and asserting the type)

storage-node-v2/src/services/runtime/extrinsics.ts
```diff
@@ -28,18 +28,21 @@ export async function createStorageBucket(
   allowedNewBags = true,
   sizeLimit = 0,
   objectsLimit = 0
-): Promise<boolean> {
-  return await extrinsicWrapper(() => {
+): Promise<[boolean, number | void]> {
+  let bucketId: number | void = 0
+  const success = await extrinsicWrapper(async () => {
     const invitedWorkerValue = api.createType('Option<WorkerId>', invitedWorker)
```
Lezek123, line 33: You can also directly pass `number | null` to `api.tx.storage.createStorageBucket` as arg and `@polkadot/api` will just use `createType` under the hood in that case

storage-node-v2/src/services/runtime/extrinsics.ts
```diff
@@ -28,18 +28,21 @@ export async function createStorageBucket(
...
-      api,
-      account,
-      'storage',
-      'createStorageBucket',
-      [invitedWorkerValue, allowedNewBags, sizeLimit, objectsLimit]
-    )
+    const tx = api.tx.storage.createStorageBucket(invitedWorkerValue, allowedNewBags, sizeLimit, objectsLimit)
+    bucketId = await sendAndFollowNamedTx(api, account, tx, false, (result) => {
+      const event = getEvent(result, 'storage', 'StorageBucketCreated')
+      const bucketId = event?.data[0] as BN
```
Lezek123: I don't think `as BN` is needed.

storage-node-v2/src/services/runtime/extrinsics.ts
```diff
@@ -80,35 +79,24 @@ export async function acceptStorageBucketInvitation(
...
   account: KeyringPair,
-  bucketId: number,
-  removeBucket: boolean
+  add: number[],
+  remove: number[]
 ): Promise<boolean> {
   return await extrinsicWrapper(() => {
-    let addBuckets: CodecArg
-    let removeBuckets: CodecArg
+    const removeBuckets = api.createType('StorageBucketIdSet', remove)
```
Lezek123, line 93: As pointed out before, `@polkadot/api` will do this under the hood (there are a few other instances of unnecessary `createType` in this file).

What's special about `BTreeSet` though is that the encoding will break if the values are not sorted in ascending order. This was fixed in `@polkadot/api` [`5.2.1`](https://github.com/polkadot-js/api/blob/master/CHANGELOG.md#521-jul-26-2021), which we can upgrade to before the release (but for now the values here can be just sorted with `remove.sort((a, b) => a - b)`)

_Edit: Nvm, in this case it's actually not a problem since `JoyBTreeSet` is used as a workaround._

storage-node-v2/src/services/runtime/hireLead.ts
```diff
@@ -21,100 +16,67 @@ import logger from '../../services/logger'
...
   const SudoKeyPair = getAlicePair()
   const LeadKeyPair = getAlicePair()
 
-  const nullValue = null as unknown as CodecArg
-
   // Create membership if not already created
-  const members = (await api.query.members.memberIdsByControllerAccountId(
-    LeadKeyPair.address
-  )) as Vec<MemberId>
+  const members = (await api.query.members.memberIdsByControllerAccountId(LeadKeyPair.address)) as Vec<MemberId>
```
Lezek123: The `as Vec<MemberId>` assertion can be removed

storage-node-v2/src/services/runtime/hireLead.ts
```diff
@@ -21,100 +16,67 @@ import logger from '../../services/logger'
...
 
   let memberId: MemberId | undefined = members.toArray()[0] as MemberId
```
Lezek123: The `as MemberId` assertion can be removed (this is the case generally throughout the file)

storage-node-v2/src/services/runtime/hireLead.ts
```diff
@@ -21,100 +16,67 @@ import logger from '../../services/logger'
...
   }
 
   logger.info(`Making member id: ${memberId} the content lead.`)
 
-  const newOpeningId =
-    (await api.query.storageWorkingGroup.nextOpeningId()) as OpeningId
-  const newApplicationId =
-    (await api.query.storageWorkingGroup.nextApplicationId()) as ApplicationId
+  const newOpeningId = (await api.query.storageWorkingGroup.nextOpeningId()) as OpeningId
+  const newApplicationId = (await api.query.storageWorkingGroup.nextApplicationId()) as ApplicationId
```
Lezek123: Those can be now fetched easly from `OpeningAdded` / `AppliedOnOpening` events

storage-node-v2/src/services/webApi/app.ts
```diff
@@ -67,25 +59,28 @@ export async function createApp(
...
-      req: express.Request,
-      res: express.Response,
-      next: express.NextFunction
-    ) => {
+  app.use(errorLogger())
+
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  app.use((err: Error, req: express.Request, res: express.Response, next: express.NextFunction) => {
+    // Request validation error handler.
+    if (err instanceof HttpError) {
```
Lezek123, line 69: Nice find!

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -60,10 +53,7 @@ export async function getFile(
 /**
  * A public endpoint: sends file headers by CID.
  */
-export async function getFileHeaders(
-  req: express.Request,
-  res: express.Response
-): Promise<void> {
+export async function getFileHeaders(req: express.Request, res: express.Response): Promise<void> {
```
Lezek123, line 80: Here the server is still responding with `410 Gone` for an unknown error.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -88,10 +78,7 @@ export async function getFileHeaders(
 /**
  * A public endpoint: receives file.
  */
-export async function uploadFile(
-  req: express.Request,
-  res: express.Response
-): Promise<void> {
+export async function uploadFile(req: express.Request, res: express.Response): Promise<void> {
```
Lezek123, line 101: Same here with `410`

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -138,10 +120,7 @@ export async function uploadFile(
 /**
  * A public endpoint: creates auth token for file uploads.
  */
-export async function authTokenForUploading(
-  req: express.Request,
-  res: express.Response
-): Promise<void> {
+export async function authTokenForUploading(req: express.Request, res: express.Response): Promise<void> {
```
Lezek123, line 140: Also a `410` here.

One example of why I think it's not valid is that `getAccount` can potentially throw `No Joystream account loaded.`, which is not a request error, it's an error with node configuration. Same goes for `getApi`. There may also be other errors than will end up in `catch` block, so we're not really sure what was the cause of the `throw`.

`410` response means that _access to the target resource is no longer available at the origin server and that this condition is likely to be permanent._ ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410)). It may end up beeing cached (even forever!) by the client and the request will not be repeated again even after the issue is resolved.

Lezek123 (changes requested):
Great work, I suggested a few small improvements.
The only big issue I have are the `410 Gone` responses, which in my opinion still don't take into account some (or most) of the possible causes of an error. (and are very bad as "generic error" responses, since they may end up getting cached forever)
I would suggest:
- using `400` when 100% sure the request is invalid and it's not an issue with server config or some other unexpected state
- using `500` when not exactly sure what caused the error (this is mostly the case when using big `try-catch` blocks)
- using `410` when, for example, some resource has actually been permanently removed and we can be sure it's never going to be available again (I think that's rarely the case and can't think of any reason to use it currently)

storage-node-v2/src/commands/dev/upload.ts
shamil-gadelshin: Removed 'as BN'

storage-node-v2/src/commands/leader/update-dynamic-bag-policy.ts
shamil-gadelshin: Ok.

storage-node-v2/src/services/runtime/api.ts
shamil-gadelshin: Fixed.

storage-node-v2/src/services/runtime/extrinsics.ts
shamil-gadelshin: Turns out I can't do that:
```
Argument of type 'number | null' is not assignable to parameter of type 'string | object | Uint8Array | Option<WorkerId> | null'.
  Type 'number' is not assignable to type 'string | object | Uint8Array | Option<WorkerId> | null'.

34     const tx = api.tx.storage.createStorageBucket(invitedWorker, allowedNewBags, sizeLimit, objectsLimit)
```

storage-node-v2/src/services/runtime/extrinsics.ts
shamil-gadelshin: Removed it.

storage-node-v2/src/services/runtime/extrinsics.ts
shamil-gadelshin: It works without these types, but `yarn build` has errors:
```
Argument of type 'number[]' is not assignable to parameter of type 'BTreeSet<StorageBucketId>'.
  Type 'number[]' is missing the following properties from type 'BTreeSet<StorageBucketId>': #private, registry, encodedLength, hash, and 13 more.

93     const tx = api.tx.storage.updateStorageBucketsForBag(bagId, add, remove)
```

storage-node-v2/src/services/runtime/hireLead.ts
shamil-gadelshin: Removed.

storage-node-v2/src/services/runtime/hireLead.ts
shamil-gadelshin: Removed all assertions.

storage-node-v2/src/services/runtime/hireLead.ts
shamil-gadelshin: Changed to event data fetching.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced fine-grained error distinction.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced fine-grained error distinction.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Introduced fine-grained error distinction.

storage-node-v2/src/services/runtime/extrinsics.ts
Lezek123: I see. It should work, but there's probably some issue with `@polkadot/typegen` not allowing that

storage-node-v2/src/services/runtime/extrinsics.ts
Lezek123: Ok, my bad, I think `@polkadot/typegen` types just won't allow passing an array for `BTreeSet` (only `Vec`)

storage-node-v2/src/services/runtime/hireLead.ts
```diff
@@ -52,7 +47,16 @@ export async function hireStorageWorkingGroupLead(api: ApiPromise): Promise<void
     'storage opening', // human_readable_text
     'Leader' // opening_type
   )
-  await sendAndFollowSudoNamedTx(api, SudoKeyPair, tx)
+  const newOpeningId = await sendAndFollowSudoNamedTx(api, SudoKeyPair, tx, (result) => {
+    const event = getEvent(result, 'storageWorkingGroup', 'OpeningAdded')
+    const bucketId = event?.data[0]
+
+    return bucketId.toNumber()
```
Lezek123:
```suggestion
    const openingId = event?.data[0]

    return openingId.toNumber()
```

storage-node-v2/src/services/runtime/hireLead.ts
```diff
@@ -64,7 +68,12 @@ export async function hireStorageWorkingGroupLead(api: ApiPromise): Promise<void
     null, // opt appl. stake
     'bootstrap opening' // human_readable_text
   )
-  await sendAndFollowNamedTx(api, LeadKeyPair, tx)
+  const newApplicationId = await sendAndFollowNamedTx(api, LeadKeyPair, tx, false, (result) => {
+    const event = getEvent(result, 'storageWorkingGroup', 'AppliedOnOpening')
+    const bucketId = event?.data[1]
+
+    return bucketId.toNumber()
```
Lezek123:
```suggestion
    const applicationId = event?.data[1]

    return applicationId.toNumber()
```

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -363,3 +352,29 @@ function sendResponseWithError(res: express.Response, err: Error, errorType: str
...
+
+  if (err instanceof WebApiError) {
+    return 400
+  }
+
+  if (err instanceof CLIError) {
+    return 400
+  }
+
+  return 500
```
Lezek123: I think it's good that there are no `410`s anymore, but:
- `openapi` spec still includes `410` response for `uploadFile`
- `ExtrinsicFailedError`, `WebApiError` and `CLIError` are not necessarly all `400`, a few examples of how I think they should be categorized:
  - `Max file size exceeded.` is a `400`, because the request is invalid (it's obviously client-side error)
  - `Invalid upload token request signature.` / `Provided controller account and member id don't match.` can be `400` or `401` (more specifically), as both indicate the request was not valid (or more specifically - that authroization was unsuccesful)
  - `No Joystream API loaded.` / `No Joystream account loaded.` / `No upload directory path loaded.` are `500` or at least in `5xx` category, as those are server (node) issues, not client-side errors (ie. not caused by invalid requests).

Lezek123 (changes requested):

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Fixed in 84c14da and 2b61292

Lezek123 (approved):
Looks good!
Just leaving a note that possibly `CLIError` and `ExtrinsicFailedError` will not always be `400`, so I would suggest using some more specific errors later on.

storage-node-v2/package.json
```diff
@@ -0,0 +1,110 @@
...
+    "prettier": "^2.3.0",
+    "sinon": "^11.1.1",
+    "swagger-ui-express": "^4.1.6",
+    "ts-node": "^8.8.2",
+    "type-doc": "^0.1.41",
+    "typescript": "^3.3"
+  },
+  "engines": {
+    "node": ">=8.0.0"
+  },
```
mnaamani, line 66: We are using Volta for yarn/npm version management so its a good idea to add the volta config section here:

```suggestion
  },
    "volta": {
    "extends": "../package.json"
  },
```

storage-node-v2/src/command-base/ApiCommandBase.ts
```diff
@@ -0,0 +1,135 @@
...
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    dev: flags.boolean({ char: 'm', description: 'Use development mode' }),
+    apiUrl: flags.string({
+      char: 'u',
+      description: 'Runtime API URL. Mandatory in non-dev environment. Default is ws://localhost:9944',
+    }),
+    keyfile: flags.string({
+      char: 'k',
+      description: 'Key file for the account. Mandatory in non-dev environment.',
```
mnaamani, line 28:
```suggestion
      description: 'Key file for the account. Mandatory in non-dev environment.',
      required: (flags) => {   "true if dev flag is not set"  }
```
I'm not familiar with oclif but can the value for `required` be a function that evaluates to a bool depending on what other flags were set? With `meow` we do something like this: https://github.com/Joystream/joystream/blob/8b1f4a5df3b1cdb903730e1d51c60e6a3d033cef/storage-node/packages/colossus/bin/cli.js#L59

storage-node-v2/src/command-base/ApiCommandBase.ts
```diff
@@ -0,0 +1,135 @@
...
+ */
+export default abstract class ApiCommandBase extends Command {
+  private api: ApiPromise | null = null
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    dev: flags.boolean({ char: 'm', description: 'Use development mode' }),
+    apiUrl: flags.string({
+      char: 'u',
+      description: 'Runtime API URL. Mandatory in non-dev environment. Default is ws://localhost:9944',
```
mnaamani:
```suggestion
      description: 'Runtime API URL. Default is ws://localhost:9944',
      default: 'ws://localhost:9944',
```

Doesn't appear to be required when --dev arg is not specified, and might make more sense to set default here rather than in `init()` function.

storage-node-v2/src/services/runtime/api.ts
```diff
@@ -0,0 +1,208 @@
...
+): Promise<T | void> {
+  logger.debug(`Sending ${tx.method.section}.${tx.method.method} extrinsic...`)
+
+  if (sudoCall) {
+    tx = api.tx.sudo.sudo(tx)
+  }
+  const nonce = await getNonce(api, account)
+
+  const result = await sendExtrinsic(api, account, tx, nonce)
+
```
mnaamani, line 159: Looking at this, I'm concerned that it is possible that two transactions could get submitted not in the same order of the nonce. To avoid this potential race condition in other implementations I've done the async lock before getting the nonce and releasing after submission.

storage-node-v2/src/services/runtime/transactionNonceKeeper.ts
```diff
@@ -0,0 +1,34 @@
+import { KeyringPair } from '@polkadot/keyring/types'
+import type { Index } from '@polkadot/types/interfaces/runtime'
+import BN from 'bn.js'
+import AwaitLock from 'await-lock'
+import { ApiPromise } from '@polkadot/api'
+import logger from '../logger'
+
+let nonce: Index | null = null
```
mnaamani, line 8: There is an assumption in this code that There will only ever be one key used by the application when command is executed but the code doesn’t really reflect this. If/when a command might require two accounts this implementation will break.

storage-node-v2/src/services/runtime/transactionNonceKeeper.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Return the current transaction nonce for an account from the runtime.
+ *
+ * @param api - runtime API promise
+ * @param account - KeyPair instance
+ * @returns promise with transaction nonce for a given account.
+ *
+ */
+export async function getNonce(api: ApiPromise, account: KeyringPair): Promise<Index> {
```
mnaamani, line 19: I think it would be safer for this helper function to be "private" in the `api.ts`. Because if its incorrectly used, for example its called to get next nonce which gets incremented but a transaction is not submitted, the next time the call is used to get a nonce and transaction is submitted it will have skipped a nonce and transaction will not finalize.

storage-node-v2/src/commands/server.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (flags.dev) {
+      await this.ensureDevelopmentChain()
+    }
+
+    const account = this.getAccount(flags)
+    const api = await this.getApi()
+
+    try {
+      const port = flags.port
+      const workerId = flags.worker ?? 0
```
mnaamani, line 47: It might be a good idea here to only proceed creating the app if the workerId matches the supplied worker key and that the worker is 'active', to avoid errors later. 
You could either exit, or have the server loop retrying until the credentials are valid.

storage-node-v2/src/command-base/ApiCommandBase.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Returns the intialized account KeyringPair instance. Loads the account
+   * JSON-file or loads 'Alice' Keypair when in the development mode.
+   *
+   * @param dev - indicates the development mode (optional).
+   * @param keyfile - key file path (optional).
+   * @param password - password for the key file (optional).
+   * @returns KeyringPair instance.
+   */
+  getAccount(flags: { dev?: boolean; keyfile?: string; password?: string }): KeyringPair {
```
mnaamani: It will be very practical to also give user option to pass in their key as a URI, through an environment variable. This will be useful when running commands in test scripts.

mnaamani (changes requested):
Only had a few minor change requests.
One area I think we need to revise is the upload authentication. I think we have over complicated it.

storage-node-v2/package.json
shamil-gadelshin: Changed to 
```
  "volta": {
    "node": "12.18.2",
    "yarn": "1.22.4"
  },
```
to maintain independent settings.

storage-node-v2/src/command-base/ApiCommandBase.ts
shamil-gadelshin: Not sure it's possible.

storage-node-v2/src/command-base/ApiCommandBase.ts
shamil-gadelshin: Ok

storage-node-v2/src/services/runtime/api.ts
shamil-gadelshin: Makes sense - I will change this code in the next PR because it already contains some algorithm changes.

storage-node-v2/src/services/runtime/transactionNonceKeeper.ts
shamil-gadelshin: Will remove the cache entirely in the next PR.

storage-node-v2/src/services/runtime/transactionNonceKeeper.ts
shamil-gadelshin: Will remove the cache entirely in the next PR.

storage-node-v2/src/commands/server.ts
shamil-gadelshin: I have an issue here for the later changes: [validate the worker id on the start](https://github.com/Joystream/joystream/issues/2573)

mnaamani (approved):

PR #2515, opened 2021-07-05
Storage V2 - query node schema and mappings

Lezek123:
The PR contains:
- **Update to hydra-v3 (`3.1.0-alpha.1`)**
- **Storage v2 mappings**
Currently only the ones that should be required for `storage-node` and `distributor-node` to work.
`Giza` and `Sumer` mappings have been separated and only the `Giza` mappings are currently beeing built.
Further work on updating `Sumer` mappings is required and will be part of another PR.
- **Protobuf metadata standards for Giza**
Currently two separate libraries exist: `@joystream/metadata-protobuf` (`Giza`) and `@joystream/content-metadata-protobuf` (`Sumer`) - they will be merged together in a separate PR.
- **Small `@joystream/types` adjustments taken from https://github.com/Joystream/joystream/pull/2582** (fixed `DynamicBagCreationPolicy`, added some utility types, renamed some distribution module types)


query-node/schemas/storage.graphql
```diff
@@ -0,0 +1,249 @@
+"Global storage system parameters"
```
shamil-gadelshin, line 1: Please, implement or remove TODOs.

query-node/mappings/giza/storage/index.ts
```diff
@@ -0,0 +1,636 @@
...
+    return owner
+  } else if (bagId.isMember) {
+    const owner = new StorageBagOwnerMember()
+    owner.memberId = bagId.asMember.toNumber()
+    return owner
+  } else {
+    throw new Error(`Unexpected dynamic bag type: ${bagId.type}`)
+  }
+}
+
```
shamil-gadelshin, line 90: Shouldn't we reuse the format for the `bagId` from the storage and distribution nodes (`static:council`)?

query-node/mappings/giza/storage/index.ts
```diff
@@ -0,0 +1,636 @@
...
+  } else {
+    storageBucket.operatorStatus = new StorageBucketOperatorStatusMissing()
+  }
+  await store.save<StorageBucket>(storageBucket)
+}
+
+export async function storage_StorageOperatorMetadataSet({ event, store }: EventContext & StoreContext): Promise<void> {
+  const [bucketId, , metadataBytes] = new Storage.StorageOperatorMetadataSetEvent(event).params
+  const storageBucket = await getStorageBucketWithOperatorMetadata(store, bucketId.toString())
+  storageBucket.operatorMetadata = await processStorageOperatorMetadata(
```
shamil-gadelshin, line 224: Why don't we throw an error here on invalid metadata similar to other invalid data?

query-node/mappings/giza/storage/index.ts
```diff
@@ -0,0 +1,636 @@
...
+        storageBucket: new StorageBucket({ id: bucketId.toString() }),
+      })
+  )
+  await Promise.all(assignmentsToRemove.map((a) => store.remove<StorageBagStorageAssignment>(a)))
+  await Promise.all(assignmentsToAdd.map((a) => store.save<StorageBagStorageAssignment>(a)))
+}
+
+export async function storage_StorageBucketDeleted({ event, store }: EventContext & StoreContext): Promise<void> {
+  const [bucketId] = new Storage.StorageBucketDeletedEvent(event).params
+  // TODO: Delete or just change status?
```
shamil-gadelshin: Please, resolve TODOs

query-node/schemas/storage.graphql
```diff
@@ -0,0 +1,245 @@
...
+  "Assignments to store a bag"
+  bagAssignments: [StorageBagStorageAssignment!] @derivedFrom(field: "storageBucket")
+
+  "Bucket's data object size limit in bytes"
+  dataObjectsSizeLimit: BigInt!
+
+  "Bucket's data object count limit"
+  dataObjectCountLimit: BigInt!
+
+  # TODO: Are those useful for storage node?
```
shamil-gadelshin: Please, resolve TODOs

shamil-gadelshin (changes requested):
Looks good. All current storage node queries could be modified to the working state. 

There are several issues found:
- unresolved TODOs in several files
- the `query-node/mappings` project  has failed checks (lints, formatting)
- the `metadata-protobuf` project fails linting
- Shouldn't we use the `bagId` format similar to the storage and distribution node (static:council) ?

query-node/mappings/giza/storage/index.ts
Lezek123: The errors are only thrown on some unexpected state which should never occur if the node is working correctly.
Invalid metadata is a different case, because the runtime does not check its validity, so it's still a _normal_ condition that the metadata is invalid. We're just logging this and performing no action in this case.

Throwing an error here would basically shut down the query node processor.

query-node/mappings/giza/storage/index.ts
Lezek123: That's a good idea, I think I just wasn't aware of this format when I was creating the mappings.

Lezek123:
> the query-node/mappings project has failed checks (lints, formatting)

The main reason for this is that on `master` the autogenerated `query-node` files are comitted, so the query node can be built without connecting to a running node and fetching the metadata. It also allows us to format those files to be inline with the prettier config, linter rules etc. and make some other custom changes in them.

Comitting autogenerated files has a few disadvantages though, which can be observed for `@joystream/types` (as one example) - it makes the repo substantially larger, the PRs much harder to parse and it generates lots of conflicts when merging different branches. I think on `master` we also had some manual changes in the `query-node` autogenerated files, which would need to be either automated or manually repeated each time the the files are re-generated. 

There are mutiple ways we can address this:
1. keep comitting autogenerated files the way it was done on `master`
2. make running a dockerized node a prerequisite for most/all of the checks in order to fetch the chain metadata for generation scripts
3. commit current chain metadata as json into the repo (it's ~500 KB currenty)

I would vote for option 3 as a final solution, as having chain metadata in the repo could be useful in many different ways and updating it seems quite trivial.

I think that, especially during development,  option number`1` could be quite inconvenient.

I'll try to address this issue in my next PR, since in this one the `sumer`/`master` mappings have not even been updated yet.

Lezek123:
> the metadata-protobuf project fails linting

I couldn't reproduce this issue, I'm only getting some warnings:
```
$ yarn workspace @joystream/metadata-protobuf checks
yarn workspace v1.22.4
yarn run v1.22.4
$ tsc --noEmit --pretty && prettier ./ --check && yarn lint
Checking formatting...
All matched files use Prettier code style!
$ eslint ./src --ext .ts

/home/leszek/projects/joystream/joystream-ws-1/metadata-protobuf/src/types.ts
  13:34  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  14:37  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

/home/leszek/projects/joystream/joystream-ws-1/metadata-protobuf/src/utils.ts
  8:54  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

✖ 3 problems (0 errors, 3 warnings)

Done in 11.07s.
Done in 11.55s.
```

query-node/schemas/storage.graphql
Lezek123: Done in https://github.com/Joystream/joystream/pull/2515/commits/3f04581173a720ac3a15a10c673d19875e0b4961

query-node/mappings/giza/storage/index.ts
Lezek123: Done in https://github.com/Joystream/joystream/pull/2515/commits/3f04581173a720ac3a15a10c673d19875e0b4961

query-node/mappings/giza/storage/index.ts
Lezek123: Done in https://github.com/Joystream/joystream/pull/2515/commits/3f04581173a720ac3a15a10c673d19875e0b4961

query-node/schemas/storage.graphql
Lezek123: Done in https://github.com/Joystream/joystream/pull/2515/commits/3f04581173a720ac3a15a10c673d19875e0b4961

shamil-gadelshin:
> I couldn't reproduce this issue, I'm only getting some warnings:

I meant the warnings. Sorry for the confusion.

shamil-gadelshin:
> There are mutiple ways we can address this:

If we don't want to commit the generated code shouldn't we just add the whole `generated` folder to the linter's ignore list as a simple solution?

Lezek123:
> I meant the warnings. Sorry for the confusion.

Fixed (https://github.com/Joystream/joystream/pull/2515/commits/924da13a98122fc481f78f120d637eaae485e791)

> If we don't want to commit the generated code shouldn't we just add the whole generated folder to the linter's ignore list as a simple solution?

Oh, I didn't realize linter will work on the project that does not build.
I added generated files to `prettierignore` and `eslintignore` and modified the `query-node` workflow - it now only runs linter and prettier checks, but does not attempt to actually build the node (https://github.com/Joystream/joystream/pull/2515/commits/97e3196aab10a0f1857e68aa2ffae89673f9154f)

I think this should work as a temporary solution before we decide on the metadata.

shamil-gadelshin (approved):
LGTM

PR #2513, opened 2021-07-05
runtime: fix unstaking from operations and gateway working groups

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/2505

- Adds runtime unstaking handlers for working groups Operations and Gateway
- Removed sumer on_runtime_upgrade code
- Bumped runtime spec version

PR #2523, opened 2021-07-14
Increase max working group mint capacity

mnaamani:
Council can now set working group mint capacity to 50 Million (up from 5 Million) via proposal.


Lezek123:
Looks like a small change in Pioneer proposal form validation may be needed: https://github.com/Joystream/joystream/blob/master/pioneer/packages/joy-proposals/src/validationSchema.ts#L80

mnaamani:
> Looks like a small change in Pioneer proposal form validation may be needed: https://github.com/Joystream/joystream/blob/master/pioneer/packages/joy-proposals/src/validationSchema.ts#L80

Thanks will prepare a PR

mnaamani:
@bwhm We can merge this, pioneer will auto-build so the limit increase will be allowed in the UI but fail in runtime. As long as community is aware I see no need to "pause" auto-deploy of pioneer. ACK ?

shamil-gadelshin (changes requested):
Looks good. However, it seems you forgot to update the node version. We are merging to master, so it is relevant.

mnaamani:
> Looks good. However, it seems you forgot to update the node version. We are merging to master, so it is relevant.

The node version is not as important to bump as the spec version if we haven't published a new binary, and since I bumped the node version in https://github.com/Joystream/joystream/pull/2513 I didn't bother doing it again here, but I think its good practice to always bump it. Will push an update.

mnaamani:
Integration tests are failing I think I need to update the test related to mint capacity..

Issue #2511, opened 2021-07-03
Channel subreddit


dobertRowneySr:
# Table of Contents

1.  [Goals](#org21c99e2)
2.  [Data Structures](#org522f5cd)
    1.  [Thread](#orgfc13473)
    2.  [Post](#org5e970f1)
    3.  [Category](#org388df75)
    4.  [Storage Elements](#org1b904ad)
3.  [Extrinsics](#orga8c8e2b)

****Target Branch****: `new_content_features`

<a id="org21c99e2"></a>

# Goals

1.  fully exploit the `forum` framework provided in the olympia runtime (in order to ease refactoring)
2.  Provided a forum-like async. comunication between members. One forum must be referred to

one channel and vice versa (much like a subreddit for each channel)

1.  Forum content will be text only (at least for the moment).
2.  I didn't consider polls. I also didn't consider payoff/expenses for posting. However such

features can be easily added later on

<a id="org522f5cd"></a>

# Data Structures

<a id="orgfc13473"></a>

## Thread

Here's the `Thread` data structure from the `olympia` runtime (the only thing missing are the payoff balance and the `poll` field):

    pub struct ForumThread<ForumUserId, CategoryId, Moment, Hash, Balance, ChannelId> {
        /// Title hash
        pub title_hash: Hash,
    
        /// Category in which this thread lives
        pub category_id: CategoryId,
    
        /// Author of post.
        pub author_id: ForumUserId,
    
        /// Number of posts in the thread
        pub number_of_posts: NumberOfPosts,
    
         /// channel to which this thread belongs to
        channel_id: ChannelId,
    }

<a id="org5e970f1"></a>

## Post

    pub struct Post<ForumUserId, ThreadId, Hash, Balance, BlockNumber> {
        /// Id of thread to which this post corresponds.
        pub thread_id: ThreadId,
    
        /// Hash of current text
        pub text_hash: Hash,
    
        /// Author of post.
        pub author_id: ForumUserId,
    
        /// When it was created or last edited
        pub last_edited: BlockNumber,
    }

Copycat of the post in the `forum` module on the `olympia` runtime.

<a id="org388df75"></a>

## Category

Category data structure is basically the same as the one from the olympia runtime

    pub struct Category_<CategoryId, Hash> {
        /// Title
        pub title_hash: Hash,
    
        /// Description
        pub description_hash: Hash,
    
        /// Whether category is archived.
        pub archived: bool,
    
[...]

PR #2525, opened 2021-07-15
Channel forum

dobertRowneySr:
As outlined in #2511

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -1552,14 +1552,32 @@ decl_module! {
                 &account_id,
                 &actor,
                 &channel_id,
+                &thread_id,
             )?;
 
-            //
-            // == MUTATION SAFE ==
-            //
+           // iterate over values that share the first key
+            let mut iter = <PostById<T>>::iter_prefix_values(thread_id);
+            let mut thread_cleanup_cost = thread.bloat_bond;
 
-            // Pay off to thread deleter
-//            Self::pay_off(thread_id, thread.bloat_bond, &account_id)?;
+           if let Some(post) = iter.next() {
+               thread_cleanup_cost = thread.bloat_bond.saturating_add(post.bloat_bond);
+           }
+
+           // Pay off to author or burn tokens
+           if let ContentActor::Member(member) = actor {
+               if member == thread.author_id {
+                   let _ = Self::pay_off(thread_id, thread_cleanup_cost, &account_id);
+               }
+           } else {
+                   let _ = balances::Module::<T>::burn(thread_cleanup_cost);
+           }
+
+           //
+           // == MUTATION SAFE ==
+           //
+
+           // burn all the other bloat bonds
+           let _ = iter.map(|post| balances::Module::<T>::burn(post.bloat_bond));
 
             // delete all the posts in the thread
             <PostById<T>>::remove_prefix(&thread_id);
@@ -2109,6 +2127,7 @@ impl<T: Trait> Module<T> {
         account_id: &T::AccountId,
         actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
         channel_id: &T::ChannelId,
+        thread_id: &T::ThreadId,
     ) -> DispatchResult {
         let channel = Self::ensure_channel_exists(channel_id)?;
         match actor {
@@ -2135,7 +2154,8 @@ impl<T: Trait> Module<T> {
                 // Ensure the member is the channel owner or is a moderator
                 ensure!(
                     channel.owner == ChannelOwner::Member(*member_id)
-                        || <ModeratorSetForSubreddit<T>>::contains_key(*channel_id, *member_id),
+                        || <ModeratorSetForSubreddit<T>>::contains_key(*channel_id, *member_id)
+                        || <ThreadById<T>>::get(thread_id).author_id == *member_id,
                     Error::<T>::ActorNotAuthorized
                 );
 
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -1492,14 +1492,15 @@ decl_module! {
             <PostById<T>>::insert(thread_id, new_post_id, new_post);
 
         let mut thread = <ThreadById<T>>::get(category_id, thread_id);
-        thread.number_of_posts = thread.number_of_posts.saturating_add(T::PostId::one());
+        let post_id = thread.number_of_posts.saturating_add(T::PostId::one());
+	thread.number_of_posts = post_id;
 
         <ThreadById<T>>::mutate(category_id, thread_id, |value| *value = thread);
 
             // Generate event
-            // Self::deposit_event(
-            //     RawEvent::PostAdded(post_id, forum_user_id, category_id, thread_id, text, editable)
-            // );
+             Self::deposit_event(
+                 RawEvent::PostAdded(post_id, forum_user_id, category_id, thread_id, text_hash)
+            );
 
             Ok(())
     }
@@ -1797,6 +1798,7 @@ decl_event!(
         ForumUserId = ForumUserId<T>,
         ThreadId = <T as Trait>::ThreadId,
         CategoryId = <T as Trait>::CategoryId,
+        PostId = <T as Trait>::PostId,
     {
         // Curators
         CuratorGroupCreated(CuratorGroupId),
@@ -1917,5 +1919,6 @@ decl_event!(
         ),
         PersonDeleted(ContentActor, PersonId),
         ThreadCreated(ThreadId, ForumUserId, CategoryId, Hash, ChannelId),
+        PostAdded(PostId, ForumUserId, CategoryId, ThreadId, Hash),
     }
 );
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -494,10 +494,38 @@ pub struct Thread_<ForumUserId, CategoryId, Hash, Balance, NumberOfPosts, Channe
 pub type Thread<T> = Thread_<
     ForumUserId<T>,
     <T as Trait>::CategoryId,
-    <T as frame_system::Trait>::Hashing,
+    <T as frame_system::Trait>::Hash,
     BalanceOf<T>,
     <T as Trait>::PostId,
     <T as StorageOwnership>::ChannelId,
+>;
+
+/// Represents a thread post
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct Post_<ForumUserId, ThreadId, Hash, Balance, BlockNumber> {
+    /// Id of thread to which this post corresponds.
+    pub thread_id: ThreadId,
+
+    /// Hash of current text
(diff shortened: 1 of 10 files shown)


bedeho (changes requested):
- This has a forum structure, with moderators and categories, what we want is a subreddit. So there are no categories, all threads are in one scope. 
- allow channel owner to turn subreddit off or on, and while its off, no threads or comments or reactions on comments are possible. during channel creation, the owner can decide whether to have a subreddit on or off to begin with.
- follow this updated scheme that is being introduced in the forum https://github.com/Joystream/joystream/issues/2545
- thread deletion should include deletion of all posts/replies. This should work in the same way as thread deletion in the forum, using witnesses to size of object tree, and unrolling over multiple blocks. Again, we only allow one simultaneous unrolling. In order to traverse the entire tree of posts/replies, an iterable double map with key space `ThreadIdxPostId`may be needed to organise the posts/replies.
- Introduce reactions to the thread itself, as distinct from the posts/replies.
- Both the channel owner and the content curators can act as moderators. The channel owner can also add a finite number (which must have some hardcoded upper bound) of arbitrary members to the channel which can moderate both video comments and the channel subreddit.
- It should not be possible to delete the channel if there are any threads alive.

ignazio-bovo:
In a channel subreddit, should a channel owner to delete posts / threads regardless of the author?

runtime-modules/content/src/lib.rs
```diff
@@ -654,17 +668,15 @@ decl_storage! {
...
+        hasher(blake2_128_concat) T::PostId => Post<T>;
 
-    /// Map categoryid to category
-        pub CategoryById get(fn category_by_id):
-    map hasher(blake2_128_concat) T::CategoryId => Category<T>;
+        /// Moderator set for each Subreddit
+        pub ModeratorSetForSubreddit get(fn moderator_set_for_subreddit): double_map
+            hasher(blake2_128_concat) T::ChannelId, hasher(blake2_128_concat) T::MemberId => ();
+        /// Number of subreddit moderators
+        pub NumberOfSubredditModerators get(fn number_of_subreddit_moderators) config(): u64;
```
bedeho: Why was this introduced, it appears to not be needed for any explicit purpose.

runtime-modules/content/src/lib.rs
```diff
@@ -654,17 +668,15 @@ decl_storage! {
...
-    double_map hasher(blake2_128_concat) T::ThreadId,
-    hasher(blake2_128_concat) T::PostId => Post<T>;
+        double_map hasher(blake2_128_concat) T::ThreadId,
+        hasher(blake2_128_concat) T::PostId => Post<T>;
 
-    /// Map categoryid to category
-        pub CategoryById get(fn category_by_id):
-    map hasher(blake2_128_concat) T::CategoryId => Category<T>;
+        /// Moderator set for each Subreddit
+        pub ModeratorSetForSubreddit get(fn moderator_set_for_subreddit): double_map
```
bedeho: 1. Lets keep the set of moderators for a the subreddit of a channel inside the channel itself, so somewhere inside `ChannelRecord`.  We can assume that there will be a small set of moderators, so we can just model it as a `moderators: BTreeSet<MemeberId>`, which has some cap on the max number of moderators allowed per channel.
2. In light of the change above, make sure to select a representation which is maximally safe w.r.t. other information about the subreddit currently, i.e. whether its enabled or not. Something like `ChannelRecord.subreddit_status: Option<ActiveSubredditStatus>` and `ActiveSubredditStatus` holds the moderator list in the prior point, and in the future can be expanded to hold other stuff.

runtime-modules/content/src/lib.rs
```diff
@@ -1664,157 +1775,121 @@ decl_module! {
...
-    }
+        }
 
     #[weight = 10_000_000]
-    fn delete_forum_category(
+    fn update_moderator_set(
         origin,
-        actor: PrivilegedActor<T>,
-        category_id: T::CategoryId
+        actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
```
bedeho, line 1832: There is no need to pass actor, the only one who can make a valid call to this is the owner of this channel, so we dont need this information. Please double check the `ContentActor` type and make sure you are not using it a bunch of other places on the subreddit side, as its generally not applicable there.

runtime-modules/content/src/lib.rs
```diff
@@ -1664,157 +1775,121 @@ decl_module! {
...
+        channel_id: T::ChannelId,
+        member_id: T::MemberId,
```
bedeho: Lets just simplify the semantics of this call to just provide a new value for the set of all moderators, rather then this editing semantics, so just provide a new `BTreeSet<MemberId>`, and overwrite the old one stored for the given channel.

runtime-modules/content/src/lib.rs
```diff
@@ -505,95 +538,82 @@ pub struct Thread_<ForumUserId, CategoryId, Hash, Balance, NumberOfPosts, Channe
...
+    pub author_id: MemberId,
 
-    /// Cleanup pay off
-    pub cleanup_pay_off: Balance,
+    /// Author account (used for slashing)
+    pub author_account: AccountId,
 
     /// When it was created or last edited
-    pub last_edited: BlockNumber,
+    pub creation_time: BlockNumber,
```
bedeho: Why do we need this?

The only reason to have this before was because we anted _anyon_ to be able to cleanup objects from the forum after a certain period of time, however we do not do this any longer, this is not part of this proposal 

https://github.com/Joystream/joystream/issues/2545

runtime-modules/content/src/lib.rs
```diff
@@ -505,95 +538,82 @@ pub struct Thread_<ForumUserId, CategoryId, Hash, Balance, NumberOfPosts, Channe
...
     pub text_hash: Hash,
 
     /// Author of post.
-    pub author_id: ForumUserId,
+    pub author_id: MemberId,
 
-    /// Cleanup pay off
-    pub cleanup_pay_off: Balance,
+    /// Author account (used for slashing)
+    pub author_account: AccountId,
```
bedeho: You do not keep funds in a dedicated account for each author, all state bloat bonds are pooled into a module specific built in account, the forum does this.

runtime-modules/content/src/lib.rs
```diff
@@ -505,95 +538,82 @@ pub struct Thread_<ForumUserId, CategoryId, Hash, Balance, NumberOfPosts, Channe
...
 
     /// When it was created or last edited
-    pub last_edited: BlockNumber,
+    pub creation_time: BlockNumber,
+
+    /// Whether the post is mutable
+    pub mutable: bool,
```
bedeho, line 584: We do not maintain about a post being mutable. Being mutable is the same as existing in storage, if you exist in storage, then we know who owns the post, and thus the runtime can authenticate any attempt at editing the post text. If a post has been deleted, its effectively immutable in the sense that i can no longer be edited. So you can just disregard this entire idea as an explicit concept for the runtime to deal with, it probably causes more confusion than clarity.

runtime-modules/content/src/lib.rs
```diff
@@ -1664,157 +1775,121 @@ decl_module! {
...
-    ) -> DispatchResult {
-            // Ensure data migration is done
-//            Self::ensure_data_migration_done()?;
-
+        #[weight = 10_000_000]
+        fn react_thread(origin,
+              member_id: T::MemberId,
+              thread_id: T::ThreadId,
+              react: T::ReactionId,
+              channel_id: T::ChannelId,
```
bedeho: Why would you need this? We have `Thread_::channel_id`.

runtime-modules/content/src/lib.rs
```diff
@@ -1435,227 +1453,320 @@ decl_module! {
...
+                &actor,
+                &channel_id,
+                &thread_id,
             )?;
 
-            //
-            // == MUTATION SAFE ==
-            //
+           // iterate over values that share the first key
+            let iter = <PostById<T>>::iter_prefix_values(thread_id);
```
bedeho: This is not a safe way to delete a thread, please re-read `Cleanup` section of https://github.com/Joystream/joystream/issues/2545.

runtime-modules/content/src/lib.rs
```diff
@@ -1435,227 +1453,320 @@ decl_module! {
...
                 )
             );
 
             Ok(())
      }
 
-    #[weight = 10_000_000]
-    fn delete_thread(
+       #[weight = 10_000_000]
+       fn delete_thread(
```
bedeho, line 1542: - When moderators delete threads, they must provide rationales. It may be that using the `ContentActor` type may get in the way of the safest representation here, so you may need a new type `DeletionActor` or something, which statically holds these rationales.
- There must be a way to signal whether this is a deletion which should be understood as archiving, i.e. where UIs can still show it, or if its deletion as its a mistake or bad, and should be hidden.

runtime-modules/content/src/lib.rs
```diff
@@ -1435,227 +1453,320 @@ decl_module! {
...
-            //
-            // == MUTATION SAFE ==
-            //
+           // iterate over values that share the first key
+            let iter = <PostById<T>>::iter_prefix_values(thread_id);
 
-            // Pay off to thread deleter
-//            Self::pay_off(thread_id, thread.cleanup_pay_off, &account_id)?;
+           // Pay off to author or burn tokens
+           if let ContentActor::Member(member) = actor {
```
bedeho: Are you implementing the new rules for when burning and payouts happen here? does not look like it, I don't see any capping.

runtime-modules/content/src/lib.rs
```diff
@@ -1435,227 +1453,320 @@ decl_module! {
...
-            // Delete thread
-            <ThreadById<T>>::remove(category_id, thread_id);
+           //
+           // == MUTATION SAFE ==
+           //
 
-            // decrease category's thread counter
-            <CategoryById<T>>::mutate(category_id, |category| category.num_direct_threads -= 1);
+           // burn all the other bloat bonds
+           let _ = iter.map(|post| balances::Module::<T>::burn(post.bloat_bond));
```
bedeho: A minor performance improvement you may want to keep in mind is to sometime try to aggregate payments, as each payment is a database hit (read and write, independently), which will add to the cost of the processing. We should not severely sacrifice clarity for this kind of optimization, but sometimes it can make sense.

bedeho (changes requested):
- Could you add a counter in the  `ChannelRecord` that keeps track of how many active threads exist , if any, under the subreddit for the given channel? We need this reference coutner for when we introduce channel deletion, as we do not want to allow deletion of channels with live threads. This may change in the future, but its the simplest place to start. Keep in mind that the lifetime of this counter is not affected by the subreddit getting turned on and off, as those events have no impact on the number of active threads.
- Seems like the indentation on parts of the `decl_storage` macro are not aligned.
- There are a few places where you do not strictly respect the semantics of the mutation safe syntactic boundry, such as `delete_thread`, here you are doing mutations before this mark.
- In some places you have ensure checks which also have side-effects, e.g. `transfer_to_state_cleanup_treasury_account` in  `create_post`, most of these should be possible to split up into a genuine check and the side-effect, so we retain the clear demarcation. In some places it will not be possible, but in those cases, leave a comment in the code.

dobertRowneySr:
again, as the previous PR fix. I set `cleanup_cost = 1`

Issue #2463, opened 2021-06-08
Pioneer v1: Tokenomics page doesn't show operations lead payments in USD


mochet:
Page: https://testnet.joystream.org/#/tokenomics
Problem: the USD field for Operations Lead is not populated, it appears to still be calculated in the total field though.

![image](https://user-images.githubusercontent.com/50830161/121152575-5f818b00-c856-11eb-94e6-d244d6f41aaf.png)

PR #2530, opened 2021-07-16
Fix operations lead group earning value

DzhideX:
This PR aims to fix the problem from this issue: https://github.com/Joystream/joystream/issues/2463

The problem was that the value being shown was the same value as the one from the operations group as opposed to the lead. The solution was to add operations in the (lead) check for the function that displays this data.

The simplest solution was to add `'operations'` to the if case but could lead to a similar problem in the future. I went with creating an array for working groups so we only have one place to add the value. Typescript doesn't like the includes function so I had to add some type assertions to get around it. Am open to changing this if there's a better way to do it!

### DIFF
--- a/pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
+++ b/pioneer/packages/joy-tokenomics/src/Overview/SpendingAndStakeDistributionTable.tsx
@@ -121,21 +121,21 @@ const SpendingAndStakeTableRow: React.FC<{
   );
 };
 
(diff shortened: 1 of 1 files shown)


Lezek123 (approved):
Looks good, I like the approach you've chosen with the array

bwhm (approved):
LGTM!

I'll let @mnaamani decide when to merge, as we should perhaps stack up a lot of these pioneer changes before we rebuild and deploy!

bwhm:
@mnaamani please merge!

Issue #2396, opened 2021-05-12
Olympia forum: create_thread extrinsic should take poll description, not description_hash, as arg


Lezek123:
Currently forum's `create_thread` extrinsic takes `Poll` struct as argument, which includes `description_hash` instead of `description` itself. This makes query node unable to retrieve this data.

`create_thread` should take poll `description` as one of the arguments and then hash it before committing to runtime state, same way that, for example, `create_category` takes the category description as arg and hashes it.

The same also applies to poll alternatives texts

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2531

PR #2527, opened 2021-07-15
Forum: Expose initial post_id in ThreadCreated event

iorveth:
Closes https://github.com/Joystream/joystream/issues/2403

PR #2531, opened 2021-07-16
Olympia forum: create_thread extrinsic should take poll description, not description_hash, as arg

iorveth:
Closes https://github.com/Joystream/joystream/issues/2396
Depends on: https://github.com/Joystream/joystream/pull/2527

### DIFF
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -934,11 +934,7 @@ decl_module! {
             )?;
 
             // Hash poll description and poll alternatives description
-            let poll = if let Some(poll_input) = poll_input.clone() {
-                Some(Self::from_poll_input(poll_input))
-            } else {
-                None
-            };
+            let poll = poll_input.clone().map(Self::from_poll_input);
 
             // Build a new thread
             let new_thread = Thread {
@@ -1688,7 +1684,7 @@ impl<T: Trait> Module<T> {
     }
 
     // Ensure poll alternative size is valid
-    fn ensure_poll_alternatives_length_is_valid<K>(alternatives: &Vec<K>) -> Result<(), Error<T>> {
+    fn ensure_poll_alternatives_length_is_valid<K>(alternatives: &[K]) -> Result<(), Error<T>> {
         Self::ensure_map_limits::<<<T>::MapLimits as StorageLimits>::MaxPollAlternativesNumber>(
             alternatives.len() as u64,
         )?;
--- a/runtime-modules/forum/src/lib.rs
+++ b/runtime-modules/forum/src/lib.rs
@@ -189,7 +189,7 @@ pub trait StorageLimits {
     type MaxPollAlternativesNumber: Get<u64>;
 }
 
-/// Represents all poll alternatives and vote count for each one
+/// Represents all poll alternative text hashes and vote count for each one
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
 pub struct PollAlternative<Hash> {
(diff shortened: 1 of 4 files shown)


runtime-modules/forum/src/lib.rs
```diff
@@ -200,6 +200,31 @@ pub struct PollAlternative<Hash> {
...
 
+/// Represents all poll alternative texts and vote count for each one
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct PollAlternativeInput {
+    /// alternative description
+    pub alternative_text: Vec<u8>,
+
+    /// Vote count for the alternative
+    pub vote_count: u32,
```
Lezek123: Vote count should not be specified in the input. It should be always `0` for each alternative on poll thread creation.
I think input alternatives can be represented just as a simple `Vec<u8>` in `PollInput`

runtime-modules/forum/src/lib.rs
```diff
@@ -200,6 +200,31 @@ pub struct PollAlternative<Hash> {
...
+}
+
+/// Represents a poll input
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct PollInput<Timestamp> {
+    /// description
+    pub description: Vec<u8>,
+
+    /// pallet_timestamp of poll end
```
shamil-gadelshin: Fix this comment, please.

runtime-modules/forum/src/lib.rs
iorveth: Good catch. Done https://github.com/Joystream/joystream/pull/2531/commits/b8cb5726d2a221f8e38f14ff572b03933b821b66

runtime-modules/forum/src/mock.rs
```diff
@@ -603,19 +603,17 @@ pub fn good_poll_alternative_text() -> Vec<u8> {
     b"poll alternative".to_vec()
 }
 
-pub fn generate_poll(
+pub fn generate_poll_input(
```
shamil-gadelshin, line 619: Add comments for public methods.

runtime-modules/forum/src/lib.rs
```diff
@@ -1536,6 +1566,24 @@ decl_module! {
 }
 
 impl<T: Trait> Module<T> {
+    /// Hash poll description and poll alternatives descriptions, coverting `PollInput` into `Poll`
+    fn from_poll_input(poll_input: PollInput<T::Moment>) -> Poll<T::Moment, T::Hash> {
```
shamil-gadelshin, line 1559: Could we implement  `From<PollInput>` trait instead?

runtime-modules/forum/src/mock.rs
```diff
@@ -624,11 +622,11 @@ pub fn generate_poll(
     }
 }
 
-pub fn generate_poll_timestamp_cases(
+pub fn generate_poll_input_timestamp_cases(
```
shamil-gadelshin, line 636: Add comments for public methods.

shamil-gadelshin (changes requested):
Looks good. Just a couple of minor issues.

runtime-modules/forum/src/lib.rs
iorveth: We can, i even thought about this approach, but i don`t like that it would require to pass a hash function as additional generic parameter in that case.

runtime-modules/forum/src/lib.rs
iorveth: Done

runtime-modules/forum/src/mock.rs
iorveth: Good point, done https://github.com/Joystream/joystream/pull/2531/commits/4bb08fdb7c9a5c49976eb313b279b62aa8973780

runtime-modules/forum/src/lib.rs
iorveth: https://github.com/Joystream/joystream/pull/2531/commits/4bb08fdb7c9a5c49976eb313b279b62aa8973780

runtime-modules/forum/src/mock.rs
iorveth: Done https://github.com/Joystream/joystream/pull/2531/commits/4bb08fdb7c9a5c49976eb313b279b62aa8973780

runtime-modules/forum/src/lib.rs
shamil-gadelshin: We use an additional generic parameter anyway (in the result type) and the usage of the `From<>` implementation is as easy as `let poll = poll_input.clone().map(|pi| pi.into());` . However, you use `T::calculate_hash()` inside of the conversion function and I don't see an easy way to use it inside of the From-trait implementation. Let's keep your initial approach.

shamil-gadelshin (approved):
LGTM

PR #2533, opened 2021-07-16
DevOps - Deploy Joystream node network on EKS with Kubernetes and Pulumi

ahhda:
Related to #2406 

Refer to README for instructions on how to deploy

## Manual testing
Screenshot 2021-07-28 at 2 17 25 PM

### DIFF
--- a/devops/infrastructure/node-network/index.ts
+++ b/devops/infrastructure/node-network/index.ts
@@ -3,6 +3,8 @@ import * as eks from '@pulumi/eks'
 import * as pulumi from '@pulumi/pulumi'
 import * as k8s from '@pulumi/kubernetes'
 import { configMapFromFile } from './configMap'
+import { CaddyServiceDeployment } from './caddy'
+import { getSubkeyContainers, getValidatorContainers } from './utils'
 
 const config = new pulumi.Config()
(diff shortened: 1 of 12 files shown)


devops/infrastructure/node-network/index.ts
```diff
@@ -0,0 +1,312 @@
...
+    },
+    spec: {
+      backoffLimit: 0,
+      template: {
+        spec: {
+          containers: [
+            ...subkeyContainers,
+            {
+              name: 'builder-node',
+              image: 'joystream/node:latest',
```
mnaamani, line 129: I assume this is being pulled from docker hub if not found will pulumi try to build the image?
Can we make the image a configuration parameter?

mnaamani (approved):
This worked really nicely. I left a small request which can be done in a separate PR.
There is now some common code across the pulumi projects, I can identify the Caddy service deployment for example. We should try to factor that out if possible.

devops/infrastructure/node-network/index.ts
ahhda: Yes, the image is pulled from docker hub. I'll make the image a configuration parameter.

Issue #2482, opened 2021-06-16
Improve forum deletion extrinsic signature


bedeho:
Currently it has the [following signature](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/forum/src/lib.rs#L1441) and initial processing

```Rust 
      fn delete_posts(
            origin,
            forum_user_id: ForumUserId,
            posts: Vec<(T::CategoryId, T::ThreadId, T::PostId, bool)>,
            rationale: Vec<u8>
            ) -> DispatchResult {
      
             // Check only unique post instances.
             let unique_posts: BTreeSet<_> = posts.into_iter().collect();
```

The computation of `unique_posts` is key for the security of this extrinsic, so we may as well update and make

```Rust 
 posts: BTreeMap<(T::CategoryId, T::ThreadId, T::PostId), bool>
 ```
 
 In fact, this appears to solve the current bug which allows a forum post deletor to delete the same post twice by varying the `bool` dimension.

iorveth:
Closed by https://github.com/Joystream/joystream/issues/2482

PR #2535, opened 2021-07-19
forum: improve deletion signature

iorveth:
Closes https://github.com/Joystream/joystream/issues/2482

### DIFF
--- a/runtime-modules/forum/src/benchmarking.rs
+++ b/runtime-modules/forum/src/benchmarking.rs
@@ -1710,17 +1710,19 @@ benchmarks! {
         let hide = false;
         let mut posts = BTreeMap::new();
         for _ in 0 .. k {
-            posts.insert((
+            posts.insert(
+                ExtendedPostIdObject {
                     category_id,
                     thread_id,
-                    add_thread_post::<T>(
+                    post_id: add_thread_post::<T>(
                         caller_id.clone(),
                         forum_user_id.saturated_into(),
                         category_id,
                         thread_id,
                         vec![0u8],
-                    )),
-                    hide
+                    ),
+                },
+                hide
             );
         }
 
@@ -1747,8 +1749,8 @@ benchmarks! {
         thread.number_of_posts -= k as u64;
         assert_eq!(Module::<T>::thread_by_id(category_id, thread_id), thread);
 
-        for (post, _) in &posts {
-            assert!(!<PostById<T>>::contains_key(post.1, post.2));
+        for (extended_post, _) in &posts {
+            assert!(!<PostById<T>>::contains_key(extended_post.thread_id, extended_post.post_id));
         }
 
         assert_last_event::<T>(
--- a/runtime-modules/forum/src/benchmarking.rs
+++ b/runtime-modules/forum/src/benchmarking.rs
@@ -1708,9 +1708,9 @@ benchmarks! {
             text.clone(), text.clone(), poll
         );
         let hide = false;
-        let mut posts = Vec::new();
+        let mut posts = BTreeMap::new();
         for _ in 0 .. k {
-            posts.push((
+            posts.insert((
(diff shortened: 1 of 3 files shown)


shamil-gadelshin (changes requested):
The linter is muted for this warning: `#![allow(clippy::type_complexity)]`

But the key (T::CategoryId, T::ThreadId, T::PostId) and the map `BTreeMap<(T::CategoryId, T::ThreadId, T::PostId), bool>` are very complex types. Can we introduce a struct instead of a tuple? Will it affect other systems? @Lezek123 @bedeho 

```
type ExtendedPostId<T> = ExtendedPostIdObject<T::CategoryId, T::ThreadId, T::PostId>


#[derive(Ord, .......)]
struct ExtendedPostIdObject<CategoryId, ThreadId, PostId>{
    category_id: CategoryId,
    thread_id: ThreadId, 
    post_id: PostId,
}
```

Lezek123:
> The linter is muted for this warning: `#![allow(clippy::type_complexity)]`
> 
> But the key (T::CategoryId, T::ThreadId, T::PostId) and the map `BTreeMap<(T::CategoryId, T::ThreadId, T::PostId), bool>` are very complex types. Can we introduce a struct instead of a tuple? Will it affect other systems? @Lezek123 @bedeho
> 
> ```
> type ExtendedPostId<T> = ExtendedPostIdObject<T::CategoryId, T::ThreadId, T::PostId>
> 
> 
> #[derive(Ord, .......)]
> struct ExtendedPostIdObject<CategoryId, ThreadId, PostId>{
>     category_id: CategoryId,
>     thread_id: ThreadId, 
>     post_id: PostId,
> }
> ```

I think the idea of using `BTreeMap` here is to avoid duplicated ids and I'm not sure if a struct can be used as key in `BTreeMap`? When I tried this for the purpose of checking the `BTreeMap` encoding rules it was giving me some errors during the runtime build, while tuple/primitive keys weren't (It's very possible that I was doing something else wrong though, because of my lack of experience in runtime development)

shamil-gadelshin:
> I think the idea of using `BTreeMap` here is to avoid duplicated ids, and I'm not sure if a struct can be used as key in `BTreeMap`? When I tried this for the purpose of checking the `BTreeMap` encoding rules it was giving me some errors during the runtime build, while tuple/primitive keys weren't (It's very possible that I was doing something else wrong though, because of my lack of experience in runtime development)

`BTreeMap` needs something that implements `Ord` trait (comparison) as a key. Most structs could get this implementation automatically in simple cases using `#[derive(Ord)]` attribute, and it will avoid duplicates. I don't see any possible problems from the runtime side (I could try to solve your mentioned issue with this if it is still actual). However, we might have a lot of other code that depends on the tuple (current implementation), and it would be impractical to change it now. Otherwise, the struct is a better choice, in my opinion.

Lezek123:
> However, we might have a lot of other code that depends on the tuple (current implementation), and it would be impractical to change it now.

There will surely be a few changes needed in types, query-node mappings, integration tests etc., not sure what would be the impact on the runtime side.

iorveth:
> > However, we might have a lot of other code that depends on the tuple (current implementation), and it would be impractical to change it now.
> 
> There will surely be a few changes needed in types, query-node mappings, integration tests etc., not sure what would be the impact on the runtime side.

It hope, it would be very local, unit tests & benchmarking adjustments mostly.

shamil-gadelshin:
Please, try to implement it in the runtime. @iorveth

iorveth:
> Please, try to implement it in the runtime. @iorveth

Done here https://github.com/Joystream/joystream/pull/2535/commits/5e0d4d0effd3b3f1756217c378d195569ddf276a

shamil-gadelshin (approved):
LGTM

Issue #2422, opened 2021-05-21
Olympia: Support for forum thread labels/tags


Lezek123:
List of changes needed for Olympia in order to support forum thread tags functionality:

### Runtime

- `create_thread` extrinsic - change `title` argument name to `metadata`, as it will become the encoded thread metadata which includes both thread title and tags
- `edit_thread_title` extrinsic and `ThreadTitleUpdated` event - change names to `edit_thread_metadata` and `ThreadMetadataUpdated`, as this extrinsic will be used to update both forum title and tags.

### Protobuf messages

We need a new protobuf message to represent the thread metadata:
```protobuf
message ForumThreadMetadata {
  optional string title = 1;
  repeated string tags = 2;
}
```

It will be used both when creating a new thread and updating a thread, as described in the __Query node mappings__ section below.

### Query node input schema

We need a new entity like `ForumThreadTag`, with many-to-many relationship to `ForumThread`, ie.:
```graphql
type ForumThreadTag @entity {
  "Tag name"
  name: String! @unique

  "Threads that use this tag"
  threads: [ForumThread!] @derivedFrom(field: "tags")

  "Number of threads that use this tag"
  threadsCount: Int!
}

type ForumThread @entity {
  "Runtime thread id"
  id: ID!

  "Related thread tags"
  tags: [ForumThreadTag!]

  # ...
}
```
_Notice that currently we need a separate `threadsCount` field in ForumThreadTag, because aggregate queries are not yet supported (https://github.com/Joystream/hydra/issues/195)_

### Query node mappings

After deserializing the `tags` from `ForumThreadMetadata`, the query node will:
- `slice` the array so that only the max. allowed number of tags per thread remains (ie. `4`)
- iterate over remaining tags and insert a new `ForumThreadTag` into the database for each tag that doesn't yet have a matching entry in the database (we could additionaly normalize all tags with, for example, `toLowerCase()` and `trim()`, in order to avoid duplications).

When handling `ThreadMetadataUpdated` event:
- no value for `tags` would mean we leave current thread tags unaffected
- any value (including empty array) means we override all existing tags relationships with the new ones (we don't support atomically adding / removing just a single tag)
- we must update `threadCount` accordingly in all unset / set tags.

### Pioneer v2

The input schema menioned above should be enough to support all the queries required by Pioneer, for example:

```graphql
query forumThreadTags(orderBy: name_DESC, limit: 50) {
  name
  threadsCount
} 
```
_Listing tags alphabetically_

[...]

PR #2537, opened 2021-07-19
Forum thread tags (runtime part)

iorveth:
Partially closes https://github.com/Joystream/joystream/issues/2422 (runtime part)


Lezek123 (commented):
Looks good, I didn't encounter any issues when updating query-node and tests.
I noticed there are a few places where title is still referenced, for example:
- `ensure_can_edit_thread_title`
- `good_thread_title`

But this is probably ok, considering that metadata is `title` + optionally some `tags`.

I was also wondering about the `title_hash` -> `metadata_hash` change.
I'm not sure if storing the `metadata_hash` is of any use at this point, because metadata updates can be partial and since the runtime is not aware of this, it will just store the lastest metadata update hash in this field, which can be confusing. Can we just drop the hash? Does this affect any planned functionality?

bedeho:
> I was also wondering about the title_hash -> metadata_hash change.
I'm not sure if storing the metadata_hash is of any use at this point, because metadata updates can be partial and since the runtime is not aware of this, it will just store the lastest metadata update hash in this field, which can be confusing. Can we just drop the hash? Does this affect any planned functionality?

I agree, lets drop.

iorveth:
> > I was also wondering about the title_hash -> metadata_hash change.
> > I'm not sure if storing the metadata_hash is of any use at this point, because metadata updates can be partial and since the runtime is not aware of this, it will just store the lastest metadata update hash in this field, which can be confusing. Can we just drop the hash? Does this affect any planned functionality?
> 
> I agree, lets drop.

Done

shamil-gadelshin:
Changed base branch to `olympia` because `olympia_dev` is deprecated.

bedeho (approved):

PR #2544, opened 2021-07-22
Storage-Node: Improve Sync Run

mnaamani:
Speedup the sync process, by simply allowing time for check for existence of content in IPFS we improve the number of new data objects that can be synced per "sync run". On startup expect high CPU as all objects will be effectively checked in a short time span.

Additionally dropping the need to create and update "storage-relationships" on chain also simplifies the sync process. We don't use that state in query-node/Atlas directly, or from helios. This also means the next sync run starts sooner and we can continue to process new set of content quicker.

Updating deployed storage nodes with this fix should significantly improve the UX from Atlas during playback and fetching assets as nodes will quickly sync up all the content.

Bumped version of colossus to v0.4.0 and the API version to 1.1.0 - the api version is now displayed when getting information from nodes with helios.

Improvements to helios tool to give better status of asset availability on nodes.

### DIFF
--- a/storage-node/packages/colossus/api-base.yml
+++ b/storage-node/packages/colossus/api-base.yml
@@ -1,7 +1,7 @@
 openapi: '3.0.0'
 info:
-  title: 'Joystream Storage Node API.'
-  version: '1.0.0'
+  title: 'Colossus - Joystream Storage Node'
+  version: '1.1.0'
 paths: {} # Will be populated by express-openapi
 
 components:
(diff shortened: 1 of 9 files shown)


bwhm (approved):
Ideally, we should have had time to optimize `MAX_CONCURRENT_SYNC_ITEMS` and `INTERVAL_BETWEEN_SYNC_RUNS_MS`, but this can be discovered through trial and error.

Seems to work a lot better, so great!

PR #2557, opened 2021-08-02
DevOps - Multi-arch docker build for Joystream node

ahhda:
Resolves https://github.com/Joystream/joystream/pull/2342

Creates amd64 and arm64 images for node and combines them to create a single manifest.

* amd64 build happens on the github action runner.
* arm64 build happens on an arm64 EC2 instance on AWS.

### How to Setup this Action

Create an EC2 Key Pair for the user with the name `joystream-github-action-key`.

The following secrets needs to be setup in the repository for the action to work:
* AWS_ACCESS_KEY_ID
* AWS_SECRET_ACCESS_KEY
* SSH_PRIVATE_KEY
* DOCKERHUB_USERNAME
* DOCKERHUB_PASSWORD

Screenshot 2021-08-03 at 12 17 10 AM

Successful run: https://github.com/ahhda/joystream/actions/runs/1104362647


.github/workflows/joystream-node-docker.yml
```diff
@@ -1,79 +1,177 @@
...
+                          docker_password=${{ secrets.DOCKERHUB_PASSWORD }} \
+                          tag_name=${{ steps.compute_shasum.outputs.shasum }}-${{ matrix.platform_tag }} \
+                          repository=${{ env.REPOSITORY }} dockerfile=${{ matrix.file }} \
+                          stack_name=${{ env.STACK_NAME }} platform=${{ matrix.platform }}"
+        if: ${{ steps.compute_image_exists.outputs.image_exists == 1 }}
 
+  push-manifest:
+    name: Create manifest using both the arch images
+    needs: [push-amd64, push-arm]
+    if: needs.push-amd64.outputs.image_exists == 1 || needs.push-arm.outputs.image_exists_arm == 1 || needs.push-arm.outputs.image_exists_arm64 == 1
```
mnaamani: I think a better condition would be "non existence of the final manifest" on docker-hub. It handles better the failure case of the 'push-manifest' job itself.

joystream-node-arm.Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM rust:1.52.1-buster AS rust
```
mnaamani: can we name this file `joystream-node-armv7.Dockerfile`

ahhda: Done

.github/workflows/joystream-node-docker.yml
ahhda: Resolved. We now only push a manifest if the main image tag does not exist.

scripts/runtime-code-shasum.sh
```diff
@@ -19,4 +19,5 @@ ${TAR} -c --sort=name --owner=root:0 --group=root:0 --mode 644 --mtime='UTC 2020
     runtime \
     runtime-modules \
     utils/chain-spec-builder \
-    joystream-node.Dockerfile | shasum | cut -d " " -f 1
+    joystream-node.Dockerfile \
+    joystream-node-armv7.Dockerfile | shasum | cut -d " " -f 1
```
mnaamani, line 24:
```suggestion
    node \
    joystream-node-armv7.Dockerfile | shasum | cut -d " " -f 1
```

In addition to the new Dockerfile for armv7, I realized we are also missing the `node/` folder

.github/workflows/joystream-node-docker.yml
```diff
@@ -18,62 +27,149 @@ jobs:
...
+          IMAGE=${{ env.REPOSITORY }}:${{ env.TAG_SHASUM }}
+          echo $IMAGE
+          docker pull $IMAGE-amd64
+          docker pull $IMAGE-arm64
+          docker pull $IMAGE-arm
+          docker manifest create $IMAGE $IMAGE-amd64 $IMAGE-arm64 $IMAGE-arm
+          docker manifest annotate $IMAGE $IMAGE-amd64 --arch amd64
+          docker manifest annotate $IMAGE $IMAGE-arm64 --arch arm64
+          docker manifest annotate $IMAGE $IMAGE-arm --arch arm
+          docker manifest push $IMAGE
```
mnaamani, line 175: Can we also tag the image as `joystream/node:latest` and push it as well if the branch push was to `master`

mnaamani (changes requested):
Working great, we just want to add one last to push `joystream/node:latest` when build happens on `master` branch

mnaamani:
I tried to run the image on raspberry pi3 (armv7) but I get an error:

```
The futex facility returned an unexpected error code.
Exit code 132
```

I think that means the process was killed with Signal 4 -> Illegal Instruction. So maybe the wrong image is being used?

.github/workflows/joystream-node-docker.yml
ahhda: Done

mnaamani:
Should the target be `arm-unknown-linux-gnueabihf` instead? To work on the raspberry pi ?

mnaamani:
After some testing we found that the joystream-node binary extracted from the docker image works well on the RP, so its not clear what docker is doing to cause crash.

mnaamani (approved):

PR #2559, opened 2021-08-02
Colossus: More enhancements and improved helios

mnaamani:
Series of optimizations for storage node syncing and assert serving of content

- Scan repo on startup, so we avoid doing file stat (slow and risks opening too many connections to ipfs rpc node on asset requests
- Speedup sync run, shorter intervals
- refresh objects every two minutes instead of every one minute
- improved caching of computed mapping and ipfs hash set
- pop instead of shift .. reducing computation per sync run

Improved Helios:

```
  yarn helios --help
```

```
  Helios - Storage Node diagnostics

  Usage:
    $ helios [arguments]

  Arguments (optional)
    --endpoint URL            Test only one specific colossus API endpoint (does not have to be 
                              an active worker).
                              When not specified all online storage providers will be tested.
    --limit-assets  N         Limit tests to N assets.
    --detailed                Includes list of failed assets in output report.
    --assets                  Path to a JSON file containing array of ContentIds to test.
                              When not specified all 'Accepted' content will be tested.
    --dump-assets             Dumps list of assets (content ids) which will be tested.
```

### DIFF
--- a/storage-node/packages/colossus/bin/cli.js
+++ b/storage-node/packages/colossus/bin/cli.js
@@ -143,15 +143,9 @@ function getStorage(runtimeApi, { ipfsHost }) {
     resolve_content_id: async (contentId) => {
       // Resolve accepted content from cache
       const hash = runtimeApi.assets.resolveContentIdToIpfsHash(contentId)
-      if (hash) return hash
-
-      // Resolve via API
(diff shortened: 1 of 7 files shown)


shamil-gadelshin (approved):
LGTM

mnaamani:
With giza around the corner this is not required any longer.

PR #2562, opened 2021-08-05
DevOps - Github Action for releasing binaries

ahhda:
Creates a release with title and tag provided by the user

## Testing

To test this workflow this action has to be a part of the default branch of the repository. Once that is done, you can run the workflow from the Action menu > Run workflow.

Screenshot 2021-08-05 at 1 59 42 PM

Screenshot 2021-08-10 at 2 27 22 PM

Successful run: https://github.com/ahhda/joystream/runs/3290277718?check_suite_focus=true

Sample release: https://github.com/ahhda/joystream/releases/tag/v9.9.5

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,11 +11,45 @@ on:
         required: true
 
 env:
-  REPOSITORY: ahhda/joystream
+  REPOSITORY: joystream/node
 
 jobs:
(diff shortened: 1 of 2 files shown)


mnaamani:
The names of the tarballs should have `linux-gnu` in them. The extracted binary for arm64 will not work on a mac m1 because it doesn't have a linux kernel.

Issue #2556, opened 2021-08-01
Bug: Pioneer v1 - Loading proposal votes/comments takes very long


mochet:
Problem: loading the votes/comments of a proposal as a CM takes up to 2.5 minutes. The proposal text/description will load at an expected speed (15-20 seconds) but once the page requests the comments + votes, and a user is logged on as a CM it will take an extremely long time.

Some points worth noting:
- If not logged on as a CM, this issue is not present (credit: freakstatic for this)
- While loading as a CM, the blockheight will not increase (the page is not frozen, but something weird happens with it)

mochet:
related: https://github.com/Joystream/pioneer/issues/867

bedeho:
# Hard solution

Would it be totally nuts to have the operations working group try to add query node support for the proposal system and actually update Pioneer v1 to use it? They will be able to look at the example code on the query node side and Pioneer side in our work on Pioneer v2, which has both updated?

It would make the system work fast and smooth, but it would require time for the developer to understand what is going on and build it out. Pioneer is still 2+ months away, so it may have some value both in use and for the developers to get some practice with the query node.

# Hacky solution

We change Pioneer v1 to only load pending proposals from the chain, everything else is stored as static data inside Pioneer v1 itself! This is kind of nuts, but may be a super fast hack.

mochet:
I've passed the task to Operations to see what they think about.

In the meantime some other information on this bug:
- A proposal accessed with a CM key, but the CM never voted and the proposal is finalized = 2.5 minutes to load
- A proposal accessed with a CM key, but the CM has already voted and the proposal is finalized = seconds to load
- This appears to be a UI issue, as when this bug triggers, the blockheight will freeze. If the user switches to the `explorer` view when encountering this error, the UI will suddenly process all of the blocks that were produced almost instantly.

The only changes I have found in recent time, were the introduction of the FM banners on Pioneer v1 and also a spelling correction on the actual voting form: https://github.com/Joystream/joystream/pull/2519/commits/f5c740858661e2cc89b8bc6e4b9960da7ba7ce32

freakstatic:
It seems that the problem can be seen debugging this lines:
https://github.com/Joystream/joystream/blob/b43d9e1737c741ea9cb6ae8e14b85fdbe7c4b4d2/pioneer/packages/joy-utils/src/transport/proposals.ts#L205

and by openning the proposal: 378 with the endpoint:
wss://rome-rpc-endpoint.joystream.org:9944

[...]

PR #2564, opened 2021-08-06
Improve proposal page performance when council key is used

freakstatic:
As described in [here](https://github.com/Joystream/joystream/issues/2556) votes in proposals were taking too long to load  
it seems to be related with `api.query.proposalsEngine.voteExistsByProposalByVoter.size`. 
Replacing it with `api.query.proposalsEngine.voteExistsByProposalByVoter.entries` and filtering the votes locally seems to be a lot faster :rocket: .

### DIFF
--- a/pioneer/packages/joy-utils/src/transport/proposals.ts
+++ b/pioneer/packages/joy-utils/src/transport/proposals.ts
@@ -185,12 +185,16 @@ export default class ProposalsTransport extends BaseTransport {
       proposals
     };
   }
+  
+  async voteByProposalAndMember(proposalId: ProposalId, voterId: MemberId): Promise<VoteKind | null> {
+    const votesEntries = await this.api.query.proposalsEngine.voteExistsByProposalByVoter.entries(proposalId);
+    const voteEntry = votesEntries.find((voteEntry) => {
+      const memberId = voteEntry[0].args[1] as MemberId;
 
-  async voteByProposalAndMember (proposalId: ProposalId, voterId: MemberId): Promise<VoteKind | null> {
-    const vote = (await this.proposalsEngine.voteExistsByProposalByVoter(proposalId, voterId)) as VoteKind;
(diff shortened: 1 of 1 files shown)


mochet:
without knowing the technical side of things, I can confirm based on testing that this fixes the problem that previously existed

DzhideX (approved):
Great work, thanks again for finding this :)

Regarding the functionality, it works identically to the previous implementation and I can't really find anything I'd change much in the code itself. On this account, I'm going to approve this PR.

That being said, I'm not 100% sure why this implementation works better compared to using `size()` to check if member has voted already. It probably has something to do with the underlying implementation of the function because I doubt that getting rid of the one await call would give the improvements that council members have confirmed to have experienced (going from around 2 minutes to about 10-15s).

ping @mnaamani

pioneer/packages/joy-utils/src/transport/proposals.ts
```diff
@@ -187,10 +187,14 @@ export default class ProposalsTransport extends BaseTransport {
   }
 
   async voteByProposalAndMember (proposalId: ProposalId, voterId: MemberId): Promise<VoteKind | null> {
-    const vote = (await this.proposalsEngine.voteExistsByProposalByVoter(proposalId, voterId)) as VoteKind;
-    const hasVoted = (await this.api.query.proposalsEngine.voteExistsByProposalByVoter.size(proposalId, voterId)).toNumber();
+    const votesEntries = await this.api.query.proposalsEngine.voteExistsByProposalByVoter.entries(proposalId);
```
mnaamani, line 190: I can confirm `.size(..)` is indeed taking an extended amount of time to complete. This seems to only affect double_maps.

The original implementation is using cachedAPI by accessing `this.proposalsEngine` instead of directly `this.api.query.proposalsEngine` would we get some extra speedup in loading if we do this?

```suggestion
    const votesEntries = await this.proposalsEngine.voteExistsByProposalByVoter.entries(proposalId);
```

cc: @Lezek123

mnaamani (approved):
There may be room to improve this further by using the cacheApi, but this is already a huge improvement.

PR #2566, opened 2021-08-06
Fix failing proposal tests for worker leads

mnaamani:
For a long time there has been observation that the tests specifically for hiring worker leads was failing occasionally. We would just re-run the tasks and they would pass. It was hard to re-produce so it was never properly fixed.

It turns out two jobs were trying to run the same flows and depending on the order in which they are executed was producing the failed outcome.

Example run when it fails:
https://github.com/Joystream/joystream/pull/2559/checks?check_run_id=3260753592#step:8:155

I think it would have been easier to diagnose the problem if the DispatchError was logged here: https://github.com/Joystream/joystream/blob/b43d9e1737c741ea9cb6ae8e14b85fdbe7c4b4d2/tests/network-tests/src/Fixture.ts#L79

### DIFF
--- a/tests/network-tests/src/fixtures/proposalsModule.ts
+++ b/tests/network-tests/src/fixtures/proposalsModule.ts
@@ -753,3 +753,12 @@ export class VoteForProposalFixture extends BaseFixture {
     this._events = proposalOutcome[1]
   }
 }
+
+export class VoteForProposalAndExpectExecutionFixture extends VoteForProposalFixture {
+  public async execute(): Promise<void> {
+    await super.execute()
+    if (!this.proposalExecuted) {
+      this.error(new Error('Proposal Expected to be executed'))
+    }
+  }
+}
--- a/tests/network-tests/src/fixtures/proposalsModule.ts
+++ b/tests/network-tests/src/fixtures/proposalsModule.ts
@@ -732,10 +732,22 @@ export class VoteForProposalFixture extends BaseFixture {
     const councilAccounts = await this.api.getCouncilAccounts()
     this.api.treasuryTransferBalanceToAccounts(councilAccounts, proposalVoteFee)
 
-    // Approving the proposal
+    // Catch the proposal execution result
     const proposalExecutionResult = await this.api.subscribeToProposalExecutionResult(this.proposalNumber)
-    const approvals = await this.api.batchApproveProposal(this.proposalNumber)
-    approvals.map((result) => this.expectDispatchSuccess(result, 'Proposal Approval Vote Expected To Be Successful'))
+
+    // Approve proposal
+    // const approvals =
+    await this.api.batchApproveProposal(this.proposalNumber)
+
+    // Depending on the "approval_threshold_percentage" paramater for the proposal type, it might happen
+    // that the proposal can pass before all votes are cast, and this can result in an approval vote
+    // failing with a 'ProposalFinalized' Dispatch error. Important to note that depending on how many
+    // votes make it into a block, it may not always manifest becasuse the check for proposal finalization
+    // happens in on_finalize. So we no longer assert that all vote dispatches are successful.
+    // "Not taking this into account has resulted in tests failing undeterminisctly in the past.."
+    // approvals.map((result) => this.expectDispatchSuccess(result, 'Proposal Approval Vote Expected To Be Successful'))
+
+    // Wait for the proposal to finalize
     const proposalOutcome = await proposalExecutionResult.promise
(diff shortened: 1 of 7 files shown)


mnaamani:
> It turns out two jobs were trying to run the same flows and depending on the order in which they are executed was producing the failed outcome.

That was only one if the issues. The second was a bad assertion that all proposal votes must succeed. A proposal may finalize and be executed before all votes are finalized. [details](https://github.com/Joystream/joystream/pull/2566/commits/9db517272c19bfc4d1fb1b77391816aec79573e6#diff-9397d8e6e67948afeb9d1ea396e7860d31c3baaf10a97ac16d5d9f1a2d4fedacR742)

tests/network-tests/src/flows/proposals/manageLeaderRole.ts
```diff
@@ -147,6 +150,7 @@ async function manageLeaderRole(api: Api, env: NodeJS.ProcessEnv, group: Working
 
   // Approve new leader reward
   await new FixtureRunner(voteForeLeaderRewardFixture).run()
+  assert(voteForeLeaderRewardFixture.proposalExecuted)
```
Lezek123: This assertion is missing for `voteForDecreaseStakeProposal`, `voteForSlashProposalFixture` and `terminateLeaderRoleProposalFixture` and also in `flows/proposals/workingGroupMintCapacityProposal.ts` and `flows/workingGroup/workerPayout.ts`

tests/network-tests/src/flows/membership/creatingMemberships.ts
```diff
@@ -12,7 +12,6 @@ import { assert } from 'chai'
 export default async function membershipCreation({ api, env }: FlowProps): Promise<void> {
   const debug = extendDebug('flow:memberships')
   debug('Started')
-  api.enableDebugTxLogs()
```
Lezek123, line 15: I'm actually enabling this log for all tests on `olympia` and it helped me debug a few issues in the past (perhaps it would also make debugging _this_ issue easier). This may only be slightly problematic when a runtime error is actually expected during the test.

tests/network-tests/src/fixtures/proposalsModule.ts
```diff
@@ -732,10 +732,22 @@ export class VoteForProposalFixture extends BaseFixture {
...
+    // const approvals =
+    await this.api.batchApproveProposal(this.proposalNumber)
+
+    // Depending on the "approval_threshold_percentage" paramater for the proposal type, it might happen
+    // that the proposal can pass before all votes are cast, and this can result in an approval vote
+    // failing with a 'ProposalFinalized' Dispatch error. Important to note that depending on how many
+    // votes make it into a block, it may not always manifest becasuse the check for proposal finalization
+    // happens in on_finalize. So we no longer assert that all vote dispatches are successful.
+    // "Not taking this into account has resulted in tests failing undeterminisctly in the past.."
+    // approvals.map((result) => this.expectDispatchSuccess(result, 'Proposal Approval Vote Expected To Be Successful'))
```
Lezek123, line 748: Nicely spotted!
Another way to fix this issue would be to calculate how many votes are required depending on approval quorum and threshold ([this is how I went about this on `olympia` branch](https://github.com/Joystream/joystream/blob/olympia_dev/tests/integration-tests/src/fixtures/proposals/DecideOnProposalStatusFixture.ts#L66)) in order to avoid sending any unnecessary transactions, but it's not strictly necessary.

mnaamani: That's an excellent suggestion, and with your approach we would be doing extra coverage on the threshold aspect of the proposal engine implementation itself.

tests/network-tests/src/flows/proposals/manageLeaderRole.ts
mnaamani: Well spotted. I'll add them for completeness.

tests/network-tests/src/flows/membership/creatingMemberships.ts
mnaamani: Yes indeed, that's how I was finally able to debug the cause of the failure. I think I removed them here to remove some noise while debugging.

Maybe a better approach is to do logging the failed transaction be in the `Fixture::expectDispatchSuccess()`, and in `Fixture.expectDispatchError( ..., expectedErrorName )` if error is not the same as error expected.

This way whenever we see a dispatch error logged we know something is wrong.

Lezek123 (approved):
LGMT!

PR #2568, opened 2021-08-11
Pioneer - Improve forum performance

freakstatic:
- Add local storage cache to get the forum posts of a thread without getting every time all the forum posts
- Improve selection of recent posts

### DIFF
--- a/pioneer/packages/joy-forum/src/ViewThread.tsx
+++ b/pioneer/packages/joy-forum/src/ViewThread.tsx
@@ -127,7 +127,7 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
 
 const POSTS_THREAD_MAP_CACHE_KEY = 'postsThreadMap';
 
-async function refreshPostsInThreadCache(nextPostId: PostId, api: ApiPromise) {
+async function refreshPostsInThreadCache (nextPostId: PostId, api: ApiPromise) {
   const newId = (id: number | BN) => api.createType('PostId', id);
   const apiCalls: Promise<Post>[] = [];
   let idToFetch = newId(1);
@@ -135,7 +135,7 @@ async function refreshPostsInThreadCache(nextPostId: PostId, api: ApiPromise) {
   let postsToThread = getPostsIdsInThreadCache();
   const nextThreadId = await api.query.forum.nextThreadId() as ThreadId;
 
-  if (postsToThread.size >= Number(nextThreadId.toString())) { // invalid cache
+  if (postsToThread.size >= nextThreadId.toNumber()) { // invalid cache
     postsToThread = new Map<number, number[]>();
   }
 
--- a/pioneer/packages/joy-forum/src/ViewThread.tsx
+++ b/pioneer/packages/joy-forum/src/ViewThread.tsx
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 import { Table, Button, Label, Icon } from 'semantic-ui-react';
 import BN from 'bn.js';
 
-import { ThreadId } from '@joystream/types/common';
+import { PostId, ThreadId } from '@joystream/types/common';
 import { Category, Thread, Post } from '@joystream/types/forum';
 import { Pagination, RepliesPerPage, CategoryCrumbs, TimeAgoDate, usePagination, useQueryParam, ReplyIdxQueryParam, ReplyEditIdQueryParam } from './utils';
 import { ViewReply } from './ViewReply';
@@ -22,6 +22,7 @@ import MemberPreview from '@polkadot/joy-utils/react/components/MemberByAccountP
 import { formatDate } from '@polkadot/joy-utils/functions/date';
 import { NewReply, EditReply } from './EditReply';
 import { useApi } from '@polkadot/react-hooks';
+import { ApiPromise } from '@polkadot/api/promise';
 
 type ThreadTitleProps = {
   thread: Thread;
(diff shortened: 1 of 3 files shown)


pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
 };
 
+const POSTS_THREAD_MAP_CACHE_KEY = 'postsThreadMap';
+
+async function refreshPostsInThreadCache (nextPostId: PostId, api: ApiPromise) {
+  const newId = (id: number | BN) => api.createType('PostId', id);
+  const apiCalls: Promise<Post>[] = [];
+  let id = newId(1);
+
+  let mapPostToThread = getPostsIdsInThreadCache();
```
DzhideX: (nit) I don't think we need to explicitly add map to the name as with typescript it's very easy to figure out the type of the variable. In my opinion it would be best to change it to something like `postsToThread`.

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
+  const nextTreadId = await api.query.forum.nextThreadId() as ThreadId;
```
DzhideX: (nit) Small spelling mistake here.

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
+
+  if (mapPostToThread.size >= Number(nextTreadId.toString())) { // invalid cache
+    mapPostToThread = new Map<number, number[]>();
+  }
+
+  if (mapPostToThread.size > 0) {
+    let lastPostIdInCache = 0;
```
DzhideX: I think this whole next part can be reduced to something along these lines:
```
const lastPostIdInCache = Math.max(...Array.from(mapPostToThread.values()).flat());
```
Of course you can separate it or add comments if you think it isn't as legible as your solution.

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
+
+async function refreshPostsInThreadCache (nextPostId: PostId, api: ApiPromise) {
+  const newId = (id: number | BN) => api.createType('PostId', id);
+  const apiCalls: Promise<Post>[] = [];
+  let id = newId(1);
+
+  let mapPostToThread = getPostsIdsInThreadCache();
+  const nextTreadId = await api.query.forum.nextThreadId() as ThreadId;
+
+  if (mapPostToThread.size >= Number(nextTreadId.toString())) { // invalid cache
```
DzhideX: I don't know if I'm missing something but here we can just write `nextThreadId.toNumber()` instead. There are other instances of this elsewhere in the code as well.

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
+
+      if (!postsInThread || !postsInThread.includes(lastPostIdInCache)) { // cache doesn't match the data in chain
+        mapPostToThread = new Map<number, number[]>();
+      }
+    } else {
+      mapPostToThread = new Map<number, number[]>();
+    }
+  }
+
+  while (nextPostId.gt(id)) {
```
DzhideX: While running this, I noticed this part of the code was always making one extra call even while the data was cached. To remedy that, we could just subtract one from `nextPostId`. E.g `nextPostId.sub(new BN(1)).gt(id)`

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -124,6 +125,78 @@ type ViewThreadProps = ApiProps & InnerViewThreadProps & {
...
+    apiCalls.push(api.query.forum.postById(id) as Promise<Post>);
+    id = newId(id.add(newId(1)));
+  }
+
+  const newPosts = await Promise.all<Post>(apiCalls);
+
+  for (const post of newPosts) {
```
DzhideX: I think this part can be optimized a little bit. Currently, if I'm not mistaken the time complexity would be something along the lines of `O(number_of_threads * number_of_posts_in_that_thread^2)`.

Now I'm not sure if it would really make that much of a difference since there aren't going to be too many new posts usually after caching, so I'll leave it up to you whether or not you want to change this.

pioneer/packages/joy-forum/src/ViewThread.tsx
```diff
@@ -154,20 +227,18 @@ function InnerViewThread (props: ViewThreadProps) {
...
       if (!nextPostId || totalPostsInThread === 0 || thread.isEmpty) return;
 
-      const newId = (id: number | BN) => api.createType('PostId', id);
-      const apiCalls: Promise<Post>[] = [];
-      let id = newId(1);
+      await refreshPostsInThreadCache(nextPostId, api);
+      const mapPostToThread = getPostsIdsInThreadCache();
+
+      const postIdsInThread = mapPostToThread.get(Number(thread.id.toString())) as number[];
+      const postsInThisThreadCalls: Promise<Post>[] = [];
```
DzhideX: This next part can be rewritten as:
```
const postsInThisThread = await Promise.all(postIdsInThread ? postIdsInThread.map((postId) => api.query.forum.postById(postId)) : []) as Post[];
```
Of course, you can separate this up if you want to (probably preferable :)).

DzhideX (changes requested):
Great work again :)

I have found some things that I think could be changed but let me know if any of them don't make sense (as I may lack context in certain scenarios) or if you disagree for whatever reason!

pioneer/packages/joy-forum/src/ViewThread.tsx
freakstatic: Not sure how this could be improved :thinking: 
Maybe grouping by posts per thread first and then update the map? :thinking:

DzhideX (approved):
Awesome work, thanks for taking care of this!

mnaamani (approved):
Nice set of optimizations. Tested and code looks correct.

mnaamani:
Merging this will also reduce some load on the backend rpc nodes. A good next optimization might be caching (in memory) member information so it doesn't need to be loaded all the time while browsing through the forum.
Nice work.

bwhm (approved):
Way overdue! Merging :)

PR #2588, opened 2021-08-18
query node - mapping of language iso to id

ondratra:
A little improvement in the video's language representation. Enables filtering by language in queries like `videos(where: {languageId_eq: 'en'})` before more sophisticated solution for `where` filtering of foreign fields.


mnaamani (approved):
Looks good, deterministic id is preserved and this will reduce the size of the language "table" too.

PR #2592, opened 2021-08-24
DevOps - Pulumi node-network deployment improvements

ahhda:
This PR makes minor improvements on the node-network deployed via Pulumi #2406 

* Adds instructions on how to access `node-network` service when running on Minikube
* Adds a container job to 7z the chain-spec and other secret files into an encrypted file
* Adds instructions on how to download that file

### DIFF
--- a/devops/infrastructure/node-network/README.md
+++ b/devops/infrastructure/node-network/README.md
@@ -39,7 +39,7 @@ After cloning this repo, from this working directory, run these commands:
    ```bash
    $ pulumi config set-all --plaintext aws:region=us-east-1 --plaintext aws:profile=joystream-user \
     --plaintext numberOfValidators=2 --plaintext isMinikube=true --plaintext networkSuffix=8122 \
-    --plaintext nodeImage=joystream/node:latest
+    --plaintext nodeImage=joystream/node:latest --plaintext encryptionKey=password
    ```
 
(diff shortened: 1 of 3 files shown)


devops/infrastructure/node-network/README.md
```diff
@@ -106,6 +111,12 @@ After cloning this repo, from this working directory, run these commands:
    $ kubectl exec --stdin --tty <PODNAME> -c colossus -- /bin/bash
    ```
 
+1. To get the chain-data and secrets, run the below command
+
+   ```bash
+   $ kubectl cp $(kubectl get pods | grep rpc-node | awk '{print $1}'):/chain-data/chain-data.7z ./chain-data.7z
```
mnaamani, line 117:
```suggestion
   $ kubectl cp $(kubectl get pods | grep rpc-node | awk '{print $1}'):chain-data/chain-data.7z ./chain-data.7z
```

When command is run:

```
tar: Removing leading `/' from member names
```

Removing that leading slash solved it.

mnaamani (changes requested):

devops/infrastructure/node-network/README.md
mnaamani: This was not required after all. The message was just informative not an error

PR #2596, opened 2021-08-26
channel collaborators

dobertRowneySr:
Refers to issue #2639


runtime-modules/content/src/lib.rs
```diff
@@ -203,6 +211,8 @@ pub struct ChannelRecord<MemberId, CuratorGroupId, DAOId, AccountId, VideoId, Pl
     is_censored: bool,
     /// Reward account where revenue is sent if set.
     reward_account: Option<AccountId>,
+    /// collaborator set
+    maybe_collaborators: Option<CollaboratorsSetType>,
```
bedeho: - Rename to `collaborators`.
- Drop `Option` wrapper, simply having a `BTreeSet<MemberId>` is sufficient, as it being empty has same semantics.
- You should also not need `CollaboratorsSetType` any longer.

bedeho (changes requested):
This looks quite good overall, I think we should make one non-trivial change when `update_channel` is being called by a collaborator

- it is not allowed to update the set of collaborators
- it is not allowed to update the reward account

This basic level of permissioning I think is actually a must to be functional.

runtime-modules/content/src/lib.rs
```diff
@@ -246,25 +257,29 @@ pub type ChannelOwnershipTransferRequest<T> = ChannelOwnershipTransferRequestRec
...
+pub struct ChannelUpdateParameters<ContentParameters, AccountId, CollaboratorsSetType> {
     /// Assets referenced by metadata
     assets: Option<Vec<NewAsset<ContentParameters>>>,
     /// If set, metadata update for the channel.
     new_meta: Option<Vec<u8>>,
     /// If set, updates the reward account of the channel
-    reward_account: Option<Option<AccountId>>,
+    reward_account: Option<AccountId>,
+    /// collaborator set
+    maybe_collaborators: Option<CollaboratorsSetType>,
```
bedeho: lets call this field `new_collaborators`

runtime-modules/content/src/lib.rs
```diff
@@ -189,8 +189,16 @@ pub struct ChannelCategoryUpdateParameters {
...
-{
+pub struct ChannelRecord<
+    MemberId,
+    CuratorGroupId,
+    DAOId,
+    AccountId,
+    VideoId,
+    PlaylistId,
+    SeriesId,
+    CollaboratorsSetType,
```
bedeho: You did not address why we need  `CollaboratorsSetType`, rather than just using `BTreeSet<>` inlined? It's fine if you disagree, but then you should provide a rationale that rebuts my review comment as a reply to the prior comment.

bedeho (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: OK sorry about that I forgot to mention. I used `CollaboratorSetType` because BTree set requires a key that implements
the Ord trait, This will produce a long list of errors. This is related to the previous [issue](https://github.com/Joystream/joystream/issues/2492#issuecomment-868261879).

bedeho: But I don't think the implication of this issue is that you have to use this solution, because we are not needing to do this many other places where we do have BTreeSet, right @shamil-gadelshin  ?

shamil-gadelshin: We can use `MemberId: Ord` as a generic type requirement for the type with the `BTreeSet`. That could solve the issue.

dobertRowneySr: managed to rewrite the code in order to satisfy the requirement above

bedeho (approved):
Nice, looks good to me.

PR #2611, opened 2021-09-03
Storage v2 auth schema fixes.

shamil-gadelshin:
This PR intends to update the authentication schema for the uploading. It changes the runtime, type, and storage-node projects.

#### Changes
- remove obsolete `authentication_key` from uploading parameters
- add `MaxDataObjectSize` runtime constant (defaulted to 10 GB)
- add `ipfs_content_id` to the `DataObject` runtime state
- add CID and size verification on the data uploading

### DIFF
--- a/storage-node-v2/src/services/webApi/app.ts
+++ b/storage-node-v2/src/services/webApi/app.ts
@@ -114,8 +114,6 @@ function validateUpload(api: ApiPromise, account: KeyringPair): ValidateUploadFu
   // We don't use these variables yet.
   /* eslint-disable @typescript-eslint/no-unused-vars */
   return (req: express.Request, scopes: string[], schema: OpenAPIV3.SecuritySchemeObject) => {
-    return true
-
     const tokenString = req.headers['x-api-key'] as string
     const token = parseUploadToken(tokenString)
 
--- a/storage-node-v2/src/services/webApi/app.ts
+++ b/storage-node-v2/src/services/webApi/app.ts
@@ -76,12 +76,17 @@ export async function createApp(
 
   /* eslint-disable @typescript-eslint/no-unused-vars */
   app.use((err: Error, req: express.Request, res: express.Response, next: express.NextFunction) => {
-    // Request validation error handler.
+    // Express error handling recommendation:
+    // https://expressjs.com/en/guide/error-handling.html
+    if (res.headersSent) {
+      return next(err)
+    }
+
+    // Request error handler.
     if (err instanceof HttpError) {
-      res.status(err.status).json({
-        type: 'request_validation',
+      res.status(err.status || 500).json({
+        type: 'request_exception',
         message: err.message,
-        errors: err.errors,
       })
     } else {
       res.status(500).json({
--- a/storage-node-v2/src/services/webApi/app.ts
+++ b/storage-node-v2/src/services/webApi/app.ts
@@ -17,13 +17,15 @@ import { httpLogger, errorLogger } from '../../services/logger'
  * @param account - KeyringPair instance
  * @param workerId - storage provider ID (worker ID)
  * @param uploadsDir - directory for the file uploading
+ * @param maxFileSize - max allowed file size
  * @returns Express promise.
  */
 export async function createApp(
   api: ApiPromise,
   account: KeyringPair,
   workerId: number,
-  uploadsDir: string
+  uploadsDir: string,
+  maxFileSize: number
 ): Promise<Express> {
   const spec = path.join(__dirname, './../../api-spec/openapi.yaml')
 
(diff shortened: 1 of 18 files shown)


storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -107,10 +110,14 @@ export async function uploadFile(req: express.Request, res: express.Response): P
     const fileObj = getFileObject(req)
     cleanupFileName = fileObj.path
 
-    verifyFileSize(fileObj.size)
+    const api = getApi(res)
+    await verifyFileSize(api, fileObj.size)
```
Lezek123: Instead of checking the filesize after it's fully uploaded, perhaps you could use `limits.fileSize` property in `multer` config? (`OpenApiValidator.middleware` -> `fileUploader`)

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -296,17 +301,59 @@ async function validateTokenRequest(api: ApiPromise, tokenRequest: UploadTokenRe
...
+  bagId: BagId,
+  dataObjectId: number,
+  fileSize: number,
+  hash: string
+) {
+  const dataObject = await api.query.storage.dataObjectsById(bagId, dataObjectId)
+
+  if (dataObject.accepted.valueOf()) {
+    throw new WebApiError(`Data object had been already accepted ID = ${dataObjectId}`, 400)
+  }
```
Lezek123: I'm not sure what was the conclusion on allowing already accepted objects, since if the "accepting" node goes down before propagating the data - reuploading it will become impossible.

Another issue is that this opens up an attack vector where storage provider could effectively block lots of uploads by sending `acceptDataObject` transactions without ever recieving the objects, which may be an issue even considering that the provider will then be slashed at some point.

And the final note is that this check will not always work either, because some other `acceptDataObject` tx(s) may be already pending on-chain confirmation (or will be sent/recieved by other nodes earlier). Therefore my suggestion is to not use it for anything other than skipping an attempt to send `acceptDataObject` if the object has already been accepted (but still keeping the actual uploaded data in that case)

storage-node-v2/src/services/webApi/controllers/publicApi.ts
```diff
@@ -296,17 +301,59 @@ async function validateTokenRequest(api: ApiPromise, tokenRequest: UploadTokenRe
...
+
+  // Cannot get 'size' as a regular property.
+  const probablyDataObjectSize = dataObject.get('size') as unknown
+  const dataObjectSize = probablyDataObjectSize as BN
```
Lezek123:
```suggestion
  const dataObjectSize = dataObject.getField('size')
```

Lezek123 (changes requested):

Lezek123:
Tested the uploads, I noticed one issue when requesting the token with invalid signature:
```
Internal Server Error: no schema defined for status code '401' in the openapi spec
```

Other than that everything seems to work fine

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Very nice finding! Implemented.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: Fixed (I got errors previously using this code but I couldn't reproduce it again).

shamil-gadelshin:
> Internal Server Error: no schema defined for status code '401' in the openapi spec

Fixed.

storage-node-v2/src/services/webApi/controllers/publicApi.ts
shamil-gadelshin: We haven't decided on the 'allowing already accepted objects' issue yet, and I didn't add the 'data object accepted' check on the runtime side. I was planning to add the 'storage node side' check as a fail-fast solution. 
Regarding possible attack vectors from the storage providers - transaction fees and slashing are our current mitigations.

However, I like your idea because we are still thinking about re-uploading data objects on the failure of the single holder (we also need to think about the DoS from the client of the storage node - multiple uploading of the same large file if we allow re-uploading the data object).

Anyway, It's easy to make it strict again (possibly during the auth-token creation to counter mentioned DoS).

storage-node-v2/src/services/webApi/app.ts
```diff
@@ -103,6 +114,8 @@ function validateUpload(api: ApiPromise, account: KeyringPair): ValidateUploadFu
   // We don't use these variables yet.
   /* eslint-disable @typescript-eslint/no-unused-vars */
   return (req: express.Request, scopes: string[], schema: OpenAPIV3.SecuritySchemeObject) => {
+    return true
+
```
Lezek123: I think this got comitted by accident
```suggestion
```

Lezek123 (changes requested):
Did a few more tests, eveything works pretty much as expected, just noticed one more issue with the upload responses:

**Example 1:** (no space left)
```
{"type":"request_validation","message":"ENOSPC: no space left on device, write"}
```
**Example 2:** (uploads dir has been deleted)
```
{"type":"request_validation","message":"ENOENT: no such file or directory, open 'uploads-dir/d030312187891a709f787fde94bec1e8'"}
```

It seems like all errors that reach express error handler are treated as `request_validation` errors, regardless of the actual cause.

storage-node-v2/src/services/webApi/app.ts
shamil-gadelshin: Yes. I implemented a flag disabling the auth in my latest branch (but not here).

Lezek123 (approved):
LGTM!

PR #2613, opened 2021-09-03
Query node - error prevention for data object deletion

### LINKED ISSUE #2575: Sumer query node - update metadata on content deletion
Sumer release query node has doesn't updated meta fields like `avatarPhotoAvailability`, etc. are not updated when content is deleted via event `content_ChannelAssetsRemoved`. Proper unsetting is done in `dataDirectory_ContentRemoved`.

Ensure `content_ChannelAssetsRemoved` properly unsets all fields that need it (like `avatarPhotoAvailability`). Then try to look for any other mappings that remove content and are having the same issue.

ondratra:
Fixes #2575.

There was only one place that needed the safeguard. I was wondering if we should also delete assets when channel/video is deleted, but it seems that runtime does no such thing. Is that correct?

### DIFF
--- a/query-node/mappings/src/content/channel.ts
+++ b/query-node/mappings/src/content/channel.ts
@@ -13,6 +13,7 @@ import {
   convertContentActorToChannelOwner,
   convertContentActorToDataObjectOwner,
 } from './utils'
+import { disconnectDataObjectRelations } from '../storage'
 
 import { Channel, ChannelCategory, DataObject, AssetAvailability } from 'query-node'
 import { inconsistentState, logger } from '../common'
@@ -127,6 +128,10 @@ export async function content_ChannelAssetsRemoved(db: DatabaseManager, event: S
 
   // delete assets
   for (const asset of assets) {
+    // ensure dataObject is nowhere used to prevent db constraint error
+    await disconnectDataObjectRelations(db, asset)
+
+    // remove data object
     await db.remove<DataObject>(asset)
   }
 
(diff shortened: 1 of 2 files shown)


query-node/mappings/src/content/channel.ts
```diff
@@ -127,6 +128,10 @@ export async function content_ChannelAssetsRemoved(db: DatabaseManager, event: S
 
   // delete assets
   for (const asset of assets) {
+    // ensure dataObject is nowhere used to prevent db constraint error
+    await disconnectDataObjectRelations(db, asset)
```
mnaamani, line 132: I think this is already handled by storage mappings which will process the `ContentRemoved` event first:

https://github.com/Joystream/joystream/blob/432e9ffb3a3503109fc7299a6b0a451960506f9a/query-node/mappings/src/storage.ts#L69

The event `ChannelAssetsRemoved` is emitted only after the content directory calls `T::StorageSystem::atomically_remove_content()`

https://github.com/Joystream/joystream/blob/432e9ffb3a3503109fc7299a6b0a451960506f9a/runtime-modules/content/src/lib.rs#L767

ondratra: You might be right. Are you sure the event `ContentRemoved` is really triggered in this flow? It's just that when looking into `atomically_remove_content` https://github.com/Joystream/joystream/blob/432e9ffb3a3503109fc7299a6b0a451960506f9a/runtime-modules/storage/src/data_directory.rs there are no events emitted.

mnaamani: Oh you are absolutely correct, event is not emitted in atomically_remove_content()

PR #2638, opened 2021-09-10
Update protobuf library and mappings for Giza

Lezek123:
Based on https://github.com/Joystream/joystream/pull/2674

**Scope:**
- Updates `@joystream/metadata-protobuf` library - merges storage and content protobufs
- Updates mappings merged from `master` to Hydra v3 and makes them compatible with the new `metadata-protobuf` library which uses `protobufjs` compiler
- Updates content mappings for compatibility with Storage v2
- Resolves https://github.com/Joystream/joystream/issues/2703


query-node/mappings/workingGroup.ts
```diff
@@ -0,0 +1,153 @@
...
+/// ///////////////// Helpers ////////////////////////////////////////////////////
+
+function getWorkerType(event: SubstrateEvent): WorkerType | null {
+  if (event.section === 'storageWorkingGroup') {
+    return WorkerType.STORAGE
+  } else if (event.section === 'gatewayWorkingGroup') {
+    return WorkerType.GATEWAY
+  }
+  return null
+}
```
ondratra: A cosmetic tip. According to the so-called Object Calisthenics, it is usually a good idea to evade writing of `else`/`elseif` part of conditions and rather escape function inside `if` clause or create a helper function. 

We can simply remove `else` part in this particular function because `if` part escapes the function already.

```suggestion
  if (event.section === 'storageWorkingGroup') {
    return WorkerType.STORAGE
  } 
  
  if (event.section === 'gatewayWorkingGroup') {
    return WorkerType.GATEWAY
  }
  
  return null
}
```

query-node/mappings/content/utils.ts
```diff
@@ -0,0 +1,517 @@
...
+  if (assets[index]) {
+    return assets[index]
+  } else {
+    invalidMetadata(`Invalid${name ? ' ' + name : ''} asset index`, {
+      numberOfAssets: assets.length,
+      requestedAssetIndex: index,
+    })
+
+    return null
+  }
```
ondratra:
```suggestion
  if (assets[index]) {
    return assets[index]
  }

  invalidMetadata(`Invalid${name ? ' ' + name : ''} asset index`, {
    numberOfAssets: assets.length,
    requestedAssetIndex: index,
  })

  return null
```

query-node/mappings/content/utils.ts
```diff
@@ -0,0 +1,517 @@
...
+    // FIXME: Ideally the runtime would provide object ids in ChannelCreated/VideoCreated/ChannelUpdated(...) events
+    const objects = await getMostRecentlyCreatedDataObjects(ctx.store, assetsUploaded)
+    return objects.map((o) => {
+      const resultAsset = new AssetJoystreamStorage()
+      resultAsset.dataObjectId = o.id
+      return resultAsset
+    })
+  } else {
+    unexpectedData('Unrecognized assets type', assets.type)
+  }
```
ondratra:
```suggestion
  if (assets.isUrls) {
    return assets.asUrls.map((assetUrls) => {
      const resultAsset = new AssetExternal()
      resultAsset.urls = JSON.stringify(assetUrls.map((u) => u.toString()))
      return resultAsset
    })
  }

  if (assets.isUpload) {
    const assetsUploaded = assets.asUpload.object_creation_list.length
    // FIXME: Ideally the runtime would provide object ids in ChannelCreated/VideoCreated/ChannelUpdated(...) events
    const objects = await getMostRecentlyCreatedDataObjects(ctx.store, assetsUploaded)
    return objects.map((o) => {
      const resultAsset = new AssetJoystreamStorage()
      resultAsset.dataObjectId = o.id
      return resultAsset
    })
  }

  unexpectedData('Unrecognized assets type', assets.type)
```

query-node/mappings/content/utils.ts
```diff
@@ -0,0 +1,517 @@
...
+      }
+    })
+    await store.save<Video>(video)
+
+    // emit log event
+    logger.info('Content has been disconnected from Video', {
+      videoId: video.id.toString(),
+      dataObjectId: dataObject.id,
+    })
+  }
```
ondratra:
```suggestion
  if (channel) {
    channelAssets.forEach((assetName) => {
      if (channel[assetName] && (channel[assetName] as AssetJoystreamStorage).dataObjectId === dataObject.id) {
        channel[assetName] = new AssetNone()
      }
    })
    await store.save<Channel>(channel)

    // emit log event
    logger.info('Content has been disconnected from Channel', {
      channelId: channel.id.toString(),
      dataObjectId: dataObject.id,
    })
  }

  if (video) {
    videoAssets.forEach((assetName) => {
      if (video[assetName] && (video[assetName] as AssetJoystreamStorage).dataObjectId === dataObject.id) {
        video[assetName] = new AssetNone()
      }
    })
    await store.save<Video>(video)

    // emit log event
    logger.info('Content has been disconnected from Video', {
      videoId: video.id.toString(),
      dataObjectId: dataObject.id,
    })
  }
```

query-node/mappings/membership.ts
```diff
@@ -32,26 +23,27 @@ export async function members_MemberRegistered(db: DatabaseManager, event: Subst
     id: memberId.toString(),
     rootAccount: accountId.toString(),
     controllerAccount: accountId.toString(),
-    handle: convertBytesToString(handle.unwrapOr(null)),
-    about: convertBytesToString(about.unwrapOr(null)),
-    avatarUri: convertBytesToString(avatarUri.unwrapOr(null)),
+    // Handle is required by the runtime during registration. Lack of it will throw an error
+    handle: bytesToString(handle.unwrap()),
+    about: about.isSome ? bytesToString(about.unwrap()) : undefined,
+    avatarUri: avatarUri.isSome ? bytesToString(avatarUri.unwrap()) : undefined,
```
ondratra, line 29: Are we ok with `unwrap`s there, or should there rather be `unwrapOr`? Handle etc. are optional parameters in the runtime extrinsic.

ondratra (approved):
Looks great. I've added some cosmetic improvement suggestions.

My only question is: were the mapping tested with Giza chain used by other developers? Running mappings on the actively used chain could reveal problems in some complex/unexpected situations like it did on Sumer.

query-node/mappings/membership.ts
Lezek123: `isSome` checks whether the option is set, so we can then safely use `unwrap()`. I've seen this pattern in https://github.com/polkadot-js/apps

ondratra: That's right. `handle` is missing `isSome` check, maybe we should add it there?

Lezek123: There is a comment above explaining this:
```
 Handle is required by the runtime during registration. Lack of it will throw an error.
```
Source: https://github.com/Joystream/joystream/blob/giza/runtime-modules/membership/src/lib.rs#L612

Lezek123:
> My only question is: were the mapping tested with Giza chain used by other developers? Running mappings on the actively used chain could reveal problems in some complex/unexpected situations like it did on Sumer.

If you mean the content mappings, they weren't very thoroughly tested yet, but I did some tests locally using updated CLI (https://github.com/Joystream/joystream/pull/2708), for example, by running https://github.com/Joystream/joystream/pull/2708/files#diff-421846dcf3d6068b65f5c289c38252a96a2e682d58ffb4f1b8d63d4ef076c6ae and executing some queries.

The Giza staging network will be setup soon, so we will have a more solid testing ground.
There will be also integration tests created for the mappings.
And finally we're going to export & verify content from the current testnet, which is going to be another big test.

Lezek123:
New changes:
- Modified `DistributionBucketFamilyMetadata` standard to better support the use-cases described in [distributor node docs](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/docs/node/index.md#distributionbucketfamilymetadata)
- Bug fixes
- [Updates following runtime changes in content and storage modules](https://github.com/Joystream/joystream/pull/2638/commits/075914abf8b431146baae34ee89241bf156e050b) (mainly https://github.com/Joystream/joystream/pull/2729 and https://github.com/Joystream/joystream/pull/2741)

ondratra:
I looked through the recent changes and ran the tests after merging in CLI PR locally. LGTM!

ondratra (approved):

PR #2653, opened 2021-09-14
Giza relationships querying workaround

Lezek123:
1. Input schema workaround for allowing the use of queries like:
```
  storageBags(
    where: {
      storageAssignments_some: {
        storageBucketId_in: ["1", "2"]
      }
    }
  ) {
    id
  }
```

Since relations like `storageBucket` and `storageBag` already create `storage_bucket_id` and `storage_bag_id` fields in the database that are populated when the relationship is saved, no changes in the mappings are required.

2. Fix for `@joystream/metadata-protobuf` typings which caused query node build to fail


shamil-gadelshin (approved):
Both fixes worked.

Issue #2656, opened 2021-09-14
NFT: English auction extension period


iorveth:
- Extension period means that the remaining time on a lot will automatically reset to to the preset extension time (between 1-10 minutes) if a new bid is placed within that period. The lot extension period is set up by the auctioneer, and can vary from auction to auction.

Should be useful to avoid nasty last minute battles & software bidders;)

iorveth:
Closed by https://github.com/Joystream/joystream/pull/2667

PR #2512, opened 2021-07-03
VNFT auction

iorveth:
PR tracking progress with https://github.com/Joystream/joystream/issues/2481

- [x] vNFT auction pallet extrinsics:
  - [x] start_video_auction
  - [x] cancel_video_auction
  - [x] complete_video_auction
  - [x] settle_open_auction
  - [x] make_bid
  - [x] cancel_open_auction_bid
  - [x] offer_vnft
  - [x] cancel_offer
  - [x] accept_incoming_offer
  - [x] sell_vnft
  - [x] buy_vnft
  - [x] issue
- [x] runtime integration

Closes https://github.com/Joystream/joystream/issues/2702, https://github.com/Joystream/joystream/issues/2481

PR #2667, opened 2021-09-15
 Feature: auction extension period

iorveth:
Closes https://github.com/Joystream/joystream/issues/2656,
https://github.com/Joystream/joystream/issues/2771

Depends on https://github.com/Joystream/joystream/pull/2512


runtime-modules/content/src/nft/types.rs
```diff
@@ -436,14 +450,32 @@ pub struct AuctionParams<VideoId, BlockNumber, Balance, MemberId: Ord> {
...
     }
 }
+
+/// English auction details
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct EnglishAuctionDetails<BlockNumber> {
+    // the remaining time on a lot will automatically reset to to the preset extension time
+    // if a new bid is placed within that period
+    pub extension_period: BlockNumber,
```
bedeho, line 472: Is providing a value of `0` the way you avoid having any extension at all?

bedeho (changes requested):
- Am I correct that we are still automatically closing auctions if a buy now bid is made, even if its within the extension window? I think that is the most appropriate.
_ I think it would be useful to clearly signal in the event `AuctionBidMade`  when the auction is extended.

runtime-modules/content/src/nft/types.rs
iorveth: Yes, it's correct and possible in theory.
Though we also have safety bounds for both `English` `auction_duration` and `extension_period` parameters.

iorveth: I added additional safety check here as well https://github.com/Joystream/joystream/pull/2667/commits/5e271e3a29aac70b7be46f23eb83570c8286e4e3

iorveth:
> * Am I correct that we are still automatically closing auctions if a buy now bid is made, even if its within the extension window? I think that is the most appropriate.
>
You are correct, this features don't cause any conflict with each other 
>   _ I think it would be useful to clearly signal in the event `AuctionBidMade`  when the auction is extended.
>
Done here https://github.com/Joystream/joystream/pull/2667/commits/d6f95b93b126a3cfbcfa7b34d11859360e7e0779

bedeho (approved):
Looks good!

PR #2547, opened 2021-07-26
Olympia mappings - runtime forum/proposals changes adjustments, support for forum tags and proposals discussion

Lezek123:
Includes merged and updated work from https://github.com/Joystream/joystream/pull/2436 (proposals discussion mappings) and adjustments related to changes introduced in:
- https://github.com/Joystream/joystream/pull/2527 - `PostId` is now retriveable from `ThreadCreated` event data
- https://github.com/Joystream/joystream/pull/2531 - `PollInput` type was introduced, which fixes the issue related to forum poll input data
- https://github.com/Joystream/joystream/pull/2537 - `ThreadTitle` -> `ThreadMetadata` runtime renaming for consistency due to the introduction of `ForumThread` metadata in order tu support forum tags (https://github.com/Joystream/joystream/issues/2422)
- https://github.com/Joystream/joystream/pull/2539 - `ProposalId` was incuded as part of `proposalCodex.ProposalCreated` event and `proposalEngine.ProposalCreated` event was removed, unifying the source of new proposal data and allowing to remove the temporary workaround with cached proposal id.

**Notable changes:**
- Forum tags are now supported as desribed in https://github.com/Joystream/joystream/issues/2422. `create_thread` extrinsic and `edit_thread_metadata` (previously `edit_thread_title`) now take encoded `ForumMetadata` as argument. The `ThreadTitleUpdatedEvent` entity has been renamed to `ThreadMetadataUpdatedEvent`
- `ForumTag.visibleThreadsCount` and `ForumThread.visiblePostsCount` fields have been added, which will faciliate querying forum threads / tags by popularity (required by Pioneer 2). `visible` means that the posts/threads beeing counted are either `Active` or `Locked`, so this excludes posts/thread that are (or in case of post - are part of a thread that is) `Moderated` or `Removed`.
- Initial post in the thread will now have a valid `ID`
- The input format for polls in `create_thread` has been changed to take the actual input data instead of hashes.
- `delete_posts` extrinsic signature has been changed. The extrinsic now takes `BTreeMap` as argument, which may require updating to `@polkadot/api` `5.2.1` (see: https://github.com/polkadot-js/api/pull/3789) _(this update was already done on this branch)_

PR #2672, opened 2021-09-17
query node - council & referndum input schema + generic event mappings

ondratra:
This PR contains a draft for council & referendum input schema + mappings for generic event data.

Before merging this PR, the #2547 should be merged first as it's needed to compile properly.

Output schema including changes from this PR as well as #2547 can be found here https://gist.github.com/ondratra/200850865b0c5b6de908bc3cedfdc554 .


query-node/schemas/membership.graphql
```diff
@@ -80,6 +80,14 @@ type Membership @entity {
 
   "Content channels the member owns"
   channels: [Channel!] @derivedFrom(field: "ownerMember")
+
+  # Council & Referendum relations
+
```
jodator, line 88: It would be helpful to have `isCouncil` flag or any other means to allow filtering members by council status. I don't remember if we could filter on the `councilMembers` field.

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,224 @@
...
+  "Council that was elected in this election round."
+  nextElectedCouncil: ElectedCouncil
+
+  "Candidates in this election round."
+  candidates: [Candidate!] @derivedFrom(field: "cycleId")
+}
+
+# TODO:
+# - is it useful as defined below?
+# - do we need to add records for BudgetRefill, BudgetIncrement, etc. or is it ok to read this info from councilEvents?
```
jodator: In Pioneer, we're taking some of the data from the Joystream node using Polkadot API, so not from the query-node.

jodator (commented):
I've glimpsed through this draft. For now I can add that Membership could have `isCouncil` flag to allow easy filtering for this. It is used on the Members page, where you can browse all members and search for particular ones.

query-node/schemas/membership.graphql
ondratra: `isCouncilMember` has been added to the membership.

query-node/schemas/council.graphql
ondratra: Ok. In case it does makes sense, you can also query events themselves as they are saved into query node as well (as can be seen in `councilEvents.graphql`).

query-node/schemas/membership.graphql
jodator: Thanks!

bedeho: Hey, this information is implicitly available through field `Membership::councilMembers`, is that not usable?

jodator: We had some difficulties filtering by relation. Additionally, we would have to construct a `where` that will account for:

*   current council member
*   either is in the relationship or not (depending on what we would like to search)

In other words, the `where` clause might be complicated. Thus having this is preferable:

```graphql
query GetMemers(where:{isCouncilMember_eq:true, isVerified_eq:true}) {
  id
  handle
  #...
}
```

ondratra: There are some improvements for relation filtering coming so we can revisit this part after that. Until then we can stick with possibly redudant `isCouncilMember`.

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
+################### Council ####################################################
+
+type AnnouncingPeriodStartedEvent @entity {
```
jodator: @ondratra 

Just found out during working on the Forum events. We need events to implement the `Event` interface so we could query them at once.

See here: [https://github.com/Joystream/joystream/blob/olympia_dev/query-node/schemas/membershipEvents.graphql#L1](https://github.com/Joystream/joystream/blob/olympia_dev/query-node/schemas/membershipEvents.graphql#L1)

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,220 @@
...
+}
+
+union CouncilStage = CouncilStageAnnouncing | CouncilStageElection | CouncilStageIdle | VariantNone
+
+type Candidate @entity {
+  "Account used for staking currency needed for the candidacy."
+  stakingAccountId: String!
+
+  "Account that will receive rewards if candidate's elected to the council."
+  rewardAccountId: String!
```
jodator, line 54:
```suggestion
  "Candidate's membership."
  member: Membership!

  "Account used for staking currency needed for the candidacy."
  stakingAccountId: String!

  "Account that will receive rewards if candidate's elected to the council."
  rewardAccountId: String!
```

A member is needed to display information about the candidate.

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,220 @@
...
+
+  "Election cycle"
+  cycleId: ElectionRound!
+  stake: BigInt!
+  votePower: BigInt!
+  note: String!
```
jodator: @ondratra I think that we also need the `Note` [metadata](https://joystream.gitbook.io/joystream-handbook/key-concepts/encodings#council-candidacy-note) to be expended and included in the GraphQL mappings – as other metadata is stored in the query-node. For instance, the [membership metadata](https://joystream.gitbook.io/joystream-handbook/key-concepts/encodings#membership-metadata) is available on the [`Member` entity](https://github.com/ondratra/joystream/blob/fb2d6b604bb98c338db415e65627e9baae6d9f24/query-node/schemas/membership.graphql#L38-L39).

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/council.graphql
ondratra: Done.

query-node/schemas/council.graphql
ondratra: I can add the metadata field for sure, but I don't see such info in the runtime. I will check how it's done for membership - maybe it'ss somehow packed.

jodator: AFAIR those are protobuf-encoded fields: https://github.com/Joystream/joystream/blob/olympia_dev/metadata-protobuf/compiled/index.d.ts#L220-L233 and those were unpacked and stored directly into the query-node.

In the runtime you only store `hash` of the note IIRC.

ondratra:
Changes to input schema and mappings were pushed into this PR. To run the processor on a local machine run this:
```
docker-compose build joystream-node

yarn workspace query-node-root build # builds everything query-node related
# output schema can now be found at `query-node/generated/graphql-server/generated/schema.graphql`

# optional - sometimes might be needed
docker-compose up -d db # run db so we can make sure it's clean
yarn workspace query-node-root db:drop # clear db

yarn workspace query-node-root start # starts all services
```

jodator:
> Changes to input schema and mappings were pushed into this PR. To run the processor on a local machine run this:

@ondratra The mappings so far look OK. We still need that `note` to be available to query as this holds the all information about the candidacy.

I wasn't able to run yarn commands – I get errors about missing `query-node` workspace. I also, for some reason, build the docker image. I can build & run the joystream-node locally, but the docker command fails on my Mac M1 :/ In other words, I will probably test those on testnet to not spend too much time debugging my joystream repo setup.

ondratra:
@jodator Candidacy metadata has been added. I've put the new output schema into the gist mentioned at the top of this thread, so you don't have to generate it on your own. At this point, I think it will be better to wait for testnet instead of trying to debug the build, as you suggested. 

Regarding the Mac M1 - I know there was some work done on that, and I got an impression it's already fixed, but it's possible that the fix is not present in this PR as various PRs are being merged into Olympia branch these days. So I think it will work in Olympia branch in a couple of days.

query-node/schemas/council.graphql
ondratra: Ok, thx for the hint. The property is called `noteMetadata`, raw `note` was left intact. Done.

jodator:
> @jodator Candidacy metadata has been added. I've put the new output schema into the gist mentioned at the top of this thread, so you don't have to generate it on your own. At this point, I think it will be better to wait for testnet instead of trying to debug the build, as you suggested.

Thank you, @ondratra  :) Fortunately, I can generate a query-node server using `build.sh` script.

The setup worked for me at some point, so I think that after all of the merges into Olympia branch I should be able to run it.

ps.: I saw that you left a `note` on the `Candidacy`. I think that this could be removed as it doubles the info from `noteMedata`.

jodator:
@ondratra we also have a tracking issue with some additional issues with the schema: https://github.com/Joystream/pioneer/issues/1536. Most of them are non-blocking but some improvements are welcome.

Lezek123 (changes requested):
The integration tests are failing, the processor logs are showing:
```
ERROR: Stopping the proccessor due to errors: {} name: Error, message: Unexpected council stage "CouncilStageAnnouncing", stack: Error: Unexpected council stage "CouncilStageAnnouncing"
    at council_NewCandidate (/joystream/query-node/mappings/lib/council.js:225:15)
    at async MappingsLookupService.call (/joystream/node_modules/@dzlzv/hydra-processor/lib/executor/MappingsLookupService.js:80:9)
    at async /joystream/node_modules/@dzlzv/hydra-processor/lib/executor/TransactionalExecutor.js:35:17
```
Source: https://github.com/Joystream/joystream/runs/3869128272?check_suite_focus=true

ondratra:
Tests have been fixed. I think we should merge this PR and address additional change requests from Pioneer team that @jodator mentioned in separate PR.

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  }
+
+  return councilMember
+}
+
+/*
+  Returns the current election round record.
+*/
+async function getCurrentElectionRound(store: DatabaseManager): Promise<ElectionRound> {
+  const electionRound = await store.get(ElectionRound, { order: { id: 'DESC' } })
```
Lezek123: I think `id` will be arbitrary for election rounds (it's not explicitly set), you should order by `cycleId: 'DESC'`

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  const initialElectionRound = new ElectionRound({
+    cycleId: 0,
+    isFinished: false,
+    castVotes: [],
+    electedCouncil,
+    candidates: [],
+  })
+  await store.save<ElectionRound>(initialElectionRound)
+
+  return initialElectionRound
```
Lezek123: Should probably be moved to `/query-node/mappings/genesis.ts`

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+}
+
+/*
+  Returns the last council stage update.
+*/
+async function getCurrentStageUpdate(store: DatabaseManager): Promise<CouncilStageUpdate> {
+  const stageUpdate = await store.get(CouncilStageUpdate, { order: { id: 'DESC' } })
```
Lezek123: Avoid ordering by `id`, which is not explicitly set

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+}
+
+/*
+  Returns current elected council record.
+*/
+async function getCurrentElectedCouncil(
+  store: DatabaseManager,
+  canFail: boolean = false
+): Promise<ElectedCouncil | undefined> {
+  const electedCouncil = await store.get(ElectedCouncil, { order: { id: 'DESC' } })
```
Lezek123: Avoid ordering by `id`, which is not explicitly set

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  store: DatabaseManager,
+  account: string,
+  electionRound?: ElectionRound
+): Promise<CastVote> {
+  const where = { castBy: account } as FindConditions<Candidate>
+  if (electionRound) {
+    where.electionRound = electionRound
+  }
+
+  const castVote = await store.get(CastVote, { where, order: { id: 'DESC' } })
```
Lezek123: Avoid ordering by `id`, which is not explicitly set

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  return councilStage.isTypeOf == 'CouncilStageAnnouncing'
+}
+
+/*
+  Custom typeguard for council stage - election.
+* /
+function isCouncilStageElection(councilStage: typeof CouncilStage): councilStage is CouncilStageElection {
+  return councilStage.isTypeOf == 'CouncilStageElection'
+}
+*/
```
Lezek123: Remove commented-out code

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+
+  // update council record
+  const electedCouncil = await getCurrentElectedCouncil(store, true)
+  if (!electedCouncil) {
+    return
+  }
+
+  // electedCouncil.updates.push(councilStageUpdate) // uncomment after solving https://github.com/Joystream/hydra/issues/462
+  electedCouncil.updates = (electedCouncil.updates || []).concat([councilStageUpdate])
+  await store.save<ElectedCouncil>(electedCouncil)
```
Lezek123: 1. It's better to update the owning side of the relation (`CouncilStageUpdate.electedCouncil`) instead of inverse side (`ElectedCouncil.updates`)
2. You're not fetching the `updates` relation in `getCurrentElectedCouncil`, so `electedCouncil.updates` is always empty

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+    numOfCandidates,
+  })
+
+  await store.save<VotingPeriodStartedEvent>(votingPeriodStartedEvent)
+
+  // specific event processing
+
+  // add stage update record
+  const stage = new CouncilStageElection()
+  stage.candidatesCount = new BN(0)
```
Lezek123: Should be set to `numOfCandidates`?

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  })
+
+  await store.save<CandidacyWithdrawEvent>(candidacyWithdrawEvent)
+
+  // specific event processing
+
+  // mark candidacy as withdrawn
+  const electionRound = await getCurrentElectionRound(store)
+  const candidate = await getCandidate(store, memberId.toString(), electionRound)
+  candidate.candidacyWithdrawn = false
```
Lezek123:
```suggestion
  candidate.candidacyWithdrawn = true
```

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  const electionRound = await getCurrentElectionRound(store)
+  const candidate = await getCandidate(store, memberId.toString(), electionRound)
+
+  // unpack note's metadata and save it to db
+  const metadata = deserializeMetadata(CouncilCandidacyNoteMetadata, note)
+  const noteMetadata = candidate.noteMetadata
+  noteMetadata.header = metadata?.header || undefined
+  noteMetadata.bulletPoints = metadata?.bulletPoints || []
+  noteMetadata.bannerImageUri = metadata?.bannerImageUri || undefined
+  noteMetadata.description = metadata?.description || undefined
```
Lezek123: When you set field to `undefined` it's actually not getting updated at all (which could be a good thing, because we could support partial/atomic updates this way), you need to set it to `(null as any)` in order to unset it (see: https://github.com/Joystream/hydra/issues/435).

I would suggest something like:
```
import { isSet } from '@joystream/metadata-protobuf/utils'

noteMetadata.header = isSet(metadata?.header) ? (metadata.header || (null as any)) : undefined

// OR (for better readability)

noteMetadata.header = isSet(metadata?.header) ? (metadata.header || (null as any)) : noteMetadata.header
```

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  The event is emitted when a new budget refill is planned.
+*/
+export async function council_BudgetRefillPlanned({ event, store }: EventContext & StoreContext): Promise<void> {
+  // common event processing
+
+  const [] = new Council.BudgetRefillPlannedEvent(event).params
+
+  const budgetRefillPlannedEvent = new BudgetRefillPlannedEvent({
+    ...genericEventFields(event),
+  })
```
Lezek123: Add event data? (`next_refill`)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+
+  await store.save<RewardPaymentEvent>(rewardPaymentEvent)
+
+  // specific event processing
+
+  // update (un)paid reward info
+  const councilMember = await getCouncilMember(store, memberId.toString())
+  councilMember.accumulatedReward = councilMember.accumulatedReward.add(paidBalance)
+  councilMember.unpaidReward = missingBalance
+  await store.save<CouncilMember>(councilMember)
```
Lezek123: `lastPaymentBlock` should also be updated

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+    where: { id: optionResultsRaw.map((item) => item.option_id.toString()) },
+  })
+
+  const referendumFinishedEvent = new ReferendumFinishedEvent({
+    ...genericEventFields(event),
+    optionResults: optionResultsRaw.map(
+      (item, index) =>
+        new ReferendumStageRevealingOptionResult({
+          votePower: item.vote_power,
+          optionId: members[index],
```
Lezek123: Members array is not sorted in any way, so `members[index]` may point to an invalid option.
Also `optionId` as field name is a little confusing when querying (as usually you would expect `ID` fields to end with `Id`, like `memberId` or `proposalId`)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+
+/*
+  The event is emitted when a vote is casted in the council election.
+*/
+export async function referendum_VoteCast({ event, store }: EventContext & StoreContext): Promise<void> {
+  // common event processing
+
+  const [account, hash, stake] = new Referendum.VoteCastEvent(event).params
+  const votePower = calculateVotePower(account.toString(), stake)
+  const hashString = bytesToString(hash)
```
Lezek123: It would be better to store the hex instead (especially because the hash may contain `0x00`)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  // common event processing
+
+  const [account, memberId, salt] = new Referendum.VoteRevealedEvent(event).params
+  const member = await getMembership(store, memberId.toString())
+
+  const voteRevealedEvent = new VoteRevealedEvent({
+    ...genericEventFields(event),
+    account: account.toString(),
+    member: member,
+    salt: bytesToString(salt),
```
Lezek123: It would be better to store the hex instead (especially because the salt may contain `0x00`)

query-node/run-tests.sh
```diff
@@ -23,7 +23,7 @@ trap cleanup EXIT
 # Clean start
 docker-compose down -v
 
-./start.sh
+QUERY_NODE_TESTS=1 ./start.sh
```
Lezek123:
```suggestion
./start.sh
```

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+################### Council ####################################################
+
+type CouncilStageUpdate @entity {
+  "The new stage council got into."
+  stage: CouncilStage!
+
+  "Block number at which change happened."
+  changedAt: BigInt!
+
+  "Council elected in this stage update if any."
```
Lezek123:
```suggestion
  "Council term during which the update happened (if any)."
```

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+}
+
+type ElectionProblemNewCouncilNotElected @variant {
+  # no properties
+
+  # TODO: remove me - variant needs to have at least 1 property now
+  dummy: Int
+}
+
+union ElectionProblem = ElectionProblemNotEnoughCandidates | ElectionProblemNewCouncilNotElected | VariantNone
```
Lezek123: When all variants are "empty" you can just use `enum` instead of `union`

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+  "Intermediate winning options"
+  intermediateWinners: [ReferendumStageRevealingOptionResult!]!
+
+  "Index of current election"
+  electionRound: ElectionRound!
+}
+
+type ReferendumStageRevealingOptionResult @entity {
+  "Member that received votes."
+  optionId: Membership!
```
Lezek123:
```suggestion
  option: Membership!
```

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+
+  "Election round."
+  electionRound: ElectionRound!
+
+  "Event that concluded the referendum."
+  referendumFinishedEvent: ReferendumFinishedEvent!
+}
+
+# TODO: this maybe needs to be @entity - so it's saved in db
+union ReferendumStage = ReferendumStageInactive | ReferendumStageVoting | ReferendumStageRevealing | VariantNone
```
Lezek123: This is currently never used

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+  stake: BigInt!
+
+  "Reflects if the stake is still locked for candidacy or has been already released by the member."
+  stakeLocked: Boolean!
+
+  "Account that cast the vote."
+  castBy: String!
+
+  "Member receiving the vote."
+  voteFor: Membership
```
Lezek123: This value is never set

query-node/mappings/council.ts
```diff
@@ -0,0 +1,907 @@
...
+  const electionRound = await getCurrentElectionRound(store)
+
+  const candidates = await store.getMany(Candidate, { where: { electionRoundId: electionRound.id } })
+
+  // create new council record
+  const electedCouncil = new ElectedCouncil({
+    councilMembers: await convertCandidatesToCouncilMembers(store, candidates, event.blockNumber),
+    updates: [],
+    electedAtBlock: event.blockNumber,
+    endedAtBlock: event.blockNumber,
```
Lezek123: This setting is invalid

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,281 @@
...
+  councilMembers: [CouncilMember!]! @derivedFrom(field: "electedInCouncil")
+
+  "Changes to council status that were made during it's reign."
+  updates: [CouncilStageUpdate!]! @derivedFrom(field: "electedCouncil")
+
+  "Block number at which the council was elected."
+  electedAtBlock: Int!
+
+  "Block number at which the council reign ended and a new council was elected."
+  endedAtBlock: Int
```
Lezek123, line 192: This is currently not correctly set

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  member: Membership!
+
+  "Candidate's account used to stake currency."
+  stakingAccount: String!
+
+  "Candidate's account that will be recieving rewards if candidate's elected."
+  rewardAccount: String!
+
+  "Amount of currency to be staked for the candidacy."
+  balance: BigInt!
```
Lezek123: Just referencing `Candidate` entity should work

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  "Network the block was produced in"
+  network: Network!
+
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
+
+  "Candidate's membership."
+  member: Membership!
```
Lezek123: `Candidate` would be a better reference

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
```
Lezek123: `Candidate` would be a better reference

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
```
Lezek123: `Candidate` would be a better reference

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+
+  "The metadata contained in note."
+  noteMetadata: CandidacyNoteMetadata!
```
Lezek123, line 219: The metadata entity is updateable (it's not a snapshot), so currently all updates (per candidate) will point to the same `CandidacyNoteMetadata`

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  "Network the block was produced in"
+  network: Network!
+
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
+
+  "Candidate's membership."
+  member: Membership!
```
Lezek123: `Candidate` would be a better reference

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  "Blocknumber of the block in which the event was emitted."
+  inBlock: Int!
+
+  "Network the block was produced in"
+  network: Network!
+
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
```
Lezek123, line 321: Add `next_refill` data (block number)

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+
+  "Funding account."
```
Lezek123:
```suggestion
  "Target account."
```

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  ### SPECIFIC DATA ###
+
+  "Account that cast the vote."
+  account: String!
+
+  "Hash of sealed vote."
+  hash: String!
+
+  "Vote power decided by amount staked."
+  votePower: BigInt!
```
Lezek123: Reference to `CastVote` should do

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  ### SPECIFIC DATA ###
+
+  "Account that cast a vote."
+  account: String!
+
+  "Membership of user that cast the vote."
+  member: Membership!
+
+  "Salt that has been used for the vote's hash creation."
+  salt: String!
```
Lezek123: Reference to `CastVote` should do

query-node/schemas/membership.graphql
```diff
@@ -71,6 +71,9 @@ type Membership @entity {
   "Whether member is founding member."
   isFoundingMember: Boolean!
 
+  "Whether member is elected in the current council."
+  isCouncilMember: Boolean!
```
Lezek123, line 75: This value is never set to `true`

query-node/schemas/councilEvents.graphql
```diff
@@ -0,0 +1,576 @@
...
+  "Network the block was produced in"
+  network: Network!
+
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
+
+  "Candidates' memberships."
+  electedMembers: [Membership!]
```
Lezek123: Reference to `ElectedCouncil` should do

Lezek123 (changes requested):

query-node/mappings/council.ts
ondratra: `cycleId` has been used here as suggested. On similar places, `createdAt` was used instead when no better option was available.

query-node/mappings/council.ts
ondratra: Moved to `genesis.ts`.

query-node/mappings/council.ts
ondratra: `createdAt` has been used as ordering parameter.

query-node/mappings/council.ts
ondratra: `createdAt` has been used as ordering parameter.

query-node/mappings/council.ts
ondratra: Done.

query-node/mappings/council.ts
ondratra: You're right. This whole section was completely unnecessary, and an update on the other side of the relationship was the proper fix.

query-node/mappings/council.ts
ondratra: Yes, the `numOfCandidates` is supposed to be there. Done.

query-node/mappings/council.ts
ondratra: Done.

query-node/mappings/council.ts
ondratra: `next_refill` info added.

query-node/mappings/council.ts
ondratra: Done.

query-node/mappings/council.ts
ondratra: I didn't realize the results are not sorted. Fixed.

query-node/mappings/council.ts
ondratra: Done.

query-node/mappings/council.ts
ondratra: Done.

query-node/schemas/council.graphql
ondratra: Done.

query-node/schemas/council.graphql
ondratra: Changed to `option`.

query-node/schemas/council.graphql
ondratra: Done.

query-node/mappings/council.ts
ondratra: `endedAtBlock` is not set now.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: `nextRefillInBlock` info added.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/schemas/membership.graphql
ondratra: The sign is being set and unset properly now.

query-node/schemas/councilEvents.graphql
ondratra: Done.

query-node/mappings/council.ts
ondratra: createdAt has been used as ordering parameter.

query-node/schemas/council.graphql
ondratra: Not used items have been commented out until we solve variant references and uncomment all relevant references.

query-node/schemas/council.graphql
ondratra: Properly set now.

query-node/schemas/councilEvents.graphql
ondratra: Permanent record created there.

ondratra:
Issues from the last review are all addressed. @jodator, there were some changes to the schema. Most of them were minor, but check it all still works for you.

Tests are failing due to some error when building `joystream-node`. This should not be caused by changes in this PR. Tests passed on my local machine.

jodator:
@ondratra Sorry for the long response. I had issues running the query-node.

So far I found:

1.  candidates from the failed announcing period are not removed/hidden
2.  For past councils, we would like to display: time, block, and network when the council ended. Usually, we take that from some event, but I can't find something for the end time. There's only `newCouncilElectedEvents`.

Ad. 1 Failed announcing period

This happens if the announcing period ends up and there was only one candidate. Because the election cycle on the test node is around 1 min I was creating candidates every minute.

Now, since both were created at the different announcing periods they are returned by the query-node as candidates for the "current" election. I think that candidates should be tied to the announcing round.

Ad. 2. This is not critical, but a nice to have:

ondratra:
@jodator The changes you requested are done. Each council has block, time, and network at the time of election/resignation associated. Failed elections start a proper next election round in query node. Also, integration tests for failed elections were added.

query-node/mappings/genesis.ts
```diff
@@ -30,6 +41,27 @@ export async function loadGenesisData({ store }: StoreContext): Promise<void> {
       )
     )
   )
+}
 
-  // TODO: members, workers
+async function initFirstElectionRound(store: DatabaseManager) {
+  const electedCouncil = new ElectedCouncil({
```
Lezek123: Why do we need an empty `ElectedCouncil` entity at genesis? Is it just to reference it in `ElectionRound`? I think it would make sense to make the reference to `ElectedCouncil` in `ElectionRound` optional instead (because during first election ever there will be no elected council at all, I think)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+    balance,
+  })
+
+  await store.save<NewCandidateEvent>(newCandidateEvent)
+}
+
+/*
+  The event is emitted when the new council is elected. Sufficient members were elected and there is no other problem.
+*/
+export async function council_NewCouncilElected({ event, store }: EventContext & StoreContext): Promise<void> {
```
Lezek123: This should actually create `CouncilStageUpdate` to `CouncilStageIdle` instead of calling `startNextElectionRound` and progressing to `CouncilStageAnnouncing` immediately.

See: https://github.com/Joystream/joystream/blob/olympia/runtime-modules/council/src/lib.rs#L1284

Progressing to `Announcing` stage should only happen on `council_AnnouncingPeriodStarted` and `council_NotEnoughCandidates` (actually ideally during `NotEnoughCandidates` scenario `AnnouncingPeriodStarted` would also be emitted, but this would require a change in the runtime, as it is not currently the case)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+}
+
+/*
+  The event is emitted when a previously casted vote is revealed.
+*/
+export async function referendum_VoteRevealed({ event, store }: EventContext & StoreContext): Promise<void> {
+  // common event processing - init
+
+  const [account, memberId, salt] = new Referendum.VoteRevealedEvent(event).params
+  const member = await getMembership(store, memberId.toString())
```
Lezek123: Unused variable

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  // load candidate recored
+  const electionRound = await getCurrentElectionRound(store)
+  const candidate = await getCandidate(store, memberId.toString(), electionRound)
+
+  // unpack note's metadata and save it to db
+  const metadata = deserializeMetadata(CouncilCandidacyNoteMetadata, note)
+  const noteMetadata = candidate.noteMetadata
+  // `XXX || (null as any)` construct clears metadata if requested (see https://github.com/Joystream/hydra/issues/435)
+  noteMetadata.header = isSet(metadata?.header) ? metadata?.header || (null as any) : metadata?.header
```
Lezek123: I suggest using either:
```
  noteMetadata.header = isSet(metadata?.header) ? metadata?.header || (null as any) : undefined
```
or:
```
  noteMetadata.header = isSet(metadata?.header) ? metadata?.header || (null as any) : noteMetadata.header
```

If `isSet(metadata?.header)` is `false`, the `metadata?.header` is technically still `null | undefined` (which have different meanings in `TypeORM`).
Typescript only allows this assignment, because if we use:
```
const x = isSet(metadata?.header) ? metadata?.header || (null as any) : metadata?.header
```
Then `x` is of type `any`

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+    where: { id: optionResultsRaw.map((item) => item.option_id.toString()) },
+  })
+
+  const referendumFinishedEvent = new ReferendumFinishedEvent({
+    ...genericEventFields(event),
+    optionResults: optionResultsRaw.map(
+      (item, index) =>
+        new ReferendumStageRevealingOptionResult({
+          votePower: item.vote_power,
+          option: members.find((tmpMember) => tmpMember.id.toString() == optionResultsRaw[index].option_id.toString()),
```
Lezek123: I think this is overcomplicated. You could just use:
```
option: new Membership({ id: optionResultsRaw[index].option_id.toString() })
```
It will throw during query execution if the member does not exist, but we expect an error beeing thrown in that case anyway.

query-node/schemas/council.graphql
```diff
@@ -0,0 +1,292 @@
...
+  "Election cycle ID."
+  cycleId: Int!
+
+  "Sign if election has already finished."
+  isFinished: Boolean!
+
+  "Vote cast in the election round."
+  castVotes: [CastVote!]! @derivedFrom(field: "electionRound")
+
+  # TODO: reference variant (how?)
```
Lezek123: You could have `referrendumStage: ReferendumStage` (an union), you just can't use `@derivedFrom` and have to set it explicitly in the mappings in that case. Is this an issue?

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+  store: DatabaseManager,
+  memberId: string,
+  electionRound?: ElectionRound
+): Promise<Candidate> {
+  const where = { memberId: memberId } as FindConditions<Candidate>
+  if (electionRound) {
+    where.electionRound = electionRound
+  }
+
+  const candidate = await store.get(Candidate, { where, order: { createdAt: 'DESC' } })
```
Lezek123: Would it be possible to:
`announce_candidacy -> withdraw_candidacy -> announce_candidacy` within the same block?
It's a very edge case, but may still cause `{ createdAt: 'DESC' }` to not always return the last candidate (if we assume `createdAt` is set to the block timestamp).
If the runtime was providing `candidateId` or we were able to use `INT AUTO_INCREMENT` field in the database as id, that would be the best choice for `orderBy`, but since this is not available I think we need some other workaround.

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  return stageUpdate
+}
+
+/*
+  Returns current elected council record.
+*/
+async function getCurrentElectedCouncil(
+  store: DatabaseManager,
+  canFail: boolean = false
```
Lezek123: From the typesafety perspective it would be better to just have 2 functions like:
`getCurrentElectedCouncil` and `getRequiredCurrentElectedCouncil`, where one throws and the other doesn't (and returns `ElectedCouncil | undefined`)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+    oldElectedCouncil.endedAtTime = new Date(event.blockTimestamp)
+    oldElectedCouncil.endedAtNetwork = CURRENT_NETWORK
+    await store.save<ElectedCouncil>(oldElectedCouncil)
+  }
+
+  // get election round and its candidates
+  const electionRound = await getCurrentElectionRound(store)
+
+  // TODO: uncomment when following query will be working (after some QN patches make it to Olympia)
+  ////const electedCandidates = await store.getMany(Candidate, { where: { electionRoundId: electionRound.id, member: { id_in: electedMemberIds } } })
```
Lezek123: I'm not 100% sure if this will work, but did you try:
```typescript
import { In } from 'typeorm'
/* ... */
  store.getMany(Candidate, {
    where: { member: { id: In(memberIds.map((id) => id.toString())) } },
    relations: ['member'],
  })
/* ... */
```

tests/integration-tests/src/fixtures/council/common.ts
```diff
@@ -0,0 +1,19 @@
...
+export async function assertCouncilMembersRuntimeQnMatch(api: Api, query: QueryNodeApi) {
+  const runtimeCouncilMembers = await api.query.council.councilMembers()
+  runtimeCouncilMembers.map((item) => console.log(item.toString()))
+
+  const tmp = await query.tryQueryWithTimeout(
+    () => query.getCurrentCouncilMembers(),
+    (qnElectedCouncil) => {
+      console.log(qnElectedCouncil, qnElectedCouncil?.councilMembers)
+      assert.sameMembers(
+        (qnElectedCouncil?.councilMembers || []).map((item: any) => item.member.id.toString()),
```
Lezek123, line 24: Please don't use `item: any`, it defeats the typesafety. Just using `.map((item) => /* ... */)` will make TS derive the type of `item`.

tests/integration-tests/src/fixtures/council/common.ts
```diff
@@ -0,0 +1,19 @@
+import { assert } from 'chai'
+import { Api } from '../../Api'
+import { QueryNodeApi } from '../../QueryNodeApi'
+
+export async function assertCouncilMembersRuntimeQnMatch(api: Api, query: QueryNodeApi) {
+  const runtimeCouncilMembers = await api.query.council.councilMembers()
+  runtimeCouncilMembers.map((item) => console.log(item.toString()))
+
+  const tmp = await query.tryQueryWithTimeout(
```
Lezek123: Unused variable

tests/integration-tests/src/fixtures/council/FailToElectCouncilFixture.ts
```diff
@@ -0,0 +1,214 @@
...
+  candidatesMemberAccounts: string[]
+  councilCandidateStake: Balance
+  councilMemberIds: MemberId[]
+}
+
+export class FailToElectCouncilFixture extends BaseQueryNodeFixture {
+  /*
+    Execute all scenarios when a new council election is expected to fail.
+  */
+  public async execute(): Promise<void> {
```
Lezek123: I think a fixture is supposed to be more atomic/reusable piece of code, this sounds more like a flow/scenario.

tests/integration-tests/src/scenarios/council.ts
```diff
@@ -0,0 +1,10 @@
+import electCouncil from '../flows/council/elect'
+import failToElectCouncil from '../flows/council/failToElect'
+import { scenario } from '../Scenario'
+
+scenario(async ({ job }) => {
+  const councilJob = job('electing council', electCouncil)
+  const secondCouncilJob = job('electing second council', electCouncil).requires(councilJob)
```
Lezek123, line 7: Why is the `electCouncil` flow ran twice?

tests/integration-tests/src/scenarios/full.ts
```diff
@@ -75,4 +76,8 @@ scenario(async ({ job, env }) => {
   job('forum polls', polls).requires(sudoHireLead)
   job('forum posts', posts).requires(sudoHireLead)
   job('forum moderation', moderation).requires(sudoHireLead)
+
+  // Council
+  const secondCouncilJob = job('electing second council', electCouncil).requires(membershipSystemJob)
```
Lezek123, line 81: Why is the `electCouncil` flow ran twice?

Lezek123 (changes requested):

Lezek123:
I forgot to include this in the review:

I'm not sure why the query-node linter checks are not part of the CI in olympia (we should probably fix it in another PR), but my eslint plugin for vscode is showing quite a lot of valid linter errors/warnings in the council mappings (they can also be retrieved with `yarn workspace query-node-mappings eslint . --ext .ts --ignore-pattern lib --ignore-pattern generated`), for example:
```
   58:1   error    Expected space or tab after '///' in comment                                                               spaced-comment
  143:3   error    Type boolean trivially inferred from a boolean literal, remove type annotation                             @typescript-eslint/no-inferrable-types
  191:32  error    Expected '===' and instead saw '=='                                                                        eqeqeq
...
```

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+    ? metadata?.bannerImageUri || (null as any)
+    : metadata?.bannerImageUri
+  noteMetadata.description = isSet(metadata?.description)
+    ? metadata?.description || (null as any)
+    : metadata?.description
+  await store.save<CandidacyNoteMetadata>(noteMetadata)
+
+  // save metadata set by this event
+  const noteMetadataSnapshot = new CandidacyNoteMetadata({
+    header: metadata?.header || undefined,
```
Lezek123: `header === ''` and `header === undefined` will have different meaning (the first one will unset the header, while the second one will not change it), but if you use `header || undefined`, it will result in `undefined` in both these cases.
You should probably use:
```
    header: metadata?.header ?? undefined,
```
(same goes for `description` etc.)

query-node/mappings/council.ts
```diff
@@ -0,0 +1,928 @@
...
+  // load candidate recored
+  const electionRound = await getCurrentElectionRound(store)
+  const candidate = await getCandidate(store, memberId.toString(), electionRound)
+
+  // unpack note's metadata and save it to db
+  const metadata = deserializeMetadata(CouncilCandidacyNoteMetadata, note)
+  const noteMetadata = candidate.noteMetadata
+  // `XXX || (null as any)` construct clears metadata if requested (see https://github.com/Joystream/hydra/issues/435)
+  noteMetadata.header = isSet(metadata?.header) ? metadata?.header || (null as any) : metadata?.header
+  noteMetadata.bulletPoints = metadata?.bulletPoints || []
```
Lezek123: Wouldn't it reset `bulletPoints` even if they are not provided in the metadata? (You could add a test case for that)
For arrays I think you need custom logic, because `[]` is treated same way as `undefined` in the protobuf. You could use `['']` (array with empty string) as "unset all" signal and `undefined | null | []` as "do not change" signal, see: https://github.com/Joystream/joystream/blob/giza/query-node/mappings/storage/metadata.ts#L144

ondratra: Done. `['']` has been used to signal deletion.

query-node/mappings/council.ts
ondratra: That's right. Idle stage follows successful election. Fixed.

query-node/mappings/council.ts
ondratra: It was removed with bunch of other unused variables.

query-node/mappings/council.ts
ondratra: Fixed (second variant used).

query-node/mappings/council.ts
ondratra: You're right, that is working well.

query-node/mappings/council.ts
ondratra: Nice catch. It can be easily fixed by adding ordering by `candidacyWithdrawn`(bool) in addition to `createdAt`. Done.

tests/integration-tests/src/fixtures/council/common.ts
ondratra: Fixed

ondratra: Fixed

tests/integration-tests/src/fixtures/council/common.ts
ondratra: Fixed

tests/integration-tests/src/fixtures/council/FailToElectCouncilFixture.ts
ondratra: The fixture has been split into two, and the flow calls them in a sequence.

query-node/mappings/council.ts
ondratra: Fixed.

query-node/mappings/council.ts
ondratra: Loaded relations, unfortunately, don't have an effect on this. It's an error `"column worker.undefined does not exist"`  that we have encountered a couple of times lately, and it is not (fully?) integrated to Hydra yet. https://github.com/Joystream/hydra/issues/444

query-node/mappings/src/council.ts
```diff
@@ -0,0 +1,938 @@
...
+  The event is emitted when the candidate changes its candidacy note.
+*/
+export async function council_CandidacyNoteSet({ event, store }: EventContext & StoreContext): Promise<void> {
+  // common event processing
+
+  const [memberId, note] = new Council.CandidacyNoteSetEvent(event).params
+
+  // load candidate recored
+  const electionRound = await getCurrentElectionRound(store)
+  const candidate = await getCandidate(store, memberId.toString(), electionRound)
```
Lezek123: `noteMetadata` is not eagerly loaded, you need to pass: `relations: ['noteMetadata']` as part of `find` parameters.

query-node/schemas/council.graphql
```diff
@@ -229,14 +229,12 @@ type ElectedCouncil @entity {
   "Network running at the time of resignation."
   endedAtNetwork: Network
 
-  # it might seem that derived field is wrongly set to `nextElectedCouncil`, but that's how it should be
+  # it might seems that derived field is wrongly set to `nextElectedCouncil`, but that's how it should be
```
Lezek123: I think this was correct before

tests/integration-tests/src/fixtures/council/common.ts
```diff
@@ -1,19 +1,67 @@
...
-      console.log(qnElectedCouncil, qnElectedCouncil?.councilMembers)
       assert.sameMembers(
         (qnElectedCouncil?.councilMembers || []).map((item: any) => item.member.id.toString()),
         runtimeCouncilMembers.map((item: any) => item.membership_id.toString())
       )
     }
   )
 }
+
+export async function prepareFailtToElectResources(api: Api, query: QueryNodeApi): Promise<IFailToElectResources> {
```
Lezek123:
```suggestion
export async function prepareFailToElectResources(api: Api, query: QueryNodeApi): Promise<IFailToElectResources> {
```

query-node/mappings/src/council.ts
```diff
@@ -0,0 +1,938 @@
...
+  store: DatabaseManager,
+  memberId: string,
+  electionRound?: ElectionRound
+): Promise<Candidate> {
+  const where = { memberId: memberId } as FindConditions<Candidate>
+  if (electionRound) {
+    where.electionRound = electionRound
+  }
+
+  const candidate = await store.get(Candidate, { where, order: { createdAt: 'DESC', candidacyWithdrawn: 'ASC' } })
```
Lezek123: This solves the issue partially, but candidacy can still be withdrawn and posted multiple times within the same block. It's a very improbable case (the user would probably have to act maliciously), but I think it's worth securing against. I created a [PR](https://github.com/ondratra/joystream/pull/6) to your branch with a proposed solution

Lezek123 (changes requested):
Good job addressing the issues!
I tested the mappings locally and created a PR to your branch with a few fixes for issues that could potentially break the query node: https://github.com/ondratra/joystream/pull/6

tests/integration-tests/src/fixtures/council/common.ts
ondratra: Fixed.

query-node/mappings/src/council.ts
ondratra: Changes from your PR has been accepted.

query-node/schemas/council.graphql
ondratra: Fixed.

query-node/mappings/src/council.ts
ondratra: Relations loading added for `noteMetadata` as well as other relevant relations needed in other places.

query-node/schemas/council.graphql
ondratra: Due to complications with variants in Hydra (for example, https://github.com/Joystream/hydra/issues/467), referendum stages were replaced by entities. Thus, `@derivedFrom` can be used with ease now.

query-node/mappings/genesis.ts
ondratra: There were actually two reasons:
1) I was not sure if there will be no council before the first election is complete or we will appoint a temporary council with preselected members (in that case, we could only add member ids into `genesis.ts`)
2) To remove checks for council existence in council mappings as well as in applications using QN

That being said, there is an inconsistency because there are a couple of places in `council.ts` that actually check that council record exists (as can be seen in your comment about `getCurrentElectedCouncil`/`getRequiredCurrentElectedCouncil`). 

I am for keeping the initial council in `genesis.ts` and removing all checks for elected council emptiness through `council.ts`. Do you think that's ok?

query-node/mappings/council.ts
ondratra: Done. This is related to `initFirstElectionRound` discussion above.

query-node/mappings/genesis.ts
ondratra: After private discussion we've decided to keep initial council insertion in `genesis.ts`.

ondratra:
Issues from previous reviews were addressed. QN code lint was added to CI.

Note: to get intermediate winners in the current election, use this query:
```
candidates(
  where: { electionRound: { cycleId_eq: $electionRoundCycleId }, votePower_gt: 0 }
  orderBy: [votePower_DESC, lastVoteReceivedAtBlock_ASC, lastVoteReceivedAtEventNumber_ASC]
  limit: $councilSize
) {
  id
  member {
    id
  }
  # ...
}
```
Removing `limit` enables inspection of whole election results.

Lezek123 (approved):
LGTM!

Issue #2703, opened 2021-09-23
Giza: Query node mappings - assign dynamic bag to buckets on creation


Lezek123:
Currently dynamic bags are not assigned to buckets on `DynamicBagCreated` event processing, because the `DynamicBagCreated` event lacks data about assigned buckets. This needs to be fixed both on the runtime and mappings side.

The issue only estimates the time on the mappings side.

PR #2706, opened 2021-09-24
runtime: storage: Update DynamicBagCreated event.

shamil-gadelshin:
- add storage and distribution bucket IDs

Relates to https://github.com/Joystream/joystream/issues/2703


Lezek123 (approved):
Works as expected!

PR #2710, opened 2021-09-27
Merge `new working groups` into the giza branch.

shamil-gadelshin:
### Changes
- merge `new-operations-wg` branch (with new code from `master`)
- fixed incorrect tests for working groups 
- new distribution working group event handler
- updated pallets' versions
- removed DaoId and Dao concepts
- removed commented code

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -127,7 +127,7 @@ pub enum ChannelOwner<MemberId, CuratorGroupId> {
 // simplification type
 pub(crate) type ActorToChannelOwnerResult<T> = Result<
     ChannelOwner<
-        <T as membership::Trait>::MemberId,
+        <T as common::MembershipTypes>::MemberId,
         <T as ContentActorAuthenticator>::CuratorGroupId,
     >,
     Error<T>,
@@ -185,7 +185,7 @@ pub struct ChannelRecord<MemberId, CuratorGroupId, AccountId> {
 
 // Channel alias type for simplification.
 pub type Channel<T> = ChannelRecord<
-    <T as membership::Trait>::MemberId,
+    <T as common::MembershipTypes>::MemberId,
     <T as ContentActorAuthenticator>::CuratorGroupId,
     <T as frame_system::Trait>::AccountId,
 >;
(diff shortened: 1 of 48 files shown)


runtime-modules/content/src/lib.rs
```diff
@@ -185,7 +185,7 @@ pub struct ChannelRecord<MemberId, CuratorGroupId, AccountId> {
 
 // Channel alias type for simplification.
 pub type Channel<T> = ChannelRecord<
-    <T as membership::Trait>::MemberId,
+    <T as common::MembershipTypes>::MemberId,
```
dobertRowneySr, line 188: Maybe I'm not getting it. Why are we moving from the membership to the common pallet?

runtime-modules/content/src/lib.rs
```diff
@@ -1477,7 +1470,7 @@ decl_event!(
         ContentActor = ContentActor<
             <T as ContentActorAuthenticator>::CuratorGroupId,
             <T as ContentActorAuthenticator>::CuratorId,
-            <T as membership::Trait>::MemberId,
+            <T as common::MembershipTypes>::MemberId,
```
dobertRowneySr, line 1473: Maybe I'm not getting it. Shouldn't we use exclusively MemberId out of the membership pallet.
Since it is also used in the storage module+

dobertRowneySr, line 1473: Maybe I'm not getting it. Why are we moving from the membership to the common pallet?

runtime-modules/content/src/tests/mock.rs
```diff
@@ -153,10 +153,8 @@ parameter_types! {
 
 impl membership::Trait for Test {
     type Event = MetaEvent;
```
dobertRowneySr, line 155: same as above

runtime/src/integration/proposals/membership_origin_validator.rs
```diff
@@ -6,7 +6,7 @@ use common::origin::ActorOriginValidator;
 use frame_system::ensure_signed;
 
 /// Member of the Joystream organization
-pub type MemberId<T> = <T as membership::Trait>::MemberId;
+pub type MemberId<T> = <T as common::MembershipTypes>::MemberId;
```
dobertRowneySr, line 9: same as above

dobertRowneySr (commented):
Looks fine to me. Just it is not clear why are we moving `MemberId` from the `membership` 
module to the `common` module

shamil-gadelshin:
The merge revealed `MemberId` and `ActorId` in both `memberhip::Trait` and `common::MembershipTypes`. It caused errors with typecasting. Since in the Olympia release we have a type similar to `common::MembershipTypes` - I appropriated a similar approach and deleted `MemberId` and `ActorId` in favor of `common::MembershipTypes`. In the Olympia release, we need this type separation to avoid circular references (for example, between membership and working group pallets for benchmarking). I'm not sure that we have such a problem here, but since we already have `common::MembershipTypes` - it looks natural to fix conflicts in favor of the separate common trait.

dobertRowneySr (approved):

Issue #2726, opened 2021-09-30
Minor fixes agreed upon for Giza


bedeho:
- Drop data object id association in video: `maybe_data_objects_id_set`
- Drop incrementing `ChannelRecord::num_videos` in `update_video`.
- introduce atomic data object deletion in: for all of these, user provides `objects: BTreeSet<DataObjectId>`, for all each, they will check that deletion is feasible (by checking bag membership of all objects), and then this is done as first side-effect after mutation barrier.
  - update_video
  - update_channel
  - delete_video
  - delete_channel

dobertRowneySr:
Maybe I didn't realized during the call: why do we need to need to feed the `delete_channel` extrinsic
with a `BTreeSet` containing all the data obj.s? Don't we need to delete the storage bag for a channel
upon channel deletion? @bedeho

bedeho:
> Maybe I didn't realized during the call: why do we need to need to feed the delete_channel extrinsic
with a BTreeSet containing all the data obj.s? Don't we need to delete the storage bag for a channel
upon channel deletion? @bedeho

We do need to delete the bag, but this means it must be empty, which means all objects must be deleted first, and there is no way (that I recall), of enumerating all objects in a bag, is there? I am also not 100% sure about whether a weight function can access storage to compute weight, which would be required in order to recover the counter of how many data objects live in the bag.  @shamil-gadelshin @iorveth

iorveth:
> > Maybe I didn't realized during the call: why do we need to need to feed the delete_channel extrinsic
> > with a BTreeSet containing all the data obj.s? Don't we need to delete the storage bag for a channel
> > upon channel deletion? @bedeho
> 
> We do need to delete the bag, but this means it must be empty, which means all objects must be deleted first, and there is no way (that I recall), of enumerating all objects in a bag, is there? I am also not 100% sure about whether a weight function can access storage to compute weight, which would be required in order to recover the counter of how many data objects live in the bag. @shamil-gadelshin @iorveth

Weight function cannot access storage directly, but we can provide witness if applicable, which would be used to calculate weight. Afterwards we check whether witness is correct during extrinsic execution. https://github.com/Joystream/joystream/issues/2541

bedeho:
Ok, so if the enumeration was possible, the only input required would the number of data objects in the channel bag, but alas, I believe it is not? I don't recall 100%

iorveth:
[...]

PR #2729, opened 2021-10-01
Giza bug fixed no urls

dobertRowneySr:
reference #2726

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -672,6 +672,7 @@ decl_module! {
             actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
             channel_id: T::ChannelId,
             params: ChannelUpdateParameters<T>,
+            assets: BTreeSet<<T as storage::Trait>::DataObjectId>,
         ) {
             // check that channel exists
             let channel = Self::ensure_channel_exists(&channel_id)?;
@@ -681,6 +682,8 @@ decl_module! {
                 &actor,
                 &channel.owner,
             )?;
+
+            let _num_assets_removed = Self::remove_assets_from_storage(&assets, &channel_id, &channel.deletion_prize_source_account_id);
 
             // atomically upload to storage and return the # of uploaded assets
             let _num_assets_uploaded = params.assets.as_ref().map_or(0u64,|assets| Self::upload_assets_to_storage(assets, &channel_id, &channel.deletion_prize_source_account_id));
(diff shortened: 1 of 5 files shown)


runtime-modules/content/src/lib.rs
```diff
@@ -1589,16 +1504,17 @@ decl_event!(
...
             ContentActor,
             PersonId,
-            NewAssets,
-            PersonUpdateParameters<NewAssets>,
+            StorageAssets,
+            PersonUpdateParameters<StorageAssets>,
         ),
         PersonDeleted(ContentActor, PersonId),
         ChannelDeleted(ContentActor, ChannelId),
+        //        VideoAssetsDeleted(ContentActor, VideoId, BTreeSet<AssetsToBeRemoved>),
```
shamil-gadelshin: Please, delete commented code.

runtime-modules/content/src/lib.rs
```diff
@@ -1469,6 +1353,37 @@ impl<T: Trait> Module<T> {
...
+        Ok(())
+    }
+
+    fn remove_assets_from_storage(
+        assets: &BTreeSet<<T as storage::Trait>::DataObjectId>,
+        channel_id: &T::ChannelId,
+        prize_source_account: &T::AccountId,
+    ) -> DispatchResult {
+        // remove assets if any
+        if assets.len() > 0 {
```
shamil-gadelshin:
```suggestion
        if !assets.is_empty() {
```

shamil-gadelshin (changes requested):
Looks OK.

There is a couple of minor issues. Also, please introduce some type alias for `<T as storage::Trait>::DataObjectId` and replace all occurrences.

runtime-modules/content/src/lib.rs
```diff
@@ -706,9 +676,8 @@ decl_module! {
             actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
             channel_id: T::ChannelId,
             params: ChannelUpdateParameters<T>,
+            assets: BTreeSet<DataObjectId<T>>,
```
Lezek123: I suggest renaming it to `assets_to_remove` for more clarity. Perhaps also moving it to `ChannelUpdateParameters` may be a good idea? (and analogically with video)

ignazio-bovo: Ok I believe understand your thinking @Lezek123, ("Why do I need to specify the assets when I have to delete the whole bag regardless?") The problem lies in the fact that first the user is required to empty the channel bag by specifying objects ids. And there are not ways (at least that I am aware of) to retrieve such ids from the on-chain state

shamil-gadelshin (approved):
LGTM

runtime-modules/content/src/lib.rs
```diff
@@ -727,7 +728,7 @@ decl_module! {
             origin,
             actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
             channel_id: T::ChannelId,
-            assets_to_remove: BTreeSet<DataObjectId<T>>,
+            num_objects_to_delete: u64,
```
shamil-gadelshin, line 731: Why do we need this number? What we're trying to achieve using this check `storage::Bags::<T>::get(&bag_id).objects_number == num_objects_to_delete` ?

dobertRowneySr: https://github.com/Joystream/joystream/issues/2726#issuecomment-933485454

shamil-gadelshin (approved):
Still LGTM

Lezek123 (approved):
LGTM

PR #2741, opened 2021-10-07
Modify DataObjectsUploaded event for the storage runtime pallet.

shamil-gadelshin:
We need to  add `deletion_prize` to `DataObjectsUploaded` event to enable query-node mappings.


Lezek123 (approved):
LGTM

PR #2742, opened 2021-10-07
Second part of @joystream/types updates for Giza

Lezek123:
Updates related to:
- https://github.com/Joystream/joystream/pull/2710
- https://github.com/Joystream/joystream/pull/2741
- https://github.com/Joystream/joystream/pull/2729

Include:
- Updating `WorkingGroup` type (added new operations working group)
- Removing `DaoId` type
- Updating `content` types (changes related to removal of `NewAssets` type, optional metadata etc.)
- Updating api/types augmentations

Additional improvements related to issues spotted when working on the CLI:
- added `createTypeFromConstructor` - a significantly faster (and still typesafe) version of `createType`
- changed generic `EnumVariant<T>` to allow providing values as both `{ Variant: null }` and `'Variant'` for variants with no additional values.


types/src/common.ts
```diff
@@ -97,9 +96,11 @@ export const WorkingGroupDef = {
   // _Reserved1
```
shamil-gadelshin, line 96: Shouldn't we set some placeholders here? Reserved0 and Reserved1 ?

shamil-gadelshin (changes requested):
Looks ok, but files generated locally differ from committed. 

Also, there is a question for working groups.

types/src/common.ts
Lezek123: The reason why I added the placeholders here: https://github.com/Lezek123/substrate-runtime-joystream/blob/giza-types-update-2/types/src/common.ts#L107 instead of in `WorkingGroupDef` is that I wanted `WorkingGroupDef` const to include all values that can actually be used (for example, the Pioneer uses `Object.keys(WorkingGroupDef)` to get a list of possible working group varaints), while `_Reserved0` and `_Reserved1` are not "valid" variants, they're just used to workaround the non-zero indexing on the runtime side.

shamil-gadelshin: Ok. Makes sense.

Issue #2718, opened 2021-09-29
Update docker-compose to support running multiple storage nodes


mnaamani:
In test workflow 3 we require to run 2 storage nodes and two distributor nodes to test replication of the new storage system.

Do we just add additional services in the compose file or is there a better approach?

This is related to https://github.com/Joystream/joystream/issues/2711

mnaamani:
Done in #2746

PR #2746, opened 2021-10-11
DevOps - Support running multiple storage nodes (Giza)

ahhda:
Related to issue #2718 

## Steps to run
* `yarn start`
* `./multi-storage.sh`

## Manual testing

<details>
<summary>Script and logs</summary>

Screenshot 2021-10-11 at 12 49 01 PM

Screenshot 2021-10-11 at 12 48 37 PM

</details>

### DIFF
--- a/docker-compose.multi-storage.yml
+++ b/docker-compose.multi-storage.yml
@@ -11,11 +11,15 @@ services:
     volumes:
       - /data
       - /keystore
+      - ${ACCOUNT_KEYFILE}:/joystream/storage-node-v2/keyfile
     ports:
-      - '127.0.0.1:3335:${COLOSSUS_PORT_2}'
+      - '127.0.0.1:${COLOSSUS_PORT_2}:${COLOSSUS_PORT}'
     env_file:
       # relative to working directory where docker-compose was run from
       - .env
+    environment:
+      WORKER_ID: ${WORKER_ID}
+      ACCOUNT_KEYFILE: /joystream/storage-node-v2/keyfile
 
   distributor-node-2:
     image: joystream/distributor-node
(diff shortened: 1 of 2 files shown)


docker-compose.multi-storage.yml
```diff
@@ -0,0 +1,36 @@
...
+    build:
+      context: .
+      dockerfile: colossus.Dockerfile
+    depends_on:
+      - graphql-server
+    volumes:
+      - /data
+      - /keystore
+    ports:
+      - '127.0.0.1:3335:${COLOSSUS_PORT_2}'
```
mnaamani: I would have expected to only need to bind on host to a different port, but in container port is unchanged.
```suggestion
      - '127.0.0.1:${COLOSSUS_PORT_2}:3334'
```

docker-compose.multi-storage.yml
```diff
@@ -0,0 +1,36 @@
...
+    build:
+      context: .
+      dockerfile: distributor-node.Dockerfile
+    depends_on:
+      - graphql-server
+    volumes:
+      - /data
+      - /cache
+    ports:
+      - 127.0.0.1:3336:${DISTRIBUTOR_PORT_2}
```
mnaamani: Same comment about colossus port mapping.

multi-storage.sh
```diff
@@ -0,0 +1,36 @@
...
+    docker-compose -f docker-compose.yml -f docker-compose.multi-storage.yml rm -vsf colossus-2
+}
+
+down
+
+trap down EXIT
+
+docker-compose -f docker-compose.yml -f docker-compose.multi-storage.yml run -d --name \
+    colossus-2 --entrypoint sh colossus-2 -c "yarn storage-node server --queryNodeHost ${QUERY_NODE_HOST} --port ${COLOSSUS_PORT_2} \
+        --uploads /data --worker ${WORKER_ID} --apiUrl ${WS_PROVIDER_ENDPOINT_URI} --sync --syncInterval=1 \
```
mnaamani: The second instance needs to have a unique identity, so WORDER_ID, and ACCOUNT_KEYFILE should be different and ideally passed as docker `ARG`s

multi-storage.sh
```diff
@@ -0,0 +1,36 @@
...
+        --keyFile=${ACCOUNT_KEYFILE} --elasticSearchHost=${ELASTIC_SEARCH_HOST}"
+
+docker-compose -f docker-compose.yml -f docker-compose.multi-storage.yml run -d --name distributor-node-2 distributor-node-2
```
mnaamani, line 31: I'm sure there are other arguments for starting a distributor node that specify its identity, because its behavior does depend on its worker id.

mnaamani (changes requested):

mnaamani:
My reference to docker ARGs in the multi-storage PR is to make it flexible so perhaps call to docker-compose command comes from inside tests with and they would know what worker id and keys would be needed to start those nodes.

Also running the docker images in the local test environment and with the way keys might get generated it might be simpler to make use of the argument for passing a key as an SURI string. Is that argument available in colossus yet. (It might not be merged yet into this base branch)

mnaamani (approved):

PR #2751, opened 2021-10-13
Giza: Update @polkadot/api to 5.9.1

Lezek123:
This PR features an upgrade of `@polkadot/api` to `5.9.1` and other upgrades that were necessary for the compatibility, among others:
- `typescript` to `4.4.1` (>4 is required by new `@polkadot` packages)
- `ts-node` to `^10.2.1`
- `@polkadot/util`, `@polkadot/keyring` etc. to `7.3.1`
- `@polkadot/ui-keyring`, `@polkadot/ui-settings` to  `^0.85`
- `rxjs` to `^7.4.0`

`bn.js` was downgraded to `4.12`, as the current version (`5.1.2`) was not compatible with the latest `@polkadot/api`, causing an issue described here: https://github.com/polkadot-js/api/issues/4024

### Query node

- Because [`@polkadot/metadata` package has been removed](https://github.com/polkadot-js/api/blob/master/CHANGELOG.md#501-jul-11-2021), version `> 5` of the `@polkadot/api` is not compatible with the version used by `hydra-cli` and `hydra-typegen`. Those packages have been moved out of the monorepo workspaces in order to work correctly. (into `query-node/codegen`)
- Some small changes were needed in the query-node autogenerated files. [Currently they are beeing overriden in the query-node `build` script](https://github.com/Joystream/joystream/pull/2751/files#diff-18af0d2a4c43b9ab6d71fd7b418697c131aea3fcd348c5ddca5753a09b33a7c3), pinging @ondratra to investiate implementing them in Hydra.

### Pioneer

- [`apps-electron` package was removed](https://github.com/Joystream/joystream/pull/2751/commits/1b5354e0b8d65cafd5c3659bbc18d1bf30faf62f) as it was never used and required additional work to update
- All other changes are part of [this commit](https://github.com/Joystream/joystream/pull/2751/commits/1b4fa78c69cd85928507e406b5c7336ada813f67) and include:
   - adding `Error` as possible argument type of tx failed callback (this was also done in `polkadot-apps`, due to TS4 `useUnknownInCatchVariables`)
   - adding type argument in `UseTranslationResponse` generic type
   - fixing `entriesByIds` in `joy-utils/src/transport/base.ts` (using `undefined` as key in `entries` caused problems after the update)
   - using `isFunction` from `@polkadot/util` to sidestep TS4 `This condition will always return true since this function is always defined.` error (same way it was done in `polkadot-apps`)
[...]

PR #2756, opened 2021-10-14
Giza staging: Fix Mac OS issue with query node build

Lezek123:
Brings latest changes from https://github.com/Joystream/joystream/pull/2751


shamil-gadelshin (approved):
The fix works.

PR #2758, opened 2021-10-14
Giza: Query node bootstrap scripts

Lezek123:
Added scripts to fetch `members` / `workingGroups` data from `Sumer` chain at given block and support for bootstraping this data at specified block.

### Usage
`AT_BLOCK_NUMBER=1000000 WS_PROVIDER_ENDPOINT_URI=wss://rome-rpc-endpoint.joystream.org:9944 yarn workspace query-node-mappings bootstrap-data:fetch`

After running the script, mappings need to be rebuilt with `yarn workspace query-node-mappings build` before starting the node (in order to include the bootstrap data)

### Alternative approach
An alternative approach would be to use the query-node itself as a data source.
An example flow might look like this:
1. There is a running, up-to-date `sumer` query-node
2. The node is stopped right before the upgrade to `Giza` is executed
3. A `range` in the query-node `manifest.yml` is set, that ends on the upgrade block
4. The query-node is re-built
5. The processor is restarted
6. The processor stops processing data at the upgrade block
7. We can export the data we need for the new processor that will start from the upgrade block

Pros:
- We preserve `updatedAt` value for members (the current approach sets it to `=== createdAt`)
- We preserve `createdAt` and `updatedAt` values for workers (the current approach sets it to `=== scriptRunDate`)

Cons:
- Additional work needed
- We copy data that is partially flawed itself due to previous bootstrapping, which also suffered from `updatedAt`/`createdAt` issues and was setting all members entry to `genesis` and entryblock to `0` (which was fixed in the current script)
- The process is slightly more complicated


ondratra (approved):
LGTM! 

Unless `createdAt/updatedAt` times are not super important to us, it is better to read data from the `joystream-node` rather than query node because it requires running the processor locally via the process Leszek described. When reading data from `joystream-node` we can use a remote node, and that will be easier (and likely much faster).

ahhda:
## Testing
### On Mac
* Started by building the code first, `yarn && yarn workspace @joystream/types build && yarn workspace @joystream/metadata-protobuf build && yarn workspace query-node-root build`
* Ran the main bootstrap command, `AT_BLOCK_NUMBER=1000000 WS_PROVIDER_ENDPOINT_URI=wss://rome-rpc-endpoint.joystream.org:9944 yarn workspace query-node-mappings bootstrap-data:fetch`
Screenshot 2021-10-19 at 12 56 28 PM

* This successfuly updated the `members.json` and `workingGroups.json` files
Screenshot 2021-10-19 at 12 57 09 PM

* Rebuilt the mappings using the command `yarn workspace query-node-mappings build`
Screenshot 2021-10-19 at 1 33 04 PM

### Testing with local node
* Created a docker container with sumer branch code
* Ran the command `AT_BLOCK_NUMBER=1 WS_PROVIDER_ENDPOINT_URI=ws://localhost:9944 yarn workspace query-node-mappings bootstrap-data:fetch`
Screenshot 2021-10-19 at 3 47 20 PM

✅ Similar output on Linux

ahhda (approved):
We would need to update the `query-node/bootstrap.sh` script which can be a separate PR. Otherwise this PR works as expected.

mnaamani:
Really nice. It is a bit awkward that the bootstrap data has to be part of the build but I think we can work around that during deployments.

Tested and confirmed that bootstrapping occurs after processor sees 1st event only.

mnaamani (approved):

PR #2582, opened 2021-08-13
Distributor node

Lezek123:
Joystream Distributor Node (Argus)
See detailed documentation at: https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/README.md

PR #2774, opened 2021-10-19
Giza staging update

mnaamani:
Update from Giza branch (include chain metadata into repo) simplifying build steps.


types/package.json
```diff
@@ -8,7 +8,7 @@
     "prepublishOnly": "npm run build",
     "prepack": "npm run build",
     "compile": "tsc --build tsconfig.json",
-    "build": "git clean -xdf -e node_modules && yarn compile && yarn generate:all",
+    "build": "git clean -xdf -e node_modules || yarn compile && yarn generate:all",
```
Lezek123: This seems to have caused an issue in CI checks, try changing `||` to `;` (see: https://github.com/Joystream/joystream/pull/2582/commits/c46c8f49561f05fcd546fb5f30ef5b0119a22f6e)

mnaamani: thanks will fix

mnaamani: Done in https://github.com/Joystream/joystream/pull/2774/commits/851e47d5c78b15b6c13e3dd04a5a494c8f887500 and 202aedd8acc2d7b1acde85b394f05769d4b42044

build-npm-packages.sh
```diff
@@ -2,12 +2,9 @@
 
 set -e
 
-yarn
+yarn --frozen-lockfile
```
Lezek123, line 5: Oddly enough, now with `--frozen-lockfile` I'm running into issues on Ubuntu:
```
src/commands/working-groups/updateWorkerReward.ts:65:56 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(...text: string[]): string', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'string'.
  Overload 2 of 2, '(text: TemplateStringsArray, ...placeholders: string[]): string', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'TemplateStringsArray'.

65     this.log(chalk.green(`Worker ${chalk.magentaBright(workerId)} reward has been updated!`))
```
I think the magic `chalk` resolution needs to be added back

Lezek123: After adding `"chalk": "^4.0.0"` back to `resolutions` (https://github.com/Joystream/joystream/pull/2582/commits/8fe1e13a7ecbe0b2901bb236ea4239b98c3c96c6) I managed to run both `yarn build:packages` and `yarn build:apps:docker` succesfully.

Lezek123:
Please include changes from https://github.com/Joystream/joystream/pull/2708/commits/7a299eda57cf1a2c3c04fed46e988dd620c890c0 if possible (CLI fix for uploading large files)

mnaamani:
> Please include changes from [7a299ed](https://github.com/Joystream/joystream/commit/7a299eda57cf1a2c3c04fed46e988dd620c890c0) if possible (CLI fix for uploading large files)

done in 9be96b5

build-npm-packages.sh
mnaamani: Ah its a shame that dependencies are being resolved differently on linux and mac with yarn. Perhaps we are using different versions of yarn?

If you have `volta` installed it should at least ensure we are using the same version. Can you confirm. With volta I get version `1.22.4`

Lezek123: Yes, I'm using `volta` and I've got the same version of `yarn` (`1.22.4`)

Lezek123 (changes requested):
I'm still running into issues with `chalk`, please add it to resolutions, I already did that on `distributor-node` branch, where the issue was introduced: https://github.com/Joystream/joystream/commit/8fe1e13a7ecbe0b2901bb236ea4239b98c3c96c6

**Edit:** After some more investigation with fresh monorepo clone I found that even though initailly the `4.1.1` version of `chalk` is installed in root `node_modules`, it changes to `2.4.2` after the query-node is built, which means that probably some dependencies inside query-node autogenerated code are affecting this.

mnaamani:
Added chalk to resolutions :+1:

Lezek123 (approved):
Everything working as expected!

PR #2776, opened 2021-10-20
devops infra deployment parameterize deployment type

mnaamani:
Make the deployment type configurable when creating the chainspec.


mnaamani:
Screen Shot 2021-10-20 at 2 02 52 PM

Confirmed deployment with `staging` type uses expected grace periods

PR #2778, opened 2021-10-20
Giza staging update max mint capacity

mnaamani:
https://github.com/Joystream/joystream/pull/2523 May get deployed to live network before Giza.
So we are incorporating that work into Giza and bumping runtime spec version for Giza.


bwhm (approved):
LGTM!

PR #2780, opened 2021-10-20
DevOps - Add initial members, balances support

ahhda:
This PR adds the ability to specify initial members and balances while creating the chain spec file

Also resolves the following bugs:
* Docker upgrades via (`apt-get upgrade -y`) shows a prompt that prevents the user data script from completing. It has been marked as on hold when upgrading.
* Caddy requires port 80 to be open for downloading the certificates. The security group has been updated.

### How to run
* Update the config file with members and balances
Screenshot 2021-10-20 at 6 40 49 PM

* Run `./deploy-infra.sh ~/Joystream/bash-config.cfg`

### DIFF
--- a/devops/infrastructure/cloudformation/infrastructure.yml
+++ b/devops/infrastructure/cloudformation/infrastructure.yml
@@ -74,6 +74,10 @@ Resources:
           ToPort: 443
           CidrIp: 0.0.0.0/0
         - IpProtocol: tcp
+          FromPort: 80
+          ToPort: 80
+          CidrIp: 0.0.0.0/0
+        - IpProtocol: tcp
           FromPort: 22
           ToPort: 22
           CidrIp: 0.0.0.0/0
@@ -112,7 +116,8 @@ Resources:
             # Update all packages
             apt-get update -y
 
-            # Install the updates
+            # Install the updates except docker
+            apt-mark hold docker.io
             apt-get upgrade -y
 
             # Get latest cfn scripts and install them;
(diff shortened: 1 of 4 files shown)


devops/infrastructure/cloudformation/infrastructure.yml
```diff
@@ -112,7 +116,8 @@ Resources:
             # Update all packages
             apt-get update -y
 
-            # Install the updates
+            # Install the updates except docker
```
mnaamani:
```suggestion
            # Install the updates except docker, to avoid interactive prompt which blocks the flow of the script.
```

devops/infrastructure/roles/common/tasks/chain-spec-node-keys.yml
```diff
@@ -3,25 +3,44 @@
...
+  copy:
+    src: '{{ initial_balances_file }}'
+    dest: '{{ admin_code_dir }}/initial-balances.json'
+  when: initial_balances_file is defined and initial_balances_file|length > 0
+  run_once: true
+
+- name: Run chain-spec-builder to generate chainspec.json file (with initial data)
+  shell: >
+    {{ admin_code_dir }}/target/release/chain-spec-builder generate -a {{ number_of_validators }}
+    --chain-spec-path {{ chain_spec_path }} --deployment {{ deployment_type }}
```
mnaamani: the `--deployment` argument is also optional really and defaults to "live" if not specified. So maybe we can also have it wrapped in a conditional like you did for initial-members-path and balances.. ?

mnaamani (changes requested):

PR #2781, opened 2021-10-20
Giza - Bootstrap video and channel categories data

ahhda:
Adds scripts to bootstrap `channelCategories` and `videoCategories` data from the Graphql endpoint.

### How to run
* `export QUERY_NODE_URL=https://hydra.joystream.org/graphql`
* `yarn workspace query-node-mappings bootstrap-data:fetch:categories`
Screenshot 2021-10-20 at 8 11 04 PM

* Two new files `channelCategories.json` and `videoCategories.json` will be generated


query-node/.gitignore
```diff
@@ -1,2 +1,3 @@
 lib
 generated
+mappings/bootstrap-data/data/*
```
mnaamani: I agree that it doesn't make sense to include all the bootstrapping date in the git repo.

However it seems the storage system needs some initial data.

The ideal solution would be to indeed ignore this directory and improve `query-node/mappings/bootstrap-data/index.ts` to dynamically load files if found here (generated by running the "fetch:*" script) instead of using javascript `import` statement. When file is not found, the default "json" data would be used.

mnaamani (changes requested):
The fetching and import works perfectly.
I left a suggestion about how to deal with including the bootstrap data in the repo.

One improvement that can be made but is not critical right now is for the categories fetching script to "double check" on chain state at a particular block.

query-node/.gitignore
Lezek123: The reason I chose static import is that this way we get the additional benefit of typesafety check during build, so `yarn workspace query-node-mappings build` will fail if the json data doesn't match the expected format.

mnaamani: Yes perhaps type safety is more important.
Ok lets not .gitignore the data folder, and we just have to be sure not to check in bootstrap data.

query-node/mappings/bootstrap.ts
```diff
@@ -55,4 +69,22 @@ export async function bootstrapData({ store }: StoreContext): Promise<void> {
...
+  await Promise.all(channelCategories.map((m) => store.save<ChannelCategory>(m)))
+
+  const videoCategories = videoCategoriesData.map(
+    (m) =>
+      new VideoCategory({
+        id: m.id,
+        name: m.name,
+      })
+  )
+  await Promise.all(channelCategories.map((m) => store.save<VideoCategory>(m)))
```
mnaamani:
```suggestion
  await Promise.all(videoCategories.map((m) => store.save<VideoCategory>(m)))
```

query-node/mappings/bootstrap.ts
```diff
@@ -55,4 +69,22 @@ export async function bootstrapData({ store }: StoreContext): Promise<void> {
     )
   })
   await Promise.all(workers.map((w) => store.save<Worker>(w)))
+
+  const channelCategories = channelCategoriesData.map(
+    (m) =>
+      new ChannelCategory({
+        id: m.id,
+        name: m.name,
+      })
```
mnaamani, line 82: I think for both the ChannelCategory you must also include the other fields:
channels, createdInBlock, createdAt and updatedAt.

mnaamani: As I see this error in processor when bootstrapping occurs:
```
ERROR: Stopping the proccessor due to errors: {
  "message": "null value in column \"created_in_block\" violates not-null constraint",
  "length": 354,
  "name": "QueryFailedError",
  "severity": "ERROR",
  "code": "23502",
  "detail": "Failing row contains (1, 2021-10-21 12:34:02.825962, YSVbUNz5DM, 2021-10-21 12:34:02.825962, null, null, null, 1, Film & Animation, null, 'anim':2A 'film':1A, Film & Animation).",
  "schema": "public",
  "table": "channel_category",
  "column": "created_in_block",
  "file": "execMain.c",
  "line": "1948",
  "routine": "ExecConstraints",
  "query": "INSERT INTO \"channel_category\"(\"id\", \"created_at\", \"created_by_id\", \"updated_at\", \"updated_by_id\", \"deleted_at\", \"deleted_by_id\", \"version\", \"name\", \"created_in_block\") VALUES ($1, DEFAULT, $2, DEFAULT, DEFAULT, DEFAULT, DEFAULT, $3, $4, DEFAULT) RETURNING \"created_at\", \"updated_at\", \"version\"",
  "parameters": [
    "1",
    "YSVbUNz5DM",
    1,
    "Film & Animation"
  ]
}
```

query-node/mappings/bootstrap.ts
mnaamani: ~I'm not sure there are actually any mappings for VideoCategories in the query node~
I was looking in wrong place.

query-node/mappings/bootstrap.ts
```diff
@@ -55,4 +69,22 @@ export async function bootstrapData({ store }: StoreContext): Promise<void> {
     )
   })
   await Promise.all(workers.map((w) => store.save<Worker>(w)))
+
+  const channelCategories = channelCategoriesData.map(
+    (m) =>
+      new ChannelCategory({
+        id: m.id,
+        name: m.name,
```
mnaamani, line 77:
```suggestion
        name: m.name,
        channels: [],
        createdInBlock: 1,
        createdAt: new Date(),
        updatedAt: new Date(),
```

Something like that would work.

query-node/mappings/bootstrap.ts
```diff
@@ -55,4 +69,22 @@ export async function bootstrapData({ store }: StoreContext): Promise<void> {
...
+      })
+  )
+  await Promise.all(channelCategories.map((m) => store.save<ChannelCategory>(m)))
+
+  const videoCategories = videoCategoriesData.map(
+    (m) =>
+      new VideoCategory({
+        id: m.id,
+        name: m.name,
```
mnaamani, line 90:
```suggestion
        name: m.name,
        videos: [],
        createdInBlock: 1,
        createdAt: new Date(),
        updatedAt: new Date(),
```

mnaamani (approved):

query-node/mappings/bootstrap.ts
```diff
@@ -55,4 +69,30 @@ export async function bootstrapData({ store }: StoreContext): Promise<void> {
...
+        videos: [],
+        createdInBlock: 1,
```
mnaamani: @Lezek123 recommended that we should fetch the actual values when fetching the categories.

mnaamani (changes requested):
I added comment with suggestion from leszek.

mnaamani (approved):

PR #2783, opened 2021-10-21
DevOps - Update wait time for AMI creation to 1 hour

ahhda:
Updates the wait timeout for creation of AWS AMI from 20 minutes (default) to 1 hour.

### DIFF
--- a/devops/infrastructure/github-action-playbook.yml
+++ b/devops/infrastructure/github-action-playbook.yml
@@ -24,6 +24,8 @@
       amazon.aws.ec2_ami:
         instance_id: '{{ instance_id }}'
         wait: yes
+        # How long before wait gives up, in seconds
+        wait_timeout: 3600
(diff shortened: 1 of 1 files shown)


mnaamani:
A potentially dumb question .. if the build completes in less than an hour, the flow will still end before the full hour correct?

ahhda:
> A potentially dumb question .. if the build completes in less than an hour, the flow will still end before the full hour correct?

That is correct. `wait_timeout` only sets the maximum time to wait for the task to complete. The ping happens every 15s and as soon as the task is complete, the build will be complete.

This is the logic internally:
```python
if wait:
    delay = 15
    max_attempts = wait_timeout // delay
    waiter = get_waiter(connection, 'image_available')
    waiter.wait(ImageIds=[image_id], WaiterConfig=dict(Delay=delay, MaxAttempts=max_attempts))
```

PR #2785, opened 2021-10-21
DevOps - Run setup in non interactive mode

ahhda:
This will prevent the system to prompt for anything


ahhda:
Closing this as the issue was with https://getsubstrate.io/ not being accessible.

PR #2790, opened 2021-10-25
query-node: fix date types (exported as string by fetch scripts)

mnaamani:
exported dates from graphql query are in string format.


Lezek123 (approved):
LGTM

PR #2794, opened 2021-10-25
Giza staging query node fixes

mnaamani:
- date type fix from giza branch
- avoid using old types in bootstrap fetch scripts. Typically we are running the scripts after the upgrade and connecting to joystream-node with new runtime using old types is failing because new types are not defined. In this particular scenario for Giza, the types being exported are unchanged in Giza so we can use the new types version, and since we add "legacy" types the new version also works when connecting to old (sumer) runtime.


mnaamani:
cc: @ahhda

ahhda (approved):
LGTM

Screenshot 2021-10-26 at 12 04 58 PM

Lezek123 (approved):
LGTM
In the future we can force using older chain metadata in `@polkadot/api` in case the types change.

PR #2792, opened 2021-10-25
Refactor distribution buckets.

shamil-gadelshin:
#### Changes
- updated distribution bucket storage (from distribution type family container to a double map)
- types

#### Review comments
- Please, pay attention to the newly introduced iterations for the double map prefix (by family id). The most dangerous code change is in the `DistributionBucketPicker` struct implementation - it introduces the NxM iterations over the distribution bucket families and distribution buckets. It looks bad. The mitigation would be introducing and supporting a cache (denormalization) in the distribution bucket family to save `accepting new bags` status for buckets. Should we just roll back the changes from this PR?

PR #2797, opened 2021-10-26
Refactor distribution buckets (with extra state).

shamil-gadelshin:
#### Changes
- updated distribution bucket storage (from distribution type family container to a double map)
- introduced a new type - `DistributionBucketState`
- types

#### Review comments
- This is the second option for the distribution bucket family changes ([this is the first one](https://github.com/Joystream/joystream/pull/2792))


types/src/storage.ts
```diff
@@ -220,13 +220,23 @@ export class DistributionBucket
   })
   implements IDistributionBucket {}
 
+export type IDistributionBucketState = {
+  accepting_new_bags: bool
+}
+
+export class DistributionBucketState
```
Lezek123, line 227: Please add the new type to `storageTypes` const below and regenerate `defs.json`

Lezek123 (changes requested):

types/src/storage.ts
Lezek123: Also please update `chain-metadata.json` with `yarn update-chain-metadata`

shamil-gadelshin: Fixed.

Lezek123 (approved):
LGTM
Didn't encounter any issues during my local tests

bedeho (approved):
Approved, with 2nd review of core concept later.

shamil-gadelshin:
Superseded by https://github.com/Joystream/joystream/pull/2897

Issue #2739, opened 2021-10-07
Unlock off-chain overloading of video


bedeho:
# Background
Currently, we have represented concepts like videos, playlists and episodic content (series) as explicit & distinct blockchain level entities within channels. They all have off-chain metadata standards, which allows the ability to have multiple and evolving standard for the write-only part of the information & assets that only the channel owner controls, stuff like descriptions, titles, etc. That unlocks a lot of innovation at the presentation layer, as app-developers (e.g. Gateways) can unilaterally introduce new ways of representing these concepts.

# Problem
It's not really clear why we need these distinct concepts, as anyone can in principle (if they disregard the naming) represent anything as a video by overloading the metadata suitably. So they can represent all the other concepts above for example. Having the distinct entity types then leads to two problems

1) It then seems a bit misleading, perhaps to the extent that it can cause a more limited understanding of what is possible, to have these distinct concepts. It could lead to developers not understanding that they are entirely able to introduce totally new things like books, tracks, albums, etc. 

2) It means general purpose concepts, like NFTs, do not work for all concepts out of the box, unless we manually go and add it to them one by one.

# Solution 
Drop all the distinct types, just have one universal channel entity called `Content`, which works exactly like `Video`  currently, just with methods and fields renamed to be more neutral.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202329093222211) by [Unito](https://www.unito.io)

dobertRowneySr:
I approve

mnaamani:
Putting all the "details" into the metadata seems like a very workable idea. The less the runtime code needs to change the better. On the point of NFTs we may want to add ownership transfer capability to individual items of content otherwise the NFT would encompass the whole channel. Maybe that is acceptable?

bedeho:
> On the point of NFTs we may want to add ownership transfer capability to individual items of content otherwise the NFT would encompass the whole channel. Maybe that is acceptable?

NFTs already only exist for individual videos, i.e. individual content items, not entire channels. So what this would mean under the new approach would be that all possible content items had the same fundamental NFT functionality, at least technically.

I guess both of you are open to it, but the question is then timing, do we do this now or later?

dobertRowneySr:
I would prefer later, so I can focus on integration tests

bedeho:
Ok, sounds wise, lets keep this open then, Ill add the `tech-debt` label.

dobertRowneySr:
[...]

PR #2798, opened 2021-10-26
Video / Channel map clearing unrolling implementation

dobertRowneySr:
Things done:
- The amount of channel/videos migrated each block is an arbitrary number, we should think that through anyway, when we will compute extrinsics weights
- Migration state is saved on chain
- I didn't rename the VideoById map into ContentById, I think it's best to rename everything when we will implement idea #2739 
- Both VideoById and ChannelById maps are cleared (devoided of their content)
- extrinsics that perform a storage read are ensured to access the chain in a post-migration state (once migration is started)
- Counters like `NextVideoId` are **NOT** resetted
- corresponding types have been generated to `@joystream/types`
- `tests/network-tests/run_tests.sh` has been updated in order to perform the tests
- the migration tests itself are `joystream-cli`-based and extremely rudimental (I am not proficient in typescript yet)
- updated json-schemas to create video / channels
#### Specs:
- `joystream-cli` version used to send TXs to `sumer` is 0.5.1
- `sumer` docker image pulled directly from docker-hub
- added extra image to the `tests/assets` folder called `joystream.png`, not added to the PR


node/Cargo.toml
```diff
@@ -70,6 +70,9 @@ sc-cli = { package = 'sc-cli', git = 'https://github.com/paritytech/substrate.gi
 frame-benchmarking-cli = { package = 'frame-benchmarking-cli', git = 'https://github.com/paritytech/substrate.git', rev = '2cd20966cc09b059817c3ebe12fc130cdd850d62', optional = true }
 node-inspect = { package = 'node-inspect', git = 'https://github.com/paritytech/substrate.git', rev = '2cd20966cc09b059817c3ebe12fc130cdd850d62', optional = true }
 
+# config dependencies
+content = { package = 'pallet-content', default-features = false, path = '../runtime-modules/content' }
```
mnaamani: This should not be necessary, the type you need in `chain_spec/mod.rs` is available on the runtime crate. See later comment.

node/src/chain_spec/mod.rs
```diff
@@ -366,6 +366,14 @@ pub fn testnet_genesis(
                 next_series_id: 1,
                 next_person_id: 1,
                 next_channel_transfer_request_id: 1,
+                video_migration: content::MigrationConfigRecord {
```
mnaamani: `use node_runtime::content::MigrationConfigRecord`

There is no need to add the content module dependency explicitly in `node/Cargo.toml`

tests/network-tests/pre_migration_hook.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -e
+
+# SCRIPT_PATH="$(dirname "${BASH_SOURCE[0]}")"
+# cd $SCRIPT_PATH
+
+export AUTO_CONFIRM=true
+
+sleep 5 # needed otherwise docker image won't be ready yet
+echo "creating 1 channel"
```
mnaamani: The script needs to ensure that Alice account is added to the cli config because I get the following error:
```
creating 1 channel
 ›   Error: Membership required to access this command!
adding 1 video to the above channel
 ›   Error: Channel by id 1 not found!
```

And we will want this test to run in CI so you have to assume cli has not been configured.

I assume the `|| true` is there to allow the command to fail gracefully since there is not storage node up to handle upload?

How can we force this script to use the older version of joystream-cli that works for sumer?

tests/network-tests/post_migration_hook.sh
```diff
@@ -4,10 +4,13 @@ set -e
...
 
-# verify existence of the three new groups
+echo "*** verify existence of the 5 new groups ***"
 yarn joystream-cli working-groups:overview --group=operationsAlpha
 yarn joystream-cli working-groups:overview --group=operationsBeta
 yarn joystream-cli working-groups:overview --group=operationsGamma
 yarn joystream-cli working-groups:overview --group=curators
 yarn joystream-cli working-groups:overview --group=distributors
 
+echo "*** verify previously created channel and video are cleared ***"
```
mnaamani: Depending on how many blocks are expected to be needed for the migration to complete we would need to `sleep` for some period of time.

runtime-modules/content/src/tests/migration.rs
```diff
@@ -0,0 +1,71 @@
...
+        );
+
+        // 1 channel & 1 video
+        assert_eq!(VideoById::<Test>::iter().count(), 1);
+        assert_eq!(ChannelById::<Test>::iter().count(), 1);
+
+        // performing migration
+        Content::on_runtime_upgrade();
+
+        // assertions
```
mnaamani: Perhaps we want to have some assertions while migration has not yet completed that dispatches correctly fail with expected error.

runtime-modules/content/src/lib.rs
```diff
@@ -101,8 +106,25 @@ pub trait Trait:
 
     /// The storage type used
     type DataObjectStorage: storage::DataObjectStorage<Self>;
+
+    /// Video migrated in each block during migration
+    type VideosMigratedEachBlock: Get<u64>;
```
mnaamani: nit picking here, but a better name could be:
```suggestion
    type VideoMigrationsPerBlock: Get<u64>;
```

mnaamani (changes requested):
Implementation looks correct. I left some thoughts and questions.

We could probably increase the number of videos and channels migrated per block to something a little higher.

I think the integration testing still needs some work.

runtime-modules/content/src/tests/migration.rs
mnaamani: For calls to `runt_to_block(N)`, `N` should also be a function of number of videos and channels that exists and the number of videos and channels that are migrated per block.

mnaamani:
You missed the the guard `Self::ensure_channel_migration_done()?;` on dispatchables:
`create_channel()`, 
~`update_channel()`~, `create_video()`,
`update_channel_censorship_status()`,

dobertRowneySr:
> You missed the the guard `Self::ensure_channel_migration_done()?;` on dispatchables: `create_channel()`, `update_channel()`, `update_channel_censorship_status()`,

Shouldn't it be safe to perform `create_video/channel` operation even during a migration?
`on_runtime_upgrade` is executed before the `on_initialize` (that is before any extrinsics in the block). This means that any create channel / video extrinsics targeting the giza runtime should be able to have a channel on the `ChannelById` map at id `NextChannelId` which is outside the scope of migration (same for create video )?
It should be only the range `[start, NextChannelId_before_migration)` which is invalid during the upgrade and hence should not be accessible, right? Am I missing something?

mnaamani:
> > You missed the the guard `Self::ensure_channel_migration_done()?;` on dispatchables: `create_channel()`, `update_channel()`, `update_channel_censorship_status()`,
> 
> Shouldn't it be safe to perform `create_video/channel` operation even during a migration? `on_runtime_upgrade` is executed before the `on_initialize` (that is before any extrinsics in the block). This means that any create channel / video extrinsics targeting the giza runtime should be able to have a channel on the `ChannelById` map at id `NextChannelId` which is outside the scope of migration (same for create video )? It should be only the range `[start, NextChannelId_before_migration)` which is invalid during the upgrade and hence should not be accessible, right? Am I missing something?

You are correct. It is safe. I assumed since update_channel and update_video are disabled then it could be inconvenient if someone creates a video or channel and shortly after tries to update them (while migration is still ongoing). I suppose we could have `ensure_channel_migration_done()` take an argument for a video id, channel id to determine if it is safe to update? But I don't want to over-complicate it.

mnaamani:
There is another concern also. call to `create_video()` mutates a channel's `num_videos`. If member tries to create a new video for an "old" channel it will fail because the `Channel` runtime type has changed. I'm concerned this results not in just transaction failing, but the joystream-node crashing. Can you test this to be sure?

runtime-modules/content/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub use common::{
     AssetUrls, MembershipTypes, StorageOwnership,
 };
 
+// alias for weight type
+type Weight = u64;
```
shamil-gadelshin: Why did you introduce this type? There is a similar type in the `frame-support/weights` namespace.

runtime-modules/content/src/lib.rs
```diff
@@ -1267,10 +1309,97 @@ decl_module! {
         ) {
             Self::not_implemented()?;
         }
+
+        fn on_initialize(_n: T::BlockNumber) -> Weight {
+            Self::perform_video_migration();
+            Self::perform_channel_migration();
+
+            10_000_000
```
shamil-gadelshin: Add please a TODO comment here for weight adjusting for the Olympia release (similar to the extrinsics' comments).

runtime-modules/content/src/lib.rs
```diff
@@ -1267,10 +1309,97 @@ decl_module! {
...
+                current_id + T::ChannelsMigratedEachBlock::get().into(),
+                final_id,
+            );
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            // clear maps: (iterator are lazy and do nothing unless consumed)
+            for id in current_id.into()..next_id.into() {
```
shamil-gadelshin, line 1362: Does the exclusive range mean that we don't remove final IDs?

runtime-modules/content/src/lib.rs
```diff
@@ -1267,10 +1309,97 @@ decl_module! {
...
+    }
+
+    /// Migrate Channels
+    fn perform_channel_migration() {
+        let MigrationConfigRecord {
+            current_id,
+            final_id,
+        } = <ChannelMigration<T>>::get();
+
+        if current_id != final_id {
```
shamil-gadelshin: What will happen if `final_id` is set less than `current_id`? Will it go infinitely?

runtime-modules/content/src/lib.rs
```diff
@@ -101,8 +106,25 @@ pub trait Trait:
...
+
+    /// Channel migrated in each block during migration
+    type ChannelsMigratedEachBlock: Get<u64>;
+}
+
+/// Data structure in order to keep track of the migration
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct MigrationConfigRecord<NumericId> {
+    pub current_id: NumericId,
```
shamil-gadelshin, line 122: Please, comment on these IDs. Their meaning is not clear even after reading the usage code (see `final_id` comment below).

bedeho:
Just turn everything off here, no reason to try to be clever here, its is all risk and no value.

shamil-gadelshin (changes requested):
Found some minor runtime issues - as a general comment I suggest changing migration conditions for ID comparison from 'equals' to 'greater equal - that will make it safer in case of the invalid config data. But I couldn't run the tests in the local environment:

```
shamixs-MacBook-Pro:network-tests shamixRUNTIME=latest TARGET_RUNTIME=giza ./run-tests.sh
Oct 27 12:42:53.327  INFO Building chain spec    
Oct 27 12:42:54.708  INFO 💸 new validator set of size 1 has been elected via ElectionCompute::OnChain for era 0    
creating 1 channel
 ›   Error: Unexpected error while trying to read from the data directory.(/Users/shamix/Library/Application Support/joystream-cli)! Permissions issue?
adding 1 video to the above channel
 ›   Error: Unexpected error while trying to read from the data directory.(/Users/shamix/Library/Application Support/joystream-cli)! Permissions issue?
Extracting wasm blob from target joystream/node image.
150c82d5d154a06fbeae2f674a822e055262ab163839c671b7383b10339feb16
Runtime Version: 9.7.0
Performing runtime upgrade.
yarn workspace v1.22.4
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
yarn run v1.22.4
$ node -r ts-node/register --unhandled-rejections=strict src/dev-set-runtime-code.ts /Users/shamix/tmp/runtime.compact.wasm
WASM bytes: 2290258
✨  Done in 15.89s.
✨  Done in 16.37s.
Runtime upgraded.
Performing migration tests
*** verify existence of the 5 new groups ***
yarn run v1.22.4
$ /Users/shamix/rust/joystream/substrate-runtime-joystream/node_modules/.bin/joystream-cli working-groups:overview --group=operationsAlpha
 ›   Error: Unexpected error while trying to read from the data directory.(/Users/shamix/Library/Application Support/joystream-cli)! Permissions issue?
error Command failed with exit code 245.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Oct 27 12:43:29.328  WARN Libp2p => Failed to put record: QuorumFailed { key: Key(b"\x12 F \x87\x98\xc8\x0b\xe6S\x1cjdT1-\xb7\xd1PYb7k<\xca\x16\xcd\x18\xf9\x10\xd6\xfe\xa8^"), success: [], quorum: 20 }    
Oct 27 12:43:30.005  INFO 🙌 Starting consensus session on top of parent 0xf372d7b3a821c79c4acfd5abee01261fd3a10051f54119ea4163d0ac4f9fa6e0    
Oct 27 12:43:30.009  INFO 🎁 Prepared block for proposing at 6 [hash: 0x7854d7cb55807c3d9714b371b693bf81b9badcfb324b0988e749f20b5761c277; parent_hash: 0xf372…a6e0; extrinsics (2): [0xbdda…c30d, 0xd423…75fe]]    
Oct 27 12:43:30.013  INFO 🔖 Pre-sealed block for proposal at 6. Hash now 0x35450e24dc8f792594876c28d0017caadea4914e22c758a0fc5a4282ad8d1f13, previously 0x7854d7cb55807c3d9714b371b693bf81b9badcfb324b0988e749f20b5761c277.    
Oct 27 12:43:30.013  INFO ✨ Imported #6 (0x3545…1f13)    
Oct 27 12:43:34.330  INFO 💤 Idle (0 peers), best: #6 (0x3545…1f13), finalized #4 (0x29ca…eb56), ⬇ 0 ⬆ 0    
Oct 27 12:43:36.005  INFO 🙌 Starting consensus session on top of parent 0x35450e24dc8f792594876c28d0017caadea4914e22c758a0fc5a4282ad8d1f13    
Oct 27 12:43:38.108  INFO 🎁 Prepared block for proposing at 7 [hash: 0x13bff36df11a07b595e14dd91f89a8393491568dcca9759547667b71c8989e57; parent_hash: 0x3545…1f13; extrinsics (3): [0x8c83…e077, 0x1b99…d2cf, 0x37c0…9b19]]    
Oct 27 12:43:38.114  INFO 🔖 Pre-sealed block for proposal at 7. Hash now 0x1ac31be3ce6484d69e7ea71e1de9d2341431e7bb430ec9c711a04308cc0ca611, previously 0x13bff36df11a07b595e14dd91f89a8393491568dcca9759547667b71c8989e57.    
Oct 27 12:43:38.129  INFO ✨ Imported #7 (0x1ac3…a611)    
Oct 27 12:43:39.335  INFO 💤 Idle (0 peers), best: #7 (0x1ac3…a611), finalized #4 (0x29ca…eb56), ⬇ 0 ⬆ 0    
Oct 27 12:43:42.004  INFO 🙌 Starting consensus session on top of parent 0x1ac31be3ce6484d69e7ea71e1de9d2341431e7bb430ec9c711a04308cc0ca611    
Oct 27 12:43:42.024  INFO 🎁 Prepared block for proposing at 8 [hash: 0xc364f8ae9db4e28167ef26ada2f52e0eb5e886eba99829ef3664810ed3d76aa6; parent_hash: 0x1ac3…a611; extrinsics (2): [0xa107…bb53, 0x844f…bee6]]    
Oct 27 12:43:42.027  INFO 🔖 Pre-sealed block for proposal at 8. Hash now 0x1835582f5094be37dae8b14f7c6f5368637cd029eb6fcedf9ee5ad4fc3b16062, previously 0xc364f8ae9db4e28167ef26ada2f52e0eb5e886eba99829ef3664810ed3d76aa6.    
Oct 27 12:43:42.028  INFO ✨ Imported #8 (0x1835…6062)    
0910bce4580ac40634341c30712765a33b87d94f12d8b3cd42eea4cf10b3169c
0910bce4580ac40634341c30712765a33b87d94f12d8b3cd42eea4cf10b3169c
```

runtime-modules/content/src/lib.rs
dobertRowneySr: yes, ids removed are in half-open range `[current_id, next_id)`. If that is what you are asking

runtime-modules/content/src/lib.rs
dobertRowneySr: Nothing since the for loop execution will detect `final_id < current_id` and the range will not be valid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=54cf09969618e92a25e93ce6e8339c19

tests/network-tests/pre_migration_hook.sh
dobertRowneySr: The only way I know is `npm install @joystream/cli -g` to get the latest testnet-compatible cli version (currently
0.5.1).
I can maybe add a guard in the .sh file
The || true part is used to not get a non-zero exit code that I otherwise would get not having a storage node available

mnaamani:
> Just turn everything off here, no reason to try to be clever here, its is all risk and no value.

~@ignazio-bovo This hasn't been addressed yet. Just prevent all extrinsics until migration is complete.~
My bad I didn't see the commit.

dobertRowneySr:
As discussed during the last storage call, this PR will address the runtime part of the migration only

docker-compose.yml
```diff
@@ -4,7 +4,7 @@
 version: '3.4'
 services:
   joystream-node:
-    image: joystream/node:latest
+    image: joystream/node:giza
```
mnaamani: Our build script which downloads cached builds tags image with latest. So I prefer this tag remains `latest` for now.

runtime-modules/content/src/lib.rs
```diff
@@ -1290,15 +1408,18 @@ impl<T: Trait> Module<T> {
...
         ensure!(
             ChannelById::<T>::contains_key(channel_id),
             Error::<T>::ChannelDoesNotExist
         );
         Ok(ChannelById::<T>::get(channel_id))
     }
 
-    fn ensure_video_exists(video_id: &T::VideoId) -> Result<Video<T>, Error<T>> {
+    fn ensure_video_validity(video_id: &T::VideoId) -> Result<Video<T>, Error<T>> {
+        let _ = Self::ensure_migration_done();
```
mnaamani: error is ignored..
```suggestion
        Self::ensure_migration_done()?;
```

runtime-modules/content/src/lib.rs
```diff
@@ -1290,15 +1408,18 @@ impl<T: Trait> Module<T> {
         Ok(Self::curator_group_by_id(curator_group_id))
     }
 
-    fn ensure_channel_exists(channel_id: &T::ChannelId) -> Result<Channel<T>, Error<T>> {
+    fn ensure_channel_validity(channel_id: &T::ChannelId) -> Result<Channel<T>, Error<T>> {
+        let _ = Self::ensure_migration_done();
```
mnaamani: error is ignored
```suggestion
        let _ = Self::ensure_migration_done()?;
```

mnaamani (changes requested):
call to ensure_migration_done() in the ensure_video/channel_validity is pointless .. the result is ignored.

runtime-modules/content/src/lib.rs
dobertRowneySr: This was very stupid on my end. 
I added tests for all the relevant extrinsics in order to make sure that no such problems arise.
I cannot commit your suggestion because I cannot convert from `DispatchError` to `Error<T>`. I have found another solution however

runtime-modules/content/src/lib.rs
```diff
@@ -1371,22 +1371,22 @@ impl<T: Trait> Module<T> {
     /// Ensure Channel Migration Finished
 
     /// Ensure Video Migration Finished
-    fn ensure_migration_done() -> DispatchResult {
+    fn ensure_migration_done() -> bool {
```
mnaamani: with the change of return type a better name for the function would now be `is_migration_done() -> bool`

runtime-modules/content/src/tests/migration.rs
```diff
@@ -54,22 +54,104 @@ fn migration_test() {
 
         // only 20 videos migrated so far (specifically 1..=20)
         run_to_block(2);
+        assert_eq!(Content::ensure_migration_done(), false,);
```
mnaamani: `assert!(!Content::ensure_migration_done())` is cleaner

runtime-modules/content/src/tests/migration.rs
```diff
@@ -54,22 +54,104 @@ fn migration_test() {
...
+                Origin::signed(FIRST_MEMBER_ORIGIN),
+                ContentActor::Member(FIRST_MEMBER_ID),
+                <Test as storage::Trait>::ChannelId::one(),
+                false,
+                b"test".to_vec()
+            ),
+            Error::<Test>::MigrationNotFinished
+        );
+
+        run_to_block(6); // video migration finished 5 blocks later
```
mnaamani: The number of blocks to run to should be a function the number of channels/videos migrated per block as well as the number of videos and channels actually created in the test, to avoid needing to update the value here when those values are changed.

runtime-modules/content/src/tests/migration.rs
```diff
@@ -57,97 +146,15 @@ fn migration_test() {
...
-            Content::update_channel_censorship_status(
-                Origin::signed(FIRST_MEMBER_ORIGIN),
-                ContentActor::Member(FIRST_MEMBER_ID),
-                <Test as storage::Trait>::ChannelId::one(),
-                false,
-                b"test".to_vec()
-            ),
-            Error::<Test>::MigrationNotFinished
-        );
+        helper_test_all_relevant_extrinsics();
```
mnaamani: Its not clear from the name what you are testing for.

mnaamani (changes requested):

runtime-modules/content/src/tests/migration.rs
mnaamani: Also for completeness it makes sense to test also that after migration is complete that extrinsics succeed.

mnaamani:
One more missing item, the two runtime types `ChannelMigrationConfig, VideoMigrationConfig` also need to be added to types package.

mnaamani (commented):
Updating types was missing a step.
The migration test looked a little fragile still with some assumption. I suggest this improvement in https://github.com/ignazio-bovo/joystream/pull/1

PR #2801, opened 2021-10-27
Giza fix working group enum

mnaamani:
Fix joystream types definition for runtime `enum common::WorkingGroup` type.

Parity scale-coded encoding of `enum`s is simply the [index of the variant](https://docs.substrate.io/v3/advanced/scale-codec/#enumerations-tagged-unions).

Technically only the changes in `@joystream/types` package is needed, but I removed the redundant rust code to avoid any confusion.

### DIFF
--- a/runtime-modules/common/src/working_group.rs
+++ b/runtime-modules/common/src/working_group.rs
@@ -11,28 +11,24 @@ use strum_macros::EnumIter;
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Copy, Debug, PartialOrd, Ord)]
 pub enum WorkingGroup {
-    /* Reserved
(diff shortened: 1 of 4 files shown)


runtime-modules/common/src/working_group.rs
```diff
@@ -11,28 +11,24 @@ use strum_macros::EnumIter;
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Copy, Debug, PartialOrd, Ord)]
 pub enum WorkingGroup {
-    /* Reserved
-        /// Forum working group: working_group::Instance1.
-        Forum,
-    */
     /// Storage working group: working_group::Instance2.
-    Storage = 2isize,
```
mnaamani, line 19: The value  "2isize" assigned here is not used when the `enum` is parity scale-encoded (how it is stored both in chain state kv-store and in extrinsics).

PR #2802, opened 2021-10-27
runtime: do not init persisted working group modules

mnaamani:
We discovered that runtime modules instances of type WorkingGroup even if renamed persist their state presumably because the key (in key value store) is derived based on the module type name and instance number and not the unique name in decl_runtime construction macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2332,7 +2332,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
-version = "5.8.0"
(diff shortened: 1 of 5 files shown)


dobertRowneySr (commented):
So do we need only to rename the persisted WG in the CLI mappings in order to send correct TXs to the new runtime?

mnaamani:
> So do we need only to rename the persisted WG in the CLI mappings in order to send correct TXs to the new runtime?

Correct

PR #2804, opened 2021-10-28
Giza fix working group enum

mnaamani:
We concluded that chancing the enum variant ordering for WorkingGroup type in a runtime upgrade while existing state (proposals) references the WorkingGroup vlaue should be avoided.

### DIFF
--- a/runtime-modules/common/src/working_group.rs
+++ b/runtime-modules/common/src/working_group.rs
@@ -11,28 +11,24 @@ use strum_macros::EnumIter;
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Copy, Debug, PartialOrd, Ord)]
 pub enum WorkingGroup {
-    /* Reserved
-        /// Forum working group: working_group::Instance1.
-        Forum,
-    */
     /// Storage working group: working_group::Instance2.
-    Storage = 2isize,
+    Storage,
 
     /// Storage working group: working_group::Instance3.
-    Content = 3isize,
+    Content,
 
     /// Operations working group: working_group::Instance4.
-    OperationsAlpha = 4isize,
+    OperationsAlpha,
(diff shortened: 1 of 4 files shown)


types/src/common.ts
```diff
@@ -84,8 +82,6 @@ export const WorkingGroupDef = {
 } as const
 export type WorkingGroupKey = keyof typeof WorkingGroupDef
 export class WorkingGroup extends JoyEnum({
```
shamil-gadelshin: Could be just `WorkingGroupDef` without expanding.

runtime-modules/common/src/working_group.rs
```diff
@@ -11,28 +11,24 @@ use strum_macros::EnumIter;
...
-    OperationsAlpha = 4isize,
+    OperationsAlpha,
 
     /// Gateway working group: working_group::Instance5.
-    Gateway = 5isize,
+    Gateway,
 
     /// Distribution working group: working_group::Instance6.
-    Distribution = 6isize,
+    Distributione,
```
shamil-gadelshin: Here is the typo that affects the building process (check failed github actions)

shamil-gadelshin (changes requested):
The build is failing because of the typo. Please, fix it.

runtime-modules/common/src/working_group.rs
mnaamani: Thanks for spotting this.
Fixed in 43d27f230c67c493e301ebaf44d6479ac6fd11ca

types/src/common.ts
mnaamani: fixed in ca4b7d778cf2235c529ad52e2c57cc3814e04864
Tested in pioneer proposal creation

shamil-gadelshin (approved):
LGTM

Tested with storage node scripts including:
`yarn storage-node leader:update-bag -a=0 -i static:wg:distribution --dev` that covers the use cases for this PR.

PR #2805, opened 2021-10-28
Drop old legacy types

mnaamani:
Dropped all pre-sumer runtime types from @joystream/types


Lezek123 (approved):
LGTM!
No errors in Pioneer/CLI when connecting to Sumer chain (unless trying to actually create any of the legacy types)

PR #2809, opened 2021-11-02
Tests: Setup new chain scenario

mnaamani:
Deterministic key derivation for testing accounts.
Add scenario for initializing a new staging network with wokers, a council and some mock content.

run scenario against local node:
`yarn network-tests run-test-scenario setup-new-chain`

By default scenario expect sudo account to be `//Alice` and uses deterministic key derivation that is not protected with a secret.

For a public staging network you will need to run with additional configuration env variables:

```
# The sudo account secret or SURI
export SUDO_ACCOUNT_URI=0x1239d81f293f87...
# secret key or SURI for account that has balance that will be transferred to newly created accounts
export TREASURY_ACCOUNT_URI=0x2a383f28e497..
# mini secret used for key derivation of new accounts
export SURI_MINI_SECRET=0xa9a37c4dd9457728ae16a1b521acde1035395b913a630185e4a6f3e64523de9a
# websocker rpc endpoint for the network
NODE_URL=wss://rpc.remote-node.net:9944
```

After the scenario runs to completion a list of all created account addresses and their keyId will be written to `output.json`, example (with mini secret set above)

```
{
  "accounts": {
     '5CJmbXwMxPZD4ku4YCbmB5vFBoU4YJBbxxztLUzez52axNwH': 0,
     '5DwVZ9CHEm97mjVFzGTvtAjxmc2WhYbwhp1fRnShyWNR7rXP': 1,
     '5HB2jdFi9r4Rw935iasTWwGxxBiBtHoKV2e5caYkyH4o4eMX': 2,
     '5FLZWCy347n1czhHV1dUdjjVnk4czKZsBzPuWPnQ98x2sb8o': 3,
     '5Houy1dDJ94dcchpGWvN7PJLJV2gHNZrqAJuh4TEQnziVr5y': 4,
     '5Gbaoto9Prad31VuK2pWLwPZabQbh5UKykpp4jNqNWM2E28W': 5,
      .
      .
      .
  },
  "keyIds": {
    "start": 0,
    "final": 39
  }
}
```

The keys can then be derived with `subkey` for example by inspecting the SURI of the form:

`SURI_MINI_SECRET//testing//KeyId`

so the following account secret key `'5Gbaoto9Prad31VuK2pWLwPZabQbh5UKykpp4jNqNWM2E28W' => 5,` can be derived like so:

```
subkey inspect 0xa9a37c4dd9457728ae16a1b521acde1035395b913a630185e4a6f3e64523de9a//testing//5
Secret Key URI `0xa9a37c4dd9457728ae16a1b521acde1035395b913a630185e4a6f3e64523de9a//testing//5` is account:
  Secret seed:       0x14b34dc9773fcdd0ff08dabf16f9815d5a8a44ae922b64686d86d3f65edafbcb
  Public key (hex):  0xc87dd0f1f4465f77c3c7cf53df50bac868e5986bea99a8299cfd21a0b3319757
  Public key (SS58): 5Gbaoto9Prad31VuK2pWLwPZabQbh5UKykpp4jNqNWM2E28W
  Account ID:        0xc87dd0f1f4465f77c3c7cf53df50bac868e5986bea99a8299cfd21a0b3319757
  SS58 Address:      5Gbaoto9Prad31VuK2pWLwPZabQbh5UKykpp4jNqNWM2E28W
```

The key can be imported similarly into pioneer by providing the raw seed and the path.
Screen Shot 2021-11-08 at 1 49 35 PM

When running a second scenario against the same chain, all they keys can be derived by using the same secret and setting the `START_KEY_ID` to the final one generated by the first scenario.


mnaamani:
The deterministic keys generation is fine, but what will be better pattern is to be able to derive they keys for a member or worker role based on member id and worker id and group such as:

Member controller and root keys:
`//testing/member/{member_id}`

Worker role account:
`//testing/worker/{group name}/{worker_id}`

This will be more convenient for example when starting a storage node where we need to provide the worker id and account uri.

Will push an update.

dobertRowneySr (approved):

mnaamani:
I chose to avoid refactoring fixture too much so instead of creating keypairs from the beginning to follow those SURI patterns, they can be done at the end of the scenario.

The helper api call to change the role account with sudoAs calls can be used on demand and the same technique can be applied to member controller accounts if needed.

dobertRowneySr (approved):

mnaamani:
Merging this after preliminary approval of counterpart for giza #2830
Additional improvements are will be in giza_staging branch

PR #2812, opened 2021-11-03
DevOps - Deploy Argus node using Pulumi

ahhda:
Code to deploy distributor-node (Argus) on a Minikube or EKS cluster

<details>

* With the following configuration (started other services using `yarn start`)
Screenshot 2021-12-22 at 5 21 10 PM

* The deployment works well
Screenshot 2021-12-22 at 5 21 59 PM
</details>

### DIFF
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -1,0 +1,15 @@
+name: argus
+runtime: nodejs
+description: A Pulumi program to deploy Argus node
+template:
+  config:
+    aws:profile:
+      default: joystream-user
+    aws:region:
+      default: us-east-1
+    isLoadBalancerReady:
+      description: Whether the load balancer service is ready and has been assigned an IP
+      default: false
+    queryNodeHost:
+      description: Query node GraphQL endpoint
+      default: 'https://hydra.joystream.org/graphql'
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -16,3 +16,6 @@ template:
     wsProviderEndpointURI:
       description: Chain RPC endpoint
       default: 'wss://rome-rpc-endpoint.joystream.org:9944/'
+    argusImage:
+      description: The distributor node image to use for running the node
+      default: joystream/distributor-node:latest
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -18,13 +18,10 @@ template:
       default: 'wss://rome-rpc-endpoint.joystream.org:9944/'
     argusImage:
       description: The distributor node image to use for running the node
-      default: joystream/distributor-node:latest
     keys:
       description: Specifies the keys available within distributor node CLI
-      default: "[{ 'suri': '//Alice' }]"
     buckets:
       description: Specifies the buckets distributed by the node
       default: 'all'
     workerId:
       description: ID of the node operator (distribution working group worker)
-      default: 0
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -19,3 +19,12 @@ template:
     argusImage:
       description: The distributor node image to use for running the node
       default: joystream/distributor-node:latest
+    keys:
+      description: Specifies the keys available within distributor node CLI
+      default: "[{ 'suri': '//Alice' }]"
+    buckets:
+      description: Specifies the buckets distributed by the node
+      default: 'all'
+    workerId:
+      description: ID of the node operator (distribution working group worker)
+      default: 0
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -13,3 +13,6 @@ template:
     queryNodeHost:
       description: Query node GraphQL endpoint
       default: 'https://hydra.joystream.org/graphql'
+    wsProviderEndpointURI:
+      description: Chain RPC endpoint
+      default: 'wss://rome-rpc-endpoint.joystream.org:9944/'
--- a/devops/kubernetes/argus/Pulumi.yaml
+++ b/devops/kubernetes/argus/Pulumi.yaml
@@ -22,6 +22,14 @@ template:
       description: Specifies the keys available within distributor node CLI
     buckets:
       description: Specifies the buckets distributed by the node
-      default: 'all'
     workerId:
       description: ID of the node operator (distribution working group worker)
+    dataStorage:
+      description: Amount of storage (in Gi) assigned for the data directory
+      default: 10
+    logStorage:
+      description: Amount of storage (in Gi) assigned for the logs directory
+      default: 2
+    cacheStorage:
+      description: Amount of storage (in Gi) assigned for the cache directory
+      default: 10
--- a/devops/kubernetes/argus/index.ts
+++ b/devops/kubernetes/argus/index.ts
@@ -11,28 +11,17 @@ const config = new pulumi.Config()
 
 const queryNodeHost = config.require('queryNodeHost')
 const wsProviderEndpointURI = config.require('wsProviderEndpointURI')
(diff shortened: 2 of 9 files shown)


mnaamani:
There seems to be some important config params missing, like the websocket endpoint. The default is `ws://localhost:99444` which will not work for this deployment.

Why can't we specify a custom docker image like we do for other stacks?
I think we should have a consistent strategy for all stacks. 

After exploring the `ingress` functionality of kubernetes in https://github.com/Joystream/joystream/pull/2820, it is worth looking into possibility of using it in place of our custom caddy approach (if the nginx ingress controller supports dynamically assigning letsencrypt certificates or possibly looking for a custom ingress controller implementation that does?) 

Finally it looks to me like we should factor out the EKS deployment from the scripts and have that be its own stack. And all stack can pass a custom kubeconfig to use as the "provider"? This will also make it more practical to deploy a stack to various cloud providers and not be hard coded to AWS EKS.

mnaamani:
Had trouble with deploying to minikube:

```
Updating (argus):
     Type                              Name                        Status                  Info
 +   pulumi:pulumi:Stack               argus-argus                 **creating failed**     1 error
 +   ├─ pulumi:providers:kubernetes    local                       created                 
 +   ├─ docker:image:Image             joystream/distributor-node  created                 1 warning
 +   ├─ kubernetes:core/v1:Namespace   argus-node                  created                 
 +   ├─ kubernetes:apps/v1:Deployment  argus-node                  **creating failed**     1 error
 +   └─ kubernetes:core/v1:Service     argus-node                  created                 
 
Diagnostics:
  kubernetes:apps/v1:Deployment (argus-node):
    error: 4 errors occurred:
    	* resource argus-node-v5bmtftc/argus-node-5ka8bb44 was successfully created, but the Kubernetes API server reported that it failed to fully initialize or become live: 'argus-node-5ka8bb44' timed out waiting to be Ready
    	* [MinimumReplicasUnavailable] Deployment does not have minimum availability.
    	* Minimum number of live Pods was not attained
    	* [Pod argus-node-v5bmtftc/argus-node-5ka8bb44-8dff85cb6-8jk6w]: containers with unready status: [argus] -- [ImagePullBackOff] Back-off pulling image "joystream/distributor-node:latest"
 
  pulumi:pulumi:Stack (argus-argus):
    error: update failed
```

My guess is because even though the build worked, the policy was still to pull from docker hub because the image tag is `latest`:

```
[ImagePullBackOff] Back-off pulling image "joystream/distributor-node:latest"
```

I modified code to deploy with "joystream/distributor-node:giza" and manually loaded the with `minikube image load` and it succeeded.

mnaamani:
When argus node starts up I'm seeing there multiple storage locations can be configured.
So it would make sense to have these as persisted volumes.

```
{
    "config": {
        ...
        "directories": {
            "assets": "/data",
            "cacheState": "/cache",
            "logs": "/logs"
        },
  ...
}
```

ahhda:
Updates:
- [x]  Specify custom docker image argus
- [x]  Add WS Config param
- [x]  Provide Persistent volumes to directories config
- [x] Add bucket, keys and worker Id config params
- [ ] Factor out the EKS deployment from the scripts and have that be its own stack (This will be a separate PR)

devops/kubernetes/argus/Pulumi.yaml
```diff
@@ -0,0 +1,30 @@
...
+      default: 'https://hydra.joystream.org/graphql'
+    wsProviderEndpointURI:
+      description: Chain RPC endpoint
+      default: 'wss://rome-rpc-endpoint.joystream.org:9944/'
+    argusImage:
+      description: The distributor node image to use for running the node
+      default: joystream/distributor-node:latest
+    keys:
+      description: Specifies the keys available within distributor node CLI
+      default: "[{ 'suri': '//Alice' }]"
```
mnaamani:
```suggestion
      default: "[{ \"suri\": \"//Alice\" }]"
```

Escape `"` to make valid JSON.

As a general point, When is this default value even used? If we create a new stack with `pulumi stack init` we just get a blank with no default config values, so its not like Pulumi.yaml is used as a "template". 

And if there is a default value set in this yaml file, why do we also do `const keys = config.get('keys') || '[{ "suri": "//Alice" }]'` in index.ts effectively overriding it.

devops/kubernetes/argus/index.ts
```diff
@@ -0,0 +1,213 @@
...
+import { CaddyServiceDeployment } from 'pulumi-common'
+
+const awsConfig = new pulumi.Config('aws')
+const config = new pulumi.Config()
+
+const queryNodeHost = config.require('queryNodeHost')
+const wsProviderEndpointURI = config.require('wsProviderEndpointURI')
+let configArgusImage = config.require('argusImage')
+const lbReady = config.get('isLoadBalancerReady') === 'true'
+const keys = config.get('keys') || '[{ "suri": "//Alice" }]'
```
mnaamani: I think keys, and worker id should be required instead of using these faults. Better to be explicit.

devops/kubernetes/argus/index.ts
```diff
@@ -0,0 +1,213 @@
...
+                  value: keys,
+                },
+                {
+                  name: 'JOYSTREAM_DISTRIBUTOR__BUCKETS',
+                  value: buckets,
+                },
+                {
+                  name: 'JOYSTREAM_DISTRIBUTOR__WORKER_ID',
+                  value: workerId,
+                },
```
mnaamani, line 133:
```suggestion
                },
                {
                  name: 'JOYSTREAM_DISTRIBUTOR__PORT',
                  value: 3334,
                },
```

Better to be explicit just in-case argus implementation changes default port.

mnaamani (changes requested):
Works like a charm. Left a couple minor points.

devops/kubernetes/argus/Pulumi.yaml
ahhda: The default value in `Pulumi.yaml` is more of a reference guide for a user to see what is the default value for a parameter so that they don't have to dive into the code.

devops/kubernetes/argus/index.ts
ahhda: Done

devops/kubernetes/argus/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    metadata: {
+      labels: appLabels,
+      namespace: namespaceName,
+      name: `${name}-pvc`,
+    },
+    spec: {
+      accessModes: ['ReadWriteOnce'],
+      resources: {
+        requests: {
+          storage: `10Gi`,
```
mnaamani: I think each volume size should be separately configurable. we don't need 10G for logs for example.

devops/kubernetes/argus/index.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const awsConfig = new pulumi.Config('aws')
+const config = new pulumi.Config()
+
+const queryNodeHost = config.require('queryNodeHost')
+const wsProviderEndpointURI = config.require('wsProviderEndpointURI')
+let configArgusImage = config.require('argusImage')
+const lbReady = config.get('isLoadBalancerReady') === 'true'
+const keys = config.require('keys')
+const buckets = config.get('buckets') || 'all'
```
mnaamani: the fallback value `all` is invalid, perhaps make the config value required instead and don't provide a default/fallback value.

devops/kubernetes/argus/index.ts
```diff
@@ -0,0 +1,218 @@
...
+import * as k8s from '@pulumi/kubernetes'
+import * as pulumi from '@pulumi/pulumi'
+import { CaddyServiceDeployment } from 'pulumi-common'
+
+const awsConfig = new pulumi.Config('aws')
+const config = new pulumi.Config()
+
+const queryNodeHost = config.require('queryNodeHost')
+const wsProviderEndpointURI = config.require('wsProviderEndpointURI')
+let configArgusImage = config.require('argusImage')
```
mnaamani:
```suggestion
const configArgusImage = config.require('argusImage')
```

mnaamani (changes requested):

PR #2814, opened 2021-11-03
Fix pioneer linter issue

mnaamani:
In https://github.com/Joystream/joystream/pull/2804 I broke the [linter](https://github.com/Joystream/joystream/runs/4081599274?check_suite_focus=true) for pioneer. After refactoring of WorkingGroup enum it was no longer necessary to do a type assertion where it was used in the proposals app.


Lezek123 (approved):
LGTM! (I was about to mention it in https://github.com/Joystream/joystream/pull/2804)

Issue #2577, opened 2021-08-12
Temporary content directory god mode


bedeho:
Both for populating the content after redeploying the storage system, such as in `Giza` network launch, but also to support planned marketing initiatives where we can have auto-synching with YouTube for creator with a pre-existing presence, we need a way to use SUDO to

- create a channel for a member
- update a channel
- publish a video in a channel
- update a video
- delete a video

This is functionality which is temporary before mainnet, so we can just introduce ability to use SUDO for these extrinsics

bedeho:
I think there may be a more useful way to solve this which also aligns with some long features the creators likely will have.

Allow a channel owner to have up to a fixed number of other members as _collaborators_ on a channel, meaning that they are allowed to do anything the owner can do. In the future we could add permissions, but lets start simple. This is a very realistic feature, as people who own channels may want to delegate part of the work.

With this feature, we can do the bootstrapping and marketing stuff we need by just having a designated `Jsgenesis` member setup as a collaborator during channel creation. There is no need for a SUDO privilege. Even the channel creation, which is done by a member (not a channel owner or collaborator, like the others) does not require SUDO, we can just create the member at our own expense with our account, with the collaborator set, and then we change the controller account of the member to the provided account from the new user.

dobertRowneySr:
The collaborator set is a `BTreeSet<MemberId>`, that can be created and update together with the channel by providing a the set during creation/update (so that is responsibility of the caller to ensure that the set is valid).
This scheme is the one I used on the previous PRs.
As far as overlapping between channel owner and collaborator goes I have found that only video operation {update, create, delete} needs collaboration so far. Since:
- playlist extrinsics are yet to be implemented 
- same thing with `-person` extrinsics
- members are restricted from censoring and managing categories
- feature videos can only be set by Lead

I had to split up the `actor_authorized_to_update_channel` into subchecks in order to reuse them

bedeho:
Excellent work, but what about updating the channel tiself? Like changing its metadata, like the cover photo or title etc.  Should this not be possible?

PR #2817, opened 2021-11-04
Channel collab rebase

dobertRowneySr:
Rebase of an old commit issue (https://github.com/Joystream/joystream/issues/2577)
- Collaborators can add/remove/update video & assets to channels
- Currently collaborators dispose of `deletion_prize_source_account_id` (i.e. channel's owner account) for funding storage operations
- Channel owner can add himself as a collaborator
- added features are basically the same as the old PR, I just refactored some of the tests


Lezek123:
I suggest to include 2 small adjustments in this PR:
- Update `chain_metadata.json` by executing `yarn update-chain-metadata`
- Update `@joystream/types` (add `collaborators` to `Channel`, `ChannelCreationParametersRecord`, `ChannelUpdateParametersRecord` and re-build)

runtime-modules/content/src/lib.rs
```diff
@@ -206,36 +208,45 @@ pub type ChannelOwnershipTransferRequest<T> = ChannelOwnershipTransferRequestRec
...
-pub struct ChannelUpdateParametersRecord<StorageAssets, AccountId, DataObjectId: Ord> {
+pub struct ChannelUpdateParametersRecord<StorageAssets, AccountId, DataObjectId: Ord, MemberId: Ord>
+{
     /// Asset collection for the channel, referenced by metadata    
     assets_to_upload: Option<StorageAssets>,
     /// If set, metadata update for the channel.
     new_meta: Option<Vec<u8>>,
     /// If set, updates the reward account of the channel
-    reward_account: Option<Option<AccountId>>,
+    reward_account: Option<AccountId>,
```
Lezek123: `Option<Option<AccountId>>` allowed updating reward account to `none`, which I think will not be possible now?

dobertRowneySr: right, thanks, good call

types/src/content/index.ts
```diff
@@ -28,6 +28,7 @@ export class CuratorGroup extends JoyStructDecorated({
 }) {}
 
 export class ContentActor extends JoyEnum({
+  Collaborator: MemberId,
```
Lezek123: The variant ordering must match the one in the runtime, so:
`Curator` -> `Member` -> `Collaborator` -> `Lead`

I was also wondering if it wouldn't be better have `Collaborator` as last variant in both the runtime and `@joystream/types` (`Curator` -> `Member` -> `Lead` -> `Collaborator`) just in case for extra safety and backward-compatibility (the main rule is we shouldn't change ordering of enum varaints if the enum is stored somewhere in the runtime state, although it may not be the case in this specific example)

runtime-modules/content/src/lib.rs
```diff
@@ -708,22 +720,29 @@ decl_module! {
...
-            // == MUTATION SAFE ==
-            //
-
             let mut channel = channel;
 
-            // Maybe update the reward account
-            if let Some(reward_account) = &params.reward_account {
+            // maybe update the reward account if actor is not a collaborator
+            if let Some(reward_account) = params.reward_account.as_ref() {
+                ensure_actor_not_a_collaborator::<T>(&actor)?;
```
Lezek123: This check should be done before calls to `Self::remove_assets_from_storage` and `Self::upload_assets_to_storage`.

When I tried to update both channel avatar and `reward_account` as `Collaborator` I noticed that:
- the old avatar is removed
- the new avatar is "uploaded", but it's not assigned to the channel by the query node, because `ChannelUpdated` event is never emitted
- extrinsic fails with `ActorNotAuthorized`, but the runtime state is already changed
- this results in channel having no assigned avatar and a dangling data object of `DataObjectTypeUnknown` can be seen in query node

Events emitted as a result of the call:
```
balances.Transfer
storage.BagObjectsChanged
storage.VoucherChanged
storage.DataObjectsDeleted
balances.Transfer
storage.BagObjectsChanged
storage.VoucherChanged
storage.DataObjectsUploaded
system.ExtrinsicFailed
```

Please make sure no state changes are done before all of the extrinsic parameters are validated. (From what I understand that's why all methods have the `// == MUTATION SAFE ==` comment)

Lezek123:
Another thing I wanted to suggest is to add a check on the runtime side whether the collaborators provided as part of `ChannelCreationParameters` or `ChannelUpdateParameters` are actually valid (existing) members.

This would have two benefits:
- Would prevent the caller from accidently adding a non-existing member as collaborator
- Would allow us to store `collaborators` as references to `Membership` entity on the query-node side (instead of storing just the ids as array of `Int`s/`BigInt`s/`String`s)

dobertRowneySr:
> Another thing I wanted to suggest is to add a check on the runtime side whether the collaborators provided as part of `ChannelCreationParameters` or `ChannelUpdateParameters` are actually valid (existing) members.
> 
> This would have two benefits:
> 
> * Would prevent the caller from accidently adding a non-existing member as collaborator
> * Would allow us to store `collaborators` as references to `Membership` entity on the query-node side (instead of storing just the ids as array of `Int`s/`BigInt`s/`String`s)

I think this should be a good idea too, @bedeho can you weight in?

types/src/content/index.ts
```diff
@@ -1,7 +1,7 @@
 import { Vec, Option, Tuple, BTreeSet } from '@polkadot/types'
 import { bool, u64, u32, u128, Null, Bytes } from '@polkadot/types/primitive'
 import { MemberId } from '../members'
-import { JoyStructDecorated, JoyEnum, ChannelId } from '../common'
```
Lezek123, line 4: I think line this was accidently removed

bedeho:
> > Another thing I wanted to suggest is to add a check on the runtime side whether the collaborators provided as part of `ChannelCreationParameters` or `ChannelUpdateParameters` are actually valid (existing) members.
> > This would have two benefits:
> > 
> > * Would prevent the caller from accidently adding a non-existing member as collaborator
> > * Would allow us to store `collaborators` as references to `Membership` entity on the query-node side (instead of storing just the ids as array of `Int`s/`BigInt`s/`String`s)
> 
> I think this should be a good idea too, @bedeho can you weight in?

Yep, great idea ;)

bedeho:
1. When should I review this?
2. Perhaps it's just escaping my memory, but why do we need `Channel::deletion_prize_source_account_id` ? From what I gather, this is where funds will be drawn from when uploading to finance deletion prize? But why can't this just be the controller account of the calling member?

runtime-modules/content/src/lib.rs
```diff
@@ -656,22 +669,20 @@ decl_module! {
...
 
-            //
-            // == MUTATION SAFE ==
-            //
-
             // Only increment next channel id if adding content was successful
             NextChannelId::<T>::mutate(|id| *id += T::ChannelId::one());
 
+           // ensure collaborator member ids are valid
+            Self::validate_collaborator_set(&params.collaborators)?;
```
Lezek123: I think this should be done before any state mutation.

Lezek123 (changes requested):

bedeho:
I think @shamil-gadelshin is better positioned to review this, and he is back next Monday, so assigned to him.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -213,7 +259,7 @@ pub fn ensure_actor_authorized_to_censor<T: Trait>(
                 Ok(())
             }
         }
-        ContentActor::Member(_) => {
+        _ => {
```
shamil-gadelshin, line 360: Obsolete comment

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -236,7 +282,7 @@ pub fn ensure_actor_authorized_to_manage_categories<T: Trait>(
             // Authorize curator, performing all checks to ensure curator can act
             CuratorGroup::<T>::perform_curator_in_group_auth(curator_id, curator_group_id, &sender)
         }
-        ContentActor::Member(_) => {
+        _ => {
```
shamil-gadelshin, line 384: Obsolete comment

runtime-modules/content/src/tests/mock.rs
```diff
@@ -1,7 +1,7 @@
 #![cfg(test)]
 
+use crate::sp_api_hidden_includes_decl_storage::hidden_include::traits::Currency;
```
shamil-gadelshin: It looks like an auto-added line. `hidden_include` is non-documented export. We should use `Currency` trait from the frame_support crate.

runtime-modules/storage/src/lib.rs
```diff
@@ -675,7 +675,7 @@ pub struct UploadParametersRecord<MemberId, ChannelId, AccountId, Balance> {
     /// Account for the data object deletion prize.
     pub deletion_prize_source_account_id: AccountId,
 
-    /// Expected data size fee value for this extrinsic call.
+    ///a Expected data size fee value for this extrinsic call.
```
shamil-gadelshin: Typo?

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -115,61 +116,106 @@ pub fn ensure_actor_authorized_to_create_channel<T: Trait>(
...
             ensure!(
-                *owner == ChannelOwner::Member(*member_id),
+                channel.collaborators.contains(member_id),
                 Error::<T>::ActorNotAuthorized
             );
-
             Ok(())
         }
     }
 }
```
shamil-gadelshin, line 279: Please, add a comment for the public method.

runtime-modules/content/src/lib.rs
```diff
@@ -1437,6 +1477,7 @@ decl_event!(
         VideoCreationParameters = VideoCreationParameters<T>,
         VideoUpdateParameters = VideoUpdateParameters<T>,
         StorageAssets = StorageAssets<T>,
+        //        CollaboratorsSetType = BTreeSet<<T as MembershipTypes>::MemberId>,
```
shamil-gadelshin: Remove commented code, please.

runtime-modules/content/src/lib.rs
```diff
@@ -1407,6 +1438,15 @@ impl<T: Trait> Module<T> {
         }
         Ok(())
     }
+
+    fn validate_collaborator_set(collaborators: &BTreeSet<T::MemberId>) -> DispatchResult {
+        // check if all members are valid
+        let res = collaborators.iter().fold(true, |acc, member_id| {
+            acc && <T as ContentActorAuthenticator>::validate_member_id(member_id)
```
shamil-gadelshin: Isn't `any` function simpler than `fold` in this case?

runtime-modules/content/src/lib.rs
```diff
@@ -688,42 +700,56 @@ decl_module! {
...
 
             ensure_actor_authorized_to_update_channel::<T>(
                 origin,
                 &actor,
-                &channel.owner,
+                &channel,
             )?;
 
-            Self::remove_assets_from_storage(&params.assets_to_remove, &channel_id, &channel.deletion_prize_source_account_id)?;
+            let mut channel = channel;
```
shamil-gadelshin: What is the purpose of the shadowing here? Why don't add 'mut' on initialization?

shamil-gadelshin (changes requested):
Looks ok. Found several minor issues.

Please, add comments to the public methods and investigate the Github actions failed build ( my local machine has no issues during the testing).

runtime-modules/content/src/tests/mock.rs
```diff
@@ -597,3 +615,55 @@ pub fn update_video_mock(
...
 }
+
+pub fn delete_video_mock(
+    sender: u64,
+    actor: ContentActor<CuratorGroupId, CuratorId, MemberId>,
+    video_id: <Test as Trait>::VideoId,
+    assets_to_remove: BTreeSet<DataObjectId<Test>>,
+    result: DispatchResult,
+) {
+    // let channel_id = Content::video_by_id(video_id.clone()).in_channel;
```
shamil-gadelshin: Please, remove commented code.

runtime-modules/content/src/lib.rs
```diff
@@ -733,14 +765,15 @@ decl_module! {
             channel_id: T::ChannelId,
             num_objects_to_delete: u64,
         ) -> DispatchResult {
+
+            let sender = ensure_signed(origin.clone())?;
```
shamil-gadelshin: Why do we clone origin here?

runtime-modules/content/src/lib.rs
```diff
@@ -688,42 +709,53 @@ decl_module! {
             channel_id: T::ChannelId,
             params: ChannelUpdateParameters<T>,
         ) {
+            let sender = ensure_signed(origin.clone())?;
```
shamil-gadelshin: Why do we clone origin here?

runtime-modules/content/src/lib.rs
```diff
@@ -1407,6 +1444,15 @@ impl<T: Trait> Module<T> {
         }
         Ok(())
     }
+
+    fn validate_collaborator_set(collaborators: &BTreeSet<T::MemberId>) -> DispatchResult {
+        // check if all members are valid
+        let res = !collaborators
```
shamil-gadelshin: You're using `!any(!)`. Isn't it similar to `all` ?

shamil-gadelshin (changes requested):
Just a couple more issues.

Please, check all unnecessary `origin.clone()` entries.

shamil-gadelshin (approved):
LGTM

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -1,3 +1,20 @@
...
+* The following table summarizes the permissions in the content subsystem.
+* - Actor role as columns, controller account is Tx sender.
+* - operations on a given channel (=channel=) are rows, which are basically the guards to be
+*   implemented
+* - Entries are conditions to be verified / assertions
+*
+* |                       | *Lead*                   | *Curator*                | *Member*                | *Collaborator*                   |
+* |-----------------------+--------------------------+--------------------------+-------------------------+----------------------------------|
+* | *assets mgmt*         | channel.owner is curator | curator is channel.owner | member is channel.owner | collaborators in channel.collabs |
+* | *censorship mgmt*     | channel.owner is curator | curator is channel.owner | false                   | false                            |
```
Lezek123: This is incorrect, the lead can censor all channels, curator can only censor member channels:
```suggestion
* | *censorship mgmt*     | true                     | channel.owner is member  | false                   | false                            |
```

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -1,3 +1,20 @@
...
+* | *category mgmt*       | true                     | true                     | false                   | false                            |
+* | *collab. set mgmt*    | channel.owner is curator | curator is channel.owner | member is channel.owner | false                            |
+* | *reward account mgmt* | false                    | curator is channel.owner | member is channel.owner | false                            |
+* | *create channel*      | false                    | true                     | true                    | false                            |
+* | *delete channel*      | false                    | curator is channel.owner | member is channel.owner | false                            |
```
Lezek123: I'm not sure why the lead is not authorized to delete channel or update reward account using `Lead` context, while it is authorized to update assets or collaborators using this context. Previously I believe the lead could act as owner of any curator channel (which I think made sense, as (s)he is the one managing the curator groups

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -91,86 +111,196 @@ pub fn ensure_lead_auth_success<T: Trait>(account_id: &T::AccountId) -> Dispatch
...
-            let sender = ensure_signed(origin)?;
+        ContentActor::Member(member_id) => ensure_member_auth_success::<T>(sender, member_id),
+        // Lead & collaborators should use their member or curator role to create channels.
+        _ => Err(Error::<T>::ActorCannotOwnChannel.into()),
+    }
+}
 
-            ensure_member_auth_success::<T>(member_id, &sender)
+/// Ensure actor is authorized to delete channel
+pub fn ensure_actor_authorized_to_delete_channel<T: Trait>(
```
Lezek123, line 130: This function is never used, instead `ensure_actor_authorized_to_update_channel_assets` is used during `delete_channel`

Lezek123 (changes requested):

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -152,7 +158,7 @@ pub fn ensure_actor_authorized_to_delete_channel<T: Trait>(
             ensure_member_is_channel_owner::<T>(channel_owner, member_id)?;
             Ok(())
         }
-        // Lead & collaborators should use their member or curator role in order to update reward account.
+        // collaborators should use their member or curator role in order to update reward account.
```
Lezek123: The comment is about reward account, but this is `ensure_actor_authorized_to_delete_channel` function

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -1,20 +1,19 @@
...
+//   implemented
+// - Entries are conditions to be verified / assertions
+//
+// |                       | *Lead*                   | *Curator*                | *Member*                | *Collaborator*            |
+// |-----------------------+--------------------------+--------------------------+-------------------------+---------------------------|
+// | *assets mgmt*         | channel.owner is curator | curator is channel.owner | member is channel.owner | collab in channel.collabs |
+// | *censorship mgmt*     | true                     | channel.owner is member  | false                   | false                     |
+// | *category mgmt*       | true                     | true                     | false                   | false                     |
+// | *collab. set mgmt*    | channel.owner is curator | curator is channel.owner | member is channel.owner | false                     |
+// | *reward account mgmt* | channel.owner is curator | curator is channel.owner | member is channel.owner | false                     |
```
Lezek123, line 13: You need to update `ensure_actor_can_manage_reward_account` to reflect this change (allow `Lead` context)

Lezek123 (changes requested):

Lezek123 (approved):
LGTM

PR #2829, opened 2021-11-12
DevOps - Update Colossus deployment for Giza

ahhda:
* Remove docker build within deployment
* Add additional parameters like `queryNodeHost`, `workerId`, `accountURI`, `colossusImage`
* Refactor generic `configMap` to `pulumi-common`

<details>

* Running stack with these configurations
Screenshot 2021-12-23 at 6 26 43 PM

* The node runs without any issues
Screenshot 2021-12-23 at 6 26 38 PM

</details>

### DIFF
--- a/devops/kubernetes/storage-node/index.ts
+++ b/devops/kubernetes/storage-node/index.ts
@@ -13,7 +13,7 @@ const config = new pulumi.Config()
 const name = 'storage-node'
 
 const wsProviderEndpointURI = config.require('wsProviderEndpointURI')
-const queryNodeHost = config.require('queryNodeHost')
+const queryNodeHost = config.require('queryNodeEndpoint')
 const workerId = config.require('workerId')
 const accountURI = config.get('accountURI')
 const keyFile = config.get('keyFile')
@@ -166,7 +166,7 @@ const deployment = new k8s.apps.v1.Deployment(
                   value: `${colossusPort}`,
                 },
                 {
-                  name: 'QUERY_NODE_HOST',
+                  name: 'QUERY_NODE_ENDPOINT',
                   value: queryNodeHost,
                 },
                 {
@@ -198,7 +198,8 @@ const deployment = new k8s.apps.v1.Deployment(
                 'server',
                 '--worker',
                 workerId,
-                '--port=3333',
+                '--port',
+                `${colossusPort}`,
                 '--uploads=/data',
                 '--sync',
                 '--syncInterval=1',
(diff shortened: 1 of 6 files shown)


devops/kubernetes/storage-node/index.ts
```diff
@@ -96,74 +89,22 @@ const pvc = new k8s.core.v1.PersistentVolumeClaim(
...
+).configName
 
-    volumes.push({
-      name: 'keyfile-volume',
-      configMap: {
-        name: keyConfigName,
-      },
-    })
+const remoteKeyFilePath = '/joystream/key-file.json'
+additionalParams = ['--key-file', remoteKeyFilePath]
```
mnaamani: keyfile is required above but additionalParams are not used, since we are passing ACCOUNT_URI.

So we need better handling I think here of keyFile, passphrase, and accountUri configs, especially if both are set, which one takes precedence (based on storage-node implementation of course)

mnaamani (changes requested):
Deployment works well. Handling of keyFile/passphrase and accountUri configs needs improvement.

devops/kubernetes/storage-node/index.ts
ahhda: So according to the method `getAccount` in `storage-node-v2/src/command-base/ApiCommandBase.ts` if `accountURI` is specified it takes precedence and overrides `password` and `keyFile` flags.

Also either `Keyfile` or `accountURI` must be set otherwise we throw an error.

Will update this in the code.

devops/kubernetes/storage-node/index.ts
```diff
@@ -217,21 +161,65 @@ const deployment = new k8s.apps.v1.Deployment(
...
+                },
+                ...additionalVolumeMounts,
+              ],
+              command: ['yarn'],
+              args: [
+                'storage-node',
+                'server',
+                '--worker',
+                workerId,
+                '--port=3333',
```
mnaamani: Shouldn't this be the the pulumi configuration parameter 'colossusPort' instead of hard-coded 3333?

mnaamani (changes requested):
Minor issue with the port configuration otherwise looks good.

devops/kubernetes/storage-node/index.ts
```diff
@@ -217,21 +161,65 @@ const deployment = new k8s.apps.v1.Deployment(
                   name: 'DEBUG',
                   value: 'joystream:*',
                 },
+                {
+                  name: 'COLOSSUS_PORT',
+                  value: `${colossusPort}`,
+                },
+                {
+                  name: 'QUERY_NODE_HOST',
```
mnaamani: this has now changed to QUERY_NODE_ENDPOINT and is a full Url not just the host-name

PR #2830, opened 2021-11-13
Tests: Giza setup new chain scripts

mnaamani:
Counterpart to https://github.com/Joystream/joystream/pull/2809 but for giza with some additional refactoring and code cleanup.

Refactored docker-compose to not require building docker images for query-node, colossus, argus, pioneer.
docker-compose has two services instances of colossus and argus with well known keys.

I substituted blake3 for blake3-wasm (used by cli, colossus and argus) to be able to run docker containers, otherwise the docker image for `node:14` with mounted bind volume, is not bale to execute the native npm module due to differing cpu architecture. I couldn't figure out why docker wasn't pulling an image that matches the host arch (on mac). Maybe we can find a solution before release. (For the cli its probably the right choice to make the cli more portable and not require user have build tools to install?)

Had to drop `chalk` from root package.json resolutions to get distributor-node to work again, but was able to add it again after adding `@oclif/plugin-help` dependency.

Updated yarn.lock (for chalk and blake3 related package changes)


.env
```diff
@@ -9,39 +9,49 @@ INDEXER_DB_NAME=query_node_indexer
...
-# WS_PROVIDER_ENDPOINT_URI=ws://host.docker.internal:9944/
-
-######################
-#    Storage Node    #
-######################
-COLOSSUS_PORT=3333
-QUERY_NODE_HOST=${GRAPHQL_SERVER_HOST}:${GRAPHQL_SERVER_PORT}
-WORKER_ID=0
-ACCOUNT_URI=//Alice
+# GraphQL ports (WARTHOG_APP_HOST overrides GRAPHQL_SERVER_PORT)
```
Lezek123:
```suggestion
# GraphQL ports (WARTHOG_APP_PORT overrides GRAPHQL_SERVER_PORT)
```

docker-compose.yml
```diff
@@ -18,48 +18,121 @@ services:
...
-      - graphql-server
     volumes:
       - /data
       - /cache
+      - /logs
+      - type: bind
+        source: .
+        target: /joystream
+    # let the working_dir be the distributor node to pickup the config.yml file
+    working_dir: /joystream/distributor-node
```
Lezek123, line 60: Previously I was using the config file from `distributor-node/config/docker/config.docker.yml` by providing:
`ENV CONFIG_PATH ./distributor-node/config/docker/config.docker.yml`
in `distributor-node.Dockerfile`

One of the differences that is not reflected in the current env setting is the `directories` config.
You can either set:
```
JOYSTREAM_DISTRIBUTOR__DIRECTORIES__ASSETS: /data
JOYSTREAM_DISTRIBUTOR__DIRECTORIES__CACHE_STATE: /cache
JOYSTREAM_DISTRIBUTOR__DIRECTORIES__LOGS: /logs
```

Or point to the config file in `distributor-node/config/docker/config.docker.yml`.

Personally I think it's a good idea to just use env variables and remove `distributor-node/config/docker/config.docker.yml` so we don't need to maintain two separate config files.

.env
```diff
@@ -9,39 +9,49 @@ INDEXER_DB_NAME=query_node_indexer
...
+# So keep them the safe to avoid confusion, as long as each service is
+# run in a separate container on same host then there is no issue.
+# It does not make sense to have both but hydra codegen demands them.
+GRAPHQL_SERVER_PORT=4001
+GRAPHQL_SERVER_HOST=localhost
+WARTHOG_APP_PORT=${GRAPHQL_SERVER_PORT}
+WARTHOG_APP_HOST=localhost
```
Lezek123: Since both `hydra-indexer-gateway` and `graphql-server` are relying on `WARTHOG_APP_PORT`, perhaps we could simplify the config a little and use:
```
GRAPHQL_SERVER_PORT=8081
HYDRA_INDEXER_GATEWAY_PORT=4000
```

And then set `WARTHOG_APP_PORT` per-service in the `docker-compose.yml`:
```
  hydra-indexer-gateway:
    # ...
    environment:
      # ...
      - WARTHOG_APP_PORT=${HYDRA_INDEXER_GATEWAY_PORT}
      - PORT=${HYDRA_INDEXER_GATEWAY_PORT}
    ports:
      - '127.0.0.1:${HYDRA_INDEXER_GATEWAY_PORT}:${HYDRA_INDEXER_GATEWAY_PORT}'
  graphql-server:
    environment:
      - WARTHOG_APP_PORT=${GRAPHQL_SERVER_PORT}
      - PORT=${GRAPHQL_SERVER_PORT}
    ports:
      - '127.0.0.1:${GRAPHQL_SERVER_PORT}:${GRAPHQL_SERVER_PORT}'
```
It would also make it clearer which port/host are we referring to, for example:
```
DISTRIBUTOR_QUERY_NODE_URL=http://graphql-server:${GRAPHQL_SERVER_PORT}/graphql
PROCESSOR_INDEXER_GATEWAY=http://hydra-indexer-gateway:${HYDRA_INDEXER_GATEWAY_PORT}/graphql
```

I created a PR to your PR ilustrating this change: https://github.com/mnaamani/joystream/pull/10

tests/network-tests/src/Api.ts
```diff
@@ -117,29 +159,39 @@ export class Api {
...
+
+  public createCustomKeyPair(path: string): KeyringPair {
+    return this.factory.createCustomKeyPair(path)
+  }
+
+  public keyGenInfo(): { start: number; final: number } {
+    return this.factory.keyGenInfo()
+  }
+
+  public getAllgeneratedAccounts(): { [k: string]: number } {
```
Lezek123:
```suggestion
  public getAllGeneratedAccounts(): { [k: string]: number } {
```

tests/network-tests/src/Api.ts
```diff
@@ -1710,4 +1787,134 @@ export class Api {
...
+    account: string
+  ): Promise<ISubmittableResult> {
+    if (!(await this.isWorker(workerId, group))) {
+      throw new Error('Worker not found')
+    }
+    const worker = await this.getWorkerById(workerId, group)
+
+    const memberController = await this.getMemberControllerAccount(worker.member_id.toNumber())
+    // there cannot be a worker associated with member that does not exist
+    assert(memberController, 'Member controller not found')
```
Lezek123: You could use `if (!memberController) throw` to avoid using non-null assetion (`memberController!`) later.
There is also a way to make [typesafe `assert` function](https://github.com/Joystream/joystream/blob/olympia/tests/integration-tests/src/utils.ts#L102).

tests/network-tests/src/Scenario.ts
```diff
@@ -9,13 +9,32 @@ import { Job } from './Job'
...
   env: NodeJS.ProcessEnv
   debug: Debugger.Debugger
   job: (label: string, flows: Flow[] | Flow) => Job
 }
 
+function writeOutput(api: Api, miniSecret: string) {
+  const outputFilename = 'output.json'
+  console.error('Writing generated account to', outputFilename)
+  // account to key ids
+  const accounts = api.getAllgeneratedAccounts()
```
Lezek123:
```suggestion
  const accounts = api.getAllGeneratedAccounts()
```

tests/network-tests/src/fixtures/membershipModule.ts
```diff
@@ -60,10 +63,12 @@ export class BuyMembershipWithInsufficienFundsFixture extends BaseFixture {
...
-    // Assertions
-    const membership = await this.api.getMemberIds(this.account)
-
-    assert(membership.length === 0, 'Account must not be associated with a member')
+    try {
+      await this.api.getMemberId(this.account)
+      assert(false, 'Account must not be associated with a member')
+    } catch (err) {
+      // member id not found
+    }
```
Lezek123: This will always end up in `catch` because of `assert(false)`.

tests/network-tests/src/Api.ts
```diff
@@ -161,8 +218,18 @@ export class Api {
...
-    return this.api.query.members.memberIdsByControllerAccountId<Vec<MemberId>>(address)
+  // Many calls in the testing framework take an account id instead of a member id when an action
+  // is intended to be in the context of the member. This function is used to do a reverse lookup.
+  // There is an underlying assumption that each member has a unique controller account even
+  // though the runtime does not place that constraint. But for the purpose of the tests we throw
+  // if that condition is found to be false to esnure the tests do not fail. As long as all memberships
+  // are created through the membership fixture this should not happen.
+  public async getMemberId(address: string): Promise<MemberId> {
+    const ids = await this.api.query.members.memberIdsByControllerAccountId<Vec<MemberId>>(address)
+    if (ids.length > 1) {
```
Lezek123: Should we also handle `ids.length === 0` case?

tests/network-tests/src/flows/council/assign.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  debug('Assigning new council of size', councilSize)
+
+  const council = []
+
+  for (let i = 0; i < councilSize; i++) {
+    council.push(api.createCustomKeyPair(`CouncilMember//${i}`).address)
+  }
+
+  const paidTerms: PaidTermId = api.createPaidTermId(new BN(+env.MEMBERSHIP_PAID_TERMS!))
```
Lezek123, line 39: It would be nice to avoid non-null assertions

tests/network-tests/src/flows/workingGroup/manageWorkerAsWorker.ts
```diff
@@ -17,10 +18,10 @@ import { extendDebug } from '../../Debugger'
 
 export default {
   storage: async function ({ api, env }: FlowProps): Promise<void> {
-    return manageWorkerAsWorker(api, env, WorkingGroups.StorageWorkingGroup)
+    return manageWorkerAsWorker(api, env, WorkingGroups.Storage)
```
Lezek123, line 21: Shouldn't the default export be changed here to match with `manageWorkerAsLead.ts`? (a function that takes group as arg)

tests/network-tests/src/flows/workingGroup/workerPayout.ts
```diff
@@ -23,10 +24,10 @@ import { Resource, ResourceLocker } from '../../Resources'
 
 export default {
   storage: async function ({ api, env, lock }: FlowProps): Promise<void> {
```
Lezek123, line 26: Shouldn't the default export be changed here to match with `manageWorkerAsLead.ts`? (a function that takes group as arg)

tests/network-tests/src/misc/mockContentFlow.ts
```diff
@@ -0,0 +1,68 @@
...
+  const numberOfRoundsVideo = 100
+
+  const channelIds: number[] = []
+
+  // create mock channels
+  debug('Creating Channels')
+  for (let n = 0; n < numberOfRoundsChannel; n++) {
+    const createChannels = new CreateChannelsAsMemberFixture(api, memberId, numberOfChannelsPerRound)
+    await new FixtureRunner(createChannels).run()
+    createChannels.getCreatedChannels().forEach((id) => channelIds.push(id!.toNumber()))
```
Lezek123: It would be nice to avoid non-null assertions

.env
```diff
@@ -9,39 +9,49 @@ INDEXER_DB_NAME=query_node_indexer
 DB_NAME=query_node_processor
 DB_USER=postgres
 DB_PASS=postgres
-DB_HOST=db
+DB_HOST=localhost
```
Lezek123: I think we should actually stick with `db`, because otherwise the value in `/query-node/generated/graphql-server/.env` will vary between `db` and `localhost` during query node startup process, which seems quite fragile.

Lezek123 (changes requested):
The improvements are great and very useful!
I suggested a few minor changes.

Some additional observations:
- Needed to update `docker-compose` from `1.27.4` to `1.29.2` locally, otherwise the scripts like `query-node/start.sh` or `tests/network-tests/run-tests.sh` were not working without explicitly set environment (through `set -a . path/to/.env set +a`)
- The migration tests need to be fixed. Previously they relied on the assumption that `@joystream/cli@0.5.1` is installed globally, but this wasn't executed as step in the CI / script. There is also an issue with the app data dir not beeing accessible in Github CI (https://github.com/Joystream/joystream/runs/4235401998?check_suite_focus=true#step:8:51)

.env
Lezek123: **EDIT:** Actually nevermind this. The reason why this was occuring for me was because I:
1. Changed `DB_HOST=db` in `.env`
2. Removed `- DB_HOST=db` under `graphql-server` environment in `docker-compose.yml` as it was no longer necessary
3. Changed `DB_HOST` back to `localhost` in `.env` for some testing, but didn't add back the config removed in step `2.`, which resulted in the issue mentioned in the comment above.

I think both approaches are valid then.

ahhda (approved):

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,111 +4,65 @@ set -e
...
-    docker rm ${CONTAINER_ID}
-    rm tests/network-tests/assets/TestChannel__rejectedContent.json
-    rm tests/network-tests/assets/TestVideo__rejectedContent.json
-    
+    docker-compose -f ../../docker-compose.yml down -v
+    rm ./assets/TestChannel__rejectedContent.json || true
+    rm ./assets/TestVideo__rejectedContent.json || true
 }
 
 function pre_migration_hook() {
```
dobertRowneySr, line 29: I wrote this part before I knew how to setup pre/post scenario on the typescript test framework that we currently use.
Shouldn't this part be redundant? (Since you already create fixtures & flows in order to perform exactly these operations)

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,111 +4,65 @@ set -e
...
+  yarn workspace api-scripts tsnode-strict src/status.ts | grep Runtime
+
+  yarn joystream-cli account:choose --address 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY # Alice
+  echo "creating 1 channel"
+  yarn joystream-cli content:createChannel --input=./assets/TestChannel.json --context=Member || true
+  echo "adding 1 video to the above channel"
+  yarn joystream-cli content:createVideo -c 1 --input=./assets/TestVideo.json || true
 }
 
 function post_migration_hook() {
```
dobertRowneySr, line 49: Same as the previous comment:
I would drop the use of `joystream-cli` commands in the testing framework in favor of the usual fixture/flows scenarios, and if you agree I would propose the following scheme:
1. `yarn workspace setup_pre_migration_scenario` (if needed)
2. replace runtime effectively triggering the migration
3. `yarn workspace setup_post_migration_scenario` (which despite the bad naming I have chosen should only assert that the post migration state is valid)

tests/network-tests/src/scenarios/setup-new-chain.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const leads = job(
+    'Set WorkingGroup Leads',
+    AllWorkingGroups.map((group) => leaderSetup(group, true))
+  )
+
+  const updateWorkerAccounts = job('Update worker accounts', updateAccountsFlow).after(leads)
+
+  if (!process.env.SKIP_MOCK_CONTENT) {
+    // Create some mock content in content directory - without assets or any real metadata
```
dobertRowneySr, line 20: Is the whole mock content flow an alternative to the content creation via joystream cli?

dobertRowneySr (commented):
Left a couple of questions

mnaamani:
> The improvements are great and very useful! I suggested a few minor changes.
> 
> Some additional observations:
> 
>     * Needed to update `docker-compose` from `1.27.4` to `1.29.2` locally, otherwise the scripts like `query-node/start.sh` or `tests/network-tests/run-tests.sh` were not working without explicitly set environment (through `set -a . path/to/.env set +a`)
> 
>     * The migration tests need to be fixed. Previously they relied on the assumption that `@joystream/cli@0.5.1` is installed globally, but this wasn't executed as step in the CI / script. There is also an issue with the app data dir not beeing accessible in Github CI (https://github.com/Joystream/joystream/runs/4235401998?check_suite_focus=true#step:8:51)

Thanks for excellent review will start working to address the requested changes. On the issue regarding docker-compose, @ahhda reported the same issue and shared this link that confirms it: https://docs.docker.com/compose/environment-variables/#the-env-file which seems to recommend using `--project-directory` argument for older versions.

docker-compose.yml
mnaamani: I agree a single default `./config.yml` is better, and env variables can be used to override their values.

.env
mnaamani: Yes I like this approach, its much clearer.

mnaamani: applied in 8d1f42b7a46004a2ae5bca3413b5e55de3205d99

docker-compose.yml
mnaamani: fixed in d3e0d343b53f5a087d5f221ee61ddbfa6cdea30b

tests/network-tests/src/fixtures/membershipModule.ts
mnaamani: good catch. fixed in c4e4bd947f14b40f498c0ffeba668a24944a0f74

tests/network-tests/src/Api.ts
mnaamani: fixed in d442cbe9d231f913f5112e60867e902fa660dd17

tests/network-tests/src/Scenario.ts
mnaamani: fixed in 97654d26bec9ac86584189484161bec53204976e

tests/network-tests/src/Api.ts
mnaamani: fixed in 97654d26bec9ac86584189484161bec53204976e

tests/network-tests/src/Api.ts
mnaamani: Indeed we should, I chose to assert that ids.length == 1. It most closely communicates our intention.
This is a really ugly hack to enforce unique member controller. It would be much better to do a full refactoring of the fixtures to always require a memberId as a parameter rather than do a reverse lookup by account id.

Done in 7d63b04fac263910ba8f45c55730788fb37b6373

tests/network-tests/src/flows/workingGroup/manageWorkerAsWorker.ts
mnaamani: Yes I just forgot to update all the flows. Although I was starting to wonder whether running the same set of tests on multiple instances of working group. Maybe we can add those as needed?

tests/network-tests/src/flows/workingGroup/workerPayout.ts
mnaamani: same comment as above

.env
mnaamani: done in b3ca360396da7a1c76a6eeb4b39c118bf5f433ee

mnaamani:
> > The improvements are great and very useful! I suggested a few minor changes.
> > Some additional observations:
> > ```
> > * Needed to update `docker-compose` from `1.27.4` to `1.29.2` locally, otherwise the scripts like `query-node/start.sh` or `tests/network-tests/run-tests.sh` were not working without explicitly set environment (through `set -a . path/to/.env set +a`)
> > 
> > * The migration tests need to be fixed. Previously they relied on the assumption that `@joystream/cli@0.5.1` is installed globally, but this wasn't executed as step in the CI / script. There is also an issue with the app data dir not beeing accessible in Github CI (https://github.com/Joystream/joystream/runs/4235401998?check_suite_focus=true#step:8:51)
> > ```
> 
> Thanks for excellent review will start working to address the requested changes. On the issue regarding docker-compose, @ahhda reported the same issue and shared this link that confirms it: https://docs.docker.com/compose/environment-variables/#the-env-file which seems to recommend using `--project-directory` argument for older versions.

Or would removing the instances of env_file work? Funnily enough I'm seeing I have that comment about which file get picked up.. which indicates I faced this issue in the past :D and perhaps why I had introduced this awful hack of loading the .env 
```
    env_file:
      # relative to working directory where docker-compose was run from
      - .env
 ```

File: tests/network-tests/src/misc/mockContentFlow.ts (hunk 12 above)
Comment (mnaamani) [reply tests/network-tests/src/misc/mockContentFlow.ts]: done in 5480e207fe70447a1537741063325b9f48aa36a7

File: tests/network-tests/run-migration-tests.sh (hunk 15 above)
Comment (mnaamani) [reply tests/network-tests/run-migration-tests.sh:49]: I don't think it has to be one or the other. We can have to separate scripts and CI jobs. One that uses joystream-cli and the second as you suggest using the testing framework scenarios.

Having the job using the joystream-cli is a nice thing to have for some real integration testing. Although it is a little harded right now .. see db45330885276aae01181d56fb460233939d70f2

As far as I know @Lezek123 is working on writing tests which might also use the cli, so maybe he might find ways to make it easier to use for testing. The only downside however of using the cli is that it was designed for a human to use, doing one transaction at a time. It is less convenient and possibly slower if its needed for doing many sequential steps like creating 100 videos.. etc.

**Comment by mnaamani** [comment]
Tried to fix the permission issue with running cli on github runner https://github.com/Joystream/joystream/pull/2830/commits/db45330885276aae01181d56fb460233939d70f2 but I can't seem to shake this write permission issue:

```
@joystream/cli/0.5.1 linux-x64 node-v14.18.1
 ›   Error: Unexpected error while trying to initialize the data directory.(/ho
 ›   me/runner/work/joystream/joystream/.local/share/joystream-cli)! 
 ›   Permissions issue?
```
Tried changing HOME env to change target location, but now I'm wondering does joystream-cli `mkdir -p` `.local/share` or does it expect it to exist?

**Review by dobertRowneySr — APPROVED** [review]

File: tests/network-tests/run-migration-tests.sh [hunk 17]
```diff
@@ -4,129 +4,97 @@ set -e
...
+  id=$(docker create joystream/node:${TARGET_RUNTIME})
   docker cp $id:/joystream/runtime.compact.wasm ${DATA_PATH}
   docker rm $id
 
-  # Display runtime version before runtime upgrade
-  yarn workspace api-scripts tsnode-strict src/status.ts | grep Runtime
-
   echo "Performing runtime upgrade."
   yarn workspace api-scripts tsnode-strict \
     src/dev-set-runtime-code.ts -- ${DATA_PATH}/runtime.compact.wasm
```
dobertRowneySr, line 88: The file `src/dev-set-runtime-code.ts` is missing (at least from your branch). Should I use the one from Sumer?

dobertRowneySr (commented):
left a question

tests/network-tests/run-migration-tests.sh
mnaamani: that file is actually in `utils/api-scripts/src/` directory. When running `yarn workspace api-scripts ...`, the working directory becomes `utils/api-scripts/`  so `src/dev-set-runtime-code.ts` is relative to that directory. That is why it still works :)

.github/workflows/run-network-tests.yml
```diff
@@ -104,8 +104,15 @@ jobs:
           yarn workspace @joystream/cli build
       - name: Ensure tests are runnable
         run: yarn workspace network-tests build
+      - name: Install joystream-cli @joystream/cli/0.5.1
+        run: npm -g install @joystream/cli
```
Lezek123, line 108: Perhaps we should lock the version to `@joystream/cli@^0.5.1` just in case

Lezek123 (approved):
Perfect!
I added a small suggestion, but it's not really that important.

During my tests I also found a small issue not related to this PR. After building & starting the node and running `yarn verify-chain-metadata`, the script reported that the `chain-metadata.json` does not match.
I didn't add this check to any CI workflow, so this was probably overlooked in some older PR.

Answering your question:

> Tried changing HOME env to change target location, but now I'm wondering does joystream-cli mkdir -p .local/share or does it expect it to exist?

Yes, the CLI currently expects `~/.local/share` or `~/Library/Application Support` to exist (depending on os)

.github/workflows/run-network-tests.yml
mnaamani: I add that in next PR :+1:

mnaamani:
> 

Made an issue for it. https://github.com/Joystream/joystream/issues/2856

PR #2831, opened 2021-11-15
DevOps - Separate stack for deploying a Kubernetes cluster

ahhda:
This PR adds the ability to use an existing Kubernetes cluster for the deployment of a node. A new stack has been added which is responsible for deploying the cluster on either `minikube` or `EKS`.

## Steps to run
* Go to the `cluster` directory
* Run `npm install` & `pulumi stack init`
* For minikube run `pulumi config set-all --plaintext platform=aws`
* Run `pulumi up -y`

When deploying `node-network`, make sure to set the `clusterStackRef` config if you are logged into Pulumi

Screenshot 2021-11-16 at 4 17 18 PM

If you are using Pulumi local (`pulumi login -l`), use the below format:
```
$ pulumi config set clusterStackRef dev
```

If you are using an already deployed cluster, save the Kubeconfig as a file and set the `kubeconfigFile` config

   ```bash
   $ pulumi config set kubeconfigFile <PATH>
   ```


ahhda:
@mnaamani Please review

devops/kubernetes/node-network/index.ts
```diff
@@ -1,342 +1,323 @@
...
-const config = new pulumi.Config()
-const awsConfig = new pulumi.Config('aws')
-const isMinikube = config.getBoolean('isMinikube')
+export = async () => {
+  const config = new pulumi.Config()
 
-export let kubeconfig: pulumi.Output<any>
+  // Get's the Cluster provider and platform based on clusterStackRef platform config
+  const clusterStackRef = new pulumi.StackReference(config.require('clusterStackRef'))
+  const provider = await getProvider(clusterStackRef)
```
mnaamani: I'm getting a compiler error at this line when trying to bring up the stack:
```
Previewing update (network):
     Type                 Name                  Plan       Info
 +   pulumi:pulumi:Stack  node-network-network  create     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (node-network-network):
    error: Running program '/Users/mokhtar/joystream/joystream/devops/kubernetes/node-network' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(14,38): error TS2345: Argument of type 'import("/Users/mokhtar/joystream/joystream/devops/kubernetes/node-network/node_modules/@pulumi/pulumi/stackReference").StackReference' is not assignable to parameter of type 'import("/Users/mokhtar/joystream/joystream/devops/kubernetes/node-network/node_modules/pulumi-common/node_modules/@pulumi/pulumi/stackReference").StackReference'.
      Types have separate declarations of a private property 'readOutputValue'.
    index.ts(15,47): error TS2345: Argument of type 'import("/Users/mokhtar/joystream/joystream/devops/kubernetes/node-network/node_modules/@pulumi/pulumi/stackReference").StackReference' is not assignable to parameter of type 'import("/Users/mokhtar/joystream/joystream/devops/kubernetes/node-network/node_modules/pulumi-common/node_modules/@pulumi/pulumi/stackReference").StackReference'.
```

devops/kubernetes/cluster/index.ts
```diff
@@ -0,0 +1,32 @@
...
+const awsConfig = new pulumi.Config('aws')
+const config = new pulumi.Config()
+
+export let kubeconfig: pulumi.Output<any>
+export let provider: k8s.Provider
+export let platform = config.require('platform')
+
+if (platform === 'minikube') {
+  provider = new k8s.Provider('local', {})
+} else if (platform === 'aws') {
```
mnaamani: I think the following parameters should be configurable when deploying the cluster:

- number of availability zones
- the cluster name
- instanceType

mnaamani (changes requested):
I'm not able to fully test this because of the compiler issue, but a comment on general approach.

When deploying a new stack to a k8 provider, the ability to reference a cluster by pulumi specific stack name is very handy (such as an aws eks cluster we deployed with his new cluster stack), but it shouldn't be the only way. We may not have deployed a kubernetes cluster via pulumi.

We need an additional option of passing a reference to a `kubeconfig` file, to make it as general as possible. Can this be done and are there any downsides to this approach?

ahhda:
> We need an additional option of passing a reference to a `kubeconfig` file, to make it as general as possible. Can this be done and are there any downsides to this approach?

Sounds like a good idea to me. Added the feature 👍

devops/kubernetes/cluster/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const numberOfAvailabilityZones = parseInt(config.get('numberOfAvailabilityZones') || '2')
+const clusterName = config.get('clusterName') || 'eksctl-joystream'
+const instanceType = (config.get('instanceType') || 't2.medium') as pulumi.Input<aws.ec2.InstanceType>
+
+export let kubeconfig: pulumi.Output<any>
+export let provider: k8s.Provider
+export let platform = config.require('platform')
+
+if (platform === 'minikube') {
```
mnaamani: Looking back at these options and trying things out, it doesn't seem necessary to me to have to create a pulumi stack when the provider is just going to be "local" cluster. So I suggest we drop the minikube option, and this script should only deploy clusters to aws, and later maybe add support for google and azure.

devops/kubernetes/node-network/index.ts
```diff
@@ -1,342 +1,339 @@
...
 
-if (isMinikube) {
-  provider = new k8s.Provider('local', {})
-} else {
-  // Create a VPC for our cluster.
-  const vpc = new awsx.ec2.Vpc('joystream-node-vpc', { numberOfAvailabilityZones: 2 })
+  if (kubeconfigFile) {
+    const kubeconfig = readFileSync(kubeconfigFile).toString()
+    provider = new k8s.Provider('cloud', { kubeconfig })
+    isMinikube = false
```
mnaamani: Assumption that passing a kubeconfig file automatically means it is a cloud provider is not correct.
I can for example pass the default kubeconfig file path or a custom config at an arbitrary location with:
```
KUBECONFIG=/tmp/mini-kubeconfig minikube start
or
KUBECONFIG=/tmp/kind-kubeconfig kind create
```

Determining if the selected cluster is a cloud provider or a local cluster (minikube or kind) should be done by inspecting the kubeconfig instead.

devops/kubernetes/node-network/index.ts
```diff
@@ -1,342 +1,339 @@
...
-const awsConfig = new pulumi.Config('aws')
-const isMinikube = config.getBoolean('isMinikube')
+export = async () => {
+  const config = new pulumi.Config()
 
-export let kubeconfig: pulumi.Output<any>
+  const kubeconfigFile = config.get('kubeconfigFile')
+  const clusterStackRefInput = config.get('clusterStackRef')
+  let provider: k8s.Provider = new k8s.Provider('local', {})
+  let isMinikube: boolean = false
```
mnaamani: I think we should drop the minikube config but replace it with `useLocalProvider` for user to have freedom to just use the currently selected context in their default kubeconfig?

mnaamani (changes requested):

mnaamani:
if I create a cloud cluster stack, and then the node-network stack which references the cluster stack, what happens when I run `pulumi destroy` on the node-network stack. Will it also destroy the cluster stack?

ahhda:
> if I create a cloud cluster stack, and then the node-network stack which references the cluster stack, what happens when I run `pulumi destroy` on the node-network stack. Will it also destroy the cluster stack?

It will only destroy the resources created by the `node-network` stack and not the cluster.

PR #2766, opened 2021-10-15
Commit chain metadata + scripts + builds adjustments

Lezek123:
Running `yarn update-chain-metadata` will update `chain-metadata.json` from current local node

Running `yarn verify-chain-metadata` will compare `chain-metadata.json` against the current local node metadata

**Build adjustments:**
- `hydra-typegen` will now use `chain-metadata.json` to generate the types, no need to connect to a running chain during query-node build step
- `@joystream/types` will now use `chain-metadata.json` to generate augmentations, no need to connect to a running chain during `generate:all` step
- most of the files generated by `@joystream/types generate:all` are now excluded from the repository
- `@joystream/types build` will now also run `@joystream/types generate:all`
- added pre-build cleanup in `@joystream/types build` via `git clean -xdf -e node_modules` in order to remove old build artifacts

PR #2854, opened 2021-11-23
Olympia: Improve query-node setup and update README

Lezek123:
- Add `chain-metadata.json` to monorepo and use it during query-node's `hydra-codegen` step (following Giza setup: https://github.com/Joystream/joystream/pull/2766)
- Remove old `graphql-server` and `processor` services and replace them with `graphql-server-mnt` and `processor-mnt` (now renamed to `graphql-server` and `processor`) following similar changes in Giza (https://github.com/Joystream/joystream/pull/2830)
- Updated `start.sh` script
- Updated `query-node/start.sh` scripts (removed unnecessary `warthog:migrate` step)
- Updated `README.md` and `query-node/README.md`


query-node/README.md
```diff
@@ -2,60 +2,47 @@
...
+- Bring up `db` service (query node is using a PostgreSQL database to store the data)
+- Configure the database (`yarn workspace query-node config:dev`)
+- Create the database and initialize schema (`yarn workspace query-node-root db:prepare && yarn workspace query-node-root db:processor:migrate`)
+- Start the GraphQL server service (`docker-compose up -d graphql-server`)
+- Start the Hydra Processor responsible for processing the runtime events and running the mappings' functions (`docker-compose up -d processor`)
 
-```bash
-yarn processor:start
-```
+## 4. Stoping the node and removing all associated containers
```
jodator:
```suggestion
## 4. Stopping the node and removing all associated containers
```

jodator (commented):
@Lezek123 The updated readme is much clearer now. I was able to run the whole setup and query-node only.

However, I see some errors from `joystream_processor_1` service:

```
ERROR: {
  "response": {
    "errors": [
      {
        "message": "Expected type Int, found -9007199254740991.",
        "locations": [
          {
            "line": 3,
            "column": 7424
          }
        ],
        "extensions": {
          "code": "GRAPHQL_VALIDATION_FAILED"
        }
      }
    ],
    "status": 400
  },
  "request": {
    "query": "query ....
...
,\"content.FeaturedVideosSet\",\"dataDirectory.ContentAdded\",\"dataDirectory.ContentRemoved\",\"dataDirectory.ContentAccepted\"], blockNumber_gt: -1, bloc
kNumber_lte: -9007199254740991 }, afterID: \"0000000000-000000\", limit: 2000)
```

---

ps.: I had the same [error](https://github.com/Joystream/joystream/runs/4302627177?check_suite_focus=true#step:9:2122) locally when building joystream node:

> error[E0658]: use of unstable library feature 'btree_retain'

It looks like some dependencies changes as this also fail on other PRs that I've seen.

jodator:
@Lezek123 The processor error is not blocking AFAICS. We just tested this together with #2672 branch and it process everything nicely.

Again, thank you for that update. It helped a lot with running everything together.

Lezek123:
As it turned out `warthog:migrate` step is actually necessary to support search queries (like `threadsByTitle`).
Currently this step was failing on `olympia` branch due to some index name clashes. This probably needs further investigation.

Lezek123:
> As it turned out warthog:migrate step is actually necessary to support search queries (like threadsByTitle).
Currently this step was failing on olympia branch due to some index name clashes. This probably needs further investigation.

Fixed in https://github.com/Joystream/joystream/pull/2854/commits/d2333be7a0a51f0d0798d2b91a902a859f07ffed
There were 2 unreladed issues that caused `warthog:migrate` to fail:
- Lack of `.bin/typeorm` link in `generated/graphql-server/node_modules`
- The `Proposal` entity defined 2 `fulltext` queries, but only 1 is supported per entity (in the current version of Hydra)

query-node/start.sh
```diff
@@ -20,11 +20,11 @@ docker-compose up -d db
 # Make sure we use dev config for db migrations (prevents "Cannot create database..." and some other errors)
 yarn workspace query-node config:dev
 
-# Migrate the databases
+# Prepare & migrate the databases
 yarn workspace query-node-root db:prepare
-yarn workspace query-node-root db:migrate
+yarn workspace query-node-root db:migrate # FIXME
```
jodator:
```suggestion
yarn workspace query-node-root db:migrate
```

This should be fixed by recent changes, right?

jodator (approved):
LGTM, but @thesan please double check.

query-node/start.sh
```diff
@@ -20,11 +20,11 @@ docker-compose up -d db
 # Make sure we use dev config for db migrations (prevents "Cannot create database..." and some other errors)
 yarn workspace query-node config:dev
```
thesan, line 27: Maybe it's due to my laptop being slow but sometime I get a `Error: Connection terminated unexpectedly` when running the `yarn workspace query-node-root start`. It's due to the db not being up yet. Relaunching the command once or at most twice works, but it might be confusing for someone new to the project.

Would this be an acceptable fix?
```suggestion

# Wait for the db to be up
until docker-compose logs db | grep "server started"; do
  echo "Waiting for the db to be ready..."
  sleep 1
done

```

thesan (commented):
Everything is clearer and works well for me, apart from the small issue with the start command.

query-node/start.sh
Lezek123: Looks good, thank you

query-node/start.sh
Lezek123: Done in https://github.com/Joystream/joystream/pull/2854/commits/0c27ce4d5323d84180b47830f639aadd3defe9f7

PR #2857, opened 2021-11-24
Configure basic storage config for playground

mnaamani:
Thanks to handy script from @Lezek123 playground can now start with a properly configured storage infrastructure.
Updated docker-compose to make storage nodes publicly accessible so storage node uri's can be set to public interface of the host.


storage-playground-config.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_PATH="$(dirname "${BASH_SOURCE[0]}")"
+cd $SCRIPT_PATH
+
+# Basic script to be run to configure dev chain storage infrastructure.
+# Assumes one storage lead and distributor lead are already hired
+# and their identity information are stored in .env
+
```
mnaamani, line 10: I'm thinking to add just a step to get the docker local host interface address which will work better as the initial node_uri to be set in the metadata.

storage-playground-config.sh
```diff
@@ -0,0 +1,52 @@
...
+${CLI} leader:update-voucher-limits -o 10000 -s 1000000000000 --accountUri ${COLOSSUS_1_ACCOUNT_URI}
+BUCKET_ID=`${CLI} leader:create-bucket -i=${COLOSSUS_1_WORKER_ID} -a -n=10000 -s=1000000000000  --accountUri ${COLOSSUS_1_ACCOUNT_URI}`
+${CLI} operator:accept-invitation -w=${COLOSSUS_1_WORKER_ID} -i=${BUCKET_ID} --accountUri ${COLOSSUS_1_ACCOUNT_URI}
+${CLI} leader:update-dynamic-bag-policy -n 1 -t Channel --accountUri ${COLOSSUS_1_ACCOUNT_URI}
+${CLI} leader:update-data-fee -f 10 --accountUri ${COLOSSUS_1_ACCOUNT_URI} # Optionally - set some data fee per megabyte
+
+# This is only set initially to the name of the container to make it accessible by the distributors running as containers
+# in the same network. This should be updated to the "public" endpoint so the node is accessible from containers and
+# external applications like Atlas for example that depend on discovering the endpoint uri by querying the query-node.
+COLOSSUS_1_NODE_URI=${COLOSSUS_1_NODE_URI:="http://colossus-1:3333"}
```
Lezek123: > I'm thinking to add just a step to get the docker local host interface address which will work better as the initial node_uri to be set in the metadata.

Yes, that's a good idea. I somehow missed that initially, but with `http://colossus-1:3333` set even the CLI won't be able to connect to storage node in order to upload assets, unless it's ran from a container belonging to the same network.

docker-compose.yml
```diff
@@ -16,8 +16,8 @@ services:
       - chain-data:/data
     command: --dev --alice --validator --unsafe-ws-external --unsafe-rpc-external --rpc-methods Unsafe --rpc-cors=all --log runtime --base-path /data
     ports:
-      - '127.0.0.1:9944:9944'
-      - '127.0.0.1:9933:9933'
+      - 9944:9944
+      - 9933:9933
```
Lezek123, line 20: Since we're using `--rpc-methods Unsafe` and `--rpc-cors=all` flags, perhaps we should add a comment here somewhere that this is not a safe production setup (or modify the setup)

Lezek123 (changes requested):

storage-playground-config.sh
mnaamani: Yes indeed. The same could be said about the test scenarios even.

So there are several approaches here: https://stackoverflow.com/questions/22944631/how-to-get-the-ip-address-of-the-docker-host-from-inside-a-docker-container

.. to get what we need. I'm leaning towards using:

`docker network inspect joystream_default`

which gives us a JSON output of the host machine IP address as well as each individual containers ip address.
Will test that out.

docker-compose.yml
mnaamani: Perhaps joystream-node container doesn't in-fact need to listen on `0.0.0.0`

mnaamani: But more importantly why do we need to enable the unsafe methods?

Lezek123: I remember I suggested that because it allows calling `author_rotateKeys` from localhost / Pioneer when the node is running as docker container. It also allows testing [status server](https://github.com/Joystream/status-endpoint-joystream/tree/development), as it uses `api.rpc.system.peers`

storage-playground-config.sh
mnaamani: I settled on this approach c8de48305a74dd2e50f7f837db2008a14f9a66e9

docker-compose.yml
mnaamani: Ah yes thanks for reminder. I think its best to have the safer default of not allowing the "unsafe rpc" and when the unsafe rpc commands are needed the test node can be started with `./tests/network-tests/run-test-node-docker.sh` which allows the Unsafe commands.

Done in ff8f82be895d18b4529ee4ac73795713da3066a5

tests/network-tests/get-host-ip.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+    # We could get the specific network address for the docker network with
+    #   docker network inspect --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}' joystream_default
+    ip route | awk '/default/ {print $3}'
```
Lezek123: For me locally this approach didn't work.
`ip route | awk '/default/ {print $3}'`
Returned `192.168.0.1`, while:
`docker network inspect --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}' joystream_default`
Returned `172.19.0.1`

When I set the endpoints to `172.19.0.1:3333` and `172.19.0.1:3334` everything worked as expected.
Should we use `docker network inspect --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}' joystream_default` approach instead?

Lezek123 (changes requested):

tests/network-tests/get-host-ip.sh
mnaamani: The only downside of course is any client connecting to `172.19.0.1` would have to be from the same host. For the the purpose of local testing this is acceptable I think.

mnaamani: done in 23ff187

mnaamani:
Updated from giza_staging. In docker-compose for distributors I kept the 'admin' port to only be exposed on 127.0.0.1

Lezek123 (approved):
Works as expected now

PR #2858, opened 2021-11-25
rollback to rust nightly-2021-02-20

mnaamani:
To resolve build failure of `serde_json v1.0.71` we are reverting to older nightly. 

```
error: failed to run custom build command for `joystream-node-runtime (/joystream/runtime)`

Caused by:
  process didn't exit successfully: `/joystream/target/release/build/joystream-node-runtime-....../build-script-build` (exit code: 1)
  --- stdout
  cargo 1.52.0-nightly (90691f2bf 2021-03-16)


  --- stderr
     Compiling serde_json v1.0.71
  error[E0658]: use of unstable library feature 'btree_retain'
     --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_json-1.0.71/src/map.rs:277:18
      |
  277 |         self.map.retain(f);
      |                  ^^^^^^
      |
      = note: see issue #79025 <https://github.com/rust-lang/rust/issues/79025> for more information
      = help: add `#![feature(btree_retain)]` to the crate attributes to enable

  error: aborting due to previous error

  For more information about this error, try `rustc --explain E0658`.
  error: could not compile `serde_json`
```

Had to also make clippy happy by allowing `DAOId` improper uppercase style.

```
error: name `DAOId` contains a capitalized acronym
  --> runtime/src/primitives.rs:18:10
   |
18 | pub type DAOId = u64;
   |          ^^^^^ help: consider making the acronym lowercase, except the initial letter: `DaoId`
   |
   = note: `-D clippy::upper-case-acronyms` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms
```

PR #2859, opened 2021-11-25
Giza update master

mnaamani:
Updates to Giza from master

Including nightly compiler rollback https://github.com/Joystream/joystream/pull/2858


mnaamani:
Perhaps this is a bit of overkill for now. I think I will just cherry-pick the compiler rollback instead.

PR #2860, opened 2021-11-25
Giza staging update from giza and master

mnaamani:
Update from master and giza branches, primarily to bring in rollback of rust nightly. It has brought also some recent updates from master for pioneer and some devops scripts.

There does seem to be an issue with running the tests though, seems the joystream-node is not finding the generated test chainspec file:

```
Error: Input("Error opening spec file: No such file or directory (os error 2)")
```

Will investigate. * edit: Fixed in b1a0148ba22f87dd05705aa2afd258b668aaf61a


Lezek123:
>There does seem to be an issue with running the tests though, seems the joystream-node is not finding the generated test chainspec file:
> Error: Input("Error opening spec file: No such file or directory (os error 2)")
> Will investigate. * edit: Fixed in b1a0148

The issue still seems to occur in `Integration tests: Runtime upgrade` check: https://github.com/Joystream/joystream/runs/4332735960?check_suite_focus=true#step:9:346

tests/network-tests/run-test-node-docker.sh
```diff
@@ -9,7 +9,7 @@ cd $SCRIPT_PATH
 
 # Location that will be mounted as the /data volume in containers
 # This is where the initial members and balances files and generated chainspec files will be located.
-DATA_PATH=${DATA_PATH:=$(pwd)/data}
+DATA_PATH=${DATA_PATH:=$PWD/data}
```
Lezek123: The `Error opening spec file: No such file or directory (os error 2)` still occurs occasionally (in [CI checks](https://github.com/Joystream/joystream/runs/4332735960?check_suite_focus=true) and also on my local machine).

I wonder if it's not because we're mounting the chainspec to `/data` and in `docker-compose.yml` we're also mounting `chain-data` to `/data`. After switching from `-v ${DATA_PATH}:/data` to `-v ${DATA_PATH}:/spec` (and adjusting the location of the chainspec files etc.) I didn't run into this issue anymore.

Lezek123 (changes requested):

tests/network-tests/run-test-node-docker.sh
mnaamani: Well spotted, this issue does seem to have popped up after using local volumes. Although on mac it seems docker-compose was behaving differently. Fixed in f31ba786c985fa784e38c4b4053d528ddbd8a3c3

tests/network-tests/run-test-node.sh
```diff
@@ -7,7 +7,7 @@ cd $SCRIPT_PATH
 # Location that will be mounted as the /data volume in containers
 # This is how we access the initial members and balances files from
 # the containers and where generated chainspec files will be located.
```
Lezek123: The comment mentiones `containers` and `mounting`, which seems unrelated to what the script actually does (starting the node locally, directly from the binary)

Lezek123 (commented):
I suggest just a small fix in the comment and I think it's good to go!

tests/network-tests/run-test-node.sh
mnaamani: done in 048654aed51b5b72d76588aece08dac28c702144

mnaamani:
Pushed last suggested fix and updated from giza_staging

Lezek123 (approved):
LGTM!

Issue #2828, opened 2021-11-12
Runtime Migration


dobertRowneySr:
*This is a note to be used as a reference in case of future (storage) migrations*
# Runtime Side
1. Every pallet should have a version (`u32` is fine) in his storage space, I call this `on_chain_version`
2. Every pallet should have a version (`u32` is fine) hardcoded in the module. I call this `current_storage_version`
*Hence I would like to propose this approach this implementation approach for each pallet*
3. 1 + 2 together allows to implement the [GetStorage](https://paritytech.github.io/substrate/monthly-2021-09+1/frame_support/dispatch/trait.GetStorageVersion.html) trait for `Module<T>`. Hence `let CHECK = on_chain_version !=
current_storage_version`, if `CHECK == true` migration is needed.
4. I should have a separate module for migration like
``` rust
pub mod migration {
  use super::*;

  pub mod v1 {
      // data structures definitions used in previous storage version
  }

  pub fn migrate_to_v2<T: Config>() -> frame_support::weights::Weight {
      if CHECK {
          // translate storageV1 types into storageV2 types
      }
}
```
5. Having `on_runtime_upgrade` in `runtime/src/runtime_api.rs` call the appropriate `migrate_to_v2` function

# Testing
4 current ways to test the runtime upgrade `Sumer` -> `Giza`:
1. two repos: have 2 repos (one with `Sumer`, the other with `Giza`) run one test scenario on `Sumer` in order to create some state to mess up with, then perform the migration and complete assertion launching a test scenario from giza
2. webpack: as a fixer to the previous method we can bundle the whole `Sumer` augmented api into a `sumer.js` file using [webpack](https://webpack.js.org/guides/getting-started/) and import it into the `Giza` test scenarios
3. substrate `remote externalities` / `try_runtime` utils: this allowes to download a state snapshot of a live chain at a specified block height. This is more similar to a unit test since it download and executes rust code into a sandbox environment in which you can verify pre/post migration invariants/assertions.
4. fork-off: forkoff allows to turn a state snapshot of a live chain at specified block height into a readable chainspec file.
# Fork off tool
Seems the most promising, and it is the workflow that best fit the current integration testing framework.  
[...]

PR #2861, opened 2021-11-26
added migration scripts & utils.sh file

dobertRowneySr:
Refers to #2828


dobertRowneySr:
I encountered a problem on the @joystream/cli side, @Lezek123 has been kind enough to start fixing it.
Everything is almost finished, I would also change the nightly toolchain used to avoid the build same serde build problem encountered on the storage node PR

tests/network-tests/.env
```diff
@@ -1,61 +1,58 @@
...
 # Slash value for manage working group lead testing scenario
-SLASH_AMOUNT = 2
+SLASH_AMOUNT=2
 # Stake decrement amount for manage working group lead testing scenario
-STAKE_DECREMENT = 3
+STAKE_DECREMENT=3
 # Mint capacity increment value for working gorup mint capacity test
-MINT_CAPACITY_INCREMENT = 1000
-# Mini-secret or mnemonic used in SURI for deterministic key derivation
-SURI_MINI_SECRET = ""
```
mnaamani, line 58: Why did you drop these two variables? START_KEY_ID and SURI_MINI_SECRET

tests/network-tests/index.js
```diff
@@ -0,0 +1,110 @@
...
+	});
+    }
+
+    // storage.json is guaranteed to exists
+
+    const metadata = await api.rpc.state.getMetadata();
+    // Populate the prefixes array
+    const modules = metadata.asLatest.pallets;
+    modules.forEach((module) => {
+	if (module.storage) {
```
mnaamani: there is some bad indentation here. try running `yarn format` or `yarn prettier ./index.js` to fix

tests/network-tests/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "network-tests",
   "version": "0.1.0",
   "license": "GPL-3.0-only",
-  "scripts": {
+    "scripts": {
```
mnaamani: bad indentation

tests/network-tests/run-migration-tests.sh
```diff
@@ -68,33 +133,174 @@ function post_migration_hook() {
   fi
   # This cammand doesn't give error exit code if videos not found in a channel
   yarn joystream-cli content:videos 1
-}    
+}
+
```
mnaamani: Code is repeating here??

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,29 +4,87 @@ set -e
...
+# If state modification by means of typescript scenarios are required before migration
+PRE_MIGRATION_ASYNC_SETUP=${PRE_MIGRATION_ASYNC_SETUP:=$false}
+# Post migration assertions by means of joystream-cli required
+POST_MIGRATION_CLI_ASSERTIONS=${POST_MIGRATION_CLI_ASSERTIONS=$false}
+# Post migration assertions by means of typescript scenarios required
+POST_MIGRATION_ASYNC_ASSERTIONS=${POST_MIGRATION_ASYNC_ASSERTIONS=$true}
 
-# The joystream/node docker image tag to start chain
-export RUNTIME=${RUNTIME:=latest}
+export WS_RPC_ENDPOINT="wss://testnet-rpc-3-uk.joystream.org"
```
mnaamani: its not clear which part of the code uses this variable. Perhaps put it closer to the point where it is used.

tests/network-tests/node-utils.sh
```diff
@@ -0,0 +1,140 @@
...
+#######################################
+# Start a chain with generated chain spec
+# Globals:
+#   DATA_PATH
+# Arguments:
+#   None
+#######################################
+function start_node {
+    docker-compose -f ../../docker-compose.yml run \
+		   -d -v ${DATA_PATH}:/data --name "${JOYSTREAM_NODE_TAG}" \
```
mnaamani: mount the DATA_PATH in a different location than the `/data` volume which is mounted by docker-compose to a local volume.

tests/network-tests/index.js
```diff
@@ -0,0 +1,110 @@
+const fs = require('fs');
```
mnaamani: When and how should this script be used.
index.js is not very informative is its not really the root file for the network-tests package. Did you replace this file with `utils/api-scripts/src/fork-off.ts` perhaps ?

tests/network-tests/run-migration-tests.sh
```diff
@@ -68,33 +133,174 @@ function post_migration_hook() {
...
+    # verify that the number of outstanding channels & videos == 0
+    # if ( $POST_MIGRATION_CLI_ASSERTIONS ); then
+    # 	# verify assertion using cli
+    # 	echo "***** POST MIGRATION CLI *****"
+    # 	post_migration_cli
+    # fi
+    
+    if ( $POST_MIGRATION_ASYNC_ASSERTIONS ); then
+	# verify assertion using typsecript
+	sleep 10 # TODO: 3m and 52 sec needed!!
```
mnaamani: It took a little bit less on my machine, but yes its not surprising since the chainspec file is around 100MB.

tests/network-tests/run-migration-tests.sh
```diff
@@ -68,33 +133,174 @@ function post_migration_hook() {
...
+#   None
+#######################################
+function fork_off_init() {
+    # chain-spec-raw already existing
+
+    if ! [[ -f $(pwd)/storage.json ]]; then
+	curl http://testnet-rpc-3-uk.joystream.org:9933 -H \
+	     "Content-type: application/json" -d \
+	     '{"jsonrpc":"2.0","id":1,"method":"state_getPairs","params":["0x"]}' \
+	     > storage.json
```
mnaamani: perhaps save this directly to the data folder

mnaamani (changes requested):
Managed to get the run-migration-tests to run. Had to wait for node to come up and then run the 'post-migration' scenario again manually. But the test fails.

```
yarn workspace network-tests node-ts-strict src/scenarios/post-migration.ts
yarn workspace v1.22.4
yarn run v1.22.4
$ node -r ts-node/register --unhandled-rejections=strict src/scenarios/post-migration.ts
2021-12-06 17:52:01        RPC-CORE: queryStorageAt(keys: Vec<StorageKey>, at?: BlockHash): Vec<StorageChangeSet>:: Unable to decode storage content.channelById: entry 0:: createType(Channel):: decodeU8a: failed at 0xba16000000000000… on reward_account: Option<GenericAccountId>:: Invalid AccountId provided, expected 32 bytes, found 11
2021-12-06 17:52:01             DRR: Unable to decode storage content.channelById: entry 0:: createType(Channel):: decodeU8a: failed at 0xba16000000000000… on reward_account: Option<GenericAccountId>:: Invalid AccountId provided, expected 32 bytes, found 11
Error: Unable to decode storage content.channelById: entry 0:: createType(Channel):: decodeU8a: failed at 0xba16000000000000… on reward_account: Option<GenericAccountId>:: Invalid AccountId provided, expected 32 bytes, found 11
    at RpcCore._newType (/Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:450:13)
    at RpcCore._formatStorageSetEntry (/Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:430:17)
    at /Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:410:25
    at Array.reduce (<anonymous>)
    at RpcCore._formatStorageSet (/Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:409:17)
    at /Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:384:55
    at Array.map (<anonymous>)
    at RpcCore._formatOutput (/Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:381:29)
    at callWithRegistry (/Users/mokhtar/joystream/joystream/node_modules/@polkadot/rpc-core/bundle.cjs:255:42)
Job Results:
Verify post-migration chain state: Failed

```

Failure to decode the channel type indicates that the channels from sumer runtime still exist in the state and so when reading the `content.channelById.entries()` return some results. This suggests the code for on_runtime_migration didn't get executed?

However inspecting with pioneer I see what is expected so I'm not sure what is going on...

Screen Shot 2021-12-06 at 6 08 04 PM

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,32 @@
...
+    assert(nextVideoCategoryId.toNumber() > 1);
+    assert(nextVideoId.toNumber() > 1);
+    assert(nextChannelId.toNumber() > 1);
+
+    // asserting the number of outstanding channel & videos is 0
+    debug('Checking that number of outstanding channels & videos == 0');
+
+    assert(num_channels === 0);
+    assert(num_videos === 0);
+    assert(num_categories === 0);
```
mnaamani: we don't actually delete categories on runtime upgrade so num_categories is actually expected to not be 0.
Ideally we should confirm that it is same as before the runtime upgrade.

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const nextVideoCategoryId = await api.query.content.nextVideoCategoryId()
+    const nextVideoId = await api.query.content.nextVideoId()
+    const nextChannelId = await api.query.content.nextChannelId()
+
+    const num_channels = await api.getNumberOfOutstandingChannels()
+    const num_videos = await api.getNumberOfOutstandingVideos()
+    const num_categories = await api.getNumberOfOutstandingVideoCategories()
+
+    assert(nextVideoCategoryId.toNumber() > 1);
```
mnaamani: These assertions are failing because `.toNumber()` is returning a `BN` type. so you want:
```suggestion
    assert(nextVideoCategoryId.toNumber().gt(1));
```

dobertRowneySr:
The seconds to sleep are a rough estimate based on the time I observed on my machine.
Pre-migration setup does not work, since it needs `sumer` `@joystream/types`.
For the purpose of this migration pre-setup should be not needed however.
I also removed Post-migration CLI-checks, since there are no cli-commands in order to check the status of the pending migration

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+    assert(wg_gateway.toNumber() === 0);
+    assert(wg_content.toNumber() === 0);
+
+    debug('Checking that Video, Channel, Categories  counters have not been re-set')
+
+    const nextVideoCategoryId = await api.query.content.nextVideoCategoryId()
+    const nextVideoId = await api.query.content.nextVideoId()
+    const nextChannelId = await api.query.content.nextChannelId()
+
+    assert(nextVideoCategoryId.toNumber().gt(1));
```
mnaamani: perhaps I was mistaken about the type? see: https://github.com/Joystream/joystream/runs/4434220829?check_suite_focus=true#step:7:7

tests/network-tests/.env
```diff
@@ -1,61 +1,60 @@
...
+SLASH_AMOUNT=2
 # Stake decrement amount for manage working group lead testing scenario
-STAKE_DECREMENT = 3
+STAKE_DECREMENT=3
 # Mint capacity increment value for working gorup mint capacity test
-MINT_CAPACITY_INCREMENT = 1000
+MINT_CAPACITY_INCREMENT=1000
 # Mini-secret or mnemonic used in SURI for deterministic key derivation
-SURI_MINI_SECRET = ""
-# The starting key id to use when running a scenario. This will allow scenario
```
mnaamani, line 59: Please don't drop this START_KEY_ID from .env

tests/network-tests/index.js
mnaamani: This index.js file isn't used can we drop it ?

utils/api-scripts/src/fork-off.ts
```diff
@@ -0,0 +1,121 @@
...
+// paths & env variables
+let alice = process.env.SUDO_ACCOUNT
+// bad error handling TODO: fix process.env
+let schemaPath = path.join(process.env.DATA_PATH || "", 'schema.json');
+let wasmPath = path.join(process.env.DATA_PATH || "", 'runtime.wasm') || "";
+let hexPath = path.join(process.env.DATA_PATH || "", 'runtime.hex') || "";
+let specPath = path.join(process.env.DATA_PATH || "", 'chain-spec-raw.json');
+let storagePath = path.join(process.env.DATA_PATH || "", 'storage.json');
+
+// this might not be of much use
```
mnaamani, line 16: If its of no use can we drop it?

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+    assert(wg_gateway.toNumber() === 0);
+    assert(wg_content.toNumber() === 0);
+
+    debug('Checking that Video, Channel, Categories  counters have not been re-set')
+
+    const nextVideoCategoryId = await api.query.content.nextVideoCategoryId()
+    const nextVideoId = await api.query.content.nextVideoId()
+    const nextChannelId = await api.query.content.nextChannelId()
+
+    assert(nextVideoCategoryId.toNumber(). > 1);
```
mnaamani:
```suggestion
    assert(nextVideoCategoryId.toNumber() > 1);
```

mnaamani: you should really run `yarn build` in network-tests folder to catch these errors.

mnaamani (changes requested):

tests/network-tests/node-utils.sh
```diff
@@ -0,0 +1,140 @@
...
+#######################################
+# Start a chain with generated chain spec
+# Globals:
+#   DATA_PATH
+# Arguments:
+#   None
+#######################################
+function start_node {
+    docker-compose -f ../../docker-compose.yml run \
+		   -d -v $(pwd) --name "${JOYSTREAM_NODE_TAG}" \
```
mnaamani:
```suggestion
		   -d -v ${DATA_PATH}:/spec --name "${JOYSTREAM_NODE_TAG}" \
```

tests/network-tests/node-utils.sh
```diff
@@ -0,0 +1,140 @@
...
+		   -p 9944:9944 -p 9933:9933 joystream-node \
+		   --alice --validator --unsafe-ws-external --unsafe-rpc-external \
+		   --rpc-methods Unsafe --rpc-cors=all -l runtime \
+		   --chain /data/chain-spec-raw.json
```
mnaamani:
```suggestion
		   --chain /spec/chain-spec-raw.json
```

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+export default async function postMigrationAssertions({ api }: FlowProps): Promise<void> {
+    const debug = extendDebug('flow:postMigrationAssertions')
+    debug('Started')
+
+    debug('Ensure migration is done')
+
+    const channel_migration = await api.query.content.channelMigration();
+    const video_migration = await api.query.content.videoMigration();
+
+    assert(channel_migration.current_id.toNumber() == channel_migration.final_id.toNumber())
```
mnaamani:
```suggestion
    assert.equal(channel_migration.current_id.toNumber(), channel_migration.final_id.toNumber())
```
Better to use assert.equal to compare values when the assertion fails

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+    const wg_gamma = await api.query.operationsWorkingGroupGamma.activeWorkerCount();
+    const wg_gateway = await api.query.gatewayWorkingGroup.activeWorkerCount();
+    const wg_content = await api.query.contentWorkingGroup.activeWorkerCount();
+
+    assert(wg_alpha.toNumber() === 0);
+    assert(wg_beta.toNumber() === 0);
+    assert(wg_gamma.toNumber() === 0);
+    assert(wg_gateway.toNumber() === 0);
+    assert(wg_content.toNumber() === 0);
+
```
mnaamani, line 49: only new groups would have 0 workers. Renamed modules will have same number of workers as before runtime upgrade.

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+    const debug = extendDebug('flow:postMigrationAssertions')
+    debug('Started')
+
+    debug('Ensure migration is done')
+
+    const channel_migration = await api.query.content.channelMigration();
+    const video_migration = await api.query.content.videoMigration();
+
+    assert(channel_migration.current_id.toNumber() == channel_migration.final_id.toNumber())
+    assert(video_migration.current_id.toNumber() == video_migration.final_id.toNumber())
```
mnaamani: When I ran locally I got the channel assertion passing but the video assertion failing.

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,97 +4,76 @@ set -e
...
+    # use forkoff to update chainspec with the live state + update runtime code
+    fork_off_init
+
+    echo "***** STARTING NODE WITH FORKED STATE *****"
+    export JOYSTREAM_NODE_TAG=$RUNTIME_TAG
+    CONTAINER_ID=$(start_node)
+
+    # 120 sec needed to load the chainspec + 480 sec in order to perform migration, given that
+    # a new block is produced every 6 sec
+    sleep 600
```
mnaamani: Getting the timing correct here without inspecting how many videos channels exist and at what rate they are being migrated is not good enough. A better approach would be in the postMigrationAssertionsFlow.ts check on the progress of the migration, wait until a specific block has been reached. You may need some helper methods to be added in the `Api.ts`.

utils/api-scripts/src/fork-off.ts
```diff
@@ -0,0 +1,121 @@
...
+    chainSpec.protocolId = chainSpec.protocolId;
+
+    // Grab the items to be moved, then iterate through and insert into storage
+    storage
+        .result
+        .filter((i) => prefixes.some((prefix) => i[0].startsWith(prefix)))
+        .forEach(([key, value]) => (chainSpec.genesis.raw.top[key] = value));
+
+    // Delete System.LastRuntimeUpgrade to ensure that the on_runtime_upgrade event is triggered
+    delete chainSpec.genesis.raw.top['0x26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8'];
```
mnaamani, line 91: It would be better to show how these raw values are derived so this script can be generalized if we use it against a different chain.

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
mnaamani:
```
yarn run v1.22.4
$ node -r ts-node/register --unhandled-rejections=strict src/scenarios/post-migration.ts
AssertionError: expected 7600 to equal 8192
```
When I ran with assert.equal(..)

mnaamani (changes requested):

tests/network-tests/run-migration-tests.sh
dobertRowneySr: I added a while loop in `postMigrationAssertionsFlow.ts`, it should be robust but I am pretty sure that there are more "elegant" approaches. Do you have any idea how I can ping `ws://127.0.0.0:9944` in order to check when the chainspec has been finished loading? I have put a sleep 120 in order to account for loading time, but I guess there should be a more robust solution

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
dobertRowneySr: I guess it's due to the sleep 600 line (underestimation). I fixed it by adding a while loop

utils/api-scripts/src/fork-off.ts
dobertRowneySr: The raw values are derived from the pallet name + storage element (+ eventual key in the map). Deriving the value I believe doesn't prevent the script from breaking in case of a different chain (since it depends on the chain pallets configuration), but it makes it easier to fix...

tests/network-tests/src/misc/postMigrationAssertionsFlow.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    let channel_migration = await api.query.content.channelMigration();
+    let video_migration = await api.query.content.videoMigration();
+
+    // wait for migration to be done
+    while (channel_migration.current_id.toNumber() < channel_migration.final_id.toNumber() ||
+        video_migration.current_id.toNumber() < video_migration.final_id.toNumber()) {
+        channel_migration = await api.query.content.channelMigration();
+        video_migration = await api.query.content.videoMigration();
+
```
mnaamani, line 22: This is better, I would just add:
- a short delay in the loop 
- a check for condition that current_id's are indeed changing otherwise this will be an infinite loop.

There is a utility method you could use:

```
import { Utils } from '../../utils'

await Utils.wait(6000) // wait for 6second

```

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,97 +4,75 @@ set -e
...
+
+    # use forkoff to update chainspec with the live state + update runtime code
+    fork_off_init
+
+    echo "***** STARTING NODE WITH FORKED STATE *****"
+    export JOYSTREAM_NODE_TAG=$RUNTIME_TAG
+    CONTAINER_ID=$(start_node)
+
+    # 120 sec needed to load the chainspec
+    sleep 120
```
mnaamani, line 78: I propose we do the following to have the initial genesis state be written to db so we don't have to wait an arbitrary amount of time which needs to be different depending on what hardware is running:
```suggestion
   docker-compose -f ../../docker-compose.yml run -v ${DATA_PATH}:/spec joystream-node export-state --chain /spec/chain-spec-raw.json --base-path /data --pruning archive > ${DATA_PATH}/exported-state.json
   # short delay for node to startup
    sleep 30
```

The `--base-path /data` argument is important to be the same as one used in node-utils.sh to start the node with the same chain data generated in this step.

tests/network-tests/node-utils.sh
```diff
@@ -0,0 +1,140 @@
...
+# Arguments:
+#   None
+#######################################
+function start_node {
+    docker-compose -f ../../docker-compose.yml run \
+		   -d -v ${DATA_PATH}:/spec --name "${JOYSTREAM_NODE_TAG}" \
+		   -p 9944:9944 -p 9933:9933 joystream-node \
+		   --alice --validator --unsafe-ws-external --unsafe-rpc-external \
+		   --rpc-methods Unsafe --rpc-cors=all -l runtime \
+		   --chain /spec/chain-spec-raw.json
```
mnaamani:
```suggestion
		   --chain /spec/chain-spec-raw.json --base-path /data
```

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,97 +4,75 @@ set -e
...
-    rm ./assets/TestVideo__rejectedContent.json || true
+TARGET_RUNTIME_TAG=${TARGET_RUNTIME_TAG:=latest}
+# The joystream/node docker image tag to start the chain with
+RUNTIME_TAG=${RUNTIME_TAG:=sumer}
+# Post migration assertions by means of typescript scenarios required
+POST_MIGRATION_ASYNC_ASSERTIONS=${POST_MIGRATION_ASYNC_ASSERTIONS=$true}
+# The joystream/node docker image tag to start the chain with
+RUNTIME_TAG=${RUNTIME_TAG:=sumer}
+# RPC endpoint for live RUNTIME testnet 
+export WS_RPC_ENDPOINT="wss://testnet-rpc-3-uk.joystream.org"
```
mnaamani:
```suggestion
```
Put it closer to where it is used.

tests/network-tests/run-migration-tests.sh
```diff
@@ -4,97 +4,75 @@ set -e
...
+    fi
+
+    if ! [[ -f ${DATA_PATH}/schema.json ]]; then
+	cp $SCRIPT_PATH/../../types/augment/all/defs.json ${DATA_PATH}/schema.json
+    fi
+
+    id=$(docker create joystream/node:${TARGET_RUNTIME_TAG})
+    docker cp $id:/joystream/runtime.compact.wasm ${DATA_PATH}/runtime.wasm
+    
+    yarn workspace api-scripts tsnode-strict src/fork-off.ts
```
mnaamani, line 45:
```suggestion
WS_RPC_ENDPOINT="wss://testnet-rpc-3-uk.joystream.org" yarn workspace \
    api-scripts tsnode-strict src/fork-off.ts
```

mnaamani (changes requested):
Almost there, made a few last suggestions.

In `test/network-tests/` working directory:
There are a few linter fixes to do in the typescript code.
run `yarn checks` to check them (or you might see them in your editor with eslint plugin enabled)
run `yarn format` to fix formatting.

PR #2862, opened 2021-11-26
Fix cross-platform issue with tests

mnaamani:
Fixes getting different behaviors on linux vs mac in some bash scripts.

use `$PWD`  not $(pwd)
and `$(command)` not \`command\`

ref: 
https://stackoverflow.com/questions/10795014/pwd-vs-pwd-regarding-portability
https://stackoverflow.com/questions/22709371/backticks-vs-braces-in-bash


dobertRowneySr (approved):

mnaamani:
Part of https://github.com/Joystream/joystream/pull/2860

PR #2876, opened 2021-11-30
Query node: Fix for "table name x specified more than once"

Lezek123:
Fixes `table name x specified more than once` issue spotted by @kdembler while executing a query:
```graphql
{
  videos(
    limit: 10
    where: {
      media: { isAccepted_eq: true }
      thumbnailPhoto: { isAccepted_eq: true }
    }
  ) {
    id
    title
    media {
      id
      isAccepted
    }
    thumbnailPhoto {
      id
      isAccepted
    }
  }
}
```
Previously this resulted in:
```
# ...
"message": "table name \"storage_data_object\" specified more than once"
# ...
```

**Reason:**
No aliasing was used when joining related tables. When the same table was joined twice (`storage_data_object` in this case), the query was failing due to name clash.

**Fix:**
I implemented a patch for `@joystream/warthog` to use `uuid/v4` as table alias when joining `OneToOne`, `ManyToOne` and `OneToMany` relationships. The logic for `ManyToMany` relationships is a bit different, but since we currently have no case of joining two separate `ManyToMany` relationships in the same query with the same table names, I decided to skip patching this part.

**Testing:**
The query mentioned above appears to work as expected now.
I also tested all [Colossus](https://github.com/Joystream/joystream/blob/giza_staging/storage-node-v2/src/services/queryNode/queries/queries.graphql) and [Argus](https://github.com/Joystream/joystream/blob/giza_staging/distributor-node/src/services/networking/query-node/queries/queries.graphql) - all of them also worked as expected.

### DIFF
--- a/query-node/codegen/package.json
+++ b/query-node/codegen/package.json
@@ -4,6 +4,9 @@
   "description": "Hydra codegen tools for Joystream Query Node",
   "author": "",
   "license": "ISC",
+  "scripts": {
+    "postinstall": "cd .. && yarn workspace query-node-mappings postHydraCLIInstall"
+  },
   "dependencies": {
     "@joystream/hydra-cli": "3.1.0-alpha.13",
     "@joystream/hydra-typegen": "3.1.0-alpha.13"
--- a/query-node/mappings/package.json
+++ b/query-node/mappings/package.json
@@ -12,6 +12,7 @@
     "checks": "prettier ./ --check && yarn lint",
     "format": "prettier ./ --write ",
     "postinstall": "yarn ts-node ./scripts/postInstall.ts",
+    "postHydraCLIInstall": "yarn ts-node ./scripts/postHydraCLIInstall.ts",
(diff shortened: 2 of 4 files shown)


kdembler:
The mentioned query works fine now but there's another issue when running the search query: `missing FROM-clause entry for table \"c524a952049f-41c1-9612-7ad5495091f7\"`

Full query:
```
{
  search(text: "test", whereVideo: { media: { isAccepted_eq: true } }) {
    item {
      __typename
    }
  }
}
```

Lezek123:
> The mentioned query works fine now but there's another issue when running the search query: `missing FROM-clause entry for table \"c524a952049f-41c1-9612-7ad5495091f7\"`
> 
> Full query:
> 
> ```
> {
>   search(text: "test", whereVideo: { media: { isAccepted_eq: true } }) {
>     item {
>       __typename
>     }
>   }
> }
> ```

Should be fixed in https://github.com/Joystream/joystream/pull/2876/commits/90f8f97ed7d9b1271413e21de41af79adbe22981

PR #2885, opened 2021-12-03
Giza: rollback to rust nightly-2021-02-20

mnaamani:


shamil-gadelshin (approved):
LGTM

There are no entries with the old compiler version. Docker build works (tested via /query-node/start.sh), cargo build works (tested via ./scripts/run-dev-chain.sh).

PR #2888, opened 2021-12-03
Giza staging: query-node/kill.sh script fix

Lezek123:
Remove `query-node-data` volume during `query-node/kill.sh` to allow restarting the query node via `./query-node/kill.sh && ./query-node/start.sh`


mnaamani:
I suffered from this the other day until I finally realized it was the persisted volume :)
I'd falsely assumed the `-v` argument would remove them but that only removed anonymous volumes apparently.

mnaamani (approved):

PR #2896, opened 2021-12-06
Pioneer: Show active proposals per default

traumschule:
Small change to save load time for often used route:
https://pioneer.joystreamstats.live/#/proposals


mochet:
I think this could be quite a helpful change, some have already said the change made on the joystreamstats pioneer server is good for being able to vote quicker.

Issue #2865, opened 2021-11-26
Scaleable distribution bucket & family data model


bedeho:
# Background

In the original on-chain data model for distribution bucket families and their distribution buckets, an entire such family, and corresponding buckets, were stored as one single storage value.

```Rust
pub struct DistributionBucketFamilyRecord<DistributionBucketId: Ord, WorkerId: Ord> {
    pub distribution_buckets: BTreeMap<DistributionBucketId, DistributionBucketRecord<WorkerId>>,
}
```

This model means that one only needs a single database operation, reading this record, when one, during bag creation, is trying to establish what distribution buckets should be responsible for a bag. This is nice. The problem is however that one has to constrain the size of this object, in particular `distribution_buckets`, as the read/write time is eventually going to start rising as well. 

https://github.com/Joystream/joystream/issues/2234

Each byte contributes `0.0003838545018007194` and `0.002405831452581034` seconds to read and write times on reference hardware, respectively. It also tradeoff between complex+brittle or wasteful+expensive fees for impacted extrinsics, as they will either need to have the user commit to a map size as a way to charge precise fees, or to just always charge as if the object has max size, respectively.

The most fundamental problem is that any limit we impose on size of `distribution_buckets` will impact the ability of the system to scale, and it is very hard to make reliable predictions about what what limits we will need into the future on mainnet. What compounds this is that if our utilization runs into this limit, then trying to move away from a limit based model will cause the need for a in-production runtime upgrade+migration of the most epic proportions.

# Possible Solution

The easiest solution is to go for a standard model where `DistributionBucketRecord` are not stored inline like this, but in a separate double map with key space `DistributionBucketFamilyIdxDistributionBucketId´. The bag creation process will now try to find a suitable buckets subset of buckets, in each family, just like storage buckets are found, using a combination of random sampling of the key space + a fallback linear method. The number of database operations will go up, but they will each be small, and the number of operations does not scale with size of the system. This will imply

1. introducing a double map.
2. distributor buckets are identified by combination of family and bucket id.

# Preliminary Work

Before we make any change, it would be useful to map out what the full range of consequences would be on trying to update this now, meaning

- runtime
  - extrinsics
  - events
  - storage
  - unit tests
- colossus
- query node: mapping & schemas
- argus
- cli
- integration tests
- anything else?

Can each party assigned here comment on how much time it require them to update the part of the surface area for this change? 
[...]

PR #2897, opened 2021-12-06
Giza: introduce complex distribution bucket IDs.

shamil-gadelshin:
This PR modifies the `DistributionBucketId` from the simple u64 type to the structure based composed of `DistributionBucketFamilyId` and `DistributionBucketIndex`. 

Changes:
- reworked complex DistributionBucketId
- added DistributionBucketIndex type and changed DistributionBucketFamily type
- changed the random bucket selection algorithm for the distribution buckets.
- `types` library was changed

It supersedes PRs - https://github.com/Joystream/joystream/pull/2797 and https://github.com/Joystream/joystream/pull/2792

Relates to https://github.com/Joystream/joystream/issues/2865


types/src/storage.ts
```diff
@@ -220,13 +232,23 @@ export class DistributionBucket
...
 
+export type IDistributionBucketState = {
+  accepting_new_bags: bool
+}
+
+export class DistributionBucketState
+  extends JoyStructDecorated({
+    accepting_new_bags: bool,
+  })
+  implements IDistributionBucketState {}
```
Lezek123: I couldn't find this type defined anywhere in the runtime

Lezek123 (changes requested):

types/src/storage.ts
shamil-gadelshin: This is dead code. I removed it here 14c4633

runtime-modules/storage/src/lib.rs
```diff
@@ -826,45 +828,42 @@ impl<Balance: Saturating + Copy> BagUpdate<Balance> {
...
 
-    // Checks inner buckets for bag assignment number. Returns true only if all 'assigned_bags' are
-    // zero.
-    fn no_bags_assigned(&self) -> bool {
-        self.distribution_buckets
-            .values()
-            .all(|b| b.no_bags_assigned())
-    }
+/// Type alias for the DistributionBucketIdRecord.
+pub type DistributionBucketId<T> = DistributionBucketIdRecord<
```
bedeho, line 851: Is it risky, or confusing, to have `DistributionBucketId` and `DistributionBucketIndex` denote distinct but intimately related concepts? This caused me a bit of confusion when reading through.

Perhaps the latter could be something like `WithinFamilyDistributionBucketId` or something? just a tip, you decide what you want to do.

runtime-modules/storage/src/random_buckets/storage_bucket_picker.rs
```diff
@@ -0,0 +1,65 @@
...
+pub(crate) use super::{RandomBucketIdIterator, SequentialBucketIdIterator};
+use crate::{DynamicBagType, Module, Trait};
+
+// Generates storage bucket IDs to assign to a new dynamic bag.
+pub(crate) struct StorageBucketPicker<T> {
+    trait_marker: PhantomData<T>,
+}
+
+impl<T: Trait> StorageBucketPicker<T> {
+    // Selects storage bucket ID sets to assign to the storage bucket.
```
bedeho: `assign to the bag`

bedeho (approved):
This looks good to me, approving under the assumption that there will be fix later which introduces the ability, and perhaps replaces wholesale the current permissive mode, to create dynamic bags which are guaranteed to be stored and distributed in a policy compliant way.

runtime-modules/storage/src/random_buckets/storage_bucket_picker.rs
shamil-gadelshin: Fixed.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: `DistributionBucketIndex` seemed to satisfy the rest of the team. So, I tend to keep the name for now.

Issue #2904, opened 2021-12-08
Unbiased fallback bucket selection algorithm


shamil-gadelshin:
We need to pick storage buckets during the bag creation, and in order to distribute the load for the storage provider, we should pick them randomly. However, random selection could result in a large number of choices of inappropriate buckets (deleted, disabled, full, etc.), so we limit the number of random attempts. If we fail all random picking attempts, we fall back to sequential bucket selection starting from zero ID. 

And this, in turn, could lead to disproportionate load to first buckets because the fallback algorithm will always start from the beginning. To mitigate the issue, we could introduce the last chosen bucket ID for the sequential fallback algorithm and use it as a starting point for the algorithm updating it after each selection.

Relates to https://github.com/Joystream/joystream/issues/2902

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201810380192325) by [Unito](https://www.unito.io)

bedeho:
Superseded by https://github.com/Joystream/joystream/issues/3289

Issue #2882, opened 2021-12-02
Giza: Storage bucket to bag assignment runtime bug


Lezek123:
During recent tests a major bug in the Giza runtime related to storage-buckets-to-storage-bag assignments was discovered

**Bug description**

1. When we try to create a new channel with 2 assets and one of the storage buckets assigned to store the channel's bag has `voucher.objectsLimit < voucher.objectsUsed + 2`, the runtime will throw `StorageBucketObjectNumberLimitReached` error
2. At this point something will "get stuck" in the runtime and no matter how we change the replication factor for channel bags (`update-dynamic-bag-policy`) or how many other valid buckets we have (we can also try creating new ones), any attempt to create the channel (as mentioned in step 1) will always result in `StorageBucketObjectNumberLimitReached`
3. The only way to break out of the situation described in step 2 seems to be increasing the `objectsLimit` of the bucket mentioned in step 1. If we make the limits high enough for the bucket to be able to accept 2 new assets, we should be able to finally create the channel, but the current status of the buckets (`acceptingNewBags` flag) and `DynamicBagPolicy` of `Channel` will have **no effect** on the new bag's assignment. The runtime will assign the new channel bag to the buckets it has "already chosen" in step 1. For example: If in step 1 we had `dynamicBagCreationPolicies(Channel).numberOfStorageBuckets == 3` and all buckets were accepting new bags and now we have `dynamicBagCreationPolicies(Channel).numberOfStorageBuckets == 1` and none of the buckets are accepting new bags, the new channel bag will still be assigned to 3 buckets (one of them beeing the bucket which limits we just increased).
4. Any channel created after step 3 will have its bag correctly assigned according to current policy and bucket statuses.

_**Edit:** It turned out that the channel's dynamic bag is actually created and assigned to buckets in step `1.` (unsafe mutation) even though the extrinsic fails and the channel itself it not created. See discussion below._

**How to reproduce**

The easiest way is to use the script I created on my `giza-integration-tests` branch: https://github.com/Lezek123/substrate-runtime-joystream/blob/giza-integration-tests/reproduce-giza-issue.sh

bedeho:
- Really great find, how was this found?
- I would the initial extrinsic invoked in the content directory for channel creation should just fail because the checks before the `mutation safe` barrier should prevent any side-effects when it asks the data directory whether the imminent addition of data objects is feasible. Is this not what is happening? Seems to not be the case, as behavior of the system seems to change, i.e. there has been some side-effect.

[...]

PR #2907, opened 2021-12-09
Storage bucket content fix

dobertRowneySr:


runtime-modules/content/src/lib.rs
```diff
@@ -1419,12 +1397,24 @@ impl<T: Trait> Module<T> {
...
+
+        let upload_params = UploadParametersRecord {
+            bag_id: bag_id.clone(),
+            object_creation_list: assets.object_creation_list.clone(),
+            deletion_prize_source_account_id: prize_source_account.clone(),
+            expected_data_size_fee: assets.expected_data_size_fee,
+        };
+
+        // create dynamic bag if it doesn't exist (does not emits conflicting created/deleted bag events)
+        if let Err(_) = Storage::<T>::ensure_bag_exists(&bag_id) {
```
shamil-gadelshin: Try this one: https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err

shamil-gadelshin (changes requested):
This PR contains a major issue:
It doesn't solve the main problem of preventing failed bag creation. The problem is rooted in bucket resources exhaustion but the code doesn't contain code for this check.

The possible solution could be modification and invoking later the  `StorageBucketPicker` in one of the verification methods.

Additionally, tests don't contain a 'failed' scenario that allows verifying the fix.

runtime-modules/content/src/tests/channels.rs
```diff
@@ -749,3 +749,43 @@ fn channel_censoring() {
         ));
     })
 }
+
+#[test]
+fn channel_creation_doesnt_live_bags_dangling() {
```
Lezek123:
```suggestion
fn channel_creation_doesnt_leave_bags_dangling() {
```

runtime-modules/storage/src/lib.rs
```diff
@@ -2673,34 +2711,58 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
...
 
-    fn ensure_bag_exists(bag_id: &BagId<T>) -> Result<Bag<T>, DispatchError> {
-        Self::ensure_bag_exists(bag_id)
-    }
+        <StorageTreasury<T>>::deposit(
+            &params.deletion_prize_source_account_id,
+            bag_change.total_deletion_prize,
+        )?;
+        //
+        // == MUTATION SAFE ==
```
Lezek123: Isn't `<StorageTreasury<T>>::deposit` already a mutation?

Lezek123 (changes requested):
As @shamil-gadelshin pointed out, this PR doesn't really solve [the issue](https://github.com/Joystream/joystream/issues/2882).

It still can be reproduced, ie. with:
```
yarn storage-node dev:init
yarn storage-node leader:update-bag-limit -l 7 --dev
yarn storage-node leader:update-voucher-limits -o 100 -s 10000000 --dev
yarn storage-node leader:create-bucket -i=0 -a -n=1 -s=100  --dev
yarn storage-node operator:accept-invitation -w=0 -i=0 --dev
yarn storage-node leader:update-dynamic-bag-policy -t Channel -n 1
```

And then from https://github.com/Joystream/joystream/pull/2708:
```
yarn joystream-cli content:createChannel -i ./cli/examples/content/CreateChannel.json --context Member
```

Which causes:
image

The bucket capacity check should happen **before** the bag is created when `create_dynamic_bag_with_objects` is called (and should be part of bucket selection process itself)

dobertRowneySr:
the `reproduce-giza-issue.sh` script now succeeds in creating all 3 channels.
This is because the picking mechanism filters out buckets with not enough capacity (both in size and num. objects). In the reproduction there are 2 buckets (0 & 1), bucket 0 is set to a low capacity (beforehand) and gets filtered out. All the assets for the first 2 channels subsequently created are then placed in bucket 1. Then limits for bucket 0 are enlarged and channel 3 assets can be placed in both 0 or 1.
Tested also with no outstanding buckets large enough. Create channel fails and rightfully so.
~But I now have a problem when re-establishing the bucket size: it still fails to create the channel~ (I had to re-activate the bucket on the reproduction script)

bedeho:
Very nice job ser ;)

runtime-modules/content/src/lib.rs
```diff
@@ -1419,12 +1397,24 @@ impl<T: Trait> Module<T> {
...
+            Storage::<T>::create_dynamic_bag_with_objects(dyn_bag, None, upload_params.clone())?;
+        };
 
-        // attempt to upload objects att
-        Storage::<T>::upload_data_objects(upload_params.clone())?;
+        // remove bag in case of upload error
```
Lezek123: I don't understand this comment, is it a leftover of some earlier approach or am I missing something?

runtime-modules/storage/src/lib.rs
```diff
@@ -2673,34 +2741,59 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
...
```
Lezek123: It looks like this was not fixed. The `== MUTATION SAFE ==` comment should be above `<StorageTreasury<T>>::deposit()`

runtime-modules/storage/src/tests/mod.rs
```diff
@@ -4995,3 +5019,177 @@ fn set_distribution_bucket_family_metadata_fails_with_invalid_distribution_bucke
...
+            dynamic_bag_id,
+            Some(deletion_prize),
+            BTreeSet::from_iter(bag.stored_by),
+            BTreeSet::from_iter(bag.distributed_by),
+        ));
+    });
+}
+
+#[test]
+fn create_dynamic_bag_with_objects_fails_with_no_bucket_availables_with_enough_number_objects() {
```
Lezek123:
```suggestion
fn create_dynamic_bag_with_objects_fails_with_no_buckets_available_with_sufficient_objects_limit() {
```

runtime-modules/storage/src/tests/mod.rs
```diff
@@ -4995,3 +5019,177 @@ fn set_distribution_bucket_family_metadata_fails_with_invalid_distribution_bucke
...
+        CreateDynamicBagWithObjectsFixture::default()
+            .with_bag_id(dynamic_bag_id.clone())
+            .with_deletion_prize(deletion_prize.clone())
+            .with_objects(upload_parameters)
+            .call_and_assert(Err(Error::<Test>::StorageBucketIdCollectionsAreEmpty.into()));
+    })
+}
+
+#[test]
+fn create_dynamic_bag_with_objects_fails_with_no_bucket_availables_with_enough_size() {
```
Lezek123:
```suggestion
fn create_dynamic_bag_with_objects_fails_with_no_buckets_available_with_sufficient_size_limit() {
```

Lezek123 (commented):
Looks like the main issue is fixed :ok_hand: 
I just suggested a few cosmetic changes

runtime-modules/content/src/lib.rs
shamil-gadelshin: Did you consider `is_err` method?

runtime-modules/content/src/lib.rs
```diff
@@ -1419,12 +1397,24 @@ impl<T: Trait> Module<T> {
...
+        let upload_params = UploadParametersRecord {
+            bag_id: bag_id.clone(),
+            object_creation_list: assets.object_creation_list.clone(),
+            deletion_prize_source_account_id: prize_source_account.clone(),
+            expected_data_size_fee: assets.expected_data_size_fee,
+        };
+
+        // create dynamic bag if it doesn't exist (does not emits conflicting created/deleted bag events)
+        if let Err(_) = Storage::<T>::ensure_bag_exists(&bag_id) {
+            Storage::<T>::create_dynamic_bag_with_objects(dyn_bag, None, upload_params.clone())?;
```
shamil-gadelshin: Why do you pass `None` here?

runtime-modules/content/src/lib.rs
```diff
@@ -1419,12 +1397,24 @@ impl<T: Trait> Module<T> {
         channel_id: &T::ChannelId,
         prize_source_account: &T::AccountId,
     ) -> DispatchResult {
```
shamil-gadelshin: `upload_assets_to_storage` is used incorrectly. In most scenarios, it shouldn't create the bag but this possibility exists. Probably, this method should be split for different scenarios.

runtime-modules/content/src/lib.rs
shamil-gadelshin: This comment should be changed or removed.

runtime-modules/storage/src/lib.rs
```diff
@@ -2646,21 +2631,104 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
...
+    }
+
+    fn can_create_dynamic_bag(
+        bag_id: &DynamicBagId<T>,
+        deletion_prize: &Option<DynamicBagDeletionPrize<T>>,
+    ) -> DispatchResult {
+        Self::validate_create_dynamic_bag_params(bag_id, deletion_prize)
+    }
+
+    fn can_create_dynamic_bag_with_objects(
```
shamil-gadelshin: Please, rename it - it's confusing now because it doesn't create bags with objects. `can_create_dynamic_bag_with_object_constraints` could be an option.

runtime-modules/content/src/lib.rs
dobertRowneySr: Yes leftover

shamil-gadelshin (changes requested):
There are several issues with `remove_assets_from_storage` and `upload_assets_to_storage` - both are encountered in the same extrinsics. They are both fallible and should be protected by their `can_*` operations that are part of the storage interface before the `mutation safe` mark. The order should be the same for both operations and for their `can_*` validators  (see `upload_video` for different order). Otherwise, these extrinsics could damage the state. The `update_video` is an example of the unsafe extrinsic.

Also `delete_channel` has an incorrect `mutation-safe` mark - no protection for the storage pallet operations.

`upload_assets_to_storage` should be redesigned because its ability to create bags is incorrect in all but one extrinsic (`create_channel`).

The current implementation depends on the fact that subsequent calls for the random bucket selection return the same result. That is not generally the case: https://github.com/Joystream/joystream/issues/2904 The optimal solution would be to introduce a test that will demand a similar result for those bucket selection calls and fails otherwise. I'm not sure it's part of the current scope and a minimum solution would be to introduce a comment with that issue closer to the bucket selection algorithm.

runtime-modules/content/src/lib.rs
```diff
@@ -674,10 +674,24 @@ decl_module! {
...
+
+                let upload_params = UploadParametersRecord {
+                    bag_id: bag_id.clone(),
+                    object_creation_list: upload_assets.object_creation_list.clone(),
+                    deletion_prize_source_account_id: sender.clone(),
+                    expected_data_size_fee: upload_assets.expected_data_size_fee,
+                };
+
+                // verify that bag exists before hand
+                Storage::<T>::can_create_dynamic_bag_with_objects_constraints(&dyn_bag, &None, &upload_params)?;
```
Lezek123: 1. The comment here seems invalid, the function does not verify whether the bag exists, but whether it can be created
2. I'm not very familiar with the runtime development, but wouldn't that be considerably computionaly wasteful to run this function twice? (first here and then inside `create_dynamic_bag_with_objects_constraints`) @shamil-gadelshin also pointed out that the result of the random bucket selection may be different when its ran the second time.

runtime-modules/content/src/lib.rs
```diff
@@ -1024,6 +1063,21 @@ decl_module! {
...
+                let bag_id = BagIdType::from(dyn_bag.clone());
+
+                let upload_params = UploadParametersRecord {
+                    bag_id: bag_id.clone(),
+                    object_creation_list: upload_assets.object_creation_list.clone(),
+                    deletion_prize_source_account_id: sender.clone(),
+                    expected_data_size_fee: upload_assets.expected_data_size_fee,
+                };
+
+                Storage::<T>::can_upload_data_objects(&upload_params)?;
```
Lezek123: `can_delete_data_objects` check is still missing

Lezek123 (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: `remove_assets_from_storage` just calls `delete_data_objects` from the storage pallet:
``` rust
    fn can_delete_data_objects(
        bag_id: &BagId<T>,
        objects: &BTreeSet<T::DataObjectId>,
    ) -> DispatchResult {
        Self::validate_delete_data_objects_params(bag_id, objects).map(|_| ())
    }

    fn delete_data_objects(
        deletion_prize_account_id: T::AccountId,
        bag_id: BagId<T>,
        objects: BTreeSet<T::DataObjectId>,
    ) -> DispatchResult {
        let bag = Self::ensure_bag_exists(&bag_id)?;

        let bag_change = Self::validate_delete_data_objects_params(&bag_id, &objects)?;
...
```
The check is still performed beforehand. So if `validate_delete_data_objects_params` fails then all the subsequent operations fails, and the storage is not modified.
What I want to ensure is that it cannot be that the first storage operation succeeds and the second operation fails. Hence the check `can_upload_data_parameters`

runtime-modules/content/src/lib.rs
```diff
@@ -674,10 +674,13 @@ decl_module! {
 
             // upload to storage
             if let Some(upload_assets) = params.assets.as_ref() {
+
```
shamil-gadelshin: Invalid formatting here.

runtime-modules/content/src/lib.rs
```diff
@@ -1418,13 +1440,30 @@ impl<T: Trait> Module<T> {
...
+        let upload_params = UploadParametersRecord {
+            bag_id: bag_id.clone(),
+            object_creation_list: assets.object_creation_list.clone(),
+            deletion_prize_source_account_id: prize_source_account.clone(),
+            expected_data_size_fee: assets.expected_data_size_fee,
+        };
+
+        // create dynamic bag if it doesn't exist (does not emits conflicting created/deleted bag events)
+        if channel_creation && Storage::<T>::ensure_bag_exists(&bag_id).is_err() {
```
shamil-gadelshin: Please, move this code to the `create_channel` extrinsic since it is the only place that needs it. It also will highlight the missing check for creation in that extrinsic.

runtime-modules/content/src/lib.rs
```diff
@@ -735,21 +738,40 @@ decl_module! {
...
+
+                Storage::<T>::can_upload_data_objects(&upload_params).map(|_| ())?;
+            }
+
+
             //
             // == MUTATION SAFE ==
             //
 
+            // remove eassets from storage
```
shamil-gadelshin: typo

runtime-modules/content/src/lib.rs
```diff
@@ -674,10 +674,13 @@ decl_module! {
...
             if let Some(upload_assets) = params.assets.as_ref() {
+
+        // upload assets to storage after creating eth bag
                 Self::upload_assets_to_storage(
                     upload_assets,
                     &channel_id,
                     &sender,
+                    true,
                 )?;
             }
```
shamil-gadelshin, line 722: Is it correct?
```
// a newly create channel has zero videos ??
num_videos: 0u64,
```

runtime-modules/storage/src/lib.rs
```diff
@@ -97,6 +97,9 @@
 //! - delete_dynamic_bag
 //! - can_create_dynamic_bag
 //! - create_dynamic_bag
+//! - can_create_dynamic_bag_with_objects_constraints
+//! - create_dynamic_bag_with_objects_constraints
+
```
shamil-gadelshin, line 103: please, add `can_delete_non_empty_dynamic_bag` here

runtime-modules/storage/src/lib.rs
```diff
@@ -2616,6 +2604,20 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
         Self::validate_delete_dynamic_bag_params(bag_id).map(|_| ())
     }
 
+    fn can_delete_non_empty_dynamic_bag(dynamic_bag_id: &DynamicBagId<T>) -> DispatchResult {
+        Self::ensure_dynamic_bag_exists(dynamic_bag_id)?;
+
+        let dynamic_bag = Self::dynamic_bag(dynamic_bag_id);
+
```
shamil-gadelshin: You don't check whether we can delete the data objects here (but just `can_delete_data_objects` is not applicable here because the fees for objects and bags won't be summed in separate methods).

runtime-modules/content/src/lib.rs
```diff
@@ -791,13 +813,18 @@ decl_module! {
...
                 );
 
+                // construct collection of assets to be removed
+                let assets_to_remove = T::DataObjectStorage::get_data_objects_id(&bag_id);
+
+                // delete data objects guard
+                Storage::<T>::can_delete_non_empty_dynamic_bag(
+                    &dyn_bag,
+                )?;
+
```
shamil-gadelshin, line 871: No check for the deleting data objects before the `mutation safe` mark.

runtime-modules/content/src/lib.rs
```diff
@@ -791,13 +813,18 @@ decl_module! {
                     Error::<T>::InvalidBagSizeSpecified
                 );
 
+                // construct collection of assets to be removed
+                let assets_to_remove = T::DataObjectStorage::get_data_objects_id(&bag_id);
+
+                // delete data objects guard
+                Storage::<T>::can_delete_non_empty_dynamic_bag(
```
shamil-gadelshin: Invalid method - no check for combined fees for bag and objects - see a comment from the storage pallet.

runtime-modules/content/src/lib.rs
```diff
@@ -674,10 +674,13 @@ decl_module! {
 
             // upload to storage
             if let Some(upload_assets) = params.assets.as_ref() {
+
+        // upload assets to storage after creating eth bag
                 Self::upload_assets_to_storage(
```
shamil-gadelshin: No checks for dynamic bag creation and object uploading before the `mutation safe` mark.

shamil-gadelshin (changes requested):
There are several issues with the `mutation safe` mark left. 

Is this comment relevant?  Seems like a hidden error.
`// Include Option<AccountId> in ChannelUpdateParameters to update reward_account` . 
Please, remove it if it's obsolete.

This comment  is still not addressed:
```
The current implementation depends on the fact that subsequent calls for the random bucket selection return the same result. That is not generally the case: #2904 The optimal solution would be to introduce a test that will demand a similar result for those bucket selection calls and fails otherwise. I'm not sure it's part of the current scope and a minimum solution would be to introduce a comment with that issue closer to the bucket selection algorithm.
```

runtime-modules/storage/src/lib.rs
dobertRowneySr: so the correct prize should be:
`sizePerMegabyte * object_total_size + bag_deletion_prize`?
@shamil-gadelshin

shamil-gadelshin: Yes. This is a check to verify that sum. They are separately present in `can_delete_data` and `can_delete_bag` but they should be combined to check the sum correctly.

runtime-modules/content/src/lib.rs
```diff
@@ -1473,6 +1477,10 @@ impl<T: Trait> Module<T> {
         Ok(())
     }
 
+    // validate _with bojcets
```
shamil-gadelshin: What is this?

runtime-modules/storage/src/lib.rs
```diff
@@ -122,7 +122,9 @@
 #![allow(clippy::unused_unit)]
 // needed for step iteration over DataObjectId range
 #![feature(step_trait)]
-
+// silence clippy
```
shamil-gadelshin: Why do we silence clippy? Why don't we satisfy it?

runtime-modules/storage/src/lib.rs
```diff
@@ -2835,44 +2818,98 @@ impl<T: Trait> Module<T> {
...
+
+    fn ensure_sufficient_balance_for_upload(
+        deletion_prize_source_account_id: Option<T::AccountId>,
+        required_balance: BalanceOf<T>,
+    ) -> DispatchResult {
+        let usable_balance = deletion_prize_source_account_id.map_or(Zero::zero(), |account_id| {
+            Balances::<T>::usable_balance(account_id)
+        });
+
+        println!(
```
shamil-gadelshin: Please, remove it.

runtime-modules/storage/src/lib.rs
```diff
@@ -2835,44 +2818,98 @@ impl<T: Trait> Module<T> {
...
     fn validate_delete_dynamic_bag_params(
         dynamic_bag_id: &DynamicBagId<T>,
+        with_objects: bool,
     ) -> Result<Option<BalanceOf<T>>, DispatchError> {
         Self::ensure_dynamic_bag_exists(dynamic_bag_id)?;
 
         let dynamic_bag = Self::dynamic_bag(dynamic_bag_id);
 
+        // deletion prize = bag.deletion_prize + total_objects fees if any
+        let deletion_prize = match with_objects {
```
shamil-gadelshin: we should never match bool. Please, use if else in other cases. 

Here you can place ensure in 
```
if !with_objects {
   ensure!()
} 

let deletion_prize = .....
```

runtime-modules/storage/src/lib.rs
```diff
@@ -2835,44 +2818,98 @@ impl<T: Trait> Module<T> {
...
+            false => {
+                // TODO: should this be checked ?
```
shamil-gadelshin: Please, fix or remove TODO.

runtime-modules/content/src/lib.rs
```diff
@@ -1462,7 +1463,10 @@ impl<T: Trait> Module<T> {
         if channel_creation && Storage::<T>::ensure_bag_exists(&bag_id).is_err() {
             Storage::<T>::create_dynamic_bag_with_objects_constraints(
                 dyn_bag,
-                None,
+                Some(storage::DynamicBagDeletionPrize::<T> {
+                    prize: One::one(), // TODO: fix prize
```
shamil-gadelshin: Please, fix the prize.

shamil-gadelshin (changes requested):
It's better. The storage side looks ok, but there are multiple issues left:
- `create_channel` is still unsafe (no check for uploading method).
- create_bag is still within `upload_assets` method
- no comment about the storage picking issue (https://github.com/Joystream/joystream/issues/2904)
- clippy warnings shouldn't be silenced until we have major reasons (Substrate, compiler, third-party libraries, old/new release issues)
- multiple unresolved TODOs

These extrinsics have formatting issues and more than 1 empty line between the code lines:
- create_channel
- update_channel 
- add_curator_to_group

runtime-modules/content/src/lib.rs
```diff
@@ -668,22 +667,43 @@ decl_module! {
...
+                        &Some(deletion_prize.clone()),
+                        &params
+                    )?;
+                }
+
+                //
+                // == MUTATION SAFE ==
+                //
+
+                if Storage::<T>::ensure_bag_exists(&params.bag_id).is_err() {
```
shamil-gadelshin: Don't we need to create a bag even without assets to upload?

runtime-modules/content/src/lib.rs
```diff
@@ -668,22 +667,43 @@ decl_module! {
...
+                        DynamicBagIdType::<T::MemberId, T::ChannelId>::Channel(channel_id),
+                        Some(deletion_prize),
+                        params.clone(),
+                    )?;
+                }
+                // can create dynamic guards ensures that this invocation succeeds
+                Storage::<T>::upload_data_objects(params)?;
             }
 
+
```
shamil-gadelshin, line 722: Please, remove all double empty lines in this method.

runtime-modules/content/src/lib.rs
```diff
@@ -1119,12 +1147,24 @@ decl_module! {
...
             //
             // == MUTATION SAFE ==
             //
 
-            // remove specified assets from channel bag in storage
-            Self::remove_assets_from_storage(&assets_to_remove, &channel_id, &sender)?;
+            if !assets_to_remove.is_empty() {
+                Storage::<T>::delete_data_objects(
+            sender,
```
shamil-gadelshin: Formatting

runtime-modules/content/src/lib.rs
```diff
@@ -1058,35 +1081,40 @@ decl_module! {
...
-                    &sender,
-                    false,
+                    &sender
+                );
+                Storage::<T>::upload_data_objects(params)?;
+            }
+
+            if !params.assets_to_remove.is_empty() {
+                Storage::<T>::delete_data_objects(
+            sender,
```
shamil-gadelshin: Please, fix the formatting.

runtime-modules/content/src/lib.rs
```diff
@@ -1119,12 +1147,24 @@ decl_module! {
             // ensure video can be removed
             Self::ensure_video_can_be_removed(&video)?;
 
+            if !assets_to_remove.is_empty() {
```
shamil-gadelshin, line 1169: Is it ok to delete a video without having the data objects to remove? Is it a correct use-case?

runtime-modules/storage/src/lib.rs
```diff
@@ -2888,21 +2887,20 @@ impl<T: Trait> Module<T> {
...
-                // TODO: should this be checked ?
-                ensure!(
-                    dynamic_bag.objects_number == 0,
-                    Error::<T>::CannotDeleteNonEmptyDynamicBag
-                );
-                dynamic_bag.deletion_prize.unwrap_or(Zero::zero())
-            }
-            true => dynamic_bag
+        let deletion_prize = if !with_objects {
+            // TODO: should this be checked ?
```
shamil-gadelshin: Please, resolve this TODO

runtime-modules/storage/src/lib.rs
```diff
@@ -2826,21 +2827,24 @@ impl<T: Trait> Module<T> {
...
-            Self::check_global_uploading_block()?;
-
-            Self::ensure_objects_creation_list_validity(&params.object_creation_list)?;
+        // call can upload data explicitly
+        // add extra tests in the
+        let bag_change = upload_params
+            .as_ref()
+            .map(|params| {
+                // ensure coherent account ids for prize
+                if let Some(deletion_prize) = deletion_prize {
```
shamil-gadelshin, line 2854: This check doesn't cover all options for incoherent accounts. A test could highlight it.

runtime-modules/storage/src/random_buckets/storage_bucket_picker.rs
```diff
@@ -31,6 +31,7 @@ impl<T: Trait> StorageBucketPicker<T> {
 
         let required_bucket_num = creation_policy.number_of_storage_buckets as usize;
 
+        // TODO: the selection algorithm will change
```
shamil-gadelshin: Please, add the link here https://github.com/Joystream/joystream/issues/2904

shamil-gadelshin (changes requested):
Almost there. 

However, I would recommend adding tests to both storage and contact pallets to ensure that we don't miss different scenarios.

runtime-modules/content/src/lib.rs
ignazio-bovo: If I can add a video with optional assets then I can also delete it even if it has no assets

dobertRowneySr: I suppose yes, If I can add a video with None assets to upload I must be able to remove it

runtime-modules/content/src/lib.rs
```diff
@@ -699,10 +699,27 @@ decl_module! {
...
-            }
+            } else {
+                if Storage::<T>::ensure_bag_exists(&params.bag_id).is_err() {
+                    Storage::<T>::can_create_dynamic_bag(
+                        &DynamicBagIdType::<T::MemberId, T::ChannelId>::Channel(channel_id),
+                        &Some(deletion_prize.clone()),
+                    )?;
+                }
+
+                //
```
shamil-gadelshin: We can't have several `mutation safe` marks. The method became overcomplicated with code duplication. Please, refactor it. It also contains a double empty line - please fix it.

runtime-modules/content/src/lib.rs
```diff
@@ -699,10 +699,27 @@ decl_module! {
                         params.clone(),
                     )?;
                 }
-                // create_dynamic_bag_with_objects with its can* guard ensures that this invocation succeeds
+                // create_dynamic_bag_with_objects with its can* guard ensures that this invocation succee ds
```
shamil-gadelshin: typo: succeeds

runtime-modules/storage/src/lib.rs
```diff
@@ -2815,6 +2815,19 @@ impl<T: Trait> Module<T> {
         }
     }
 
+    fn ensure_upload_and_deletion_prize_account_are_coherent(
+        deletion_prize: &Option<DynamicBagDeletionPrize<T>>,
+        params: &UploadParameters<T>,
+    ) -> DispatchResult {
+        ensure!(deletion_prize.is_some(), Error::<T>::AccountsNotCoherent,);
```
shamil-gadelshin: This check is too strict. It renders useless `Option` of the deletion_prize parameter which contradicts the interface. I was wrong - let's delete this check and allow passing None as dynamic bag deletion prize having only check for a non-empty deletion_prize_account_id.

runtime-modules/content/src/lib.rs
shamil-gadelshin: What's about this comment?

shamil-gadelshin (changes requested):
Current issues:
- `create_channel` extrinsic needs refactoring - it contains two `mutation safe` marks and duplicate code
- the deletion prize account check violates the interface (too strict check) - let's roll it back
- unresolved comment:
```
// a newly create channel has zero videos ??
num_videos: 0u64,
```
-  merge conflict

runtime-modules/content/src/lib.rs
dobertRowneySr: fixed on commit 650b024138

runtime-modules/content/src/lib.rs
dobertRowneySr: fixed on commit 650b024138

runtime-modules/content/src/lib.rs
dobertRowneySr: Nothing really, the only way to increase video counter for a channel is by the `create_video` extrinsic

runtime-modules/storage/src/lib.rs
dobertRowneySr: fixed in fe90c57a04

runtime-modules/content/src/lib.rs
shamil-gadelshin: Mokhtar confirmed that this comment should be deleted.

shamil-gadelshin (approved):
Looks good. 

I'm approving the PR with a condition of the additional PR with extensive testing for added changes in the storage and content pallet. 

`create_channel` still has several formatting errors, including the double empty lines - please fix it.

runtime-modules/content/src/lib.rs
dobertRowneySr: Ok

runtime-modules/storage/src/lib.rs
```diff
@@ -2737,43 +2838,91 @@ impl<T: Trait> Module<T> {
...
             );
-        }
+            dynamic_bag.deletion_prize.unwrap_or_else(Zero::zero)
+        } else {
+            dynamic_bag
+                .deletion_prize
+                .unwrap_or_else(Zero::zero)
+                .saturating_add(Self::calculate_data_storage_fee(
+                    dynamic_bag.objects_total_size,
+                ))
```
Lezek123: This is wrong. When you delete data objects you don't recieve `storage_fee` back, only the deletion prize. So this needs to be the sum of:
- `dynamic_bag.deletion_prize`, plus:
- the sum all `data_object.deletion_prize` for each `data_object` in `dynamic_bag`

Because `storage_fee` is slashed on `upload_data_objects`, this will cause invalid `InsufficientTreasuryBalance` error when `storage_fee` > the sum described above.

You can reproduce it with:
```
docker-compose up -d joystream-node
./query-node/start.sh
./storage-node-v2/scripts/init-dev-bucket.sh
./storage-node-v2/bin/run leader:update-data-fee -f 10 --dev
yarn joystream-cli content:createChannel -i ./cli/examples/content/CreateChannel.json
yarn joystream-cli content:deleteChannel -c 1 --force
```

runtime-modules/content/src/lib.rs
```diff
@@ -970,11 +1045,12 @@ decl_module! {
...
             if let Some(upload_assets) = params.assets.as_ref() {
-                Self::upload_assets_to_storage(
+                let params = Self::construct_upload_parameters(
                     upload_assets,
                     &channel_id,
-                    &sender,
-                )?;
+                    &sender
+                );
+                 Storage::<T>::upload_data_objects(params)?;
```
Lezek123, line 1053: Missing `can_upload_data_objects` before mutation-safe?

runtime-modules/storage/src/lib.rs
```diff
@@ -97,6 +97,10 @@
 //! - delete_dynamic_bag
 //! - can_create_dynamic_bag
 //! - create_dynamic_bag
+//! - can_create_dynamic_bag_with_objects_constraints
+//! - create_dynamic_bag_with_objects_constraints
+//! - can_delete_non_empty_dynamic_bag
```
```suggestion
//! - can_delete_dynamic_bag_with_objects
```

Lezek123 (changes requested):
Additional issues:
- `chain-metadata.json` needs to be updated

runtime-modules/content/src/lib.rs
dobertRowneySr: Checks for `upload_data_objects` are also inside the `can_create_bag_with_objects`. Plus`can_upload_data_objects` throws if the bag does not exist. So I cannot use it before the bag is created

runtime-modules/storage/src/lib.rs
dobertRowneySr: fixed in f2fbb1a993

runtime-modules/storage/src/lib.rs
dobertRowneySr: fixed in 7344c8e218

runtime-modules/content/src/lib.rs
dobertRowneySr: added a comment in 5dc18dc1a4

Lezek123: The comment was about `create_video`, not `create_channel`

dobertRowneySr: In the `create_video` extrinsic the `upload_data_objects` verifies parameters before uploading. so It will not leave content/storage state in a non valid status.
I will add a comment

Lezek123 (approved):
LGTM

PR #2908, opened 2021-12-09
Giza query node: Support new distribution bucket ids

Lezek123:
Update query node following runtime changes in https://github.com/Joystream/joystream/pull/2897


shamil-gadelshin (approved):
Looks good and works.

I have a concern about whether a concatenated string is a good ID for complex keys. It will work OK on a small number of buckets but will become less performant if we increase that number. But probably we could optimize it later.

PR #2918, opened 2021-12-14
query node - pioneer update

thesan:
This PR addresses:
- https://github.com/Joystream/pioneer/issues/1536
- https://github.com/Joystream/pioneer/issues/1741
- https://github.com/Joystream/pioneer/issues/1524

### DIFF
--- a/query-node/mappings/src/council.ts
+++ b/query-node/mappings/src/council.ts
@@ -171,7 +171,7 @@ async function getAccountCastVote(
 
   if (!castVote) {
     throw new Error(
-      `No vote cast by the given account in the curent election round. accountId '${account}', cycleId '${electionRound?.cycleId}'`
+      `No vote cast by the given account in the current election round. accountId '${account}', cycleId '${electionRound?.cycleId}'`
     )
   }
 
@@ -315,7 +315,7 @@ export async function council_AnnouncingPeriodStarted({ event, store }: EventCon
 }
 
 /*
-  The event is emitted when a candidacy announcment period has ended, but not enough members announced.
+  The event is emitted when a candidacy announcement period has ended, but not enough members announced.
 */
 export async function council_NotEnoughCandidates({ event, store }: EventContext & StoreContext): Promise<void> {
   // common event processing
@@ -381,7 +381,7 @@ export async function council_NewCandidate({ event, store }: EventContext & Stor
 
   const electionRound = await getCurrentElectionRound(store)
 
-  // prepare note metadata record (empty until explicitily set via different extrinsic)
+  // prepare note metadata record (empty until explicitly set via different extrinsic)
   const noteMetadata = new CandidacyNoteMetadata({
     bulletPoints: [],
   })
@@ -398,7 +398,7 @@ export async function council_NewCandidate({ event, store }: EventContext & Stor
     stakeLocked: true,
     votePower: new BN(0),
     noteMetadata,
-    votesRecieved: [],
+    votesReceived: [],
   })
   await store.save<Candidate>(candidate)
 
@@ -426,7 +426,7 @@ export async function council_NewCouncilElected({ event, store }: EventContext &
 
   // specific event processing
 
-  // mark old council as resinged
+  // mark old council as resigned
   const oldElectedCouncil = await getCurrentElectedCouncil(store)
   oldElectedCouncil.isResigned = true
   oldElectedCouncil.endedAtBlock = event.blockNumber
@@ -589,7 +589,7 @@ export async function council_CandidacyNoteSet({ event, store }: EventContext & 
 
   const [memberId, note] = new Council.CandidacyNoteSetEvent(event).params
 
-  // load candidate recored
+  // load candidate recorded
   const electionRound = await getCurrentElectionRound(store)
   const candidate = await getCandidate(store, memberId.toString(), electionRound, ['noteMetadata'])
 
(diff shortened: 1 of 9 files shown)


query-node/mappings/src/council.ts
```diff
@@ -48,6 +48,10 @@ import {
   ElectedCouncil,
   CastVote,
   CandidacyNoteMetadata,
+  CandidacyStatusActive,
+  CandidacyStatusWithdrawn,
+  CandidacyStatusElected,
+  CandidacyStatusLost,
```
Lezek123: It seems like `CandidacyStatusLost` got _lost_ during some conflict resolutions(?).
If I compere this PR branch to `thesan/olympia` it looks like there is some code missing in `council_NewCouncilElected` handler.

query-node/mappings/src/council.ts
```diff
@@ -433,11 +436,15 @@ export async function council_NewCouncilElected({ event, store }: EventContext &
   // get election round and its candidates
   const electionRound = await getCurrentElectionRound(store)
 
-  // TODO: uncomment when following query will be working (after some QN patches make it to Olympia)
-  // const electedCandidates = await store.getMany(Candidate, { where: { electionRoundId: electionRound.id, member: { id_in: electedMemberIds } } })
-  const electedCandidates = (
-    await store.getMany(Candidate, { where: { electionRoundId: electionRound.id }, relations: ['member'] })
-  ).filter((item: Candidate) => electedMemberIds.find((tmpId) => tmpId === item.member.id.toString()))
+  const where = { electionRoundId: electionRound.id, status_json: { isTypeOf_eq: 'CandidacyStatusActive' } }
```
Lezek123: This doesn't look like a valid TypeORM where condition.
Note that we're not querying the graphql server here, but the PostgreSQL database instead.
To query by `status` you'd need a condidion like:
```suggestion
  const where = {
    electionRoundId: electionRound.id,
    status: Raw((alias) => `${alias}::json->'isTypeOf' = :status`, {
      status: 'CandidacyStatusActive',
    })
  }
```
Where `Raw` is:
```typescript
import { Raw } from 'typeorm'
```

query-node/mappings/src/council.ts
```diff
@@ -433,11 +436,15 @@ export async function council_NewCouncilElected({ event, store }: EventContext &
...
+  const candidates = await store.getMany(Candidate, { where })
```
Lezek123: Because later you're using `candidate.member.id`, I think you need to eagerly load a member
```suggestion
  const candidates = await store.getMany(Candidate, { where }, relations: ['member'])
```
Or to use a "hackier", but more optimal way: replace `candidate.member.id` with:
```typescript
(candidate as Candidate & { memberId: string }).memberId
```

query-node/mappings/src/council.ts
```diff
@@ -433,11 +436,15 @@ export async function council_NewCouncilElected({ event, store }: EventContext &
...
+
+  const electedCandidates = candidates.filter((candidate) => electedMemberIds.includes(candidate.member.id))
+
+  // Set candidates elected status
+  electedCandidates.forEach((candidate) => {
+    candidate.status = new CandidacyStatusElected()
```
Lezek123: I think you can put this under `councilMember.electedInCouncil = electedCouncil`

query-node/schemas/council.graphql
```diff
@@ -46,6 +46,36 @@ enum ElectionProblem {
...
+}
+
+type CandidacyStatusLost @variant {
+  # no properties
+
+  # TODO: remove me - variant needs to have at least 1 property now
+  dummy: Int
+}
+
+union CandidacyStatus = CandidacyStatusActive | CandidacyStatusWithdrawn | CandidacyStatusElected | CandidacyStatusLost
```
Lezek123: Since all variants have no additional properties, perhaps using `enum` for `CandidacyStatus` would make things simpler.

query-node/schemas/membership.graphql
```diff
@@ -1,9 +1,21 @@
...
+
+union Avatar = AvatarObject | AvatarUri
+
 type MemberMetadata @entity {
   "Member's name"
   name: String
 
   "Avatar data object"
-  avatar: DataObject
+  avatar: Avatar
```
Lezek123, line 18: This will require changes in `tests/integration-tests`, but I can take care of that.

Lezek123 (changes requested):
Additional note: Do you want to change the status of candidates on `council_NotEnoughCandidates` / `council_NewCouncilNotElected`? Currently in those cases it seems like the status of the candidates will remain `Active`, even though a brand new election is started.

query-node/mappings/src/council.ts
thesan: Sorry about that I'm not familiar with TypeORM, I wrote that as a placeholder and forgot about it :facepalm:

thesan:
@Lezek123 thanks for the review. I had completely forgotten about the `council_NotEnoughCandidates` / `council_NewCouncilNotElected`. Also using `enum` for `CandidacyStatus` solves a few problems in itself.
Finally I didn't manage fix the integration tests, could you have a look please?

Lezek123:
> Finally I didn't manage fix the integration tests, could you have a look please?

Looks like there weren't actually any changes needed related to member avatar, as it just currently ignored in the integration tests.
The tests were failing because of another `candidate.member.id` reference, created a PR to your branch with a fix: https://github.com/thesan/joystream/pull/3

Lezek123 (approved):
Great work!

PR #2849, opened 2021-11-23
Giza integration tests: Reusable functionality

Lezek123:
Partially implements https://github.com/Joystream/joystream/issues/2846
- support for typesafe queries in the integration tests framework
- support for executing storage CLI, distributor CLI and joystream CLI commands from within the integration tests flows / fixtrues
- flows to quickly initialize storage and distribution system to a given state (creating buckets, setting the limits, dynamic bag policy etc.)

Tests that have been added (as part of [full new chain test](https://github.com/Joystream/joystream/pull/2849/files#diff-7b6f74331bfd37c9fcfbd0ad2fbee8eb1111e286da0c2e7f8a6bd729cb7be58f))
- basic storage CLI test (initializing storage bucket)
- basic distributor CLI test (initializing distributor bucket)
- creating channel using Joystream CLI (along with uploading data objects) and checking the result in the query node

Other changes:
- Updated yarn to `1.22.15` to fix and issue with broken package resultion
- Some minor CLI fixes
- Some changes in the framework - typesafe retrieval of events in `Api.ts`, ability to retrieve suri by address (needed in order to support CLIs), preserving _custom_ keys between scenarios etc.

PR #2935, opened 2021-12-16
Giza transaction nonce fix

shamil-gadelshin:
We need to introduce an additional account ID for the storage providers to prevent transaction nonce conflicts.

### Affected components
- runtime change of the `accept_storage_bucket_invitation` extrinsic and related code
- storage-node support
- types
- query-node mappings


storage-node-v2/scripts/init-dev-bucket.sh
```diff
@@ -13,6 +13,6 @@ ${CLI} dev:init
 ${CLI} leader:update-bag-limit -l 7 --dev
 ${CLI} leader:update-voucher-limits -o 10000 -s 1000000000 --dev
 BUCKET_ID=`${CLI} leader:create-bucket -i=0 -a -n=10000 -s=1000000000  --dev` 
-${CLI} operator:accept-invitation -w=0 -i=${BUCKET_ID} --dev
+${CLI} operator:accept-invitation -w=0 -i=${BUCKET_ID} --dev -t=5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
```
Lezek123, line 16: Similar update is needed for `run-all-commands.sh` script

Lezek123 (approved):
Looks good, I've 2 additional comments:

- I was wondering about the case when multiple storage buckets are assigned to the same worker. Technically it's possible, although I suppose it's not recommended. Currently Colossus will try to fulfill all obligations by `workerId`, which means the worker will need to make sure he uses the same transactor account for all buckets.

- Since you already updated storage-node-v2, you could also run `yarn workspace storage-node-v2 generate:types:graphql` to update the graphql schema typings, however when I tried to do this locally I encountered:
    ```
      code: 'MODULE_NOT_FOUND',
      requireStack: [
        '/home/leszek/projects/joystream/joystream-ws-3/node_modules/graphql-config/node_modules/@graphql-tools/url-loader/index.js',
        '/home/leszek/projects/joystream/joystream-ws-3/node_modules/graphql-config/index.cjs.js',
        '/home/leszek/projects/joystream/joystream-ws-3/node_modules/@graphql-codegen/cli/bin.js'
      ]
    ```
    I already dealt with similar issue on `olympia-staging` branch. It seems like a bug in yarn package resolution, I fixed it by updating `yarn` to `1.22.15` (see: https://github.com/Joystream/joystream/pull/2849)

PR #2938, opened 2021-12-17
Fix update_families_in_dynamic_bag_creation_policy extrinsic.

shamil-gadelshin:
Add default storage bucket number for the new policy.


Lezek123 (approved):
LGTM, the fix works as expected

PR #2939, opened 2021-12-17
Giza: Storage node - folder structure, readme, docker, file logging.

shamil-gadelshin:
### Changes
- renamed `storage-node-v2` folder to `storage-node`
- updated the README
- added file logger and refactored the logging service
- changed the Docker files


storage-node/src/services/logger.ts
```diff
@@ -210,6 +188,24 @@ function createElasticTransport(logSource: string, elasticSearchEndpoint: string
...
+ *
+ * @param fileName - log file name.
+ * @returns winston file transport
+ */
+function createFileTransport(filename: string): winston.transport {
+  const options = {
+    filename,
+    maxsize: 50000000, // 50 MB
+    maxFiles: 3,
+    level: 'debug',
```
Lezek123: Shouldn't those be easily configurable? The values seem kind of arbitraty.

storage-node/src/services/webApi/app.ts
```diff
@@ -56,7 +57,8 @@ export async function createApp(config: AppConfig): Promise<Express> {
     })
   ) // Required signature.
 
-  app.use(errorLogger())
+  // Error logger
+  app.use(errorLogger(expressLoggerOptions))
```
Lezek123: I'm not sure if it's a good idea to use the same `level` for successful and failed requests.
The default `level` for `expressWinston.errorLogger` is `error`, but here you're overriding it with `http`

storage-node/src/services/logger.ts
```diff
@@ -210,6 +188,24 @@ function createElasticTransport(logSource: string, elasticSearchEndpoint: string
...
+    format: ecsformat(),
+  }
+  return new winston.transports.File(options)
```
Lezek123: For distributor node I used [`winston-daily-rotate-file` transport](https://github.com/Lezek123/substrate-runtime-joystream/blob/distributor-node/distributor-node/src/services/logging/LoggingService.ts#L102), because I noticed a strange behavior when using the default `winston.transports.File` with both `maxsize` and `maxFiles` set.

To ilustrate this: I changed the `maxsize` for Colossus to `10KB`, gathered some logs and then ran `distributor-node dev:batchUpload` script to execute 200 uploads at once. The result is as follows:
image
First there is `log.txt` (generated before I decreased the maxsize) and `log1.txt`. You would then expect `log.txt`, `log1.txt` and `log2.txt`, but next time I'm checking the logs (while the `batchUpload` is executing), there is only `log1.txt`. Then when the `batchUpload` finishes executing we suddenly have `log275.txt` out of nowhere.

This matches the experience I had when testing distributor node file logging, I think `winston.transports.File` is just not a reliable transport for this use-case.

Lezek123 (changes requested):

storage-node/src/services/logger.ts
shamil-gadelshin: I also experimented with edge cases but didn't encounter such behavior. But it's good that you noticed it - I copied most part of the implementation from Argus (https://github.com/Joystream/joystream/pull/2939/commits/2f527e86e1c34fa5a2fcd610cb0bdf9c48e0b8db)

storage-node/src/services/webApi/app.ts
shamil-gadelshin: Fixed here (https://github.com/Joystream/joystream/commit/2f527e86e1c34fa5a2fcd610cb0bdf9c48e0b8db)

storage-node/src/services/logger.ts
shamil-gadelshin: Fixed here (https://github.com/Joystream/joystream/commit/2f527e86e1c34fa5a2fcd610cb0bdf9c48e0b8db)

storage-node/src/services/logger.ts
```diff
@@ -82,16 +85,27 @@ const proxy = new Proxy(InnerLogger, {
...
 }
 
+/**
+ * Creates Express-Winston error logger options.
+ *
+ */
+export function createExpressErrorLoggerOptions(): expressWinston.LoggerOptions {
+  return {
+    winstonInstance: proxy,
+    level: 'error',
```
Lezek123, line 105: After some testing I suggest making a small change to this config. It seems possible to include some additional information in the error message (like `err.message` or `res.statusCode`) [by setting the value of `msg` config parameter](https://www.npmjs.com/package/express-winston/v/2.6.0#options-1). The default value for `msg` seems to be `middlewareError`, which doesn't say much about the error when you go through the logs in console / elasticsearch:
image
image

Lezek123 (commented):
I just added one improvement suggestion, but otherwise everything seems to work as expected.
There is also a small `yarn.lock` conflict left to resolve.

PR #2945, opened 2021-12-17
query node - bounty schema and mappings

thesan:
Closes https://github.com/Joystream/pioneer/issues/1948
Closes https://github.com/Joystream/pioneer/issues/1955

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703952879841/1201704775392491) by [Unito](https://www.unito.io)


thesan:
For now this is a first draft so we get started working with MirageJs mocks next week.

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,190 @@
...
+  judgingPeriod: Int!
+
+  # Bounty current state:
+  "Current bounty stage"
+  stage: BountyStage!
+
+  totalFunding: BigInt!
+
+  "Bounty's discussion thread"
+  discussionThread: ForumThread!
```
thesan: According to the design there is a discussion thread under the "Bounty tab". But the extrinsics don't refer to this. Should a forum thread be created before calling `createBounty` and the threadId be stored in the metadata ?

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,190 @@
...
+
+  "Staking account with the work entry stake"
+  stakingAccount: String
+
+  # Work entry current state:
+  "Whether at least one work has been submitted"
+  workSubmitted: Boolean!
+
+  # TODO find an acceptable structure for this
+  works: [String]!
```
thesan: Any idea on how to store the works ?

query-node/schemas/bounty.graphql
jodator: The extrinsic will not have this for sure. The metadata should have this.

AFAIR there should be a predefined forum category in which we would create a thread. How to get this is another story for sure ;)

query-node/schemas/bounty.graphql
jodator: Do you mean work entry?

thesan: Kind of. Sometime work entry refers to the data created by `submitWork` and sometime to the data created by `announceWorkEntry`. Here I meant the data created by `submitWork`.

However I managed to clarify this a bit: works will be markdown strings which might contain hyperlinks to the actual work. So I should simply replace `String` by an object (e.g `BountyWorkData`) containing a title and a description.
(I originally got confused by [this line](https://github.com/Joystream/joystream/blob/0e05e05f964e317fb4d40d374e44b373579a5b37/runtime-modules/bounty/src/lib.rs#L566))

query-node/schemas/bounty.graphql
jodator: BTW: #2961

jodator (commented):
I'd say good first step to build on :+1:

query-node/mappings/src/scheduler.ts
```diff
@@ -0,0 +1,56 @@
+import { In } from 'typeorm'
+import { DatabaseManager, StoreContext } from '@joystream/hydra-common'
+import { eventEmitter, ProcessorEvents } from '@joystream/hydra-processor/lib/start/processor-events'
```
thesan: @Lezek123, @ondratra because the runtime doesn't implement the events necessary to map the transition between funding, work, judgment, and failed stages, I have added this "scheduler" logic based on the hydra `eventEmitter`. According to you would this be an acceptable solution ?

mnaamani:
The query node does not build (When I merge it into olympia-playground branch)

`Error: src/index.ts(17,15): error TS2307: Cannot find module './genesis' or its corresponding type declarations.`

Looks like the branch is a little bit dated. Could you rebase it from `olympia-playground` branch? 

I think the runtime at that point has the latest bounty work, and its the current version deployed for the bounty playground testing.

mnaamani:
Hey @theo I merged andy's PR and did a quick merge conflict fix on this PR to merge it as well.
There is a small break with the bounty types:

```
Error: src/bounty.ts(243,61): error TS2339: Property 'map' does not exist on type 'AssuranceContractType_Closed'.
```

That type has been reverted back to a BTreeSet instead of a Vec, so the `.map` function doesn't exist anymore.

mnaamani:
To fix the typescript formatting causing linter checks to fail https://github.com/Joystream/joystream/runs/5287300293?check_suite_focus=true#step:4:1356

just run `yarn format` in `query-node/`

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,483 @@
...
+  store: DatabaseManager,
+  event: SubstrateEvent,
+  bountyId: BountyId,
+  changes: (bounty: Bounty) => Partial<Bounty>
+) {
+  const bounty = await getBounty(store, bountyId)
+  bounty.updatedAt = new Date(event.blockTimestamp)
+  Object.assign(bounty, changes(bounty))
+
+  await store.save<BountyEntry>(bounty)
```
Lezek123:
```suggestion
  await store.save<Bounty>(bounty)
```

query-node/mappings/src/scheduler.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function scheduleAtBlock(blockNumber: number, job: () => void): void {
+  toBeScheduled.push([blockNumber, job])
+}
+
+function runScheduler(): void {
+  isSchedulerRunning = true
+  const scheduleRecord: { [n: number]: (() => void)[] } = {}
+
+  eventEmitter.on(ProcessorEvents.INDEXER_STATUS_CHANGE, (indexerStatus) => {
```
Lezek123: There is a standarized way to execute an action before/after a block is processed in the Hydra framework, which is by using `preBlockHooks` / `postBlockHooks` as described in https://docs.subsquid.io/v/v3/mappings/mapping-filters#mapping-filter

If there are any problems with using `preBlockHooks` / `postBlockHooks` for the execution of scheduled tasks, they should probably be fixed on the Hydra side.

The reason why this workaround is not acceptable is because it's executing some database tasks outside of the standard processing loop (by listening to `INDEXER_STATUS_CHANGE` event). With Hydra processor we need to be sure that each event is beeing processed separately and there is nothing that affects the database "from the outside" during this time. If this assumption is broken it can lead to many unexpected & undeterministic behaviors. Also, note that indexer runs independently from the processor, so `indexer.head` and especially `indexer.chainHeight` is changing way in advance compared to processor's `lastCompleteBlock` (last processed block), so the job scheduled to execute at block `150` may actually execute while block `140` is still beeing processed etc.

query-node/mappings/src/scheduler.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (toBeScheduled.length) {
+      toBeScheduled.forEach(([blockNumber, job]) => {
+        if (blockNumber < indexerStatus.chainHeight) {
+          job()
```
Lezek123: `job` is esentially an async task (the way it's currently used, it performs some database operations), so it's risky to execute all jobs at once without `await`ing the result of the previous ones, because we cannot then guarantee in which order the operations will be executed.

For example, because the processor uses:
```
for ... {
  await ...
}
```
syntax when processing the events, we can be sure that each event is processed after all the previous ones have been already processed. I would expect the scheduler to work the same way: we execute all the tasks in the deterministic order, ie.:
- tasks scheduled for older blocks have priority over tasks scheduled for newer blocks, and then:
- tasks scheduled first have priority over tasks scheduled later

query-node/mappings/src/scheduler.ts
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+          scheduleRecord[blockNumber] = [...(scheduleRecord[blockNumber] ?? []), job]
+        }
+      })
+      toBeScheduled = []
+    }
+
+    if (scheduleRecord[indexerStatus.chainHeight]) {
+      scheduleRecord[indexerStatus.chainHeight].forEach((job) => job())
```
Lezek123: The previous comment also applies here.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+ */
+
+export function bountyScheduleFundingEnd(store: DatabaseManager, bounty: Bounty): void {
+  const { fundingType } = bounty
+  if (bounty.stage !== BountyStage.Funding || !('fundingPeriod' in fundingType)) return
+
+  const fundingPeriodEnd = bounty.createdInEvent.inBlock + fundingType.fundingPeriod
+  scheduleAtBlock(fundingPeriodEnd, () => {
+    if (bounty.stage === BountyStage.Funding) {
+      const isFunded = bounty.totalFunding >= fundingType.minFundingAmount
```
Lezek123: Those condidtions will be checked in the future, but you're not reloading the bounty data in any way, you're using a stale `bounty` object which will not have an up-to-date state once this scheduled job is executed.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+    }
+  })
+}
+
+export function bountyScheduleWorkSubmissionEnd(store: DatabaseManager, bounty: Bounty): void {
+  if (bounty.stage !== BountyStage.WorkSubmission) return
+
+  const workingPeriodEnd = fundingPeriodEnd(bounty) + bounty.workPeriod
+  scheduleAtBlock(workingPeriodEnd, () => {
+    if (bounty.stage === BountyStage.WorkSubmission) {
```
Lezek123: Previous comment also applies here

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+}
+
+export function bountyScheduleJudgementEnd(store: DatabaseManager, bounty: Bounty): void {
+  if (bounty.stage !== BountyStage.Judgment) return
+
+  const judgementPeriodEnd = fundingPeriodEnd(bounty) + bounty.workPeriod + bounty.judgingPeriod
+  scheduleAtBlock(judgementPeriodEnd, () => {
+    if (bounty.stage === BountyStage.Funding) {
+      bounty.updatedAt = new Date()
+      bounty.stage = BountyStage.Failed
```
Lezek123: Previous comment also applies here.
Also, is that even the right condition? Shouldn't it be `bouty.stage === BountyStage.Judgment`?

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+    description: metadata?.description ?? undefined,
+    bannerImageUri: metadata?.bannerImageUri ?? undefined,
+    cherry: bountyParams.cherry,
+    entrantStake: bountyParams.entrant_stake,
+    creator: bountyActorToMembership(bountyParams.creator),
+    oracle: bountyActorToMembership(bountyParams.oracle),
+    fundingType: asFundingType(bountyParams.funding_type),
+    contractType: asContractType(bountyParams.contract_type),
+    workPeriod: bountyParams.work_period.toNumber(),
+    judgingPeriod: bountyParams.judging_period.toNumber(),
```
Lezek123: As recently discovered, there are issues with inserting `u32` numbers into `Int` database columns, especially when the `u32` is a custom user input and can be > than `2147483647`

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+    stage: BountyStage.Funding,
+    totalFunding: bountyParams.cherry,
+    discussionThread: whenDef(metadata?.discussionThread, (id) => new ForumThread({ id })),
```
Lezek123: This allows anyone to crash the query node by providing invalid thread id in the metadata, the mappings should be secured against that.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  function asFundingType(funding: FundingType) {
+    if (funding.isPerpetual) {
+      const perpetualFunding = new BountyFundingPerpetual()
+      perpetualFunding.target = funding.asPerpetual.target
+      return perpetualFunding
+    } else {
+      const limitedFunding = new BountyFundingLimited()
+      limitedFunding.maxFundingAmount = funding.asLimited.max_funding_amount
+      limitedFunding.minFundingAmount = funding.asLimited.min_funding_amount
+      limitedFunding.fundingPeriod = funding.asLimited.funding_period.toNumber()
```
Lezek123: Unsafe `u32` => `Int`

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+  function asFundingType(funding: FundingType) {
+    if (funding.isPerpetual) {
+      const perpetualFunding = new BountyFundingPerpetual()
+      perpetualFunding.target = funding.asPerpetual.target
+      return perpetualFunding
+    } else {
+      const limitedFunding = new BountyFundingLimited()
+      limitedFunding.maxFundingAmount = funding.asLimited.max_funding_amount
+      limitedFunding.minFundingAmount = funding.asLimited.min_funding_amount
```
Lezek123: There are currently some issues with assigning `Codec` values (ie. `u128`) like that directly into variant properties. This has not yet been fixed in Hydra and it causes an error like:
```
processor                | ERROR: Stopping the proccessor due to errors: {} name: TypeError, message: attempted to use private field on non-instance, stack: TypeError: attempted to use private field on non-instance
processor                |     at _classPrivateFieldBase (/joystream/node_modules/@babel/runtime/helpers/classPrivateFieldLooseBase.js:3:11)
processor                |     at u128.bitLength (/joystream/node_modules/@polkadot/types/codec/AbstractInt.cjs:137:53)
processor                |     at u128.toHex (/joystream/node_modules/@polkadot/types/codec/AbstractInt.cjs:186:23)
processor                |     at u128.toJSON (/joystream/node_modules/@polkadot/types/codec/AbstractInt.cjs:223:66)
processor                |     at JSON.stringify (<anonymous>)
processor                |     at PostgresDriver.preparePersistentValue (/joystream/node_modules/typeorm/driver/postgres/PostgresDriver.js:497:25)
processor                |     at /joystream/node_modules/typeorm/query-builder/InsertQueryBuilder.js:399:57
processor                |     at Array.forEach (<anonymous>)
processor                |     at /joystream/node_modules/typeorm/query-builder/InsertQueryBuilder.js:379:25
processor                |     at Array.forEach (<anonymous>)

```

I suggest using:
```suggestion
      limitedFunding.maxFundingAmount = new BN(funding.asLimited.max_funding_amount.toString())
      limitedFunding.minFundingAmount = new BN(funding.asLimited.min_funding_amount.toString())
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+// Store BountyFundingWithdrawal events (also update the contribution deleteAt time)
+export async function bounty_BountyFundingWithdrawal({ event, store }: EventContext & StoreContext): Promise<void> {
+  const fundingWithdrawalEvent = new BountyEvents.BountyFundingWithdrawalEvent(event)
+  const [bountyId, contributorActor] = fundingWithdrawalEvent.params
+  const eventTime = new Date(event.blockTimestamp)
+
+  // Update the contribution
+  const contributor = bountyActorToMembership(contributorActor)
+  const contribution = await getContribution(store, bountyId, contributor?.id)
```
Lezek123, line 336: You're creating a new `BountyContribution` each time `BountyFunded` event is fired, but the runtime allows a single member to contribute multiple times (see: https://github.com/Joystream/joystream/blob/olympia/runtime-modules/bounty/src/lib.rs#L902), so there may be many contributions per member. Here you are fetching only a single record.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+  // Record the event
+  const entry = new BountyEntry({ id: String(entryId) })
+  const slashedInEvent = new WorkEntrySlashedEvent({ ...genericEventFields(event), entry })
+  await store.save<WorkEntrySlashedEvent>(slashedInEvent)
+}
+
+// Store WorkSubmitted events
+export async function bounty_WorkSubmitted({ event, store }: EventContext & StoreContext): Promise<void> {
+  const entrySlashedEvent = new BountyEvents.WorkEntrySlashedEvent(event)
```
Lezek123: Wrong event beeing used here (`WorkEntrySlashedEvent` instead of `WorkSubmittedEvent`)

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,226 @@
...
+  Terminated
+}
+
+type Bounty @entity {
+  # Bounty creation parameters:
+  "Bounty's runtime id"
+  id: ID!
+
+  "Bounty title"
+  title: String!
```
Lezek123: Should be optional, otherwise providing no title results in:
```
processor                | ERROR: Stopping the proccessor due to errors: {
processor                |   "message": "null value in column \"title\" violates not-null constraint",
```

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,226 @@
...
+
+  "Bounty description"
+  description: String!
```
Lezek123: Should be optional, same as `title`

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,226 @@
...
+
+  # Bounty current state:
+  "Current bounty stage"
+  stage: BountyStage!
+
+  "Total amount currently available for the reward"
+  totalFunding: BigInt!
+
+  "Bounty discussion thread"
+  discussionThread: ForumThread!
```
Lezek123: Should be optional (metadata may not contain any ref)

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+/**
+ * Schedule Periods changes
+ */
+
+export function bountyScheduleFundingEnd(store: DatabaseManager, bounty: Bounty): void {
+  const { fundingType } = bounty
+  if (bounty.stage !== BountyStage.Funding || !('fundingPeriod' in fundingType)) return
+
+  const fundingPeriodEnd = bounty.createdInEvent.inBlock + fundingType.fundingPeriod
```
Lezek123: `createdInEvent` is undefined here:
```
processor                | ERROR: Stopping the proccessor due to errors: {} name: TypeError, message: Cannot read property 'inBlock' of undefined, stack: TypeError: Cannot read property 'inBlock' of undefined
processor                |     at bountyScheduleFundingEnd (/joystream/query-node/mappings/lib/src/bounty.js:77:52)
processor                |     at bounty_BountyCreated (/joystream/query-node/mappings/lib/src/bounty.js:165:5)
```

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,226 @@
...
+  # no properties
+
+  # TODO: remove me - variant needs to have at least 1 property now
+  dummy: Int
+}
+
+"Only preselected members can submit work entries to the bounty"
+type BountyContractClosed @variant {
+  # Members authorized to submit entries to the bounty
+  whitelist: [Membership!]!
```
Lezek123: You shouldn't use one-to-many directly inside a variant like that, this is not supported and results in strange errors when trying to query the bounties.

Even a simple query like:
```
bounties { id }
```

Results in:
```
  "errors": [
    {
      "message": "Cannot read property 'whitelistIds' of undefined",
```

For reference implementation, see: https://github.com/Joystream/joystream/blob/olympia/query-node/schemas/proposalDiscussion.graphql#L13

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  bountyScheduleFundingEnd(store, bounty)
+
+  // Record the event
+  const createdInEvent = new BountyCreatedEvent({ ...genericEventFields(event), bounty })
+  await store.save<BountyCreatedEvent>(createdInEvent)
+
+  function asFundingType(funding: FundingType) {
+    if (funding.isPerpetual) {
+      const perpetualFunding = new BountyFundingPerpetual()
+      perpetualFunding.target = funding.asPerpetual.target
```
Lezek123:
```suggestion
      perpetualFunding.target = new BN(funding.asPerpetual.target.toString())
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+function bountyActorToMembership(actor: BountyActor): Membership | undefined {
+  if (actor.isMember) {
+    return new Membership({ id: String(actor.asMember) })
+  }
+}
+
+function fundingPeriodEnd(bounty: Bounty): number {
+  return (
+    bounty.maxFundingReachedEvent?.inBlock ??
+    bounty.createdInEvent.inBlock + (bounty.fundingType as BountyFundingLimited).fundingPeriod
```
Lezek123: In order for `createdInEvent.inBlock` to be available here you need to query the bounty along with the event, ie.:
```
await store.get(Bounty, { where: { id: bountyId }, relations: ['createdInEvent'] })
```
Otherwise it's `undefined` and throws an error:
```
processor                | ERROR: Stopping the proccessor due to errors: {} name: TypeError, message: Cannot read property 'inBlock' of undefined, stack: TypeError: Cannot read property 'inBlock' of undefined
processor                |     at fundingPeriodEnd (/joystream/query-node/mappings/lib/src/bounty.js:64:27)
processor                |     at bountyScheduleWorkSubmissionEnd (/joystream/query-node/mappings/lib/src/bounty.js:89:30)
processor                |     at endFundingPeriod (/joystream/query-node/mappings/lib/src/bounty.js:115:9)
processor                |     at bounty_BountyMaxFundingReached (/joystream/query-node/mappings/lib/src/bounty.js:241:11)
```

query-node/schemas/bountyEvents.graphql
```diff
@@ -0,0 +1,341 @@
...
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
+
+  "The entry this work was submitted into"
+  entry: BountyEntry!
+
+  "Title of the work"
+  title: String!
```
Lezek123: Nothing is inserted here, which causes:
```
processor                | ERROR: Stopping the proccessor due to errors: {
processor                |   "message": "null value in column \"title\" violates not-null constraint",
```

query-node/schemas/bountyEvents.graphql
```diff
@@ -0,0 +1,341 @@
...
+
+  "Description which contains the work itself as a URL, a BLOB, or just text"
+  description: String!
```
Lezek123: Same as above

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+
+function bountyActorToMembership(actor: BountyActor): Membership | undefined {
+  if (actor.isMember) {
+    return new Membership({ id: String(actor.asMember) })
+  }
+}
+
+function fundingPeriodEnd(bounty: Bounty): number {
+  return (
+    bounty.maxFundingReachedEvent?.inBlock ??
```
Lezek123: In order for `maxFundingReachedEvent.inBlock` to be available here you need to query the bounty along with the event, ie.:
```
await store.get(Bounty, { where: { id: bountyId }, relations: ['maxFundingReachedEvent'] })
```

query-node/mappings/src/bounty.ts
Lezek123: It's a good practice to avoid assertions like `(bounty.fundingType as BountyFundingLimited)`.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  })
+}
+
+function endFundingPeriod(store: DatabaseManager, bounty: Bounty, isFunded = true): Promise<void> {
+  bounty.updatedAt = new Date()
+  if (isFunded) {
+    bounty.stage = BountyStage.WorkSubmission
+    bountyScheduleWorkSubmissionEnd(store, bounty)
+  } else {
+    bounty.stage = BountyStage[bounty.contributions?.length ? 'Failed' : 'Expired']
```
Lezek123: In order for `bounty.contributions` to be available here you need to query the bounty along with the `contributions`, ie.:
```
await store.get(Bounty, { where: { id: bountyId }, relations: ['contributions'] })
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  }
+  return store.save<Bounty>(bounty)
+}
+
+function endWorkingPeriod(store: DatabaseManager, bounty: Bounty): Promise<void> {
+  bounty.updatedAt = new Date()
+  if (bounty.entries?.length) {
```
Lezek123: In order for `bounty.entries` to be available here you need to query the bounty along with the `entries`, ie.:
```
await store.get(Bounty, { where: { id: bountyId }, relations: ['entries'] })
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+    entrantStake: bountyParams.entrant_stake,
+    creator: bountyActorToMembership(bountyParams.creator),
+    oracle: bountyActorToMembership(bountyParams.oracle),
+    fundingType: asFundingType(bountyParams.funding_type),
+    contractType: asContractType(bountyParams.contract_type),
+    workPeriod: bountyParams.work_period.toNumber(),
+    judgingPeriod: bountyParams.judging_period.toNumber(),
+
+    stage: BountyStage.Funding,
+    totalFunding: bountyParams.cherry,
```
Lezek123: It seems like in the runtime `totalFunding` and `cherry` are two independent values, but here they were "merged", which makes some of the logic invalid, ie.:
In the runtime `minimum_funding_reached` depends on `bounty.total_funding >= min_funding_amount`, in the mappings inside `bountyScheduleFundingEnd` we have:
```
const isFunded = bounty.totalFunding >= fundingType.minFundingAmount
```
But in the mappings `totalFunding` includes `cherry`, while in the runtime it doesn't, so the logic becomes inconsistent

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  // Update winner entries status
+  await Promise.all(
+    bounty.entries?.map((entry) => {
+      const judgement = entryJudgments.find(([entryId]) => String(entryId) === entry.id)?.[1]
+
+      if (!judgement) {
+        entry.status = new BountyEntryStatusPassed()
+      } else if (judgement?.isWinner) {
+        const status = new BountyEntryStatusWinner()
+        status.reward = judgement.asWinner.reward
```
Lezek123: Due to Hydra bug `u128` cannot be directly assigned to variant field (as mentioned before)
```suggestion
        status.reward = new BN(judgement.asWinner.reward.toString())
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// Terminate removed bounties
+export async function bounty_BountyRemoved({ event, store }: EventContext & StoreContext): Promise<void> {
+  const bountyRemovedEvent = new BountyEvents.BountyRemovedEvent(event)
+
+  // Terminate the bounty
+  const bounty = await updateBounty(store, event, bountyRemovedEvent.params[0], (bounty) => ({
+    deletedAt: bounty.updatedAt,
+    stage: BountyStage.Terminated,
```
Lezek123: Note that `BountyRemoved` is also emitted, for example, after all workers cash out, in which case the bounty status will change from `Successful` to `Terminated`. Is this what Pioneer expects?

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+      const isFunded = bounty.totalFunding >= fundingType.minFundingAmount
+      endFundingPeriod(store, bounty, isFunded)
+    }
+  })
+}
+
+export function bountyScheduleWorkSubmissionEnd(store: DatabaseManager, bounty: Bounty): void {
+  if (bounty.stage !== BountyStage.WorkSubmission) return
+
+  const workingPeriodEnd = fundingPeriodEnd(bounty) + bounty.workPeriod
```
Lezek123: This is incorrectly calculated, because `bounty.maxFundingReachedEvent.inBlock` is not yet available here when the function is called from `bounty_BountyMaxFundingReached`.

I added some console logs to ilustrate this (the bounty funding type was `Perpetual` and the funding target was reached):
```
[36mprocessor                |[0m Calculating funding period end {
[36mprocessor                |[0m   maxFundingReachedEventBlock: undefined,
[36mprocessor                |[0m   bountyCreatedInEventBlock: 123,
[36mprocessor                |[0m   bountyFundingPeriod: undefined
[36mprocessor                |[0m }
[36mprocessor                |[0m Scheduling work submission end at NaN
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,485 @@
...
+  scheduleAtBlock(workingPeriodEnd, () => {
+    if (bounty.stage === BountyStage.WorkSubmission) {
+      endWorkingPeriod(store, bounty)
+    }
+  })
+}
+
+export function bountyScheduleJudgementEnd(store: DatabaseManager, bounty: Bounty): void {
```
Lezek123: There is a recurring _Judg**e**ment_ instead of _Judgment_ typo (this includes schema, mappings etc.)

Lezek123 (changes requested):
I'm not sure if I spotted all the individual issues, but I think I covered most of them and many are related to similar wrong assumptions etc., so to summarize:
- Change the scheduling logic to use `preBlockHooks` / `postBlockHooks` instead of listening to `INDEXER_STATUS_CHANGE` event and execute all scheduled jobs separately, in a deterministic order
- Make sure the `bounty` fields you're trying to access in functions like `bountyScheduleFundingEnd`, `bountyScheduleWorkSubmissionEnd` etc. are actually populated with the up-to-date data (either by querying `bounty` along with the relations you're relying on or explicitly setting those values before you pass `bounty` argument to the function)
- Make sure you always reload the bounty data within the scheduled jobs
- Make sure all entity fields that may be `undefined` are optional in the schema!
- Make sure there are no `one-to-many` relations directly inside `variant`s in the schema (ie. `BountyContractClosed.whitelist`)
- Make sure you're not assigning `Codec` values (like `u128`, `u64`, `u32` etc.) directly to variant fields (ie. instead of `status.reward = judgement.asWinner.reward`, use `status.reward = new BN(judgement.asWinner.reward.toString()`), since the former causes issues due to `Codec` type having a custom `toJson()` implementation
- Make sure the mappings are inline with the runtime implementation, ie. the statuses / field values change under the same conditions etc. I spotted a few cases where your assumptions about the runtime implementation were wrong (ie. `bounty.totalFunding` mismatch, `bounty_BountyFundingWithdrawal` logic)
- Make sure `toNumber()` conversions are safe and values larger than `2147483647` are not stored in the `Int` database fields (see: https://github.com/Joystream/joystream/pull/3299)
- Make sure that ids provided as part of the metadata (ie. bounty thread id) cannot break query node due to missing entity - those cases should be handled without producing excpetions.

There were a few typo-like issues also, but there's not much to say about that.

Hope this review was helpful, I'm going to do a second round after the fixes are implemented, because the testing is very limited currently and not everything that I spotted can be addressed with a "quickfix" in order to allow for the testing to continue.

query-node/mappings/src/bounty.ts
thesan: Yes this is my understanding of the bounty cycle, but I will make sure.

query-node/mappings/src/bounty.ts
thesan: Right the cherry shouldn't be included in `totalFunding` since it's not included in the funding targets. Also I realized that it makes more sense for Pioneer to not have the funds subtracted on funding withdrawn or cash out.

query-node/mappings/src/bounty.ts
thesan: I see your point though! I've removed the `Terminated` stage and replaced it by an `isTerminated` flag, so the `Successful` / `Failed` information doesn't get overwritten.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,490 @@
...
+  const metadata = deserializeMetadata(BountyMetadata, metadataBytes)
+  const discussionThread = await whenDef(metadata?.discussionThread, (id) => store.get(ForumThread, { where: { id } }))
+
+  // Create the bounty
+  const bounty = new Bounty({
+    id: String(bountyId),
+    createdAt: eventTime,
+    updatedAt: eventTime,
+    title: metadata?.title ?? undefined,
+    description: metadata?.description ?? undefined,
```
Lezek123: During my tests I noticed that one can still break the query node by using a null character (`/u0000`) in a string that gets inserted to database.

This was supposed to be fixed in Hydra, but looks like the solution that was implemented is not sufficient.
As a temporary workaround you could use `perpareString` function from `common.ts`, but I'll try to make a patch for Hydra to cover this.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,490 @@
...
+    bannerImageUri: metadata?.bannerImageUri ?? undefined,
+    cherry: bountyParams.cherry,
+    entrantStake: bountyParams.entrant_stake,
+    creator: bountyActorToMembership(bountyParams.creator),
+    oracle: bountyActorToMembership(bountyParams.oracle),
+    fundingType: asFundingType(bountyParams.funding_type),
+    entrantWhitelist: asEntrantWhitelist(bountyParams.contract_type),
```
Lezek123: Child entities will not be persisted along with the parent, you need to save `BountyEntrantWhitelist` before you can use it as reference (otherwise `entrantWhitelist` will always be empty)

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,490 @@
...
+    return creationBlock + bounty.fundingType.fundingPeriod
+  }
+}
+
+function scheduleBountyStageEnd(
+  stage: BountyStage,
+  endStage: (store: DatabaseManager, bounty: Bounty, blockNumber: number) => Promise<void>,
+  relations: string[] = []
+): (store: DatabaseManager, bounty: Bounty, blockNumber: number | undefined) => void {
+  return (store, bounty, blockNumber) => {
```
Lezek123: Unfortunately this is not a valid approach. You can only use `store` within the scope of the event handler to which it was provided, otherwise you'll run into:
```
[36mprocessor                |[0m   "name": "QueryRunnerAlreadyReleasedError",
[36mprocessor                |[0m   "message": "Query runner already released. Cannot run queries anymore."
```

So, in other words: If you're processing `BountyCreated` event in `bounty_BountyCreated` handler, you should only use the instance of `store` within `bounty_BountyCreated` scope. If you use it in a function that's scheduled to be executed later, the `store` instance will have a reference to an outdated query runner.

You'd need to pass `store` directly from `runScheduler` into the actual job.

Lezek123 (changes requested):

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+
+/**
+ * Schedule Periods changes
+ */
+
+export const bountyScheduleFundingEnd = scheduleBountyStageEnd(
+  BountyStage.Funding,
+  async (store, bounty, fundingPeriodEnd) => {
+    if ('minFundingAmount' in bounty.fundingType) {
+      const isFunded = bounty.totalFunding.gte(bounty.fundingType.minFundingAmount)
```
Lezek123: Unfortunately looks like there is some typings bug in hydra and `bounty.fundingType.minFundingAmount` is not actually a `BN` (once fetched from the database), but a `string` instead. This is another issue on the long list of issues to be fixed in Hydra (hopefully most of them will be gone once we migrate to Subsquid)

When comparing a BN to a string you will get unexpected results.

A temporary workaround:
```suggestion
      const isFunded = bounty.totalFunding.gte(new BN(bounty.fundingType.minFundingAmount))
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+  }
+  return bounty
+}
+
+async function getContribution(
+  store: DatabaseManager,
+  bountyId: BountyId,
+  contributor: string | undefined
+): Promise<BountyContribution> {
+  const contribution = await store.get(BountyContribution, { where: { bountyId, contributor } })
```
Lezek123: The query always returns empty result, it looks like you need to modify the query slightly in order for it to work:
```suggestion
  const contribution = await store.get(BountyContribution, {
    where: { bounty: { id: bountyId.toString() }, contributor: { id: contributor } },
  })
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+
+  // Update the bounty totalFunding
+  const bounty = await updateBounty(store, event, bountyId, (bounty) => ({
+    totalFunding: bounty.totalFunding.add(amount),
+  }))
+
+  // Create or update the contribution
+  let contribution: BountyContribution
+  const contributor = bountyActorToMembership(contributorActor)
+  const existing = await store.get(BountyContribution, { where: { bountyId, contributor } })
```
Lezek123:
```suggestion
  const existing = await store.get(BountyContribution, {
    where: { bounty: { id: bountyId.toString() }, contributor: { id: contributor?.id } },
  })
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+    contribution = existing
+    contribution.amount = existing.amount.add(amount)
+  } else {
+    contribution = new BountyContribution({ createdAt: eventTime, bounty, contributor, amount })
+  }
+  contribution.updatedAt = eventTime
+  await store.save<BountyContribution>(contribution)
+
+  // Record the event
+  const fundedEvent = new BountyFundedEvent({ ...genericEventFields(event), contribution })
```
Lezek123, line 309: You have one-to-one relationship between `BountyFundedEvent` (`contribution`) and `BountyContribution` (`createdInEvent`), but now you may have multiple `BountyFundedEvent`s pointing to the same `BountyContribution`, which causes error on the database level:
```
processor                | ERROR: Stopping the proccessor due to errors: {
processor                |   "message": "duplicate key value violates unique constraint \"REL_dabfa47c320c4ea07c57235bf1\"",
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+      await endFundingPeriod(store, bounty, fundingPeriodEnd, isFunded)
+    }
+  }
+)
+
+export const bountyScheduleWorkSubmissionEnd = scheduleBountyStageEnd(BountyStage.WorkSubmission, endWorkingPeriod, [
+  'entries',
+])
+
+export const bountyScheduleJudgmentEnd = scheduleBountyStageEnd(BountyStage.WorkSubmission, async (store, bounty) => {
```
Lezek123:
```suggestion
export const bountyScheduleJudgmentEnd = scheduleBountyStageEnd(BountyStage.Judgment, async (store, bounty) => {
```

Lezek123:
```suggestion
export const bountyScheduleJudgmentEnd = scheduleBountyStageEnd(BountyStage.Judgment, async (store, bounty) => {
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+export async function bounty_WorkSubmitted({ event, store }: EventContext & StoreContext): Promise<void> {
+  const entrySlashedEvent = new BountyEvents.WorkSubmittedEvent(event)
+
+  // Update the entry status
+  const entry = await updateEntry(store, event, entrySlashedEvent.params[1], () => ({
+    workSubmitted: true,
+  }))
+
+  // Record the event
+  const submittedInEvent = new WorkSubmittedEvent({ ...genericEventFields(event), entry })
```
Lezek123:
```suggestion
  const workSubmittedEvent = new BountyEvents.WorkSubmittedEvent(event)

  // Update the entry status
  const entry = await updateEntry(store, event, workSubmittedEvent.params[1], () => ({
    workSubmitted: true,
  }))

  // Record the event
  const meta = deserializeMetadata(BountyWorkData, workSubmittedEvent.params[3])
  const submittedInEvent = new WorkSubmittedEvent({
    ...genericEventFields(event),
    entry,
    title: whenDef(meta?.title, perpareString),
    description: whenDef(meta?.description, perpareString),
  })
```

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+  await store.save<WorkEntrySlashedEvent>(submittedInEvent)
+}
+
+// Start bounties withdrawal period and set entries status to either passed, winner, or rejected
+export async function bounty_OracleJudgmentSubmitted({ event, store }: EventContext & StoreContext): Promise<void> {
+  const judgmentSubmittedEvent = new BountyEvents.OracleJudgmentSubmittedEvent(event)
+  const [bountyId, , bountyJudgment] = judgmentSubmittedEvent.params
```
Lezek123: Is it intentional to omit and not map `rationale` parameter here?

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+    bounty.stage = BountyStage[bounty.totalFunding.gtn(0) ? 'Failed' : 'Expired']
+  }
+  return store.save<Bounty>(bounty)
+}
+
+function endWorkingPeriod(store: DatabaseManager, bounty: Bounty, blockNumber: number): Promise<void> {
+  bounty.updatedAt = new Date()
+  if (bounty.entries?.length) {
+    bounty.stage = BountyStage.Judgment
+    bountyScheduleJudgmentEnd(bounty, blockNumber + bounty.judgingPeriod)
```
Lezek123: Invalid assumption: The runtime will only progress a bounty to `Judgement` stage if there was at least one work submitted.

Here you're only checking if there are any entries, but not if they contain any submitted works.

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,496 @@
...
+
+  // Update the bounty status
+  const hasWinners = entryJudgments.some(([, judgment]) => judgment.isWinner)
+  const bounty = await updateBounty(store, event, bountyId, () => ({
+    stage: BountyStage[hasWinners ? 'Successful' : 'Failed'],
+  }))
+
+  // Update winner entries status
+  await Promise.all(
+    bounty.entries?.map((entry) => {
```
Lezek123, line 465: `bounty.entries` is always empty here (bounty is not fetched with `['entries']` relationship.

Lezek123 (changes requested):

query-node/mappings/src/bounty.ts
thesan: No the rationale wasn't implemented by the runtime when I wrote this one :grin:

query-node/mappings/src/bounty.ts
thesan: Sorry about that, I had not realized that I barely implemented this and just forgot about it :facepalm:

query-node/mappings/src/bounty.ts
```diff
@@ -0,0 +1,509 @@
...
+  return bounty
+}
+
+async function getContribution(
+  store: DatabaseManager,
+  bountyId: BountyId,
+  contributorId: string | undefined
+): Promise<BountyContribution | undefined> {
+  const contributor = whenDef(contributorId, (id) => ({ contributor: id })) ?? null
+  return await store.get(BountyContribution, { where: { bountyId, contributor } })
```
Lezek123:
```suggestion
  const contributor = whenDef(contributorId, (id) => ({ id })) ?? ( { id: null })
  return await store.get(BountyContribution, { where: { bounty: { id: bountyId }, contributor } })
```

query-node/schemas/bounty.graphql
```diff
@@ -0,0 +1,219 @@
...
+
+  "Member making the contribution (if a member and not the council)"
+  contributor: Membership
+
+  "Amount of the contribution"
+  amount: BigInt!
+
+  # Events
+  "The event the contribution was created in"
+  createdInEvent: [BountyFundedEvent] @derivedFrom(field: "contribution")
```
Lezek123: The field name and description seems incorrect now

Lezek123 (changes requested):
Almost everything works as expected now, just 2 remaining issues to fix

query-node/mappings/src/bounty.ts
```diff
@@ -50,13 +50,12 @@ async function getBounty(store: DatabaseManager, bountyId: BountyId | string, re
...
 
-async function getContribution(
+function getContribution(
   store: DatabaseManager,
   bountyId: BountyId,
   contributorId: string | undefined
 ): Promise<BountyContribution | undefined> {
-  const contributor = whenDef(contributorId, (id) => ({ contributor: id })) ?? null
-  return await store.get(BountyContribution, { where: { bountyId, contributor } })
+  return store.get(BountyContribution, { where: { bountyId, contributor: { id: contributorId ?? null } } })
```
Lezek123:
```suggestion
  return store.get(BountyContribution, { where: { bounty: { id: bountyId }, contributor: { id: contributorId ?? null } } })
```

Lezek123 (changes requested):

query-node/mappings/src/bounty.ts
thesan: I followed your suggestion but just to make sure, when the council is the contributor, will e.g:
```ts
store.get(BountyContribution, { where: { bounty: { id: 0 }, contributor: { id: null } } })
```
returns the correct contribution ?

query-node/mappings/src/bounty.ts
thesan: Sorry I just realized I forgot about the bounty. But please check my reply above :point_up:

query-node/mappings/src/bounty.ts
Lezek123: Yes, just did a custom test locally and it works as expected (even though Council contributions are not possible currently)

Lezek123 (approved):
LGTM!

PR #2952, opened 2021-12-20
"Abstain" votes shouldn't count as "No"

singulart:
In the current implementation, "Abstain" votes are treated as votes against the proposal. This simple improvement is about not taking such votes into account when calculating thresholds. I've also added some more unit tests. 

See https://mrsc.org/Home/Stay-Informed/MRSC-Insight/April-2013/How-Are-Abstentions-Handled-When-Counting-Votes.aspx

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703462643764) by [Unito](https://www.unito.io)

### DIFF
--- a/runtime-modules/proposals/engine/src/types/mod.rs
+++ b/runtime-modules/proposals/engine/src/types/mod.rs
@@ -112,7 +112,7 @@ impl VotingResults {
 
     /// Calculates number of votes so far
     pub fn votes_number(&self) -> u32 {
-        self.abstentions + self.approvals + self.rejections + self.slashes
+        self.approvals + self.rejections + self.slashes
     }
 }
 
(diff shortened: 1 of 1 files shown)


bwhm:
I think this is the right idea, but I'm not sure if this fix would be obsoleted by the many changes in `olympia`. @bedeho?

bedeho:
This is being addressed here https://github.com/Joystream/joystream/issues/2953

shamil-gadelshin (commented):
We opted for another algorithm: https://github.com/Joystream/joystream/issues/2953

This PR helped to advance the discussion. Thank you.

mnaamani:
Looks like this PR can be closed?

PR #2954, opened 2021-12-20
query node - active video counters

ondratra:
This PR adds counters to query node's channels, channel categories, and video categories.

Use it in a query like this:

```
channels {id, activeVideosCounter}
channelCategories {id, activeVideosCounter}
videoCategories {id, activeVideosCounter}
```

Videos are considered active if all the following conditions are true: `isPublic == true`, `isCensored == false`, `thumbnailPhotoAvailability == AssetAvailability.Accepted`, and `mediaAvailability == AssetAvailability.Accepted`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703601809148) by [Unito](https://www.unito.io)

### DIFF
--- a/query-node/mappings/src/common.ts
+++ b/query-node/mappings/src/common.ts
@@ -96,42 +96,105 @@ export async function prepareDataObject(
   return dataObject
 }
 
-export function isVideoFullyActive(video: Video): boolean {
-  return (
+export interface IVideoActiveStatus {
+  isFullyActive: boolean
+  videoCategory: VideoCategory | undefined
+  channel: Channel
+  channelCategory: ChannelCategory | undefined
+}
+
+export function getVideoActiveStatus(video: Video): IVideoActiveStatus {
+  const isFullyActive =
     !!video.isPublic &&
     !video.isCensored &&
     video.thumbnailPhotoAvailability === AssetAvailability.ACCEPTED &&
     video.mediaAvailability === AssetAvailability.ACCEPTED
-  )
+  const videoCategory = video.category
+  const channel = video.channel
+  const channelCategory = channel.category
+
+  return {
+    isFullyActive,
+    videoCategory,
+    channel,
+    channelCategory,
+  }
 }
 
 export async function updateVideoActiveCounters(
   store: DatabaseManager,
-  video: Video,
-  isFullyActive: boolean
+  initialActiveStatus: IVideoActiveStatus | undefined,
+  activeStatus: IVideoActiveStatus | undefined
 ): Promise<void> {
-  const counterChange = isFullyActive ? 1 : -1
-
-  if (video.channel) {
-    const channel = video.channel
-    channel.activeVideosCounter += counterChange
-
-    await store.save<Channel>(channel)
+  // definition of generic type for Hydra DatabaseManager's methods
+  type EntityType<T> = {
+    new (...args: any[]): T
   }
 
-  if (video.channel && video.channel.category) {
-    const channelCategory = video.channel.category
-    channelCategory.activeVideosCounter += counterChange
+  async function updateSingleEntity<Entity extends VideoCategory | Channel>(
+    entity: Entity,
+    counterChange: number
+  ): Promise<void> {
+    entity.activeVideosCounter += counterChange
 
-    await store.save<ChannelCategory>(channelCategory)
+    await store.save<EntityType<Entity>>(entity)
   }
 
-  if (video.category) {
-    const category = video.category
-    category.activeVideosCounter += counterChange
+  async function reflectUpdate<Entity extends VideoCategory | Channel>(
+    oldEntity: Entity | undefined,
+    newEntity: Entity | undefined,
+    initFullyActive: boolean,
+    nowFullyActive: boolean
+  ): Promise<void> {
+    if (!oldEntity && !newEntity) {
+      return
+    }
 
-    await store.save<VideoCategory>(category)
+    const didEntityChange = oldEntity?.id.toString() !== newEntity?.id.toString()
+    const didFullyActiveChange = initFullyActive !== nowFullyActive
+
+    // escape if nothing changed
+    if (!didEntityChange && !didFullyActiveChange) {
+      return
+    }
+
+    if (!didEntityChange) {
+      // && didFullyActiveChange
+      const counterChange = nowFullyActive ? 1 : -1
+
+      await updateSingleEntity(newEntity as Entity, counterChange)
+
+      return
+    }
+
+    // didEntityChange === true
+
+    if (oldEntity) {
+      // if video was fully active before, prepare to decrease counter; increase counter otherwise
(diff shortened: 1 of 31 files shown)


kdembler:
I'm not aware of query node's internal workings enough to properly review. I'll happily test it out once we have some public deployment though

query-node/mappings/src/content/video.ts
```diff
@@ -179,6 +181,11 @@ export async function content_VideoCreated(db: DatabaseManager, event: Substrate
   // save video
   await db.save<Video>(video)
 
+  // increase active video counters for channel, channel category, and video category
+  if (isVideoFullyActive(video)) {
+    await updateVideoActiveCounters(db, video, true)
```
Lezek123: `video.channel.category` is not fetched along with the `channel`, so it's always empty and the channel category activeVideosCounter is not updated
**_Edit: Actually I noticed there is a bug and channel category is never really assigned, so it doesn't matter unless said bug is also fixed._**

Lezek123 (changes requested):
You need to add proper handling for the case when video / channel is moved to a different category or the video's/channel's category is unset.

query-node/mappings/src/content/video.ts
ondratra: ~I've inspected Sumer's runtime, and I think categories and channels can't be changed in this release. I think mappings are aligned with that and correct.~ A ~formal~ load of `video.channel.category` was added,~but it will actually have no effect, as you stated.~

~In light of that, this PR is actually complete, right?~

query-node/mappings/src/content/video.ts
```diff
@@ -139,7 +139,9 @@ export async function content_VideoCreated(db: DatabaseManager, event: Substrate
   })
 
   // load channel
-  const channel = await db.get(Channel, { where: { id: channelId.toString() } as FindConditions<Channel> })
+  const channel = await db.get(Channel, {
+    where: { id: channelId.toString(), relations: ['category'] } as FindConditions<Channel>,
```
Lezek123:
```
ERROR: Stopping the proccessor due to errors: {
  "name": "EntityColumnNotFound",
  "message": "No entity column \"relations\" was found."
```

```suggestion
    where: { id: channelId.toString() } as FindConditions<Channel>,
    relations: ['category']
```

query-node/mappings/src/content/video.ts
Lezek123: 1. You can't set/update channel's category, but this is only because those actions are not be processed by the mappings. Channel categories do exist and it's possible to include them in the metadata (`ChannelMetadata.category`), just like video categories. There's nothing in the runtime that prevents this.
2. You can still update video's category (`VideoMetadata.category`) in which case you want to reduce the `activeVideosCounter` in the old category (if it exists and the video was active), and possibly increase the `activeVideosCounter` in the new category (if it exists, !== oldCategory and the video is active). This is not correctly handled currently, for example: You have an active video in category `1`, you update its category to `2` and you get invalid state in QN:
  image

query-node/mappings/src/content/video.ts
ondratra: Sorry... I edited that recklessly. Fixed in another commit

query-node/mappings/src/content/video.ts
ondratra: Ok. I reworked the feature, so it handles these kinds of updates. There still needs to be some work and especially testing done.

query-node/mappings/src/common.ts
```diff
@@ -96,42 +96,105 @@ export async function prepareDataObject(
...
-    const channel = video.channel
-    channel.activeVideosCounter += counterChange
+  async function updateSingleEntity<Entity extends VideoCategory | Channel>(
+    entity: Entity,
+    counterChange: number
+  ): Promise<void> {
+    entity.activeVideosCounter += counterChange
 
-    await store.save<Channel>(channel)
+    await store.save<EntityType<Entity>>(entity)
```
Lezek123, line 140: This doesn't seem to add any value in terms of typesafety (I can still pass invalid object or even a number as argument and the compiler won't complain). I think you can just use: `store.save(entity)` instead.

query-node/mappings/src/common.ts
```diff
@@ -96,42 +96,105 @@ export async function prepareDataObject(
...
-  const counterChange = isFullyActive ? 1 : -1
+  // definition of generic type for Hydra DatabaseManager's methods
+  type EntityType<T> = {
+    new (...args: any[]): T
+  }
 
-  if (video.channel) {
-    const channel = video.channel
-    channel.activeVideosCounter += counterChange
+  async function updateSingleEntity<Entity extends VideoCategory | Channel>(
```
Lezek123, line 134:
```suggestion
  async function updateSingleEntity<Entity extends VideoCategory | ChannelCategory | Channel>(
```

query-node/mappings/src/common.ts
```diff
@@ -96,42 +96,105 @@ export async function prepareDataObject(
...
+    entity.activeVideosCounter += counterChange
 
-    await store.save<Channel>(channel)
+    await store.save<EntityType<Entity>>(entity)
   }
 
-  if (video.channel && video.channel.category) {
-    const channelCategory = video.channel.category
-    channelCategory.activeVideosCounter += counterChange
+  async function reflectUpdate<Entity extends VideoCategory | Channel>(
```
Lezek123, line 143:
```suggestion
  async function reflectUpdate<Entity extends VideoCategory | ChannelCategory | Channel>(
```

query-node/mappings/src/common.ts
```diff
@@ -96,42 +96,105 @@ export async function prepareDataObject(
...
-    await store.save<VideoCategory>(category)
+      await updateSingleEntity(newEntity as Entity, counterChange)
+
+      return
+    }
+
+    // didEntityChange === true
+
+    if (oldEntity) {
+      // if video was fully active before, prepare to decrease counter; increase counter otherwise
```
Lezek123, line 173: This behavior is incorrect.
For example, if we change the category of an inactive video from 1 to 2, we don't want to increment the activeVideoCounter in category 1, that doesn't make sense.

query-node/mappings/src/common.ts
```diff
@@ -96,42 +96,105 @@ export async function prepareDataObject(
...
+      const counterChange = initFullyActive ? -1 : 1
+
+      await updateSingleEntity(oldEntity, counterChange)
+    }
+
+    if (newEntity) {
+      // if video is fully active now, prepare to increase counter; decrease counter otherwise
```
Lezek123, line 180: This is also not correct. If we change category of an inactive video from 1 to 2 we don't want to decrement the activeVideoCounter in category 2.

Lezek123 (changes requested):

ondratra:
Superseded by https://github.com/Joystream/joystream/pull/2964

PR #2955, opened 2021-12-21
Lower validator reward, bump runtime and node versions

mnaamani:


shamil-gadelshin (approved):
Looks good. 

However, `chain-spec-builder` version update is missing. I'll approve and merge it to speed up the parameters change. However, we need to update the `chain-spec-builder` as well.

PR #2958, opened 2021-12-21
DevOps - AWS CloudFormation launch fixes

ahhda:
This PR fixes the below-mentioned issues:

* Enable services to start on boot
* Use `data_path` in bash script

Screenshot 2021-12-21 at 6 55 47 PM

Screenshot 2021-12-21 at 6 57 14 PM

### DIFF
--- a/devops/aws/deploy-infra.sh
+++ b/devops/aws/deploy-infra.sh
@@ -75,7 +75,7 @@ if [ $? -eq 0 ]; then
   then
     echo -e "\n\n=========== Compile joystream-node on build server ==========="
     ansible-playbook -i $INVENTORY_PATH --private-key $KEY_PATH build-code.yml \
-      --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO build_local_code=$BUILD_LOCAL_CODE data_path=data-$NEW_STACK_NAME"
+      --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO build_local_code=$BUILD_LOCAL_CODE data_path=$DATA_PATH"
 
     echo -e "\n\n=========== Install additional utils on build server ==========="
     ansible-playbook -i $INVENTORY_PATH --private-key $KEY_PATH setup-admin.yml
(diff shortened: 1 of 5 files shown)


devops/aws/deploy-infra.sh
```diff
@@ -75,7 +75,7 @@ if [ $? -eq 0 ]; then
   then
     echo -e "\n\n=========== Compile joystream-node on build server ==========="
     ansible-playbook -i $INVENTORY_PATH --private-key $KEY_PATH build-code.yml \
-      --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO build_local_code=$BUILD_LOCAL_CODE data_path=data-$NEW_STACK_NAME"
+      --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO build_local_code=$BUILD_LOCAL_CODE data_path=$DATA_PATH"
```
mnaamani, line 78: I think the playbook also needs to be adjusted so that you can have a different local path.
Say I wanted an absolute path DATA_PATH=/home/mokhtar/artifacts
That doesn't exist on the remote machine..

ahhda: The `data_path` variable is to set the directory where the binaries, chainspec files and auth information would reside on the local machine (the machine which was used to run the scripts). It works with both relative and absolute paths.

This variable doesn't interfere with how data is stored on the remote machines i.e. the inventory (validators and rpc nodes). Do we want to change the location of the data stored on the remote machine?

mnaamani: Oh thanks for correcting me here. In that case it should be fine. I'll test to confirm. Otherwise looks good

mnaamani: Perhaps I was correct?

I set DATA_PATH=~/tmp/deployment/giza-staging-dec-22

The inventory file was correctly placed in my `/Users/mokhtar/tmp/deployment/giza-staging-dec-22/`

But it tried to use that same full path on the remote machine:

```
TASK [common : Run chain-spec-builder to generate chainspec.json file (with initial data)] ******************************************************************
fatal: [54.174.240.128]: FAILED! => changed=true 
  cmd: |-
    /home/ubuntu/joystream/target/release/chain-spec-builder generate -a 1 --chain-spec-path /Users/mokhtar/tmp/deployment/giza-staging-dec-22/chainspec.json --endowed 1 --keystore-path /Users/mokhtar/tmp/deployment/giza-staging-dec-22 --deployment staging
  delta: '0:00:01.266905'
  end: '2021-12-22 17:08:15.925265'
  msg: non-zero return code
  rc: 1
  start: '2021-12-22 17:08:14.658360'
  stderr: 'Error: "Permission denied (os error 13)"'
  stderr_lines: <omitted>
  stdout: |-

```

ahhda: You are right, I missed testing a different relative path. I was reusing the `data_path` variable for both localhost and build server causing confusion. Will correct this.

mnaamani (approved):
Works as expected now :+1:

PR #2959, opened 2021-12-21
Fork off script: Fix module.storage.prefix, fix linter errors

Lezek123:
**`utils/api-scripts/src/fork-off.ts `:**
- Replace `xxhashAsHex(module.name.toString(), 128)` with `xxhashAsHex(module.storage.unwrap().prefix.toString(), 128)` to fix missing Membership module state.
- Fix linter errors


mnaamani:
Nice catch. I'm curios. Is there also a difference between module.name and module.storage.prefix for the working group module instances?

mnaamani (approved):

PR #2964, opened 2021-12-23
query node - Giza - active video counters

ondratra:
This PR adds the save active-video-counters feature as https://github.com/Joystream/joystream/pull/2954 to Giza release.

Counters are added to query node's channels, channel categories, and video categories.

Use it in a query like this:

```
channels {id, activeVideosCounter}
channelCategories {id, activeVideosCounter}
videoCategories {id, activeVideosCounter}
```

Videos are considered active if all the following conditions are true: `isPublic == true`, `isCensored == false`, `thumbnailPhoto is set`, and `media is set`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703462334862) by [Unito](https://www.unito.io)


query-node/mappings/content/utils.ts
```diff
@@ -518,10 +521,54 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
+
     // emit log event
     logger.info('Content has been disconnected from Video', {
       videoId: video.id.toString(),
       dataObjectId: dataObject.id,
     })
   }
+
+  // remove data object
+  await store.remove<StorageDataObject>(dataObject)
```
Lezek123, line 535: I think this should either remain a separate action or this function should replace `removeDataObject` completely

query-node/mappings/content/utils.ts
```diff
@@ -518,10 +521,54 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
+}
+
+export function isVideoFullyActive(video: Video): boolean {
+  return !!video.isPublic && !video.isCensored && !!video.thumbnailPhoto && !!video.media
```
Lezek123: To make it consistent with https://github.com/Joystream/joystream/pull/2954, this should be:
```suggestion
  return !!video.isPublic && !video.isCensored && !!(video.thumbnailPhoto?.isAccepted) && !!(video.media?.isAccepted)
```
And there should be additional logic added in `storage_PendingDataObjectsAccepted`

query-node/mappings/content/video.ts
```diff
@@ -154,6 +162,14 @@ export async function content_VideoUpdated(ctx: EventContext & StoreContext): Pr
   // save video
   await store.save<Video>(video)
 
+  // check if video is fully active
+  const isFullyActive = isVideoFullyActive(video)
+
+  // update video counters for channel, channel category, and video category if needed
+  if (wasFullyActive !== isFullyActive) {
+    await updateVideoActiveCounters(store, video, isFullyActive)
```
Lezek123: Same as in https://github.com/Joystream/joystream/pull/2954 - there needs to be special handling in case the video category has been changed / unset.
Similarly for channel category in `content_ChannelUpdated`

Lezek123 (changes requested):
On channel creation:
```
"message": "null value in column \"active_videos_counter\" violates not-null constraint"
```
I think the same will apply to: `content_ChannelCategoryCreated`, `content_VideoCategoryCreated`

query-node/mappings/content/video.ts
ondratra: The feature has been reworked. Channel and categories updates are handled now.

query-node/mappings/content/utils.ts
ondratra: Fixed.

query-node/mappings/content/utils.ts
ondratra: `removeDataObject` has been removed - fully replaced by `unsetAssetRelations`.

cli/src/base/AccountsCommandBase.ts
```diff
@@ -376,5 +390,9 @@ export default abstract class AccountsCommandBase extends ApiCommandBase {
       throw this.createDataDirInitError()
     }
     await this.initKeyring()
+
+    const availableMemberships = await this.getKnownMembers()
```
Lezek123: This means all memberships will be fetched from the node each time a command that extends `AccountsCommandBase` is ran, which is very suboptimal. A better way would be to just fetch a member by `selectedMemberId` and just see if the controller key is available.

cli/src/base/ApiCommandBase.ts
```diff
@@ -89,9 +89,6 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
       if (this.requiresQueryNode && !queryNodeUri) {
         this.warn('Query node endpoint uri is required in order to run this command!')
         queryNodeUri = await this.promptForQueryNodeUri(true)
-      } else if (queryNodeUri === undefined) {
```
Lezek123: I'd prefer to leave it, query node will be useful for almost all of the commands soon, so prompting to provide it on the first run makes sense. A user can always choose `none` and set it later though. (`none` === `null`, which is different from `undefined`)

cli/src/commands/account/chooseMember.ts
```diff
@@ -0,0 +1,43 @@
+import AccountsCommandBase, { ISelectedMember } from '../../base/AccountsCommandBase'
+import chalk from 'chalk'
+import { flags } from '@oclif/command'
+import ExitCodes from '../../ExitCodes'
+
+export default class AccountChoose extends AccountsCommandBase {
+  static description = 'Choose default account to use in the CLI'
```
Lezek123:
```suggestion
export default class AccountChooseMember extends AccountsCommandBase {
  static description = 'Choose default member to use in the CLI'
```

cli/src/commands/account/chooseMember.ts
```diff
@@ -0,0 +1,43 @@
...
+  static flags = {
+    address: flags.string({
```
Lezek123: Why not use a `memberId` flag instead? It's much easier to retrieve a member by id than to do a lookup of member by their controller address (which requires fetching all existing memberships). It doesn't seem like there's a lot of utility in choosing a membership by controller address instead of `memberId` either.

query-node/mappings/content/utils.ts
```diff
@@ -492,11 +505,11 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
         id: dataObject.id,
       },
     })),
-    relations: [...videoAssets, 'category', 'channel', 'channel.category'],
+    relations: [...videoAssets, ...videoRelationsForCountersBare],
   })
 
   // remember if video is fully active before update
-  const wasFullyActive = video && isVideoFullyActive(video)
+  const wasFullyActive = video && getVideoActiveStatus(video)
```
Lezek123, line 499: Perhaps you could move it under `if (video) {` to prevent the need for assetion (`wasFullyActive as IVideoActiveStatus`) later.

query-node/mappings/content/utils.ts
```diff
@@ -540,35 +548,136 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
+}
+
+export function getVideoActiveStatus(video: Video): IVideoActiveStatus {
+  const productionEnv = () => {
+    const isFullyActive = !!video.isPublic && !video.isCensored && !!video.thumbnailPhoto?.isAccepted
+    !!video.media?.isAccepted
+
+    return isFullyActive
+  }
+  const testEnv = () => {
```
Lezek123: Shouldn't we try to simulate the production environment / behavior as closely as possible when running the tests? It should be quite easy now to setup a test with a running storage node which would accept the data obejcts (you could just add an additional flow / scenario as part of `run-full-tests.sh`)
With the separation of `testEnv` and `productionEnv` behavior the tests will be very limited and unreliable

query-node/mappings/content/utils.ts
```diff
@@ -540,35 +548,136 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
-  video: Video,
-  isFullyActive: boolean
+  initialActiveStatus: IVideoActiveStatus | null | undefined,
+  activeStatus: IVideoActiveStatus | null | undefined
 ): Promise<void> {
-  const counterChange = isFullyActive ? 1 : -1
+  // definition of generic type for Hydra DatabaseManager's methods
+  type EntityType<T> = {
+    new (...args: any[]): T
+  }
```
Lezek123: Same comment I added in the PR targetting Sumer: https://github.com/Joystream/joystream/pull/2954#discussion_r778298197

query-node/mappings/content/utils.ts
```diff
@@ -540,35 +548,136 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
+
+      await updateSingleEntity(newEntity as Entity, counterChange)
+
+      return
+    }
+
+    // didEntityChange === true
+
+    if (oldEntity) {
+      // if video was fully active before, prepare to decrease counter; increase counter otherwise
```
Lezek123: https://github.com/Joystream/joystream/pull/2954#discussion_r778300936

query-node/mappings/content/utils.ts
```diff
@@ -540,35 +548,136 @@ export async function unsetAssetRelations(store: DatabaseManager, dataObject: St
...
+      const counterChange = initFullyActive ? -1 : 1
+
+      await updateSingleEntity(oldEntity, counterChange)
+    }
+
+    if (newEntity) {
+      // if video is fully active now, prepare to increase counter; decrease counter otherwise
```
Lezek123: https://github.com/Joystream/joystream/pull/2954#discussion_r778302695

query-node/mappings/storage/index.ts
```diff
@@ -227,13 +230,63 @@ export async function storage_DataObjectsUploaded({ event, store }: EventContext
...
+    We can upgrade this algorithm in the future if this event mapping proves to have serious
+    performance issues. In that case, a unit test for this mapping will be required.
+  */
+  // load relevant videos one by one and update related active-video-counters
+  await initialActiveStates.reduce(async (accPromise, initialActiveState) => {
+    await accPromise
+
+    // load refreshed version of videos and related entities (channel, channel category, category)
+
+    const video = await store.get(Video, {
```
Lezek123: Why not pre-load the videos using `getMany` with `where: { id: In(initialActiveStates.map(s => s.video.id.toString())) }` instead?

Of course the best way would be to calculate the status when processing data object updates instead of re-fetching all the videos afterwards, but as you mentioned, it may be slightly more complex.

query-node/mappings/storage/index.ts
```diff
@@ -227,13 +230,63 @@ export async function storage_DataObjectsUploaded({ event, store }: EventContext
...
+    shared by mutliple entities (at least now). Because of that when updating for example
+    `dataObject.videoThumnail.channel.activeVideoCounter` on dataObject A, this change is not
+    reflected on `dataObject.videoMedia.channel.activeVideoCounter` on dataObject B.
+
+    We can upgrade this algorithm in the future if this event mapping proves to have serious
+    performance issues. In that case, a unit test for this mapping will be required.
+  */
+  // load relevant videos one by one and update related active-video-counters
+  await initialActiveStates.reduce(async (accPromise, initialActiveState) => {
+    await accPromise
```
Lezek123: In cases like that I think much cleaner approach is using `for item of array { ... }` instead, ie.:
```suggestion
  for initialActiveState of initialActiveStates {
```
You can then safely use `await` inside the loop and the processing will be pretty much the same.

query-node/mappings/storage/index.ts
```diff
@@ -250,8 +303,32 @@ export async function storage_DataObjectsMoved({ event, store }: EventContext &
...
+    ...videoRelationsForCounters.map(item => `videoMedia.${item}`),
+  ])
+
+  await Promise.all(
+    dataObjects.map(async (dataObject) => {
+      // remember if video is fully active before update
+      const initialVideoActiveStatusThumbnail = dataObject.videoThumbnail
+        ? getVideoActiveStatus(dataObject.videoThumbnail)
+        : null
+      const initialVideoActiveStatusMedia = dataObject.videoMedia ? getVideoActiveStatus(dataObject.videoMedia) : null
```
Lezek123: I'd reduce it to just 1 variable.
```suggestion
      const initialVideoActiveStatus = dataObject.videoThumbnail
        ? getVideoActiveStatus(dataObject.videoThumbnail)
        : dataObject.videoMedia
          ? getVideoActiveStatus(dataObject.videoMedia)
          : null
```

query-node/schemas/storage.graphql
```diff
@@ -213,6 +213,12 @@ type StorageDataObject @entity {
 
   "If the object is no longer used as an asset - the time at which it was unset (if known)"
   unsetAt: DateTime
+
+  "Video that has this data object associated as thumbnail photo."
+  videoThumbnail: Video @derivedFrom(field: "thumbnailPhoto")
+
+  "Video that has this data object associated as media."
+  videoMedia: Video @derivedFrom(field: "media")
```
Lezek123, line 221: The names are a little confusing, as both of those fields reffer to a `Video` entity.
Perhaps you could rename them to `thumbnailParentVideo` / `mediaParentVideo` (or something smiliar)

tests/network-tests/src/Api.ts
```diff
@@ -114,8 +118,7 @@ export class ApiFactory {
   }
 
   public createCustomKeyPair(customPath: string): KeyringPair {
-    const uri = `${this.miniSecret}//testing//${customPath}`
-    return this.keyring.addFromUri(uri)
+    return this.keyring.addFromUri(customPath + uuid().substring(0, 8))
```
Lezek123: A lot of the functionality on `giza_staging` still relies on the previous logic of this function, why was it changed?

tests/network-tests/src/Api.ts
```diff
@@ -907,6 +910,14 @@ export class Api {
     }
   }
 
+  public findStorageBucketCreated(events: EventRecord[]): DataObjectId | undefined {
+    const record = this.findEventRecord(events, 'storage', 'StorageBucketCreated')
+
+    if (record) {
+      return (record.event.data[0] as unknown) as DataObjectId
+    }
+  }
```
Lezek123: After merging the latest `giza_staging` you can just use `this.getEvent(events, 'storage', 'StorageBucketCreated').data[0]` (see: https://github.com/Joystream/joystream/blob/giza_staging/tests/network-tests/src/Api.ts#L910)

tests/network-tests/src/CliApi.ts
```diff
@@ -0,0 +1,256 @@
...
+
+export interface ICreatedVideoData {
+  videoId: number
+  assetContentIds: string[]
+}
+
+/**
+  Adapter for calling CLI commands from integration tests.
+*/
+export class CliApi {
```
Lezek123: Mostly a duplicate of https://github.com/Joystream/joystream/blob/giza_staging/tests/network-tests/src/cli/joystream.ts, you can move the logic there after merging `giza_staging`

tests/network-tests/src/Api.ts
```diff
@@ -1982,24 +1966,25 @@ export class Api {
     channelId: string,
     addStorageBuckets: StorageBucketId[]
   ) {
-    const bagId = {Dynamic: {Channel: channelId}}
-    const encodedStorageBucketIds = new BTreeSet<StorageBucketId>(this.api.registry, "StorageBucketId", addStorageBuckets.map(item => item.toString()))
-    const noBucketsToRemove = new BTreeSet<StorageBucketId>(this.api.registry, "StorageBucketId", [])
+    const bagId = { Dynamic: { Channel: channelId } }
+    const encodedStorageBucketIds = new BTreeSet<StorageBucketId>(
```
Lezek123: Please use `this.api.createType` or `createType` function from `@joystream/types` for consistency across the codebase.

tests/network-tests/src/consts.ts
```diff
@@ -0,0 +1,2 @@
+// Test chain blocktime
+export const BLOCKTIME = 1000
```
Lezek123: For Giza it's `6000`

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// QN connection paramaters
+const qnConnection = {
+  numberOfRepeats: 20, // QN can take some time to catch up with node - repeat until then
+  repeatDelay: 3000, // delay between failed QN requests
+}
+
+// settings
+const contentDirectoryWorkingGroupId = 1 // TODO: retrieve group id programmatically
```
Lezek123: What is this for?

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,383 @@
...
+const sufficientTopupAmount = new BN(1000000) // some very big number to cover fees of all transactions
+
+/**
+  Fixture that test Joystream content can be created, is reflected in query node,
+  and channel and categories counts their active videos properly.
+*/
+export class ActiveVideoCountersFixture extends BaseQueryNodeFixture {
```
Lezek123, line 33: As I already pointed once, a `Fixture` should be more atomic and reusable, this one is more like an entire flow / scenario

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,383 @@
...
+    // end
+    */
+
+    this.debug('Done')
+  }
+
+  /**
+    Prepares storage requisites.
+  */
+  private async prepareAssetStorage(storageLeader: Worker, storageGroupWorker: IMember) {
```
Lezek123: This is handled by https://github.com/Joystream/joystream/blob/giza_staging/tests/network-tests/src/flows/storagev2/initStorage.ts on `giza_staging`, which is part of the scenario executed during `run-all-tests`: https://github.com/Joystream/joystream/blob/giza_staging/tests/network-tests/src/scenarios/combined.ts

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,383 @@
...
+  private async assertCounterMatch(
+    entityName: 'channels' | 'channelCategories' | 'videoCategories',
+    entityId: number,
+    expectedCount: number
+  ) {
+    const qnConnectionNumberOfRepeats = 10
+
+    const getterName = `get${entityName[0].toUpperCase()}${entityName.slice(1)}`
+    await this.query.tryQueryWithTimeout(
+      () => (this.query as any)[getterName](),
```
Lezek123: Merging `giza_staging` will allow you to use typesafe queries (https://github.com/Joystream/joystream/blob/giza_staging/tests/network-tests/src/QueryNodeApi.ts)

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,383 @@
...
+    )) as number[]
+
+    return createdIds
+  }
+
+  /**
+    Creates a bunch of content entities.
+  */
+  private async createCommonEntities<T>(count: number, createPromise: (index: number) => Promise<T>): Promise<T[]> {
+    const createdIds = await Array.from(Array(count).keys()).reduce(async (accPromise, index: number) => {
```
Lezek123: I think you should parse the CLI output to find the created entity id.
In the future the CLI will be adjusted to make this output easier to parse, but it should be still possible now with some simple regexp.

tests/network-tests/src/flows/workingGroup/leaderSetup.ts
```diff
@@ -61,6 +63,15 @@ async function leaderSetup(
   assert.notEqual(hiredLead, undefined, `${group} group Lead was not hired!`)
   assert(hiredLead!.role_account_id.eq(leadKeyPair.address))
 
+  // setup (ensure) CLI connection to node and query node
+
+  await cli.setApiUri()
+
+  await cli.setQueryNodeUri()
```
Lezek123: This doesn't seem like a good place for this setup

Lezek123 (changes requested):
Please merge the latest `giza_staging` to your branch, it looks like a lot of the code will be affected by that (I mentioned some examples in the comments).

cli/src/base/AccountsCommandBase.ts
ondratra: Fixed. Only `selectedMemberId` is loaded now (via `membershipById`).

cli/src/base/ApiCommandBase.ts
ondratra: Reverted. I didn't realize there is `none === null` setting that makes query node optional.

cli/src/commands/account/chooseMember.ts
ondratra: Done

cli/src/commands/account/chooseMember.ts
ondratra: Done - flag changed from `address` to `memberId`. 

I was reworking `AccountChoose` from Sumer release since something like that was missing in Giza and mistakenly considered address as 1st class citizen here, but it should be a member, as you pointed out.

query-node/mappings/content/utils.ts
ondratra: Typecast `wasFullyActive as IVideoActiveStatus` is no longer needed in that place. Thanks to that, no change is needed for this line.

query-node/mappings/content/utils.ts
ondratra: `testEnv`/`productionEnv` has been removed.

This was debugging relic from the time when I was skipping acceptance `thumbnailPhoto` and `media` data storage objects.

query-node/mappings/content/utils.ts
ondratra: Removed.

This time is a debugging relic that was needed at one time to prevent a compilation error.

query-node/mappings/content/utils.ts
ondratra: Fixed. Now will decrease counter for `oldEntity` only when the video was previously active.

Code changed to 
```
if (oldEntity && initFullyActive) {
  // if video was fully active before, prepare to decrease counter
  const counterChange = -1
  await updateSingleEntity(oldEntity, counterChange)
}
```

query-node/mappings/content/utils.ts
ondratra: Fixed. Now will increase the counter for `oldEntity` only when the video is now active.

Code changed to 
```
if (newEntity && nowFullyActive) {
  // if video is fully active now, prepare to increase counter
  const counterChange = 1

  await updateSingleEntity(newEntity, counterChange)
}
```

query-node/mappings/storage/index.ts
ondratra: Done.

query-node/mappings/storage/index.ts
ondratra: Done. 

NOTE: This implementation relies on dataObject being associated to either thumbnail or video, but not both together. In current runtime implementation, that's always true (and will likely ever be).

tests/network-tests/src/Api.ts
ondratra: Multiple imports of the same `uri` into CLI throws an error. The functionality was reverted back with only exception being randomness into ````const uri = `${this.miniSecret}//testing//${suriPath}//${uuid().substring(0, 8)}` ````. This should not interfere with the rest of the code.

tests/network-tests/src/Api.ts
ondratra: Done

tests/network-tests/src/Api.ts
ondratra: Done

tests/network-tests/src/consts.ts
ondratra: Constant updated. This made extra repeats for QN queries in `activeVideoCounters` fixture obsolete.

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
ondratra: Obsolete code. Removed.

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
ondratra: Fixed. Cast to `any` no longer needed there thx to typecasting `getterName`
````
const getterName = `get${entityName[0].toUpperCase()}${entityName.slice(1)}` as 'getChannels' | 'getChannelCategories' | 'getVideoCategories'
````

File: tests/network-tests/src/flows/workingGroup/leaderSetup.ts (hunk 27 above)
Comment (ondratra) [reply tests/network-tests/src/flows/workingGroup/leaderSetup.ts]: CLI init has been removed from there.

File: tests/network-tests/src/fixtures/content/activeVideoCounters.ts (hunk 24 above)
Comment (ondratra) [reply tests/network-tests/src/fixtures/content/activeVideoCounters.ts]: The function has been removed and the existing `initStorage` flow is used instead.

File: tests/network-tests/src/fixtures/content/activeVideoCounters.ts (hunk 23 above)
Comment (ondratra) [reply tests/network-tests/src/fixtures/content/activeVideoCounters.ts:33]: This has been split into 3 fixtures that makes senses on their own.

File: tests/network-tests/src/CliApi.ts (hunk 19 above)
Comment (ondratra) [reply tests/network-tests/src/CliApi.ts]: CLIs has been merged together.

File: query-node/mappings/storage/index.ts (hunk 13 above)
Comment (ondratra) [reply query-node/mappings/storage/index.ts]: Because if we load entities via `getMany` in the beginning, the changes that will be made into their related entities (channels, etc.) will not be reflected. 

That's because the current implementation of Hydra seems to return duplicate/non-identical objects. So if we have `VideoA` that belongs to `ChannelA` and `VideoB` that also belongs to `ChannelA`, if we use `getMany` with `relations: ['channel']` parameter, we will receive two different `channelB` objects. If we update `activeVideoCounters` on one of them, the second one will be unaffected.

This algorithm could definitely be improved, but it would require a proper unit test due to the complexity of all possible situations. That seems to be too much work at the moment, and there is a chance that the behavior will change after we migrate to Subsquid (making the effort needles).

File: tests/network-tests/src/fixtures/content/activeVideoCounters.ts (hunk 26 above)
Comment (ondratra) [reply tests/network-tests/src/fixtures/content/activeVideoCounters.ts]: The id is actually parsed from CLI output. This function `createCommonEntities` is just a way how to create many entities at once and aggregate all ids together. Parsing is done inside of `createPromise` callback.

File: query-node/schemas/storage.graphql (hunk 16 above)
Comment (ondratra) [reply query-node/schemas/storage.graphql:221]: I agree that it's a little confusing, but I find `video` prefix better than the suffix there. Do you think `videoParentThumbnail`/`videoParentMedia` is ok?

File: cli/src/base/AccountsCommandBase.ts [hunk 28]
```diff
@@ -352,10 +355,21 @@ export default abstract class AccountsCommandBase extends ApiCommandBase {
     return this.selectedMember
   }
 
+  private async getKnownMembers(allowedIds?: MemberId[]): Promise<ISelectedMember[]> {
+    const membersEntries = allowedIds
+      ? await this.getApi().memberEntriesByIds(allowedIds)
+      : await this.getApi().allMemberEntries()
+    const availableMemberships = await Promise.all(
```
Lezek123: Redundant `Promise.all`

tests/network-tests/src/cli/base.ts
```diff
@@ -38,17 +43,53 @@ export abstract class CLI {
...
+            cwd: this.rootPath,
+          })
+
+          // return outputs and exit code
+          return {
+            ...execOutputs,
+            exitCode: 0,
+          }
+        } catch (error: unknown) {
+          const errorTyped = error as ExecFileException & { stdout: string; stderr: string }
```
Lezek123:
```suggestion
          const errorTyped = error as ExecFileException & { stdout?: string; stderr?: string }
```

tests/network-tests/src/cli/joystream.ts
```diff
@@ -14,34 +20,196 @@ export class JoystreamCLI extends CLI {
...
+
+    if (stderr) {
+      throw new Error(`Unexpected CLI failure on choosing account: "${stderr}"`)
+    }
+  }
+
+  /**
+    Creates a new channel.
+  */
+  async createChannel(channel: unknown): Promise<number> {
```
Lezek123: Why not use `chnnel: ChannelInputParameters`?

tests/network-tests/src/cli/joystream.ts
```diff
@@ -14,34 +20,196 @@ export class JoystreamCLI extends CLI {
...
+    const jsonFile = this.tmpFileManager.jsonFile(channel)
+
+    const { stdout, stderr, exitCode } = await this.run('content:createChannel', [
+      '--input',
+      jsonFile,
+      '--context',
+      'Member',
+    ])
+
+    if (exitCode && !this.containsWarningNoStorage(stderr)) {
```
Lezek123: There is a specific exit code reserved for this case, so it'd be better to check for exit codes other than `ExitCodes.ActionCurrentlyUnavailable`, ie:
```typescript
    if (exitCode && exitCode !== ExitCodes.ActionCurrentlyUnavailable) {
```

Actually, you shouldn't ever recieve an exitCode here, since you're using `requireSuccess === true`

tests/network-tests/src/cli/joystream.ts
```diff
@@ -14,34 +20,196 @@ export class JoystreamCLI extends CLI {
...
+
+  /**
+    Selects active member for CLI commands.
+  */
+  async chooseMemberAccount(memberId: MemberId) {
+    const { stderr } = await this.run('account:chooseMember', ['--memberId', memberId.toString()])
+
+    if (stderr) {
+      throw new Error(`Unexpected CLI failure on choosing account: "${stderr}"`)
+    }
```
Lezek123: Why would you use `stderr` to determine if there was an error instead of relying on `requireSuccess` and the `exitCode`? In the future CLI version everything will be logged to `stderr` and `stdout` will only contain simple output values like created entity id.

tests/network-tests/src/cli/joystream.ts
```diff
@@ -14,34 +20,196 @@ export class JoystreamCLI extends CLI {
...
+
+    const { stdout, stderr, exitCode } = await this.run(
+      'content:createVideo',
+      ['--input', jsonFile, '--channelId', channelId.toString()],
+      undefined,
+      !canOmitUpload
+    )
+
+    // prevent error from CLI that create
+    if (canOmitUpload && exitCode && !this.containsWarningNoStorage(stderr)) {
```
Lezek123: As suggested before:
```
exitCode && exitCode !== ExitCodes.ActionCurrentlyUnavailable
```

tests/network-tests/src/QueryNodeApi.ts
```diff
@@ -119,4 +120,40 @@ export class QueryNodeApi {
...
+        channels {
+          id
+          activeVideosCounter
+        }
+      }
+    `
+    return await this.queryNodeProvider.query({ query })
+  }
+
+  public async getChannelCategories(): Promise<ApolloQueryResult<any>> {
```
Lezek123: Please use typesafe queries, like those above (`channelById`, `getDataObjectsByIds`).
You can do this by adding queries to `tests/network-tests/src/graphql/queries`, running `yarn workspace network-tests generate:types:graphql` and importing the generated types here.

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -0,0 +1,146 @@
...
+      | 'getChannels'
+      | 'getChannelCategories'
+      | 'getVideoCategories'
+    await this.query.tryQueryWithTimeout(
+      () => this.query[getterName](),
+      (tmpEntity) => {
+        const entities = (tmpEntity as any).data[entityName]
+        assert(entities.length > 0) // some entities were loaded
+
+        const entity = entities.find((item: any) => item.id === entityId.toString())
```
Lezek123: Why are you querying for all entities, instead of just a single one by `entityId`?

tests/network-tests/src/fixtures/content/createChannelsAndVideos.ts
```diff
@@ -0,0 +1,222 @@
...
+      this.videoCount,
+      this.createdItems.channelIds[0],
+      this.videoCategoryId
+    )
+
+    const { storageBucketId, storageGroupWorkerId, storageGroupWorkerAccount } = await this.retrieveBucket()
+
+    this.debug('Accepting content to storage bag')
+    const allAssetIds = this.createdItems.videosData.map((item) => item.assetContentIds).flat()
+    await this.api.acceptPendingDataObjects(
```
Lezek123: Wouldn't it be easier/better to assume those tests are beeing ran in a setup with a working Argus / Colossus, like the one in `tests/network-tests/run-full-tests.sh`?
In that case:
- There would be no need to allow for the CLI uploads to fail
- There would be no need to manually accept pending data objects
- This will be more in the spirit of integration tests (we'll be testing a realistic scenario, instead of executing some additional extrnisics manually)

tests/network-tests/src/fixtures/content/createContentStructure.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+    Creates a new video category. Can only be executed as content group leader.
+  */
+  private async createVideoCategories(count: number): Promise<number[]> {
+    const createdIds = (await this.createCommonEntities(count, (index) =>
+      this.cli.createVideoCategory({
+        ...getVideoCategoryDefaults(index),
+      })
+    )) as number[]
```
Lezek123: Unnecessary assertion

tests/network-tests/src/fixtures/content/createContentStructure.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+    Creates a new channel category. Can only be executed as content group leader.
+  */
+  private async createChannelCategories(count: number): Promise<number[]> {
+    const createdIds = (await this.createCommonEntities(count, (index) =>
+      this.cli.createChannelCategory({
+        ...getChannelCategoryDefaults(index),
+      })
+    )) as number[]
```
Lezek123: Unnecessary assertion

tests/network-tests/src/scenarios/combined.ts
```diff
@@ -37,8 +37,10 @@ scenario(async ({ job }) => {
...
-  const createChannelJob = job('create channel via CLI', createChannel)
-  job('init storage and distribution buckets via CLI', [initDistributionBucket, initStorageBucket]).after(
-    createChannelJob
-  )
+  const initBucketsJob = job('init storage and distribution buckets via CLI', [
+    initDistributionBucket,
+    initStorageBucket,
+  ]).requires(leadSetupJob)
+
+  const createChannelJob = job('create channel via CLI', createChannel).requires(initBucketsJob)
```
Lezek123: `createChannelJob` should be before `initBucketsJob`, because it assumes the storage setup prepared by `/tests/network-tests/run-full-tests.sh` (and `initBucketsJob` modifies this setup by adding new buckets)

Lezek123 (changes requested):
Aside from things pointed out in the comments below, I noticed a lot of unused imports / variables in the `network-tests` workspace. Are you using an eslint plugin for vscode? It should help you spot those. Alternatively you could add `"noUnusedLocals": true,` to `tsconfig.json` and then run `yarn workspace network-tests build`

The mappings seem to work as expected :+1:

cli/src/base/AccountsCommandBase.ts
ondratra: Fixed

tests/network-tests/src/cli/base.ts
ondratra: Fixed

tests/network-tests/src/cli/joystream.ts
ondratra: `ChannelInputParameters` is not exactly 1:1 with what is accepted there (for example, `ChannelInputParameters.category` is of type `Long`, but `number` is actually required here), but with a little change, it works here. Fixed

tests/network-tests/src/cli/joystream.ts
ondratra: Fixed

tests/network-tests/src/cli/joystream.ts
ondratra: Fixed.

tests/network-tests/src/cli/joystream.ts
ondratra: Fixed. The `stderr` was parsed there because of weird exit codes returned by CLI, but the cause of that has been identified and workaround implemented. (See https://github.com/Joystream/joystream/issues/3251)

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
ondratra: Changed to loading entities by id.

tests/network-tests/src/fixtures/content/createChannelsAndVideos.ts
ondratra: That's a good idea. Storage and distribution services are used there now.

tests/network-tests/src/fixtures/content/createContentStructure.ts
ondratra: Fixed

tests/network-tests/src/scenarios/combined.ts
ondratra: Fixed.

tests/network-tests/src/fixtures/content/createContentStructure.ts
ondratra: Fixed

tests/network-tests/src/QueryNodeApi.ts
ondratra: Fixed

ondratra:
The latest `master` was merged into the branch, and this PR now targets `master` instead of `giza_staging`.

tests/network-tests/src/cli/utils.ts
```diff
@@ -18,7 +18,7 @@ export class TmpFileManager {
     )
     mkdirSync(this.tmpDataDir, { recursive: true })
     nodeCleanup(() => {
-      rmSync(this.tmpDataDir, { recursive: true, force: true })
+      //      rmSync(this.tmpDataDir, { recursive: true, force: true })
```
Lezek123: Is there a reason for commenting it out, or was it a mistake?

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
```diff
@@ -122,21 +122,16 @@ export class ActiveVideoCountersFixture extends BaseQueryNodeFixture {
...
     await this.query.tryQueryWithTimeout(
-      () => this.query[getterName](),
+      () => this.query[getterName](entityId.toString()),
       (tmpEntity) => {
-        const entities = (tmpEntity as any).data[entityName]
-        assert(entities.length > 0) // some entities were loaded
-
-        const entity = entities.find((item: any) => item.id === entityId.toString())
+        assert(tmpEntity)
+        const entity = tmpEntity as Exclude<typeof tmpEntity, null>
```
Lezek123: `Utils.assert` provides typesafety support for this case

tests/network-tests/start-storage.sh
```diff
@@ -0,0 +1,22 @@
+TMP=$0
+THIS_DIR=`dirname $TMP`
+
+# Run proposals tests first, since they require no leads hired
+$THIS_DIR/run-test-scenario.sh proposals
```
Lezek123: Running proposals tests as part of `start-storage.sh` script makes it less general-purpose and reusable.

This is just a special case we need to take into account in `run-full-tests.sh`, as those (proposals) tests affect the hired leads.

Generally, if we expect the new leads hired to have `WORKER_ID=0`, we can setup a storage with:
```
set -a
. ../../.env
set +a

HOST_IP=`$THIS_DIR/get-host-ip.sh`
export COLOSSUS_1_URL="http://${HOST_IP}:3333"
export COLOSSUS_1_TRANSACTOR_KEY=$(docker run --rm --pull=always docker.io/parity/subkey:2.0.1 inspect ${COLOSSUS_1_TRANSACTOR_URI} --output-type json | jq .ss58Address -r)
export DISTRIBUTOR_1_URL="http://${HOST_IP}:3334"
$THIS_DIR/run-test-scenario.sh init-storage-and-distribution
docker-compose -f $THIS_DIR/../../docker-compose.yml up -d colossus-1
docker-compose -f $THIS_DIR/../../docker-compose.yml up -d distributor-1
```
(just like in `start.sh`)

Lezek123 (changes requested):
Just a few minor things and I think we're ready to merge!

tests/network-tests/src/cli/utils.ts
ondratra: This was commented out for debug purposes. Fixed. On Olympia there is a better mechanism built in to debug CLI results.

tests/network-tests/src/fixtures/content/activeVideoCounters.ts
ondratra: `Utils.assert` is used here now, thx.

tests/network-tests/start-storage.sh
ondratra: Amount of code in `start-storage.sh` was reduced. Proposals scenario is run in `run-full-tests.sh` only. Done.

Lezek123 (approved):
LGTM!

mnaamani:
@Lezek123 I recollect that this work been merged into Olympia branch as part of the integration tests migration?
I ask because I'm seeing many changes to cli and network tests so hopefully all merge conflicts have already been addressed if so.

ondratra:
> @Lezek123 I recollect that this work been merged into Olympia branch as part of the integration tests migration? I ask because I'm seeing many changes to cli and network tests so hopefully all merge conflicts have already been addressed if so.

Yes, the feature is already merged in Olympia - it was merged with basic QN NFT features because it shared the code of integration tests. (https://github.com/Joystream/joystream/pull/3195)

PR #2965, opened 2021-12-23
content migration script: add excludeVideoIds argument

mnaamani:
Add optional argument `--excludeVideoIds` to filter videos which we want to exclude from migration.


Lezek123 (approved):
LGTM

PR #2966, opened 2021-12-23
Migration scripts: Fix for --forceChannelOwnerMemberId=0

Lezek123:
SSIA


mnaamani (approved):
Good one, hadn't tested with member id 0 which I'm assuming that how you found this bug.

Issue #2961, opened 2021-12-22
Add metadata to the create_bounty extrinsic


jodator:
The `create_bounty` [extrinsic](https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/bounty/src/lib.rs#L727) parameters are missing `metadata` field:

[https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/bounty/src/lib.rs#L212-L240](https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/bounty/src/lib.rs#L212-L240)

The `metadata` should hold description-like fields for bounty, similarly to other entities like `Member` or `Proposal`. [https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/membership/src/lib.rs#L196-L197](https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/membership/src/lib.rs#L196-L197)

The `metadata` structure will be provided by @thesan while implementing the bounties schema for query-node.

shamil-gadelshin:
The `create_bounty` extrinsic contains the metadata separated from the `BountyParameters` structure:         

https://github.com/Joystream/joystream/blob/273a6240b14d5c7cd9e234a227510e7b69c77ec3/runtime-modules/bounty/src/lib.rs#L727

We use separated parameters in order to reduce storage size because we don't need the metadata for the smart-contract logic - so we keep it externally in the form of the QueryNode data. @jodator

jodator:
@shamil-gadelshin Thanks for the clarification. I don't know how, but I missed that on the `create_bounty` signature.

PR #2968, opened 2021-12-23
Add bounty metadata

thesan:
As part of the development on pioneer is blocked due to #2961, I'm adding on a separate PR from the graphql schema.


metadata-protobuf/proto/Bounty.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto2";
+
+message BountyMetadata {
+  optional string title = 1; // Bounty title
+  optional string description = 2; // Bounty description
+  optional uint32 discussionThread = 3; // Id of the forum thread used to discuss the bounty
```
ondratra: I think `discussionThread` should be of type `uint64` because threads are defined as a 64-bit number in runtime. https://github.com/Joystream/joystream/blob/ed663cccf7f0a3d2cef2abdbf7e416233911ec2a/runtime/src/primitives.rs#L85

ondratra (changes requested):

metadata-protobuf/proto/Bounty.proto
thesan: Thanks I missed that

thesan:
@ondratra different tests fails at different times. Here it's [council election failures](https://github.com/Joystream/joystream/runs/4703813806?check_suite_focus=true#step:8:4525), but on my laptop it's just *proposals & proposal discussion* and on #2945 it's [*forum threads* and *forum thread tags*](https://github.com/Joystream/joystream/runs/4667350788?check_suite_focus=true#step:8:4484). Any ideas ?

Lezek123:
The reason why forum tests fail is that we need to update `@joystream/types` after https://github.com/Joystream/joystream/commit/1c37a713511cfaf35b9577069f78827492324f2c which removed `metadata_hash` from `Thread` struct.

As for _council election failures_ test, it looks like the announcment period number changed between `await this.api.untilCouncilStage('Announcing')` and `await this.api.untilCouncilStage('Voting')`. Perhaps block "reserve" of 3 (3 seconds) was not enough to submit all candidacies. Perhaps @ondratra can look into it.

ondratra:
Fix for forum types and increase of `reserve` blocks for council tests is done in PR https://github.com/thesan/joystream/pull/4 that can be merged into this PR.

I'm not sure if the fix for council tests is a long-term solution because, as I understand it, the tests were failing only when run with limited hardware resources (such as CI virtual machines). Since resources allocated for tests are quite unpredictable, we might want to revisit the issue and find a better solution if it keeps making problems.

ondratra (approved):
LGTM

thesan:
@ondratra thanks. Who is able to merge this PR ?

PR #2971, opened 2021-12-28
chain-spec-builder bump version

mnaamani:
Bumping version as suggested in https://github.com/Joystream/joystream/pull/2955#pullrequestreview-840375553


shamil-gadelshin (approved):
LGTM

PR #2982, opened 2022-01-03
Giza: Make public services reachable on all interfaces

mnaamani:
Let public services listen on all interfaces to make them more easily accessible from other services running in containers from a different docker-compose network. (They will be reachable on their public interface address)


Lezek123 (approved):
LGTM

PR #2987, opened 2022-01-04
Giza: docker-compose: add missing env var for latest version of orion

mnaamani:
Latest version of orion which is published on docker hub `joystream/orion:latest` is now the main "backend" endpoint for Atlas, and needs to talk to a query-node.

Added configuration in docker-compose.


kdembler (approved):

Lezek123 (approved):
LGTM

PR #2991, opened 2022-01-05
DevOps - Github action to deploy node network

ahhda:
Github action to deploy node-network using CloudFormation and Ansible.

The following secrets needs to be setup in the repository for the action to work:

* AWS_ACCESS_KEY_ID
* AWS_SECRET_ACCESS_KEY
* SSH_PRIVATE_KEY

> :warning:  NOTE: To trigger the workflow_dispatch event, your workflow must be in the default branch.

### How to run
* Open the file `.pipelines/deploy-node-network-inputs.json`
* Edit the parameter values according to your preference
* Copy the contents of the json file
* Go to Github Actions > Deploy node network and click on Run workflow
* Paste the contents of the json file and click Run workflow
 Screenshot 2022-01-07 at 12 51 43 PM
* After the action runs successfully you can download the artifacts (node binary, keys, auth data, chain-spec file) from the summary tab
 Screenshot 2022-01-07 at 12 57 01 PM

> :warning:  NOTE: To extract the 7z file on a Mac use something other than the Archive Utility app as it has some problems with opening 7z password protected files.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703601304526) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/deploy-node-network.yml
+++ b/.github/workflows/deploy-node-network.yml
@@ -3,42 +3,10 @@ name: Deploy node network
 on:
   workflow_dispatch:
     inputs:
-      gitRepo:
-        description: 'Code repository'
-        required: false
-        default: 'https://github.com/Joystream/joystream.git'
-      branchName:
-        description: 'Branch to deploy'
-        required: false
-        default: 'master'
-      keyName:
-        description: 'SSH key pair on AWS'
-        required: false
-        default: 'anuj-key'
-      numberOfValidators:
-        description: 'Number of validators to deploy'
-        required: false
-        default: '2'
-      validatorInstanceType:
-        description: 'AWS EC2 instance type for Validators (t2.micro, t2.large)'
-        required: false
-        default: 't2.large'
-      buildInstanceType:
-        description: 'AWS EC2 instance type for Build Instance (t2.micro, t2.large)'
-        required: false
-        default: 't2.xlarge'
-      rpcInstanceType:
-        description: 'AWS EC2 instance type for Validators (t2.micro, t2.large)'
-        required: false
-        default: 't2.large'
-      ec2AMI:
-        description: 'Pre-built AMI ID (ami-0090c9c49dc765fbb)'
-        required: false
+      jsonInput:
+        description: 'Json input (refer to .pipelines/deploy-node-network-inputs.json)'
+        required: true
         default: ''
-      networkSuffix:
-        description: 'Network suffix that will be added to the network name'
-        required: false
-        default: '8129'
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-node-network.yml
```diff
@@ -0,0 +1,160 @@
...
+          key: ${{ secrets.SSH_PRIVATE_KEY }}
+          options: |
+            --inventory inventory
+            --extra-vars "network_suffix=${{ steps.myoutputs.outputs.networkSuffix }} data_path=mydata \
+                          number_of_validators=${{ steps.myoutputs.outputs.numberOfValidators }} \
+                          git_repo=${{ steps.myoutputs.outputs.gitRepo }} deployment_type=live \
+                          bucket_name=${{ steps.deploy_stack.outputs.S3BucketName }} \
+                          branch_name=${{ steps.myoutputs.outputs.branchName }} \
+                          initial_members_file=${{ steps.myoutputs.outputs.initialMembersFilePath }} \
+                          initial_balances_file=${{ steps.myoutputs.outputs.initialBalancesFilePath }}"
```
mnaamani, line 144: there was a variable also to select staging or production type

.pipelines/deploy-node-network-inputs.json
```diff
@@ -0,0 +1,46 @@
...
+    "value": "8129"
+  },
+  "initialBalancesFile": {
+    "description": "HTTP Link to the Initial Balances file",
+    "value": ""
+  },
+  "initialMembersFile": {
+    "description": "HTTP Link to the Initial Members file",
+    "value": ""
+  }
```
mnaamani, line 53: I think we should probably have a password here that can be passed to encrypt the artifact files produces with the secrets generated. Or we can keep it simple, don't include secrets in the artifact, just the chainspec and expect the user to ssh into the build instance and get the necessary files?

.github/workflows/deploy-node-network.yml
mnaamani: I recently tested the deploy-infra script where I omitted the initial members file but somehow the playbook still tried adding the argument to include in the chain spec builder. Perhaps you can look into that. I had updated to latest version of ansible maybe that has something to do with it.

mnaamani:
Looks good, just a couple of suggestions.

Who can access the artifacts. I know you have to be signed into github, but then who has permission to fetch them?

mnaamani (changes requested):
deploy-node-network is note really the best name really for the action.
I suggest to rename it to just `Deploy Network` (similar to our other action `Deploy Playground`)

ahhda:
> Who can access the artifacts. I know you have to be signed into github, but then who has permission to fetch them?

Anyone logged into Github can access and download the artifacts for a public repository.

.pipelines/deploy-node-network-inputs.json
ahhda: Added `encryptionKey` parameter. We now encrypt the artifacts.

PR #2998, opened 2022-01-07
QN - tweaks for council CI testing

ondratra:
This PR is increasing `blockReserve` value in wait-for functions in council QN tests. Some behavior seems to behave differently on CI test servers than local machines due to different machines' performance limits.

This PR is never meant to be merged and only tries to find a solution for failing tests in https://github.com/Joystream/joystream/pull/2968.


ondratra:
Council tests made it through, thus, validating the changes. This PR is no longer needed.

PR #3007, opened 2022-01-10
metadata-protobuf: do not run build script on package install

mnaamani:
While preparing to publish the package I tried to installed locally `npm pack`ed package and got:

```
npm i

> protobufjs@6.11.2 postinstall /Users/mokhtar/tmp/package/node_modules/protobufjs
> node scripts/postinstall

npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
```

The `prepublish` script was running on package install.

This PR fixes it as recommended in the warning message. (The same way we do for our @joystream/types package)

### DIFF
--- a/metadata-protobuf/package.json
+++ b/metadata-protobuf/package.json
@@ -27,8 +27,7 @@
     "lint": "eslint ./src --ext .ts",
     "checks": "tsc --noEmit --pretty && prettier ./ --check && yarn lint",
     "format": "prettier ./ --write",
-    "prepublishOnly": "npm run prepack",
-    "prepack": "npm run checks && npm run build"
+    "prepack": "npm run build && npm run checks"
   },
   "files": [
     "lib/**/*",
--- a/metadata-protobuf/package.json
+++ b/metadata-protobuf/package.json
@@ -27,7 +27,8 @@
     "lint": "eslint ./src --ext .ts",
     "checks": "tsc --noEmit --pretty && prettier ./ --check && yarn lint",
     "format": "prettier ./ --write",
-    "prepublish": "yarn build"
+    "prepublishOnly": "npm run prepack",
+    "prepack": "npm run checks && npm run build"
   },
   "files": [
     "lib/**/*",
--- a/metadata-protobuf/package.json
+++ b/metadata-protobuf/package.json
@@ -19,6 +19,10 @@
   "author": "Joystream Contributors",
   "license": "MIT",
   "private": false,
+  "publishConfig": {
+    "access": "public",
+    "registry": "https://registry.npmjs.org"
+  },
   "scripts": {
     "build": "yarn compile && rm -rf lib && tsc",
     "compile": "yarn ts-node ./scripts/compile.ts",


metadata-protobuf/package.json
```diff
@@ -27,7 +27,8 @@
     "lint": "eslint ./src --ext .ts",
     "checks": "tsc --noEmit --pretty && prettier ./ --check && yarn lint",
     "format": "prettier ./ --write",
-    "prepublish": "yarn build"
+    "prepublishOnly": "npm run prepack",
```
Lezek123: According to https://docs.npmjs.com/cli/v8/using-npm/scripts, `prepack` is always executed during `npm publish`, so this seems redundant

metadata-protobuf/package.json
```diff
@@ -27,7 +27,8 @@
...
+    "prepack": "npm run checks && npm run build"
```
Lezek123: In general I'd advise running `checks` after `build`, because some of the checks can only be executed against a built code (ie. circular references check, or in case of `@joystream/types` for example - checks against autogenerated files). This is not really the case for `@joystream/metadata-protobuf` currently, but it would be good to have those executed in the right order just in case some checks are added in the future or in case someone decides to use those scripts as a reference.

Lezek123 (commented):
Looks good, just added 2 small suggestions

metadata-protobuf/package.json
mnaamani: resolved in 3eba5de

metadata-protobuf/package.json
mnaamani: resolved in 3eba5de

mnaamani:
Tested with `npm pack` and `npm publish --dry-run`

mnaamani:
Also added publish config to package.json
```
  "publishConfig": {
    "access": "public",
    "registry": "https://registry.npmjs.org"
  }
```

Lezek123 (commented):
I'm running into a strange issue with `npm publish --dry-run`, not sure if it's worth investigating.

When I clear `node_modules` and re-install with `yarn`, then cd into `metadata-protobuf` and run `npm publish --dry-run` I'm getting:
```
$ npm publish --dry-run

> @joystream/metadata-protobuf@1.0.0 prepack .
> npm run build && npm run checks


> @joystream/metadata-protobuf@1.0.0 build /home/leszek/projects/joystream/joystream-ws-2/metadata-protobuf
> yarn compile && rm -rf lib && tsc

yarn run v1.22.10
$ yarn ts-node ./scripts/compile.ts
$ /home/leszek/projects/joystream/joystream-ws-2/node_modules/.bin/ts-node ./scripts/compile.ts
installing semver@^7.1.2
installing jsdoc@^3.6.3
installing semver@^7.1.2
installing jsdoc@^3.6.3
/home/leszek/projects/joystream/joystream-ws-2/metadata-protobuf/compiled/index.js updated
Error: Cannot find module 'jsdoc/jsdoc.js'
Require stack:
- /home/leszek/projects/joystream/joystream-ws-2/node_modules/protobufjs/cli/pbts.js
- /home/leszek/projects/joystream/joystream-ws-2/metadata-protobuf/scripts/compile.ts
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.resolve (internal/modules/cjs/helpers.js:98:19)
    at callJsdoc (/home/leszek/projects/joystream/joystream-ws-2/node_modules/protobufjs/cli/pbts.js:105:55)
    at exports.main (/home/leszek/projects/joystream/joystream-ws-2/node_modules/protobufjs/cli/pbts.js:96:9)
    at Object.<anonymous> (/home/leszek/projects/joystream/joystream-ws-2/metadata-protobuf/scripts/compile.ts:22:5)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Module.m._compile (/home/leszek/projects/joystream/joystream-ws-2/node_modules/@endemolshinegroup/cosmiconfig-typescript-loader/node_modules/ts-node/src/index.ts:1056:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Object.require.extensions.<computed> [as .ts] (/home/leszek/projects/joystream/joystream-ws-2/node_modules/@endemolshinegroup/cosmiconfig-typescript-loader/node_modules/ts-node/src/index.ts:1059:12)
    at Module.load (internal/modules/cjs/loader.js:950:32)
error Command failed with exit code 1.
```
Running `npm run build` separately before `npm publish --dry-run` seems to fix the issue. `npm pack` works without any issues.

Neither adding back `prepublishOnly` nor changing `npm run build` to `yarn build` has any effect on this issue, so I suppose we can just ignore it, since it's not critical.

mnaamani:
I was able to reproduce the same problem. I can't say I know the exact reason but I'm sure it is because of mixing use of yarn and npm.. maybe it makes more sense to run `yarn publish` instead of `npm publish` ?

PR #3018, opened 2022-01-11
Giza: zero storage fees

mnaamani:
Make the storage deletion prize 0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,7 +731,7 @@ dependencies = [
 
 [[package]]
 name = "chain-spec-builder"
-version = "3.2.0"
+version = "3.3.0"
 dependencies = [
  "ansi_term 0.12.1",
  "enum-utils",
@@ -2332,7 +2332,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
-version = "5.12.0"
+version = "5.13.0"
 dependencies = [
  "frame-benchmarking",
  "frame-benchmarking-cli",
@@ -2393,7 +2393,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node-runtime"
-version = "9.12.0"
+version = "9.13.0"
 dependencies = [
  "frame-benchmarking",
  "frame-executive",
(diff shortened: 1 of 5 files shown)


dobertRowneySr (approved):

shamil-gadelshin (changes requested):
We need to change the version in the `runtime/lib.rs` as well.

mnaamani:
Despite the change I still see the error `InsufficientBalance` from the storage system.
Can it possibly be due to the rounding up to nearest greater integer?


```
// Calculate data storage fee based on size. Fee-value uses megabytes as measure value.
    // Data size will be rounded to nearest greater MB integer.
    pub(crate) fn calculate_data_storage_fee(bytes: u64) -> BalanceOf<T> {
        let mb_fee = Self::data_object_per_mega_byte_fee();

        const ONE_MB: u64 = 1_048_576;

        let mut megabytes = bytes / ONE_MB;

        if bytes % ONE_MB > 0 {
            megabytes += 1; // rounding to the nearest greater integer
        }

        mb_fee.saturating_mul(megabytes.saturated_into())
    }
```

mnaamani:
> We need to change the version in the `runtime/lib.rs` as well.

Oh yes I totally missed that :D ty

mnaamani:
> Despite the change I still see the error `InsufficientBalance` from the storage system. Can it possibly be due to the rounding up to nearest greater integer?
> 
> ```
> // Calculate data storage fee based on size. Fee-value uses megabytes as measure value.
>     // Data size will be rounded to nearest greater MB integer.
>     pub(crate) fn calculate_data_storage_fee(bytes: u64) -> BalanceOf<T> {
>         let mb_fee = Self::data_object_per_mega_byte_fee();
> 
>         const ONE_MB: u64 = 1_048_576;
> 
>         let mut megabytes = bytes / ONE_MB;
> 
>         if bytes % ONE_MB > 0 {
>             megabytes += 1; // rounding to the nearest greater integer
>         }
> 
>         mb_fee.saturating_mul(megabytes.saturated_into())
>     }
> ```

of course it cannot be that lol, as long as mb_fee is zero. Most likely I was running wrong runtime because I didn't bump the version as pointed out.

mnaamani:
Ok thanks for helping with debugging @shamil-gadelshin the issue was that in my test environment I had forgotten to set the data fee to zero. Atlas works now with new membership with 0 balance.

shamil-gadelshin (approved):
LGTM

PR #3027, opened 2022-01-11
Content migration script: Fix alternativeEndpoints

Lezek123:
Adds missing `queryNodeApi` to `AssetsManager.create` call, allowing the `AssetsManager` to use alternative Sumer storage node endpoints when the default storage endpoint(s) fail(s).

Adds more `debug` logs and allows controlling the log level through `DEBUG` env, ie:
```
# Log level = debug:
DEBUG=true yarn migration-scripts ...

# Log level = info (default):
yarn migration-scripts ...
```


mnaamani (approved):
Tested with storage-1.joystream.org off and confirmed with DEBUG logs that other nodes were selected.

PR #3043, opened 2022-01-14
Allow to filter entity by WorkingGroup id directly

thesan:
PR for https://github.com/Joystream/joystream/issues/3095

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201710183771244) by [Unito](https://www.unito.io)


query-node/schemas/workingGroups.graphql
```diff
@@ -34,6 +34,9 @@ type Worker @entity {
   "The group that the worker belongs to"
   group: WorkingGroup!
 
+  "The id the group that the worker belongs to"
+  groupId: ID!
```
ondratra, line 40: This PR creates `groupId` attribute, but never explicitely uses it. It reIies on internal mechanics of Hydra that sets up this field automatically, I'm not mistaken. We could add comment about that.

```suggestion
  # This exposes internal Hydra value related to `group`. It will be internally set and updated by Hydra.
  # See https://github.com/Joystream/joystream/pull/3043 for more info.
  "The id the group that the worker belongs to"
  groupId: ID!
```

ondratra (approved):
LGTM. I just added one commentary suggestion.

PR #3053, opened 2022-01-18
Olympia - Fix the OracleJudgment type

thesan:
The way `OracleJudgment` was defined:
```ts
export class OracleJudgment extends JoyEnum({
  Winner: OracleJudgment_Winner,
  Rejected: Null,
}) {}
```
Is not usable to map `OracleJudgmentSubmitted` each entry status. Plus in the runtime code [`OracleJudgment`](https://github.com/Joystream/joystream/blob/2d970a177ee2acbc1063eede7116b87c8994a659/runtime-modules/bounty/src/lib.rs#L459) is:
```rust
pub type OracleJudgment<EntryId, Balance> = BTreeMap<EntryId, OracleWorkEntryJudgment<Balance>>;
```


Lezek123 (approved):
LGTM

PR #3055, opened 2022-01-19
Olympia membership migration.

shamil-gadelshin:
For this issue: https://github.com/Joystream/joystream/issues/3096

#### Changes
* runtime (removed name from the genesis config)
* import membership data on the query-node genesis
* updated metadata-protobuf compiled files (side-effect)
 _UPDATED_ I've rebased on the latest olympia and compiled metadata-protobuf files gone because of the latest changs

#### Runtime genesis data
* handle (hashed)
* root account
* controller account

Member ID is preserved.

#### Query-node genesis data
* handle
* about
* avatar URI
* member ID
* root account
* controller account

All other data were set to default values.

#### Comments
The genesis import for the query-node requires `members.json` file (exported data) to be set in the `.query-node/mappings/src/genesis-data` folder. As was agreed with mnaamani

* an additional modification of the Docker file (outside of the scope of this PR) is required to pass this data to the `query-node` service.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201710230154052) by [Unito](https://www.unito.io)


mnaamani:
Needs update from olympia ref the query-node code. My fault for talking a while to review this. Olympia has been updated from giza since the PR was opened.
genesis.ts is renamed to bootstrap.ts 

example from giza for member bootstrapping:   https://github.com/Joystream/joystream/blob/1a2e90b2e2888b53ccf5e85c289706ef6dffb649/query-node/mappings/bootstrap.ts#L32

Runtime code looks correct.

query-node/mappings/src/bootstrap-data/index.ts
```diff
@@ -1,11 +1,11 @@
-import { MemberJson, StorageSystemJson, WorkingGroupJson, MembershipSystemJson } from './types'
+import { StorageSystemJson, WorkingGroupJson, MembershipSystemJson } from './types'
 import storageSystemJson from './data/storageSystem.json'
 import membersJson from './data/members.json'
 import workingGroupsJson from './data/workingGroups.json'
 import membershipSystemJson from './data/membershipSystem.json'
 
 const storageSystemData: StorageSystemJson = storageSystemJson
-const membersData: MemberJson[] = membersJson
```
mnaamani: I see why you removed the type here, as you tried to accommodate to use the same .json file from the export scripts in https://github.com/Joystream/runtime-state-export-tools/blob/master/src/export_members.ts 
so same file can be used for query-node and for chainspec builder `--initial-members`.

For background the 'MemberJson' type is defined as it is for the fetch script from https://github.com/Joystream/joystream/blob/master/query-node/mappings/bootstrap-data/scripts/fetchMembersData.ts which we used during sumer -> giza migration.

This is resulting in need to do type casting in `initMembers()`  in bootstrap.ts.

I don't think it is worth sacrificing type safety which I'm sure @Lezek123 will not approve. So why don't we just update `MemberJson` to match?

PR #3088, opened 2022-01-24
Giza. Add storage buckets bag assignments.

shamil-gadelshin:
We need to fix a bug in the storage pallet: storage buckets deletion doesn't change the bag.

#### Changes
- runtime(storage pallet)
- types
- chain-spec.json

#### Storage pallet changes
- added `assigned_bag` field to the `StorageBucket`
- updated tests
- modified extrinsics and methods
  - `delete_storage_bucket`
  - `delete_dynamic_bag`
  - `create_dynamic_bag`


yarn.lock
```diff
@@ -3142,7 +3142,7 @@
     lodash "^4.17.15"
     moment "^2.24.0"
 
-"@joystream/warthog@2.41.2", "@joystream/warthog@~2.41.2":
+"@joystream/warthog@2.41.2", "@joystream/warthog@^2.40.0":
```
mnaamani: this change doesn't seem correct. When I ran yarn build:packages in your branch it reverted to:
`"@joystream/warthog@2.41.2", "@joystream/warthog@~2.41.2":`

mnaamani: perhaps you are not using the same version of yarn. (which may be the case if you haven't installed volta)

shamil-gadelshin:
```
$ yarn --version
1.22.15

$ volta --version
1.0.5
```

mnaamani:
Implementation looks correct.
I think we should bump the runtime spec to 14, node to 5.14.0 and chainspec builder to 3.4.0 as usual

runtime-modules/storage/src/lib.rs
```diff
@@ -2657,7 +2687,15 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
 
         <Bags<T>>::remove(&bag_id);
 
-        Self::change_bag_assignments(&BTreeSet::new(), &deleted_dynamic_bag.distributed_by);
+        Self::change_bag_assignments_for_distribution_buckets(
+            &BTreeSet::new(),
+            &deleted_dynamic_bag.distributed_by,
```
bedeho, line 2692: Should we, perhaps in a future iteration, add commitment in extrinsic to this length, as in principle this set could be different to what caller thinks, for the purposes of weight function? Same for `stored_by`, below.

If you agree, please make new issues for this with correct labels, in particular `storage-pallet`, `runtime` and `mainnet`.

bedeho (approved):
Looks good :)

mnaamani (approved):

PR #3115, opened 2022-01-26
Giza: Playground fix

mnaamani:
- Allows COLOSSUS_1_URL and DISTRIBUTOR_1_URL to be configured outside of start.sh
- Fix renamed env variables above in github workflow playground deployment
- Allow custom prefix to be added to cloud formation stack name for playground deployment to more easily identify it in list of stacks in aws console.


mnaamani:
please review @ahhda

Issue #3121, opened 2022-01-27
Giza type issues


traumschule:
1. jsstats using `@joystream/types": "^0.17.1`
```
$ yarn
yarn install v1.22.17
[1/4] Resolving packages...
success Already up-to-date.

$ yarn build
./node_modules/@joystream/types/JoyEnum.js
SyntaxError: /home/joystream/jsstats/node_modules/@joystream/types/JoyEnum.js: Missing class properties transform.
   6 | function JoyEnum(types) {
   7 |     return class JoyEnumObject extends codec_1.Enum.with(types) {
>  8 |         static typeDefinitions = types;
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   9 |         typeDefinitions = JoyEnumObject.typeDefinitions; // Non-static version
  10 |         static create(registry, typeKey, value) {
  11 |             return new JoyEnumObject(registry, { [typeKey]: value });
```

2. The cli has issues too: `Unable to resolve type StorageBucketsPerBagValueConstraint`
```
Compiling joystream-node v5.14.0 (/home/joystream/joystream/node)                                                                                           
    Finished release [optimized] target(s) in 5m 04s

joystream$ cd cli
joystream/cli$ ./bin/run working-groups:openings -g operationsAlpha
(node:14465) [MODULE_NOT_FOUND] Error Plugin: @joystream/cli: Cannot find module '/home/joystream/joystream/node_modules/@joystream/metadata-protobuf/lib/utils.js'
module: @oclif/config@1.17.0
task: toCached
plugin: @joystream/cli
root: /home/joystream/joystream/cli
See more details with DEBUG=*
(repeated like five million times)
2022-01-27 09:30:17        REGISTRY: Unable to resolve type StorageBucketsPerBagValueConstraint, it will fail on construction
2022-01-27 09:30:17        API/INIT: Error: FATAL: Unable to initialize the API: createType(StorageBucketsPerBagValueConstraint):: DoNotConstruct: Cannot construct unknown type StorageBucketsPerBagValueConstraint
    at EventEmitter.value (/home/joystream/joystream/node_modules/@polkadot/api/base/Init.cjs:82:25)                                                          
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    Error: FATAL: Unable to initialize the API: createType(StorageBucketsPerBagValueConstraint):: DoNotConstruct: Cannot construct unknown type               
    StorageBucketsPerBagValueConstraint

[...]

PR #3134, opened 2022-01-28
Types and metadata-protobuf: Patch to target es2017

Lezek123:
Targetting `esnext` in `tsconfig.json` for reusable packages like `@joystream/types` and `@joystream/metadata-protobuf` turned out to be problematic, because not all tools will support the cutting-edge version of ECMAScript, see point 1. in [this issue](https://github.com/Joystream/joystream/issues/3121)

In this PR I switched the target to `es2017` for those packages and bumped the minor version number.

### DIFF
--- a/metadata-protobuf/package.json
+++ b/metadata-protobuf/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/metadata-protobuf",
-  "version": "1.0.0",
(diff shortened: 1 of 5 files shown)


mnaamani (approved):
Assuming joystream-cli work well without need to depend on these newer versions.

PR #3135, opened 2022-01-28
Integration tests: Run CLI tests after workerPayoutsJob

Lezek123:
Fixes occasionally failing integration tests, ie.: https://github.com/Joystream/joystream/runs/4983207184?check_suite_focus=true

When `workerPayoutsJob` is executed in parallel to `createChannelJob`, there may be a significant delay between [getting the current block number](https://github.com/Lezek123/substrate-runtime-joystream/blob/integration-tests-fix/tests/network-tests/src/fixtures/workingGroupModule.ts#L347) and [sending the `fillOpening` tx with `nextPaymentBlock === now.add(this.firstPayoutInterval`](https://github.com/Lezek123/substrate-runtime-joystream/blob/integration-tests-fix/tests/network-tests/src/fixtures/workingGroupModule.ts#L353) which may cause `FillOpeningInvalidNextPaymentBlock`

### DIFF
--- a/tests/network-tests/src/scenarios/combined.ts
+++ b/tests/network-tests/src/scenarios/combined.ts
@@ -26,7 +26,11 @@ scenario(async ({ job }) => {
   job('at least value bug', atLeastValueBug).requires(leadSetupJob)
 
   // tests minting payouts (requires council to set mint capacity)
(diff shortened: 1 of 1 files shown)


mnaamani:
Great, thanks for fixing this

Issue #3119, opened 2022-01-26
Name working groups in Pioneer v1


bedeho:
The abstract names are not helpful:

- `Alpha` -> Builders
- `Beta` -> Human Resources
- `Gamma` -> Marketing

Since these names are indeed used in CLI , it could be sensible to somehow mention these cryptic names still as well, but less prominent.

traumschule:
Done.

PR #3136, opened 2022-01-30
Rename working groups in Pioneer v1

DzhideX:
This PR addresses the changes proposed here: https://github.com/Joystream/joystream/issues/3119

Updated in:
- Tokenomics page
- Proposals page
- Working Groups page

I haven't updated the descriptions for these new working groups (yet) as I think Ben could probably do a better job. We also probably want to mention which group name here corresponds to which group name in the runtime (like Bedeho says to reduce confusion) as not only are they used in CLI but also in the chain state tab and more..

Edit: I've went ahead and added descriptions on my own. Let me know if it needs changing!

### DIFF
--- a/pioneer/packages/joy-proposals/src/forms/GenericWorkingGroupProposalForm.tsx
+++ b/pioneer/packages/joy-proposals/src/forms/GenericWorkingGroupProposalForm.tsx
@@ -14,6 +14,8 @@ import { usePromise, useTransport } from '@polkadot/joy-utils/react/hooks';
 import PromiseComponent from '@polkadot/joy-utils/react/components/PromiseComponent';
 import { WorkerData } from '@polkadot/joy-utils/types/workingGroups';
 import { LeadInfo } from '@polkadot/joy-utils/react/components/working-groups/LeadInfo';
+import styled from 'styled-components';
+import _ from 'lodash';
 
 export type FormValues = GenericFormValues & {
   workingGroup: WorkingGroupKey;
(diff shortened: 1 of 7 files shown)


Lezek123 (changes requested):
Looks good, but a few suggestions:
- `Opportunities` tab (`working-groups/opportunities`) still shows "old" working groups names in filter dropdown as opening labels etc., there is no reference to `Builders` / `Marketing` / `HR`
- On Tokenomics page only new names are shown and there is no reference to `Alpha`, `Beta`, `Gamma`, which I think goes against what's been stated in [the issue](https://github.com/Joystream/joystream/issues/3119):
    > Since these names are indeed used in CLI , it could be sensible to somehow mention these cryptic names still as well, but less prominent.

DzhideX:
Thank you for these suggestions! I went and addressed them but some notes wrt them:
1. I've updated the filter dropdown and opening labels but I couldn't find any other ones (that aren't user input) as the code is a bit convoluted (for example I thought maybe `<OpeningTitle/>`/`JobSpecifics.title` but wasn't quite sure). Let me know if there are any other ones that I may have missed!
2. I wasn't too sure where to put them as the amount of space in the table is already pretty sparse. I've put the extra data into the `helpText` for each of them and hopefully that should be enough?

Edit: Initially I also didn't want to clutter the space in the filter/label but maybe we can add a (for example) `(Operations Group ...)` there or would that be too much?

Lezek123 (commented):
I think that's good enough, although I was imagining `Operations Working Group X` more like a small text under the "common" working group name, for example:

Working groups overveiw:
image

Opportunities filtering:
image

Tokenomics:
image

_Those are just quick examples, so they're not very well styled, but you get the idea._

Let me know what do you think.

DzhideX:
I think that's actually a great idea. It looks much better that way!

>Those are just quick examples, so they're not very well styled, but you get the idea.

I don't necessarily think so, maybe just the padding/margins need some work.

I'll do it that way then!

DzhideX:
Okay, I've went ahead and implemented it in:
- Tokenomics
- Proposals
- WGs
- WGs Opportunities filter

Since you didn't say anything against my label comment I didn't add any further info there (I assume you agree that it would clutter it)?

Lezek123 (approved):
LGTM!

> Since you didn't say anything against my label comment I didn't add any further info there (I assume you agree that it would clutter it)?

Yeah, I think it looks good as is, there is already enough information on the page now for the user to be able to figure out which group is which.

freakstatic:
@DzhideX I believe that Beta is Marketing and Gamma is HR :sweat:

freakstatic:
image

Source: https://blog.joystream.org/sumer-kpis/

freakstatic:
in this PR it seems to be the other way around
image

PR #3140, opened 2022-02-01
DevOps - Add query node to CloudFormation node-network deployment [WIP]

ahhda:
This PR adds starting a query-node stack along with a new chain using CloudFormation and Ansible.

Steps to run:

```
./deploy-infra.sh ~/Joystream/bash-config.1.cfg
```

### DIFF
--- a/devops/aws/roles/rpc/tasks/main.yml
+++ b/devops/aws/roles/rpc/tasks/main.yml
@@ -22,10 +22,48 @@
     enabled: yes
   become: yes
 
+- name: Build required packages
+  shell: yarn build:packages
+  args:
+    chdir: '{{ remote_code_path }}'
+  # Run in async fashion for max duration of 1 hr
+  async: 3600
+  poll: 0
+  register: build_result
+
+- name: Check on build async task
+  async_status:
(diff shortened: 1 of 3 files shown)


devops/aws/roles/rpc/tasks/main.yml
```diff
@@ -22,10 +22,48 @@
...
+    jid: '{{ build_result.ansible_job_id }}'
+  register: job_result
+  until: job_result.finished
+  # Max number of times to check for status
+  retries: 36
+  # Check for the status every 100s
+  delay: 100
+
+- name: Run query node containers
+  command: yarn workspace query-node-root start:dev
```
mnaamani, line 45: the `.env` which is picked up by docker-compose has `JOYSTREAM_NODE_WS=ws://joystream-node:9944/` assumes the local joystream-node was run as docker-container.

we need to override this for indexer to work correctly.
So setting it to `wss://{{ inventory_hostname }}.nip.io/ws-rpc` or simply `ws://{{ inventory_hostname }}:9944/`

mnaamani (changes requested):
needs a fix for the query-node config, otherwise looks good.

PR #3142, opened 2022-02-01
DevOps: update aws cloudformation sample configs for giza

mnaamani:
Update default AMI in sample config to use Giza release.

cc: @ahhda


ahhda (approved):
Looks good

PR #3155, opened 2022-02-01
Query node: Fix expectedEndingTimestamp / expectedStart

Lezek123:
Small fix for working-groups mappings - the timestamp should be provided in seconds, but was treated as miliseconds when converted to `Date`


ondratra (approved):

ondratra (approved):
LGTM

PR #3158, opened 2022-02-01
build-node-docker without docker-compose

mnaamani:
- build script checks local repo before fetching from docker-hub
- immediately tags image with code hash
- build without docker-compose, remove build directive in docker-compose. benefits:
   -  on mac if local image not found, docker pulls but doesn't use the pulled image and proceeds to build
   - avoid tagging arbitrary image through docker-compose build that doesn't match code hash

### DIFF
--- a/.env
+++ b/.env
@@ -62,4 +62,6 @@ DISTRIBUTOR_2_WORKER_ID=1
 DISTRIBUTOR_2_ACCOUNT_URI=//testing//worker//Distribution//${DISTRIBUTOR_2_WORKER_ID}
 
 # joystream/node docker image tag
-JOYSTREAM_NODE_TAG=latest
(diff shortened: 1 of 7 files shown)


mnaamani:
cc: @ahhda could you test this

ahhda (approved):
Tested on both mac and linux. Works well.

### Manual testing
Mac (`yarn start`)
Screenshot 2022-02-02 at 3 28 19 PM

Linux (`yarn build:node:docker`)
Screenshot 2022-02-02 at 4 08 38 PM

mnaamani:
Went further to make it required to be explicit in specifying which tag to use when starting up joystream-node service.
This will be more useful in olympia where we have multiple compiled versions from same commit. (test node vs production node, and soon staging node) where various proposal config constants are defined at compile time.

PR #3164, opened 2022-02-03
Olympia: Playground updates

mnaamani:
Various fixes primarily to deploy playground, and fix multi-arch docker build.
- Update cargo package `librocksdb-sys` which was not building correctly for armv7. Since olympia is going to be a new chain it is safe to use newer version. Docker builds now works in the build-node-docker action so caching images is back.
- Fix in Deploy Playground to async wait for build to complete when a new joystream-node image is being built (when there is no cached image)
- Apply same idea from https://github.com/Joystream/joystream/pull/3158
- Install newer docker-compose in setup.sh


mnaamani:
Not at all clear to me why the benchmarking CI check is failing.. https://github.com/Joystream/joystream/runs/5047098186?check_suite_focus=true

Was something merge into Olympia that broken it.
cc: @shamil-gadelshin

mnaamani:
Just for context the benchmarks seem to not be in sync with the runtime:

```
cargo build --release --features runtime-benchmarks
```

error output 

```
     Compiling joystream-node-runtime v10.0.0 (/Users/mokhtar/joystream/joystream/runtime)
  error[E0432]: unresolved import `crate::ContentDirectoryWorkingGroup`
     --> /Users/mokhtar/joystream/joystream/runtime/src/runtime_api.rs:290:17
      |
  290 |             use crate::ContentDirectoryWorkingGroup;
      |                 ^^^^^^^----------------------------
      |                 |      |
      |                 |      help: a similar name exists in the module: `ContentWorkingGroup`
      |                 no `ContentDirectoryWorkingGroup` in the root

  error[E0432]: unresolved import `crate::StorageV2`
     --> /Users/mokhtar/joystream/joystream/runtime/src/runtime_api.rs:300:17
      |
  300 |             use crate::StorageV2;
      |                 ^^^^^^^---------
      |                 |      |
      |                 |      help: a similar name exists in the module: `Storage`
      |                 no `StorageV2` in the root

  error: aborting due to 2 previous errors

  For more information about this error, try `rustc --explain E0432`.
  error: could not compile `joystream-node-runtime`
```

from runtime/src/runtime_api.rs

```
    #[cfg(feature = "runtime-benchmarks")]
    impl frame_benchmarking::Benchmark<Block> for Runtime {
        fn dispatch_benchmark(
            config: frame_benchmarking::BenchmarkConfig
        ) -> Result<Vec<frame_benchmarking::BenchmarkBatch>, sp_runtime::RuntimeString> {
            use sp_std::vec;
            use frame_benchmarking::{Benchmarking, BenchmarkBatch, add_benchmark, TrackedStorageKey};

            use pallet_session_benchmarking::Module as SessionBench;
            use frame_system_benchmarking::Module as SystemBench;
            use frame_system::RawOrigin;
            use crate::ProposalsDiscussion;
            use crate::ProposalsEngine;
            use crate::ProposalsCodex;
            use crate::Constitution;
            use crate::Forum;
            use crate::Members;
            use crate::ContentDirectoryWorkingGroup;  // This is from pre-sumer I think
            use crate::Utility;
            use crate::Timestamp;
            use crate::ImOnline;
            use crate::Council;
            use crate::Referendum;
            use crate::Bounty;
            use crate::Blog;
            use crate::JoystreamUtility;
            use crate::Staking;
            use crate::StorageV2; // ??
```

So I assume the benchmarks for olympia were not actually updated and we can safely ignore this for now?

mnaamani:
Updated to correct names in [239a4c3](https://github.com/Joystream/joystream/pull/3164/commits/239a4c3a36aac4ef7390adec1fc875ee5e2f74df) which seems to have resolved the issue.

shamil-gadelshin:
> Not at all clear to me why the benchmarking CI check is failing.. https://github.com/Joystream/joystream/runs/5047098186?check_suite_focus=true
> 
> Was something merge into Olympia that broken it. cc: @shamil-gadelshin

Looks like an incorrect merge because the error indicates the old code.

mnaamani:
Disabled generation of weights for storage module for now as advised by shamil

ahhda (commented):
A minor issue I faced when running the setup:

1. When we run `yarn start`, we get this warning/error because we haven't set the `COLOSSUS_1_TRANSACTOR_URI` env variable. We can either add instructions to set this in the README or provide a default value in the `start.sh` script.

`Io(Os { code: 6, kind: Uncategorized, message: "No such device or address" })`

Screenshot 2022-02-02 at 3 10 46 PM

README.md
```diff
@@ -77,7 +77,14 @@ A step by step guide to setup a full node and validator on the Joystream testnet
 ### Integration tests
 
 ```bash
-tests/integration-tests/run-full-tests.sh
+# Make sure yarn packages are built
+yarn build:packages
+
+# Build the test joystream-node
+TEST_NODE=true yarn build:node:docker
```
Lezek123, line 84: `TEST_NODE` env will not be passed to docker build context, which will still result in standard node beeing built (added a related comment below)

build-node-docker.sh
```diff
@@ -2,23 +2,31 @@
...
-  docker-compose build joystream-node
+  # Not found, try to fetch from remote repo
+  echo "Trying to fetch cached ${IMAGE} image"
+  docker pull ${IMAGE} || :
+
+  # If we didn't find it, build it
+  if ! docker inspect ${IMAGE} > /dev/null;
+  then
+    echo "Building ${IMAGE}.."
+    docker build . --file joystream-node.Dockerfile --tag ${IMAGE}
```
Lezek123:
```suggestion
    docker build . --file joystream-node.Dockerfile --tag ${IMAGE} --build-arg TEST_NODE=${TEST_NODE}
```

build-node-docker.sh
```diff
@@ -2,23 +2,31 @@
...
 set -e
 
-if ! command -v docker-compose &> /dev/null
-then
-  echo "docker-compose not found. Skipping docker image builds."
-  exit 0
-fi
+# Looks for a cached joystream/node image matching code shasum.
+# Search order: local repo then dockerhub. If no cached image is found we build it.
+# Finally image is tagged as "latest"
```
Lezek123: Seems like this last step is not actually done

mnaamani:
> 1. When we run `yarn start`, we get this warning/error because we haven't set the `COLOSSUS_1_TRANSACTOR_URI` env variable. We can either add instructions to set this in the README or provide a default value in the `start.sh` script.

Good catch. That was supposed to have been picked up from the `.env` file actually, but since we dropped the loading of the `.env` file into the start.sh script its blank. Its actually a problem on master as well. 

Luckily the initStorage flow in the "setupNewChain" scenario which runs following that issue, got our back, and it defaults to same expected value in .env which is `//Colossus1`.

Can you suggest a way to solve this without doing equivalent of:
```
set -a
. ./.env
set +a
```

Because env variables in the shall should have precedence over any vars set in .env and that approach would break that.

Lezek123 (changes requested):

README.md
mnaamani: Interesting, I also tested this on mac and it seemed to work. Looks like different behavior between mac desktop docker and linux. I will commit your suggestion in https://github.com/Joystream/joystream/pull/3164#discussion_r798465612

mnaamani: I think given this, we will probably want to add an extra `ARG ALL_PROPOSALS_PARAMETERS_JSON` to the Dockerfile as I was considering the need to have more than the current two possible configurations (defaults from node build and the `integration-tests/proposal-parameters.json`), one more config for production "live". Unless @bwhm is okay with having staging networks use same voting/grace periods as production.

build-node-docker.sh
mnaamani: Yes I was toying with idea of never tagging `latest` when building this script and had commented out the final tagging step. So the only image ever tagged with latest will be the one built from master branch and published to docker-hub.

What do you think?

README.md
mnaamani: > Interesting, I also tested this on mac and it seemed to work. Looks like different behavior between mac desktop docker and linux. I will commit your suggestion in [#3164 (comment)](https://github.com/Joystream/joystream/pull/3164#discussion_r798465612)

My bad, I was only looking at the computed hash and seeing it was different. That might explain why running the tests locally thinking the test node was running, the proposals job was sometimes failing?

Lezek123: Yes, I think that with default proposal parameters, where grace period may be mutiple days etc., those tests are bound to fail

build-node-docker.sh
Lezek123: Since `joystream-node` service does not use the `latest` tag anymore I think this step can be skipped

mnaamani: cleaned up comments in [3747ada](https://github.com/Joystream/joystream/pull/3164/commits/3747adaa3ecb2067971c0f6cc349c1473a157d65)

mnaamani:
I've observing something strange. With 

The code shasum computed on github runner when TEST_NODE arg is set:
`744949833f811c538c10113aead8e3c59bd1ad87` 
This is as shown in run integration tests run: - https://github.com/Joystream/joystream/runs/5056914354?check_suite_focus=true#step:6:12

However I checked on ubuntu and osx running ./scripts/runtime-code-shasum.sh is producing:
`6719f0e3dc493b1dfd09843d4c36c87ccc424ca3`

Without TEST_NODE being set both github runner, and running script on ubuntu and osx machines is producing same value of `920d676f233066d875a3dcb8723d0d135bc0b1e9`  - https://github.com/Joystream/joystream/runs/5056913755?check_suite_focus=true#step:6:1

Can you confirm you get the same values as I do.

My hunch is `sed` on github runner is behaving differently than on local machines..

mnaamani:
> My hunch is `sed` on github runner is behaving differently than on local machines..

testing my hunch in https://github.com/Joystream/joystream/pull/3170

mnaamani:
> > 1. When we run `yarn start`, we get this warning/error because we haven't set the `COLOSSUS_1_TRANSACTOR_URI` env variable. We can either add instructions to set this in the README or provide a default value in the `start.sh` script.
> 
> Good catch. That was supposed to have been picked up from the `.env` file actually, but since we dropped the loading of the `.env` file into the start.sh script its blank. Its actually a problem on master as well.
> 
> Luckily the initStorage flow in the "setupNewChain" scenario which runs following that issue, got our back, and it defaults to same expected value in .env which is `//Colossus1`.
> 
> Can you suggest a way to solve this without doing equivalent of:
> 
> ```
> set -a
> . ./.env
> set +a
> ```
> 
> Because env variables in the shall should have precedence over any vars set in .env and that approach would break that.

One crude way of achieving this:

```
# Load the variables from .env in local function shell variables and get specific one of interest
function load_my_var() {
  set +a
  . ./.env
  echo $MY_VAR
}

# re-export variable of interest setting it to value from environment if set, or falling back to value loaded from .env file
export MY_VAR=${MY_VAR:=$(load_my_var)}
```

I'm sure there is a better way, certainly having function reusable by taking the variable name as argument would be good start. Some resources that might help. [1](https://www.digitalocean.com/community/tutorials/how-to-read-and-set-environmental-and-shell-variables-on-linux#common-environmental-and-shell-variables) [2](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/V3_chap02.html#tag_18_06_02) [3](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)

Lezek123:
> I've observing something strange. With
> The code shasum computed on github runner when TEST_NODE arg is set:
744949833f811c538c10113aead8e3c59bd1ad87
This is as shown in run integration tests run: - https://github.com/Joystream/joystream/runs/5056914354?check_suite_focus=true#step:6:12
> However I checked on ubuntu and osx running ./scripts/runtime-code-shasum.sh is producing:
6719f0e3dc493b1dfd09843d4c36c87ccc424ca3
> Without TEST_NODE being set both github runner, and running script on ubuntu and osx machines is producing same value of 920d676f233066d875a3dcb8723d0d135bc0b1e9 - https://github.com/Joystream/joystream/runs/5056913755?check_suite_focus=true#step:6:1
> Can you confirm you get the same values as I do.
> My hunch is sed on github runner is behaving differently than on local machines..

Hmm... Strangely enough it seems like on commit https://github.com/Joystream/joystream/pull/3164/commits/5463e4f0943c339625588cb43695082ba3e317e1 the CI was producing [the same hash we're getting locally now (`6719f0e3dc493b1dfd09843d4c36c87ccc424ca3`)](https://github.com/Joystream/joystream/runs/5049815529?check_suite_focus=true), but it changed to `744949833f811c538c10113aead8e3c59bd1ad87` after https://github.com/Joystream/joystream/pull/3164/commits/51e523c66f479c4006a744ac6e2976d1642fa87b

mnaamani:
> Hmm... Strangely enough it seems like on commit [5463e4f](https://github.com/Joystream/joystream/commit/5463e4f0943c339625588cb43695082ba3e317e1) the CI was producing [the same hash we're getting locally now (`6719f0e3dc493b1dfd09843d4c36c87ccc424ca3`)](https://github.com/Joystream/joystream/runs/5049815529?check_suite_focus=true), but it changed to `744949833f811c538c10113aead8e3c59bd1ad87` after [51e523c](https://github.com/Joystream/joystream/commit/51e523c66f479c4006a744ac6e2976d1642fa87b)

Interesting, but I can't see `scripts/runtime-code-shasum.sh` is producing different hashes. `./build-node-docker.sh` is not part of the files that the `tar` includes.

I'm wondering if you and I both ran the benchmarks locally and there are some `.rs` artifact files produced in node/ or runtime/ which .gitignore is ignoring but making it into the computed hash? But that can't be the case, [my test](https://github.com/Joystream/joystream/pull/3170) to not include use of `sed` is now resulting in same hash value.

mnaamani:
I updated olympia-playground from olympia but cli and integration-tests builds are failing there.

mnaamani:
They don't seem to tricky to fix:
cli 
```
$ tsc --noEmit --pretty && prettier ./ --check && yarn lint
src/commands/content/video.ts:21:28 - error TS2339: Property 'in_series' does not exist on type 'Video'.

21         'InSeries': aVideo.in_series.unwrapOr('NONE').toString(),
                              ~~~~~~~~~

src/commands/content/videos.ts:29:25 - error TS2339: Property 'in_series' does not exist on type 'Video'.

29           'InSeries': v.in_series.unwrapOr('NONE').toString(),
```

I'll push a fix

Lezek123 (approved):
Looks good to me.
Even the test node shasum now seems to match the one calculated within the CI:
```
$ TEST_NODE=true ./scripts/runtime-code-shasum.sh
e0d176b5effee4ca2dec58267219635074298d98
```

mnaamani:
@ahhda could you look over the AWS cloudformation changes and playground actions.

ahhda (approved):
LGTM

ahhda:
> @ahhda could you look over the AWS cloudformation changes and playground actions.

We can change the instance VolumeSize to be an input in the CF template. If required I can make it a part of #3167.

mnaamani:
> We can change the instance VolumeSize to be an input in the CF template. If required I can make it a part of #3167.

Yes please do

PR #3168, opened 2022-02-03
closes #2947 Bounty oracle switch

L3pereira:
"This only applies bounties where oracle is currently member, not council oracles, or coming working group oracles, however member oracles can change oracle to any other type: https://github.com/Joystream/joystream/issues/2979"

As previously discussed the council will also be able to switch oracles.

Oracle can be switched only in the following stages:
Funding period
Working period

Tests
oracle_council_switch_to_oracle_member_successful
oracle_member_switch_to_oracle_council_successful
oracle_member_switch_to_oracle_member_successful
oracle_member_switch_to_same_member_fails_same_new_oracle
oracle_council_switch_to_council_fails_same_new_oracle
oracle_switch_fails_new_oracle_not_a_member
oracle_switch_fails_invalid_stage_successful_withdrawal
oracle_switch_fails_invalid_stage_failed_bounty_withdrawal_without_judgement
oracle_switch_fails_invalid_stage_failed_bounty_withdrawal_with_judgement

Benchmarking 
oracle_council_switch_to_oracle_member
oracle_member_switch_to_oracle_member
oracle_member_switch_to_oracle_council

PR #3169, opened 2022-02-03
 closes #2975 Proposal to switch bounty oracle 

### LINKED ISSUE #2975: Proposal to switch bounty oracle
As an extension of

https://github.com/Joystream/joystream/issues/2947

This should also be possible to do through a proposal from the council, in case an oracle goes truly MiA, however, the bounty creator gets to set a flag during bounty creation which can block this sort of switch.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703601732275) by [Unito](https://www.unito.io)

L3pereira:
In
```rust
pub fn switch_oracle(
            origin,
            new_oracle: BountyActor<MemberId<T>>,
            bounty_id: T::BountyId,
        )
```
The oracle can be switched in the following conditions:

1. The origin is the actual oracle (either the council or simply a member), in this case it can nominate a new oracle and fire the event `BountyOracleSwitched`. If origin is not the council nor the actual oracle, a `OriginIsNotCouncilAndIsNotOracle` error is returned.
2. The origin is root (council), but is not the actual oracle, in this case if the flag allow_council_switch_inactive_oracle was set to `true` by the bounty creator at creation time, a successful event is fired `BountyInactiveOracleSwitchedByCouncilApproval`, If the flag is false a `CouncilIsNotAllowedToSwitchInactiveOracle` error is returned

The new oracle and the origin must not be the same, otherwise a `MemberIsAlreadyAnOracle` error is returned.

tests
council_switch_by_approval_new_oracle_member_successful
council_switch_by_approval_new_oracle_member_fails_council_flag_switched_off

### DIFF
--- a/runtime-modules/bounty/src/actors.rs
+++ b/runtime-modules/bounty/src/actors.rs
@@ -58,7 +58,7 @@ impl<T: Trait> BountyActorManager<T> {
         }
     }
 
-    // Validate balance is sufficient for the bounty
+    /// Validate balance is sufficient for the bounty
     pub(crate) fn validate_balance_sufficiency(
         &self,
         required_balance: BalanceOf<T>,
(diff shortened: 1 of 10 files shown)


L3pereira:
To be merged after #3168

PR #3170, opened 2022-02-03
runtime code shasum: different tag scheme for test node

mnaamani:
Testing a new approach to add `-test` suffix to docker image tag computed, to avoid use of sed which is giving some inconsistent results between local machine and github runner. see: https://github.com/Joystream/joystream/pull/3164#issuecomment-1029354574


mnaamani:
This is only a half solution, we still need to reflect that the docker build will also modify code (the block interval).

mnaamani:
it doesn't seem that this is required anymore as we are not seeing github runner produce a different / "unexpected" hash.
closing for now

Issue #3166, opened 2022-02-03
DevOps - Deploy Playground improvements


ahhda:
The Deploy Playground workflow (`.github/workflows/deploy-playground.yml`) needs some improvements:

* Delete CloudFormation stack if deployment fails
  * Have a step to check for duplicate stack names before deleting a stack on error. Prevent deletion of existing stacks.
* Save the output URL file as an artifact
* Remove Pioneer deployment
* Better naming for job `Create an EC2 instance and configure docker-compose stack` -> `Deploy Playground`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201765726303131) by [Unito](https://www.unito.io)

PR #3171, opened 2022-02-04
DevOps - Olympia Playground improvements

ahhda:
Resolves issues marked in #3166

### DIFF
--- a/.github/workflows/deploy-playground.yml
+++ b/.github/workflows/deploy-playground.yml
@@ -34,7 +34,7 @@ defaults:
 
 jobs:
   deploy-playground:
-    name: Create an EC2 instance and configure docker-compose stack
+    name: Deploy Playground Job
     runs-on: ubuntu-latest
     env:
       STACK_NAME: ${{ github.event.inputs.stackNamePrefix }}-${{ github.event.inputs.branchName }}-${{ github.run_number }}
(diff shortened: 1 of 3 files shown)


devops/aws/deploy-playground-playbook.yml
```diff
@@ -106,16 +106,28 @@
         caddy_systemd_capabilities_enabled: true
         caddy_update: false
 
+    - name: Set endpoints
+      set_fact:
+        all_services: |
+          The services should now be accesible at:
```
mnaamani: Perhaps it would be best to output this in a format that is easier to parse programmatically. like an env file or even better json (why not both) or both.
Exposing this information over one more endpoint in caddy would also support idea here: https://github.com/Joystream/pioneer/issues/2167#issuecomment-1029636565

.github/workflows/deploy-playground.yml
```diff
@@ -74,3 +84,18 @@ jobs:
...
+
+      - name: Save the endpoints file as an artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: endpoints
+          path: devops/aws/endpoints.txt
+
+      - name: Delete CloudFormation Stack if any step failed
+        # Skip if stack already existed
+        if: failure() && steps.stack_exists.outcome != 'failure'
```
mnaamani, line 96: Would the job progress to this step if "stack_exists" step infact failed ? So it would be sufficient to just check failure.

mnaamani (changes requested):
Just left suggestions for artifact file format, and making it available on an additional caddy endpoint.

.github/workflows/deploy-playground.yml
ahhda: The step `Delete CloudFormation Stack if any step failed` will only run when any of the steps failed except the `stack_exists` step.

Here are the scenarios:
1. Stack does not exist & workflow goes smoothly -> Delete CloudFormation step is skipped
2. Stack exists -> `stack_exists` steps throws failure -> Delete CloudFormation step is skipped
3. Stack does not exist but some ansible step failed -> `failure()` is true  && `steps.stack_exists.outcome` is 'success' -> Delete CloudFormation step runs

devops/aws/templates/Playground-Caddyfile.j2
```diff
@@ -42,3 +42,9 @@
     uri strip_prefix /member-faucet
     reverse_proxy localhost:3002
 }
+
+{{ nip_domain }}/config.json {
+    root * /home/ubuntu
+    rewrite * /config.json
```
mnaamani, line 48: is it endpoints.json or config.json ?
I wouldn't serve up the whole /home/ubuntu directory, even though I see the rewrite rule :)

mnaamani: looks correct actually.

Issue #3163, opened 2022-02-02
Creator payout data format - extended


zeeshanakram3:
This issue is an extension of #2943 and provides a rough solution to problems addressed in the said issue.

The content here also assemble some ideas from the following:

* #2365 
* #2420
* https://github.com/Joystream/joystream/issues/2943#issuecomment-1026701296
* [Claiming Code](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/lib.rs#L1412) and relevant test [functions](https://github.com/Joystream/joystream/blob/6b95450f750d6adc233933b2578951e1eff1bb85/runtime-modules/content/src/tests/fixtures.rs#L1544)

Let
`c_i`  = channel id that is owed some reward
`p_i`  = cumulative reward owed to channel with id `c_i`
`m_i` = some commitment that justifies the amount `p_i` for `c_i`
`d_i`  = description explaining the rationale for for reward or deduction for `c_i`

The set `S` of all payout claims where each claim consists of `(c_i, p_i, m_i, d_i)` is
:

1. Orderable on channel Id.
1. Unique - only one payout entry `{c_i, p_i, m_i, d_i}` exists for each channel `c_i` (emphasizing on this because there are multiple payout entries in the Set for same channel Id in implemented [test cases](https://github.com/Joystream/joystream/blob/6b95450f750d6adc233933b2578951e1eff1bb85/runtime-modules/content/src/tests/fixtures.rs#L1634)

If the above two properties are satisfied then `S` can be represented in the form of the following data structures:

## Ordered Set
The time complexity for:

* Searching the existence of payout for `c_i` in S is _`O(log n)`_
* Updating payout for some already existing channel Id `c_i` to new reward `p_new` is _`O(log n)`_
* Inserting payout for some new channel Id `c_new` is _`O(n)`_

## Binary Search Tree
The time complexity for:

* Searching the existence of payout for `c_i` in S is _`O(log n)`_ (average case)
* Updating payout for some already existing channel Id `c_i` to new reward `p_new` is _`O(log n)`_ (average case)
* Inserting payout for some new channel Id `c_new` is _`O(n)`_ (average case)

However, for unbalanced tree complexity can in order of `n` so AVL tree seems better alternative.

## AVL Tree
The time complexity for:

* Searching the existence of payout for `c_i` in S is _`O(log n)`_
* Updating payout for some already existing channel Id `c_i` to new reward `p_new` is _`O(log n)`_
[...]

PR #3182, opened 2022-02-07
Creator payouts implementation in Joystream.js package

zeeshanakram3:
The code for this implementation exists in the `joystreamjs` package at the root level of monorepo.
 The implementation assumption is that the payout payload would be fetched from some remote over HTTP

 The following CLI commands have been implemented for creator payout management
* [x] `content:updateCreatorPayoutCommitment` Update creator payout payload commitment hash in the runtime storage.
* [x] `content:creatorPayoutRecord` Show payout information for creator given a channel id.
* [x] `content:claimCreatorPayoutReward` Claim creator payout reward for a given channel id.
* [x] `content:creatorPayoutPayloadHeader` Get header from serialized payload file.
* [x] `content:creatorPayoutRecordAtByteOffset` Get creator payout record from serialized payload file at given byte.
* [x] `content:generateCreatorPayoutPayload` Create serialized creator payout payload from JSON input.
* [x] `content:generateCreatorPayoutMerkleRoot `  Generate merkle root for creator payout payload passed as JSON input.
* The PR contains the documentation of these commands

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203042475932932) by [Unito](https://www.unito.io)


bedeho:
1. I think its better to turn the entire format into one message, otherwise (se)deserializing this properly will be more cumbersome.
2. What is purpose of `NonLeafNodeMerkleBranchByteOffset` message?
3. `CreatorPayoutBody::merkle_branches_of_non_leaf_nodes` field ?

zeeshanakram3:
1. Do you mean both header and body should be the single message? Also, doesn't Protobuf provide out-of-the-box functionality to (se)deserializing different message properties?

For 2 & 3, please read the detailed answer:

If the payout vector is`[c_0||p_0||m_0,...,c_7||p_7||m_7]` , then the payload `body` according to [proposed](https://github.com/Joystream/joystream/issues/3163#issuecomment-1029854088) format would be:

```
c_0||p_0||proof_0||d_0 ... c_7||p_7||proof_7||d_7        
```

Now suppose we need to prove the membership of payout record for channel id `c_2`(shown green in the diagram), to prove that and claim a reward from runtime we need to pass the following args to the extrinsic

 ```rust
pub fn claim_channel_reward(
            ...
            proof: Vec<ProofElement<T>>, //  [`proof_3`, `proof_01`, `proof_4567`] (shown blue in the diagram)
            item: PullPayment<T>, // { c_2, p_2, m_2 }
        ) -> DispatchResult {

```
image


### Problem 
The payload body does not have all the Merkle branches required to construct the membership proof. For this specific example, `proof_01` & `proof_4567` are missing from the payload body.
To construct proof we need Merkle branches of both leaf and non-leaf(see diagram for differentiation) nodes, 

### Solution 
Include the remaining Merkle branches in the payload body

### Why need `CreatorPayoutBody::merkle_branches_of_non_leaf_nodes`?
`merkle_branches_of_non_leaf_nodes` is needed to construct the complete Merkle path required to prove the membership of payout entry(`c_i||p_i||m_i`)  for channel id `c_i`.

### Why need `NonLeafNodeMerkleBranchByteOffset`?
It is required to index into Merkle branch (of the non-leaf node) at a given byte offset in the payload body to construct Merkle path

**Comment by zeeshanakram3** [comment]
## Update

> a merkle branch is a full sequence of hashes that make up the required proof

As you said in the discussion, _**another**_ way to solve the problem is (and I assume this is what you actually meant):

to change the structure of payout record, considering the above example, the payout record for channel id `c_2` in the payload body would be:

`c_2||p_2||proof_3||proof_01||proof_4567||d_2`


### Schema Changes
The payout record for single channel would be:
```proto
  message CreatorPayout {
    required uint32 channel_id = 1;  // c_i
    required float cumulative_payout_owed = 2; // p_i
   
    enum Side {
        Left = 0;
        Right = 1;
    }

    message MerkleBranch {
        required bytes merkle_branch = 1;
        required Side side = 2;
    }

    repeated MerkleBranch merkle_branches = 3; 
    required string payout_rationale = 4; // d_i
  }
```

It will be the responsibility of someone who actually uploads the raw payload to provide the correct sequence of hashes that make up the membership proof. And, not the responsibility of client software to calculate the Merkle path and then fetch relevant branches/hashes/nodes that make up the required proof

joystreamjs/content/src/creatorPayout.ts
```diff
@@ -0,0 +1,261 @@
...
+ * fetch creator payouts payload from remote source
+ * @param context `Header | Body` based on part to be fetch from remote source
+ * @param offset If `context` arg is `Body` then offset of payload record must be provided
+ **/
+async function fetchCreatorPayout(
+  context: PayloadContext,
+  offset?: number
+): Promise<CreatorPayoutPayload.Header | CreatorPayoutPayload.Body.CreatorPayout> {
+  // HTTP Url of remote host where payload is stored
+  let url = process.env.CREATOR_PAYOUT_DATA_URL as string
```
kdembler: In the browser env, we won't be able to define this at build time since this URL can be dynamic. So it need to provided via some config/parameter instead

zeeshanakram3: Sure. In fact, that was discussed earlier too, I had planned to change that, Now I will prioritize it.

kdembler: Just making a note, not a priority at all from my perspective, realistically, the earliest time for using this in atlas is like 2 months

zeeshanakram3:
superseded by #4334

Issue #3167, opened 2022-02-03
DevOps - deploy-infra.sh and Deploy Network improvements


ahhda:
* Have a step to check for duplicate stack names before deleting a stack on error. Prevent deletion of existing stacks.
* Remove Pioneer deployment
* Fixes to make it work with Olympia changes
* Add VolumeSize as an input for CloudFormation

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201765726923592) by [Unito](https://www.unito.io)

mnaamani:
Once https://github.com/Joystream/joystream/pull/3055 in merged make sure to update genesis data file for query-node  if deployment includes initial members.

mnaamani:
- Deploy one storage node and one distributor node
- Deploy member faucet
- Similar to suggestion for playground produce artifact with all service endpoints
- Just as the playground uses start.sh and run a setupNewChain scenario, I think we need to have a special scenario suitable for network deployment
- decide what to do about the "build/admin" instance after deployment. terminate it? also there is confusion in the naming across the cf templates and ansible where build and admin are used to refer to the same instance.
- perhaps user deploying needs to supply their own SSH public key to be able to access the instances. ie not a named in the AWS profile account used for deployment. (To be added to ./ssh/authorized_keys)

mnaamani:
Both for deploy network and playground actions, the action should handle the flow being manually cancelled and delete cf stack.

ahhda:
* Add ability to set `ALL_PROPOSALS_PARAMETERS_JSON` as Github Action input

mnaamani:
> * Just as the playground uses start.sh and run a setupNewChain scenario, I think we need to have a special scenario suitable for network deployment

Expanding on this, the scenario run needs access to the sudo key generated for the new chain. This of course is provided in the "secrets" generated by chainspec builder.

For the faucet to function correctly, a couple things are missing in the newChainSetup script:
- ensuring the configured inviting member has enough "invites" granted by the membership working group lead and has sufficient balance as well to send to newly invited members. The inviting member should not be the sudo account btw. Its preferable that a dedicated account be used.

mnaamani:
[...]

PR #3184, opened 2022-02-07
DevOps - Deploy Network improvements

ahhda:
This PR resolves the issues mentioned in #3167 

Task list:
- [x] Have a step to check for duplicate stack names before deleting a stack on error. Prevent deletion of existing stacks.
- [x] Remove Pioneer deployment
- [x] Fixes to make it work with Olympia changes
- [x] Add VolumeSize as an input for CloudFormation
- [ ] Deploy one storage node and one distributor node
- [ ] Deploy member faucet
- [x] Similar to suggestion for playground produce artifact with all service endpoints
- [x] ~~Just as the playground uses start.sh and run a setupNewChain scenario, we need to have a special scenario suitable for network deployment~~ (Part of #3206)
- [x] There is confusion in the naming across the cf templates and ansible where build and admin are used to refer to the same instance
- [x] Terminate Build instance?
- [ ] perhaps user deploying needs to supply their own SSH public key to be able to access the instances. ie not a named in the AWS profile account used for deployment. (To be added to ./ssh/authorized_keys)
- [x] The action should handle the flow being manually cancelled and delete cf stack
- [x] Add ability to set ALL_PROPOSALS_PARAMETERS_JSON as Github Action input
- [x] Copy the initial members.json to `query-node/mappings/src/bootstrap-data/data/members.json`
- [ ] Add `Cache-Control: no-cache` to caddy config

Successful run without parameters json: https://github.com/ahhda/joystream/actions/runs/1845690533
Successful run with parameters json: https://github.com/ahhda/joystream/actions/runs/1845612209


query-node/package.json
```diff
@@ -23,14 +21,17 @@
...
+    "hydra-cli": "./codegen/node_modules/@joystream/hydra-cli/bin/run",
+    "hydra-typegen": "./codegen/node_modules/@joystream/hydra-typegen/bin/run",
     "codegen": "yarn hydra-cli codegen --schema ../../schemas/",
     "codegen:noinstall": "yarn codegen --no-install",
+    "typegen:configure": "NODE_URL=${NODE_URL:-ws://localhost:9000} envsub typegen.template.yml typegen.yml",
     "typegen": "rm -rf ./mappings/generated && yarn hydra-typegen typegen manifest.yml --debug",
     "mappings:build": "yarn workspace query-node-mappings build",
-    "format": "prettier ./ --write"
+    "start:dev": "./start.sh",
+    "kill:dev": "./kill.sh"
```
ahhda: These were part of the `giza` branch but were somehow missed in the merging process I believe.

mnaamani:
Some good improvements here, I parsed through the code and had a couple of points, didn't test deployment yet:

- Each EC2 instance really needs needs different volume size as they have different roles.
    -  For example RPC instance will also be the first storage node and distributor node so has higher requirements.
- Will the build step work if ami already has Joystream directory (when we create pre-built ami's)
- I saw one last reference to "admin" in setup-admin script/playbook I think, so should be renamed
- No docker services are started on the rpc node ?
   - Query-node
   - Colossus, Storage, Faucet, Orion..
- The setupNewChain script is not called to configure the chain - with sudo account and endowed account (taken from secrets generated in chain-spec network configuration step)
- is the initial-members file copied to query-node on the rpc instance? yarn build in query-node must be re-run where the query-node is going to be started.
- I'd rather you don't change make changes to the query-node/package.json at this stage, all the scripts works as intended.

One thing would be worth adding also is to give each validator instance its own aws region so we can have more geographically diverse deployment. We don't want all the instances in one center. (Depending on number of validators selected of course lets try to spread over US, EUROPE, ASIA)

ahhda:
> * Each EC2 instance really needs needs different volume size as they have different roles.
>   
>   * For example RPC instance will also be the first storage node and distributor node so has higher requirements.

Will add this 👍 

> * Will the build step work if ami already has Joystream directory (when we create pre-built ami's)

As part of the `build-code` playbook, the step in `get-code-git` first deletes the directory and then downloads the code from git

```mermaid
flowchart LR
build-code --> get-code-git
get-code-git --> delete[Delete remote code directory if exists]
delete --> checkout[Git checkout]
```

> * I saw one last reference to "admin" in setup-admin script/playbook I think, so should be renamed

Will add this 👍 

> * No docker services are started on the rpc node ?
>   
>   * Query-node
>   * Colossus, Storage, Faucet, Orion..

We do run `yarn workspace query-node-root start:dev` to start query-node. I forgot to replace it with `yarn start` step in the playbook.

> * The setupNewChain script is not called to configure the chain - with sudo account and endowed account (taken from secrets generated in chain-spec network configuration step)

Will add this 👍 

> * is the initial-members file copied to query-node on the rpc instance? yarn build in query-node must be re-run where the query-node is going to be started.

Yes. The initial members' file is downloaded to the location `query-node/mappings/src/bootstrap-data/data/members.json` in the playbook `devops/aws/roles/common/tasks/chain-spec-node-keys.yml`. The query node is built in the RPC part of the playbook later on.

> * I'd rather you don't change make changes to the query-node/package.json at this stage, all the scripts works as intended.

Was getting errors in starting query node for olympia as in giza the step was `start:dev`. I'll change the step in the playbook to run `start` 👍 

> One thing would be worth adding also is to give each validator instance its own aws region so we can have more geographically diverse deployment. We don't want all the instances in one center. (Depending on number of validators selected of course lets try to spread over US, EUROPE, ASIA)

Currently, the instances are spread across multiple data centres in one region. Launching a multi-region stack involves much more involved work. Maybe we can create a separate issue for it.

devops/aws/roles/rpc/tasks/main.yml
```diff
@@ -41,10 +41,13 @@
   # Check for the status every 100s
   delay: 100
 
-- name: Run query node containers
-  command: yarn workspace query-node-root start:dev
+- name: Run service containers
+  command: yarn start
```
mnaamani, line 45: yarn start will also try to start joystream-node service which will try to listen on same port 9944 as node started with systemd.

Issue #3099, opened 2022-01-24
Councils Can Veto Bounty Proposal


bedeho:
Facilitatet this [https://github.com/Joystream/pioneer/issues/2028](https://github.com/Joystream/pioneer/issues/2028).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201905998755806) by [Unito](https://www.unito.io)

bedeho:
This is no longer part of Rhodes scope.

Issue #3098, opened 2022-01-24
Withdraw council bounty funding proposal


bedeho:
Facilitate this https://github.com/Joystream/pioneer/issues/2074.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201906000308442) by [Unito](https://www.unito.io)

bedeho:
This is no longer part of Rhodes scope.

PR #3186, opened 2022-02-07
Olympia feature bounty proposals

dobertRowneySr:
fixes:
- #3099 
- #3098 

## Changes
- Added 2 variants on `ProposalDetails`, 
- Only BountyId is needed since the withdraw bounty funding proposal should be added by `BountyActor::Council`.
- `BountyId` type added  to `proposal_codex::Trait` mirrors `bounty::Trait::BountyId`
- Benchmarks added
- `types` & `metadata` generated


scripts/cargo-build.sh
```diff
@@ -2,4 +2,4 @@
 
 export WASM_BUILD_TOOLCHAIN=nightly-2021-02-20
 
-cargo +nightly-2021-02-20 build --release
+cargo +nightly-2021-02-20 build --release --features runtime-benchmarks
```
shamil-gadelshin, line 5: Why did you change this script? It doesn't look like part of the PR.

runtime/src/proposals_configuration/defaults.rs
```diff
@@ -357,3 +357,30 @@ pub(crate) fn veto_proposal_proposal() -> ProposalParameters<BlockNumber, Balanc
         constitutionality: 1,
     }
 }
+
+// Proposal parameters for the 'Veto Bounty' proposal (same as veto proposal)
+pub(crate) fn veto_bounty_proposal() -> ProposalParameters<BlockNumber, Balance> {
+    ProposalParameters {
+        voting_period: 72200, // The longest grace period
```
shamil-gadelshin, line 364: Did Martin recommend those constants?

runtime-modules/proposals/codex/src/benchmarking.rs
```diff
@@ -234,7 +234,7 @@ benchmarks! {
 
     create_proposal_funding_request {
         let i in 1 .. MAX_FUNDING_REQUEST_ACCOUNTS.try_into().unwrap();
-        let t in ...;
+//        let t in ...;
```
shamil-gadelshin: Why did you comment such lines?

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -135,6 +137,9 @@ pub trait Trait:
         ProposalParameters<Self::BlockNumber, BalanceOf<Self>>,
     >;
 
+    /// Bounty Id type (mirroring bounty::Trait::BountyId)
+    type BountyId: From<u32> + Parameter + Default + Copy;
```
shamil-gadelshin: Is there a reason why you introduce a new type here? Can't we get away without it like with `blog::PostId` for other proposals?

shamil-gadelshin (changes requested):
Looks good. However, this is a major question: how does it come that we have proposals to veto the bounty and don't have for creating one? Is there another PR for it?

Also, I didn't understand why did you comment the title variables in benchmarks for some proposals.  They could affect weights in benchmarking. Also, your local benchmarking runs discarded those variables in the autogenerated runtime file. Did you try to run it several times? It could flip back and forth - so, it's better to keep those parameters included.

scripts/cargo-build.sh
dobertRowneySr: It was required for the benchmarks. Apparently  the `generate_weights` script complains if you build it without

runtime-modules/proposals/codex/src/lib.rs
dobertRowneySr: `blog::PostId` is a `pub` alias for `u64`. `BountyId` is a defined in `bounty::Trait` as 
```rust
type BountyId: From<u32> + Parameter + Default + Copy;
```
So 3 alternative comes to my mind:
1. We move `BountyId` to `common` pallet
2. We define something like `pub type BountyId = u64`; in the `bounty` pallet
3. implement the `bounty::Trait` for `proposal_codex::Trait`

runtime/src/proposals_configuration/defaults.rs
dobertRowneySr: No actually. Those are copied from `veto_proposal_proposal`. Maybe @bwhm can weight in on those

runtime-modules/proposals/codex/src/benchmarking.rs
dobertRowneySr: > Looks good. However, this is a major question: how does it come that we have proposals to veto the bounty and don't have for creating one? Is there another PR for it?

I am not sure on those they were requested by the Pioneer team. Maybe @thesan or @dmtrjsg can provide better answers

> Also, I didn't understand why did you comment the title variables in benchmarks for some proposals. They could affect weights in benchmarking. Also, your local benchmarking runs discarded those variables in the autogenerated runtime file. Did you try to run it several times? It could flip back and forth - so, it's better to keep those parameters included.

Ok

dmtrjsg: According to the designs, bounties are created with its own "create bounty" button, so we never thought that we need a proposal for that. I don't think it is needed to be fair. 

Screenshot from https://pioneer-2-bounties.vercel.app/#/bounty/current

Screenshot 2022-02-16 at 15 11 24

dobertRowneySr: and about the commented lines should be fixed in 12c507d7c0335e3a38dccb0c6d393bf8a96fda88

runtime-modules/proposals/codex/src/lib.rs
dobertRowneySr: Can you weight in @shamil-gadelshin ?

runtime-modules/proposals/codex/src/benchmarking.rs
dmtrjsg: @shamil-gadelshin @ignazio-bovo guys, I think I jumped the gun here with this comment and I may be wrong. 

Functionality-wise Councils can participate in bounties through proposals mechanism.

Council should be able to:

Create Bounty - via Create Bounty proposal that replicates Create Bounty flow (plus some proposal specific-steps)
Contribute Funding to Bounty
Withdraw Funding
And Switch Oracle https://github.com/Joystream/joystream/issues/2975 
(We don't need a Proposal to assign Oracle, since that is part of "Create Bounty" flow)
Veto Bounty

It does not seem to be logical to me for Work entries and work to get submitted via proposals, so this is probably safe to leave out.

Adding @bedeho here, as he has the best view on phasing out the Council participation in Bounties. From purely Pioneer perspective, that was not in my focus/ scope for release.

runtime-modules/proposals/codex/src/lib.rs
shamil-gadelshin: I recommend option 3 similar to `proposals_engine::Trait` and its `ProposalId`

runtime/src/proposals_configuration/defaults.rs
bwhm: Hey,
I think there are two competing forces at play here:
1. You could argue you want a proposal to be open for voting for the entire duration of the `grace_period` of the proposal you want to veto.
2. You could just as easily argue it shouldn't be open for voting any more once the proposal you want to veto is `executed`-

When `olympia` goes live, the longest `grace_period` I have set in the WIP parameters, that is 72200.

Unless it's trivial to have this particular `voting_period` match the remaining duration of the `grace_period` of the proposal you want to `veto`, I'd say the only thing that makes sense is 72200, although that will cause issues as well...

dobertRowneySr:
~~I will fix the `lock.yml` as soon all the rest is approved~~ `lock.yml` fixed after rebase

runtime-modules/proposals/codex/src/lib.rs
dobertRowneySr: fixed in 57206bdd4604b15286909acb641129139e5f10bd

mnaamani:
You might want to merge https://github.com/Joystream/joystream/pull/3395 into rhodes, then update this PR to show only relevant changes.

bedeho:
Closing because these improvements are not relevant to new bounty enhancements.

PR #3189, opened 2022-02-08
DevOps - Olympia Deploy playground parameter updates

### LINKED ISSUE #3187: DevOps - Olympia Deploy playground updates 2
Some updates required to deploy playground workflow:

* The action should handle the flow being manually cancelled and delete cf stack
* Allow user to specify custom JSON which we will then pass to `yarn build:node:docker` as `ALL_PROPOSALS_PARAMETERS_JSON` while building the image on the instance
* Change endpoint from `{{ nip_domain }}/config.json` -> `{{ nip_domain }}/network/config.json`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201785009927821) by [Unito](https://www.unito.io)

ahhda:
Resolves #3187 

Successful run: https://github.com/ahhda/joystream/actions/runs/1813177662

### DIFF
--- a/devops/aws/deploy-playground.sh
+++ b/devops/aws/deploy-playground.sh
@@ -42,5 +42,6 @@ if [ $? -eq 0 ]; then
 
   echo -e "\n\n=========== Configuring node ==========="
   ansible-playbook -i $SERVER_IP, --private-key $KEY_PATH deploy-playground-playbook.yml \
-    --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO skip_chain_setup=$SKIP_CHAIN_SETUP"
+    --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO skip_chain_setup=$SKIP_CHAIN_SETUP
+                  stack_name=$SINGLE_NODE_STACK_NAME proposal_parameters=$ALL_PROPOSALS_PARAMETERS_JSON"
 fi
--- a/devops/aws/deploy-single-node.sample.cfg
+++ b/devops/aws/deploy-single-node.sample.cfg
@@ -21,3 +21,4 @@ CHAIN_SPEC_FILE="https://github.com/Joystream/joystream/releases/download/v9.14.
 GIT_REPO="https://github.com/Joystream/joystream.git"
 BRANCH_NAME="master"
 SKIP_CHAIN_SETUP="false"
+ALL_PROPOSALS_PARAMETERS_JSON=""
--- a/scripts/runtime-code-shasum.sh
+++ b/scripts/runtime-code-shasum.sh
@@ -15,12 +15,17 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
 fi
 
 export TEST_NODE_BLOCKTIME=1000
-export TEST_PROPOSALS_PARAMETERS_PATH="./tests/integration-tests/proposal-parameters.json"
 
-if [[ -z "$ALL_PROPOSALS_PARAMETERS_JSON" ]]; then
+if [[ -n "$ALL_PROPOSALS_PARAMETERS_JSON" ]] && [[ -n "$TEST_NODE" ]]; then
+  echo "Do not set both TEST_NODE and ALL_PROPOSALS_PARAMETERS_JSON env variables"
+  exit 1
+elif [[ -z "$ALL_PROPOSALS_PARAMETERS_JSON" ]] && [[ -z "$TEST_NODE" ]]; then
   PROPOSALS_PARAMETERS_FILE=""
+elif [[ -n "$TEST_NODE" ]]; then
+  PROPOSALS_PARAMETERS_FILE="./tests/integration-tests/proposal-parameters.json"
 else
-  PROPOSALS_PARAMETERS_FILE="./tmp.proposal_parameters.json"
+  mkdir -p runtime-inputs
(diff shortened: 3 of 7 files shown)


scripts/runtime-code-shasum.sh
```diff
@@ -17,6 +17,13 @@ fi
 export TEST_NODE_BLOCKTIME=1000
 export TEST_PROPOSALS_PARAMETERS_PATH="./tests/integration-tests/proposal-parameters.json"
 
+if [[ -z "$ALL_PROPOSALS_PARAMETERS_JSON" ]]; then
+  PROPOSALS_PARAMETERS_FILE=""
+else
+  PROPOSALS_PARAMETERS_FILE="./tmp.proposal_parameters.json"
```
ahhda: Saving the file as `tmp.proposal_parameters.json` on the target machine so that it doesn't interfere with git. Can be changed if required.

scripts/runtime-code-shasum.sh
```diff
@@ -28,6 +35,7 @@ ${TAR} -c --sort=name --owner=root:0 --group=root:0 --mode 644 --mtime='UTC 2020
     joystream-node-armv7.Dockerfile \
     node \
     $(test -n "$TEST_NODE" && echo "$TEST_PROPOSALS_PARAMETERS_PATH") \
+    $PROPOSALS_PARAMETERS_FILE \
```
mnaamani, line 42: Its possible to to have both `TEST_NODE` and `ALL_PROPOSALS_PARAMETERS_JSON` set and the hash will cover two files. So we need to fix this to only include one file.

Question: Given that joystream-node.Dockerfile will override the `ALL_PROPOSALS_PARAMETERS_JSON` variable if TEST_NODE is set with value from `tests/integration-tests/proposal-parameters.json`, we should either replicate this behavior in this script:

So lets have a single PATH to a .json file here, path being:
1 - empty "" string if both ALL_PROPOSALS_PARAMETERS_JSON and TEST_NODE are not set
2 - path to the test json `./tests/integration-tests/proposal-parameters.json` if TEST_NODE is set
3 - path to the file where you have written the values from env variable 

.. or modify joystream-node.Dockerfile to honor proposal parameters set in the environment.

Wondering wethere the built docker image should also store the proposals parameters.json file which ends up being used for the build as well.

scripts/runtime-code-shasum.sh
mnaamani: I see what you mean `tmp.*` is already in the .gitignore file..
But I think we need a more descriptive name than tmp.proposal_parameters.json

Maybe we can have a folder that is `.gitignore`d which will hold such artifacts

devops/aws/deploy-single-node.sample.cfg
```diff
@@ -20,4 +20,5 @@ CHAIN_SPEC_FILE="https://github.com/Joystream/joystream/releases/download/v9.14.
 # For deploy playground playbook only
 GIT_REPO="https://github.com/Joystream/joystream.git"
 BRANCH_NAME="master"
-SKIP_CHAIN_SETUP="false"
\ No newline at end of file
+SKIP_CHAIN_SETUP="false"
+ALL_PROPOSALS_PARAMETERS_JSON=""
```
mnaamani, line 24: is there a difference between setting this to an empty string explicitly and not setting it at all in the bash script test `if [[ -z "$ALL_PROPOSALS_PARAMETERS_JSON" ]];` ?

mnaamani (changes requested):

devops/aws/deploy-single-node.sample.cfg
ahhda: Kept this as an empty string to let the user know that they have the option to set this env variable, as they will refer the sample config while running the script. This doesn't affect the bash script. Let me know if it's not needed and I'll remove it.

scripts/runtime-code-shasum.sh
ahhda: > Wondering wethere the built docker image should also store the proposals parameters.json file which ends up being used for the build as well.
Since we are storing ALL_PROPOSALS_PARAMETERS_JSON as a file, it will be picked up by the command `COPY . /joystream` of our Dockerfile

scripts/runtime-code-shasum.sh
ahhda: Added a `runtime-inputs` dir as `gitignore` directory and the JSON file is stored here now.

devops/aws/deploy-playground.sh
```diff
@@ -42,5 +42,6 @@ if [ $? -eq 0 ]; then
 
   echo -e "\n\n=========== Configuring node ==========="
   ansible-playbook -i $SERVER_IP, --private-key $KEY_PATH deploy-playground-playbook.yml \
-    --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO skip_chain_setup=$SKIP_CHAIN_SETUP"
+    --extra-vars "branch_name=$BRANCH_NAME git_repo=$GIT_REPO skip_chain_setup=$SKIP_CHAIN_SETUP
+                  stack_name=$SINGLE_NODE_STACK_NAME proposal_parameters=$ALL_PROPOSALS_PARAMETERS_JSON"
```
mnaamani, line 46: Do we need some quotes around "$ALL_PROPOSALS_PARAMETERS_JSON"

ahhda: Not needed as the user will have to set the env variable value with quotes

Screenshot 2022-02-14 at 2 06 47 PM

PR #3191, opened 2022-02-08
Olympia. Runtime. Fix staking bug in the staking-handler pallet.

### LINKED ISSUE #3161: StakingHandler: is_enough_balance_for_stake check is not taking reusable stakes into account
**Issue:**
[`is_enough_balance_for_stake`](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/staking-handler/src/lib.rs#L174) only checks if `usable_balance` of an account is `>= amount`, while `usable_balance` excludes any tokens that are already locked. This is not correct, because we reuse locked tokens for multiple staking purposes, for example [StakingCandidateLock should be compatible with any working group lock](https://github.com/Joystream/joystream/blob/olympia/runtime/src/constants.rs#L148).

**Example 1:**
Let's say we setup a staking account with:

```
lockedBalance: 200
availableBalance: 1800
totalBalance: 2000
```

200 JOY is locked with `StakingCandidateLock`.
If we want to apply to an opening which requires a stake of `2000 JOY`, we'll get `InsufficientBalanceToCoverStake`.

**Example 2:**
Let's take another example and setup a staking account with:

```
lockedBalance: 200
availableBalance: 2000
totalBalance: 2200
```

In this case also 200 JOY is locked with `StakingCandidateLock`.
We can now succesfully apply to an opening which requires a stake of `2000 JOY`, but after applying we still have `200 JOY` available:

```
# Balances after applying:
lockedBalance: 2000
availableBalance: 200
totalBalance: 2200
```

**How to reproduce:**

1. Use https://github.com/Lezek123/substrate-runtime-joystream/tree/olympia-initialize-lead-console-logs branch for improved `console.log` outputs
1. `yarn && yarn workspace @joystream/types build`
1. Start the node
1. Execute `GROUP=distributionWorkingGroup yarn workspace api-scripts initialize-lead`
1. The output should be as follows:

```
Creating distributionWorkingGroup lead opening...
Setting up staking account...
Staking account balances before apply:  { available: 2000, locked: 200, total: 2200 }
Applying to distributionWorkingGroup lead opening...
Staking account balances after apply:  { available: 200, locked: 2000, total: 2200 }
```

1. Now reduce the initial balance of the staking account in the script by chaning [`STAKING_ACCOUNT_CANDIDATE_STAKE`](https://github.com/Lezek123/substrate-runtime-joystream/blob/olympia-initialize-lead-console-logs/utils/api-scripts/src/initialize-lead.ts#L24) const to `new BN(0)`
1. Re-run the script using different working group, ie.: `GROUP=storageWorkingGroup yarn workspace api-scripts initialize-lead`
1. The output should be as follows:

```
Creating storageWorkingGroup lead opening...
Setting up staking account...
Staking account balances before apply:  { available: 1800, locked: 200, total: 2000 }
Applying to storageWorkingGroup lead opening...
Error: Failed to apply on lead opening! - Extrinsic execution error: InsufficientBalanceToCoverStake ( Insufficient balance to cover stake.)
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201758363490308) by [Unito](https://www.unito.io)

shamil-gadelshin:
The PR changes `usable_balance` check to `free_balance` check on creating locks in the staking handler crate.

Fixes https://github.com/Joystream/joystream/issues/3161

### DIFF
--- a/runtime-modules/staking-handler/src/lib.rs
+++ b/runtime-modules/staking-handler/src/lib.rs
@@ -7,6 +7,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use frame_support::dispatch::{DispatchError, DispatchResult};
+use frame_support::ensure;
 use frame_support::traits::{Currency, Get, LockIdentifier, LockableCurrency, WithdrawReasons};
 use sp_arithmetic::traits::Zero;
 use sp_std::marker::PhantomData;
@@ -152,11 +153,10 @@ impl<
             return Ok(());
         }
 
-        let free_balance = <pallet_balances::Module<T>>::free_balance(account_id);
-
-        if new_stake > free_balance {
-            return Err(DispatchError::Other("Not enough balance for a new stake."));
-        }
+        ensure!(
+            Self::is_enough_balance_for_stake(account_id, new_stake),
+            DispatchError::Other("Not enough balance for a new stake.")
+        );
 
         Self::lock(account_id, new_stake);
 
--- a/runtime-modules/staking-handler/src/lib.rs
+++ b/runtime-modules/staking-handler/src/lib.rs
@@ -152,9 +152,9 @@ impl<
             return Ok(());
         }
 
-        let usable_balance = <pallet_balances::Module<T>>::usable_balance(account_id);
+        let free_balance = <pallet_balances::Module<T>>::free_balance(account_id);
 
-        if new_stake > current_stake + usable_balance {
+        if new_stake > free_balance {
             return Err(DispatchError::Other("Not enough balance for a new stake."));
         }
 
@@ -175,7 +175,7 @@ impl<
         account_id: &<T as frame_system::Trait>::AccountId,
         amount: <T as pallet_balances::Trait>::Balance,
     ) -> bool {
-        <pallet_balances::Module<T>>::usable_balance(account_id) >= amount
+        <pallet_balances::Module<T>>::free_balance(account_id) >= amount
     }
 
     fn lock_id() -> LockIdentifier {
--- a/runtime-modules/working-group/src/tests/mod.rs
+++ b/runtime-modules/working-group/src/tests/mod.rs
@@ -119,13 +119,13 @@ fn add_opening_fails_with_insufficient_balance() {
             .hire_lead();
 
         let add_opening_fixture = AddOpeningFixture::default().with_stake_policy(StakePolicy {
-            stake_amount: <Test as Trait>::MinimumApplicationStake::get() + 1,
+            stake_amount: <Test as Trait>::MinimumApplicationStake::get(),
             leaving_unstaking_period: <Test as Trait>::MinUnstakingPeriodLimit::get() + 1,
         });
 
-        add_opening_fixture.call_and_assert(Err(DispatchError::Other(
-            "Not enough balance for a new stake.",
-        )));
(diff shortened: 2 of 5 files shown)


dobertRowneySr (approved):
LGTM

runtime-modules/staking-handler/src/lib.rs
```diff
@@ -175,7 +175,7 @@ impl<
         account_id: &<T as frame_system::Trait>::AccountId,
         amount: <T as pallet_balances::Trait>::Balance,
     ) -> bool {
-        <pallet_balances::Module<T>>::usable_balance(account_id) >= amount
+        <pallet_balances::Module<T>>::free_balance(account_id) >= amount
```
Lezek123, line 178: This breaks `ensure_stake_for_opening_type`, which also calls `is_enough_balance_for_stake` but actually needs to check `usable_balance`, because during `add_opening` lead's stake is **increased** with `set_stake`.

runtime-modules/working-group/src/tests/mod.rs
```diff
@@ -115,14 +115,17 @@ fn add_opening_fails_with_zero_reward() {
...
+            stake_amount: <Test as Trait>::MinimumApplicationStake::get() + 1,
+            leaving_unstaking_period: <Test as Trait>::MinUnstakingPeriodLimit::get() + 1,
+        });
 
-        add_opening_fixture.call_and_assert(Err(
-            Error::<Test, DefaultInstance>::InsufficientBalanceToCoverStake.into(),
-        ));
+        add_opening_fixture.call_and_assert(Err(DispatchError::Other(
+            "Not enough balance for a new stake.",
+        )));
```
Lezek123: I see that the situation I described above is treated as expected test result here, but I think it shouldn't be for 2 reasons:
- `set_stake` is under the `MUTATION SAFE` mark, so we shouldn't expect it to ever throw
- `ensure_stake_for_opening_type` becomes useless at this point

Lezek123 (changes requested):

runtime-modules/working-group/src/tests/mod.rs
shamil-gadelshin: The `set_stake` method and changing the usable balance to free balance check are separate concerns. The previous `usable_balance` check was a bug. Even if the `set_stake` has issues we should fix them as well. But I don't see how it is the case. The `set_stake` method call is the first mutation method after the `mutation safe` mark on each occurrence and it returns the `DispatchResult` so it's potentially unsafe. But its first position (after the mark) grants us safety guarantees because if it fails the state remains unchanged. As to `ensure_stake_for_opening_type` it contains other checks in addition to balance verification but even if it would contain only the balance check for a stake - it's easier to reason about the `mutation safe` mark having all the checks set prior to mutation (even with embedded checks).

runtime-modules/staking-handler/src/lib.rs
Lezek123: My suggestion was incorrect, `ensure_stake_for_opening_type` should actually check if `free_balance` >= `current_lead_stake + opening_stake`, as there may be a higher stake than `current_lead_stake` that could be reusable (in which case `usable_balance` check would be invalid).

That also will be inline with the check inside `set_stake`.

Lezek123 (approved):
LGTM!

PR #2930, opened 2021-12-15
Content NFT schema and mappings

ondratra:
Schema and mappings for content NFTs.

Originates from PR https://github.com/Joystream/joystream/pull/2666 , fixes compilations problems from that PR and aims to reduce the number of commits included (to make PR reviewable). 

List of changes not related to query node made:
```
- merge `giza` branch (on 14.12.2021) - fixes compilation error(s) and adds latest changes to content directory, storage, etc.
- add `use sp_std::borrow::ToOwned;` to `runtime-modules/content/src/types.rs` - fixes compilation error
- `cargo update -p environmental` due to `feature(const_fn)) - feature has been removed` error - fixes compilation error
- fill empty values in tests for some newly added properties (e.g. `collaborators`)

- merge `giza_staging` branch (15.12.2021) - merges latest giza staged changes in an attempt to fix some content directory test errors (not fixed by the merge - see the line below)
- fix content directory `migration_test()` test by changing actor to `lead` and censor sign to `true` in for calls `update_channel_censorship_status` and `update_video_censorship_status` inside of `assert_video_and_channel_existrinsics_with` function
```

PR #3195, opened 2022-02-09
Olympia vnft schema mappings

ondratra:
QN mappings for video NFT feature.

Supersedes https://github.com/Joystream/joystream/pull/2930 . Targets Olympia.


ondratra:
Current status:
- NFT features are working on QN with Olympia with some limitations
  - issuing NFT during video creation is not working yet
  - setting NFT transactional status during video creation or NFT issuance is not working yet
- buy-now feature is now NOT caught by processor and thus doesn't work (reason still unknown)
- history of NFT buy/sell offers (including those canceled) is not done yet
- there are still some extra things in the runtime-module that are likely artifacts from previous branch merging. I think it should not affect Atlas now, and it will be resolved soon
- NFT metadata are not properly tested in tests due to some problems with their encoding (only empty metadata works there). This might or might not be an issue for Atlas.

ondratra:
Remaining unresolved issues:
- NFT metadata are not properly tested in tests due to some problems with their encoding (only empty metadata works there). This might or might not be an issue for Atlas. (https://github.com/Joystream/joystream/pull/3195/files#diff-64a6b3108d189ec8ae3b896171805c387b702ca3bf7faec1663dd57ff84b26ebR865 - this is what I tried so far)
- polish tests - there are some TODOs about reading min/max boundaries from runtime + add more comments where needed

Issue #3149, opened 2022-02-01
Simplify lock combinatino representation ALLOWED_LOCK_COMBINATIONS


bedeho:
# Background

Currently, `ALLOWED_LOCK_COMBINATIONS`, which is supposed to be the source of truth for account locking attempts, has representation `BTreeSet<(LockIdentifier, LockIdentifier)>`. This means assymetric locking relationships are in principle allowed, namely that lock x can be applied when lock y is already present, but not vice versa. This is however not how the locking rules actually work in Joystream, and so this is too expressive, and opens the door for sloppy initilization, which indeed appears to already be present.

https://github.com/Joystream/joystream/blob/5920eaeedf19aaf868970657c1536b17f9a8eeb3/runtime/src/constants.rs#L126

# Proposal

Come up with a succinct and restrictive representation of locking rules that mirrors what we actually want to enforce in Joystream.

shamil-gadelshin:
The [current implementation](https://joystream.gitbook.io/joystream-handbook/key-concepts/staking) of the staking mechanism seems to have only two types (or classes) of compatible lock combinations: `all` and `regular`(for 0, 10, *b"vesting " IDs) with a single exception for the invitation locks. 

It makes sense to take advantage of this fact and move away from the table structure of the lock combinations (up to N^2 combinations) to a much simpler class-based representation.

bedeho:
I updated the handbook to reflect a very similar model, except invitation locks are just treated as `all`, more or less.

https://joystream.gitbook.io/joystream-handbook/key-concepts/staking#locks-1

So one way to incorporate this would be to have

```Rust

enum Lock {
Vote,
Invitation,
...
}
pub static ref RIVALROUS_LOCKS BTreeSet<Lock>

//alternatively just have 

pub fn is_rivalrous(Lock) -> Boolean {
  // match on type and just return... right value
}
```

and now checking whether there are any locks conflicting with adding some lock `x` to an account `A` you just check
- make sure same lock is not already appliead
- make sure that if `x` is rivalrous, that all existing locks on `A` are not.

its not clear anything really needs to be exported here.

What do you think?

Issue #2825, opened 2021-11-10
Expose compatible Locks API


jodator:
In Pioneer 2.0 we need to check which locks are compatible. ATM we copy the logic of [`ALLOWED_LOCKS_COMBINATION`](https://github.com/Joystream/joystream/blob/db3885858a7812377a19390968bdbf65221f0270/runtime/src/constants.rs#L126-L142).

This might lead to errors and maintenance overhead when new locks are introduced on the runtime.

I think that this could be exposed as `api.consts.staking.allowedLocksCombination(lock: LockId)` so any front-end client could easily check allowed locks combinations.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958628539932) by [Unito](https://www.unito.io)

PR #3201, opened 2022-02-10
Runtime. Olympia change compatible locks combination.

shamil-gadelshin:
We need to update compatible lock ID combinations according to the current [handbook version](https://github.com/Joystream/handbook/blob/master/key-concepts/staking.md).

#### Affected issues
- fixes https://github.com/Joystream/joystream/issues/3149.
- reverts https://github.com/Joystream/joystream/issues/2825

#### Changes
- removed 'allowed lock combinations' constant from the utility pallet
- simplified lock ID representation
- updated affected runtime tests

### DIFF
--- a/runtime-modules/utility/src/lib.rs
+++ b/runtime-modules/utility/src/lib.rs
@@ -24,16 +24,15 @@ pub(crate) mod tests;
 
 mod benchmarking;
 
-use common::{working_group::WorkingGroup, AllowedLockCombinationProvider, BalanceKind};
+use common::{working_group::WorkingGroup, BalanceKind};
 use council::Module as Council;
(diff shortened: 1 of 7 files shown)


Lezek123:
Should we now expose `NON_RIVALROUS_LOCKS` const in case it is changed in the future? (since as I understand, all the frontend apps would need to have the same const hardcoded now)

dobertRowneySr (approved):
LGTM

shamil-gadelshin:
> Should we now expose NON_RIVALROUS_LOCKS const in case it is changed in the future? (since as I understand, all the frontend apps would need to have the same const hardcoded now)

@bedeho suggested skipping this task (this is why I reverted the const exposure). 

We could try to introduce it in a separate task if we need it after all.

PR #3204, opened 2022-02-11
Integration tests: Separate council and proposals tests

Lezek123:
Council and proposals tests should not be executed in parallel, since proposals tests make some assumptions about council and also re-elect council for the purpose of testing proposals with `constitutionality` > 1


ondratra (approved):
LGTM

PR #3202, opened 2022-02-11
increase voting and revealing periods and longer council idle time

mnaamani:
We are using adjusted periods for testing in playground. But with these values, council mappings seems to break:

```
ERROR: Stopping the proccessor due to errors: {} name: Error, message: No stage update found., stack: Error: No stage update found.
    at getCurrentStageUpdate (/joystream/query-node/mappings/lib/src/council.js:72:15)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async council_NewCandidate (/joystream/query-node/mappings/lib/src/council.js:244:29)
    at async MappingsLookupService.call (/joystream/node_modules/@joystream/hydra-processor/lib/executor/MappingsLookupService.js:80:9)
    at async /joystream/node_modules/@joystream/hydra-processor/lib/executor/TransactionalExecutor.js:40:17
ERROR: {} name: Error, message: Error: No stage update found., stack: Error: Error: No stage update found.
    at MappingsProcessor.processingLoop (/joystream/node_modules/@joystream/hydra-processor/lib/process/MappingsProcessor.js:75:23)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Promise.all (index 1)
    at async MappingsProcessor.start (/joystream/node_modules/@joystream/hydra-processor/lib/process/MappingsProcessor.js:25:9)
    at async ProcessorRunner.process (/joystream/node_modules/@joystream/hydra-processor/lib/start/ProcessorRunner.js:47:9)
    at async Run.run (/joystream/node_modules/@joystream/hydra-processor/lib/commands/run.js:27:13)
    at async Run._run (/joystream/node_modules/@oclif/command/lib/command.js:43:20)
    at async Config.runCommand (/joystream/node_modules/@oclif/config/lib/config.js:173:24)
    at async Main.run (/joystream/node_modules/@oclif/command/lib/main.js:27:9)
    at async Main._run (/joystream/node_modules/@oclif/command/lib/command.js:43:20)
INFO: Shutting down...
INFO: Closing the database connection...
error Command failed with exit code 1.
```

The chain is started up with setupNewChain scenario which is electing the council.

Is this only an issue in the mappings, or potentially indication of a bug in the council runtime module?

To replicate the issue, simply build joystream-node and start local playground with `yarn start`, then look at the processor logs with `docker logs processor`

PR #3229, opened 2022-02-14
Council mappings: Bootstrapt initial CouncilStageUpdate

Lezek123:
The council mappings rely on `CouncilStageUpdate` entity always beeing present which was discovered in https://github.com/Joystream/joystream/pull/3202 due to prolonged initial `Announcing` period.

To fix this I added initial `CouncilStageUpdate` bootstrapping to `bootstrap.ts`.


ondratra (approved):
LGTM!

PR #3253, opened 2022-02-17
Audit3 cleaner storage & content integration

### LINKED ISSUE #3082: Tracking Issue: content <-> storage integration cleanup
* [x] https://github.com/Joystream/joystream/issues/3071
* [x] https://github.com/Joystream/joystream/issues/3075
* [x] https://github.com/Joystream/joystream/issues/3076
* [x] https://github.com/Joystream/joystream/issues/3077
* [x] https://github.com/Joystream/joystream/issues/3078
* [x] https://github.com/Joystream/joystream/issues/3079
* [x] https://github.com/Joystream/joystream/issues/3083
* [x] https://github.com/Joystream/joystream/issues/3081

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958631338934) by [Unito](https://www.unito.io)

dobertRowneySr:
attempt to fix #3082 

# Changes 
- added function `upload_and_delete` data objects in order improve hygiene in `update_channel` & `update_video` extr. in the `content` pallet.
- (dynamic) bag deletion now works also with non empty bags (deleting all stored objects along with the bag)
- (dynamic) bag creation now performs upload if objects are specified in `DynBagCreation` record.

Due to the increasingly common operations in bag creation / objects upload / objects deletion / bag deletion I create a utility function 
`try_mutate_storage_state` that performs most of the work for the above methods, in order to avoid code duplication. The con of this approach is that such function turns out to be somewhat complex.


dobertRowneySr:
Runtime is ready to be reviewed. I will deal with `@joystream/types` errors later

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -14,7 +14,7 @@ use crate::{Call, Module, StorageBucketById, Trait};
 benchmarks! {
     where_clause {
         where T: balances::Trait,
-              T: membership::Trait,
+//              T: membership::Trait,
```
shamil-gadelshin: Please, remove commented code.

runtime-modules/referendum/src/lib.rs
```diff
@@ -285,7 +285,7 @@ decl_storage! {
     trait Store for Module<T: Trait<I>, I: Instance = DefaultInstance> as Referendum {
         /// Current referendum stage.
         pub Stage get(fn stage) config():
-            ReferendumStage<T::BlockNumber, T::MemberId, T::VotePower>;
+        ReferendumStage<T::BlockNumber, T::MemberId, T::VotePower>;
```
shamil-gadelshin, line 288: Why is this changed?

runtime-modules/content/src/lib.rs
```diff
@@ -807,42 +723,22 @@ decl_module! {
...
-            }
-
-            if !params.assets_to_remove.is_empty() {
-                Storage::<T>::delete_data_objects(
-                    sender,
-                    Self::bag_id_for_channel(&channel_id),
-                    params.assets_to_remove.clone(),
+            // upload/delete video assets from storage with commit or rollback semantics
+            Storage::<T>::upload_and_delete_data_objects(
+                UploadParameters::<T> {
```
shamil-gadelshin: Hard to read. Please, move the creation of the parameters in a separate variable and format the code.

runtime-modules/content/src/lib.rs
```diff
@@ -508,44 +474,21 @@ decl_module! {
...
-
-            if !params.assets_to_remove.is_empty() {
-                Storage::<T>::delete_data_objects(
-                    sender,
-                    Self::bag_id_for_channel(&channel_id),
-                    params.assets_to_remove.clone()
-                )?;
-            }
+            Storage::<T>::upload_and_delete_data_objects(
+                UploadParameters::<T> {
```
shamil-gadelshin: Hard to read. Please, move the creation of the parameters in a separate variable and format the code.

runtime-modules/storage/src/lib.rs
```diff
@@ -568,8 +531,50 @@ impl<StorageBucketId: Ord, DistributionBucketId: Ord, Balance>
...
+            deletion_prize: Some(deletion_prize),
+            ..self
+        }
+    }
 }
 
+// Helper enum for performing bag operation: no default since it is non trivial
+#[derive(Clone, PartialEq, Eq, Debug)]
+enum BagOperationParamsTypes<Balance: Unsigned, ObjectId: Clone> {
+    Update(Vec<DataObjectCreationParameters>, BTreeSet<ObjectId>),
```
shamil-gadelshin: Please, comment the options and parameters. Consider switching to the named parameters for enum options.

runtime-modules/storage/src/lib.rs
```diff
@@ -737,6 +745,18 @@ pub struct Voucher {
     pub objects_used: u64,
 }
 
+impl Voucher {
+    fn update(self, new_voucher: VoucherUpdate) -> Result<Self, &'static str> {
+        ensure!(new_voucher.objects_number <= self.objects_limit, "number");
```
shamil-gadelshin: Please, remove "magic consts" and consider placing `ensure` checks in the calling code or as a separate function.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+
+    fn compute_net_prize(
+        init_value: NetDeletionPrize<T>,
+        num_obj_to_create: usize,
+        num_obj_to_delete: usize,
+    ) -> NetDeletionPrize<T> {
+        let amnt = T::DataObjectDeletionPrize::get();
+        init_value
+            .add_balance(
+                iter::repeat(amnt)
```
shamil-gadelshin: Can't we multiply here?

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+                    .take(num_obj_to_create)
+                    .fold(BalanceOf::<T>::zero(), |acc, x| acc.saturating_add(x)),
+            )
+            .sub_balance(
+                iter::repeat(amnt)
```
shamil-gadelshin: Can't we multiply here?

runtime-modules/storage/src/lib.rs
```diff
@@ -991,28 +1079,28 @@ decl_storage! {
 
         /// DynamicBagCreationPolicy by bag type storage map.
         pub DynamicBagCreationPolicies get (fn dynamic_bag_creation_policy):
-            map hasher(blake2_128_concat) DynamicBagType =>
+        map hasher(blake2_128_concat) DynamicBagType =>
```
shamil-gadelshin: Please, fix the format.

runtime-modules/storage/src/lib.rs
```diff
@@ -991,28 +1079,28 @@ decl_storage! {
...
             DynamicBagCreationPolicy<T::DistributionBucketFamilyId>;
 
         /// 'Data objects for bags' storage double map.
         pub DataObjectsById get (fn data_object_by_id): double_map
             hasher(blake2_128_concat) BagId<T>,
-            hasher(blake2_128_concat) T::DataObjectId => DataObject<BalanceOf<T>>;
+        hasher(blake2_128_concat) T::DataObjectId => DataObject<BalanceOf<T>>;
```
shamil-gadelshin: Please, fix the format.

runtime-modules/storage/src/lib.rs
```diff
@@ -991,28 +1079,28 @@ decl_storage! {
...
         /// Distribution bucket families.
         pub DistributionBucketFamilyById get (fn distribution_bucket_family_by_id):
-            map hasher(blake2_128_concat) T::DistributionBucketFamilyId =>
+        map hasher(blake2_128_concat) T::DistributionBucketFamilyId =>
             DistributionBucketFamily<T>;
 
         /// 'Distribution bucket' storage double map.
         pub DistributionBucketByFamilyIdById get (fn distribution_bucket_by_family_id_by_index):
-            double_map
+        double_map
```
shamil-gadelshin: Please, fix the format.

runtime-modules/storage/src/lib.rs
```diff
@@ -1024,19 +1112,19 @@ decl_storage! {
...
-        <T as Trait>::DataObjectId,
-        UploadParameters = UploadParameters<T>,
-        BagId = BagId<T>,
-        DynamicBagId = DynamicBagId<T>,
-        <T as frame_system::Trait>::AccountId,
-        Balance = BalanceOf<T>,
-        <T as Trait>::DistributionBucketFamilyId,
-        DistributionBucketId = DistributionBucketId<T>,
-        <T as Trait>::DistributionBucketIndex,
+    WorkerId = WorkerId<T>,
```
shamil-gadelshin, line 1240: Please, fix the format.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
             distribution_bucket_index,
         }
     }
+
+    fn generate_new_storage_buckets(
```
shamil-gadelshin: The name is confusing - we don't create new buckets here.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+                    r.map(|(id, mut bk)| {
+                        bk.unregister_bag_assignment();
+                        (id, bk)
+                    })
+                })
+                .collect(),
+        }
+    }
+
+    fn generate_new_distribution_buckets(
```
shamil-gadelshin: The name is confusing - we don't create new buckets here.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+                    .ok()
+            })
+            // attempt to update bucket voucher
+            .map(|(id, bk)| {
+                bk.voucher
+                    .clone()
+                    .update(new_voucher_update)
+                    .map_or_else(
+                        |e| match e {
+                            "size" => Err(Error::<T>::StorageBucketObjectSizeLimitReached),
```
shamil-gadelshin: Please, refactor "magic consts" in favor of the separate function(s).

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+        new_distribution_buckets.iter().for_each(|(id, bucket)| {
+            DistributionBucketByFamilyIdById::<T>::insert(
+                &id.distribution_bucket_family_id,
+                &id.distribution_bucket_index,
+                bucket,
+            )
+        });
+
+        // FIRST refund or request deletion prize first: no-op if amnt = 0
+        match net_prize {
```
shamil-gadelshin: These calls should be made first after the mutation safe mark. Because they are potentially fallible.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+                Ok(id.clone())
+            } else {
+                Err(Error::<T>::CannotCreateStaticBag)
+            }?;
+
+            Self::ensure_bag_exists(&bag_op.bag_id).map_or_else(
+                |_| {
+                    Ok(Bag::<T>::default()
+                        .with_prize(deletion_prize)
+                        .with_storage_buckets(StorageBucketPicker::<T>::pick_storage_buckets(
```
shamil-gadelshin, line 3823: Why did we remove the bucket selection with the correct object size and number? 

`StorageBucketPicker::<T>::pick_storage_buckets(bag_type, voucher_update)`

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3602,342 @@ impl<T: Trait> Module<T> {
...
+                let obj_id = NextDataObjectId::<T>::get();
+                DataObjectsById::<T>::insert(&bag_op.bag_id, obj_id, obj);
+                NextDataObjectId::<T>::put(obj_id.saturating_add(One::one()));
+            })
+        };
+
+        // mutate bags set
+        if bag_op.params.is_delete() {
+            // remove bag for deletion
+            Bags::<T>::remove(&bag_op.bag_id)
```
shamil-gadelshin: Should we have an event here?

shamil-gadelshin (changes requested):
Huge work - the code became better!

Issues found:
- lost formatting 
- why did we rollback bucket selection with proper object size and number?
- there are a lot of long matches within the big functions. Consider encapsulating those matches in the `impl` sections for the enums `NetDeletionPrize` and `BagOperationParams` in the places that will profit from the refactoring (similar to `construct-*` functions.

bedeho:
I have two questions and a request:

# Question 1

The `try_mutating_storage_state` is 1000x too complex, too many activities ( `BagOperationParamsTypes` are getting multiplexed in here without a clear reason as far as I can see

https://github.com/Joystream/joystream/blob/63dcc503cc866e8ba7459da8cf9e8e2fe634f3df/runtime-modules/storage/src/lib.rs#L3740

# Question 2

Why are we returning `DispatchResult` from the public methods? These are not dispatchables, they are just methods. The cost of returning this result type is that we are forced to return the same set of generic `DispatchErrors`. Instead we would like to return defined error type for each method, which then self-documents and clearly shows exactly what can and cannot go wrong in each individual call. 

# Request

The public methods

```Rust

//! - upload_data_objects
//! - can_move_data_objects
//! - move_data_objects
//! - delete_data_objects
//! - delete_dynamic_bag
//! - create_dynamic_bag
//! - upload_and_delete_data_objects
```

Are badly under documented. This is in part due to error types above, but also what each requires to run, and what it will do. For example, its not clear whether`delete_dynamic_bag` works if its non-empty.

shamil-gadelshin:
>Why are we returning DispatchResult from the public methods? These are not dispatchables, they are just methods. The cost of returning this result type is that we are forced to return the same set of generic DispatchErrors. Instead we would like to return defined error type for each method, which then self-documents and clearly shows exactly what can and cannot go wrong in each individual call.

This type is very convenient.
`pub type DispatchResult = Result<(), sp_runtime::DispatchError>;`

It contains errors that are easily convertible between pallets.

The alternative would be to create a separate set of errors for the public methods, conversion code for the storage pallet between local errors and new dedicated set, and new errors and conversion code for each client pallet. We tested this approach with old (currently deleted) ContentWorkingGroup module. The overhead for these errors was huge: https://github.com/Joystream/joystream/blob/251840fee6c8f29792dec6b1e11dc0406132d176/runtime-modules/content-working-group/src/lib.rs#L863

bedeho:
> The alternative would be to create a separate set of errors for the public methods, conversion code for the storage pallet between local errors and new dedicated set, and new errors and conversion code for each client pallet. We tested this approach with old (currently deleted) ContentWorkingGroup module. The overhead for these errors was huge: 

This is only impractical in the sense that having strict types is always impractical. We are only trading superficial simplicity for opaque public APIs. The reality is that this mapping is simply i reflecting the reality of the situation, this code simply is long (but actually trivial in complexity, as a pure mapping) because the library has such a rich set of errors for a given call. That simply is making that library much easier to understand, document and use correctly. Imagine trying to some how communicate all of these possible nested failure modes in any other way to the reader? I don't, and its not like you can actually get away with someone not knowing.

shamil-gadelshin:
> This is only impractical in the sense that having strict types is always impractical. We are only trading superficial simplicity for opaque public APIs. ...

The current approach is also Substrate-native and could be integrated seamlessly with new pallets. What if we mitigate its downsides by enumerating all the possible errors in the code comment for each method? This way the code will be much cleaner and we will get an API description in the comment form (no type checking though). I would really prefer to avoid introducing the dedicated error type, another reason would be that a single change in the storage pallet could cause a cascade effect in the dependent pallets which is not necessary.

runtime-modules/referendum/src/lib.rs
dobertRowneySr: I guess it was unintended.

runtime-modules/storage/src/lib.rs
dobertRowneySr: Even when I already checked that Treasury/signer account have enough balance?

bedeho:
Conclusion about my prior comment: https://github.com/Joystream/joystream/pull/3253#issuecomment-1065024183

- Question 1: Revisit here https://github.com/Joystream/joystream/issues/3412.
- Question 2: Revisit here https://github.com/Joystream/joystream/issues/3413.
- Request: will be done in next iteration in this PR.

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/content/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/content/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/referendum/src/lib.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/benchmarking.rs
dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: Pattern used by the `try_mutating_storage` : 
1. Get initial bag state: return bag on storage if present else create default one by selecting **some** storage bucket
2. Construct creation & removal parameters 
3. create `voucher_update` with params @ (2)
4. Try update storage buckets with `voucher_update` @ (3) by checking that storage bucket object number and total size bounds are respected. If it successful return a `BTreeMap` storage bucket id x storage bucket else throw.
Analogous with distribution bucket.
5. check sender/ treasury balance funds are sufficient for the bag operation specified
6. Update chain storage & balance after the mutation safe mark

This is the common pattern of all `*-bag` extrinsic. But:
1. Is complex and the resulting function is long and complex
2. It is definitely not obvious
So this makes me question having introduced such complex artifact

dobertRowneySr: :+1:

runtime-modules/storage/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/storage/src/lib.rs
dobertRowneySr: `BalanceOf<T> * usize` multiplication does not work, so I didn't change the accounting method for the deletion prize (meaning even before my refactoring `iter::repeat + fold` was used instead of simple multiplication, and I believe for the same reason)

dobertRowneySr: `BalanceOf<T>` is defined on runtime + `usize` is architecture-dependent [see here](https://doc.rust-lang.org/nightly/std/primitive.usize.html). So even converting both operands into say `u64` multiply and then convert the result back to `BalanceOf<T>` is not feasible, since Balance does not implement `From<u64>`

runtime-modules/storage/src/lib.rs
dobertRowneySr: Same as previous comment
:+1:

runtime-modules/storage/src/lib.rs
ignazio-bovo: :+1:

dobertRowneySr:
I addressed the comments. I am updating the docs after that I will generate metadata & types

runtime-modules/storage/src/lib.rs
shamil-gadelshin: You can cast usize to Balance instead.

This code compiles:
```
        let temp: BalanceOf::<T> = num_obj_to_create.saturated_into();
        let result = temp * amnt;

Comment (shamil-gadelshin) [reply runtime-modules/storage/src/lib.rs]: You can cast usize to Balance instead.

This code compiles:
```...
        let temp: BalanceOf::<T> = num_obj_to_create.saturated_into();
        let result = temp * amnt;
...
```

runtime-modules/storage/src/lib.rs
shamil-gadelshin: https://github.com/Joystream/joystream/pull/3253#discussion_r828184114

runtime-modules/storage/src/lib.rs
shamil-gadelshin: This code has no dependency on the previous code and we can move it freely. However, if the check you mentioned would be changed or removed (by mistake) the first position of the snippet would make the method still safe in contrast with the current situation.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3651,335 @@ impl<T: Trait> Module<T> {
...
+                NextDataObjectId::<T>::put(obj_id.saturating_add(One::one()));
+            })
+        };
+
+        // mutate bags set
+        if bag_op.params.is_delete() {
+            // remove bag for deletion
+            Bags::<T>::remove(&bag_op.bag_id);
+            // unfallible: bag id / operation constraints already checked
+            let dyn_bag_id = bag_id_variant.ensure_is_dynamic_bag::<T>()?;
```
shamil-gadelshin: Please, move this conversion before the 'mutation safe' mark

shamil-gadelshin (changes requested):
There is a couple of minor issues left but the code is not compilable. Please, fix the build.

shamil-gadelshin (changes requested):
Not all issues were addressed and CI checks are still failing (benchmarking). Please, fix it.

runtime-modules/storage/src/lib.rs
```diff
@@ -3942,4 +3738,336 @@ impl<T: Trait> Module<T> {
...
+        let Bag::<T> {
+            stored_by,
+            distributed_by,
+            deletion_prize,
+            objects_total_size,
+            objects_number,
+        } = Self::retrieve_dynamic_bag(&bag_op)?;
+
+        let bag_id_variant = bag_op.ensure_bag_id_valid_for_op::<T>()?;
+        let dyn_bag_id = bag_id_variant.ensure_is_dynamic_bag::<T>().ok();
```
shamil-gadelshin: Why did you change the previous line with `?` ?
`let dyn_bag_id = bag_id_variant.ensure_is_dynamic_bag::<T>()?;`

You added a comment about `Some(..)` later which is avoidable.

shamil-gadelshin (changes requested):
Your last change for `dyn_bag_id` is suboptimal - there is a comment about it. Please, also have a look on comments from the previous reviews. You still didn't address them.

runtime-modules/storage/src/lib.rs
ignazio-bovo: I cannot address the comment: 
If I call `bag_id_variant.ensure_is_dynamic_bag` then it throws errors when you try to update assets on static bags

ignazio-bovo: This is really the best I can do

runtime-modules/storage/src/lib.rs
dobertRowneySr: :+1:  I moved above before the mutation barrier

runtime-modules/storage/src/lib.rs
dobertRowneySr: I tried a different solution since:
``` rust
         let bag_id_variant = bag_op.bag_id.ensure_is_dynamic_bag::<T>()?; 
```
Makes the function fail for static bags updates

runtime-modules/storage/src/lib.rs
shamil-gadelshin: This is incorrect - I meant "below" the 'mutation safe' mark.

dobertRowneySr: So 
``` rust
// == MUTATION SAFE ==
// ...
let dyn_bag_id = bag_id_variant.ensure_is_dynamic_bag::<T>()?; 
```
?

dobertRowneySr: At the end I realized that `DynamicBagDeleted` event is already emitted on `delete_dynamic_bag` after `try_mutating_storage` is successful. So there is no need for 
that in `try_mutating_storage` -> there is no need for `dyn_bag_id`

shamil-gadelshin (approved):
LGTM

PR #3258, opened 2022-02-17
CLIs - fix bash exit codes

### LINKED ISSUE #3251: CLI - exit codes numbers too high
Bash commands should always return exit codes between 0-255. (See https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html). Currently, Joystream CLI as well Distributor node CLI are returning values higher than 255 on error. The value gets modulated by 256 (`errorCode % 256`). That breaks comparisons between `ExitCode` defined in JS file and the actual number returned in bash.

Lower the exit codes numbers. 
See files `distributor-node/src/command-base/ExitCodes.ts` and `cli/src/ExitCodes.ts`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201840692098596) by [Unito](https://www.unito.io)

ondratra:
Fixes #3251

### DIFF
--- a/cli/src/ExitCodes.ts
+++ b/cli/src/ExitCodes.ts
@@ -1,18 +1,20 @@
 enum ExitCodes {
   OK = 0,
 
-  InvalidInput = 400,
-  FileNotFound = 401,
-  InvalidFile = 402,
-  NoAccountFound = 403,
-  NoAccountSelected = 404,
-  AccessDenied = 405,
+  InvalidInput = 40,
+  FileNotFound = 41,
+  InvalidFile = 42,
+  NoAccountFound = 43,
+  NoAccountSelected = 44,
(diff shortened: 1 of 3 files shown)


mnaamani:
Looks good, just wondering are there any tests or utilities in this repo or otherwise that were expecting the the original codes. I assume not since it would have been hard to compare since this is the whole point of this fix :)

ondratra:
> Looks good, just wondering are there any tests or utilities in this repo or otherwise that were expecting the the original codes. I assume not since it would have been hard to compare since this is the whole point of this fix :)

I think not. Otherwise, the issue would have been discovered sooner. Theoretically, somebody outside of Joystream could have a custom bash script that relies on particular exit codes returned by our CLI(s), but I think it's unlikely.

Issue #3231, opened 2022-02-15
Olympia: BountyCreated and NftIssued event decoding fails 


Lezek123:
Due to `@polkadot/typegen` not supporting nested `BTreeSet` inside an `enum` type, I temporarly changed the definition of `AssuranceContractType` to:
```typescript
export class AssuranceContractType extends JoyEnum({
  Open: Null,
  Closed: Vec.with(MemberId), // FIXME: @polkadot/typegen Error: Enum: AssuranceContractType: Unhandled nested "BTreeSet"
}) {}
```

This caused issues when decoding `BountyCreated` event, for example:
```
Unable to retrieve the specified block details. Unable to decode storage system.events:: createType(Vec<EventRecord>):: Vec<EventRecord>:: Decoded input doesn't match input, received 0x1c00000000000000482d7c090000000002000000010000000000000000000000…000000000a0000000a000000000000020000000000d8f0e05000000000000000 (349 bytes), created 0x1c00000000000000482d7c090000000002000000010000000000000000000000…0000000000000000102700000000000000000000000000000010270000000000 (312 bytes)
```

### Trying out solutions

As an alternative I tried creating separate type for `AssuranceContractType_Closed`:
```typescript
export class AssuranceContractType_Closed extends BTreeSet.with(MemberId) {}

export class AssuranceContractType extends JoyEnum({
  Open: Null,
  Closed: AssuranceContractType_Closed,
}) {}
```

The type definitions were then correctly generated:
```
    "AssuranceContractType_Closed": "BTreeSet<MemberId>",
    "AssuranceContractType": {
        "_enum": {
            "Open": "Null",
            "Closed": "AssuranceContractType_Closed"
        }
    },
```

However some issues still seem to occur when trying to create a bounty with `AssuranceContractType: Closed`:

![image](https://user-images.githubusercontent.com/19978260/154045517-f15a1e17-4ddc-4144-9411-5588027c72cd.png)

As you can notice - even though the `BountyCreationParameters` seem valid, the actual `bounty.createBounty` call cannot be decoded:

```
[...]

PR #3268, opened 2022-02-18
Fix NftIssued and BountyCreated event decoding issues

Lezek123:
Fixes issues discovered while working on https://github.com/Joystream/joystream/issues/3231

Namely:
- The [runtime was using `Metadata` alias for `Vec<u8>`](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/nft/types.rs#L4) and then [as part of `NftIssued` event](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/lib.rs#L2224). `Metadata` is a reserved typename in `@polkadot/types` which has a special meaning (it's used do decode chain metadata), so this caused issues when decoding `NftIssued` event.
- `BountyId` and `EntryId` were incorrectly defined in `@joystream/types` (`u32` instead of `u64`)
- `AssuranceContractType.Closed` was defined as `Vec.with(MemberId)` instead of `BTreeSet.with(MemberId)` due to `@polkadot/typegen` not handling a nested `BTreeSet` inside enum unless it's part of some other type. This was fixed by introducing an `AssuranceContractType_Closed` alias.

I also fixed a small issue with `yarn clean` beeing executed during `yarn build` for `@joystream/types`. Since `yarn clean` removes all files unrecognized by git, this could've caused a loss of pending work (ie.: a developer adds a new file to `./types/src`, then runs `yarn build` before adding the file via `git add`. The file is getting "cleaned up" and lost.)


mnaamani:
Fantastic fixe :+1: 
Looks like @ondratra also addressed the same problem by renaming of the Metadata type to NftMetadata in the NFT mappings PR https://github.com/Joystream/joystream/pull/3195

I applied the bounty type fixes to the running stating network where some bounties were created and broke the query-node indexer and it worked :+1:

Issue #3273, opened 2022-02-21
Improve handling ALL_PROPOSALS_PARAMETERS_JSON


mnaamani:
There a few issues with how invalid ALL_PROPOSALS_PARAMETERS_JSON affects behavior

1 - invalid JSON is not caught at build time or when running tests
2 - comments describe behavior such that improper json will cause a fallback to defaults. This is not the case, the runtime panics instead. Which behavior do we want?
3 - Its not clear how negative numbers (valid json number) in the JSON processed.
4 - `option_env!("ALL_PROPOSALS_PARAMETERS_JSON")` handles empty string and null differently on Linux vs Mac this has resulted in compiling a runtime where an empty string was set for the variable, which caused the runtime to panic with ""Invalid JSON with proposals parameters provided." since an empty string returned an error by `lite_json::parse_json()`

Proposal
1 - try to catch invalid json during build if possible, or at minimum in unit tests
2 - I think fallback is better than panic! 
3 - Will test
4 - Add explicit handling of empty string or variable not set to both result in "default" values being used.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201857679218381) by [Unito](https://www.unito.io)

mnaamani:
> 3 - Its not clear how negative numbers (valid json number) in the JSON processed.

As expected it will convert to highest value of unsigned value

mnaamani:
done in  #3275

PR #3275, opened 2022-02-21
Olympia: Improved handling of proposal params through json config

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3273

1 - Catch invalid JSON when running tests.
2 - ~Fallback to default values instead of `panic!` in runtime. (panic only in unit tests)~
3 - Empty string and unset variable treated as no explicit JSON config is being passed, and defaults are used.
4 - Pass ALL_PROPOSALS_PARAMETERS_JSON as build arg to dockerfile


runtime/src/proposals_configuration/mod.rs
```diff
@@ -109,12 +109,24 @@ lazy_static! {
 fn get_all_proposals_parameters_objects() -> AllProposalsParameters {
     let json_str: Option<&'static str> = option_env!("ALL_PROPOSALS_PARAMETERS_JSON");
 
+    // Handle undefined variable (null) and variable set to empty string the same
+    // to work cross platform.
+    if json_str.is_none() || json_str.unwrap().is_empty() {
```
shamil-gadelshin: The `unwap()` method should be avoided in the production code. The first part of the condition `.is_none()` could be modified and we would end up with panic. Using the `map_or` looks like a safer approach.

runtime/src/proposals_configuration/mod.rs
```diff
@@ -109,12 +109,24 @@ lazy_static! {
...
+        return default_parameters();
+    }
+
     json_str
         .map(lite_json::parse_json)
         .map(|res| match res {
             Ok(json) => Some(json),
             Err(_) => {
-                panic!("Invalid JSON with proposals parameters provided.");
+                if cfg!(test) {
```
shamil-gadelshin: Why do we allow invalid JSON in non-test cases?

shamil-gadelshin (changes requested):
Looks good. There are a couple of minor comments. It looks like there are clippy issues as well present in the code.

I corrupted the `./tests/integration-tests/proposal-parameters.json` file, ran the `./build-node-docker.sh`, and expected the build to fail but it was OK.

runtime/src/proposals_configuration/mod.rs
mnaamani: I was trying to follow the description of the expected behavior in https://github.com/Joystream/joystream/blob/8bb0a84bc822eb817401e8ddd0f00e1613a40ccb/runtime/src/proposals_configuration/mod.rs#L6

Specifically "If it doesn't exists or contains invalid or empty JSON then the default parameters are returned."

mnaamani:
> Looks good. There are a couple of minor comments. It looks like there are clippy issues as well present in the code.
> 

The clippy warning are because I commented out the proposals_integration module in `runtime/src/tests/mod.rs` but didn't comment out some common code which was being used from those tests.

> I corrupted the `./tests/integration-tests/proposal-parameters.json` file, ran the `./build-node-docker.sh`, and expected the build to fail but it was OK.

Did you run with `TEST_NODE=true ./build-node-docker.sh` ?

runtime/src/proposals_configuration/mod.rs
mnaamani: But if we want compile time and runtime behavior to be the same, I can drop the condition `cfg!(test)` and always return `None` here.

runtime/src/proposals_configuration/mod.rs
mnaamani: Fixed in 565dd4f536bb879de3dad829fa64656f7e98f7a3

runtime/src/proposals_configuration/mod.rs
mnaamani: As discussed reverted to panic, and updated the comments to reflect actual behavior

shamil-gadelshin (commented):
LGTM

PR #3299, opened 2022-02-23
Query node: Fix unsafe toNumber() conversion

Lezek123:
Fixes a known issue where a very high number provided as  `ForumPoll.end_time` can break the query node processor and a few other similar issues.

There are two ways in which a high number can cause issues in the query node:
- When `.toNumber()` is called on `u64` / `u128`, the `bn.js` lib will throw an error in case the number is > `Number.MAX_SAFE_INTEGER`
- When a number is inserted into the database, the column type is `Int` and the number exceeds `2147483647`, this will cause a database error

I tried to address both security issues in places where they could have been abused by a single member (ie. assuming the entire council will not conspire to approve proposals with unreasonably high values)

### DIFF
--- a/query-node/mappings/src/common.ts
+++ b/query-node/mappings/src/common.ts
@@ -5,8 +5,11 @@ import { Worker, Event, Network } from 'query-node/dist/model'
 import { BaseModel } from '@joystream/warthog'
 import { metaToObject } from '@joystream/metadata-protobuf/utils'
 import { AnyMetadataClass, DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
+import BN from 'bn.js'
 
 export const CURRENT_NETWORK = Network.OLYMPIA
+
+export const INT32MAX = 2147483647
 /*
   Simple logger enabling error and informational reporting.
 
@@ -301,4 +304,14 @@ export function deterministicEntityId(createdInEvent: SubstrateEvent, additional
     `${createdInEvent.blockNumber}-${createdInEvent.indexInBlock}` +
     (additionalIdentifier ? `-${additionalIdentifier}` : '')
   )
+}
+
+// Convert a BN to number without throwing an error if > Number.MAX_SAFE_INTEGER
+// add with a custom limit to maxValue if needed
+export function toNumber(value: BN, maxValue = Number.MAX_SAFE_INTEGER): number {
+  try {
+    return value.toNumber() > maxValue ? maxValue : value.toNumber()
+  } catch (e) {
+    return maxValue
+  }
 }
--- a/query-node/mappings/src/common.ts
+++ b/query-node/mappings/src/common.ts
@@ -316,8 +316,18 @@ export function deterministicEntityId(createdInEvent: SubstrateEvent, additional
 // add with a custom limit to maxValue if needed
 export function toNumber(value: BN, maxValue = Number.MAX_SAFE_INTEGER): number {
   try {
-    return value.toNumber() > maxValue ? maxValue : value.toNumber()
+    if (value.toNumber() > maxValue) {
+      logger.info(`toNumber() Warning: Input value ${value.toNumber()} exceeds maxValue: ${maxValue}.`)
+      return maxValue
+    }
+
+    return value.toNumber()
   } catch (e) {
+    logger.info(
+      `toNumber() Warning: BN.toNumber() conversion error: ${
+        e instanceof Error ? e.message : e
+      }. Returning maxValue: ${maxValue}`
+    )
     return maxValue
   }
 }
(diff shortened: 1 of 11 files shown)


query-node/mappings/src/common.ts
```diff
@@ -302,3 +305,13 @@ export function deterministicEntityId(createdInEvent: SubstrateEvent, additional
     (additionalIdentifier ? `-${additionalIdentifier}` : '')
   )
 }
+
+// Convert a BN to number without throwing an error if > Number.MAX_SAFE_INTEGER
+// add with a custom limit to maxValue if needed
+export function toNumber(value: BN, maxValue = Number.MAX_SAFE_INTEGER): number {
+  try {
+    return value.toNumber() > maxValue ? maxValue : value.toNumber()
+  } catch (e) {
```
mnaamani, line 325: could BN.toNumber() throw for a different reason?

We might want to log a warning that this is happening, it could be useful for debugging since we are not storing the real values.

query-node/mappings/src/common.ts
```diff
@@ -5,8 +5,11 @@ import { Worker, Event, Network } from 'query-node/dist/model'
 import { BaseModel } from '@joystream/warthog'
 import { metaToObject } from '@joystream/metadata-protobuf/utils'
 import { AnyMetadataClass, DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
+import BN from 'bn.js'
 
 export const CURRENT_NETWORK = Network.OLYMPIA
+
+export const INT32MAX = 2147483647
```
mnaamani, line 13: I would just add comment to describe this as the max value the database can store of the `Int` type.

query-node/mappings/src/forum.ts
```diff
@@ -307,7 +309,7 @@ export async function forum_ThreadCreated(ctx: EventContext & StoreContext): Pro
       createdAt: eventTime,
       updatedAt: eventTime,
       description: bytesToString(pollInput.unwrap().description),
-      endTime: new Date(pollInput.unwrap().end_time.toNumber()),
+      endTime: new Date(toNumber(pollInput.unwrap().end_time, new Date('3000-01-01').getTime())),
```
mnaamani: I don't think the query-node should be opinionated about what the max timestamp should be. I know you picked a number far in the future, but still. Might as well just make it INT32MAX.

Perhaps we should change the schema to allow storing full u64? This general issue has been covered before?

query-node/schemas/forumEvents.graphql
```diff
@@ -360,7 +360,7 @@ type PostReactionResultValid @variant {
 }
 
 type PostReactionResultInvalid @variant {
-  reactionId: Int!
+  reactionId: String!
```
mnaamani, line 363: I don't know if Pioneer currently uses reactions, but they should be aware of this change probably
cc: @thesan

mnaamani (changes requested):
Awesome finding all other places where conversion could have broken the query-node.
I suppose this is the best option without changing the database schema.
Left a couple of points, but otherwise LGTM.

query-node/mappings/src/forum.ts
Lezek123: > It should be noted that the maximum Date is not of the same value as the maximum safe integer (Number.MAX_SAFE_INTEGER is 9,007,199,254,740,991). Instead, it is defined in ECMA-262 that a maximum of ±100,000,000 (one hundred million) days relative to January 1, 1970 UTC (that is, April 20, 271821 BCE ~ September 13, 275760 CE) can be represented by the standard Date object (equivalent to ±8,640,000,000,000,000 milliseconds).

_Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date_

So according to this, we can cap to `~8640000000000000`, but not to `MAX_SAFE_INTEGER`.
I have just chosen some very high value, since I noticed that `new Date(Number.MAX_SAFE_INTEGER)` doesn't work.

Lezek123: Introduced `TIMESTAMPMAX` const in https://github.com/Joystream/joystream/pull/3299/commits/c7a129d856126767c1b2b43962de4728dddba3a2

query-node/mappings/src/common.ts
Lezek123: Done in https://github.com/Joystream/joystream/pull/3299/commits/c7a129d856126767c1b2b43962de4728dddba3a2

query-node/mappings/src/common.ts
Lezek123: Done in https://github.com/Joystream/joystream/pull/3299/commits/fa542bd8a57372d856a932c4f2c4395bb9f17ef8

Issue #3288, opened 2022-02-23
Fix Proposal Tests and Re-enable


mnaamani:
After modifying default proposal parameters many of the tests broke (they were too coupled with the default chosen parameters) so they were disabled:
https://github.com/Joystream/joystream/blob/cfd7faa9253cb08a007a1e0e4f8cd8eb69a2df0c/runtime/src/tests/mod.rs#L5

Some tests that where coupled to the size of council 3, where fixed in https://github.com/Joystream/joystream/pull/3252

Need to fix remaining tests to be sensitive to the actual values, instead of making false assumption about what the values were initially set to.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201870467439723) by [Unito](https://www.unito.io)

mnaamani:
done in https://github.com/Joystream/joystream/pull/3316

Issue #3242, opened 2022-02-16
Olympia (staging): Testing Parameters and Configurations


bwhm:
We'll have a "public" network, with two `invuls` and a query-node.
Besides `sudo` and `endowed` key(s), no state imported.

## Parameters
These are needed for the testers.

### Proposal Parameters
From: `runtime/src/proposals_configuration/defaults.rs`

| proposalType                                        | Voting P | Grace P | Appr. Q % | Appr. T % | Slash. Q % | Slash. T % | Req. Stake | Const. |
|-----------------------------------------------------|----------|---------|-----------|-----------|------------|------------|------------|--------|
| set_max_validator_count_proposal                    | 200      | 0       | 66        | 80        | 60         | 80         | 100000     | 1      |
| runtime_upgrade_proposal                            | 200      | 200     | 80        | 100       | 60         | 80         | 1000000    | 1      |
| signal_proposal                                     | 200      | 0       | 60        | 80        | 60         | 80         | 25000      | 1      |
| funding_request_proposal                            | 200      | 100     | 60        | 80        | 60         | 80         | 25000      | 1      |
| create_working_group_lead_opening_proposal          | 200      | 0       | 60        | 80        | 60         | 80         | 100000     | 1      |
| fill_working_group_lead_opening_proposal            | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| update_working_group_budget_proposal                | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| decrease_working_group_lead_stake_proposal          | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| slash_working_group_lead_proposal                   | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| set_working_group_lead_reward_proposal              | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| terminate_working_group_lead_proposal               | 200      | 0       | 66        | 80        | 60         | 80         | 100000     | 1      |
| amend_constitution_proposal                         | 200      | 100     | 80        | 100       | 60         | 80         | 1000000    | 2      |
| cancel_working_group_lead_opening_proposal          | 200      | 0       | 60        | 75        | 60         | 80         | 50000      | 1      |
| set_membership_price_proposal                       | 200      | 100     | 60        | 75        | 60         | 80         | 50000      | 1      |
[...]

PR #3316, opened 2022-02-25
Fix runtime proposal integration cargo tests

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3288

- [x] Fix issues with false assumptions about size of council and member account re-use causing staking conflict issues
- [x] Make tests sensitive to real values of required stakes
- [x] Make tests sensitive to change in proposal grace periods
  - didn't change working group related proposals assuming they will always be 0 grace period
- [x] Ensure proposed changes are indeed different than current values?
- [x] Fixed one proposal default parameter values as per https://github.com/Joystream/joystream/issues/3242


shamil-gadelshin (approved):
LGTM

Issue #3070, opened 2022-01-22
Should caller decide bag deletion prize in storage pallet


bedeho:
# Background
From the signature of bag creation and bag creation feasibility checkers, it seems the caller gets to decide what - if any, prize should be used, since it has payload `deletion_prize: &Option<DynamicBagDeletionPrize<T>>,`.

# Question
Is this really sensible, as all bags are effectively equal in terms of what the deletion prize is trying to police: state bloat, regardless of whether its a channel or membership bag, or certainly , whether its channel A or channel B. It seems the simplest approach is to either have a tunable value in `storage` pallet which is indeed universal, or at worst , oner per dynamic bag type, but this feels over-engineered.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201810379947921) by [Unito](https://www.unito.io)

shamil-gadelshin:
The code was merged.

PR #3317, opened 2022-02-25
Audit3. Runtime: convert dynamic bag and data object deletion prizes to storage values.

shamil-gadelshin:
We need to convert deletion prizes for the dynamic bags and data objects to the updatable values in the storage pallets.

#### Major changes
- changed `DataObjectDeletionPrize` from the runtime const to the `storage` pallet's storage value
- changed `DynamicBagDeletionPrize` from the external parameters for methods to the `storage` pallet's storage value
- new `update_dynamic_bag_deletion_prize` extrinsic for the storage pallet
- new `update_data_object_deletion_prize` extrinsic for the storage pallet
- updated `content` pallet to support changes in the `storage` pallet
- updated tests

[Related issue](https://github.com/Joystream/joystream/issues/3070)


bedeho:
This is not for olympia, its not even for Rhodes, its for audit branch, but lets try to postpone merging until we must.

runtime-modules/storage/src/lib.rs
```diff
@@ -2785,6 +2834,19 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
 }
 
 impl<T: Trait> Module<T> {
+    // Creates optional DynamicBagDeletionPrize. Return None when the deletion prize is zero.
+    fn create_dynamic_bag_deletion_prize_record(
+        account_id: T::AccountId,
+    ) -> Option<DynamicBagDeletionPrize<T>> {
+        let prize = Self::dynamic_bag_deletion_prize_value();
+
+        if prize == Zero::zero() {
```
dobertRowneySr: Have you tried the `is_zero()` https://docs.substrate.io/rustdocs/latest/sp_arithmetic/traits/trait.Zero.html#tymethod.is_zero ?

dobertRowneySr (approved):

runtime-modules/content/src/lib.rs
```diff
@@ -363,22 +363,17 @@ decl_module! {
...
 
-            let deletion_prize = storage::DynamicBagDeletionPrize::<T> {
-                prize: Zero::zero(), // put 0 for Giza release
-                account_id: sender,
-            };
-
             if Storage::<T>::ensure_bag_exists(&channel_bag_id).is_err() {
                 if let Some(params) = upload_params.clone() {
                     Storage::<T>::can_create_dynamic_bag_with_objects_constraints(
                         &DynamicBagIdType::<T::MemberId, T::ChannelId>::Channel(channel_id),
```
dobertRowneySr: I am not sure if it is an issue or not, but I will point it out:
Storage lead can race against `sender` (in this case) with a `update_dynamic_bag_deletion_prize` extrinsic. Changing the `deletion_prize` just before this one is executed

dobertRowneySr (commented):
It looks ok. I just left a small remark. But I am not sure if it is an issue or not

runtime-modules/storage/src/lib.rs
shamil-gadelshin: The affected method was removed, so no change is needed.

runtime-modules/content/src/lib.rs
shamil-gadelshin: It could be executed. @bedeho Should we restore commitment for the bag and object deletion prizes?

bedeho: Nicely spotted @ignazio-bovo .

Well, its not clear that this really solves the full problem in a realistic way, because all that would happen is that 99% of users using apps which just read the current deletion prize would just end up sucking up all the funds in the end anyway. So if a lead is willing to do this, he will get the funds locked none the less.

I think adding commitment is fine, but probably this should be up to council to update in the end anyway.

shamil-gadelshin:
#### New updates
- merged lastest audit3 code
- added deletion prizes commitment for dynamic bags and data objects for storage pallet
- affected extrinsics in the content pallet:
  - create_channel
  - update_channel
  - create_video
  - update_video

dobertRowneySr (commented):
LGTM

dobertRowneySr (approved):

PR #3327, opened 2022-02-28
Query node: Trigger preBlock/postBlock hooks for each block

Lezek123:
Since this issue is hard to address on the Hydra side, due to the fact that the entire flow is based around event processing, I decided to implement a workaround and add `system.ExtrinsicSuccess` event to the list of supported events in `query-node/manifest.yml`.

Because `timestamp.set` extrinsic is part of basically every block, each block will have at least 1 event recognized by the processor, meaning `preBlockHooks` and `postBlockHooks` will never be delayed due to lack of recognized events in a given block.


mnaamani:
That's a pretty neat trick :+1: 
LGTM

Issue #2963, opened 2021-12-22
Giza query node playgrounds - query templates


ondratra:
Query templates for GraphQL's playground introduced in https://github.com/Joystream/joystream/pull/2461 are not present in Sumer got lost in Giza. The cause is that in Giza, graphql files are generated on each build, while on Sumer those generated files were committed to git.

Recover query templates from Sumer and add the feature to Giza query node. Copy `queryTemplates.ts` and make sure it's passed as QN server start parameters in `query-node/generated/graphql-server/src/index.ts` `getServer()` (see https://github.com/Joystream/joystream/pull/2461/files#diff-228448314cf659f33b14c4089e3c2d74454221a775f1279362da4fef27f80924).

Additionally, create PR for GraphQL Playground to the upstream repository. Right now, the feature is only supported in our fork https://github.com/Joystream/graphql-playground .

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201904299016570) by [Unito](https://www.unito.io)

ondratra:
Missing feature(s) as well as generally broken playground is related to https://github.com/Joystream/joystream/issues/3271 .

PR #3334, opened 2022-03-02
Giza query node - graphql playground fix and query templates

ondratra:
Fixes endpoint issues in Query node's GraphQL Playgrounds and enables query templates support.

Fixes https://github.com/Joystream/joystream/issues/2963 and partially fixes Playground problems described in  https://github.com/Joystream/joystream/issues/3271 - for Giza release.

Issue #3271, opened 2022-02-19
Question/problem: How to check query node status externally?


mochet:
# Problem

Query Nodes seem to have at least a few reliability issues right now (👷) and it seems like there isn't any way for someone without server access to the QN to check what its sync status is (i.e. what is the most recent blockheight processed). I could be wrong, but just using what I know to describe this!

Via graphql there does seem to be an option for this: https://olympia-staging-a2.joystream.app/server/graphql
```
stateSubscription: ProcessorState!
type ProcessorState {
lastCompleteBlock: Float!
lastProcessedEvent: String!
indexerHead: Float!
chainHead: Float!
}
```

However running this results in the following error:
```
{
  "error": "Could not connect to websocket endpoint ws://localhost:8081/graphql. Please check if the endpoint url is correct."
}
```

Alternatively if I try this via https://3.93.173.102.nip.io/query-node/indexer/graphql I get the following error:
```
{
  "error": "Could not connect to websocket endpoint wss://3.93.173.102.nip.io/graphql. Please check if the endpoint url is correct."
}
```

Via the indexer URL it is possible to run the following (which works), it is just unclear if this can be relied upon when accessing the query node:
```query{
  indexerStatus{head}
}
```

# Questions
1. Is using `indexerStatus` via the indexer URL as good as the Query Node's status? Or would `indexerStatus` be unreliable for determining current sync?

# Solution
1. Either have `stateSubscription` work, or
2. Have some public facing URL from each QN that shows its status (something like https://status.joystream.org/status)

# Reason
Besides the benefits given when troubleshooting, the other benefit with solving this would be that a UI can be made which shows the current "network" of QNs and shows which are currently having issues.

bedeho:
> and it seems like there isn't any way for someone without server access to the QN to check what its sync status is

This def. can't be true, because Atlas already has the behavior that when the user is making a transaction, such as when making a channel, it blocks the user in a modal until it confirms that the query node has processed the block in which this transaction was included. I also believe Pioneer has been built to have a progress indicator type dialogue which shows similar processing information about the query node.

> However running this results in the following error:

Yeah not sure what that is about, but I think this is what should work.

> Is using indexerStatus via the indexer URL as good as the Query Node's status? Or would indexerStatus be unreliable for determining current sync?

[...]

PR #3335, opened 2022-03-02
Olympia graphql playground fix

ondratra:
Fixes endpoint issues in Query node's GraphQL Playgrounds and enables query templates support. Includes changes from https://github.com/Joystream/joystream/pull/3334 .

Fixes https://github.com/Joystream/joystream/issues/2963 and partially fixes Playground problems described in https://github.com/Joystream/joystream/issues/3271 - for Olympia release.


.env
```diff
@@ -24,8 +24,15 @@ BLOCK_HEIGHT=0
 # Query node GraphQL server port
 GRAPHQL_SERVER_PORT=8081
 
-# Query node playground subscription endpoint
-GRAPHQL_PLAYGROUND_SUBSCRIPTION_ENDPOINT=ws://localhost:8081/graphql
+# Query node playground endpoint
+GRAPHQL_PLAYGROUND_ENDPOINT=http://localhost:8081/graphql
```
mnaamani: Could you clarify how this is used? If I host the query-node on a public endpoint using a reverse proxy like caddy or nginx for example`https://www.somewhere.com/server/graphql` does this env variable have to be updated to match ?

mnaamani: We would also want to apply similar config for the hydra-indexer-gateway graphql endpoint as well?

Can we solve this through proper configuration of HTTP headers forwarding at the proxy level? So that the graphql playground can properly determine what endpoint it was visited on, rather than having to manually confiugre it through env variables? Or Does the webserver used by the playground not look at the appropriate HTTP headers?

ondratra: The variable enables overwrite of playgrounds default endpoint. It seems that this value was not optimal. I changed `http://localhost:8081/graphql` to an empty value and added a comment about this. The empty value should mean that the endpoint from the browser's URL will be used. So if a proper proxy is set up, it should work ok but can be overwritten with this variable if needed.

I'll look into how to set this value for `hydra-indexer-gateway` in case we ever needed it; I forgot about gateway before you mentioned it.

mnaamani:
So far was able to test the url fix :+1: Hosted it on a custom url and it was automatically picked up in the playground

Screen Shot 2022-03-07 at 5 21 46 PM

mnaamani:
The query templates are showing up but a lot of them I think need to be updated because they are for giza schema so I get errors in the query results:

Screen Shot 2022-03-08 at 10 21 07 AM

mnaamani:
Tested now and I don't see the Graphql validations errors anymore

Issue #3322, opened 2022-02-27
Metaprotocol Comments & Reactions - feature list


zeeshanakram3:
This issue lists the Metaprotocol based implementation of basic features related to comment & reaction.

For each action/feature listed here I have described different pieces involved from start to end (i.e. from calling the `*_remark` extrinsic, to processing the message in mappings, and then updating the QN state):

`Message`: Protocol buffers message used for given action, 
`Event`: The event emitted (containing serialized protobuf message) by the `*_remark` extrinsic,
`Validations`: Validations performed on the message in mappings before making it part of Query Node state,

## Schema
```graphql
enum ReactionOptions {
  CANCEL
  LIKE
  UNLIKE
}

type VideoReaction @entity {
  "Unique identifier"
  id: ID!

  "The Reaction"
  reaction: ReactionOptions!

  "The member that reacted"
  member: Membership!

  "The video that has been reacted to"
  video: Video!
}

type CommentReaction @entity {
  "Unique identifier"
  id: ID!

  "The Reaction"
  reaction: ReactionOptions!

  "The member that reacted"
  member: Membership!

  "The comment that has been reacted to"
  comment: Comment!
}

type Comment @entity {
  "Unique identifier"
  id: ID!

  "Author of the video comment"
  author: Membership!

  "Comment text"
  text: String @fulltext(query: "commentText")

  "Video the comment was added to"
  video: Video!

  "True if the comment is visible; moderators can set this flag"
  isVisible: Boolean!

  "List of all reactions to the comment"
  reactions: [CommentReaction!] @derivedFrom(field: "comment")

  "A (parent) comment that this comment replies to (if any)"
  parentCommentId: Comment
}
```

#### Message 
```proto
message ReactVideo {
  required uint64 video_id = 1;
  enum Reaction {
    CANCEL = 0;
    LIKE = 1;
    UNLIKE = 2;
   }
   required Reaction reaction = 2;
}
```
#### Event
`MemberRemarked`

#### Validations 
- _Ensure video exists_ 
- _Ensure comment section is enabled_

### Message 
Same message used to react a video, with enum field set to `CANCEL`
#### Event
`MemberRemarked`

#### Validations
- _Ensure video exists_ 
- _Ensure comment section is enabled_
- _Ensure member reacted previously_

#### Message 
```proto
message ReactComment {
  required uint64 comment_id = 1;
  enum Reaction {
    CANCEL = 0;
    LIKE = 1;
    UNLIKE = 2;
  }
[...]

Issue #2932, opened 2021-12-15
QN Video Comments and Reactions Metaprotocol


bedeho:
### Context
:information_source: Based on this https://github.com/Joystream/joystream/issues/1990, 
:warning:  Depends on this https://github.com/Joystream/joystream/issues/2919.
:no_entry_sign:  Blocking this https://github.com/Joystream/atlas/issues/1876 

### Scope
* [ ] Add QN support for Video Comments and Reactions 

### Note
A missing part of current on-chain implementation is "unreaction" to video and comment, we should have this.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958580174348) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1201958687417145/1201958580174348

bedeho:
Done.

PR #3340, opened 2022-03-02
Comments-Reactions: review followups

zeeshanakram3:
implements #2932 

list at #3322


kdembler (commented):
All right, I feel really excited finally getting to this, can't wait to get this integrated! In general, parts I was able to test worked pretty well! Here's first round of feedback:

1. Processor seems to be crashing if an invalid message is provided. I ran into that by a mistake when I didn't wrap the message into member remark. We need to be very careful here, remark can contain _anything_ and the processor shouldn't crash regardless of what that is.
2. `Video.reactions` isn’t really useful, could we add something like `Comment.reactionsCountByReactionId` instead?
3. `Comment.reactionsCount` is missing, this will be very useful for sorting
4. I don’t think I’m able to fetch only top-level comments, excluding replies. Maybe we could have `Comment.isTopLevel`?
5. `editComment` and `deleteComment` messages require providing `videoId`. Is that necessary since `commentId` is already unique?
6. Like discussed in Slack, I think we will need a separate message for channel owner to remove comments, because the owner may not be in the moderators list.
7. `editComment` did not work for me. When processor handled the block in which it was emitted it thrown this error and the edit wasn’t applied in any way:

```jsx
(node:40) UnhandledPromiseRejectionWarning: QueryRunnerAlreadyReleasedError: Query runner already released. Cannot run queries anymore.
    at new QueryRunnerAlreadyReleasedError (/joystream/node_modules/typeorm/error/QueryRunnerAlreadyReleasedError.js:10:28)
    at PostgresQueryRunner.<anonymous> (/joystream/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:218:35)
    at step (/joystream/node_modules/tslib/tslib.js:143:27)
    at Object.next (/joystream/node_modules/tslib/tslib.js:124:57)
    at /joystream/node_modules/tslib/tslib.js:117:75
    at new Promise (<anonymous>)
    at Object.__awaiter (/joystream/node_modules/tslib/tslib.js:113:16)
    at PostgresQueryRunner.query (/joystream/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:212:24)
    at SelectQueryBuilder.<anonymous> (/joystream/node_modules/typeorm/query-builder/SelectQueryBuilder.js:1654:62)
    at step (/joystream/node_modules/tslib/tslib.js:143:27)
```

8. Could we add `Comment.isEdited` flag? This way we don’t need to fetch all the edits for every comment just to know which ones were edited
9. I’m not able to apply any sorting to edits made to a comment. Is there a guarantee that this list will come be sorted? If so, in which direction and by what param? Maybe it would make sense to have a separate query just for edit made to a comment? This way we can apply sorting ourselves. Separate query is not a must though, if this is most-recent first, that will work
10. I see that when removing a message as a moderator, I can provide a `rationale`. Is that data returned by any query?
11. There is still one important piece of data missing from both video and comments reactions and that's info on whether user running the query has submitted a given reaction. I mentioned this [here](https://github.com/Joystream/joystream/issues/3322#issuecomment-1054685917) (2nd open question) and [here](https://github.com/Joystream/joystream/issues/3322#issuecomment-1068405144). This may be tricky to solve, if nothing comes to mind, let's sync up on this and brainstorm workarounds

ondratra:
`UnhandledPromiseRejectionWarning: QueryRunnerAlreadyReleasedError: Query runner already released. Cannot run queries anymore.` error means there is a promise that's not being `await`ed.

metadata-protobuf/doc/index.md
```diff
@@ -64,6 +92,56 @@
 
 
 
+<a name="proto/Bounty.proto"></a>
+<p align="right"><a href="#top">Top</a></p>
+
+## proto/Bounty.proto
+
+
+
```
ondratra: Let's remove all consecutive empty lines in this file.

metadata-protobuf/proto/Metaprotocol.proto
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// edit comment by author
+message EditComment {
+    // ID of the comment whose text is being edited
+    required string comment_id = 3;
+
+    // New comment body
+    required string new_body = 4;
+}
```
ondratra: While most `message`s start with item number `1`, but `EditComment` and couple of others start with a higher number (`3` in this case), is that ok?

query-node/mappings/src/content/channel.ts
```diff
@@ -236,3 +270,130 @@ export async function content_ChannelDeleted({ store, event }: EventContext & St
...
+
+export async function content_ChannelOwnerRemarked(ctx: EventContext & StoreContext): Promise<void> {
+  const { event, store } = ctx
+  const [owner, channelId, message] = new Content.ChannelOwnerRemarkedEvent(ctx.event).params
+
+  // unique identifier for metaprotocol tx
+  const { id: metaprotocolTxIdentifier } = genericEventFields(event) as BaseModel
+
+  const metaprotocolTxStatusEvent = new MetaprotocolTransactionStatusEvent({
+    ...genericEventFields(event),
```
ondratra: Let's not call `geenericEventFields` twice here. Please fix this on other places as well.

kdembler:
Issues left from our perspective:

- [x] Still no way to fetch only top-level comments I think? (unless Hydra allows to somehow filter by field being `null` which I'm not aware of)
- [x] It seems there is no way to fetch the original comment once it's been edited? I can query specific edits but the original text is lost? Or maybe we don't know how to get it?
- [ ] Point 11. from my comment above is still unresolved I think?

query-node/mappings/src/content/channel.ts
```diff
@@ -236,3 +270,130 @@ export async function content_ChannelDeleted({ store, event }: EventContext & St
...
+      await updateMetaprotocolTransactionStatus(store, metaprotocolTxIdentifier, statusSuccessful)
+    }
+  } catch (e) {
+    const statusErrored = new MetaprotocolTransactionErrored()
+
+    if (typeof e === 'string') {
+      statusErrored.message = e
+    } else if (e instanceof Error) {
+      statusErrored.message = e.message
+    }
```
ondratra:
```suggestion
    statusErrored.message = e instanceof Error
      ? e.message
      : e.toString()
```

query-node/mappings/src/content/channel.ts
```diff
@@ -3,13 +3,47 @@ eslint-disable @typescript-eslint/naming-convention
...
+  processPinOrUnpinCommentMessage,
+  processBanOrUnbanMemberFromChannelMessage,
+  processCommentSectionPreferenceMessage,
+  processVideoReactionsPreferenceMessage,
+} from './commentAndReaction'
 import { getAllManagers } from '../derivedPropertiesManager/applications'
+import { BaseModel } from '@joystream/warthog'
 
 export async function content_ChannelCreated(ctx: EventContext & StoreContext): Promise<void> {
   const { store, event } = ctx
```
ondratra:
```suggestion
export async function content_ChannelCreated({ store, event }: EventContext & StoreContext): Promise<void> {
```
Let's keep the new code consistent with existing mappings. Please fix it in all other places as well.

query-node/mappings/src/content/channel.ts
```diff
@@ -236,3 +270,130 @@ export async function content_ChannelDeleted({ store, event }: EventContext & St
...
+    if (!messageType) {
+      invalidMetadata('Unsupported message type in channel_owner_remark action')
+    }
+
+    if (messageType === 'pinOrUnpinComment') {
+      await processPinOrUnpinCommentMessage(ctx, contentActor, channelId, decodedMessage.pinOrUnpinComment!)
+      const statusSuccessful = new MetaprotocolTransactionSuccessful()
+      await updateMetaprotocolTransactionStatus(store, metaprotocolTxIdentifier, statusSuccessful)
+      return
+    }
```
ondratra: There are multiple very similar parts of code that differ only in `decodedMessage`'s property used. Let's generalize it to a function and use it on all places.

query-node/mappings/src/content/channel.ts
```diff
@@ -236,3 +270,130 @@ export async function content_ChannelDeleted({ store, event }: EventContext & St
...
+    const statusErrored = new MetaprotocolTransactionErrored()
+
+    if (typeof e === 'string') {
+      statusErrored.message = e
+    } else if (e instanceof Error) {
+      statusErrored.message = e.message
+    }
+
+    await updateMetaprotocolTransactionStatus(store, metaprotocolTxIdentifier, statusErrored)
+  }
```
ondratra: This code is also used in `content_ChannelOwnerRemarked`. Let's generalize it.

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -0,0 +1,708 @@
...
+  VideoReaction,
+  VideoReactionOptions,
+  VideoReactionsCountByReactionType,
+  VideoReactionsPreferenceEvent,
+} from 'query-node/dist/model'
+import { genericEventFields, inconsistentState, newMetaprotocolEntityId } from '../common'
+
+// TODO: Ensure video is actually a video
+// TODO: make sure comment is fully removed (all of its reactions)
+// TODO: make sure video is fully removed (all of its comments & reactions)
```
ondratra: This will be part of further PRs, right?

zeeshanakram3:
> Still no way to fetch only top-level comments I think? (unless Hydra allows to somehow filter by field being null which I'm not aware of)

Won't the following query achieve the purpose (I tested this & it works)
```graphql
query {
  comments(where: {parentComment : {id_eq: null}}) {
    id
  }
}
```

> It seems there is no way to fetch the original comment once it's been edited? I can query specific edits but the original text is lost? Or maybe we don't know how to get it?

You can get the original text of the comment querying [CommentCreatedEvent](https://github.com/zeeshanakram3/joystream/blob/7fe3f5bb1431b6e03e1ed9c943bf1f6e70abfbd7/query-node/schemas/contentEvents.graphql#L61) entity (there is a [text](https://github.com/zeeshanakram3/joystream/blob/7fe3f5bb1431b6e03e1ed9c943bf1f6e70abfbd7/query-node/schemas/contentEvents.graphql#L85) field). Is this fine or do we need to query the original text through `comment` query

> Point 11. from my comment above is still unresolved I think?

Yes, not really found a way yet.

kdembler:
Okay I had no idea `id_eq: null` is doable, that should solve the issue then 👍 

>You can get the original text of the comment querying [CommentCreatedEvent](https://github.com/zeeshanakram3/joystream/blob/7fe3f5bb1431b6e03e1ed9c943bf1f6e70abfbd7/query-node/schemas/contentEvents.graphql#L61) entity (there is a [text](https://github.com/zeeshanakram3/joystream/blob/7fe3f5bb1431b6e03e1ed9c943bf1f6e70abfbd7/query-node/schemas/contentEvents.graphql#L85) field). Is this fine or do we need to query the original text through comment query

Okay we will use this

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -0,0 +1,708 @@
...
+
+  // load previous reaction by member to the video (if any)
+  const previousReactionByMember = await store.get(VideoReaction, {
+    where: { id: videoReactionEntityId({ memberId, videoId: videoId.toString() }) },
+  })
+
+  // load video reaction count by reaction type (LIKE/UNLIKE)
+  const reactionsCountByReactionType = await getOrCreateVideoReactionsCountByReactionId(store, video, reactionResult)
+
+  if (previousReactionByMember) {
```
ondratra, line 265: Let's avoid [Hadouken code](https://external-preview.redd.it/6TTm4tS5pMCiq2KlgviubpGKt1QzvetOi2Z6Z3lB1Nc.jpg?auto=webp&s=c1df961d84c8d76048a93c1d261e425d3737d779) (excesive indent) here. Create helper functions and flatten the structure. Also, some parts like `reactionsCountByReactionType.count` incrementing can be generalized.

Instead of
```
if (previousReactionByMember) {
  if (reactionResult === previousReactionByMember.reaction) {
    ...
  } else {
    ...
  }
} else {
  ...
}
```
let's have something like
```
myHelperFunction(video, ...)
...

function myHelperFunction(video, ...) {
  if (previousReactionByMember) {
    myHelperFunction2(...)
    return
  }
  
  // !previousReactionByMember
  ...
}

```

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -0,0 +1,708 @@
...
+  // ensure member is not banned from channel
+  ensureMemberIsNotBannedFromChannel(channel, memberId.toString(), 'Cannot add comment')
+
+  // ensure video's comment section is enabled
+  ensureCommentSectionIsEnabled(video, 'Cannot add comment')
+
+  // if new comment is replying to some parent comment, 1. validate that comment existence,
+  //  2. set `parentComment` to the parent comment, otherwise set `parentComment` to undefined
+  const parentComment =
+    parentCommentId && parentCommentId !== '' ? await getComment(store, parentCommentId.toString()) : undefined
```
ondratra: Redundant expression
```suggestion
    parentCommentId ? await getComment(store, parentCommentId.toString()) : undefined
```

ondratra (changes requested):
The features look generally good. I've suggested a couple of changes, most of them related to code readability. 

In addition to that, tests are failing. https://github.com/Joystream/joystream/runs/6147194020?check_suite_focus=true

metadata-protobuf/proto/Metaprotocol.proto
zeeshanakram3: Since the fields in the serialized messages are identified by their field numbers, so as long as the field numbers are not conflicting the protobuf de/serialization will keep working fine. However, I have changed the field numbers to start from `1` to remove any sort of confusion.

query-node/mappings/src/content/channel.ts
zeeshanakram3: Done.

query-node/mappings/src/content/channel.ts
zeeshanakram3: Done.

query-node/mappings/src/content/channel.ts
zeeshanakram3: Done.

metadata-protobuf/doc/index.md
zeeshanakram3: The doc is generated by running `yarn generate-doc` in `metadata-protobuf` package. So will have to inspect the script to see the problem.

query-node/mappings/src/content/commentAndReaction.ts
zeeshanakram3: Done.

query-node/mappings/src/content/commentAndReaction.ts
zeeshanakram3: Yes

zeeshanakram3:
Also for a failing [test](https://github.com/Joystream/joystream/runs/6147194020?check_suite_focus=true#step:8:6810), there was an issue with the approach of the calculating index of **_event_** in the block. Previously, the index was calculated based on the event content. However, if two events had the same content (one use case would be a member reacting to a video/comment two times with the same reaction in a given block; equivalent to reacting, and then unreacting), then we would get the same index for both events, which obviously caused the issue.

### Fix

For correct calculation, find the index of the event in the event list based on the whole event(metadata+content), and not just based on content. 

fix commit -> https://github.com/Joystream/joystream/pull/3340/commits/d5453bfce772a95a1008eb3282749a4d34634d16

ondratra (approved):
LGTM. Many commits that were initially included in this PR have already been merged in #3726, but my previous change requests seem to be solved.

Issue #3289, opened 2022-02-23
Idea: User selects buckets


bedeho:
There is a lot of complexity around trying to do random sampling of both storage and distribution buckets, and it also leads to more expensive interactions with the storage system. What if the user would provide the buckets to use, and the runtime simply verified that the selection was compatible with current constraints? For example that there is sufficient redundancy in storage buckets.

I can't see any serious issue security risks of doing this, and it would open up the opportunity for clients to exercise some judgement about what storage hosts act as good liasons.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201871056332439) by [Unito](https://www.unito.io)

bwhm:
@bedeho 

For storage: 
- Are we talking about new bags only, or for adding data to existing bags?
- Are we talking about which bucket you want upload to only or which bucket(s) will hold your bag? 

There are many potential pitfalls here, so let's see if we can narrow it down first...

Do you want this to apply to distribution as well?

bedeho:
> Are we talking about new bags only, or for adding data to existing bags?

New only.

> Are we talking about which bucket you want upload to only or which bucket(s) will hold your bag?

This is already the case, no change, user already picks this.

> Do you want this to apply to distribution as well?

Yes.

bwhm:
As discussed briefly in person:

### Storage
There are at least two clear constraints you'll have to adhere two:
- the bucket status of all the buckets you select must allow new bags
- the `StorageBucketsPerBagLimit`, ie. the maximum number of buckets a single bag can be in, is not violated

Depending on how this is implemented, the dynamic bag policy **may** also be a constraint here. By that, I mean it's not clear to me whether this would count as a dynamic bag policy, in the sense that it isn't the runtime assigning this, but the users themselves. This can be solved in one of (at least) two ways.

1. Selecting buckets yourself is constrained by the dynamic bag policy:
   - Suppose the dynamic bag policy is `n`, and the user selects `m` buckets.
   - The user selects `m = n` buckets themselves, and all is well.
   - The user selects `m < n` buckets, and the runtime selects the remaining `n-m`.
[...]

PR #3348, opened 2022-03-03
Audit3. Convert random buckets selection to user defined buckets.

shamil-gadelshin:
We remove a random selection of the storage and distribution buckets on dynamic bags creation from the storage pallet and replace it with user-defined buckets.

#### Changes
- removed a random selection of the buckets
- added new parameters (buckets) to the bag creation methods
- modified extrinsics for dynamic bag creation policy modifications (enforced new checks for parameters)
- fixed tests for storage and content pallets.

[Original issue](https://github.com/Joystream/joystream/issues/3289)


shamil-gadelshin:
We need to update the `audit3` branch with the latest `olympia` code in order to review this PR properly.

runtime-modules/storage/src/lib.rs
```diff
@@ -2978,6 +2958,25 @@ impl<T: Trait> Module<T> {
             Self::check_buckets_for_overflow(storage_buckets, &bag_change.voucher_update)?;
         }
 
+        // Validate distribution bucket IDs
```
dobertRowneySr: This has become a long function now with several sub validation routines. Can you please refactor each unit with self contained validation logic (like distribution bucket validation) into its own separate function. I think it is clearer that way

dobertRowneySr (changes requested):
It's ok. No major issue found. I just left a small request in order to enhance code clarity

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Fixed here: https://github.com/Joystream/joystream/pull/3348/commits/32d252427ebd6c3b417662badd523e572ab3cbc4

dobertRowneySr (approved):
LGTM

PR #3350, opened 2022-03-03
Olympia: Disable nft tests

mnaamani:
Disable nft tests (they fail now because we disabled them in the runtime) from full scenario to be re-enabled in rhodes release.


Lezek123 (approved):
LGTM!

Issue #3285, opened 2022-02-22
Tweak NFT open auction featuers


bedeho:
Specifically

1. Allow multiple simultaneous bids, but at most one per member.
1. Allow a member to update the amount associated with their bid, either up or down. **Reducing the amount can only be done when the bid locking period period has passed, however increasing the amount can be done at any time.**
1. The owner can select _any_ active bid as the winner, regardless of the amount associated with the bid. Remember that when owner is accepting a bid, this acceptance must commit to the current bid amount, to avoid front-running type attack where the winning bidder races to reduce the bid amount.
1. It must be possible to cancel any non-winning bid even after the auction has been settled, this is how the losers get their money back. Bid locking period still applies to cancellations.
1. Drop minimal bid step for open auction.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201900104475634) by [Unito](https://www.unito.io)

kdembler:
Quick thought that popped into my head wrt those changes. Previously, when new bid was placed we would [validate it](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/nft/types.rs#L252) to make sure `(NEW_BID - LAST_BID) > MIN_BID_STEP`. However, I think this logic doesn't really apply anymore? Since we said we want members to be able to freely increase/decrease their bids in an open spirit, do we need any validation at all, other than making sure bid is bigger than min price? I would say no, consequence of that would be that `minimalBidStep` is unused in context of an open auction. Would it make sense to say this `minimalBidStep` exists only for an english auction then?

bedeho:
Yes, this is not even optional, its a must if we are going to allow arbitrary bid amount changes, like reductions.

kdembler:
I'm not fully sure I understand your comment, are you saying that you agree we move `minimalBidStep` to English auction details and remove it in context of an open auction?

bedeho:
# Yes <3

kdembler:
How do we handle withdrawing from past auctions? It's possible that since I placed my bid, the auction have already ended and a new one has started. Will we need some way to identify an auction when requesting bid withdrawal?

bedeho:
> How do we handle withdrawing from past auctions? It's possible that since I placed my bid, the auction have already ended and a new one has started. Will we need some way to identify an auction when requesting bid withdrawal?

[...]

PR #3353, opened 2022-03-04
Rhodes fix open auction tweak

dobertRowneySr:
fixes https://github.com/Joystream/joystream/issues/3285.

# Changes
- Auction record removed from `Nft::transactional_status` and added to on-chain storage
- `BidByAuctionByMember: AuctionId x MemberId => Bid` added to on-chain storage
- test added for each https://github.com/Joystream/joystream/issues/3285#issue-1147027460
- english auction now has an optional `top_bidder` fields which tracks the member having the highest bid so far
- Auction are added to storage with `start_nft_auction` and removed upon completion


runtime-modules/content/src/lib.rs
```diff
@@ -189,6 +200,9 @@ decl_storage! {
 
         /// Max nft auction whitelist length
         pub MaxAuctionWhiteListLength get(fn max_auction_whitelist_length) config(): MaxNumber;
+
+        /// Auction By Id map
+        pub AuctionById get(fn auction_by_id): map hasher(blake2_128_concat) T::AuctionId => Auction<T>;
```
bedeho: I don't see any reason to lift `Auction` instances out of the channels (really inside `TransactionalStatusRecord`), it just creates many new invariants rather than using the type system.

runtime-modules/content/src/lib.rs
```diff
@@ -137,6 +143,11 @@ decl_storage! {
         pub VideoPostById get(fn video_post_by_id) : double_map hasher(blake2_128_concat) T::VideoId,
         hasher(blake2_128_concat) T::VideoPostId => VideoPost<T>;
 
+        /// Map in order to keep track of users bids
+        pub BidByAuctionByMember get(fn bid_by_auction_by_member) :
```
bedeho: There can only ever be _one_ bid from a given member on a given channel, as agreed here

https://github.com/Joystream/joystream/issues/3285#issuecomment-1049897515

This means for the current map could in practice that for all `m` and `b` there would be a single `a` mapped `(a,m) -> b`.

Lets clean this up and directly do  `T::VideoId x MemberId -> Bid` , and make sure the `Bid`. This captures the fact that for a given video and member, there could only ever be one bid held by the chain.

**In fact, this representation is not optional, it is required, in order to allow the runtime to replace existing bids from the same memberxvideo**

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,160 +154,237 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
+    MemberId: Default + Ord + Copy,
+    VideoId: Default + Copy,
 > {
     pub starting_price: Balance,
     pub buy_now_price: Option<Balance>,
     /// Auction type (either english or open)
-    pub auction_type: AuctionType<BlockNumber>,
-    pub minimal_bid_step: Balance,
-    pub last_bid: Option<Bid<MemberId, AccountId, BlockNumber, Balance>>,
+    pub auction_type: AuctionType<BlockNumber, Balance, MemberId>,
```
bedeho: We also need to store what `auction_id` an auction has, given that we now do not store `AuctionRecord` in a storage map. Critically, this field should only be stored when there is an open auction, as only those need these Ids, this means you somehow need to package this with  `AuctionType::Open`. However, there is a problem, because this type is also used in user provided inputs, and the user is not supposed to provide auction ids. It seems there are two solutions

1) make some new types that are almost identical, which can be used for providing input by users
2) have the runtime automatically pick open auction ids itself, like using the block time! I suggest this above anyway.

Make sure you fully understand the nuances of this issue, as it is a bit complex, but important.

runtime-modules/content/src/lib.rs
```diff
@@ -129,6 +132,9 @@ decl_storage! {
 
         pub NextVideoId get(fn next_video_id) config(): T::VideoId;
 
+        /// Auction Nonce
+        pub NextAuctionId get(fn next_auction_id) config(): T::AuctionId;
```
bedeho: First off, we don't actually need ids for all auctions, we need unique nonces specifically for _open auctions_. So at the very least, a renaming is needed here, but I actually suggest below - related to other more complex type issues about `AuctionRecord`, to just have runtime use block height as nonce. Its not a problem that different auctions have same nonce, we only need a unique nonce _per video_.

You can hypotheically imagine a weird scenario where an auction is opened, a bid is made, a new auction is opened, and then the old bid is accepted for the new auction, all in the same block! 🤯 I find this highly highly unlikely ot be possible.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,160 +154,237 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
     pub starts_at: BlockNumber,
     pub whitelist: BTreeSet<MemberId>,
+    pub nft_video_id: VideoId,
```
bedeho: This obviosuly goes away.

runtime-modules/content/src/nft/types.rs
bedeho: Looking closer tehre are obviously multiple fields here that are coverd by this, like 

`last_bid: Option<Bid<MemberId, AccountId, BlockNumber, Balance>>,`

This now only is a relevant state variable if we are in an English auction, you would not store this last bid here in an open auction, and vice versa. See comments below about segmenting out the two different bid types for each auction.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,160 +154,237 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
+    <T as common::MembershipTypes>::MemberId,
+    CurrencyOf<T>,
+    <T as Trait>::VideoId,
+>;
+
+// Auction implementation
+/// Information on the auction being created.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct Bid<BlockNumber: BaseArithmetic + Copy + Default, Balance> {
```
bedeho: Should we actually split this into two types?

- `OpenAuctionBid`:
  - you do need a `auction_id` field, which connects the bid to a given auction, so that when owner is accepting bids, it has to be for the currently active auction.
- `EnglishAuctionBid`
  - you don't need the `made_at_block` I think.
  - you don't need to store it in the storage map `BidByAuctionByMember`, because you only need to store the last best bid inlined, like `last_bid` now, etc.
  
There is probably more here, but you get the general idea. Things are now working so differently across the two auction types that we presumably would raise the safety and clarity of our code a lot by embracing that.

runtime-modules/content/src/lib.rs
bedeho: See my comments below about this map only being for open auction bids, so it must be renamed.

bedeho (changes requested):
Some very deep changes are needed to the approach, so I am not looking at any details in the extrinsics here, just the types and storage representation.

runtime-modules/content/src/nft/types.rs
dobertRowneySr: How about this:
#### Conditions to be met 
1. `Alice`: Member can *edit* their outstanding bid when nft transactional status is in:
    a. `Auction` state  
    b. `Auction` is `OPEN` type  
    c. open auction `Auction` is ongoing  
2. `Bob`:NFT owner can select winning `bid` when for nft
    a. nft status is `Auction`
    b. `Auction` is `OPEN` type
    c. `Auction` is ongoing

#### Solution
What I am thinking is 
1. Have two structs for the different types of auction
``` rust
struct EnglishAuctionRecord {
starting_price: Balance,
buy_now_price: Option<Balance>, 
starts_at: BlockNumber, 
whitelist: BTreeSet<MemberId>,
top_bid: EnglishBid,
bid_step: Balance,
duration: BlockNumber,
extension: BlockNumber,
}
struct OpenAuctionRecord {
  bid_lock_duratino: BlockNumber,
  auction_id: OpenAuctionId,
  buy_now_price: Option<Balance>,
  white_list: BTreeSet<MemberId>,
}
```
3. change `TransactionalStatus` to:
``` rust
enum TransactionalStatus {
  ...
  OpenAuction(OpenAuctionRecord),
  EnglishAuction(EnglishAuctionRecord),
}
```
4. Update `VideoRecord`
``` rust
struct VideoRecord {
...
last_open_auction_id: Option<AuctionId>,
...
}
```
Now when `Bob` issues a new open auction for the video `V` he will use `auction_id = video.last_open_auction_id.unwrap_or_default() + 1` as nonce.
- --> `OpenAuctionId` range in `[1, ..)`

6. Differentiate also bid types for the two auctions
``` rust
struct OpenAuctionBid {
   amount: Balance,
   auction_id: AuctionId,
   made_at_block: BlockNumber,
}

struct EnglishAuctionBid {
   amount: Balance,
   bidder: MemberId
}
```

Now Alice wants to update his bid for video nft `V`: she calls `make_bid(origin, Alice, V, amount)`:
``` rust
// make_bid
if let TransactionalStatus::OpenAuction(open_a) = nft.transactional_status {
   if let Ok(previous_bid) = Self::ensure_bid_exists(V, Alice) {
       if bid.auction_id == open_a.auction_id { 
          // Alice wants to update bid for ongoing auction => Standard case
       } else { 
         // new auction has been issued since last bid so update the bid regardless of the locking period
      }
   }
}
```
Bob wants to select Alice's bid (amount = `amount`) as open auction winner: he calls `pick_open_auction_winner(origin, ContentActor::Member(Bob), Alice, V, amount)`:
``` rust
let latest_oa_id = Self::ensure_latest_oa_id_for_video(V)?; 
let alice_bid = Self::ensure_bid_exists(V, Alice)?;
if latest_oa_id == alice_bid.auction_id {
// case when alice offered the winning bid for the latest open auction issued for V
// complete payment and remove the bid from the `VideoId x MemberId` map
} else {
// alice has an open auction bid for V but it is not related to the latest open auction
}
}
```
If Alice wants to cancel her bid for video `V`, she calls `cancel_bid(origin, Alice, V)`:
``` rust
let alice_bid = Self::ensure_bid_exists(V, Alice)?;
let nft = Self::ensure_video_exists(V).and_then(|video| video.ensure_nft_issued())?;
if  let TransactionalStatus::OpenAuction(open_a) = nft.transactional_status {
  if open_a.auction_id == alice_bid.auction_id {
     // ensure lock duration before cancelling ongoing auction bid
  }
} else {
// bid is for a NON ongoing open auction -> cancel bid regardless
}
}
```

bedeho: Looks like a step in the right direction, here is my summary from our call

- drop `EnglishAuctionRecord::started_at / duration`, and replace by a new single variable `ends_at`, which is set to `block_number + duration` when auction is created, and increased if there is a sniping bid, by `extension`.
- splitting auction entirely is a good improvement.
- `OpenAuctionBid::made_at_block` is not needed.
- `EnglishAuctionRecord::top_bid` must have type `Option<EnglishBid>`
- `VideoRecord::last_open_auction_id: Option<AuctionId>` simplify to `VideoRecord::next_open_auction_id: AuctionId`
- add `OpenAuctionRecord::starting_price` as a lower bound
- `OpenAuctionRecord::white_list` -> `whitelist`
- `EnglishAuctionRecord::bid_step` -> `min_bid_step`

dobertRowneySr:
I followed almost all the details we discussed.
I decided to keep the `AuctionRecord` with common trait so that I don't have to write duplicate functions that operate on common members only.
The runtime + tests are ready to be reviewed, after approval I will solve the conflicts by re-generating `@joystream/types`

runtime-modules/content/src/lib.rs
```diff
@@ -189,6 +192,11 @@ decl_storage! {
 
         /// Max nft auction whitelist length
         pub MaxAuctionWhiteListLength get(fn max_auction_whitelist_length) config(): MaxNumber;
+
+        /// Bids for open auctions
+        pub BidByVideoByMember get(fn bid_by_video_by_member):
```
bedeho: `BidByVideoByMember` -> `BidByVideoAndMember`

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +68,160 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
 #[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
-pub struct AuctionRecord<
-    BlockNumber: BaseArithmetic + Copy,
-    Balance: Clone,
-    MemberId: Ord + Clone,
-    AccountId: Ord + Clone,
-> {
+pub struct AuctionParamsRecord<AuctionType, Balance, MemberId: Ord> {
+    // Auction type (either english or open)
+    pub auction_type: AuctionType,
```
bedeho: I don't love this approach too be honest. Its now become much harder to cleanly figure out what the type hierarchy actually looks like, and all of the naming is not longer making that much sense. I have spent >20mins now just trying to track things down and understanding all the different meanings of  `auction_type`and `AuctionRecord` vs. `AuctionParamRecord`, which by the way are _literally_ identical!

It seems wrong to me that you are basically introducing a trait parameter which _will only ever be instantiated by one type_. This smells of incorrect usage of the abstractions, the other trait parameters here, like `Balance` or `MemberId`, they could in principle be lots of wacky types when this pallet is used, but `AuctionType` can only ever be one type in any usage of this code.

I am not exactly sure what you are trying to to achieve from your description, but I don't think this is the right way, can you please change it to the direct approach. If you are adamant about there being some serious problem with that approach, then let us do a call.

bedeho (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: fixed in 7b7aa6d34c28d0521ffba79a4ba8321bad00f7c2

runtime-modules/content/src/nft/types.rs
dobertRowneySr: I removed the duplicated type in 4f26837b202f027efed81d26adc45295a8f5ee62
I see the problem you are pointing at. The only draw back I see in changing back to the original type is the following:
``` rust
/// Auction Record alias
pub type Auction<T> =
    AuctionRecord<CurrencyOf<T>, <T as common::MembershipTypes>::MemberId, AuctionTypeOf<T>>;

/// AuctionParameters used for Auction initialization, it is almost a copycat of Auction<T>
/// Except that AuctionTypeOf<T> has been replaced because auction_id member is unknown @ creation
pub type AuctionParams<T> = AuctionRecord<
    CurrencyOf<T>,
    <T as common::MembershipTypes>::MemberId,
    AuctionType<EnglishAuction<T>, <T as frame_system::Trait>::BlockNumber>,
>;
```
While this usage might seem horrendous it does the job, because `auction_id` type inside the `OpenAuction<T>` record cannot be known before creation. And it is needed in order to differentiate between:
 cancelling bid from an ongoing open auction vs. cancelling bid from a past expired open auction.

bedeho: I did not understand this post, what is the draw back of changing back? We are already all over the runtime distingushing between a type and the a type to create that type, auctions are not special.

dobertRowneySr: The main problem I have in splitting the 2 auction is situations like the following. I will have to duplicate functions for english / opens auction such as `ensure_auction_can_be_canceled` -> `ensure_open_auction_can_be_canceled`, `ensure_english_auction_can_be_canceled`
``` rust
            // Ensure auction for given video id exists
            let auction = Self::ensure_auction_state(&nft)?;

            // Ensure given auction can be canceled
            Self::ensure_auction_can_be_canceled(&auction, video_id)?;
```
This does not seems sustainable to me, for example if tomorrow we want to add an extra type of auction this will force to add extra function: `can_ancel_new_auction_variant`

bedeho: Well, if you want to keep the extrinsics lexical scope as simple as possible, you could just leave it as is, but then inside of `ensure_auction_can_be_canceled` you can have distinct calls for each case. If you are saying that even this is a problem because of new cases in the future, then I would have to disagree.

dobertRowneySr: So for example if I split the 2 auctions I would basically use AuctionTypeOf in order to differentiate between the 2 auction types, this means also having:
```rust
    pub(crate) fn new_auction_from_params(
        params: &AuctionParams<T>,
        new_nonce: <T as Trait>::OpenAuctionId,
    ) -> AuctionTypeOf<T>; 

```
?

bedeho: No `AuctionType` is not really a sensible type as is. 

I would suggest you do one of two things,
a) totally split up extrincs and types for the two auctions
b) introduce new enum types for auction state and auction creation parameters, across the two parameters. I am not sure this is really worth it, but since you are thinking more in these terms, this seems like a way to do it which does not have the same problems as the prior approach

```Rust

pub enum TransactionalStatusRecord<MemberId, Balance, AuctionType> {
    Idle,
    InitiatedOfferToMember(MemberId, Option<Balance>),
    Auction(Auction),
    BuyNow(Balance),
}

pub enum Auction {
  EnglishAuction(EnglishAuction),
  OpenAuction(OpenAuction)
}

pub struct EnglishAuction {
  /// state of an english auction
} 

pub struct OpenAuction {
 /// state of an open auction
}

pub enum AuctionParams<BlockNumber, Balance, MemberId: Ord> {
  EnglishAuctionParams(EnglishAuctionParams),
  OpenAuctionParams(OpenAuctionParams)
}

pub struct EnglishAuctionParams {
  //// creation info
}

pub struct OpenAuctionParams {
   //// creation info
}
```

Now you can do

```Rust
pub(crate) fn new_auction_from_params(
        params: &AuctionParams<T>,
        new_nonce: <T as Trait>::OpenAuctionId,
    ) -> Auction; 
```

There will be some common fields in play here, and if you don't like that you can solve it with some static composition, like

```Rust
pub struct BaseAuctionProperties {
    pub starting_price: Balance,
    pub buy_now_price: Option<Balance>,
    pub whitelist: BTreeSet<MemberId>,
}

pub struct EnglishAuction {
 base_properties: BaseAuctionProperties
 ...
 }
```

but not sure how much value it really has. 

I would say is that the main problem with this entire approach is that its trying to hang on to the idea that open and english auctions are virtually indistinguishable, when that is basically no longer the case.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,338 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
-        let last_bid = self.ensure_last_bid_exists::<T>()?;
+        self.ensure_auction_has_no_bids::<T>(video_id)
+    }
+}
 
-        // ensure caller is last bidder.
-        self.ensure_caller_is_last_bidder::<T>(who)?;
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
+pub struct OpenBidRecord<Balance, BlockNumber, AuctionId> {
```
bedeho: `OpenAuctionBidRecord`

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,338 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
-    English(EnglishAuctionDetails<BlockNumber>),
-    // Open auction details
-    Open(OpenAuctionDetails<BlockNumber>),
-}
-
-impl<BlockNumber: Default> Default for AuctionType<BlockNumber> {
-    fn default() -> Self {
-        Self::English(EnglishAuctionDetails::default())
-    }
+pub struct EnglishBid<Balance, MemberId> {
```
bedeho: `EnglishAuctionBid`

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,338 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
-    }
-}
+pub type OpenAuctionParams<T> = OpenAuctionParamsRecord<
+    <T as frame_system::Trait>::BlockNumber,
+    CurrencyOf<T>,
+    <T as common::MembershipTypes>::MemberId,
+>;
 
-pub type InitTransactionalStatus<T> = InitTransactionalStatusRecord<
+pub type OpenBid<T> = OpenBidRecord<
```
bedeho: `OpenAuctionBid`

runtime-modules/content/src/lib.rs
```diff
@@ -1636,39 +1756,33 @@ decl_module! {
...
-            let bid = auction.ensure_last_bid_exists::<T>()?;
-
-            // Ensure auction type is set to `English`
-            auction.ensure_is_english_auction::<T>()?;
-
-            // Ensure auction can be completed
-            Self::ensure_auction_can_be_completed(&auction)?;
-
-            let owner_account_id = Self::ensure_owner_account_id(&video, &nft).ok();
+            // Ensure nft & english auction validity for nft exists, retrieve top bid
```
bedeho, line 1836: This section has problems
1. Overall clarity is significantly lower, you have had to throw away statically guaranteed intermediate entities and comments are gone.
2. You have introduced a panicable unwrap which is only avoided by runtime invariants.

Please revert back to the old style: one thing at a time, producing rich static artefacts, with suitable comment.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -358,28 +321,174 @@ impl<T: Trait> Module<T> {
...
+        video_id: T::VideoId,
+        winner_id: T::MemberId,
+        amount: CurrencyOf<T>,
+    ) -> Result<Nft<T>, DispatchError> {
+        let Video::<T> {
+            in_channel,
+            nft_status,
+            ..
+        } = Self::video_by_id(video_id);
+        let nft = nft_status.unwrap();
```
bedeho: Fix.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -358,28 +321,174 @@ impl<T: Trait> Module<T> {
...
-        let bidder_account_id = last_bid.bidder_account_id;
+        video_id: T::VideoId,
+        winner_id: T::MemberId,
+        amount: CurrencyOf<T>,
+    ) -> Result<Nft<T>, DispatchError> {
+        let Video::<T> {
+            in_channel,
+            nft_status,
+            ..
+        } = Self::video_by_id(video_id);
```
bedeho: This lookup is not statically safe & clear, please fix all other cases where you did something similar.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -358,28 +321,174 @@ impl<T: Trait> Module<T> {
...
+        let nft = nft_status.unwrap();
+
+        let dest_account_id = Self::ensure_owner_account_id(in_channel, &nft)?;
```
bedeho: These fallible calls after mutation safe is not ok, solved just like above.

runtime-modules/content/src/lib.rs
```diff
@@ -1487,104 +1558,162 @@ decl_module! {
...
+            Self::ensure_has_sufficient_balance(&participant_account_id, bid_amount)?;
 
             // Ensure nft is already issued
-            let nft = video.ensure_nft_is_issued::<T>()?;
+            let nft = Self::ensure_nft_exists(video_id)?;
 
-            // Ensure auction for given video id exists
-            let auction = nft.ensure_auction_state::<T>()?;
+            // Validate parameters & return english auction
+            let auction =  Self::ensure_open_auction_state(&nft)
```
bedeho: Same here.

runtime-modules/content/src/lib.rs
```diff
@@ -1487,104 +1558,162 @@ decl_module! {
...
 
-                    let nft = Self::complete_auction(video.in_channel, nft, bid, funds_destination_account_id);
-                    let video = video.set_nft_status(nft);
+            // Ensure nft is already issued
+            let nft = Self::ensure_nft_exists(video_id)?;
 
-                    // Update the video
-                    VideoById::<T>::insert(video_id, video);
+            // Validate parameters & return english auction
+            let auction =  Self::ensure_english_auction_state(&nft)
```
bedeho: Same here.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -358,28 +321,174 @@ impl<T: Trait> Module<T> {
...
+
+    /// Ensure bid can be cancelled
+    pub(crate) fn ensure_bid_can_be_canceled(
+        who: T::MemberId,
+        nft_video_id: T::VideoId,
+    ) -> DispatchResult {
+        let nft = Self::ensure_nft_exists(nft_video_id)?;
+        let bid = Self::ensure_open_bid_exists(nft_video_id, who)?;
+
+        Self::ensure_open_auction_state(&nft).map_or(Ok(()), |open| {
```
bedeho: Even flatten this I would say.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,338 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
 
-    /// Ensure auction have been already started
-    pub fn ensure_auction_started<T: Trait>(&self, current_block: BlockNumber) -> DispatchResult {
-        ensure!(
-            self.starts_at <= current_block,
-            Error::<T>::AuctionDidNotStart
-        );
-        Ok(())
+    pub(crate) fn ensure_auction_has_valid_bids<T: Trait>(&self) -> DispatchResult {
+        self.ensure_auction_has_no_bids::<T>()
```
bedeho: Isn't this reversed? valid bids means you must ensure there are bids?

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,338 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
+        video_id: T::VideoId,
     ) -> DispatchResult {
-        if let AuctionType::Open(OpenAuctionDetails { bid_lock_duration }) = &self.auction_type {
-            ensure!(
-                current_block - bid.made_at_block >= *bid_lock_duration,
-                Error::<T>::BidLockDurationIsNotExpired
-            );
-        }
+        ensure!(
+            crate::BidByVideoAndMember::<T>::iter_prefix(video_id)
```
bedeho: This is not workable, not safe, not needed.

runtime-modules/content/src/lib.rs
```diff
@@ -189,6 +192,11 @@ decl_storage! {
 
         /// Max nft auction whitelist length
         pub MaxAuctionWhiteListLength get(fn max_auction_whitelist_length) config(): MaxNumber;
+
+        /// Bids for open auctions
+        pub BidByVideoAndMember get(fn bid_by_video_and_member):
```
bedeho: `OpenAuctionBidByVideoAndMember`

runtime-modules/content/src/lib.rs
```diff
@@ -1417,21 +1460,51 @@ decl_module! {
...
 
-            // Ensure given auction can be canceled
-            auction.ensure_auction_can_be_canceled::<T>()?;
+            // Trigger event
+            Self::deposit_event(RawEvent::AuctionCanceled(owner_id, video_id));
+        }
+
+        // Cancel video nft english auction
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn cancel_open_auction(
```
bedeho, line 1481: This is incorrectly blocked by not having bids... can be done at any time.

bedeho (changes requested):

runtime-modules/content/src/nft/types.rs
dobertRowneySr: renaming done in 6f3f47bc3dc229e7af3d3b00f4df7f68e94c4cdd

runtime-modules/content/src/nft/types.rs
dobertRowneySr: renaming done in 6f3f47bc3dc229e7af3d3b00f4df7f68e94c4cdd

runtime-modules/content/src/nft/types.rs
dobertRowneySr: renaming done in 6f3f47bc3dc229e7af3d3b00f4df7f68e94c4cdd

runtime-modules/content/src/lib.rs
dobertRowneySr: I restored the previous design starting @ 58037c2ec42956075f1403e89261828a8b20348c

runtime-modules/content/src/lib.rs
dobertRowneySr: I restored the previous design starting @ 58037c2ec42956075f1403e89261828a8b20348c

runtime-modules/content/src/lib.rs
dobertRowneySr: I restored the previous design starting @ 58037c2ec42956075f1403e89261828a8b20348c

runtime-modules/content/src/nft/types.rs
dobertRowneySr: Removed here e256eba1394bd5ce6fe5bde99645c90cde5b36e8

runtime-modules/content/src/nft/types.rs
dobertRowneySr: Removed here e256eba1394bd5ce6fe5bde99645c90cde5b36e8

runtime-modules/content/src/lib.rs
dobertRowneySr: Removed here e256eba1394bd5ce6fe5bde99645c90cde5b36e8

runtime-modules/content/src/lib.rs
dobertRowneySr: renaming done in 6f3f47bc3dc229e7af3d3b00f4df7f68e94c4cdd

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: I refactored `complete_auction` in 3fc8c8517f772e08cabdb320c987a276b7e4d1e2

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: I refactored `complete_auction` in 3fc8c8517f772e08cabdb320c987a276b7e4d1e2

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: I refactored `complete_auction` in 3fc8c8517f772e08cabdb320c987a276b7e4d1e2

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: refactored function in 58037c2ec42956075f1403e89261828a8b20348c

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: I refactored `complete_auction` in 3fc8c8517f772e08cabdb320c987a276b7e4d1e2

runtime-modules/content/src/lib.rs
```diff
@@ -1416,21 +1459,50 @@ decl_module! {
             // Authorize nft owner
             ensure_actor_authorized_to_manage_nft::<T>(origin, &owner_id, &nft.owner, video.in_channel)?;
 
-            // Ensure auction for given video id exists
-            let auction = nft.ensure_auction_state::<T>()?;
+            // Ensure auction state that can be canceled
+            Self::ensure_english_auction_state(&nft)
```
bedeho: `ensure_english_auction_state` => `ensure_in_english_auction_state`

runtime-modules/content/src/lib.rs
```diff
@@ -1416,21 +1459,50 @@ decl_module! {
...
+            // Ensure nft is already issued
+            let nft = video.ensure_nft_is_issued::<T>()?;
+
+            // Authorize nft owner
+            ensure_actor_authorized_to_manage_nft::<T>(origin, &owner_id, &nft.owner, video.in_channel)?;
 
-            // Ensure given auction can be canceled
-            auction.ensure_auction_can_be_canceled::<T>()?;
+            // Ensure auction state that can be canceled
+            Self::ensure_open_auction_state(&nft)?;
```
bedeho: `ensure_in_open_auction_state`

runtime-modules/content/src/nft/mod.rs
```diff
@@ -360,28 +323,107 @@ impl<T: Trait> Module<T> {
...
             true,
         );
 
-        nft.owner = NftOwner::Member(last_bidder);
-        nft.transactional_status = TransactionalStatus::<T>::Idle;
-        nft
+        Self::cancel_transaction(nft, Some(winner_id))
+    }
+
+    pub(crate) fn cancel_transaction(nft: &Nft<T>, winner: Option<T::MemberId>) -> Nft<T> {
```
bedeho: All but one case uses `cancel_transaction` to just make nft be idle status again, so `winner` is almost always None, only one case (in internal method `complete_auction`) is there a new owner also set. So can we just have a method `make_new_idle_nft(nft:)` which is used for these majority cases, because `cancel_transaction` is a bit too vague when being read, despite not needing to be.

runtime-modules/content/src/lib.rs
```diff
@@ -713,9 +720,9 @@ decl_module! {
                 .map_or(
                     Ok(None),
                     |issuance_params| {
-                        Some(Self::construct_owned_nft(issuance_params)).transpose()
+                        Some(Self::construct_owned_nft(issuance_params,video_id)).transpose()
```
bedeho, line 714: So just move this computation of `nft_status` above the upload check/sideeffect above, then the problem should be solved.

runtime-modules/content/src/lib.rs
```diff
@@ -1453,17 +1525,15 @@ decl_module! {
             ensure_actor_authorized_to_manage_nft::<T>(origin, &owner_id, &nft.owner, video.in_channel)?;
 
             // Ensure nft in pending offer state
-            nft.ensure_pending_offer_state::<T>()?;
+            Self::ensure_pending_offer_state(&nft)?;
```
bedeho: `ensure_in_pending_offer_state`

runtime-modules/content/src/lib.rs
```diff
@@ -1486,17 +1556,15 @@ decl_module! {
             ensure_actor_authorized_to_manage_nft::<T>(origin, &owner_id, &nft.owner, video.in_channel)?;
 
             // Ensure nft in buy now state
-            nft.ensure_buy_now_state::<T>()?;
+            Self::ensure_buy_now_state(&nft)?;
```
bedeho: `ensire_in_buy_now_state`

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
         #[weight = 10_000_000] // TODO: adjust weight
-        pub fn make_bid(
+        pub fn make_open_auction_bid(
             origin,
             participant_id: T::MemberId,
             video_id: T::VideoId,
-            bid: CurrencyOf<T>,
+            bid_amount: CurrencyOf<T>,
         ) {
+            let now = <frame_system::Module<T>>::block_number();
```
bedeho: Can you move this closer to where it is needed?

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
+            Self::ensure_has_sufficient_balance(&participant_account_id, bid_amount)?;
 
             // Ensure nft is already issued
+            let video = Self::ensure_video_exists(&video_id)?;
             let nft = video.ensure_nft_is_issued::<T>()?;
 
-            // Ensure auction for given video id exists
-            let auction = nft.ensure_auction_state::<T>()?;
+            // Validate parameters & return english auction
+            let open_auction =  Self::ensure_open_auction_state(&nft)?;
```
bedeho: `ensure_in_open_auction_state`

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
+                Some(buy_now_price) if bid_amount >= buy_now_price => {
+                    // complete auction @ buy_now_price
+                    let new_nft = Self::complete_auction(
+                        &nft,
+                        video.in_channel,
+                        participant_account_id,
+                        participant_id,
+                        buy_now_price,
+                    );
+
```
bedeho, line 1652: If there was an existing bid, which was superseded and won, should we not at least clean it up?

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
+                },
+                _ =>  {
+                    Self::ensure_open_bid_exists(video_id, participant_id)
+                        .map_or((), |bid| {
+                            T::Currency::unreserve(
+                                &participant_account_id,
+                                bid.amount
+                            );
+                        });
+
```
bedeho, line 1673: There is no reservation of balance on this new bid?

runtime-modules/content/src/lib.rs
```diff
@@ -1717,37 +1864,47 @@ decl_module! {
...
+            // remove bid
+            OpenAuctionBidByVideoAndMember::<T>::remove(video_id, winner_id);
 
             // Update the video
-            VideoById::<T>::insert(video_id, video);
+            VideoById::<T>::mutate(video_id, |v| v.set_nft_status(post_auction_nft));
 
             // Trigger event
             Self::deposit_event(RawEvent::OpenAuctionBidAccepted(owner_id, video_id));
```
kdembler: This now needs to identify which bid was accepted

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,406 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
-            // Open auction never expires
-            false
+impl<
+        BlockNumber: Copy + Saturating + PartialOrd,
+        AuctionId: Copy + PartialEq,
+        Balance: Copy + PartialOrd + Saturating,
+        MemberId: Ord + Copy,
+    > OpenAuctionRecord<BlockNumber, AuctionId, Balance, MemberId>
+{
+    pub(crate) fn ensure_can_make_bid<T: Trait>(
```
bedeho, line 272: So from what I gather, what this is doing is really specifically checking that the new bid amount is acceptable, so lets sharpen the name to something like `ensure_bid_amount_is_sufficient_now`. It would also be good if you can add some inline comments which explains what you are checking and why that is the right check given what you already checked.

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
+            let open_auction =  Self::ensure_open_auction_state(&nft)?;
 
-            let current_block = <frame_system::Module<T>>::block_number();
+            // check whitelisted participant
+            open_auction.ensure_whitelisted_participant::<T>(participant_id)?;
 
-            // Ensure nft auction not expired
-            auction.ensure_nft_auction_not_expired::<T>(current_block)?;
+            // ensure bid can be made
+            let old_bid = Self::ensure_open_bid_exists(video_id, participant_id).ok();
```
bedeho: Can we rename `old_bid` to `possible_old_bid`.

runtime-modules/content/src/lib.rs
```diff
@@ -1541,86 +1607,167 @@ decl_module! {
...
-                }
-                _ => {
-                    // Make auction bid & update auction data
+            // Validate parameters & return english auction
+            let eng_auction =  Self::ensure_english_auction_state(&nft)?;
+
+            // Ensure auction is not expired
+            eng_auction.ensure_auction_is_not_expired::<T>(now)?;
+
+            // Ensure top bid is present
```
bedeho: Comment is misleading, you are not ensuring this (ensure really means that it _has to hold, or else ...._).
Also, do you even need this variable? you are doing all this computation `eng_auction.ensure_top_bid_exists::<T>().ok()` only to get literally `eng_auction.top_bid`.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,406 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
+    ) -> Result<EnglishAuctionBid<Balance, MemberId>, DispatchError> {
+        self.top_bid
+            .to_owned()
+            .ok_or_else(|| Error::<T>::BidDoesNotExist.into())
     }
 
-    // Ensure auction has no bids
-    fn ensure_has_no_bids<T: Trait>(&self) -> DispatchResult {
-        ensure!(self.last_bid.is_none(), Error::<T>::ActionHasBidsAlready);
+    pub(crate) fn ensure_bid_can_be_made<T: Trait>(&self, amount: Balance) -> DispatchResult {
```
bedeho: Same here, this name should be sharpened to reflect that its really about enforcing sufficiency of amount.

runtime-modules/content/src/nft/types.rs
```diff
@@ -169,391 +83,406 @@ impl<MemberId> Default for NftOwner<MemberId> {
...
+        if let Some(buy_now) = &self.buy_now_price {
+            if amount > *buy_now {
+                return Ok(());
+            }
+        }
+
+        self.top_bid.as_ref().map_or_else(
```
bedeho: This is a bit subjective, but to me, a simple `if let Some(bid)  { .. }` with each ensure in sequence inside, would have been much nicer to read & process.

runtime-modules/content/src/lib.rs
```diff
@@ -1717,37 +1864,47 @@ decl_module! {
...
+            commit: CurrencyOf<T>, // amount the auctioner is committed to
         ) {
-
-            // Ensure given video exists
-            let video = Self::ensure_video_exists(&video_id)?;
+            let winner_account_id = T::MemberAuthenticator::controller_account_id(winner_id)?;
 
             // Ensure nft is already issued
-            let nft = video.ensure_nft_is_issued::<T>()?;
+            let nft = Self::ensure_nft_exists(video_id)?;
```
bedeho: Its more natural to just ensure the video exists, and then inspect the recovered video object to ensure there is an nft issued, this also requires one read rather than two, which is a tiny bit better.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -5,69 +5,36 @@ pub use types::*;
...
-    }
-
     /// Ensure auction participant has sufficient balance to make bid
     pub(crate) fn ensure_has_sufficient_balance(
         participant: &T::AccountId,
         bid: CurrencyOf<T>,
     ) -> DispatchResult {
         ensure!(
-            T::Currency::can_reserve(participant, bid),
+            T::Currency::total_balance(participant) >= bid,
```
bedeho: My best bet, while we are using reservations, is that this entire method is not needed, instead call https://docs.rs/frame-support/3.0.0/frame_support/traits/trait.ReservableCurrency.html#tymethod.can_reserve

You should double check!

bedeho (changes requested):
- A handful of naming improvements.
- There are instances of this problem https://github.com/Joystream/joystream/issues/3422, at least in `create_video`. I tried to find others, I didn't, but best to double check.

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/nft/types.rs
ignazio-bovo: :+1:

runtime-modules/content/src/nft/types.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/nft/types.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/nft/mod.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

runtime-modules/content/src/nft/mod.rs
ignazio-bovo: :+1:

runtime-modules/content/src/lib.rs
ignazio-bovo: :+1:

dobertRowneySr: Fixed that is why CI/query node checks are failing

bedeho (approved):
Looks good!

Issue #2760, opened 2021-10-14
 On-chain content dispute and resolution


bedeho:
# Question for Lezek123 
Does this proposal fully address this?
https://github.com/Joystream/joystream/issues/2911

# Background
It must be possible to have an orderly way to 

* allow curators remove content/videos: meaning its both gone from the chain, the storage and content delivery infrastructure, and query nodes.
* allow creators to contest
* allow for the dispute to be settled, on the record.

# Proposal
Here is a high level specification.

## Membership
> Assuming membership integration has happened.

* The membership lead can delete any data object from any dynamic _membership bag_ with instant effect, and only needs to provide a rationale. Deletion opens a linear discussion thread where any member can participate.
* Each membership worker has a boolean flag `can_delete_data_objects`, which allows the worker to delete any data objects from any membership bag, with a rationale. It opens up a linear discussion thread where any member can participate.

## Content Directory
* Any member, or curator, can open a complaint on any video or channel, which allows any member to participate. Has no effect on anything except alerting the relevant parties and starting the conversation. There can only be one such conversation simultaneously per entity.
* All channels are given _privilege level_, which is basically a non-negative integer identifier without any semantics. There is a default initial level for all channels, this default is updatable by the lead, and only the lead can update the level of a given channel.
* All curation groups have _permissions_, which is a (partial) map of permission level to _level permissions_, defined as a description of which of the following actions a curator belonging to a group which has permissions which for the privilege level of a given channel has the corresponding level permissions
  1. **DELETE_CHANNEL**: instantly delete a channel, with a corresponding list of data objects from the same channel bag.
  1. **DELETE_NFT_FREE_VIDEO**: instantly delete a video without an NFT, with a corresponding list of data objects from the same channel bag.
  1. **DELETE_ANY_VIDEO**: instantly delete any video, with a corresponding list of data objects from the same channel bag.
  1. **DELETE_OBJECT**: instantly delete a data objects from the channel bag.
  1. **PAUSE_CHANNEL_FUNCTION**: block, or unblock a subset of the following by channel owner
    * transfer of funds from channel account to any other account.
    * creator cashout
    * issue an NFT
    * create a video.
    * update videos.
    * update channel.
    * issue creator token.
       Critically, an invocation of this pause command does not need to block _all_ of these, it can be issued to only block a subset, as needed.
[...]

Issue #3084, opened 2022-01-23
Question: Significance of missing storage side-effects as content directory event payload


bedeho:
The content directory makes call to the storage system, however the side-effects of these calls, namely what data objects where created or dropped, or what bags were created or destroyed, and in what buckets, is not returned from the storage pallet, and thus also not part of content directory events.

This makes it harder to write correct mappings, as a single mapping cannot pickup everything needed to bind entities across content<-> storage boundary, you will need to do it across distinct mappings. In so far as this is correct, we should fix this.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202000874282720) by [Unito](https://www.unito.io)

Lezek123:
I think the only issue I faced when writing the mappings was w.r.t. to processing new assets created during channel / video creation/update, [where I had to make an assumption that `x` most recently created storage data objects correspond to the objects specified in `ChannelCreationParameters` / `VideoCreationParameters` / `ChannelUpdateParameters` / `VideoUpdateParameters`](https://github.com/Joystream/joystream/blob/e16937f1bb28a18630089472db84ea3107709fc7/query-node/mappings/content/utils.ts#L326), which due to some hydra limitations (ids are strings) required keeping track of the last inserted `dataObjectId` separately and making the processing logic a bit more complex and fragile in case of the runtime implementation changes.

bedeho:
Ok, I think this comment literally says what I was saying

`// FIXME: Ideally the runtime would provide object ids in ChannelCreated/VideoCreated/ChannelUpdated(...) events`

bedeho:
We need to discuss this and come up with an overall conclusion, but my current assumption is that the world would be much simpler for the query node mapping author if there were no events in utilty modules (i.e. modules only invoked inside the runtime), and instead have much richer events in the relevant subsystem (council, content, membership, etc.), where the relevance of a dataobjects is directly related to some high level concept like a proposal, membership, channel or video.

Lezek123:
I believe there was only one issue that the current content-storage integration caused for the query node, which was about no object ids included in `VideoCreated`, `VideoUpdated`, `ChannelCreated`, `ChannelUpdated` events (as mentioned before). This was fixed in https://github.com/Joystream/joystream/pull/3454:
- `ChannelCreated` already included `Channel` struct as part of event payload, so it got fixed without any changes in the actual event (since `Channel` struct now contains `BTreeSet<DataObjectId>`).
[...]

Issue #3071, opened 2022-01-22
Cleaner storage bag creation integration code


bedeho:
# Note
No need to change this for Giza, but seems like tech-debt to me, discovered while diagnosing

https://github.com/Joystream/joystream/issues/3069

# Background
We need a clear and crisp example in the runtime of how to plug into the storage system, as it needs to be used from other pallets in the future, but currently the variation we have in the content directory, specifically from `create_channel`,  seems rough to me. Two particularly severe issues is the presence of two error operators post mutation safe boundary:

* https://github.com/Joystream/joystream/blob/giza/runtime-modules/content/src/lib.rs#L763
* https://github.com/Joystream/joystream/blob/giza/runtime-modules/content/src/lib.rs#L769

# Proposal
1. More comments: there are large complex flows of checks and calls which have 0 explanation., in particular around bag creation.
1. Drop bag existence check: this is effectively is a runtime check of a broken chain, as the only way there could exist a bag for a new channel which exceeds the prior max id for channels, is if the content pallet or some other pallet in the chain, unwittingly already created it. If any thing, the _can_create_ call can subsume this by returning false if the bag already exists, and this is already the case it seems
https://github.com/Joystream/joystream/blob/giza/runtime-modules/storage/src/lib.rs#L2849
1. There is no reason to have two distinct calls `can_create_dynamic_bag_with_objects_constraints` and `can_create_dynamic_bag`, these are semantically indistinguishable, and infact the implementations clearly show how easy it would be to just have a single `can_create_dynamic_bag` which takes an additional `Option<UploadParameters<T>>`. Hence this should be fixed in the `storage` pallet as well.
1. Same for actual creation, so drop `create_dynamic_bag_with_objects_constraints`.
1. There should not be a second call to `uplaod_data_objects`, 
https://github.com/Joystream/joystream/blob/giza/runtime-modules/content/src/lib.rs#L769
the upload should be bundled with the bag creation, as it will have access to upload params. This probably will also open door for reducing # reads & writes and simplifying interface for storage pallet users, but more importantly, it is super clear what it means to create a bag including references to uploads yet to be initiated. Clearly some sort of `BagUpdate` is being computed based off these parameters, which means that vouchers and deletion prizes already reflect an impending upload. This begs the question of how this currently is not compounded by call to `upload_data_objects`, but in the end it takes too much work to get to the bottom of this.
1. In `can_create_dynamic_bag` signature, rename `deletion_prize` to `account_deletion_prize`.
[...]

PR #3366, opened 2022-03-07
Content directory: New permissions model

Lezek123:
Implementing content directory permissions model according to https://github.com/Joystream/joystream/issues/2760
Work in progress


runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -1,4 +1,48 @@
 use super::*;
+use sp_std::collections::btree_map::BTreeMap;
+#[cfg(feature = "std")]
+use strum_macros::EnumIter;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct ContentModerationPermissions {
```
bedeho: Would it perhaps be a more efficient representation to just have something like `BTreeSet<ContentModerationAction>`, where presence `a` <=> allowed to do action `a`. More importantly than the storage efficiency in storage or in extrinsics is that you get rid of the conversion layer you need in places like `can_perform_action`.

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -1,4 +1,48 @@
...
+        }
+    }
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, Debug)]
+pub enum ContentModerationAction {
+    HideVideo,
+    HideChannel,
+    PauseChannel,
```
bedeho: Here it was meant that pause controls were more granular, as the requirements alluded

> Critically, an invocation of this pause command does not need to block all of these, it can be issued to only block a subset, as needed.

So you may as well make life simple and just flatten all cases directly into `ContentModerationAction`.

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -13,6 +57,9 @@ where
 
     /// When `false`, curator in a given group is forbidden to act
     active: bool,
+
+    // Group's moderation permissions
+    permissions: ModerationPermissionsByLevel<T>,
```
bedeho: `permissions` => `level_permissions` or `permissions_by_level` ? what do you think

runtime-modules/content/src/lib.rs
```diff
@@ -553,6 +592,31 @@ decl_module! {
...
+            origin,
+            channel_id: T::ChannelId,
+            new_privilege_level: T::ChannelPrivilegeLevel,
+        ) {
+            let sender = ensure_signed(origin)?;
+
+            ensure_lead_auth_success::<T>(&sender)?;
+
+            // check that channel exists
+            let mut channel = Self::ensure_channel_validity(&channel_id)?;
```
bedeho: In general, I try to encourage everyone to avoid having mutables with any sort of longer life-time, as it just opens to the door to some bad mutation being introduced at some later time. This case is really mild, so not a big issue, but try to avoid when possible, like using ::mutate storage action, or constructing a new entity object using the old one combined with struct update syntax
https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

So here it would be

```Rust
let new_channel = Channel {
privilege_level: new_privilege_level
...old_channel
}
```

bedeho (changes requested):
Excellent work, minor tweaks only, looks like you are on a solid track.

bedeho:
Oh I forgot one thing in the prior review: remember to add some sort of size constraint on how many mappings can live in a map provided for a curator group, as it becomes a security issue if it is too big. You can just introduce a new constant for this value.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: I was considering that, but I noticed that I'm unable to use nested `BTreeSet` (ie. `BTreeMap<T, BTreeSet<T>>`) as an extrinsic argument, so I needed some other type to represent the permissions. Alternatively I could use `Vec`, but then I'd need to introduce additional validation to make sure each `ContentModerationAction` only existis there once, so I went for `Struct` instead.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: I'd opt for either keeping `permissions` or changing to `permissions_by_level`.
`level_permissions` seems confusing for me, because it sounds like it means _set of permissions for a given level_ instead of _map of permissions by level_.

Lezek123:
> Oh I forgot one thing in the prior review: remember to add some sort of size constraint on how many mappings can live in a map provided for a curator group, as it becomes a security issue if it is too big. You can just introduce a new constant for this value.

Because the key of the map is `u8` it can be no more than `255` currently, should we restrict it further?

bedeho:
> Because the key of the map is u8 it can be no more than 255 currently, should we restrict it further?

I don't know actually, but without an explicit limit it may be easy for someone to change the type for some other reason without keeping this concern in mind.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Renamed to `permissions_by_level`

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/4ef05ccc660879617d616332926e3cb3cd03082f

Lezek123:
> remember to add some sort of size constraint on how many mappings can live in a map provided for a curator group, as it becomes a security issue if it is too big. You can just introduce a new constant for this value.

Done in https://github.com/Joystream/joystream/pull/3366/commits/c84df57ab25527b09c911b906e679a55e88f69c6

runtime-modules/content/src/lib.rs
```diff
@@ -251,20 +268,27 @@ decl_module! {
...
+            is_active: bool,
+            permissions_by_level: ModerationPermissionsByLevel<T>
         ) {
 
             let sender = ensure_signed(origin)?;
             // Ensure given origin is lead
             ensure_lead_auth_success::<T>(&sender)?;
+            // Ensure permissions_by_level map max. allowed size is not exceeded
+            ensure!(
+                permissions_by_level.len() <= T::MaxKeysPerCuratorGroupPermissionsByLevelMap::get().into(),
```
bedeho: I think you use this a few places, probably will increase, I think its worth factoring out as a `ensure_permissions_by_level_small_enough`, making the inline extrinsic code a bit more readable.

runtime-modules/content/src/lib.rs
```diff
@@ -553,6 +609,30 @@ decl_module! {
...
+            origin,
+            channel_id: T::ChannelId,
+            new_privilege_level: T::ChannelPrivilegeLevel,
+        ) {
+            let sender = ensure_signed(origin)?;
+
+            ensure_lead_auth_success::<T>(&sender)?;
+
+            // check that channel exists
+            Self::ensure_channel_validity(&channel_id)?;
```
bedeho: Can you make a separate tech-debt issue to rename to this `ensure_channel_exists`, which is the typical pattern we use for this kind of call. Lets not start changing it here, as it just increases scope of conflicts with ongoing work.

runtime-modules/content/src/lib.rs
```diff
@@ -273,6 +297,37 @@ decl_module! {
...
+            curator_group_id: T::CuratorGroupId,
+            permissions_by_level: ModerationPermissionsByLevel<T>
+        ) {
+            let sender = ensure_signed(origin)?;
+            // Ensure given origin is lead
+            ensure_lead_auth_success::<T>(&sender)?;
+            // Ensure curator group under provided curator_group_id already exist
+            Self::ensure_curator_group_under_given_id_exists(&curator_group_id)?;
+            // Ensure permissions_by_level map max. allowed size is not exceeded
+            ensure!(
```
bedeho: Use new method recommended above.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+        Ok(())
+    }
+
+    fn ensure_channel_bag_can_be_dropped(
+        dynamic_bag_id: &storage::DynamicBagId<T>,
+        num_objects_to_delete: u64,
+    ) -> Result<(bool, BTreeSet<DataObjectId<T>>), DispatchError> {
+        let bag_id = storage::BagIdType::from(dynamic_bag_id.clone());
+
+        if let Ok(bag) = T::DataObjectStorage::ensure_bag_exists(&bag_id) {
```
bedeho, line 2533: First off, be aware that this check should _never_ fail, at least when called  from `delete_channel`. If it does, then the runtime storage has some how become corrupted, because a channel should exist if-and-only-if the dynamic bag for that channel exists.

Now there is s still a question of what to do statically in the code itself. This has been a long running issue, and it always pops up at the boundry between pallets, like this. I suggest 
- add a panic debug statement that will not be included in the WASM, so `debug_assert`.
- opt for no side-effect in this edge case, so `ensure_channel_bag_can_be_dropped` returns error with some new error case, which causes `delete_channel` to not have ny side-effects.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+            ensure!(
+                bag.objects_number == num_objects_to_delete,
+                Error::<T>::InvalidBagSizeSpecified
+            );
+
+            // construct collection of assets to be removed
+            let assets_to_remove = T::DataObjectStorage::get_data_objects_id(&bag_id);
+
+            if !assets_to_remove.is_empty() {
+                Storage::<T>::can_delete_dynamic_bag_with_objects(&dynamic_bag_id)?;
```
bedeho: I tried to map out code path of all of these `can`calls in  `Storage`, and its not clear that you need to do any of these, regardless of whether `assets_to_remove` is empty or not. Can you please validate and confirm that this is needed, and why?

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+        // Update corresponding channel
+        // Remove recently deleted video from the channel
+        ChannelById::<T>::mutate(channel_id, |channel| {
+            channel.num_videos = channel.num_videos.saturating_sub(1)
+        });
+
+        Ok(())
+    }
+
+    fn ensure_channel_bag_can_be_dropped(
```
bedeho, line 2526: Calling this routine always involves a channel id in calling code, so may as well take that as the input argument rather than `dynamic_bag_id`. If the `dynamic_bag_id` is needed for other routines in same calling code, like `execute_delete_channel_mutation`, then those probably also can be restated in terms of channel ids. At least check if this is feasible.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+            Ok((true, assets_to_remove))
+        } else {
+            Ok((false, BTreeSet::<DataObjectId<T>>::new()))
+        }
+    }
+
+    fn execute_delete_channel_mutation(
+        sender: T::AccountId,
+        channel_id: T::ChannelId,
+        dynamic_bag_id: Option<&storage::DynamicBagId<T>>,
```
bedeho: This should not be wrapped any longer given my point to not move forward if the runtime is corrupt.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+        assets_to_remove: &BTreeSet<DataObjectId<T>>,
+    ) -> DispatchResult {
+        // channel has a dynamic bag associated to it -> remove assets from storage
+        if let Some(dynamic_bag_id) = dynamic_bag_id {
+            // remove specified assets from storage
+            if !assets_to_remove.is_empty() {
```
bedeho: All of this logic about checking and deleting objects separately is quite rough, it becomes much cleaner when this is merged

https://github.com/Joystream/joystream/pull/3253

So I think it needs to get merged first, then you can pull that in to simplify.

runtime-modules/content/src/permissions/curator_group.rs
bedeho: Lets review this later in the audit work milestone: https://github.com/Joystream/joystream/issues/3387

Lezek123: This turned out to be caused by missing `PartialOrd, Ord` traits for `ContentModerationAction`. After adding those I was able to succesfully switch to the nested `BTreeSet<ContentModerationAction>` approach. Done in https://github.com/Joystream/joystream/pull/3366/commits/001745b49cb9fd57fd3c41646c95e172635ae6e5

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Introduced `PausedChannelFunctionality` enum in https://github.com/Joystream/joystream/pull/3366/commits/001745b49cb9fd57fd3c41646c95e172635ae6e5

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+        assets_to_remove: &BTreeSet<DataObjectId<T>>,
+    ) -> DispatchResult {
+        // bloat bond logic: channel owner is refunded
+        video
+            .video_post_id
+            .as_ref()
+            .map(|video_post_id| {
+                Self::video_deletion_refund_logic(&sender, &video_id, &video_post_id)
+            })
+            .transpose()?;
```
bedeho, line 2509: Why is there a `?`, not obvious to me, and this means we can fail past mutation safe barrier.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+        video_id: T::VideoId,
+        video: &Video<T>,
+        assets_to_remove: &BTreeSet<DataObjectId<T>>,
+    ) -> DispatchResult {
+        // bloat bond logic: channel owner is refunded
+        video
+            .video_post_id
+            .as_ref()
+            .map(|video_post_id| {
+                Self::video_deletion_refund_logic(&sender, &video_id, &video_post_id)
```
bedeho, line 2507: This whole section is a bit complex, but I think it goes away when comments go away, but then again it may also reappear when we do this https://github.com/Joystream/joystream/issues/3394.

runtime-modules/content/src/lib.rs
```diff
@@ -2333,6 +2423,101 @@ impl<T: Trait> Module<T> {
...
+            })
+            .transpose()?;
+
+        if !assets_to_remove.is_empty() {
+            Storage::<T>::delete_data_objects(
+                sender.clone(),
+                Self::bag_id_for_channel(&channel_id),
+                assets_to_remove.clone(),
+            )?;
```
bedeho: Be aware that you here can now fail, at least statically, with some partial mutation being done above. If the case above can be dropped, then at least you could switch places with this, so you only have one fallible operation with side-effects, which is safe.

In fact, you should really take this approach outlined here 
https://github.com/Joystream/joystream/issues/3071#issuecomment-1019269870

of not even doing a separate `can` check. I don't even know if you need to wait for this PR to be merged to do that
https://github.com/Joystream/joystream/pull/3253

runtime-modules/content/src/lib.rs
```diff
@@ -892,36 +965,93 @@ decl_module! {
...
-                |video_post_id| Self::video_deletion_refund_logic(&sender, &video_id, &video_post_id)
-            ).transpose()?;
-
             //
             // == MUTATION SAFE ==
             //
 
+            Self::execute_delete_video_mutation(&sender, channel_id, video_id, &video, &assets_to_remove)?;
+
+            Self::deposit_event(RawEvent::VideoDeleted(actor, video_id));
```
bedeho, line 895: Better to include all assets deleted also, same goes for all other new events: make sure they have as rich information as possible to avoid QN mapping authors having to stitch together events from content & storage.

runtime-modules/content/src/lib.rs
```diff
@@ -575,94 +655,89 @@ decl_module! {
...
-                Storage::<T>::delete_dynamic_bag(
-                    sender,
-                    dyn_bag,
-                )?;
-            }
+            // Permissions check
+            let actions_to_perform = if num_objects_to_delete == 0 {
+                vec![ContentModerationAction::DeleteChannel]
+            } else {
+                vec![ContentModerationAction::DeleteChannel, ContentModerationAction::DeleteObject]
```
bedeho: So I was asking about this in Rocket: what it means to delete an object in a world where we don't have extrinsics solely for this purpose. I don't think it really makes sense to distinguish between these if we don't have such a direct action. I can't imagine the lead thinking "I'll allow Bob to delete videos, but only if they have no assets", or vice versa for channels. Lets just drop the concept of object deletion from the permission model, as it is effectively not part of the moderation model. That should lead to some simplification here also.

runtime-modules/content/src/lib.rs
```diff
@@ -575,94 +655,89 @@ decl_module! {
...
+            let (bag_exists, assets_to_remove) = Self::ensure_channel_bag_can_be_dropped(&dynamic_bag_id, num_objects_to_delete)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+            Self::execute_delete_channel_mutation(sender, channel_id, if bag_exists { Some(&dynamic_bag_id) } else { None }, &assets_to_remove)?;
 
             // deposit event
-            Self::deposit_event(RawEvent::ChannelDeleted(actor, channel_id));
+            Self::deposit_event(RawEvent::ChannelDeletedByModerator(actor, channel_id, rationale));
```
bedeho, line 632: Same issue: more info about what was deleted.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -480,3 +457,24 @@ pub fn ensure_authorized_to_update_max_reward<T: Trait>(sender: &T::AccountId) -
...
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+    actions: &Vec<ContentModerationAction>,
+    channel_privilege_level: T::ChannelPrivilegeLevel,
+) -> DispatchResult {
+    ensure_actor_auth_success::<T>(&sender, &actor)?;
+    // Ensure actor is either lead of part of curators group with sufficient permissions
+    match actor {
+        ContentActor::Lead => Ok(()),
+        ContentActor::Curator(curator_group_id, ..) => {
+            let group = Module::<T>::curator_group_by_id(&curator_group_id);
```
bedeho, line 474: What ensures this curator group id is valid? I think this default constructs the group if it does not exist.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -480,3 +457,24 @@ pub fn ensure_authorized_to_update_max_reward<T: Trait>(sender: &T::AccountId) -
...
+            group.ensure_can_perform_actions(actions, channel_privilege_level)?;
```
bedeho, line 475: Don't you need to check that the group is active?

runtime-modules/content/src/lib.rs
```diff
@@ -575,94 +655,89 @@ decl_module! {
...
 
-            // remove channel from on chain state
-            ChannelById::<T>::remove(channel_id);
+            let dynamic_bag_id = storage::DynamicBagId::<T>::Channel(channel_id);
+            let (bag_exists, assets_to_remove) = Self::ensure_channel_bag_can_be_dropped(&dynamic_bag_id, num_objects_to_delete)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+            Self::execute_delete_channel_mutation(sender, channel_id, if bag_exists { Some(&dynamic_bag_id) } else { None }, &assets_to_remove)?;
```
bedeho: Same issue here with bag having to exist.

runtime-modules/content/src/types.rs
```diff
@@ -87,19 +85,10 @@ pub struct ChannelRecord<MemberId: Ord, CuratorGroupId, AccountId, Balance> {
...
-        ensure!(
-            self.is_censored != is_censored,
-            Error::<T>::ChannelCensorshipStatusDidNotChange
-        );
-        Ok(())
-    }
+    // Privilege level (curators will have different moderation permissions w.r.t. this channel depending on this value)
+    pub privilege_level: ChannelPrivilegeLevel,
+    // Whether the channel has been marked as hidden by a curator
+    pub is_hidden: bool,
```
bedeho: We don't really need to store this, as it does not impact any transaction, the most recent _is_hidden_ just needs to be an even payload for a successful transaction which updates it, query node takes care of the rest. This means we cannot block a redundant state update where the state is set to the current value, but that is fine.

runtime-modules/content/src/types.rs
```diff
@@ -254,14 +244,14 @@ pub type VideoUpdateParameters<T> = VideoUpdateParametersRecord<StorageAssets<T>
...
-    /// Whether the curators have censored the video or not.
-    pub is_censored: bool,
     /// enable or not comments
     pub enable_comments: bool,
     /// First post to a video works as a description
     pub video_post_id: Option<VideoPostId>,
     /// Whether nft for this video have been issued.
     pub nft_status: Option<OwnedNFT>,
+    // Whether the video has been marked as hidden by a curator
+    pub is_hidden: bool,
```
bedeho: Same point as `Channel::is_hidden`.

runtime-modules/content/src/lib.rs
```diff
@@ -892,36 +965,93 @@ decl_module! {
...
+
+            // get information regarding channel
+            let channel_id = video.in_channel;
+            let channel = ChannelById::<T>::get(channel_id);
+
+            // Permissions check
+            let actions_to_perform = if assets_to_remove.is_empty() {
+                vec![ContentModerationAction::DeleteVideo]
+            } else {
+                vec![ContentModerationAction::DeleteVideo, ContentModerationAction::DeleteObject]
```
bedeho: Same comment here about not needing delete object as a permission.

bedeho (changes requested):
- Upon reading the code, I think it's actually more appropriate to refer to _privilege level_ as _channel permission group_, or just _permission group_. The _level_ nomer really triggers the idea of there being an ordering where a curator who has a certain permission for a given level will also by implications have this for everything at a lower level, sort of like CPU [protection rings](https://en.wikipedia.org/wiki/Protection_ring).
- I think your work unfortunately would have been much easier to do if you had had this PR to work with https://github.com/Joystream/joystream/pull/3253. In particular the pattern of checking whether a bag can be deleted, before deleting, and auto-deletion of all data objects in a bag, would have a large impact on this code. **Perhaps you should just wait for this code to get merged, and work around the deletion issues until you can pullin that work?**.

runtime-modules/content/src/lib.rs
Lezek123: I factored out this part of the code from `delete_channel`, from what I can tell one reason as to why `can_delete_dynamic_bag_with_objects` is needed here is because it ensures `InsufficientTreasuryBalance` will not be thrown during `delete_dynamic_bag` at https://github.com/Lezek123/substrate-runtime-joystream/blob/content-directory-new-permissions-model/runtime-modules/content/src/lib.rs#L2562 (which happens after `delete_data_objects`, so that would be an unsafe mutation).

runtime-modules/content/src/lib.rs
Lezek123: This wasn't introduced by me, it's just what I factored out from `delete_video`, I suppose the reason for this is that `video_deletion_refund_logic` could return the `DispatchError` from `Self::refund` which could return the `DispatchError` from `ContentTreasury::<T>::withdraw`

runtime-modules/content/src/lib.rs
Lezek123: I don't think adding this information will be of any use for the query node currently. There is `DataObjectsDeleted` event emitted before `VideoDeleted`, the QN needs to know how to handle it, so it needs to check whether the objects have any relationship with a video/channel anyway (in order to unset those relations first). Then it removes the objects and then when handling `VideoDeleted`/`ChannelDeleted` it can assume all associated objects were already removed.

runtime-modules/content/src/permissions/mod.rs
Lezek123: `ensure_actor_auth_success` (which is executed before `match actor`) ensures that.

runtime-modules/content/src/permissions/mod.rs
Lezek123: `ensure_actor_auth_success` (which is executed before `match actor`) ensures that.

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/3d6692e4f3234e6d2c35698ceb8a5619fb22e498

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/3d6692e4f3234e6d2c35698ceb8a5619fb22e498

runtime-modules/content/src/types.rs
Lezek123: Removed in https://github.com/Joystream/joystream/pull/3366/commits/3d6692e4f3234e6d2c35698ceb8a5619fb22e498

runtime-modules/content/src/types.rs
Lezek123: Removed in https://github.com/Joystream/joystream/pull/3366/commits/3d6692e4f3234e6d2c35698ceb8a5619fb22e498

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/3d6692e4f3234e6d2c35698ceb8a5619fb22e498

runtime-modules/content/src/lib.rs
Lezek123: [This was already done](https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/lib.rs#L521) during the removal of migration-related code on `olympia` branch, we'll merge those changes once we update `audit3` branch from `olympia`.

runtime-modules/content/src/lib.rs
Lezek123: Made obsolete by introducing new `delete_dynamic_bag_with_objects` function in https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: Changed those functions to take `channel_id` instead of `dynamic_bag_id` in https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: Made obsolete by  https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: As later suggested I decided to go with my own temporary implementation of `delete_dynamic_bag_with_objects` in https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: I removed `can` checks in favor of putting the validation directly inside `delete_dynamic_bag_with_objects` function in https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/3366/commits/4aa4dac265b5e8a12d6e3428a2f16f30977175e3

runtime-modules/content/src/lib.rs
Lezek123: I removed the distinction between removing a video with assets and a video without assets, but I still think we need separate extrinsics to remove channel/video assets by curators, because now that removing a video which has an nft issued is impossible, this opens up an attack vector where one can publish sensitive content, issue nft for it and therefore prevent the sensitive video asset from beeing removed (because it can only be removed along with the video, which in that case would not possible due to an nft beeing issued)

runtime-modules/content/src/lib.rs
Lezek123: I added `ChannelBagMissing` dispatch error and `debug_assert` statement

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -1,4 +1,47 @@
 use super::*;
+use sp_std::collections::btree_map::BTreeMap;
+#[cfg(feature = "std")]
+use strum_macros::EnumIter;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub enum ChannelFeature {
```
bedeho: `PausableChannelFeature`

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -1,4 +1,47 @@
...
+pub enum ChannelFeatureStatus {
+    Paused,
+    Active,
+}
+
+pub type ChannelFeatureStatusChanges = BTreeMap<ChannelFeature, ChannelFeatureStatus>;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub enum ContentModerationAction {
```
bedeho, line 41: I think this is just an internal type? so no client pallet or off-chain code needs to understand it? if so, then we should make it non-public I think.

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -481,6 +458,27 @@ pub fn ensure_authorized_to_update_min_cashout<T: Trait>(sender: &T::AccountId)
...
+    actor: &ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+    actions: &[ContentModerationAction],
+    channel_privilege_level: T::ChannelPrivilegeLevel,
+) -> DispatchResult {
+    ensure_actor_auth_success::<T>(&sender, &actor)?;
+    // Ensure actor is either lead of part of curators group with sufficient permissions
+    match actor {
+        ContentActor::Lead => Ok(()),
+        ContentActor::Curator(curator_group_id, ..) => {
+            let group = Module::<T>::curator_group_by_id(&curator_group_id);
```
bedeho, line 474: This lookup is not 100% great, as it may seem like you could get a default constructed group in return back, ideally `ensure_actor_auth_success` should have returned some wrapped thing for this, but as that is returning `DispatchResult` (evil!), we can't, so just leave it be.

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -1,4 +1,47 @@
...
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub enum ChannelFeatureStatus {
+    Paused,
+    Active,
+}
+
+pub type ChannelFeatureStatusChanges = BTreeMap<ChannelFeature, ChannelFeatureStatus>;
```
bedeho: Do we actually need this? 

Can't we simply provide `BTreeSet<ChannelFeature>` in `change_channel_features_status_as_moderator`, and just overwrite `Channel::paused_features`? 

For the permission validation, the runtime can easily compute the set of channel features which are being updated by simply XORing the old and new version of `paused_features`, i.e. `BTreeSet::symmetric_difference`

- https://docs.substrate.io/rustdocs/latest/sp_std/collections/btree_set/struct.BTreeSet.html#method.symmetric_difference
- https://en.wikipedia.org/wiki/Symmetric_difference

runtime-modules/content/src/lib.rs
```diff
@@ -449,6 +501,69 @@ decl_module! {
...
+            rationale: Vec<u8>,
+        ) -> DispatchResult {
+
+            let sender = ensure_signed(origin)?;
+            // check that channel exists
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+            let mut new_paused_features = channel.paused_features.clone();
+
+            // Check permissions for moderation actions
+            let mut actions = Vec::<ContentModerationAction>::new();
```
bedeho: If you follow my recommendation on updating this state, I believe most of this goes away.

runtime-modules/content/src/permissions/curator_group.rs
bedeho: Would it perhaps be useful to write an explicit comment on each of these which explains exactly what extrinsics, with what parameter inputs, are off limits as a result of each variant? I was first a bit skeptical that the implementation was correct when reading `update_video`, because  I had some loose idea in my mind about these variants being mapped to explicit singular extrinsics, but they are not, as some features can be triggered by multiple extrinsics, such as issuing an NFT.

runtime-modules/content/src/lib.rs
```diff
@@ -748,40 +888,81 @@ decl_module! {
...
+            let actions_to_perform = vec![ContentModerationAction::DeleteVideo];
+            ensure_actor_authorized_to_perform_moderation_actions::<T>(&sender, &actor, &actions_to_perform, channel.privilege_level)?;
+
+            // ensure video can be removed
+            Self::ensure_video_can_be_removed(&video)?;
 
             //
             // == MUTATION SAFE ==
             //
+            Self::execute_delete_video_mutation(&sender, channel_id, video_id, &video, &assets_to_remove)?;
```
bedeho: Can we move this above mutation safe barrier?

bedeho (changes requested):
Fantastic, only tweaks & minor comments.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: It's part of `ModerationPermissionsByLevel`, which is used as an input for both `create_curator_group` and `update_curator_group_permissions`

runtime-modules/content/src/permissions/mod.rs
Lezek123: I think there's no harm in changing `DispatchResult` to `Result<SomeType, DispatchError>`, we use that in some similar checks (like `ensure_bounty_actor_manager`)

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: This sacrifaces the ability to atomically update the status of a single feature without the need to know the current set of paused features, but can be done.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/d164b4b310c6af9f0affc7a76abdbfdd29e77938

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/d164b4b310c6af9f0affc7a76abdbfdd29e77938

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/d164b4b310c6af9f0affc7a76abdbfdd29e77938

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3366/commits/d164b4b310c6af9f0affc7a76abdbfdd29e77938

runtime-modules/content/src/lib.rs
bedeho: This appears to contradict the conclusion here https://github.com/joystream/joystream/issues/3084

but let us revisit this later in a separate discussion then.

bedeho (approved):
Nice job, merging.

PR #3370, opened 2022-03-08
small fixes

bwhm:
All changes I think we want for `olympia` as well:

- Currently, the `MinUnstakingPeriodLimit` is enforced as `>` meaning it's not the actual minimum.
- `DistributionRewardPeriod` from `14400 + 50;` to `14400 + 90;` as it was in "conflict" with `GatewayRewardPeriod`
 - `ElectedMemberRewardPeriod`  from `10` to `14400`, to avoid to many of these
- The Council budget was similarly incremented too "frequently":
   - `BudgetRefillPeriod` from `1000` to `14400` (one day, not con
   - `DefaultBudgetIncrement` from `1000` to `10000000` (this may need revising)

Finally, made it so when a new council is NOT elected, the `event` doesn't say otherwise :)

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -369,7 +369,7 @@ decl_event! {
         /// New council was elected and appointed
         NewCouncilElected(Vec<MemberId>),
 
-        /// New council was elected and appointed
+        /// New council was not elected
         NewCouncilNotElected(),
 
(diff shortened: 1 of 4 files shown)


mnaamani:
LGTM, fixed test in https://github.com/bwhm/joystream/pull/5 if you want to merge that in we can run the full tests.

PR #3393, opened 2022-03-11
Audit3. Storage pallet benchmarks

shamil-gadelshin:
#### Changes
- benchmarked all storage pallet's extrinsics
- generate weights for the storage pallet
- set weights for the storage pallet's extrinsics


shamil-gadelshin:
It's probably better to review this PR after updating the `audit3` branch with the latest code from `olympia`

mnaamani:
Is this bringing in update from rhodes branch ? 1600 files :D

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -1,32 +1,1376 @@
...
+            RawEvent::NumberOfStorageBucketsInDynamicBagCreationPolicyUpdated(
+                dynamic_bag_type,
+                new_number
+            ).into()
+        );
+    }
+
+    update_blacklist {
+        let i in 0 .. BLACKLIST_SIZE_LIMIT;
+        let j in 0 .. BLACKLIST_SIZE_LIMIT;
```
Lezek123: Why not use `T::BlacklistSizeLimit::get().saturated_into()`?
The results are very different for the current limit (`10000`)

runtime/src/weights/storage.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn update_number_of_storage_buckets_in_dynamic_bag_creation_policy() -> Weight {
+        (258_000_000 as Weight)
+            .saturating_add(DbWeight::get().reads(3 as Weight))
+            .saturating_add(DbWeight::get().writes(1 as Weight))
+    }
+    fn update_blacklist(i: u32, j: u32) -> Weight {
+        (463_903_000 as Weight)
+            .saturating_add((76_353_000 as Weight).saturating_mul(i as Weight))
+            .saturating_add((36_361_000 as Weight).saturating_mul(j as Weight))
+            .saturating_add(DbWeight::get().reads(203 as Weight))
```
Lezek123: It seems like this number actually depends on `j`, when I modified the benchmarks to use `BLACKLIST_SIZE_LIMIT` of `10000` (the current limit) I got `10003 as Weight` here, which leads me to conclusion that this may be just `3 + j`.

I'm surprised the benchmark script did not recognize this dependency (the way it did for `i`)

runtime/src/weights/storage.rs
```diff
@@ -0,0 +1,199 @@
...
+            .saturating_add(DbWeight::get().reads(2 as Weight))
+            .saturating_add(DbWeight::get().writes(2 as Weight))
+    }
+    fn update_number_of_storage_buckets_in_dynamic_bag_creation_policy() -> Weight {
+        (258_000_000 as Weight)
+            .saturating_add(DbWeight::get().reads(3 as Weight))
+            .saturating_add(DbWeight::get().writes(1 as Weight))
+    }
+    fn update_blacklist(i: u32, j: u32) -> Weight {
+        (463_903_000 as Weight)
```
Lezek123: This value is very dependent on the `BLACKLIST_SIZE_LIMIT` used for the benchmarking. It's ~55 times higher when using `10000` instead of `200`

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -1,32 +1,1376 @@
...
+
+fn create_cids(i: u32) -> BTreeSet<Cid> {
+    fn create_cid(i: u32) -> Cid {
+        let bytes = i.to_be_bytes();
+        let mut buffer = Vec::new();
+
+        // Total CID = 32 bytes
+        for _ in 0..8 {
+            buffer.append(&mut bytes.to_vec());
+        }
```
Lezek123: There is no constraint on the `cid` length, so using `32 bytes` is pretty much arbitrary, yet this number has quite visible effect on the results of the benchmarks.

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -1,32 +1,1376 @@
...
+) -> BTreeSet<DistributionBucketId<T>> {
+    iter::repeat(())
+        .take(i.saturated_into())
+        .map(|_| create_distribution_bucket_with_family::<T>(account_id.clone(), Some(family_id)))
+        .collect::<_>()
+}
+
+const BLACKLIST_SIZE_LIMIT: u32 = 200;
+const STORAGE_BUCKETS_FOR_BAG_NUMBER: u32 = 7;
+const DISTRIBUTION_BUCKETS_FOR_BAG_NUMBER: u32 = 7;
```
Lezek123: Shouldn't those be based on `StorageBucketsPerBagValueConstraint` / `DistributionBucketsPerBagValueConstraint` instead? (although I haven't noticed a substantial effect on the benchmarks when modifying this value)

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -1,32 +1,1376 @@
...
+
+    }: _ (RawOrigin::Signed(lead_account_id), remove_cids.clone(), add_cids.clone())
+    verify {
+        assert_last_event::<T>(
+            RawEvent::UpdateBlacklist(
+                remove_cids,
+                add_cids
+            ).into()
+        );
+    }
```
Lezek123: Missing storage state verification

runtime-modules/storage/src/benchmarking.rs
```diff
@@ -1,32 +1,1376 @@
...
+
+    update_distribution_buckets_for_bag {
+        let i in 1 .. DISTRIBUTION_BUCKETS_FOR_BAG_NUMBER;
+        let j in 1 .. DISTRIBUTION_BUCKETS_FOR_BAG_NUMBER;
+
+        let lead_account_id = insert_distribution_leader::<T>(DISTRIBUTION_WG_LEADER_ACCOUNT_ID);
+        let bag_id = BagId::<T>::Static(StaticBagId::Council);
+        let family_id = create_distribution_family::<T>(lead_account_id.clone());
+
+        let new_limit = 7;
```
Lezek123:
```suggestion
        let new_limit: u64 = DISTRIBUTION_BUCKETS_FOR_BAG_NUMBER.into();
```

runtime-modules/storage/src/lib.rs
```diff
@@ -2852,7 +3155,7 @@ decl_module! {
         }
 
         /// Create a dynamic bag. Development mode.
-        #[weight = 10_000_000] // TODO: adjust weight
+        #[weight = WeightInfoStorage::<T>::storage_operator_remark(msg.len().saturated_into())]
```
Lezek123, line 3191: Missing `/// <weight>` comment?

runtime-modules/storage/src/lib.rs
```diff
@@ -2871,7 +3174,7 @@ decl_module! {
         }
 
         /// Create a dynamic bag. Development mode.
-        #[weight = 10_000_000] // TODO: adjust weight
+        #[weight = WeightInfoStorage::<T>::distribution_operator_remark(msg.len().saturated_into())]
```
Lezek123, line 3218: Missing `/// <weight>` comment?

Lezek123 (changes requested):
Left a couple of comments w.r.t. the choice of constants for benchmarking.
I made a little summary to ilustrate how choosing different benchmarking constants may affect the result: https://gist.github.com/Lezek123/d9befd0ceb4793798f21cad9ffcf6e0a

The main issue I found is w/ `BLACKLIST_SIZE_LIMIT` and the `CID` length (which is currently set to `32 bytes` _magic number_).

I also noticed that the benchmarks are generating possibly invalid weight calculation function for `update_blacklist` (see the comment).

Another issue I noticed is that `chain-metadata.json` is outdated on this branch and some types are missing:
```
2022-03-28 14:57:51        REGISTRY: Unable to resolve type DistributionBucketsPerBagValueConstraint, it will fail on construction
```

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Fixed.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Fixed.

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: Fixed

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: This is an actual size of the CID (32 bytes hash).

runtime/src/weights/storage.rs
shamil-gadelshin: Changed to the trait constant.

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: Ok.

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: Fixed.

runtime/src/weights/storage.rs
shamil-gadelshin: I'm not it's a good idea to change it manually at this point. It will be regenerated multiple times, runtime upgrade will happen, and we will change the hardware.

shamil-gadelshin:
We don't regenerate `chain-metadata.json` and types for the audit3 branch because we don't change TypeScript apps and it is constantly producing the PR conflicts. Please, ask Mokhtar for details.

The issues with benchmarks generation could be affected later by hardware changes, substrate upgrades, benchmarking parameters, etc.

runtime/src/weights/storage.rs
Lezek123: I was wondering if we could tweak the benchmarking script in some way to fix this issue, but if that's not the case, perhaps you could make a GH issue, just so we don't forget about it?

This bug will make `update_blacklist` fee extremely high, especially w/ the blacklist size limit of 10000

shamil-gadelshin: https://github.com/Joystream/joystream/issues/3517

runtime-modules/storage/src/benchmarking.rs
Lezek123: I understand, but user can provide a `Vec` of any arbitrary length (as "CID") in an extrinsic argument and this will not be accounted for by the weight calculation function, which is a security issue.

Currently the weight/fee will be exactly the same no matter if one provides 1000 valid 32 bytes CIDs or 1000 "CIDs" of 100 KB each, yet the difference in processing time etc. will be huge between those.

shamil-gadelshin: I would rather add CID-length checks for the `update_blacklist` extrinsic. cc @bedeho

runtime/src/weights/storage.rs
Lezek123: Actually, it appears that this problem is now solved after re-generating the weight functions (after https://github.com/Joystream/joystream/pull/3393/commits/1331e2fc5c05828d09ffb3edac85d3544710720b) :rocket: 
```
	fn update_blacklist(i: u32, j: u32, ) -> Weight {
		(0 as Weight)
			.saturating_add((58_286_000 as Weight).saturating_mul(i as Weight))
			.saturating_add((27_969_000 as Weight).saturating_mul(j as Weight))
			.saturating_add(DbWeight::get().reads(3 as Weight))
			.saturating_add(DbWeight::get().reads((1 as Weight).saturating_mul(i as Weight)))
			.saturating_add(DbWeight::get().reads((1 as Weight).saturating_mul(j as Weight)))
			.saturating_add(DbWeight::get().writes(1 as Weight))
			.saturating_add(DbWeight::get().writes((1 as Weight).saturating_mul(i as Weight)))
	}
```

runtime-modules/storage/src/benchmarking.rs
bedeho: 1. Why would it be that this has such an effect? My mental model is that basically # database operations dominate everything else in weight function.
2. I am fine with adding a bound if that resolves the issue.

runtime/src/weights/storage.rs
shamil-gadelshin: Ok!

runtime-modules/storage/src/benchmarking.rs
shamil-gadelshin: Added CID-checks here: https://github.com/Joystream/joystream/pull/3393/commits/6897b1d1c3b2aaabc0992ceb414116df51781ae1

runtime-modules/storage/src/lib.rs
```diff
@@ -188,6 +188,9 @@ pub trait WeightInfo {
 
 type WeightInfoStorage<T> = <T as Trait>::WeightInfo;
 
+/// Content ID length in bytes (32 bytes hash).
+pub const CID_LENGTH: usize = 32;
```
Lezek123: Just note that both storage and distributor node currently use B58 representation of the multihash, which is 46 bytes:
```
$ ./bin/run leader:update-blacklist -a gW2jkz44uexQjrm5buTALDmq3yfnZpvH961enoTrfmhLVv --dev
2022-03-30 12:59:33:5933 info: Initialized runtime connection: ws://localhost:9944
2022-03-30 12:59:36:5936 info: Waiting for chain to be synced before proceeding.
2022-03-30 12:59:36:5936 info: Updating the blacklist...
2022-03-30 12:59:36:5936 info: Development mode is ON.
2022-03-30 12:59:36:5936 debug: Sending storage.updateBlacklist extrinsic...
2022-03-30 12:59:42:5942 error: Api Error: Error: Extrinsic execution error: InvalidCidLength ( Invalid CID length (must be 32 bytes))
```

runtime-modules/storage/src/lib.rs
```diff
@@ -3729,25 +3735,32 @@ impl<T: Trait> Module<T> {
...
     fn get_hashes_by_predicate<P: FnMut(&&Cid) -> bool>(
         hashes: &BTreeSet<Cid>,
         predicate: P,
-    ) -> BTreeSet<Cid> {
-        hashes
+    ) -> Result<BTreeSet<Cid>, DispatchError> {
+        ensure!(
+            hashes.iter().all(|cid| cid.len() == CID_LENGTH),
+            Error::<T>::InvalidCidLength
+        );
```
Lezek123, line 3774: Shouldn't we use a fixed-length type (like `Hash` or `[u8; 32]`) instead of checking if the provided vector is of expected length?

Lezek123 (changes requested):
Added 1 additional note and 1 suggestion.
I think it would be good to also re-generate the weight functions now that the benchmarking parameters have been adjusted.

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Thanks for catching this! Fixed: https://github.com/Joystream/joystream/pull/3393/commits/7efc372a721aab7cc3243a1816bed90fbbd51b5c

runtime-modules/storage/src/lib.rs
shamil-gadelshin: `Vec` is a generally easier type to work with. We accept only `Vec` in all the runtime and I'm not sure we need to create an exception. Also, we have `Vec` also on the TypeScript side - we probably need to test the apps if we change the parameter type. I'd like to avoid it. If we change the [hashing algorithm](https://github.com/multiformats/multihash#example) it would be easier to relax the check to  the `less than` operator and support two hashing types.

Lezek123: Ok, this change would probably have big implications across the entire project, so let's leave it as is.

runtime-modules/storage/src/lib.rs
```diff
@@ -1564,6 +1612,9 @@ decl_event! {
 decl_error! {
     /// Storage module predefined errors
     pub enum Error for Module<T: Trait>{
+        /// Invalid CID length (must be 32 bytes)
```
```suggestion
        /// Invalid CID length (must be 46 bytes)
```

Lezek123 (commented):
Looks good, could you just update `/runtime/src/weights/storage.rs`?

runtime-modules/storage/src/lib.rs
shamil-gadelshin: Sure: https://github.com/Joystream/joystream/pull/3393/commits/b27e568d4364581e0ce28b0fa9b6f9ce51267f77

shamil-gadelshin:
Updated weights: https://github.com/Joystream/joystream/pull/3393/commits/74bf05308b38f1fd9ae0b4c9135edd2651d9e545

Lezek123 (approved):
LGTM!

mnaamani:
Re-running failed job to see if this clippy error is reproducible on github runner:

```
 error[E0046]: not all trait items implemented, missing: `fund_working_group_budget`
   --> runtime-modules/storage/src/tests/mocks.rs:305:1
    |
305 | impl working_group::WeightInfo for Weights {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `fund_working_group_budget` in implementation
    |
    = help: implement the missing item: `fn fund_working_group_budget() -> u64 { todo!() }`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `pallet-storage`
```

PR #3399, opened 2022-03-14
query node - active video counters generalization

ondratra:
This PR generalizes logic previously seen in active video counters to be easily used for other properties calculated from other properties (for example, active forum posts, etc.).


mnaamani:
Happy to merge this just need approval from @Lezek123

mnaamani:
Happy to merge this just need approval from @Lezek123

Issue #3300, opened 2022-02-23
NFT query node - channel owner support


ondratra:
Track the current owner of NFT when an owner of NFT is the channel owner. QN must update the value when NFT owner changes (that should be covered already) as well as when NFT owner is set to the channel owner and channel owner changes.

Since channel owner can be curator group, add `ownerCuratorGroup` property in addition to `ownerMember` to `OwnedNft`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201899833798248) by [Unito](https://www.unito.io)

ondratra:
We will also need to track if OwnedNFT is owned by the channel or not. Add `isOwnedByChannel` to `OwnedNft` as well. Only when `isOwnedByChannel` is true, track all channel owner changes and reflect them to `OwnedNft`.

PR #3405, opened 2022-03-15
Rhodes QN NFT's creator channel shortcut

zeeshanakram3:
resolves this #3314

PR #3406, opened 2022-03-15
Rhodes NFT QN - channel owner support

zeeshanakram3:
resolves this #3300


kdembler:
Unfortunately this doesn't solve the mentioned issue, at least in my understanding of what was supposed to be done here. As I see it, if the NFT is owned by the channel, then its `ownerMember` property should be populated by the member that controls this channel. Seems you have the correct logic for the `ownerCuratorGroup`, it should work the same

zeeshanakram3:
> if the NFT is owned by the channel, then its `ownerMember` property should be populated by the member that controls this channel. 

`ownerMember` would only be populated if NFT is owned by member and not the channel. If NFT is owned by the channel then the `ownerMember` would be null, implying NFT is owned by the channel. Also, the channel owner information would be stored in extra `creatorChannel` field, implemented in #3405,

The implementation is based on the graphql schema for the `OwnedNFT` entity.
https://github.com/Joystream/joystream/blob/f033654298de5c6634596ceb9414d5b3398ef5d8/query-node/schemas/contentNft.graphql#L66

query-node/mappings/src/content/nft.ts
```diff
@@ -930,3 +945,44 @@ export async function contentNft_BuyNowCanceled({ event, store }: EventContext &
...
   await store.save<BuyNowCanceledEvent>(announcingPeriodStartedEvent)
 }
+
+export async function contentNft_NftSlingedBackToTheOriginalArtist({
+  event,
+  store,
+}: EventContext & StoreContext): Promise<void> {
+  // common event processing
+
+  const [videoId, ownerId] = new Content.NftBoughtEvent(event).params
```
ondratra: `NftBoughtEvent` event is mistakenly used instead of `NftSlingedBackToTheOriginalArtist` here. That will cause `ownerId` to be of type `ContentActor` -> that will have impact on `NftSlingedBackToTheOriginalArtistEvent.owner` definition.

query-node/mappings/src/content/nft.ts
```diff
@@ -930,3 +945,44 @@ export async function contentNft_BuyNowCanceled({ event, store }: EventContext &
...
+    videoId.toString(),
+    'Non-existing video was slinged',
+    'Non-existing nft was slinged'
+  )
+
+  // load channel
+  const channel = (await getById(store, Video, video.getId(), ['channel'])).channel
+
+  // load channel ownerCuratorGroup (if any)
+  const ownerCuratorGroup = (await getById(store, Channel, channel.getId(), ['ownerCuratorGroup'])).ownerCuratorGroup
```
ondratra: This can be simplified with relation dependency loading. FYI: relations can be chained e.g. `['channel', 'channel.ownerCuratorGroup', 'channel.ownerCuratorGroup.whateverRelationExistsThere']`
```suggestion
  const { video, nft } = await getNftFromVideo(
    store,
    videoId.toString(),
    'Non-existing video was slinged',
    'Non-existing nft was slinged',
    ['channel', 'channel.ownerCuratorGroup', 'channel.ownerMember']
  )
```

query-node/mappings/src/content/nft.ts
```diff
@@ -930,3 +945,44 @@ export async function contentNft_BuyNowCanceled({ event, store }: EventContext &
...
+
+  nft.ownerMember = undefined
+  nft.ownerCuratorGroup = ownerCuratorGroup
```
ondratra: I don't think there is any reason why the channel couldn't be owned by a member here, or is it?
```suggestion
  nft.ownerMember = video.channel?.ownerMember
  nft.ownerCuratorGroup = video.channel?.ownerCuratorGroup
```

query-node/mappings/src/content/nft.ts
```diff
@@ -349,12 +351,25 @@ export async function createNft(
...
+  const isOwnedByChannel = !ownerMember
+
+  // load channel
+  const channel = (await getById(store, Video, video.getId(), ['channel'])).channel
+
+  // load channel ownerCuratorGroup (if any)
+  const ownerCuratorGroup = isOwnedByChannel
+    ? (await getById(store, Channel, channel.getId(), ['ownerCuratorGroup'])).ownerCuratorGroup
+    : undefined
+
```
ondratra: Let's get rid of both `getById`s here. We can rely on having these values loaded in `video` (`video.channel`, `video.channel.ownerCuratorGroup`) - see https://github.com/Joystream/joystream/pull/3405/files#r827498072

Since `ownerCuratorGroup` load is not ensured yet, ensure it is loaded before `createNft` is called. There are only 2 places right now - in `contentNft_NftIssued` and `content_VideoCreated`. (See https://github.com/Joystream/joystream/pull/3406#discussion_r827506457)

Also, don't forget that channel can also be owned by a member. Only the curator group owner is handled at the moment.

ondratra (changes requested):

ondratra:
> > if the NFT is owned by the channel, then its `ownerMember` property should be populated by the member that controls this channel.
> 
> `ownerMember` would only be populated if NFT is owned by member and not the channel. If NFT is owned by the channel then the `ownerMember` would be null, implying NFT is owned by the channel. Also, the channel owner information would be stored in extra `creatorChannel` field, implemented in #3405,
> 
> The implementation is based on the graphql schema for the `OwnedNFT` entity.
> 
> https://github.com/Joystream/joystream/blob/f033654298de5c6634596ceb9414d5b3398ef5d8/query-node/schemas/contentNft.graphql#L66

We had a private discussion with Klaudiusz before and decided that the value will always be set. The mentioned description in the schema is obsolete. Please adjust the code and remove `Channel owner owns the NFT if not set.` description part from the schema.

As you correctly pointed out, the information is theoretically retrievable, but due to some current query limitations, we tend to provide redundant information in some entities. (For example, https://github.com/Joystream/joystream/pull/3405 had the same point as this).

query-node/mappings/src/content/nft.ts
```diff
@@ -952,24 +958,19 @@ export async function contentNft_NftSlingedBackToTheOriginalArtist({
 }: EventContext & StoreContext): Promise<void> {
   // common event processing
 
-  const [videoId, ownerId] = new Content.NftBoughtEvent(event).params
+  const [videoId, ownerId] = new Content.NftSlingedBackToTheOriginalArtistEvent(event).params
```
ondratra: This change is ok, but `ownerId` is of type `ContentActor` now -> we need to reflect that change of definition in `NftSlingedBackToTheOriginalArtistEvent` in schema and on lane `owner: new Membership({ id: ownerId.toString() }),`

query-node/mappings/src/content/nft.ts
```diff
@@ -1,7 +1,7 @@
 // TODO: solve events' relations to videos and other entites that can be changed or deleted
 
 import { DatabaseManager, EventContext, StoreContext, SubstrateEvent } from '@joystream/hydra-common'
-import { genericEventFields, inconsistentState, logger } from '../common'
+import { genericEventFields, getById, inconsistentState, logger } from '../common'
```
ondratra: `getById` is not needed anymore in this file.

ondratra (changes requested):

ondratra (approved):

kdembler:
@zeeshanakram3 @ondratra Needs rebase and the checks are failing, is that because of the open auction changes? If so, should we merge anyway?

ondratra:
> @zeeshanakram3 @ondratra Needs rebase and the checks are failing, is that because of the open auction changes? If so, should we merge anyway?

Merge conflicts have been resolved. Auction is breaking the tests and checks at the moment, but we can fix lint later on together with auction changes.

kdembler (approved):

PR #3407, opened 2022-03-15
query node - NFT royalty overflow hotfix

ondratra:
This PR contains a hotfix for `royalty` value, causing a number overflow error. It discards actual royalty value and sets dummy value (`1`). It will need a proper fix after some testing is done (the actual number must be divided by proper constant).

### DIFF
--- a/query-node/mappings/src/content/nft.ts
+++ b/query-node/mappings/src/content/nft.ts
@@ -345,7 +345,9 @@ export async function createNft(
 
   // calculate some values
   const creatorRoyalty = nftIssuanceParameters.royalty.isSome
-    ? nftIssuanceParameters.royalty.unwrap().toNumber()
+    ? // TODO: this is causing number overload and needs to be divided by magic constant seen in tests
+      // ? nftIssuanceParameters.royalty.unwrap().toNumber()
+      1
     : undefined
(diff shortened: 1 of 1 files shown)


query-node/mappings/src/content/nft.ts
```diff
@@ -345,7 +345,9 @@ export async function createNft(
 
   // calculate some values
   const creatorRoyalty = nftIssuanceParameters.royalty.isSome
-    ? nftIssuanceParameters.royalty.unwrap().toNumber()
+    ? // TODO: this is causing number overload and needs to be divided by magic constant seen in tests
+      // ? nftIssuanceParameters.royalty.unwrap().toNumber()
+      1
```
mnaamani, line 350: I think this is the Perbill type which is also being fixed in https://github.com/Joystream/joystream/pull/3411
So shouldn't you be able to now do big num division:
```
const billion = new BN(1000000000)
royalty.unwrap().div(billion).toNumber()
```

mnaamani: or are we not able to store decimal values in the schema also?

mnaamani: Or if with the fixed type now being 32bits [asked here](https://github.com/Joystream/joystream/pull/3411/files#r826855654) then simply .toNumber() will work without overflow?

PR #3408, opened 2022-03-15
Rhodes NFT QN - fix planned end at block

### LINKED ISSUE #3304: NFT query node - `plannedEndAtBlock` doesn't take extension into account
Seems `Auction.plannedEndAtBlock` doesn't get updated when a bid is made during English auction's extension period. When that happens, the auction gets extended so the planned end should be updated as well.

Here's an example, on the NFT playground (https://54.172.37.177.nip.io/orion/graphql):

1. NFT for video with id `8`, has the following parameters:
	- "startsAtBlock": 161505
	- "extensionPeriod": 10
	- "duration": 20
	- "plannedEndAtBlock": 161525
1. So far so good, planned end is `start + duration`
1. But, if you also query bids made to this auction, there is one bid, made at block `161522`. That's 3 blocks away from the planned end date, so this should trigger extension of additional 10 blocks, moving planned end to block `161535`

Relevant runtime code is here: https://github.com/Joystream/joystream/blob/olympia/runtime-modules/content/src/nft/types.rs#L289

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201899740607978) by [Unito](https://www.unito.io)

zeeshanakram3:
fixes #3304


query-node/mappings/src/content/nft.ts
```diff
@@ -550,6 +551,20 @@ export async function contentNft_AuctionBidMade({ event, store }: EventContext &
...
+  if (extendsAuction.valueOf() === true) {
+    const { auction } = await getCurrentAuctionFromVideo(
+      store,
+      videoId.toString(),
+      'Non-existing video got new bid',
+      'Non-existing auction got new bid'
+    )
+
+    const englishAuctionExtensionPeriod = (auction.auctionType as AuctionTypeEnglish).extensionPeriod!
+    auction.plannedEndAtBlock = auction.plannedEndAtBlock! + englishAuctionExtensionPeriod
```
ondratra: While your solution here is valid, I'd suggest having default values `0` rather than non-null assertion. It is safer in case we introduce an inconsistency (a bug) in the future that will break the assertation. Saving `NaN`(`isNaN(undefined + 1) === true`) to database could cause a processor to crash; it would be better to have "just" an invalid value in such case.
```suggestion
    const englishAuctionExtensionPeriod = (auction.auctionType as AuctionTypeEnglish).extensionPeriod
    auction.plannedEndAtBlock = (auction.plannedEndAtBlock || 0) + (englishAuctionExtensionPeriod || 0)
```

ondratra (changes requested):
This issue is almost done. I just have one suggestion. CI tests haven't run on this PR (that occasionally happens). Accepting the suggested code will hopefully make them run.

ondratra (approved):
LGTM

kdembler (approved):

PR #3411, opened 2022-03-15
Rhodes fix perbill type

dobertRowneySr:
Fixes the `Royalty` type on the `@joystream/types` library

### DIFF
--- a/types/package.json
+++ b/types/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/types",
-  "version": "0.18.5",
+  "version": "0.18.19",
   "description": "Types for Joystream Substrate Runtime - Olympia release",
   "main": "index.js",
   "types": "index.d.ts",
--- a/types/package.json
+++ b/types/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/types",
-  "version": "0.18.19",
+  "version": "0.19.0",
   "description": "Types for Joystream Substrate Runtime - Olympia release",
   "main": "index.js",
   "types": "index.d.ts",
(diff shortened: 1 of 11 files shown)


types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/types",
-  "version": "0.18.5",
+  "version": "0.18.19",
```
kdembler: I think that should be `0.19.0` instead?

mnaamani: The idea was to bump this to v0.19.0
```suggestion
  "version": "0.19.0",
```

But you will need to update also all the other packages in the workspace that have dependency on `"^0.18.5"`  and change them to `"^0.19.0"`. This would be in
```
cli/package.json
distributor-node/package.json
query-node/mappings/package.json
storage-node/package.json
utilts/api-scripts/package.json
utils/migration-scripts/package.json
tests/package.json
```

kdembler: @mnaamani Just FYI your suggestion uses `0.19.10` instead of `0.19.0`

mnaamani: yep fixed it thank you :+1:

dobertRowneySr: fixed in 734c6643850421322a6dde222024b0d1f4ae679b

dobertRowneySr: fixed in 734c6643850421322a6dde222024b0d1f4ae679b

types/src/content/index.ts
```diff
@@ -19,7 +19,7 @@ export class CurrencyAmount extends CurrencyOf {}
 
 // NFT types
 
-export class Royalty extends UInt {}
+export class Royalty extends UInt.with(32, 'Perbill') {}
```
mnaamani, line 22: What does `32` here mean? It will be a 32 bit value?

dobertRowneySr:
I also confirm that `Perbill` is a `u32` in the runtime  
https://docs.substrate.io/rustdocs/latest/src/sp_arithmetic/per_things.rs.html#1593-1602

types/src/content/index.ts
dobertRowneySr: Yes

mnaamani (approved):

kdembler (approved):

PR #3427, opened 2022-03-16
Olympia devops updates

mnaamani:
Primarily fixes Network Deployment script to work with building image from source, and to ensure the rpc instance has required toolchain when we don't use a pre-built AMI with joystream setup already done. Original PR https://github.com/Joystream/joystream/pull/3262

I successfully tested this in run https://github.com/Joystream/joystream/actions/runs/1992152324

Fixed messy approach for loading required env variables in start.sh which need to be propagated to the scenario runner, by having the scenario load both the root .env and testing .env files. This makes it now possible to override settings in shell env calling start.sh to configure both the docker-compose services and the scenario run. For this reason I modified the root .env file to note use `${VAR}` syntax, which is correctly handled by newer versions of docker-compose but the npm `dotenv` will not process them in the same way. This work is done in 4f0cbd4fd00292d609b48a55203ed3de3baabd5c

The only issue not resolved is if we want to use a specific variable value from the root .env that could have been overridden in shell. We need a kind of similar behavior to `dotenv` in bash :). For now we can just run some ts script similarly to the scenario. There are many approaches .. some can be seen here: https://gist.github.com/judy2k/7656bfe3b322d669ef75364a46327836?permalink_comment_id=3654745#gistcomment-3654745


devops/aws/build-code.yml
```diff
@@ -1,18 +1,27 @@
 ---
 
 - name: Get latest Joystream code, build it and copy binary to local
```
Lezek123: Is this description still valid?
Another related question: I noticed that those tasks are ran inside `Run playbook to compile joystream-node on build server` step, but they now actually also affect the rpc node(s). Should that description be adjusted also?

start.sh
```diff
@@ -36,10 +36,6 @@ docker-compose up -d orion
 
 ## Init the chain with some state
 if [[ $SKIP_CHAIN_SETUP != 'true' ]]; then
-  set -a
-  . ./.env
-  set +a
```
Lezek123, line 41: Should this be also removed from other network-tests-related bash scripts, like `run-tests.sh`, `run-full-tests.sh` etc?

tests/network-tests/src/Scenario.ts
```diff
@@ -48,8 +49,11 @@ function writeOutput(api: Api, miniSecret: string) {
 }
 
 export async function scenario(label: string, scene: (props: ScenarioProps) => Promise<void>): Promise<void> {
-  // Load env variables
-  config()
+  // Load env variables - test framework specific
+  console.log(config({ path: path.join(__dirname, '../.env') }))
```
mnaamani: TODO: bad idea to log env variables in-case there are sensitive secrets in here (left over from debugging)

Lezek123 (commented):
I'm not very familiar with the devops workspace, but since the github action worked I'm assuming it's all good.

Left 2 small comments.

start.sh
mnaamani: yes good point it can indeed be removed

mnaamani: fixed in 483191dd0ad40b9561ea5ca6a801c0f0599b49b8

devops/aws/build-code.yml
mnaamani: Yes this is a case where refactoring missed updating task names. Fixed in 0c140543ce00819d346b7dc1048c61e67219aa04

tests/network-tests/src/Scenario.ts
mnaamani: fixed in 74909c991bb48a3114f893fba6da9a22943a7d74

Lezek123 (approved):
LGTM

Issue #2816, opened 2021-11-04
Channel ownership transfer (or swap with payment)


bedeho:
# Background
Primary use case so far has been in order to facilitate setting up channels for creators, who do not have tJOY/JOY, with the goal of being able to auto synch content from YT on their behalf. Importantly, this should be in a context where any `sudo` or `sudoAs` features are not available.

The steps would be

1. Creator signs up and authenticates ownership of channel `X` on YT.
1. Creator provides raw account `A_X` to infra.
1. Infra creates a new channel `Y` on Joystream under robot member `infra`.
1. Infra adds `infra` as a collaborator to `Y`: this could possibly be one extrinsic if we enhance creation parameters.
1. Infra creates member `member_X` on-chain, controlled by controller account `A_X`.
1. Infra transfers ownerhsip from `infra` to  `member_X`

At this point  `infra` can keep synching & uploading videos to `Y`, but creator can opt out unilaterally at any time.

# Proposal
Add simple transfer ownership feature to channels, should def. be double-opt in, and we may as well throw in a side-payment condition feature, like for NFTs, in order for members to do trustless sales of channels if they want.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958578530831) by [Unito](https://www.unito.io)

bedeho:
Actually, no reason this can't be across all possible channel owner types, can should ahve double opt-in with side-payment feature, just like we have for NFTs.

shamil-gadelshin:
> 5. Infra creates member member_X on-chain,

Do we use existing member invitation infrastructure or we need to create a third way to introduce a member (assuming 'buying the membership' is a second way)?

> 6. Infra transfers ownerhsip from infra to member_X

This the only step that requires the runtime changes, correct? 

We will introduce an extrinsic in the content pallet with ownership transfer and reward account change. 

> At this point infra can keep synching & uploading videos to Y, but creator can opt out unilaterally at any time.

Does it require the runtime change?

> Add simple transfer ownership feature to channels, should def. be double-opt in, and we may as well throw in a side-payment condition feature, like for NFTs, in order for members to do trustless sales of channels if they want.

Does it mean that we need creating the ownership transfer in two steps (extrinsics)? The first one would be an actual transfer and the second one would be a new owner confirmation.

bedeho:
> Do we use existing member invitation infrastructure or we need to create a third way to introduce a member (assuming 'buying the membership' is a second way)?

[...]

PR #3432, opened 2022-03-16
Audit3. Runtime. Add channel transfer feature to the content pallet.

shamil-gadelshin:
[Original issue](https://github.com/Joystream/joystream/issues/2816)

## Changes
- new extrinsic `update_channel_transfer_status`
- new extrinsic `accept_channel_transfer`
- added 'pending transfer` check barriers for multiple extrinsics (prevent extrinsics call for channels with pending transfer)

## Comments
1.  Parameters validation
    - existing check for the `new_collaborators` was used (validate_member_set)
    - no upper bound for `new_collaborators` was enforced similar to the existing code
    - no channel ownership or price checks were added
2. No price related logic was added

#### Modified extrinsics
- update_channel_censorship_status
- update_channel
- create_video
- update_video
- delete_video
- edit_post_text
- delete_post
- update_moderator_set
- claim_channel_reward
#### Skipped extrinsics
- delete_channel - doesn't make sense to prevent this operation
#### To be modified extrinsics ([separate issue](https://github.com/Joystream/joystream/issues/3433))
- start_nft_auction
- cancel_nft_auction
- cancel_offer
- cancel_buy_now
- make_bid
- cancel_open_auction_bid
- claim_won_english_auction
- pick_open_auction_winner
- offer_nft
- sling_nft_back
- accept_incoming_offer
- sell_nft
- buy_nft
- channel_owner_remark
- channel_collaborator_remark
- channel_moderator_remark
- nft_owner_remark


shamil-gadelshin:
We should update the `audit3` branch with the code from `rhodes` branch prior to reviewing this PR.

runtime-modules/content/src/lib.rs
```diff
@@ -1948,6 +1949,38 @@ decl_module! {
...
+            origin,
+            channel_id: T::ChannelId,
+            actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+            new_status_params: PendingTransfer<T::MemberId, T::CuratorGroupId, BalanceOf<T>>
+        ) {
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+            ensure_actor_authorized_to_transfer_channel::<T>(origin, &actor, &channel.owner)?;
+
+            ensure!(
+                channel.transfer_status == ChannelTransferStatus::NoActiveTransfer,
```
bedeho: This constraint is not required, I should be able to change terms of an ongoing transfer in any way, this is totally useful, e.g. reducing price or anything else.

runtime-modules/content/src/lib.rs
```diff
@@ -1948,6 +1949,38 @@ decl_module! {
...
+                Error::<T>::InvalidChannelTransferStatus
+            );
+
+            Self::validate_member_set(&new_status_params.new_collaborators)?;
```
bedeho: I would not even bother with this, let people provide whatever ids they want, runtime does not need to babysit, and if we are doing this check other places I would say that is the mistake, and it should be removed in the interest of consistency, rather than adding here.

bedeho (changes requested):
Minor tweaks.

runtime-modules/content/src/lib.rs
shamil-gadelshin: Should, I remove all entries of the `validate_member_set` within this task?

- create_channel
https://github.com/Joystream/joystream/blob/b3d36b7ebf348cf083b79bbe904bce6ebb1ea341/runtime-modules/content/src/lib.rs#L355

- update_channel
https://github.com/Joystream/joystream/blob/b3d36b7ebf348cf083b79bbe904bce6ebb1ea341/runtime-modules/content/src/lib.rs#L461

- update_moderator_set
https://github.com/Joystream/joystream/blob/b3d36b7ebf348cf083b79bbe904bce6ebb1ea341/runtime-modules/content/src/lib.rs#L1235

runtime-modules/content/src/lib.rs
shamil-gadelshin: We forgot about the `cancel_transfer` extrinsic. But we could skip creating it if we allow `update_channel_transfer_status` to set any transfer status.

runtime-modules/content/src/lib.rs
bedeho: yes I think so, thank you for chasing this down here in this PR. Some tests will probably break.

runtime-modules/content/src/lib.rs
bedeho: yes this was the idea.

bedeho (approved):
Excellent work.

Issue #3372, opened 2022-03-08
Introduce channel accounts


bedeho:
As described here 

https://github.com/Joystream/joystream/issues/3263

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201934969693349) by [Unito](https://www.unito.io)

PR #3443, opened 2022-03-17
Audit3 feature module owned accounts

dobertRowneySr:
addresses #3372

**Changes**:
- every channel `channel_id` has now exactly one (sub)-module account obtained by encoding `(MODULE_ID, channel_id)`
- rewards & nft precedings are credited to such accounts


node/src/chain_spec/mod.rs
```diff
@@ -273,6 +273,7 @@ pub fn testnet_genesis(
             ContentConfig {
                 next_curator_group_id: 1,
                 next_channel_category_id: 1,
+                // REMARK: do not set to 0 since it will break Content Treasury Handler logic
```
bedeho: What does this mean?

runtime-modules/content/src/types.rs
```diff
@@ -576,13 +554,18 @@ impl<ChannelId: Clone, SeriesId: Clone, VideoPostId: Clone, OwnedNFT: Clone>
...
 
     /// The account ID of the module account.
     fn module_account_id() -> T::AccountId {
-        Self::ModuleId::get().into_sub_account(Vec::<u8>::new())
+        Self::ModuleId::get().into_sub_account(T::ChannelId::zero())
+    }
+
+    /// The account ID for the specified channel_id
+    fn account_for_channel_id(channel_id: T::ChannelId) -> T::AccountId {
+        Self::ModuleId::get().into_sub_account(channel_id)
```
bedeho: What about if `channel_id`== 0? wouldn't this give same account as fore the entire module?
Wouldn't it be a bit more safe to have something like `"channel"||channel_id` to make it impossible for there to be conflict?

runtime-modules/content/src/nft/mod.rs
```diff
@@ -345,11 +345,10 @@ impl<T: Trait> Module<T> {
                 _ => (),
             };
 
-            // Should always be Some(_) at this stage, because of previously made check.
-            if let Some(creator_account_id) = Self::channel_by_id(in_channel).reward_account {
-                // Deposit royalty into creator account
-                T::Currency::deposit_creating(&creator_account_id, royalty);
-            }
+            // Deposit royalty into creator account
```
bedeho: This fragment is getting repeated a lot, probably will only increase, perhaps make it into a method?

bedeho (changes requested):
Nice work, only minor tweaks needed.

node/src/chain_spec/mod.rs
dobertRowneySr: - Content treasury account Id, used for storing `bloat_bond`s (for video comments) has address `encode(MODULE_ID, 0)`
- Any other channel account has address `encode(MODULE_ID, channel_id)`

If At genensis first channel is `0` then it can be credited with bloat bonds (for example).
I realize now that this is not a 100% robust solution. So in order to address this and the next comment:
- Treasury account = `encode((MODULE_ID, "TREASURY"))`
- Any other internal channel account = `encode((MODULE_ID, "CHANNEL", channel_id))`

runtime-modules/content/src/types.rs
dobertRowneySr: see above comment

node/src/chain_spec/mod.rs
dobertRowneySr: in this way `channel_id` s can be started at 0 with no unintended consequences

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: Ok, I had to unify the balance type used: removed `CurrencyOf<T>` in favor of `BalanceOf<T>` in order to properly reuse the new method `deposit_to_creator`

runtime-modules/content/src/types.rs
```diff
@@ -443,13 +421,18 @@ impl<ChannelId: Clone, VideoPostId: Clone, OwnedNft: Clone>
...
 
+    /// The account ID of the module account.
+    fn account_for_channel(channel_id: T::ChannelId) -> T::AccountId {
+        Self::ModuleId::get().into_sub_account(("CHANNEL", channel_id))
+    }
+
     /// The account ID of the module account.
     fn module_account_id() -> T::AccountId {
-        Self::ModuleId::get().into_sub_account(Vec::<u8>::new())
+        Self::ModuleId::get().into_sub_account("TREASURY")
```
Lezek123, line 435: This is inconsistent now with storage pallet's `ModuleAccount::module_account_id`, was there a specific reason to change the previous implementation?

runtime-modules/content/src/types.rs
```diff
@@ -473,6 +456,14 @@ pub trait ModuleAccount<T: balances::Trait> {
         )
     }
 
+    /// Deposit amount to internal creator account: infallible
+    fn deposit_to_creator(channel_id: T::ChannelId, amount: BalanceOf<T>) {
```
Lezek123: Why is it called the `creator` account here, instead of `channel_account`?
This was a little confusing for me, as it seems like it refers to channel creator (for example, a member).

Lezek123 (commented):
Looks good, just added 2 questions, but nothing major.

**Additional notes:**
`ensure_actor_can_manage_reward_account` function can now be removed

runtime-modules/content/src/types.rs
bedeho: I think this is actually better, so if we are to change anything, it should be in the storage module in my opinion, is that ok with you @Lezek123 ?

runtime-modules/content/src/nft/mod.rs
bedeho: Is that really needed? its always desirable to use trait rather than implementation, e.g. for testing purposes. We do this sort of payment many other places, why does that work fine?

bedeho (commented):
- looks good.
- I agree with renaming suggested https://github.com/Joystream/joystream/pull/3443#discussion_r830075785
- Coupling with balances module seems unwarranted.

runtime-modules/content/src/types.rs
bedeho: Agreed.

dobertRowneySr:
>   Coupling with balances module seems unwarranted

Both `BalanceOf` and `CurrencyOf` relied on the Balance pallet anyway, which is the one providing the actual implementation

bedeho:
> Both BalanceOf and CurrencyOf relied on the Balance pallet anyway, which is the one providing the actual implementation

That does not mean that it does not matter, read this:

https://docs.substrate.io/v3/runtime/pallet-coupling/

I think we can merge this now, as its not a big deal - and we have a lot going on, and it can easily be altered later, but the difference is important.

Issue #3263, opened 2022-02-18
Finalize channel payouts in runtime


bedeho:
# Background
This has already been implemented in the content directory

https://github.com/Joystream/joystream/issues/2365

# Proposal
First some simple naming stuff. Lets refer to the act of using merkle proof as _claiming_, the event of having some amount credited to you in the merkle root as _earned_ and the act of transferring funds from your channel account (read below) to some other account as _withdrawing_. 

* [x] Rename `PullPaymentElement::cumulative_payout_claimed` to `cumulative_reward_earned`.
* [x] Rename `ChannelRecord::cumulative_payout_earned` to `cumulative_reward_claimed`.
* [x] Add new `withdraw_from_channel_balance` extrinsic which allows the channel owner to actually withdraw funds from the channel account to some other account. It should not affect `cumulative_reward_claimed`, only balance of account. Collaborator members cannot call this, only owner.
* [x] Add a new extrinsic `claim_and_withdraw`, which is an atomic invocation of `claim_channel_reward` & `withdraw_from_channel_balance`, **but remember to only emit one new event**. Collaborator members cannot call this, only owner.

Then the more substantive stuff:

* [x] (Done by @ignazio-bovo ) Replace channel accounts from being optional external accounts to being hard-coded internal module accounts, one unique account per channel.
* [x] Add proposal which can block all, and unblock, cashout attempts.
* [x] ~Add an extrinsic, which lead can invoke, to block/unblock specific individual channels from doing cashouts (or revenue splits).~
* [x] Drop `MaxRewardAllowed`, semantics are not that useful, instead introduce a `MaxCashoutAllowed`, analogous to `MinCashoutAllowed`, which only looks at the actual `cashout` value, not the total. Neither value can be set by lead, only council through proposal below.
* [x] Add council proposal to update merkle root, initiate payload upload to council storage bag, and extend to support simultaneously
  * possible new value for `MinCashoutAllowed`, not required.
  * possible new value for `MaxCashoutAllowed`, not required.
  * possible to credit content WG budget some amount, not required.
  * possible to automatically unblock cashouts if they current are blocked, not required, and if they are already blocked, there should be no error executing proposal none the less.
* [x] A new council proposal for only setting `MinCashoutAllowed` & `MaxCashoutAllowed`, not updating the payout.
* [x] Add mutation safe barrier.

# Deployment
[...]

Issue #3075, opened 2022-01-23
Cleaner storage asset updating integration code


bedeho:
# Background
https://github.com/Joystream/joystream/issues/3071#issuecomment-1019269870

Same issue here with needless repetition and complexity bleeding into client pallets without any payoff.

# Proposal
Rewrite `update_video` from this

```Rust
if let Some(upload_assets) = params.assets_to_upload.as_ref() {
    let params = Self::construct_upload_parameters(
        upload_assets,
        &channel_id,
        &sender
    );
    Storage::<T>::can_upload_data_objects(&params)?;
}

if !params.assets_to_remove.is_empty() {
    Storage::<T>::can_delete_data_objects(
        &Self::bag_id_for_channel(&channel_id),
        &params.assets_to_remove,
    )?;
}

//
// == MUTATION SAFE ==
//

if let Some(upload_assets) = params.assets_to_upload.as_ref() {
    let params = Self::construct_upload_parameters(
        upload_assets,
        &channel_id,
        &sender
    );
    Storage::<T>::upload_data_objects(params)?;
}

if !params.assets_to_remove.is_empty() {
    Storage::<T>::delete_data_objects(
        sender,
        Self::bag_id_for_channel(&channel_id),
        params.assets_to_remove.clone(),
    )?;
}
```

to some variation of this

```Rust


let upload_params = // map params.assets_to_upload to Some(Self::construct_upload_parameters)

let useful_result_to_add_to_event= Storage::<T>::upload_and_delete_data_objects(&upload_params, params.assets_to_remove)?;

//
// == MUTATION SAFE ==
//

```

The signature of `upload_and_delete_data_objects` probably needs more consideration, just outlining the principle here.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958581667551) by [Unito](https://www.unito.io)

bedeho:
Done.

PR #3444, opened 2022-03-17
Finalize channel payouts

Lezek123:
Implements https://github.com/Joystream/joystream/issues/3263


runtime-modules/content/src/tests/fixtures.rs
```diff
@@ -1619,21 +1871,50 @@ fn build_merkle_path_helper<E: Encode + Clone>(
...
     let commit = generate_merkle_root_helper(payments).pop().unwrap();
-    UpdateCommitmentValueFixture::default()
-        .with_commit(commit)
+    UpdateChannelPayoutsFixture::default()
+        .with_commitment(Some(commit))
         .call_and_assert(Ok(()));
 }
+
+fn channel_reward_account(channel: &Channel<Test>) -> Option<AccountId> {
+    if let Some(reward_account) = channel.reward_account {
```
shamil-gadelshin: Please, refactor such code. Consider using Option.map() and unwrap_or() methods.

runtime-modules/proposals/codex/Cargo.toml
```diff
@@ -21,6 +22,7 @@ proposals-discussion = { package = 'pallet-proposals-discussion', default-featur
 constitution = { package = 'pallet-constitution', default-features = false, path = '../../constitution'}
 membership = { package = 'pallet-membership', default-features = false, path = '../../membership'}
 blog = { package = 'pallet-blog', default-features = false, path = '../../blog'}
+content = { package = 'pallet-content', default-features = false, path = '../../content'}
```
shamil-gadelshin, line 25: Please, add an entry to the 'std' feature.

runtime/src/weights/proposals_codex.rs
```diff
@@ -191,4 +191,13 @@ impl proposals_codex::WeightInfo for WeightInfo {
             .saturating_add(DbWeight::get().reads(6 as Weight))
             .saturating_add(DbWeight::get().writes(9 as Weight))
     }
+    fn create_proposal_update_channel_payouts(t: u32, d: u32, i: u32) -> Weight {
+        // TODO: Adjust using reference hardware
```
shamil-gadelshin: This comment will be lost during the next auto-generation.

runtime-modules/content/src/lib.rs
```diff
@@ -2416,6 +2483,65 @@ impl<T: Trait> Module<T> {
...
+    fn execute_channel_balance_withdrawal(
+        reward_account: &T::AccountId,
+        destination: &T::AccountId,
+        amount: BalanceOf<T>,
+    ) -> DispatchResult {
+        <Balances<T> as Currency<T::AccountId>>::transfer(
+            reward_account,
+            destination,
+            amount,
+            ExistenceRequirement::AllowDeath,
```
shamil-gadelshin, line 2700: Should be changed to KeepAlive: https://github.com/Joystream/joystream/issues/3497

runtime-modules/content/src/lib.rs
```diff
@@ -1255,15 +1258,52 @@ decl_module! {
         }
 
         #[weight = 10_000_000] // TODO: adjust Weight
-        pub fn update_commitment(
+        pub fn update_channel_payouts(
```
shamil-gadelshin, line 1283: Rust doesn't format the macro code. Please, do this manually in all places respecting the rust code length.

runtime-modules/content/src/lib.rs
```diff
@@ -1255,15 +1258,52 @@ decl_module! {
...
+
+            let payload_data_object_id = params.payload.as_ref().map(|_| { Storage::<T>::next_data_object_id() });
+            if let Some(payload) = params.payload.as_ref() {
+                let upload_params = UploadParameters::<T> {
+                    bag_id: storage::BagId::<T>::from(StaticBagId::Council),
+                    object_creation_list: vec![payload.object_creation_params.clone()],
+                    deletion_prize_source_account_id: payload.uploader_account.clone(),
+                    expected_data_size_fee: payload.expected_data_size_fee,
+                };
+                Storage::<T>::upload_data_objects(upload_params)?;
```
shamil-gadelshin, line 1310: Invalid 'mutation mark' place.

runtime-modules/content/src/lib.rs
```diff
@@ -1255,15 +1258,52 @@ decl_module! {
...
+                ChannelCashoutsEnabled::put(channel_cashouts_enabled);
+            }
+
+            if let Some(commitment) = params.commitment.as_ref() {
+                <Commitment<T>>::put(*commitment);
+            }
 
-            <Commitment<T>>::put(new_commitment);
-            Self::deposit_event(RawEvent::CommitmentUpdated(new_commitment));
+            Self::deposit_event(RawEvent::ChannelPayoutsUpdated(
```
shamil-gadelshin, line 1333: There is a case when no actual update happens.

runtime-modules/content/src/lib.rs
```diff
@@ -1255,15 +1258,52 @@ decl_module! {
...
+        pub fn update_channel_payouts(
             origin,
-            new_commitment: <T as frame_system::Trait>::Hash,
+            params: UpdateChannelPayoutsParameters<T>
         ) {
-            let sender = ensure_signed(origin)?;
-            ensure_authorized_to_update_commitment::<T>(&sender)?;
+            ensure_root(origin)?;
+
+            let new_min_cashout_allowed = params.min_cashout_allowed.unwrap_or_else(Self::min_cashout_allowed);
```
shamil-gadelshin: Why do you correct the value instead of throwing the error?

shamil-gadelshin (changes requested):
Looks good, there are several minor issues.

runtime-modules/content/src/lib.rs
Lezek123: I followed an example from here: https://github.com/Joystream/joystream/issues/3075
Since `upload_data_objects` performs both the parameters check (which may return `DispatchError`) and the actual mutation, I think it's a valid approach to place `MUTATION_SAFE` after this call (since the exact place for `MUTATION_SAFE` mark is in the middle of `upload_data_objects`, to be more specific, under: `validate_upload_data_objects_parameters` call)

runtime-modules/content/src/lib.rs
Lezek123: Is it worth to introduce additional checks for that? It can happen either when all parameters are set to `None` or when all parameters that are not set to `None` have a value equal to the current one. That seems like quite a few checks to make (including 4 database reads), which may bump the weight significantly.

runtime-modules/content/src/lib.rs
Lezek123: I don't understand the question, why should I throw an error here?

If `params.min_cashout_allowed` or `params.max_cashout_allowed` is `None` we want to leave the value of it intact in the storage.

If just one of the two is provided (ie. only `max_cashout_allowed`), we want to use the _current_ value of the other in order to check if `min_cashout_allowed <= max_cashout_allowed` would still hold true after the update.

That's why I'm setting `new_min_cashout_allowed` variable to either the provided limit or the current one (and same goes for `new_max_cashout_allowed`) before running `ensure!(new_min_cashout_allowed <= new_max_cashout_allowed)`.

Note that the only purpose of `new_min_cashout_allowed` and `new_max_cashout_allowed` variables is this `ensure!` check, they aren't used for performing an actual update.

runtime-modules/content/src/lib.rs
Lezek123: https://github.com/Joystream/joystream/issues/3497#issuecomment-1080531129

runtime-modules/content/src/tests/fixtures.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3444/commits/2b96f1d4d761e0c3cd360fa44502f1fb43026b52

runtime-modules/proposals/codex/Cargo.toml
Lezek123: Done in https://github.com/Joystream/joystream/pull/3444/commits/2b96f1d4d761e0c3cd360fa44502f1fb43026b52

runtime/src/weights/proposals_codex.rs
Lezek123: Removed in https://github.com/Joystream/joystream/pull/3444/commits/2b96f1d4d761e0c3cd360fa44502f1fb43026b52

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3444/commits/2b96f1d4d761e0c3cd360fa44502f1fb43026b52

Lezek123:
https://github.com/Joystream/joystream/issues/3514 resolved in https://github.com/Joystream/joystream/pull/3444/commits/371eb59266c94e6b11b793246f2ab41bfa234f79

runtime-modules/content/src/lib.rs
shamil-gadelshin: Ok

runtime-modules/content/src/lib.rs
shamil-gadelshin: Ok

runtime-modules/content/src/lib.rs
shamil-gadelshin: Ok

runtime-modules/content/src/lib.rs
shamil-gadelshin: Ok

shamil-gadelshin (approved):
LGTM

bedeho:
Change base to `ephesus`.

PR #3452, opened 2022-03-18
Rhodes NFT QN - fix: creator royalties bug

### LINKED ISSUE #3426: NFT query node - royalties not working
As discovered few days ago, setting an NFT royalty will break a query node. We have a hotfix here #3407 but we need a full fix.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201976939841732) by [Unito](https://www.unito.io)

zeeshanakram3:
fixes #3426


query-node/mappings/src/content/nft.ts
```diff
@@ -345,7 +345,7 @@ export async function createNft(
 
   // calculate some values
   const creatorRoyalty = nftIssuanceParameters.royalty.isSome
-    ? nftIssuanceParameters.royalty.unwrap().toNumber()
+    ? nftIssuanceParameters.royalty.unwrap().div(new BN(10000000)).toNumber()
```
ondratra: Let's properly define and name this magic constant; put it into `query-node/mappings/src/content/utils.ts` and import it into test `Api.ts` instead of the same magic constant that's used there. https://github.com/Joystream/joystream/blob/ba96a82dab2bff6533c371dfd1810d92b8aeecac/tests/network-tests/src/Api.ts#L823

Also, the number can be written in a little bit easier to read format `new BN(10_000_000)`.

ondratra (changes requested):

tests/network-tests/src/misc/createChannelsAsMemberFixture.ts
```diff
@@ -21,7 +21,7 @@ export class CreateChannelsAsMemberFixture extends BaseFixture {
   public async execute(): Promise<void> {
     const account = await this.api.getMemberControllerAccount(this.memberId)
 
-    const channels = []
+    const channels: any = []
```
ondratra: `createMockChannel` returns `Promise<ChannelId>`. Replace `any` with it.

tests/network-tests/src/misc/createVideosAsMemberFixture.ts
```diff
@@ -17,7 +17,7 @@ export class CreateVideosAsMemberFixture extends BaseFixture {
   public async execute(): Promise<void> {
     const account = await this.api.getMemberControllerAccount(this.memberId)
 
-    const videos = []
+    const videos: any = []
```
ondratra: `createMockVideo` returns `Promise<VideoId>`. Replace `any` with it.

tests/network-tests/tsconfig.json
```diff
@@ -3,12 +3,13 @@
     "target": "esnext",
     "module": "commonjs",
     "outDir": "dist",
-    "rootDir": "src",
+    "rootDirs": ["src", "../../query-node/mappings/src/content/utils"],
```
ondratra: Why is this needed? We are surely not using `utils.ts` as root, or do we?

ondratra (changes requested):

tests/network-tests/tsconfig.json
zeeshanakram3: Was getting the following error while building `network-tests`:
```
File '/workspace/joystream/query-node/mappings/src/content/utils.ts' is not under 'rootDir' '/workspace/joystream/tests/network-tests/src'. 'rootDir' is expected to contain all source files.ts(6059)
```
So made the change.

_Update:_
Just found out that the above error disappears if change 
```javascript
"rootDir": "src"
```
to 
```javascript
"rootDirs": ["src"]
```
:shrug:

Issue #3380, opened 2022-03-09
Introduce Channel/Video -> BTreeSet<DataObjectId> association in the runtime state


Lezek123:
To solve https://github.com/Joystream/joystream/issues/2936 we need to introduce some state in the content directory to map data objects to specific channel/video in order to determine whether the actor is trying to remove a data object associated with a video that already has an nft issued (or channel that has some creator tokens issued).

The initial idea is to add `objects: BTreeSet<DataObjectId>` to `Channel` / `Video` entity and initialize/update the set on:
- `create_channel`
- `create_video`
- `update_channel`
- `update_video`

**_Note that this will only work if there are no other extrinsics (especially outside of the content pallet) through which it is possible to remove data objects that live inside a channel bag (not counting extrinsics that remove the entire `Channel` / `Video` entity along with all it's data objects, ie. `delete_video`, `delete_channel`)_**

Then we validate the existence of data object ids provided as `assets_to_remove` arg for:
- `update_video`
- `update_channel`

In given `Channel` / `Video` `objects` set.

We could also change the signature for `delete_video` to take `num_objects_to_delete` instead of `assets_to_remove` in order to:
- Be consistent with `delete_channel`
- Make sure all video assets are always removed along with the video
- Reduce the size and complexity of the parameters that need to be provided to the extrinsic

Introducing this state opens the door for preventing the removal of data objects associated with the video / channel depending on:
- video's nft status
- channel's creator tokens status
- whether the object is associated with a video (video media / thumbnail) or a channel (avatar / cover photo)

And (optionally) for introducing separate curator permissions (see: https://github.com/Joystream/joystream/issues/2760) for removing data objects depending on those conditions, for example:
- `DELETE_VIDEO_DATA_OBJECT_NO_ASSOCIATED_ASSETS` (allow deleting video data object in case channel has no creator tokens issued and the video has no nft issued)
- `DELETE_VIDEO_DATA_OBJECT_NFT` (allow deleting video data object in case video has an associated nft issued)
- `DELETE_VIDEO_DATA_OBJECT_CREATOR_TOKENS` (allow deleting video data object in case channel has creator tokens issued)
- `DELETE_ CHANNEL_DATA_OBJECT_NO_ASSOCIATED_ASSETS` (allow deleting channel data object in case channel has no creator tokens issued)
- `DELETE_CHANNEL_DATA_OBJECT_CREATOR_TOKENS` (allow deleting channel data object in case channel has some creator tokens issued)

[...]

Issue #2936, opened 2021-12-16
Video metadata and assets immutable when NFT


bedeho:
- user metadata on videos cannot be updated
- problem: it is still possible to delete a dataobject on-chain, which would be an asset for the video. Trying to prevent this would require starting to introduce content directory level information about what data objects exist, it quickly will get complicated to manage this.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201837747825058) by [Unito](https://www.unito.io)

Lezek123:
Resolved in https://github.com/Joystream/joystream/pull/3454

PR #3454, opened 2022-03-18
Channel/Video -> data objects association in the runtime

Lezek123:
Based on https://github.com/Joystream/joystream/pull/3366

Implements https://github.com/Joystream/joystream/issues/3380, fixes related issue https://github.com/Joystream/joystream/issues/2936, adds separate extrinsics for curators to remove channel/video assets and separate permissions for removing:
- non-nft video assets
- nft video assets
- channel assets

**TODOs:**
- [x] update types and chain metadata


runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -38,6 +38,9 @@ pub enum ContentModerationAction {
     ChangeChannelFeatureStatus(ChannelFeature),
     DeleteVideo,
     DeleteChannel,
+    DeleteVideoAssets,
```
bedeho: I think we should add deeper descriptions here also.

bedeho: Perhaps combining `DeleteVideoAssets` & `DeleteNftVideoAssets` into `DeleteVideoAssets(bool)` is more clear?
or we will have to rename `DeleteNftVideoAssets` to `DeleteVideoAssetsForVideoWithNFT` to be more clear, but it still leave `DeleteVideoAssets` unclear on the relevance of NFT status for a new reader.

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -38,6 +38,9 @@ pub enum ContentModerationAction {
...
+    DeleteNftVideoAssets,
+    DeleteChannelAssets,
```
bedeho: Almost worth renaming to `DeleteNonVideoChannelAssets`, just because it makes it crystal clear that this is a separate category.

runtime-modules/content/src/lib.rs
```diff
@@ -411,6 +411,9 @@ decl_module! {
             let channel_id = NextChannelId::<T>::get();
 
             let storage_assets = params.assets.clone().unwrap_or_default();
+
+            let data_objects_ids = Storage::<T>::get_next_data_object_ids(storage_assets.object_creation_list.len());
```
bedeho, line 419: This sort of leaks implementation assumptions into client code, and is just complex. We really shoul have `create_dynamic_bag` just return this sort of information

Sidenote: @shamil-gadelshin @ignazio-bovo one example of why a proper result type is needed.

**I think we should consider this change later, but adding a reference to the long running issue on this question.**

runtime-modules/content/src/lib.rs
```diff
@@ -2590,6 +2730,28 @@ impl<T: Trait> Module<T> {
             video_id, member_id,
         ))
     }
+
+    fn ensure_assets_to_remove_are_part_of_assets_set(
+        assets_to_remove: &BTreeSet<DataObjectId<T>>,
+        assets_set: &BTreeSet<DataObjectId<T>>,
+    ) -> DispatchResult {
+        ensure!(
+            assets_to_remove.iter().all(|id| assets_set.contains(id)),
```
bedeho: `assets_to_remove.is_subset(assets_set)`

runtime-modules/content/src/lib.rs
```diff
@@ -603,6 +617,50 @@ decl_module! {
...
+            actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+            channel_id: T::ChannelId,
+            assets_to_remove: BTreeSet<DataObjectId<T>>,
+            rationale: Vec<u8>,
+        ) {
+            let sender = ensure_signed(origin)?;
+
+            // check that channel exists
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+            let mut updated_data_objects_set = channel.data_objects.clone();
```
bedeho: Declare it close to where its needed.

runtime-modules/content/src/lib.rs
```diff
@@ -828,17 +895,29 @@ decl_module! {
                 channel.ensure_feature_not_paused::<T>(ChannelFeature::VideoNftIssuance)?;
             }
 
+            let assets_to_upload = params.assets_to_upload.clone().unwrap_or_default();
```
bedeho, line 897: These `is_empty` guards don't seem to be of much use, and just make the code less readable? They don't even generate more database operations, which is the only computational cost of relevance.

runtime-modules/content/src/lib.rs
```diff
@@ -477,17 +481,27 @@ decl_module! {
                 channel.collaborators = new_collabs.clone();
             }
 
+            let assets_to_upload = params.assets_to_upload.clone().unwrap_or_default();
```
bedeho: 1. These `is_empty` guards don't seem to be of much use, and just make the code less readable? They don't even generate more database operations, which is the only computational cost of relevance.
2. You seem to do a very similar sort of process three places when it comes to computing and checking updates the data objects of a video or channel. Is there room to consolidate the core logic here, so that you just have a single method that does the checking and returns the new set? Only do if the pattern is genuinely the same, but it seems so.
3. I know this mutable `channel` has been flying around since before this PR, but good if we could avoid the mutation, or at least not do it until the mutation safe barrier, although that barrier really is about storage mutation, not variables.

bedeho (changes requested):
Very solid, minor issues.

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Added in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

Lezek123: Done in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

runtime-modules/content/src/lib.rs
Lezek123: Refactored in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

runtime-modules/content/src/lib.rs
Lezek123: Refactored in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

runtime-modules/content/src/lib.rs
Lezek123: Refactored in https://github.com/Joystream/joystream/pull/3454/commits/d96affab416d06332652e74dcb0e3b914443c1f6

bedeho (approved):
_Perfect_.

Issue #3453, opened 2022-03-18
Olympia: Query node bug


bwhm:
A bounty was created by a tester (using polkadot-js):

```
params: BountyCreationParameters
{
  oracle: {
    Member: 0
  }
  contract_type: {
    Closed: [
      0
    ]
  }
  creator: {
    Member: 3,610
  }
  cherry: 500
  entrant_stake: 500
  funding_type: {
    Perpetual: {
      target: 5,000
    }
  }
  work_period: 100
  judging_period: 50
}
```

The transaction was accepted, but the processor, built from [2d51001](https://github.com/Joystream/joystream/commit/2d510011691cbd2220dd6b02db8f2df1ae9ff7a5), crashed with:

    INFO: There are already processed blocks in the database. The indexer will continue from block 96520.
    INFO: Last block: 96519 	: 4993 blocks behind
    INFO: Starting the event queue
    INFO: Processing block: 0000096520-24f40, events count: 4 
    (node:40) UnhandledPromiseRejectionWarning: QueryFailedError: current transaction is aborted, commands ignored until end of transaction block
        at new QueryFailedError (/joystream/node_modules/typeorm/error/QueryFailedError.js:12:28)
        at PostgresQueryRunner.<anonymous> (/joystream/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:248:31)
        at step (/joystream/node_modules/tslib/tslib.js:143:27)
        at Object.throw (/joystream/node_modules/tslib/tslib.js:124:57)
        at rejected (/joystream/node_modules/tslib/tslib.js:115:69)
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
    (Use `node --trace-warnings ...` to show where the warning was created)
    (node:40) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
    (node:40) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
    ERROR: Stopping the proccessor due to errors: {
[...]

PR #3460, opened 2022-03-20
query node - bountry entrant whitelist fix

ondratra:
Fixes https://github.com/Joystream/joystream/issues/3453.

The simplest way to test it is to use already deployed indexer:
```
set -e

# ensure mappings are built up to date
yarn
yarn workspace query-node-mappings build

# use external indexer
export PROCESSOR_INDEXER_GATEWAY=https://18.207.235.254.nip.io/query-node/indexer/graphql

# reset db and start processor
docker-compose up -d db
yarn db:drop && yarn db:prepare && yarn db:migrate # reset db
docker-compose up -d processor

docker logs processor # everything is fixed if no errors are present in logs
```

### DIFF
--- a/query-node/mappings/src/bounty.ts
+++ b/query-node/mappings/src/bounty.ts
@@ -203,7 +203,7 @@ export async function bounty_BountyCreated({ event, store }: EventContext & Stor
   // Create the EntrantWhitelist
   const entrantWhitelist = asEntrantWhitelist(bountyParams.contract_type)
   if (entrantWhitelist) {
-    store.save<BountyEntrantWhitelist>(entrantWhitelist)
+    await store.save<BountyEntrantWhitelist>(entrantWhitelist)
   }
 
   // Create the bounty


mnaamani (approved):

bedeho:
What was the actual problem though? a description would have been useful.

mnaamani:
> What was the actual problem though? a description would have been useful.

It was a race condition bug.
The `async` call to update the  database with the whitelist was missing the `await` keyword. So the function progressed and the following database update was trying to add a bounty record referencing the whitelist record which was not yet stored resulting in the error seen in the issue created: `"detail": "Key (entrant_whitelist_id)=(000000) is not present in table \"bounty_entrant_whitelist\".",`

bedeho:
I see, that seems like its a bad interface generated for the purposes of interacting with the database in Hydra, should we fix it in Hydra, or perhaps introduce some sort of linter check that at least makes sure our mappings never have this?

mnaamani:
> I see, that seems like its a bad interface generated for the purposes of interacting with the database in Hydra, should we fix it in Hydra, or perhaps introduce some sort of linter check that at least makes sure our mappings never have this?

I think this linter rule might be what we are looking for: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

A fix in hydra would be the better approach.

ondratra:
This is a quite common Typescript problem. Sometimes developers mistakenly omit `await` keyword when promise resolution is expected. 

> I think this linter rule might be what we are looking for: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md


I didn't know that this handy directive existed. Excellent, we can ensure now this kind of error won't happen again. I've created an issue for this https://github.com/Joystream/joystream/issues/3461

Issue #3330, opened 2022-03-01
Make all constants configurable in node


bedeho:
Currently there are a number of constants that have inlined values, however, it would be useful to be able to set new values for specific scenarios, such as in testing environments.

mnaamani:
I propose we use conditional compilation using [cargo features](https://doc.rust-lang.org/cargo/reference/features.html) to define 3 sets of runtime configurations/parameters to cater for different environments:

`Production` for production live network (default if no explicit selection is made)
`Staging` for pre-release staging networks and playgrounds
`Testing` for running automated integration tests

move out all parameter declarations and constants from `runtime/src/lib.rs` and `runtime/src/constants.rs` into a "config_production.rs", "config_staging.rs" and "config_testing.rs" modules.
common constants that are not environment specific can be placed in a "config_common.rs"

`runtime/src/lib.rs` will then import conditionally the appropriate parameters from the config_x.rs module

It would also be advisable to have different runtime `spec name` and `authoring version` number for each build type.

For proposals parameters configuration we currently optionally parse a JSON string from the environment at compile time. We should also drop this technique which overly complicates build and deployment in addition to being overly complex in implementation. Further, I think it is sufficient to have different sets of parameters for the different scenarios, and fine grained control seems unnecessary. (At least that has been the experience during pre-release testing)

In the end the different runtime configurations would be producing like so:

`cargo build --release` // production
`cargo build --release --features "staging_env"` // staging
`cargo build --release --features "testing_env"` // testing

mnaamani:
what do you think @shamil-gadelshin about this approach ?

shamil-gadelshin:
It could work! 

If we choose this approach we should implement 'mutual exclusive feature barrier': https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

There is a very close alternative: to use environment variable flags only to choose which `constant.rs` file to use (DEV, TEST, etc) to avoid feature clash.

Also, we should run checks&tests for each configuration that could be an additional burden for our CI.

bedeho:
FYI: Whatever scheme we pick, I hope it will be very light weight to add/remove constants, because I suspect many values which currently are constants will actually become storage variables that will be updatable with proposals. Allowing people to easily provide initial values when launching a playground for such configurable storage values would arguably be just as important.

mnaamani:
[...]

PR #3462, opened 2022-03-21
Runtime profiles

### LINKED ISSUE #3451: Olympia Runtime Params
Unless noted otherwise, all changes from `runtime/src/lib.rs`

## Proposals General
[file](runtime/src/proposals_configuration/defaults.rs)
| proposalType                                  | voting_period        | grace_period          | appr_q% | appr_t% | slash_q% | slash_t% | required_stake           | const |
|-----------------------------------------------|----------------------|-----------------------|---------|---------|----------|----------|--------------------------|-------|
| set_max_validator_count_proposal              | 43200                | 0                     | 66      | 80      | 60       | 80       | 100000                   | 1     |
| runtime_upgrade_proposal                      | ~~72000~~,**28800**  | ~~72000~~,**14400**   | 80      | 100     | 60       | 80       | ~~1000000~~,**10000000** | 1     |
| signal_proposal                               | 72000                | 0                     | 60      | 80      | 60       | 80       | 25000                    | 1     |
| funding_request_proposal                      | 72000                | ~~14400~~,**0**       | 60      | 80      | 60       | 80       | 25000                    | 1     |
| create_working_group_lead_opening_proposal    | 72000                | 0                     | 60      | 80      | 60       | 80       | 100000                   | 1     |
| fill_working_group_lead_opening_proposal      | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| update_working_group_budget_proposal          | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| decrease_working_group_lead_stake_proposal    | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| slash_working_group_lead_proposal             | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_working_group_lead_reward_proposal        | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| terminate_working_group_lead_proposal         | 72200                | 0                     | 66      | 80      | 60       | 80       | 100000                   | 1     |
| amend_constitution_proposal                   | 72200                | ~~72000~~,**14400**   | 80      | 100     | 60       | 80       | 1000000                  | **2** |
| cancel_working_group_lead_opening_proposal    | 43200                | 0                     | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_membership_price_proposal                 | 43200                | ~~14400~~**0**        | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_council_budget_increment_proposal         | 72000                | ~~43200~~**100800**   | 66      | 80      | 60       | 80       | 200000                   | 1     |
| set_councilor_reward_proposal                 | 72000                | ~~201600~~,**100800** | 66      | 80      | 60       | 80       | 200000                   | 1     |
| set_initial_invitation_balance_proposal       | 72000                | ~~43200~~**0**        | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_membership_lead_invitation_quota_proposal | 72000                | ~~43200~~**0**        | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_referral_cut_proposal                     | 43200                | ~~14400~~**0**        | 60      | 75      | 60       | 80       | 50000                    | 1     |
| set_invitation_count_proposal                 | 72000                | ~~43200~~**0**        | 60      | 75      | 60       | 80       | 50000                    | 1     |
| create_blog_post_proposal                     | 72000                | ~~14400~~**0**        | 60      | 80      | 60       | 80       | 25000                    | 1     |
| edit_blog_post_proposal                       | 72000                | ~~14400~~**0**        | 60      | 80      | 60       | 80       | 25000                    | 1     |
| lock_blog_post_proposal                       | 72000                | ~~14400~~**0**        | 60      | 80      | 60       | 80       | 25000                    | 1     |
| unlock_blog_post_proposal                     | 72000                | ~~14400~~**0**        | 60      | 80      | 60       | 80       | 25000                    | 1     |
| veto_proposal_proposal                        | ~~72200~~,**100800** | 0                     | 75      | 80      | 60       | 66       | 1000000                  | 1     |

**Purpose:**

* [ ] `runtime_upgrade_proposal` - This is for Jsgenesis' convenience. With these values, we can make propose a runtime upgrade, and have it execute >72h later.
* [ ] `funding_request_proposal` - Although there are arguments for keeping some grace here, the convenience for the users outweigh the risk. In an ideal world, we'd like to have the `grace_period` be a function of the spending, but hard to get right...
* [ ] `amend_constitution_proposal` - Let's rather have use the `constitutionality` than the `grace_period` to "secure" this.
* [ ] `set_membership*` - No grace needed.
* [ ] `set_council*` - Let's make these a council terms worth. Could use constitutionality, but let's rather vary.
* [ ] `set_initial_invitation_balance_proposal`,`set_referral_cut_proposal`,`set_invitation_count_proposal` - No grace needed.
* [ ] `*blog_post_proposal` - No grace needed.
* [ ] `veto_proposal_proposal` - The `voting_period` needs to match the longest `grace_period`

### MinUnstakingPeriodLimit
`pub const MinUnstakingPeriodLimit: u32 = 43200;`

* [ ] **Preferably** change the runtime constraint from `>` to `>=43200`
This should be the case already, with #3370:
[file](/runtime-modules/working-group/src/checks.rs)

```
    ensure!(
        stake_policy.leaving_unstaking_period >= T::MinUnstakingPeriodLimit::get(),
        Error::<T, I>::UnstakingPeriodLessThanMinimum
    );
```

* [ ] **If** there are any possible secondary effects of the above, instead:
`pub const MinUnstakingPeriodLimit: u32 = 43199;`

**Purpose:**
Annoying with an effective 43201 limit :)

Note that for:

```
pub const MinimumApplicationStake: Balance = 2000;
pub const LeaderOpeningStake: Balance = 2000;
```

It's already `>=2000`

### LeaderOpeningStake
`pub const LeaderOpeningStake: Balance = 2000;`

* [ ] Change to `2500000`

**Purpose:**
This should ideally be optimized for each group, but let's instead use a somewhat higher "bad" value, so there is some downside risk for bad actors.
For workers, it's fine to give the Lead lots of wiggle room.

### GatewayRewardPeriod/DistributionRewardPeriod
Both are:

```
    pub const GatewayRewardPeriod: u32 = 14400 + 50;
    pub const DistributionRewardPeriod: u32 = 14400 + 50;
```

* [ ] Change one of them to `14400+90`:
This should be the case already, with #3370.

**Purpose:**
Error?

### Term Length
```
pub const VoteStageDuration: BlockNumber = 100;
pub const RevealStageDuration: BlockNumber = 50;
pub const IdlePeriodDuration: BlockNumber = 400;
pub const AnnouncingPeriodDuration: BlockNumber = 200;
```

* [ ] Change to:

```
pub const VoteStageDuration: BlockNumber = 14400;
pub const RevealStageDuration: BlockNumber = 14400;
pub const IdlePeriodDuration: BlockNumber = 57600;
pub const AnnouncingPeriodDuration: BlockNumber = 28800;
```

**Purpose:**

* Gives a week Council (57600+14400+14400+14400 = 100800)
* Avoids a too long "dead" period when we launch the new network.

### ElectedMemberRewardPeriod
`pub const ElectedMemberRewardPeriod: BlockNumber = 10;`
I assume this is the Council Members equivalent to `GatewayRewardPeriod` etc, although they are in `u32`.

* [ ] Change to:
`pub const ElectedMemberRewardPeriod: BlockNumber = 14400;`

**Purpose:**

* A council term will likely always be divisible by 14400, so it seems preferable to not add `n` blocks at the end.

### Council Budget
```
pub const DefaultBudgetIncrement: u64 = 1000;
pub const BudgetRefillPeriod: BlockNumber = 1000;
```

* [ ] Change to:

```
pub const DefaultBudgetIncrement: u64 = 5000000;
pub const BudgetRefillPeriod: BlockNumber = 14400;
```

**Purpose:**
Back of the envelope calculations...

## Bounties
```
    pub const ClosedContractSizeLimit: u32 = 50;
    pub const MinCherryLimit: Balance = 10;
    pub const MinFundingLimit: Balance = 10;
    pub const MinWorkEntrantStake: Balance = 100;
```

* [ ] Change to:

```
    pub const ClosedContractSizeLimit: u32 = 50;
    pub const MinCherryLimit: Balance = 1000;
    pub const MinFundingLimit: Balance = 1000;
    pub const MinWorkEntrantStake: Balance = 1000;
```
AND/OR
`pub const ExistentialDeposit: u128 = 10;`
- [ ] Change to:
`pub const ExistentialDeposit: u128 = 1;`

**Purpose:**
Considering 1JOY is so little, it might not be worth changing this, but the combination of these are pretty bad. I think we'll get a lot of questions if we go directly to 10, so lets leave it at 1.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201991222576407) by [Unito](https://www.unito.io)

mnaamani:
Fixes #3451 and #3330

Introduced 3 new cargo features for the runtime build, `staging_runtime`, `playground_runtime` and `testing_runtime`. (I think I broke rust convention by using underscore instead of hyphen?)

Without specifying any of those features the "production" live configuration is applied.

To test:
```sh
# Test and build production node and runtime
yarn cargo-checks
yarn cargo-build

# Test and build staging node and runtime
# Shorter values for proposal params, smaller council
export RUNTIME_PROFILE=STAGING
yarn cargo-checks
yarn cargo-build

# Test and build playground node and runtime
#  Identical to staging except the CouncilSize=1
export RUNTIME_PROFILE=PLAYGROUND
yarn cargo-checks
yarn cargo-build

# Test and build "testing" node and runtime
# Block interval is 1000ms instead of 6000ms and short values for
# proposal params and council/election periods
export RUNTIME_PROFILE=TESTING
yarn cargo-checks
yarn cargo-build
```

I added github workflow as suggested by Shamil to test all the runtime profiles: https://github.com/Joystream/joystream/runs/5640223092?check_suite_focus=true

And docker builds (linux/amd64 only) for the development profiles.


mnaamani:
Seems setting MinNumberOfExtraCandidates to zero broke the benchmarks @ondratra 
https://github.com/Joystream/joystream/runs/5640222896?check_suite_focus=true#step:7:27

runtime/src/lib.rs
```diff
@@ -503,6 +512,29 @@ pub type ReferendumInstance = referendum::Instance1;
...
+    pub const MaxSaltLength: u64 = 32;
+    pub const VoteStageDuration: BlockNumber = 14400;
+    pub const RevealStageDuration: BlockNumber = 14400;
+    pub const MinimumVotingStake: u64 = 10000;
+
+    // council parameteres
+    pub const MinNumberOfExtraCandidates: u64 = 0;
+    pub const AnnouncingPeriodDuration: BlockNumber = 14400;
+    pub const IdlePeriodDuration: BlockNumber = 57600;
+    pub const CouncilSize: u64 = 12;
```
bwhm: This should go back to 5!

runtime/src/lib.rs
```diff
@@ -503,6 +512,29 @@ pub type ReferendumInstance = referendum::Instance1;
...
+    pub const MinCandidateStake: u64 = 1000;
+    pub const ElectedMemberRewardPeriod: BlockNumber = 14400;
+    pub const DefaultBudgetIncrement: u64 = 5000000;
+    pub const BudgetRefillPeriod: BlockNumber = 14400;
+    pub const MaxWinnerTargetCount: u64 = 15; // should be greater than council size
```
bwhm: From what I can tell, this will only impact an election if
`MaxWinnerTargetCount < CouncilSize + MinNumberOfExtraCandidates`.
So feel free to keep it :)

runtime/src/lib.rs
```diff
@@ -503,6 +512,29 @@ pub type ReferendumInstance = referendum::Instance1;
...
+    pub const VoteStageDuration: BlockNumber = 14400;
+    pub const RevealStageDuration: BlockNumber = 14400;
+    pub const MinimumVotingStake: u64 = 10000;
+
+    // council parameteres
+    pub const MinNumberOfExtraCandidates: u64 = 0;
+    pub const AnnouncingPeriodDuration: BlockNumber = 14400;
+    pub const IdlePeriodDuration: BlockNumber = 57600;
+    pub const CouncilSize: u64 = 12;
+    pub const MinCandidateStake: u64 = 1000;
```
bwhm: Although I didn't request this one, I think it's fine :)

runtime/src/lib.rs
```diff
@@ -503,6 +512,29 @@ pub type ReferendumInstance = referendum::Instance1;
...
 pub type CouncilModule = council::Module<Runtime>;
 
+// Production coucil and elections configuration
+#[cfg(not(any(feature = "staging_runtime", feature = "testing_runtime")))]
+parameter_types! {
+    // referendum parameters
+    pub const MaxSaltLength: u64 = 32;
+    pub const VoteStageDuration: BlockNumber = 14400;
+    pub const RevealStageDuration: BlockNumber = 14400;
+    pub const MinimumVotingStake: u64 = 10000;
```
bwhm, line 517: Not sure about the mismatch between `MinimumVotingStake` and `MinCandidateStake`, but let's avoid fiddling more than necessary now!

bwhm (changes requested):
Please revert `CouncilSize` back to 5.

The rest seems fine, although I'm not entirely confident in my understanding of how the dynamic between `MinNumberOfExtraCandidates` and `MaxWinnerTargetCount` works...

runtime/src/lib.rs
mnaamani: I'll revert to original values since we tested with these values.

shamil-gadelshin (changes requested):
Looks good, I tested some configurations - they work. However, the bencmarking failed (at least the previous run). 

I have a couple of comments:
- Did you consider placing all the feature-gated code in a single file to simplify future configuration changes?
- `cargo build --release --features staging_runtime,testing_runtime` this command fails. It's possible to avoid it properly but I'm not sure it worth the trouble.

runtime/src/lib.rs
```diff
@@ -91,10 +95,15 @@ pub use content::MaxNumber;
 
 /// This runtime version.
 pub const VERSION: RuntimeVersion = RuntimeVersion {
+    #[cfg(not(any(feature = "staging_runtime", feature = "testing_runtime")))]
     spec_name: create_runtime_str!("joystream-node"),
+    #[cfg(feature = "staging_runtime")]
+    spec_name: create_runtime_str!("joystream-node-staging"),
+    #[cfg(feature = "testing_runtime")]
+    spec_name: create_runtime_str!("joystream-node-testing"),
```
Lezek123: Note that this would make `polkadot/apps` no longer recognize staging/testing/playground config nodes as Joystream nodes for the purpose of injecting the types unless we also introduce this change in https://github.com/polkadot-js/apps/pull/7134 (which of course can be done).

.github/workflows/joystream-node-docker-dev.yml
```diff
@@ -0,0 +1,55 @@
...
+          echo "::set-output name=image_exists::${IMAGE_EXISTS}"
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: joystream-node.Dockerfile
+          platforms: linux/amd64
+          push: true
+          tags: ${{ env.REPOSITORY }}:${{ steps.compute_shasum.outputs.shasum }}
```
Lezek123: Don't you need to pass a `buildArg` here depending on the value of `RUNTIME_PROFILE` env?
I tested the published `TESTING` docker image and it's still using `6s` blocktime which would confirm that this is missing.

runtime/src/lib.rs
```diff
@@ -503,6 +512,29 @@ pub type ReferendumInstance = referendum::Instance1;
...
+    pub const CouncilSize: u64 = 5;
+    pub const MinCandidateStake: u64 = 11000;
+    pub const ElectedMemberRewardPeriod: BlockNumber = 14400;
+    pub const DefaultBudgetIncrement: u64 = 5000000;
+    pub const BudgetRefillPeriod: BlockNumber = 14400;
+    pub const MaxWinnerTargetCount: u64 = 10; // should be greater than council size
+}
+
+// Common staging and testing coucil and elections configuration
+#[cfg(any(feature = "staging_runtime", feature = "testing_runtime"))]
```
Lezek123: Now that we can easily configure those I would suggest using much shorter periods for `testing_runtime` (like `20 blocks` for each stage)

scripts/cargo-build.sh
```diff
@@ -1,7 +1,18 @@
...
+  FEATURES="testing_runtime"
+fi
+
+if [[ "$RUNTIME_PROFILE" == "STAGING" ]]; then
+  FEATURES="staging_runtime"
+fi
+
+if [[ "$RUNTIME_PROFILE" == "PLAYGROUND" ]]; then
+  FEATURES="playground_runtime"
+fi
```
Lezek123: Is it worth adding to `run-dev-chain.sh` script too?
If so, perhaps it's worth refactoring those settings to a separate script? (since they're also used by `/devops/git-hooks/pre-push`)

I suggest to make the profile configurable for `start.sh` also.
It is currently only possible to use `production` profile for `start.sh`, because the env is not passed to `runtime-code-shasum.sh`.

Lezek123 (changes requested):
Very excited about this, it is an invaluable improvement! :rocket:

Lezek123:
One more suggestion: Please remove the no longer relevant `./query-node/run-tests.sh` script.

runtime/src/lib.rs
mnaamani: Fair point, I considered modifying the runtime authoring version instead of the spec name.
The only reason I wanted to make this change was prevent mistake in doing a runtime upgrade with a wasm built for a different profile. 

> Looks good, I tested some configurations - they work. However, the bencmarking failed (at least the previous run).
> 
> I have a couple of comments:
> 
>     * Did you consider placing all the feature-gated code in a single file to simplify future configuration changes?
> 

I did but I wanted to keep the refactoring of the main lib.rs to a minimum although I had proposed to do so in the intial issue. Perhaps this can be don in a future PR.

>     * `cargo build --release --features staging_runtime,testing_runtime` this command fails. It's possible to avoid it properly but I'm not sure it worth the trouble.

Yes it fails with the compiler error at the top of lib.rs as they are mutually exclusive features. Is there a better way?

scripts/cargo-build.sh
mnaamani: Good suggestion

.github/workflows/joystream-node-docker-dev.yml
mnaamani: oh yes I forgot I was not using our own build scripts. will fix

runtime/src/lib.rs
mnaamani: Excellent idea.

mnaamani:
> However, the bencmarking failed (at least the previous run).

Yes I asked Andy to take a look, it fails when minimum extra candidates is set to zero.

mnaamani:
> One more suggestion: Please remove the no longer relevant `./query-node/run-tests.sh` script.

Actually @ondratra is using it on rhodes branch to test specific scenario so will keep it. I did update the main repo README to suggest running full tests from tests/network-tests/run-full-tests.sh

scripts/cargo-build.sh
mnaamani: Well actually for start.sh you can run with different profiles because `scripts/runtime-code-shasum.sh` does use the RUNTIME_PROFILE from env.
So you can do:

```
export RUNTIME_PROFILE=TESTING
./build-node-docker.sh
./start.sh
```

.github/workflows/joystream-node-docker-dev.yml
mnaamani: done in 7c58d76e3a23d797b13cc9acd2e1ddab9fead4c8 and 463b735aa8b915dde027a564d5703d3730a84725

runtime/src/lib.rs
mnaamani: Removed any change in runtime version based on runtime profile.

runtime/src/lib.rs
mnaamani: done in 201a5c999869c2f98af9bb06809bb6074942a1e7

scripts/cargo-build.sh
mnaamani: factored our the FEATURES assignment into a separate `features.sh` that is `source`ed from the bash scripts that need it. Done in 9cba30c3aeaaf273331ed4535e4b09baf0d91ac7

PR #3466, opened 2022-03-22
Small CLI improvements, and added staking.validate

bwhm:
Changes:
- Renamed the Operations Working Group in the CLI
- Added `validate.staking`, as I was under the impression this would not work in `polkadot-js`
   - If the code is ok, I suggest we can keep it, as I'm planning to have the community expand on this
- Changed the minimum unstaking period to `43200` to be in line with the runtime.
   - Note that these should ideally be read from the state, but I don't think I would be able to add that in a good way

### DIFF
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -415,7 +415,7 @@ export default class Api {
     return await this.entriesByAccountIds(this._api.query.staking.ledger)
   }
 
-  async isControllerValid(account: string) {
+  async isControllerValid(account: string): Promise<Option<StakingLedger>> {
     return await this._api.query.staking.ledger(account)
   }
 
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -152,8 +152,6 @@ export default class Api {
     return entries
   }
 
-
-
   protected async blockHash(height: number): Promise<string> {
     const blockHash = await this._api.rpc.chain.getBlockHash(height)
 
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -415,7 +415,7 @@ export default class Api {
     return await this.entriesByAccountIds(this._api.query.staking.ledger)
   }
 
-  async isControllerValid(account: string): Promise<Option<StakingLedger>> {
+  async getStakingLedger(account: string): Promise<Option<StakingLedger>> {
     return await this._api.query.staking.ledger(account)
   }
 
--- a/cli/src/Api.ts
+++ b/cli/src/Api.ts
@@ -3,10 +3,10 @@ import { createType, types } from '@joystream/types'
 import { ApiPromise, WsProvider } from '@polkadot/api'
 import { SubmittableExtrinsic, AugmentedQuery } from '@polkadot/api/types'
 import { formatBalance } from '@polkadot/util'
-import { Balance, LockIdentifier } from '@polkadot/types/interfaces'
+import { Balance, LockIdentifier, StakingLedger, ElectionStatus } from '@polkadot/types/interfaces'
 import { KeyringPair } from '@polkadot/keyring/types'
 import { Codec, Observable } from '@polkadot/types/types'
-import { UInt } from '@polkadot/types'
+import { UInt, Option } from '@polkadot/types'
 import {
   AccountSummary,
   WorkingGroups,
@@ -48,9 +48,9 @@ export const apiModuleByGroup = {
   [WorkingGroups.Forum]: 'forumWorkingGroup',
   [WorkingGroups.Membership]: 'membershipWorkingGroup',
   [WorkingGroups.Gateway]: 'gatewayWorkingGroup',
-  [WorkingGroups.OperationsAlpha]: 'operationsWorkingGroupAlpha',
-  [WorkingGroups.OperationsBeta]: 'operationsWorkingGroupBeta',
-  [WorkingGroups.OperationsGamma]: 'operationsWorkingGroupGamma',
+  [WorkingGroups.Builders]: 'operationsWorkingGroupAlpha',
+  [WorkingGroups.HumanResources]: 'operationsWorkingGroupBeta',
+  [WorkingGroups.Marketing]: 'operationsWorkingGroupGamma',
   [WorkingGroups.Distribution]: 'distributionWorkingGroup',
 } as const
 
@@ -140,6 +140,19 @@ export default class Api {
 
     return entries.sort((a, b) => a[0].toNumber() - b[0].toNumber())
   }
+
+  async entriesByAccountIds<AccountType extends AccountId, ValueType extends Codec>(
+    apiMethod: AugmentedQuery<'promise', (key: AccountType) => Observable<ValueType>, [AccountType]>
+  ): Promise<[AccountType, ValueType][]> {
+    const entries: [AccountType, ValueType][] = (await apiMethod.entries()).map(([storageKey, value]) => [
+      storageKey.args[0] as AccountType,
+      value,
+    ])
+
+    return entries
+  }
+
+
 
   protected async blockHash(height: number): Promise<string> {
     const blockHash = await this._api.rpc.chain.getBlockHash(height)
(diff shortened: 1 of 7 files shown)


cli/src/schemas/WorkingGroups.ts
```diff
@@ -49,7 +49,7 @@ export const WorkingGroupOpeningInputSchema: JsonSchema<WorkingGroupOpeningInput
       required: ['amount', 'unstakingPeriod'],
       properties: {
         amount: { type: 'integer', minimum: 2000 },
-        unstakingPeriod: { type: 'integer', minimum: 43201 },
+        unstakingPeriod: { type: 'integer', minimum: 43200 },
```
traumschule, line 52: [api.consts.staking.bondingDuration](https://github.com/Joystream/joystream/blob/master/pioneer/packages/page-staking/src/Actions/useUnbondDuration.ts#L21)

cli/src/Types.ts
```diff
@@ -55,9 +55,9 @@ export enum WorkingGroups {
   Curators = 'curators',
   Forum = 'forum',
   Membership = 'membership',
-  OperationsAlpha = 'operationsAlpha',
-  OperationsBeta = 'operationsBeta',
-  OperationsGamma = 'operationsGamma',
+  Builders = 'builders',
```
mnaamani, line 58: I don't think it is safe to change these, they should match the runtime name of the group?

mnaamani:
to fix CI checks format code and check-in changes with:

```sh
cd cli/
yarn format
```

cli/src/schemas/WorkingGroups.ts
bwhm: Hey,
That only applies to staking for validators! I know how to get these, but they could be different both by group, and opening type (Lead vs Worker). If I were to try, it would take some non-trivial amount of time, and would likely not be very clean :)

cli/src/Types.ts
Lezek123: They can be customized, since CLI uses mappings like `apiModuleByGroup`

bwhm: Admittedly, I didn't test more than a couple groups after the change, but `operationsAlpha` isn't the runtime name of the group either.  I think these are just "shorts" used for the users convenience, whereas the actual runtime reference are from `cli/src/Api.ts`.

cli/src/commands/staking/validate.ts
```diff
@@ -0,0 +1,37 @@
+import StakingCommandBase from '../../base/StakingCommandBase'
+import { flags } from '@oclif/command'
+
+
+export default class StakingValidateCommand extends StakingCommandBase {
+  static description = 'Start validating. Takes the controller key.'
+  static args = [
```
Lezek123: I was hoping we'll move away from `args` completely in favor of `flags`, since they are just easier to maintain and more typesafe

cli/src/base/StakingCommandBase.ts
```diff
@@ -0,0 +1,91 @@
...
+      this.warn(`You don't have this account imported!`)
+      this.exit(ExitCodes.NoAccountFound)
+    }
+  }
+
+  async promptForCommission(message = 'Choose how much you reqiure as commission (between 0% and 100%'): Promise<number> {
+    return await this.simplePrompt<number>({ message, type: 'number'})
+  }
+
+  async electionStatus() {
```
Lezek123: `checkElectionStatus`

cli/src/commands/staking/validate.ts
```diff
@@ -0,0 +1,37 @@
...
+    let { controller } = this.parse(StakingValidateCommand).flags
+    await this.electionStatus()
+    const validatorPrefs = await this.getValidatorPrefs(commission)
+
+    if (controller === undefined) {
+      controller = await this.getController()
+    } else {
+      await this.isController(controller)
+    }
+    await this.electionStatus()
```
Lezek123: Does it make sense to check it twice?

cli/src/Api.ts
```diff
@@ -400,6 +413,18 @@ export default class Api {
     return this.entriesByIds(this._api.query.content.curatorGroupById)
   }
 
+  async allStakingLedgers(): Promise<[AccountId, Option<StakingLedger>][]> {
+    return await this.entriesByAccountIds(this._api.query.staking.ledger)
+  }
+
+  async isControllerValid(account: string) {
```
Lezek123:
```suggestion
  async isControllerValid(account: string): Promise<Option<StakingLedger>> {
```
Also, the name suggest it returns a boolean, so I suggest either renaming the function or adding `isSome` and changing the return type.

cli/src/base/StakingCommandBase.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.warn('Your account is not a controller!')
+        this.exit(ExitCodes.InvalidInput)
+      }
+    } else {
+      this.warn(`You don't have this account imported!`)
+      this.exit(ExitCodes.NoAccountFound)
+    }
+  }
+
+  async promptForCommission(message = 'Choose how much you reqiure as commission (between 0% and 100%'): Promise<number> {
```
Lezek123:
```suggestion
  async promptForCommission(message = 'Choose how much you reqiure as commission (between 0% and 100%)'): Promise<number> {
```

cli/src/base/StakingCommandBase.ts
```diff
@@ -0,0 +1,91 @@
...
+
+
+  private async getStakingLedgers() {
+    const controllers = await this.getApi().allStakingLedgers()
+    return controllers
+  }
+
+  async getController():Promise<string> {
+    const availableControllers = await this.availableControllers()
+    const selectedValidator = await this.promptForController('No controller input',availableControllers)
```
Lezek123: Why is the prompt message set to `No controller input`?

cli/src/base/StakingCommandBase.ts
```diff
@@ -0,0 +1,91 @@
...
+  private async availableControllers(): Promise<[AccountId,StakingLedger][]> {
+    const controllers = await this.getStakingLedgers()
+    const stakerMap: [AccountId,StakingLedger][] = []
+    controllers.forEach(([key,value]) => {
+      const ledger = value.unwrapOr(undefined)
+      if (!!ledger && this.isKeyAvailable(key)) {
+        stakerMap.push([key,ledger])
+      }
+    })
+    return stakerMap
```
Lezek123:
```suggestion
    return controllers.filter(([key,value]) => value.isSome && this.isKeyAvailable(key)).map(([key, value]) => [key, value.unwrap()])
```

Lezek123 (changes requested):
Great job, just added some small suggestions.

Lezek123:
Also, please add `staking` commands description in `cli/package.json` under `oclif` -> `topics`

cli/src/commands/staking/validate.ts
bwhm: I'll change it then :)

cli/src/base/StakingCommandBase.ts
bwhm: :+1:

cli/src/commands/staking/validate.ts
bwhm: Probably not, but for (future) commands that may take a little more time for the user to complete, it would probably be good to do one before they start. 

I'm not sure how useful the error message will be if it changed while the user is doing their thing, so I added an extra run at the end to "ensure" they get the message clearly.

Ideally, I guess I should calculate approximately when the `staking.*` commands will be "enabled" again, and rather just warn at initiation.

cli/src/Api.ts
bwhm: :+1:

cli/src/base/StakingCommandBase.ts
bwhm: Good question!

bwhm:
FYI @Lezek123 
I did what @mnaamani suggested - `yarn format`, which broke all of your commit suggestions. Wasn't trying to take away the credit from you 🗡️

cli/src/base/StakingCommandBase.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private async getStakingLedgers() {
+    const controllers = await this.getApi().allStakingLedgers()
+    return controllers
+  }
+
+  async getController(): Promise<string> {
+    const availableControllers = await this.availableControllers()
+    const selectedValidator = await this.promptForController('No controller input', availableControllers)
```
Lezek123, line 22:
```suggestion
    const selectedValidator = await this.promptForController(undefined, availableControllers)
```
I think the default prompt message (`Choose your controller account`) is more clear, but it's not a big issue.

Lezek123 (approved):
LGTM

PR #3425, opened 2022-03-16
runtime: content: Fix missing member_id check.

shamil-gadelshin:
There was a major bug in the bounty pallet. Most work entry related extrinsics didn't contain entry ownership verification.

#### Fixed extrinsics
- withdraw_work_entrant_funds
- submit_work
- withdraw_work_entry

[Related issue](https://github.com/Joystream/joystream/issues/3379)

Issue #3379, opened 2022-03-09
Bounty module, A worker can arbitrarily withdraw funds from another worker if he knows the respective entry id


L3pereira:
If you run the following test, two workers successfully contribute and they should receive reward / 2, but worker 1 not only tries to call withdraw_work_entrant_funds with entry id ,1 but also tries with entry id 2 (before worker 2), the result is that the entry ids are removed worker 1 gets the total reward and worker 2 receives 0.
Is this intentional? It seems a malicious worker can steal other workers' funds.

```rust
#[test]
fn withdraw_work_entrant_funds_steal() {
    build_test_externalities().execute_with(|| {
        let starting_block = 1;
        run_to_block(starting_block);

        let initial_balance = 500;
        let max_amount = 200;
        let winner_reward = max_amount;
        let entrant_stake = 37;
        let work_period = 1;

        set_council_budget(initial_balance);

        CreateBountyFixture::default()
            .with_max_funding_amount(max_amount)
            .with_work_period(work_period)
            .with_entrant_stake(entrant_stake)
            .call_and_assert(Ok(()));

        let bounty_id = 1;

        FundBountyFixture::default()
            .with_bounty_id(bounty_id)
            .with_amount(max_amount)
            .with_council()
            .with_origin(RawOrigin::Root)
            .call_and_assert(Ok(()));

        let member_id1 = 1;
        let account_id1 = 1;
        increase_account_balance(&account_id1, initial_balance);

        // Winner
        AnnounceWorkEntryFixture::default()
            .with_origin(RawOrigin::Signed(account_id1))
            .with_member_id(member_id1)
            .with_staking_account_id(account_id1)
            .with_bounty_id(bounty_id)
            .call_and_assert(Ok(()));

        assert_eq!(
            Balances::usable_balance(&account_id1),
            initial_balance - entrant_stake
        );

        let entry_id1 = 1;

        SubmitWorkFixture::default()
            .with_origin(RawOrigin::Signed(account_id1))
            .with_member_id(member_id1)
            .with_entry_id(entry_id1)
            .call_and_assert(Ok(()));

        // Legitimate participant
        let member_id2 = 2;
        let account_id2 = 2;
        increase_account_balance(&account_id2, initial_balance);

        AnnounceWorkEntryFixture::default()
            .with_origin(RawOrigin::Signed(account_id2))
            .with_member_id(member_id2)
            .with_staking_account_id(account_id2)
            .with_bounty_id(bounty_id)
[...]

PR #3482, opened 2022-03-24
Bounty perpetual working period, closes #2821, closes #2947, closes #3193, fixes #3446, fixes #3447, fixes #3329;

### LINKED ISSUE #2821: Built-in reward for bounty oracles
Bounty creator should be able to specify a fixed reward, as part of creating bounty, which is diverted towards the oracle, at least if its not the council, which is dispersed in scenarios where the oracle submits a judgement, otherwise it should be returned to bounty creator.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1201958578817342) by [Unito](https://www.unito.io)

### LINKED ISSUE #2947: Bounty oracle switch
An oracle may not be able to complete the work on managing a bounty, for any reason, and thus transfer role to someone else. The assumption is that this happens through some trust relationship, just as the initial choice of oracle, so no reason for double opt-in. This is only possible before judgement period is over.

**This only applies bounties where oracle is currently member, not council oracles, or coming working group oracles, however member oracles can change oracle to any other type:** https://github.com/Joystream/joystream/issues/2979

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703510692257) by [Unito](https://www.unito.io)

### LINKED ISSUE #3193: Flexibile working period duration
# Background

Currently, the working period for a bounty lasts for a fixer period of time from when the funding period is over. The problem with this is that

a. it may be hard to anticipate up-front during bounty creation, how long it will take for the right worker to show up and discover the bounty, in particular because the funding period can itself be highly variable in duration.
b. there may be one or more false starts where a worker begins, which may deter others, but then pulls out, while consuming parts of the working period.
c. there may be complications relating to the nature of the work, or the ability of an otherwise suitable candidate to perform the work.

At the same time, the total lack of constraints around the working period does leave the possible third party contributors at the mercy of the oracle in terms of not locking up their funds perpetually.

# Proposal

1. Allow bounties to be created with perpetual working periods, then backers at least know this up front. This will likely impact the whole slashing logic of entries that withdraw, probably should just be dropped in this type of bounty.
2. Same as 1, but also allow council to intervene and terminate a bounty in the working period (e.g. because the oracle becomes hostile or idle), basically equivalent effect as judgement period expiring.

### LINKED ISSUE #3329: Actor validation redundant
@shamil-gadelshin 
Aren't these lines redundant? I think they're validating twice 
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L784
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L1135
We could change these extrinsics by removing ```creator``` and ```oracle``` inputs, since we already have that information in the chain and we already do the necessary validation with  ```ensure_bounty_exists```
Something like this
```rust
        pub fn cancel_bounty(origin, bounty_id: T::BountyId) {

            let bounty = Self::ensure_bounty_exists(&bounty_id)?;
            BountyActorManager::<T>::ensure_bounty_actor_manager(
                origin,
            )?;
            ...
       }


        pub fn submit_oracle_judgment(
            origin,
            bounty_id: T::BountyId,
            judgment: OracleJudgment<T::EntryId, BalanceOf<T>>
        ) {
            let bounty = Self::ensure_bounty_exists(&bounty_id)?;
            BountyActorManager::<T>::ensure_bounty_actor_manager(
                origin,
                bounty.creation_params.oracle.clone(),
            )?;
            ...
      }

```

### LINKED ISSUE #3446: When a bounty succeeds, removing all BountyContributions by the last worker can be expensive.
When a bounty succeeds and there are  funds to be withdrawn by the workers, the last worker will call ```remove_bounty``` 
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L1243
the problem here is, inside ```remove_bounty``` an instruction tries to delete all BountyContributions entries related to a bounty id, this can be O(N) expensive
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L1393

As a solution, in the new bounty implementation, the funders will have to lock some stake amount besides the normal contribution.
If the bounty succeeds, funders will have an incentive to unlock their stakes, and in the process, they will remove their entries one by one.

### LINKED ISSUE #3447: Bounty blocked by T::MinFundingLimit in funding period
When a bounty is in funding stage and the funding period ends by eg. target reached, a funder is limited by ```T::MinFundingLimit::get()```, this can block the bounty in funding period.
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L859

This test shows that a bounty can be locked, if a funder contributes with 4951, no other funder can contribute the rest of the amount (49) as this is less than ```T::MinFundingLimit::get()``` (50), causing the target to be unreachable. 

```rust
#[test]
fn fund_bounty_fails_funding_Less_then_minimum_allowed() {
    build_test_externalities().execute_with(|| {
        set_council_budget(500);

        let max_amount = 5000;
        let amount = 100;
        let account_id = 1;
        let member_id = 1;
        let initial_balance = 500;
        let cherry = DEFAULT_BOUNTY_CHERRY;

        increase_total_balance_issuance_using_account_id(account_id, 6000);
        increase_total_balance_issuance_using_account_id(
            COUNCIL_BUDGET_ACCOUNT_ID,
            initial_balance,
        );

        CreateBountyFixture::default()
            .with_perpetual_funding(max_amount)
            .with_cherry(cherry)
            .call_and_assert(Ok(()));

        let bounty_id = 1u64;

        FundBountyFixture::default()
            .with_bounty_id(bounty_id)
            .with_amount(4951)
            .with_member_id(member_id)
            .with_origin(RawOrigin::Signed(account_id))
            .call_and_assert(Ok(()));

        FundBountyFixture::default()
            .with_bounty_id(bounty_id)
            .with_amount(49)
            .with_member_id(member_id)
            .with_origin(RawOrigin::Signed(account_id))
            .call_and_assert(Ok(()));
    });
}
```

L3pereira:
#2821 Built-in reward for bounty oracles

A new field was added to the bounty pallet "MinOracleRewardLimit"
When a creator creates a bounty he has to set a specific cherry and oracle reward,
This reward should be awarded to the oracle if he submits a judgment (with winners or not).

#2947 Bounty oracle switch

Bounty oracle can be switched by the current oracle or by the council
This can happen in the following stages  Funding, WorkSubmission, Judgment.

#3193 Flexible working period duration

working period and judging period have no longer a specific limit, after funding, workers can submit their work,
the oracle can end this period by calling ```end_working_period``` if there are no work submissions, the bounty fails and 
the oracle reward goes to the creator, on the opposite side if there are submissions, the bounty will advance to judgment.
In judgement, the oracle will select the winners, the losers and receive the oracle reward for that.
The winners will receive a reward and their stake accounts will be immediately unlocked.
The losers will have their stake accounts slashed (or not) by the oracle, a description can also be given, for example, with the reason for the slashing action.
It can be the case, the oracle becomes inactive and doesn't end the working period, if there are no entry submissions, the council can call ```terminate_bounty``` and the bounty fails. 
When work entries are present, the Council can switch the oracle and proceed with the normal process.

#3446 When a bounty succeeds, removing all BountyContributions by the last worker can be expensive.
#3447 Bounty blocked by T::MinFundingLimit in the funding period

If a bounty is successful all bounty contribution entries will be occupying space in the chain, if a funder starts sending a small amount from multiple accounts, he can perform a state bloat attack.  
The funder must have an incentive to delete those bounty contribution entries, a min fund limit doesn't work, a better solution is to use a state bloat bond, 
this is a quantity (at the time this is just a static variable) of joy the funder has to deposit (this is not stake) in the bounty. after the oracle successfully judges the work, each funder can 
call ```withdraw_state_bloat_bond_by_council``` and recover the state bloat bond.
A funder can make multiple contributions with the same membership and only has to deposit the bond once. 

#3329 Joystream/joystream Actor validation redundant

### DIFF
--- a/runtime-modules/bounty/src/actors.rs
+++ b/runtime-modules/bounty/src/actors.rs
@@ -58,7 +58,7 @@ impl<T: Trait> BountyActorManager<T> {
         }
     }
 
-    // Validate balance is sufficient for the bounty
+    /// Validate balance is sufficient for the bounty
     pub(crate) fn validate_balance_sufficiency(
         &self,
         required_balance: BalanceOf<T>,
@@ -85,7 +85,7 @@ impl<T: Trait> BountyActorManager<T> {
         T::CouncilBudgetManager::get_budget() >= amount
     }
 
-    // Validate that provided actor relates to the initial BountyActor.
+    /// Validate that provided actor relates to the initial BountyActor.
     pub(crate) fn validate_actor(&self, actor: &BountyActor<MemberId<T>>) -> DispatchResult {
         let initial_actor = match self {
             BountyActorManager::Council => BountyActor::Council,
@@ -97,7 +97,7 @@ impl<T: Trait> BountyActorManager<T> {
         Ok(())
     }
 
-    // Transfer funds for the bounty creation.
+    /// Transfer funds for the bounty creation.
     pub(crate) fn transfer_funds_to_bounty_account(
         &self,
         bounty_id: T::BountyId,
@@ -122,7 +122,7 @@ impl<T: Trait> BountyActorManager<T> {
         Ok(())
     }
 
-    // Restore a balance for the bounty creator.
+    /// Restore a balance for the bounty creator.
     pub(crate) fn transfer_funds_from_bounty_account(
         &self,
         bounty_id: T::BountyId,
--- a/runtime-modules/bounty/src/actors.rs
+++ b/runtime-modules/bounty/src/actors.rs
@@ -85,18 +85,6 @@ impl<T: Trait> BountyActorManager<T> {
         T::CouncilBudgetManager::get_budget() >= amount
     }
 
-    // Validate that provided actor relates to the initial BountyActor.
-    pub(crate) fn validate_actor(&self, actor: &BountyActor<MemberId<T>>) -> DispatchResult {
-        let initial_actor = match self {
-            BountyActorManager::Council => BountyActor::Council,
-            BountyActorManager::Member(_, member_id) => BountyActor::Member(*member_id),
-        };
-
-        ensure!(initial_actor == actor.clone(), Error::<T>::NotBountyActor);
-
-        Ok(())
-    }
-
     // Transfer funds for the bounty creation.
     pub(crate) fn transfer_funds_to_bounty_account(
         &self,
--- a/runtime-modules/bounty/src/actors.rs
+++ b/runtime-modules/bounty/src/actors.rs
@@ -58,7 +58,7 @@ impl<T: Trait> BountyActorManager<T> {
         }
     }
 
-    /// Validate balance is sufficient for the bounty
+    // Validate balance is sufficient for the bounty
     pub(crate) fn validate_balance_sufficiency(
         &self,
         required_balance: BalanceOf<T>,
@@ -85,7 +85,7 @@ impl<T: Trait> BountyActorManager<T> {
         T::CouncilBudgetManager::get_budget() >= amount
     }
 
-    /// Validate that provided actor relates to the initial BountyActor.
+    // Validate that provided actor relates to the initial BountyActor.
     pub(crate) fn validate_actor(&self, actor: &BountyActor<MemberId<T>>) -> DispatchResult {
         let initial_actor = match self {
             BountyActorManager::Council => BountyActor::Council,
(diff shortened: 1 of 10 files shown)


L3pereira:
This PR also have in consideration the following issues 
#3379 Bounty module, A worker can arbitrarily withdraw funds from another worker if he knows the respective entry id
#3425 runtime: content: Fix missing member_id check.

PR #3485, opened 2022-03-24
Bounty perpetual working period

### LINKED ISSUE #2821: Built-in reward for bounty oracles

### LINKED ISSUE #2947: Bounty oracle switch

### LINKED ISSUE #3038: Bounties: Add the work description to the `announce_work_entry` extrinsic.
Add the `description` field to the extrinsic and its event only (no storage change).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201810432142171) by [Unito](https://www.unito.io)

### LINKED ISSUE #3193: Flexibile working period duration

### LINKED ISSUE #3329: Actor validation redundant

### LINKED ISSUE #3446: When a bounty succeeds, removing all BountyContributions by the last worker can be expensive.

### LINKED ISSUE #3447: Bounty blocked by T::MinFundingLimit in funding period

### LINKED ISSUE #3489: Bounties: Oracle should transfer rewards to workers or slash their accounts at the judgement stage
When the oracle submits a judgement he already has to iterate the entries he is judging, when he decides an entry is a winner, the oracle must attribute a reward to that entry, if the oracle already knows the worker_member_id  and the amount to reward, he can get the controller worker account id and transfer that amount right away. 
If the oracle decides an entry is rejected he can slash the worker staking account immediately.

This saves the workers from having to call a new extrinsic and query the entries store.
This change makes the field `EntryRecord::oracle_judgmnt_result` useless 

If a worker submitted some work that didn't get judged, he can unlock his stake by calling `unlock_work_entrant_stake`

### LINKED ISSUE #3522: Oracle reward shouldn't be automatically paid out
Oracle reward is never automatically paid out, oracle must actively withdraw reward in all terminal cases, bounty therefore also cannot be removed before this has happened. Add a new global variable oracle_withrew_reward: bool, and a new extrinsic for this withdrawal. Remember that there may be 0 oracle reward.

### LINKED ISSUE #3523: New Terminate Bounty extrinsic, veto_bounty removal
A new extrinsic `terminate_bounty` should allow the council to terminate the bounty (as Failed Bounty) in 

- Funding Period w/o contributions
- Working Period
- Judging Period

`Veto_bounty` allows the Council to remove a bounty in the funding period without contributions, this should be removed as its functionality is very similar to terminate_bounty

### LINKED ISSUE #3545: State bloat bond requirement for creators
Add a state bloat bond requirement for creators, which is automatically paid when the bounty is removed in `remove_bounty`. 
The minimum amount is a constant. Described here: #3507

Since there will be an incentive for a creator to remove the bounty, we can drop `T::MinCherryLimit` and all associated validation.

### LINKED ISSUE #3546: Drop max_funding_limit & min_funding_limit from BountyCreationParameters
Dropping max_funding_limit & min_funding_limit from BountyCreationParameters would simplify the funding logic. 
The funding should have only a target limit

L3pereira:
* Closes #2821

- - -
* Closes #2947

Bounty oracle can be switched by the current oracle or by the council.
This can happen in the following stages:

*  Funding 
* WorkSubmission
* Judgment.

- - -
* Closes #3193
* Closes #3489

Working period and judging period have no longer a specific limit, after funding, workers can submit their work,
the oracle can end this period by calling end_working_period if there are no work submissions, the bounty fails and
the oracle reward goes to the creator, on the opposite side if there are submissions, the bounty will advance to judgment.
In judgement, the oracle will select the winners, the losers and receive the oracle reward for that, this reward can only be claimed by calling `withdraw_oracle_reward` at the end of judgment.
The winners will receive a reward and their stake accounts will be immediately unlocked.
The losers will have their stake accounts slashed (or not) by the oracle, a description can also be given, for example, with the reason for the slashing action.
It can be the case, the oracle becomes inactive and doesn't end the working period, if there are no entry submissions, the council can call terminate_bounty and the bounty fails.
When work entries are present, the Council can switch the oracle and proceed with the normal process.

- - -
* Fixes #3446
* Fixes #3447

If a bounty is successful all bounty contribution entries will be occupying space in the chain, if a funder starts sending a small amount from multiple accounts, he can perform a state bloat attack.
The funder must have an incentive to delete those bounty contribution entries, a min fund limit doesn't work, a better solution is to use a state bloat bond,
this is a quantity (at the time this is just a static variable) of joy the funder has to deposit (this is not stake) in the bounty. after the oracle successfully judges the work, each funder can
call withdraw_state_bloat_bond_by_council and recover the state bloat bond.
A funder can make multiple contributions with the same membership and only has to deposit the bond once.

- - -
This PR also have in consideration the following issues

* #3379 
* #3425
* Fixes #3329

- - -
* Closes #3523
* Closes #3522
* Closes #3546
* Closes #3545
* Closes #3038

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476088401771) by [Unito](https://www.unito.io)


runtime-modules/bounty/src/lib.rs
```diff
@@ -144,6 +152,9 @@ pub trait Trait:
     /// Defines min cherry for a bounty.
     type MinCherryLimit: Get<BalanceOf<Self>>;
 
+    /// Defines min oracle cherry for a bounty.
+    type MinOracleRewardLimit: Get<BalanceOf<Self>>;
```
bedeho: We don't need such a limit.

runtime-modules/bounty/src/lib.rs
```diff
@@ -302,17 +312,14 @@ pub enum BountyMilestone<BlockNumber> {
...
+
+    /// Council terminated this bounty
+    OraclePerpetualWorkingInterruption,
 
     /// A judgment was submitted for a bounty.
-    JudgmentSubmitted {
-        /// The bounty judgment contains at least a single winner.
-        successful_bounty: bool,
-    },
+    JudgmentSubmitted { successful_bounty: bool },
```
bedeho: Can you add a brief comment here on why we specifically need to keep track of whether the bounty was successful or not.

runtime-modules/bounty/src/lib.rs
```diff
@@ -410,28 +416,24 @@ pub struct EntryRecord<AccountId, MemberId, BlockNumber, Balance> {
...
+        ///The percent share (0 - 1) to slash.
+        slashing_share: Perbill,
 
-impl<Balance> Default for OracleWorkEntryJudgment<Balance> {
-    fn default() -> Self {
-        Self::Rejected
-    }
+        /// After slash takes place the rest of the locked balance will be unlocked,
+        /// the council has the option to give description why slash happened.
+        action_justification: Option<Vec<u8>>,
```
bedeho: We follow the convention of just having `Vec<_>` for these kind of field, and have it be empty if there is no justification.

runtime-modules/bounty/src/lib.rs
```diff
@@ -452,6 +454,36 @@ struct RequiredStakeInfo<T: Trait> {
...
+// The funder can withdraw the bond after deleting the BountyContributions entry belonging to him
+static STATE_BLOAT_BOND: u32 = 10;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct Contribution<T: Trait> {
+    // contribution amount
+    amount: BalanceOf<T>,
+    // amount of bloat bond a funder had to paid
+    state_bloat_bond: BalanceOf<T>,
```
bedeho: `state_bloat_bond_amount`

shamil-gadelshin:
Please, fix the bounty crate comments. For example: remove the description of the `withdraw_work_entry` extrinsic.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+                        // Claim the winner reward.
+
+                        Self::transfer_funds_from_bounty_account(
+                            &worker_account_id,
+                            bounty_id,
+                            reward
+                        )?;
+                        // Delete the work entry record from the storage.
+                        Self::remove_work_entry(&bounty_id, &entry_id);
+                        // println!("WorkEntrantFundsWithdrawn: {:?}-{:?}", entry_id, entry.member_id);
```
shamil-gadelshin: Remove this commented code, please.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,29 +894,34 @@ decl_module! {
...
+                origin,
+                bounty.creation_params.creator.clone(),
+            )?;
 
             let current_bounty_stage = Self::get_bounty_stage(&bounty);
 
             Self::ensure_bounty_stage_for_canceling(current_bounty_stage)?;
 
+            let bounty_account_id = Self::bounty_account_id(bounty_id);
+            let balance = balances::Module::<T>::usable_balance(&bounty_account_id);
```
bedeho: What are the constraints of `usable_balance` ? we don't want to move funds that are locked (e.g. due to invite lock) or reservations.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,29 +894,34 @@ decl_module! {
...
+
+            //Checks if there are suffiicient funds to safely perform cherry and oracle reward withdrawal
+            ensure!(balance >= bounty.creation_params.cherry.saturating_add(bounty.creation_params.oracle_reward),
```
bedeho: Remove this solvency check, this should always be satisfied by virtue of creation step correctness.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,29 +894,34 @@ decl_module! {
...
+                Error::<T>::NoBountyBalanceToOracleRewardAndCherryWithdrawal);
+
             //
             // == MUTATION SAFE ==
             //
 
             Self::return_bounty_cherry_to_creator(bounty_id, &bounty)?;
-
+            Self::return_bounty_oracle_reward_to_creator(bounty_id, &bounty)?;
```
bedeho: You have two fallible calls, with side-effects, beyond the mutation safe barrier. If side-effect in `return_bounty_cherry_to_creator`was to occur, but then `return_bounty_oracle_reward_to_creator` was to stop, then state would be corrupted. Having a single fallible call (e.g. `return_bounty_cherry_to_creator`) with side-effects is OK, but two in a row is a problem.

Now it may be that you believe that the checks done before the barrier means this should not happen, but this should be reflected in the types, this is very unclear.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1724,6 +2004,45 @@ impl<T: Trait> Module<T> {
...
+
+        Ok(())
+    }
+
+    // Transfers oracle cherry to oracle and fires an event.
+    fn withdraw_oracle_reward_to_oracle(
+        bounty_id: T::BountyId,
+        bounty: &Bounty<T>,
+    ) -> DispatchResult {
+        let bounty_oracle_manager = BountyActorManager::<T>::get_bounty_actor_manager(
```
bedeho: This check should not be needed, and it is a source of failure representation in types. This method `withdraw_oracle_reward_to_oracle` is always being called _after_  a `BountyActorManager` instance has been created with this method, so you can just pass this in as an argument to get rid of this failure case.

shamil-gadelshin (changes requested):
Benchmarks looks OK. Most issues are cosmetic:
- please, fix all linter issues for the pallet: `cargo clippy --package pallet-bounty --all-features`
- add an empty line between benchmarks and remove all double empty lines in the codebase
- use rust recommended code line length
- please, consider adding helper functions to create a bounty, add a work entry, etc... There are a lot of code duplication in the benchmarking file. It's optional though.

runtime-modules/bounty/src/lib.rs
```diff
@@ -820,12 +946,22 @@ decl_module! {
                 BountyStage::Funding { has_contributions: false }
             )?;
 
+            let bounty_account_id = Self::bounty_account_id(bounty_id);
+            let balance = balances::Module::<T>::usable_balance(&bounty_account_id);
```
bedeho: Same question here about usable balance

runtime-modules/bounty/src/lib.rs
```diff
@@ -820,12 +946,22 @@ decl_module! {
...
+
+            BountyActorManager::<T>::get_bounty_actor_manager(
```
bedeho: Why are you doing this call?

It has no side-effect and you don't use the returned value?

runtime-modules/bounty/src/lib.rs
```diff
@@ -820,12 +946,22 @@ decl_module! {
...
+                bounty.creation_params.creator.clone(),
+            )?;
+            //Checks if there are sufficient funds to safely perform cherry and oracle reward withdrawal
+            ensure!(balance >= bounty.creation_params.cherry.saturating_add(bounty.creation_params.oracle_reward),
```
bedeho: Again not needed, solvency check.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+                    OracleWorkEntryJudgment::Winner{ reward } => {
+
+                        let entry = Self::entries(entry_id);
+                        // Unstake the full work entry state.
+                        let worker_account_id = T::Membership::controller_account_id(entry.member_id)?;
+
+                        T::StakingHandler::unlock(&entry.staking_account_id);
+                        // Claim the winner reward.
+
+                        Self::transfer_funds_from_bounty_account(
```
bedeho, line 1373: I think we need to rewrite `transfer_funds_from_bounty_account`, it should not return `transfer`  result, because that should never fail. Instead, you can `debug_assert` that result of transfer is always `Ok`, and just have no return value from `transfer_funds_from_bounty_account` itself. That way we can catch any mistake during non-production builds for example, like unit tests.

runtime-modules/bounty/src/lib.rs
bedeho: After looking closely at both of these methods, I believe neither should ever fail, this is just an implicit solvency check. The following applies to both

1. First source of error is `get_bounty_actor_manager`, but I have already recommended removing this in the fist comment of this review.
2. Second source of error is `BountyActorManager::transfer_funds_from_bounty_account`, however it can only fail because `Module::<T>::transfer_funds_from_bounty_account` can fail, which is again an implicit solvency failure. I have already recommended a rewrite to avoid this below.

Also, it seems both should first check that there is an actual > 0 reward to return, because its quite strange to do database operations and emit event with 0. Wasted and confusing for query node developers.

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+
+            ensure!(matches!(current_bounty_stage,
+                BountyStage::WorkSubmission|
+                BountyStage::Judgment),
+                Self::unexpected_bounty_stage_error(current_bounty_stage));
+
+            let account_id = Self::bounty_account_id(bounty_id);
+
+            ensure!(Self::check_balance_for_account(oracle_reward, &account_id),
+                Error::<T>::NoBountyBalanceToOracleRewardWithdrawal);
```
bedeho: Clear mutation safe barrier is missing.

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+            let current_bounty_stage = Self::get_bounty_stage(&bounty);
+
+            ensure!(
+                matches!(current_bounty_stage,
+                    BountyStage::Funding{..} |
+                    BountyStage::WorkSubmission |
+                    BountyStage::Judgment
+                ),
+                Self::unexpected_bounty_stage_error(current_bounty_stage)
+            );
```
bedeho, line 1076: Clear mutation safe barrier is missing.

runtime-modules/bounty/src/lib.rs
```diff
@@ -855,10 +991,26 @@ decl_module! {
             let bounty = Self::ensure_bounty_exists(&bounty_id)?;
 
             ensure!(amount > Zero::zero(), Error::<T>::ZeroFundingAmount);
+            let mut funds_so_far = Self::contribution_by_bounty_by_actor(bounty_id, &funder);
```
bedeho: What is happening here is that if `funder?  has not yet funded this bounty, then this will in fact insert a default constructor `Contribution` into the map and return it. This mutation is happening before all validation is completed, hence it will result in dangling contributions being addable without a state bloat bond.

This is not ok, you have to explicitly check whether a contribution already exists, or not, and deal explicitly with each case carefully.

Also, please avoid mutable, it is not required.

bedeho: As the frist step of the process of validation you need to check the case where this is the first contribution of the contributor, and ensure that it is >= `STATE_BLOAT_BOND`. There is no need for an explicit `amount > Zero::zero()`  above, whether its a new or existing contributor.

runtime-modules/bounty/src/lib.rs
```diff
@@ -855,10 +991,26 @@ decl_module! {
...
 
-            ensure!(amount >= T::MinFundingLimit::get(), Error::<T>::FundingLessThenMinimumAllowed);
+            let maximum_funding_reached = bounty.is_maximum_funding_reached(
+                bounty.total_funding.saturating_add(amount)
+            );
 
-            bounty_funder_manager.validate_balance_sufficiency(amount)?;
+            let actual_funding = if maximum_funding_reached {
```
bedeho: Looking closer at this code, its too complex with multiple mutables flying around. The original sin was already there in the code base from before, but with introduction of bloat bond the conditions became more complex, revealing the problem more clearly. Can you redo this entire section that does correct amount management

- with no mutables
- with correct amount checks (e.g. no need to check for 0 or min contribution)
- with precise variable names
- with comments

**Perhaps we also should just simplify, if someone wants to exceed the limit, it will be accepted. Seems to even perhaps be useful in some cases, and makes the code simpler.**

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+            let current_bounty_stage = Self::get_bounty_stage(&bounty);
+
+            ensure!(matches!(current_bounty_stage,
+                BountyStage::WorkSubmission|
+                BountyStage::Judgment),
+                Self::unexpected_bounty_stage_error(current_bounty_stage));
+
+            let account_id = Self::bounty_account_id(bounty_id);
+
+            ensure!(Self::check_balance_for_account(oracle_reward, &account_id),
```
bedeho: Remove check.

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+                Error::<T>::NoBountyBalanceToOracleRewardWithdrawal);
+
+            Self::return_bounty_oracle_reward_to_creator(bounty_id, &bounty)?;
+
```
bedeho, line 998: What should happen to possible cherry?

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+            // Update bounty record.
+            <Bounties<T>>::mutate(bounty_id, |bounty| {
+                bounty.milestone = BountyMilestone::OraclePerpetualWorkingInterruption
+            });
+
```
bedeho, line 1002: Should we not also check if there was no funding, and if so, just return cherry to creator and kill entire bounty? Seems pointless to only return value to oracle, but not creator.

If you agree, then remember to update diagram!

bedeho: This made me wonder, why do we need both `veto_bounty` and `terminate_bounty`, when both are for council? The only distinction is that the former is to stop bounties that are being funded, but have no funding yet, and the latter is for a funded bounty in working or judgement period.

Why not just make a single `terminate` that works for both?

runtime-modules/bounty/src/lib.rs
```diff
@@ -870,43 +1022,143 @@ decl_module! {
...
+        .max(WeightInfoBounty::<T>::switch_oracle_to_council_by_council_approval_successful())]
+        pub fn switch_oracle(
+            origin,
+            new_oracle: BountyActor<MemberId<T>>,
+            bounty_id: T::BountyId,
+        ) {
+
+            let bounty = Self::ensure_bounty_exists(&bounty_id)?;
+
+            let current_oracle = bounty.creation_params.oracle.clone();
```
bedeho, line 1062: I think this whole extrinsic would be simpler if we segmented it into two distinct ones

- `switch_oracle_as_root`: do switch a Root origin
- `switch_oracle`: do switch as the current oracle

Also, no need to check that the new oracle is genuinely distinct from the old one.

runtime-modules/bounty/src/lib.rs
```diff
@@ -930,16 +1182,19 @@ decl_module! {
             let bounty = Self::ensure_bounty_exists(&bounty_id)?;
 
             let current_bounty_stage = Self::get_bounty_stage(&bounty);
-            Self::ensure_bounty_stage(current_bounty_stage, BountyStage::FailedBountyWithdrawal)?;
-
+            ensure!(
+                current_bounty_stage == BountyStage::FailedBountyWithdrawal,
+                Self::unexpected_bounty_stage_error(current_bounty_stage)
+            );
             ensure!(
```
bedeho: This check to recover a contribution is repeated across two extrinsics, so I suggest you factor it out as `ensure_bounty_contribution_exists`.

This also avoids us to keep the check and reading of `funding` together in one local scope, rather than like now where they could start to drift apart more easily.

runtime-modules/bounty/src/lib.rs
```diff
@@ -930,16 +1182,19 @@ decl_module! {
...
                 <BountyContributions<T>>::contains_key(&bounty_id, &funder),
                 Error::<T>::NoBountyContributionFound,
             );
 
-            let funding_amount = <BountyContributions<T>>::get(&bounty_id, &funder);
-            let cherry_fraction = Self::get_cherry_fraction_for_member(&bounty, funding_amount);
-            let withdrawal_amount = funding_amount + cherry_fraction;
+            let funding = <BountyContributions<T>>::get(&bounty_id, &funder);
+            let cherry_fraction = Self::get_cherry_fraction_for_member(&bounty, funding.amount);
```
bedeho: Since all checks required before mutation are effectively done when we know the contribution exists, I suggest you move these computations below the `mutation safe` barrier for clarty.

runtime-modules/bounty/src/lib.rs
```diff
@@ -952,11 +1207,13 @@ decl_module! {
             Self::deposit_event(RawEvent::BountyFundingWithdrawal(bounty_id, funder));
 
             if Self::withdrawal_completed(&current_bounty_stage, &bounty_id) {
```
bedeho: I think `withdrawal_completed` needs to be rewritten
- `withdrawal_completed` => `has_no_contributions_or_work_entries`
- It is only ever called from the withdrawal stage, hence it does not need to recheck `stage` this itself.
- drop external `work_entries_exist` because its only used by `withdrawal_completed`, and if someone unwittingly called it without proper checks, a new default constructed bounty would be inserted in storage.

runtime-modules/bounty/src/lib.rs
```diff
@@ -952,11 +1207,13 @@ decl_module! {
...
+                if Self::is_funding_expired(&bounty){
```
bedeho: I do not understand this condition, why is it required to return oracle reward?
In general I think this section having to do with cleanup requires more comments.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1052,64 +1310,65 @@ decl_module! {
...
             let current_bounty_stage = Self::get_bounty_stage(&bounty);
 
             Self::ensure_bounty_stage(current_bounty_stage, BountyStage::WorkSubmission)?;
-
-            Self::ensure_work_entry_exists(&entry_id)?;
-
             //
             // == MUTATION SAFE ==
             //
+            if bounty.active_work_entry_count == 0{
```
bedeho: 1. This does solvency check, which we should not.
2. I think its better to try to limit the number of places payouts are happening, as it becomes increasingly complex to keep track of whether it has already happened or not, lets just allow this to be done in the withdrawal stage as usual.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+            let cherry = bounty.creation_params.cherry;
 
             //
             // == MUTATION SAFE ==
             //
 
             // Return a cherry to a creator.
             if successful_bounty {
+
+                let amount = oracle_reward.saturating_add(cherry).saturating_add(bounty.total_funding);
```
bedeho: Remove solvency checks in both parts of this conditions, and you are also doing them after the mutation safe barrier.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+                RawEvent::WorkEntrantStakeUnlocked(
                     bounty_id,
-                    reward
-                )?;
+                    entry_id,
+                    entry.staking_account_id)
+            );
+            Self::remove_work_entry(&bounty_id, &entry_id);
+
+            if Self::withdrawal_completed(&current_bounty_stage, &bounty_id) {
```
bedeho: Why can we ignore destination of cherry and oracle funds in all cases?

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
-            // Delete the work entry record from the storage.
-            Self::remove_work_entry(&bounty_id, &entry_id);
+            let bounty = Self::ensure_bounty_exists(&bounty_id)?;
+
+            let current_bounty_stage = Self::get_bounty_stage(&bounty);
+
+            ensure!(current_bounty_stage == BountyStage::SuccessfulBountyWithdrawal,
+                Self::unexpected_bounty_stage_error(current_bounty_stage)
+            );
+            ensure!(
```
bedeho, line 1494: See other comment about new ensure for this.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+                <BountyContributions<T>>::contains_key(&bounty_id, &funder),
+                Error::<T>::NoBountyContributionFound,
+            );
+
+            let funding = <BountyContributions<T>>::get(&bounty_id, &funder);
+            let account_id = Self::bounty_account_id(bounty_id);
+
+            ensure!(Self::check_balance_for_account(funding.state_bloat_bond, &account_id),
```
bedeho: Same issue here with solvency check.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1118,127 +1377,210 @@ decl_module! {
...
+                Error::<T>::NoBountyBalanceToStateBloatBondWithdrawal);
+
+            //
+            // == MUTATION SAFE ==
+            //
+
```
bedeho, line 1509: I believe that reason we don't need to care about cherry & oracle reward payout is because it has already been done during judgement?

bedeho (changes requested):
1. I think in general there are many places where having more granular comments about the objective of some part of the code would be very helpful.
2. It seems you are many places checking whether there is sufficient funds in the bounty account to make a certain payment. We should def. not be doing these checks. If you are trying to rely on this state to know whether you have already paid out a certain value or not, like returning oracle or cherry, then don't.
3. Can you rewrite `get_bounty_actor_manager` to not return `Err`? It is a consistent source of misleading error conditions that basically check that an already existing member exists. In some cases its even being called in codepaths where we know creator is council, that is the Root extrinsics. I suggest you just do a `debug_assert` on result of `T::Membership::controller_account_id` being `Ok`, and then not having a return value at all.
4. Can you rewrite `remove_bounty`: this was already a problem from before, but I think its best to just address it now. 
   - It still calls `<BountyContributions<T>>::remove_prefix` 
   - Do not slash any balance in the account, it should be zero, instead just do `debug_assert` that it is indeed zero.
 5. The biggest problem I am seeing is that its very difficult to clearly reason about when exactly cherry and oracle reward should or shuold not get paid out. I therefore recommend that for every method where bounty is finally removed, write some comment which explains where various funds have ended up in this case, it will make it easier for reader to get up to speed.

runtime-modules/bounty/src/lib.rs
L3pereira: Are you referring to both limits or just the minimum oracle reward?

runtime-modules/bounty/src/lib.rs
L3pereira: This is the only balance that matters in terms of most operations on tokens. It
alone is used to determine the balance when in the contract execution environment.
**free**

The amount that `free` may not drop below when withdrawing for *anything except transaction
fee payment*.
**misc_frozen**

Get the balance of an account that can be used for transfers, reservations, or any other
non-locking, non-transaction-fee activity. Will be at most `free_balance`.
**usable_balance** = `free - misc_frozen`

`misc_frozen` is updated when locked `misc_frozen.max(locked amount)`

so I guess usable_balance doesn't move locked funds

runtime-modules/bounty/src/lib.rs
bedeho: Only oracle.

runtime-modules/bounty/src/lib.rs
L3pereira: You have to have the type in this case, it throws an error if you write `action_justification: Option<Vec<_>>`.
You can use _ for example in a colection, `a.iter().map(|x| ...).collect::<Vec<_>>();` but that's because the type  was already defined.

runtime-modules/bounty/src/lib.rs
L3pereira: Could you explain this point better, please

runtime-modules/bounty/src/lib.rs
L3pereira: Do you want to validate the oracle reward > 0 at the moment the bounty is created (create_bounty)? This saves me from checking the condition in other extrinsics.

runtime-modules/bounty/src/lib.rs
L3pereira: If you want all payouts (oracle reward and cherry) to be placed in the withdrawal, I have to modify the bountymilestone to include flags indicating which path the bounty followed, in some situations, a failed withdrawal can mean different payouts to different actors (ex creator or oracle).

runtime-modules/bounty/src/lib.rs
L3pereira: Yes, the tests also could have such helper functions, a lot of replication

bedeho:
A few things from meeting

1.  There is a bug in `cancel_bounty`, which allows cancellation when funding period has expired, this is a bug which allows removing a bounty with contribution.
2. Oracle reward is _never_ automatically paid out, oracle must actively withdraw reward in all terminal cases, bounty therefore also cannot be removed before this has happened. Add new global variable `oracle_withrew_reward: bool`, and a new extrinsic for this withdrawal. **Remember that there may be 0 oracle reward**.
3. Make sure that `terminate_bounty` will from all stages (funding, working and judgement stages) also work if there are contributions, going to the failed state.

runtime-modules/bounty/src/lib.rs
bedeho: Closing, we ended up with only `terminate`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -348,6 +381,9 @@ pub struct BountyRecord<Balance, BlockNumber, MemberId: Ord> {
 
     /// Current active work entry counter.
     pub active_work_entry_count: u32,
+
+    ///This flag is set to true, if oracle called withdraw_oracle_reward.
+    pub oracle_withrew_reward: bool,
```
bedeho: I think a better semantic here is `has_unpaid_oracle_reward: bool`, which would be `true` only if there was a >0 reward and `withdraw_oracle_reward` was not yet paid, and `false` otherwise. The problem with the current semantic is that you need two checks and errors now, which is a bit bloated and gives no extra value.

```Rust
            ensure!(!bounty.oracle_withrew_reward, Error::<T>::OracleRewardAlreadyWithdrawn);
            ensure!(!oracle_reward.is_zero(), Error::<T>::OracleRewardIsZero);
```

runtime-modules/bounty/src/lib.rs
bedeho: Ok, so as we discussed, lets turn this into `action_justification: Vec<u8>`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -452,6 +483,39 @@ struct RequiredStakeInfo<T: Trait> {
...
+}
+impl<T: Trait> Contribution<T> {
+    fn add_funds(&self, amount: BalanceOf<T>) -> Contribution<T> {
+        Contribution::<T> {
+            amount: self.amount.saturating_add(amount),
+            state_bloat_bond_amount: T::Balance::from(STATE_BLOAT_BOND_AMOUNT),
+        }
+    }
+    ///Get the sum of state bloat and amount if it's the the first contribution (zero amount)
+    fn get_total_amount(&self) -> BalanceOf<T> {
```
bedeho: Perhaps we can rename just to make sure this is not misunderstood at a later time: `total_bloat_bond_and_funding`

L3pereira:
Do we really need this?
If we get into this function and there are funds remaining, something wrong already happened, which means the bounty logic is not right.
Are we addressing total token issuance when we slash? This will cause a negative imbalance
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L1396-L1400

bedeho:
> Do we really need this?

No.

runtime-modules/bounty/src/lib.rs
L3pereira: Fixed

L3pereira:
Closes https://github.com/Joystream/joystream/issues/3546; Drop max_funding_limit & min_funding_limit from BountyCreationParameters
Closes https://github.com/Joystream/joystream/issues/3545; State bloat bond requirement for creators

Fixed https://github.com/Joystream/joystream/pull/3485#discussion_r834257875
Fixed https://github.com/Joystream/joystream/pull/3485#issuecomment-1086005945

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,138 +866,266 @@ decl_module! {
...
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            Self::return_bounty_cherry_to_creator(bounty_id, &bounty, &bounty_creator_manager);
+            let creator = bounty.creation_params.creator.clone();
+            Self::deposit_event(RawEvent::BountyCanceled(bounty_id, creator));
+
+            <Bounties<T>>::mutate(bounty_id, |bounty| {
```
bedeho: Only do this mutation in an else branch of if below, as there is no point to do it otherwise.

runtime-modules/bounty/src/lib.rs
```diff
@@ -348,6 +372,9 @@ pub struct BountyRecord<Balance, BlockNumber, MemberId: Ord> {
 
     /// Current active work entry counter.
     pub active_work_entry_count: u32,
+
+    ///This flag is set to true, if oracle called withdraw_oracle_reward.
+    pub oracle_withdrew_reward: bool,
```
bedeho: I think a better semantic here is `has_unpaid_oracle_reward: bool`, which would be `true` only if there was a >0 reward and `withdraw_oracle_reward` was not yet paid, and `false` otherwise. The problem with the current semantic is that you need two checks and errors now, which is a bit bloated and gives no extra value.

```Rust
            ensure!(!bounty.oracle_withrew_reward, Error::<T>::OracleRewardAlreadyWithdrawn);
            ensure!(!oracle_reward.is_zero(), Error::<T>::OracleRewardIsZero);
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1430,10 +1793,11 @@ impl<T: Trait> Module<T> {
...
             .is_some()
     }
-
-    // Verifies that bounty has some work entries to withdraw.
-    pub(crate) fn work_entries_exist(bounty_id: &T::BountyId) -> bool {
-        Self::bounties(bounty_id).active_work_entry_count > 0
+    // Verifies that bounty has some contribution to withdraw.
+    // Should be O(1) because of the single inner call of the next() function of the iterator.
+    pub(crate) fn can_remove_bounty(bounty_id: &T::BountyId, bounty: &Bounty<T>) -> bool {
+        (bounty.creation_params.oracle_reward.is_zero() || bounty.oracle_withdrew_reward)
```
bedeho: Again, if semantics was switched to `has_unpaid_oracle_reward`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,138 +866,266 @@ decl_module! {
...
             let current_bounty_stage = Self::get_bounty_stage(&bounty);
 
-            Self::ensure_bounty_stage_for_canceling(current_bounty_stage)?;
+            ensure!(
+                matches!(current_bounty_stage, BountyStage::Funding{..}),
+                Self::unexpected_bounty_stage_error(current_bounty_stage),
+            );
+
+            //contribution is adjusted to prevent exceeding target funding.
+            let (is_target_funding_reached,
```
bedeho: Nice job, this is a complex segment dealing with contribution amount logic, as you are dealing with
- funding cap
- state bloat bond on first attempt
- updating prior attempt

I do think this would be even cleaner if `transfer_amount` is just returned, so the logic around bloat bonds and amounts does not leak into this extrinsic top level scope. You don't end up using
- `state_bloat_bond_amount`
- `is_first_contribution`

for anything else anyway.

runtime-modules/bounty/src/lib.rs
```diff
@@ -452,6 +471,41 @@ struct RequiredStakeInfo<T: Trait> {
     account_id: T::AccountId,
 }
 
+// Current state bloat bond a funder has to pay to contribute (one time payment).
+// The funder can withdraw the bond after deleting the BountyContributions entry belonging to him
+const FUNDER_STATE_BLOAT_BOND_AMOUNT: u32 = 10;
```
bedeho: I know I said the opposite prior, but can we just make these two constants

1. FUNDER_STATE_BLOAT_BOND_AMOUNT
2. CREATOR_STATE_BLOAT_BOND_AMOUNT

pallet trait configurable, just to treat same way as `T::MinWorkEntrantStake`, which is more or less functionally equivalent. Reduces chances these just get lost in the storm later, and also may be useful to be able to change easily at compile time, for example in test environment.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,138 +866,266 @@ decl_module! {
...
+                Self::unexpected_bounty_stage_error(current_bounty_stage));
 
             //
             // == MUTATION SAFE ==
             //
 
-            Self::return_bounty_cherry_to_creator(bounty_id, &bounty)?;
-
-            Self::remove_bounty(&bounty_id);
+            Self::deposit_event(
```
bedeho: Emitting events is by convention the very last step of an extrinsic, so please move this down.

runtime-modules/bounty/src/lib.rs
```diff
@@ -299,18 +318,23 @@ pub enum BountyMilestone<BlockNumber> {
...
-        work_period_started_at: BlockNumber,
-    },
+    /// Bounty cancelled in funding period expired with no contributions.
+    CancelledBounty,
+
+    /// Oracle ended the work submission stage.
+    WorkSubmitted,
+
+    /// Council terminated this bounty
+    OraclePerpetualWorkingInterruption,
```
bedeho: This milestone is not about perpetual working period being stopped, as termination can occur during funding, work and judgement periods. A better name is `CouncilTerminatedBounty`.

runtime-modules/bounty/src/lib.rs
```diff
@@ -299,18 +318,23 @@ pub enum BountyMilestone<BlockNumber> {
...
-        max_funding_reached_at: BlockNumber,
+        target_funding_reached_at: BlockNumber,
     },
 
-    /// Some work was submitted for a bounty.
-    WorkSubmitted {
-        ///  Starting block for the work period.
-        work_period_started_at: BlockNumber,
-    },
+    /// Bounty cancelled in funding period expired with no contributions.
```
bedeho: As a general comment on `BountyMilestone`, it could be good if we could add a short explainer by each variant which explains how it maps to stage, because the bounty stage calculator is a bit hard to reason about for non-Rust devs, like front-end and QN devs, who need to understand the relationship clearly.

runtime-modules/bounty/src/lib.rs
```diff
@@ -269,6 +285,9 @@ pub enum BountyStage {
     /// Bounty funding period expired with no contributions.
     FundingExpired,
 
+    /// Bounty cancelled in funding period expired with no contributions.
+    Cancelled,
```
bedeho: After reviewing the full pallet end to end, the complexity is just so high, I don't think we can justify keeping the `cancel_bounty` extrinsic or stage & milestone, when we also have `terminate_bounty`. They only have slightly different permissions, so I think we should just allow the the creator to call the `terminate_bounty` extrinsic only while in the funding period (even if there are contributions), and then we have covered most of the cancel use case with less complexity.

runtime-modules/bounty/src/lib.rs
```diff
@@ -773,138 +866,266 @@ decl_module! {
...
         ///
         /// ## weight
         /// `O (1)`
         /// - db:
         ///    - `O(1)` doesn't depend on the state or parameters
         /// # </weight>
-        #[weight = WeightInfoBounty::<T>::veto_bounty()]
-        pub fn veto_bounty(origin, bounty_id: T::BountyId) {
+        #[weight = WeightInfoBounty::<T>::terminate_bounty()]
+        pub fn terminate_bounty(origin, bounty_id: T::BountyId) {
```
bedeho, line 985: NOTE: Recall my comment above about dropping `cancel_bounty`, and allowing creator to call `terminate_bounty` during funding stage (even after contributions).

Now to the main point:

# Bug

As discussed:

1. Bounty is in funding period.
2. Has no contributions.
3. Has no oracle reward.
4. Is terminated.

At this point, we will return cherry to creator, and remains in `BountyFailed` stage, and noone can do any final call for bounty to actually get removed.

# Fix

- If there are contributions, then just do as now: just go to `FailedBounty`.
- If instead there are no contributions, then
   - return cherry to oracle
   - if unclaimed oracle reward exists, then go to `FailedBounty`: this will allow withdrawal of reward and final cleanup
   - if there is no unclaimed oracle reward, then `remove_bounty`, which does cleanup

bedeho (changes requested):
1. I am submitting review now, I have not reviewed all extrinsics, but I think its best to do a checkin here.
2. I think documentation in the Rust file is not in synch with how things currently work.
3. Drop minimum cherry limit and all associated validation, as agreed.
4. Drop max/min fuing limit and all associated validation, as agreed.

runtime-modules/bounty/src/lib.rs
L3pereira: Creator can also call terminate_bounty, so CouncilTerminatedBounty is a little bit misleading

runtime-modules/bounty/src/lib.rs
L3pereira: Unfortunately I can't return only `transfer_amount`, this is the `adjusted_amount + funder_bloat_bond` (if it is first time) or just the `adjusted_amount`

`transfer_amount` is used to validate and transfer
`adjusted_amount` is used to update total_funding
`current_contribution` is used to mutate the contribution map combined with `adjusted_amount`

runtime-modules/bounty/src/lib.rs
```diff
@@ -297,20 +331,32 @@ pub enum BountyMilestone<BlockNumber> {
     },
 
     /// A bounty funding was successful and it exceeded max funding amount.
-    BountyMaxFundingReached {
-        ///  A bounty funding was successful on the provided block.
-        max_funding_reached_at: BlockNumber,
```
L3pereira, line 306: I removed `max_funding_reached_at`, It wasn't being used

runtime-modules/bounty/src/lib.rs
```diff
@@ -753,6 +859,7 @@ decl_module! {
                 creation_params: params.clone(),
                 milestone: created_bounty_milestone,
                 active_work_entry_count: 0,
+                has_unpaid_oracle_reward: params.oracle_reward > Zero::zero()
```
L3pereira, line 878: If oracle_reward is zero, then has_unpaid_oracle_reward should be false

runtime-modules/bounty/src/lib.rs
```diff
@@ -762,149 +869,209 @@ decl_module! {
...
+            );
+
+            let is_target_funding_reached = bounty.is_target_funding_reached(
+                bounty.total_funding.saturating_add(amount));
+
+            //contribution is adjusted to prevent exceeding target funding.
+            //If funder already contributed transfer_amount is equal to adjusted_amount
+            let (current_contribution,
+                adjusted_amount,
+                transfer_amount
```
L3pereira, line 925: Couldn't return only transfer_amount,  can be only adjusted_amount if there were already a contribution.
If it is the first time  transfer_amount is adjusted_amount + funder state bloat bond.
I need transfer_amount to validate and transfer the total amount, but I also need adjusted_amount and current_contribution  to update the storage

runtime-modules/bounty/src/lib.rs
```diff
@@ -1330,36 +1624,154 @@ impl<T: Trait> Module<T> {
...
+    ) -> Result<BountyActorManager<T>, BadOrigin> {
+        let bounty_creator_manager = BountyActorManager::<T>::ensure_bounty_actor_manager(
+            origin.clone(),
+            bounty.creation_params.creator.clone(),
+        );
+
+        let actor = match bounty_creator_manager {
+            Ok(creator_manager) => creator_manager,
+            Err(_) => {
+                ensure_root(origin)?;
```
L3pereira, line 1690: If is not creator, check if it's Council

runtime-modules/bounty/src/lib.rs
```diff
@@ -1330,36 +1624,154 @@ impl<T: Trait> Module<T> {
...
+        new_oracle: BountyActor<MemberId<T>>,
+    ) -> Result<BountyActor<MemberId<T>>, DispatchError> {
+        //Checks if the function caller (origin) is current oracle
+        let switcher_actor_manager =
+            BountyActorManager::<T>::ensure_bounty_actor_manager(origin.clone(), current_oracle);
+
+        let switcher = match switcher_actor_manager {
+            Ok(oracle_switcher_manager) => oracle_switcher_manager.get_bounty_actor(),
+            Err(_) => {
+                ensure_root(origin)?;
```
L3pereira, line 1668: If is not oracle, check if it's Council

runtime/src/lib.rs
```diff
@@ -943,9 +943,9 @@ impl bounty::Trait for Runtime {
     type StakingHandler = staking_handler::StakingManager<Self, BountyLockId>;
     type EntryId = u64;
     type ClosedContractSizeLimit = ClosedContractSizeLimit;
-    type MinCherryLimit = MinCherryLimit;
-    type MinFundingLimit = MinFundingLimit;
     type MinWorkEntrantStake = MinWorkEntrantStake;
+    type FunderStateBloatBondAmount = FunderStateBloatBondAmount;
+    type CreatorStateBloatBondAmount = CreatorStateBloatBondAmount;
```
L3pereira, line 1136: These are intended to be variable in the future

L3pereira:
@bedeho 
What do you think about this?
we drop 
`unlock_work_entrant_stake`, 
`withdraw_funder_state_bloat_bond_amount`, 
`withdraw_oracle_reward`

we rename withdraw_funding to  `withdraw_or_unlock_funds` and this function takes care of all the funds (Funder, oracle and worker)

```rust
#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
pub enum WithdrawFundsActorParam<MemberId, EntryId>{
    Funder(BountyActor<MemberId>),
    Worker(MemberId, EntryId),
    Oracle(BountyActor<MemberId>)
}
/// Alias type for the Entry.
pub type WithdrawFundsActor<T> = WithdrawFundsActorParam<
    <T as common::membership::MembershipTypes>::MemberId,
    <T as Trait>::EntryId
>;
```

```rust
        pub fn withdraw_or_unlock_funds(
            origin,
            withdraw_funds_actor: WithdrawFundsActor<T>,
            bounty_id: T::BountyId,
        ) {

            let bounty = Self::ensure_bounty_exists(&bounty_id)?;

            let current_bounty_stage = Self::get_bounty_stage(&bounty);

            let bounty_creator_manager = BountyActorManager::<T>::get_bounty_actor_manager(
            )?;

            match withdraw_funds_actor {
                WithdrawFundsActor::<T>::Funder(funder) => {

                        let bounty_funder_manager = BountyActorManager::<T>::ensure_bounty_actor_manager(
                            origin,
                            funder.clone(),
                        )?;

                        let funding = Self::ensure_bounty_contribution_exists(&bounty_id, &funder)?;

                        //
                        // == MUTATION SAFE ==
                        //

                        Self::withdraw_funds_by_funder_actor_mutation(
                            &bounty_id,
                            &bounty,
                            funder,
                            bounty_funder_manager,
                            funding,
                            current_bounty_stage
                        )?;

                    },
                    WithdrawFundsActor::<T>::Oracle(_) => {

                        let bounty_oracle_manager = BountyActorManager::<T>::ensure_bounty_actor_manager(
                            origin,
                            bounty.creation_params.oracle.clone(),
                        )?;

                        ensure!(bounty.has_unpaid_oracle_reward, Error::<T>::OracleRewardAlreadyWithdrawn);

                        //
                        // == MUTATION SAFE ==
                        //

                        Self::withdraw_oracle_reward_mutation(        
                            &bounty_id,
                            &bounty,
                            bounty_oracle_manager,
                            current_bounty_stage)?;
                    },
                    WithdrawFundsActor::<T>::Worker(member_id, entry_id) => {

                        T::Membership::ensure_member_controller_account_origin(origin, member_id)?;
            
                        let entry = Self::ensure_work_entry_exists(&entry_id)?;
            
                        Self::ensure_work_entry_ownership(&entry, &member_id)?;

                        //
                        // == MUTATION SAFE ==
                        //
                        
                        Self::unlock_worker_stake_mutation(        
                            &bounty_id,
                            entry_id,
                            entry,
                            current_bounty_stage)?;
                    }
            }

            if Self::can_remove_bounty(&bounty_id, &bounty) {
                Self::remove_bounty(
                    &bounty_id,
                    &bounty,
                    &bounty_creator_manager
                );
            }
        }
```

All mutation functions operate under two stages (`SuccessfulBountyWithdrawal`, `FailedBountyWithdrawal`) 
For example `withdraw_oracle_reward_mutation`
The only way this can fail is if the stage is not right, but then nothing is mutated

```rust
   fn withdraw_oracle_reward_mutation(        
        bounty_id: &T::BountyId,
        bounty: &Bounty<T>,
        bounty_oracle_manager: BountyActorManager<T>,
        current_bounty_stage: BountyStage) -> DispatchResult{
        let oracle_reward = bounty.creation_params.oracle_reward;
        match current_bounty_stage{
            BountyStage::SuccessfulBountyWithdrawal|
            BountyStage::FailedBountyWithdrawal => {
                bounty_oracle_manager.transfer_funds_from_bounty_account(*bounty_id, oracle_reward);
                <Bounties<T>>::mutate(bounty_id, |bounty| {
                    bounty.has_unpaid_oracle_reward = false;
                });
    
                Self::deposit_event(RawEvent::BountyOracleRewardWithdrawal(
                    *bounty_id,
                    bounty.creation_params.oracle.clone(),
                    oracle_reward
                ));

                Ok(())
            },
            _ => Err(Self::unexpected_bounty_stage_error(current_bounty_stage))
        } 
    }
```

bedeho:
I would avoid this. Its becoming a very large extrinsic, and this is not really genuinely making anything simpler from what I can see, that balances this new complex extrinsic. All of these activities are rather orthogonal, so I think we should keep them separate.

runtime-modules/bounty/src/lib.rs
```diff
@@ -930,33 +1100,45 @@ decl_module! {
...
             // == MUTATION SAFE ==
             //
 
-            bounty_funder_manager.transfer_funds_from_bounty_account(bounty_id, withdrawal_amount)?;
-
-            <BountyContributions<T>>::remove(&bounty_id, &funder);
-
-            Self::deposit_event(RawEvent::BountyFundingWithdrawal(bounty_id, funder));
+            match current_bounty_stage{
+                BountyStage::FailedBountyWithdrawal => {
```
L3pereira: Both state bloat and funds are withdrawn in `FailedBountyWithdrawal`

runtime-modules/bounty/src/lib.rs
```diff
@@ -930,33 +1100,45 @@ decl_module! {
...
+                    Self::withdraw_funding_mutation(
+                        &bounty_id,
+                        &bounty,
+                        funder,
+                        &bounty_funder_manager,
+                        funding);
+                },
+                BountyStage::SuccessfulBountyWithdrawal => {
```
L3pereira: Only funder state bloat is withdrawn in `SuccessfulBountyWithdrawal`

runtime-modules/bounty/src/lib.rs
```diff
@@ -930,33 +1100,45 @@ decl_module! {
...
+                    Self::withdraw_funding_state_bloat_bond_mutation(
+                        &bounty_id,
+                        funder,
+                        &bounty_funder_manager,
+                        funding);
+                },
+                _ => fail!(Self::unexpected_bounty_stage_error(current_bounty_stage))
```
L3pereira: If the stage is not one of the above, the function fails safely without any mutation,
There are invalid stages that can't reach `unexpected_bounty_stage_error`, for example, FundingExpired and Funding without contributions, since there are no contributions, this condition 
`let funding = Self::ensure_bounty_contribution_exists(&bounty_id, &funder)?;` will fail first.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1151,102 +1325,235 @@ decl_module! {
...
+        <BountyContributions<T>>::remove(&bounty_id, &funder);
+
+        Self::deposit_event(RawEvent::FunderStateBloatBondWithdrawn(
+            *bounty_id,
+            funder.clone(),
+            T::FunderStateBloatBondAmount::get(),
+        ));
+
+        Self::deposit_event(RawEvent::BountyFundingWithdrawal(*bounty_id, funder));
+    }
```
L3pereira: This does the logic that the previous version of witthdraw_funding did

runtime-modules/bounty/src/lib.rs
```diff
@@ -1151,102 +1325,235 @@ decl_module! {
...
+
+        //Remove contribution from
+        <BountyContributions<T>>::remove(&bounty_id, &funder);
+
+        Self::deposit_event(RawEvent::FunderStateBloatBondWithdrawn(
+            *bounty_id,
+            funder,
+            T::FunderStateBloatBondAmount::get(),
+        ));
+    }
```
L3pereira, line 1577: This does the logic that the dropped function withdraw_funder_state_bloat_bond_amount did

runtime-modules/bounty/src/lib.rs
bedeho: This `fail!` is not OK
1. Runtime code must never-ever-ever panic, ever: 
    - https://docs.substrate.io/v3/runtime/events-and-errors/.
    - https://github.com/paritytech/substrate/issues/668
2. In general, regardless of the Substrate context, panics are for a state the code _cannot handle_, its when there is no other choice, its not for failed input validation check, i.e. error handling. Please read: https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html#summary
3. If you have any other panicabl code anywhere, please highlight now. `debug_assert` is ok, because it is not included in production build.
4. You should not have this test after the mutation safe barrier, as you are breaking the semantics of what that means. You always, unless it is _impossible_ (which only happens in one case we have found so far), do 100% checks before you begin with any mutation code in the scope of an extrinsic.

bedeho: The best way to do this is to introduce a private type, e.g. called `ValidWithdrawalStage` defined as

```Rust
// Represents a valid stage for doing withdrawals,
// is used a safe internal representation of validation step in `withdraw_funding` extrinsic.
enum ValidWithdrawalStage {
FailedBountyWithdrawal,
SuccessfulBountyWithdrawal
}

ensure_withdraw_funding_in_valid_stage() -> Result<,ValidWithdrawalStage>
```

and then generate and instance of this from validation before mutation safe barrier, so `let valid_stage = ensure_withdraw_funding_in_valid_stage();`

and now you can just match cleanly on this on steps after mutation safe barrier.

runtime-modules/bounty/src/lib.rs
```diff
@@ -762,149 +870,209 @@ decl_module! {
...
+            ////If origin is creator (not council) then
+            //Stage can be FundingExpired, Funding
+            let current_bounty_stage = Self::ensure_terminate_bounty_stage(&bounty, &terminate_bounty_actor)?;
 
             //
             // == MUTATION SAFE ==
             //
 
-            Self::return_bounty_cherry_to_creator(bounty_id, &bounty)?;
+            match current_bounty_stage {
```
bedeho: I think this is fine, however you can do slightly better in terms of clarity, becuase there are really only three explicit stages the bounty can be in after passing `ensure_terminate_bounty_stage`

  - funding
  - funding expired
  - working period
  - judgement period
 
hence the catch all branch of the matching statement is overly broad, and the reader has to put in extra effort to figure out. You could introduce a new private type returned by `ensure_terminate_bounty_stage` which encodes exactly what is relevant to the later processing, similar to what I suggested in `withdraw_funding` to address your use of `fail`. It is not _critical_, but I think it would be better.

bedeho (changes requested):
- Please make sure when you address a prior review, reply ot each individual unresolved review comment and say what was done, with link to commit if a change was done.
- While I appreciate the gesture of trying to explain with a self-review, its actually just quite noisy to deal with, and will confuse third parties, so better that you just leave a comment block or something.
- Can we rename `ExpiredFundingPeriod` to `NoFundingContributed`, its a bit more precisel and helps distinguish its role from mere `BountyFailed`.
- `OraclePerpetualWorkingInterruption` is not appropriate, as I said prior, it can be triggered even during funding, or judgement. Something like `Terminated` would be better.
- I have attached a state chart diagram showing the stages of the bounty as it works now, I would appreciate if you verified that it corresponds to the implementation. No need to focus on the details of the diagram style, only that it is functionally accurate.

bounty_v2(2)

runtime-modules/bounty/src/lib.rs
L3pereira: Does fail! panic? I think It's similar to ensure!.
ensure! returns `Err($y:expr)` if the condition isn't met
fail! just returns `Err($y:expr)`
https://docs.substrate.io/rustdocs/latest/frame_support/macro.ensure.html
https://docs.substrate.io/rustdocs/latest/frame_support/macro.fail.html

In relation to the mutation barrier, yes I agree, It could be more clear with the enum

bedeho: I assumed it was this

https://stackoverflow.com/questions/26965066/rust-error-macro-undefined-fail

If not, then I was mistaken.

bedeho (approved):
Excellent.

bedeho (changes requested):
One last request!!!
1. `unlock_work_entrant_stake` -> `withdraw_entrant_stake` to follow naming convention.
2. Review updated state chart, key here is that the flow of funds are accurate.

bounty_v2(4)

L3pereira:
- `unlock_work_entrant_stake` renamed to `withdraw_entrant_stake`

- Closes https://github.com/Joystream/joystream/issues/3038;

I think the `withdraw_entrant_stake` is not correct, this extrinsic can be called from `Bounty Failed` and `Bounty Successful`.
This is used because an entry can be left not judged, since this entry is neither winner nor rejected, It shouldn't be rewarded or slashed so the stake can be unlocked

bedeho (approved):
LGTM.

runtime-modules/bounty/src/benchmarking.rs
```diff
@@ -843,119 +1075,665 @@ benchmarks! {
...
+        let bounty_id: T::BountyId = 1u32.into();
+
+        assert!(Bounties::<T>::contains_key(bounty_id));
+        assert_eq!(Bounty::<T>::bounty_count(), 1); // Bounty counter was updated.
+        assert_last_event::<T>(
+            Event::<T>::WorkSubmissionPeriodEnded(
+                bounty_id,
+                oracle).into());
+    }
+    withdraw_entrant_stake{
```
shamil-gadelshin, line 1483: Please, add lines between benchmarks.

shamil-gadelshin (changes requested):
Most comments stay the same:
- please, fix all linter issues for the pallet: `cargo clippy --package pallet-bounty --features runtime-benchmarks`
- add an empty line between benchmarks
- use rust recommended code line length: `cargo fmt` doesn't format complex macros so please format it manually using the similar style. It easy to do that installing the 80-length ruler in the VSCode or similar text editor.

runtime-modules/bounty/src/benchmarking.rs
L3pereira: But `cargo clippy --package pallet-bounty --features runtime-benchmarks` will run all the dependencies, I would have to fix all other pallets like membership etc. 
I used `cargo clippy -p pallet-bounty --all-features -- --no-deps` instead,
Should I fix the other pallets?
Is the 80 length rule, applied only in benchmarking? Because there are situations where `cargo fmt` uses more than 80 col and is not a macro

shamil-gadelshin:
1. My bad: there are no clippy warnings in the bounty pallet left. Only for other pallets as you wrote. 
2. Enforce rust line length rule only for macros in the benchmarking code.

shamil-gadelshin (approved):
Benchmarks LGTM.

bedeho:
Remember to fix #4074 for oracles and whitelist (possibly).

bedeho:
Covered by https://github.com/Joystream/joystream/pull/3840.

PR #3490, opened 2022-03-24
Auction time parameters fix

### LINKED ISSUE #3488: Inconsistency in English auction parameters
# Problem
In the runtime we have:

```rust
struct EnglishAuction {
   end: BlockNumber,
   duration: BlockNumber,
   ...
}
```

In the `start_english_auction` extrinsics however the auction is started immediately (i.e. block height the extr. is called at).
But the semantic is unclear since an invariant should be `start = end - duration`, which can be broken arbitrarely in the `EnglishAuctionParams` by specifying

* `end`
* any value for duration which gives `end - duration` > current block height

# Solution
* remove `duration` from both `EnglishAuction` & `EnglishAuctionParams` 
* update events & types for QN support
* re-generate types & metadata

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202019460587037) by [Unito](https://www.unito.io)

dobertRowneySr:
fixes #3488 
AnD:

 - English auction parameters will use duration since it easier for the end user to specify duration rather than end-block.
- extension logic as in the handbook
- English auction record will use end instead. Since it makes easier to deal with the extension logic
- You can forward start an auction again

I did not update the metadata, since for some reason I am not able to connect to port 9933 for the Joystream node
types are updated


ondratra (approved):
LGTM

bedeho (approved):
LGTM.

runtime-modules/content/src/nft/types.rs
```diff
@@ -233,18 +242,15 @@ impl<
     }
 
     pub(crate) fn with_bid(self, amount: Balance, bidder_id: MemberId, block: BlockNumber) -> Self {
-        let (auction_duration, end) = if self.end.saturating_sub(self.extension_period) < block {
-            (
-                self.auction_duration.saturating_add(self.extension_period),
-                self.end.saturating_add(self.extension_period),
-            )
+        // sniping extension logic
```
bedeho: I would explicitly write out the logic here in the comment. If bid block is in `[end - extension_period, end)` then new `end` is `end + extension_period`

bedeho (commented):
LGTM, just one cosmetic fix.

types/src/content/index.ts
```diff
@@ -75,8 +76,8 @@ export class EnglishAuctionParams extends JoyStructDecorated({
   starting_price: Balance,
   buy_now_price: Option.with(Balance),
   whitelist: BTreeSet.with(MemberId),
-  end: BlockNumber,
-  auction_duration: BlockNumber,
+  duration: BlockNumber,
+  starts_at: Option.with(BlockNumber),
```
ondratra: Order needs to match the definition in runtime.
```suggestion
  starts_at: Option.with(BlockNumber),
  duration: BlockNumber,
```

ondratra (changes requested):

PR #3512, opened 2022-03-28
Fix some more bounty mapping issues

thesan:
Once an entry gets to the `BountyEntryStatusCashedOut` status it becomes difficult to know if the contributor won or not.
 
This partially address Joystream/pioneer#2640 and Joystream/pioneer#2702


Lezek123 (approved):
LGTM

Issue #3515, opened 2022-03-29
NFT query node - add price to nft bought events


kdembler:
There are 2 events that indicate that NFT was bought immediately:

* `NftBoughtEvent` (for NFTs in BuyNow status)
* `BidMadeCompletingAuctionEvent` (for NFTs on auction that had a `buyNowPrice`)

It would be great if QN could add price for which the NFT was bought to those events when queried via respective event queries, so that Atlas can show notification like `Your NFT was bought for 50k tJOY`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202039993973328) by [Unito](https://www.unito.io)

PR #3513, opened 2022-03-28
Rhodes query node auction repair

ondratra:
This PR adapts NFT auction mappings to changes made to runtime by #3353.

Includes changes from https://github.com/Joystream/joystream/pull/3490 & https://github.com/Joystream/joystream/pull/3506 and expects them to be merged first.

~Currently, auction extension period is not handled properly.~

PR #3520, opened 2022-03-29
Rhodes QN NFT - add price to nft bought events

zeeshanakram3:
implements #3515

Based on  #3513 so expects it to be merged first

### DIFF
--- a/query-node/mappings/src/content/nft.ts
+++ b/query-node/mappings/src/content/nft.ts
@@ -284,9 +284,9 @@ async function finishAuction(
     : (auction.topBid as Bid)
 
   // load winner member
-  const winnerMemberId = (auction.lastBid as Bid).bidder.id
+  const winnerMemberId = winningBid.bidder.id
   // load bid amount for which NFT is bought
-  const boughtPrice = (auction.lastBid as Bid).amount
+  const boughtPrice = (auction.topBid as Bid).amount
   const winner = await getRequiredExistingEntity(
     store,
     Membership,
@@ -1015,11 +1015,14 @@ export async function contentNft_NftBought({ event, store }: EventContext & Stor
   // specific event processing
 
   // load video
-  const video = await getRequiredExistingEntity(store, Video, videoId.toString(), 'Non-existing video was bought', [
-    'nft',
-    'nft.transactionalStatus',
-  ])
-  const { nft } = video
+  const { video, nft } = await getNftFromVideo(
+    store,
+    videoId.toString(),
+    'Non-existing video was bought',
+    'Non-existing NFT was bought',
+    ['nft', 'nft.transactionalStatus']
+  )
+
   // read member
   const winner = new Membership({ id: memberId.toString() })
 
@@ -1039,7 +1042,7 @@ export async function contentNft_NftBought({ event, store }: EventContext & Stor
 
     video,
     member: winner,
-    price: (nft?.transactionalStatus as TransactionalStatusBuyNow).price.toNumber(),
+    price: (nft.transactionalStatus as TransactionalStatusBuyNow).price.toNumber(),
   })
 
   await store.save<NftBoughtEvent>(announcingPeriodStartedEvent)
--- a/query-node/mappings/src/content/nft.ts
+++ b/query-node/mappings/src/content/nft.ts
@@ -921,6 +921,8 @@ export async function contentNft_OfferAccepted({ event, store }: EventContext & 
 
   // read member from offer
   const memberId = (nft.transactionalStatus as TransactionalStatusInitiatedOfferToMember).memberId
+  // read price from offer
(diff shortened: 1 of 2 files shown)


runtime-modules/content/src/lib.rs
```diff
@@ -1660,7 +1660,7 @@ decl_module! {
 
                     (
                         updated_nft,
-                        RawEvent::BidMadeCompletingAuction(participant_id, video_id, None),
+                        RawEvent::BidMadeCompletingAuction(participant_id, video_id),
```
kdembler: @zeeshanakram3 I don't think your work on QN should impact those events in any way? Some rebase error or something?

zeeshanakram3: You are right. Since, I based this PR on #3513, I will update this from Rhodes once #3513 has been merged.

kdembler (approved):

query-node/mappings/src/content/nft.ts
```diff
@@ -1012,8 +1014,11 @@ export async function contentNft_NftBought({ event, store }: EventContext & Stor
   // specific event processing
 
   // load video
-  const video = await getRequiredExistingEntity(store, Video, videoId.toString(), 'Non-existing video was bought')
-
+  const video = await getRequiredExistingEntity(store, Video, videoId.toString(), 'Non-existing video was bought', [
+    'nft',
+    'nft.transactionalStatus',
+  ])
+  const { nft } = video
```
ondratra: Use `getNftFromVideo` instead of `getRequiredExistingEntity` wherever you need to get both `video` and `nft` objects.

It will allow you to simplify the line with `price: (nft?.transactionalStatus as TransactionalStatusBuyNow).price.toNumber(),` as well.

query-node/schemas/contentNftEvents.graphql
```diff
@@ -399,6 +402,9 @@ type NftBoughtEvent implements Event @entity {
 
   "Member that bought the NFT."
   member: Membership!
+
+  "Price for which NFT was bought"
+  price: Int!
```
ondratra: I think we should add `price` property to `Auction` entity as well so we have information about the actual price NFT was sold in an auction. It should be simple change since you have this value (`boughtPrice`) in `finishAuction` function already.

ondratra (changes requested):

query-node/mappings/src/content/nft.ts
```diff
@@ -671,6 +671,7 @@ export async function contentNft_AuctionBidMade({ event, store }: EventContext &
     auction.auctionType.plannedEndAtBlock = auction.auctionType.extensionPeriod
     store.save<Auction>(auction)
   }
+  */
```
ondratra: I guess this is a debugging relic, right? Remove it if it's so.

query-node/schemas/contentNftEvents.graphql
kdembler: `Auction.price` may be a bit misleading. Do I understand correctly that this value would be populated only if the action has finished? If so, maybe some more descriptive name could be used, like `finalPrice` or something along the lines

query-node/mappings/src/content/nft.ts
```diff
@@ -284,7 +284,9 @@ async function finishAuction(
     : (auction.topBid as Bid)
 
   // load winner member
-  const winnerMemberId = winningBid.bidder.id
+  const winnerMemberId = (auction.lastBid as Bid).bidder.id
+  // load bid amount for which NFT is bought
+  const boughtPrice = (auction.lastBid as Bid).amount
```
kdembler: Those should use `winningBid`, `lastBid` doesn't exist anymore

kdembler (changes requested):

query-node/mappings/src/content/nft.ts
```diff
@@ -285,6 +285,8 @@ async function finishAuction(
 
   // load winner member
   const winnerMemberId = winningBid.bidder.id
+  // load bid amount for which NFT is bought
+  const boughtPrice = (auction.topBid as Bid).amount
```
kdembler: This needs to use `winningBid` as well, in open auction the lowest bid can be chosen as the winner as well

query-node/mappings/src/content/nft.ts
```diff
@@ -1028,11 +1038,13 @@ export async function contentNft_NftBought({ event, store }: EventContext & Stor
 
   // common event processing - second
 
+  console.log('TransactionalStatusBuyNow', transactionalStatus)
```
ondratra: Remove a debug line.

ondratra (changes requested):
I've noticed that we forgot about direct offers (`OfferAccepted` event). It should remember the buying price as well.

ondratra (approved):
LGTM

Issue #3470, opened 2022-03-22
Add owner to NFT events


bedeho:
# Background
Problem 1 and 2 here: https://github.com/Joystream/atlas/issues/1740#issuecomment-1069362536

# Proposal
Add two nullable owner fields to all NFT related events in order to make filtering feasible.

* `member_owner: Member`
* `curator_group_owner: CuratorGroup`

These refers to whomever is the owner _at the time of the event_. This is a flattened version of ownership, as we can't do relationships through algebraic types. One and only one of the fields should be non-null, please mention this in comments.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202006104763717) by [Unito](https://www.unito.io)

PR #3527, opened 2022-03-30
Rhodes add owner to nft events

zeeshanakram3:
implements #3470 

Based on #3513 so expects it to be merged first


query-node/mappings/src/content/nft.ts
```diff
@@ -671,6 +676,7 @@ export async function contentNft_AuctionBidMade({ event, store }: EventContext &
     auction.auctionType.plannedEndAtBlock = auction.auctionType.extensionPeriod
     store.save<Auction>(auction)
   }
+  */
```
ondratra: I guess this is a debugging relic, right? Remove it if it's so.

query-node/mappings/src/content/nft.ts
```diff
@@ -153,6 +154,7 @@ async function resetNftTransactionalStatusFromVideo(
     return inconsistentState(errorMessage, videoId.toString())
   }
 
+  // TODO: check for curator group
   if (newOwner) {
     nft.ownerMember = newOwner
   }
```
ondratra: Is there a problem with adding support for the curator group here?

query-node/mappings/src/content/nft.ts
```diff
@@ -700,6 +706,7 @@ export async function contentNft_AuctionBidCanceled({ event, store }: EventConte
     'Non-existing auction got bid canceled',
     ['topBid', 'bids', 'bids.bidder']
   )
+  const { nft } = video
```
ondratra: Since `video.nft` is not guaranteed to exist, it creates an extra burden of checking it's not null as can be seen a few lines under.
```
ownerMember: nft?.ownerMember,
ownerCuratorGroup: nft?.ownerCuratorGroup,
```

1) `getCurrentAuctionFromVideo` ensures NFT exists, thus it will be better if it returns `nft` object with the other variables it already does (`video` & `auction`).

2) Let's add a check for `nft` existance in `getCurrentAuctionFromVideo` and get rid of optional chaining there.
```
...
const nft = video.nft

if (!nft) {
  return inconsistentState(errorMessageForNft, videoId) # add errorMessageForNft as function parameter
}

// get auction
const allAuctions = video.nft.auctions || []
...
```

query-node/mappings/src/content/nft.ts
```diff
@@ -824,7 +845,9 @@ export async function contentNft_BidMadeCompletingAuction({
   // specific event processing
 
   // create record for winning bid
-  await createBid(event, store, memberId.toNumber(), videoId.toNumber())
+  const {
+    video: { nft },
+  } = await createBid(event, store, memberId.toNumber(), videoId.toNumber())
```
ondratra: After the change of `getCurrentAuctionFromVideo` as described above, we can make `createBid` return guaranteed `nft` object as well. It will remove the need to use optional chaining used few lines below.
```
ownerMember: nft.ownerMember,
ownerCuratorGroup: nft.ownerCuratorGroup,
```

query-node/mappings/src/content/nft.ts
```diff
@@ -1012,7 +1048,12 @@ export async function contentNft_NftBought({ event, store }: EventContext & Stor
   // specific event processing
 
   // load video
-  const video = await getRequiredExistingEntity(store, Video, videoId.toString(), 'Non-existing video was bought')
+  const video = await getRequiredExistingEntity(store, Video, videoId.toString(), 'Non-existing video was bought', [
+    'nft',
+    'nft.ownerMember',
+    'nft.ownerCuratorGroup',
+  ])
+  const { nft } = video
```
ondratra: `getNftFromVideo` can be used here the same way as described here https://github.com/Joystream/joystream/pull/3520#discussion_r840458252

query-node/schemas/contentNftEvents.graphql
```diff
@@ -91,8 +103,15 @@ type NftIssuedEvent implements Event @entity {
   "NFT's metadata."
   metadata: String!
 
-  "Member NFT was originally issued to."
-  newOwner: Membership
+  # TODO: remove this field ???
+  # "Member NFT was originally issued to."
+  # newOwner: Membership
```
ondratra: I guess this can be removed now.

types/src/content/index.ts
```diff
@@ -39,6 +39,7 @@ export class EnglishAuction extends JoyStructDecorated({
   top_bid: Option.with(EnglishAuctionBid),
   whitelist: BTreeSet.with(MemberId),
   extension_period: BlockNumber,
+  auction_duration: BlockNumber,
```
ondratra: We need to revisit this `EnglishAuction` type as it has been changed recently in the runtime. See https://github.com/Joystream/joystream/blob/rhodes/runtime-modules/content/src/nft/types.rs#L134 and update the type accordingly, please. Note that the order of parameters matters.

ondratra (changes requested):

ondratra (approved):
LGTM!

kdembler (approved):

Issue #3510, opened 2022-03-28
Existential deposit problem for storage module account


bedeho:
Same as https://github.com/Joystream/joystream/issues/3508, but for storage pallet.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202037015799814) by [Unito](https://www.unito.io)

PR #3538, opened 2022-03-31
Audit3. Runtime: Add initial balance to the module account of the storage pallet.

shamil-gadelshin:
We need to deposit some initial balance to the module account of the storage pallet to prevent accounts from deletion because of the existential deposit requirements.

#### Changes
- new field `ModuleAccountInitialBalance` in the `pallet_storage::Trait`
- new genesis code for the storage pallet
- `ModuleAccountInitialBalance` is set to the `ExistentialDeposit` in the runtime configuration
- updated tests and chain-spec

[Original issue](https://github.com/Joystream/joystream/issues/3510)

### DIFF
--- a/runtime-modules/storage/src/lib.rs
+++ b/runtime-modules/storage/src/lib.rs
@@ -1229,6 +1229,14 @@ decl_storage! {
     }
     add_extra_genesis {
         build(|_| {
+            // We deposit some initial balance to the pallet's module account on the genesis block
+            // to protect the account from being deleted ("dusted") on early stages of pallet's work
+            // by the "garbage collector" of the balances pallet.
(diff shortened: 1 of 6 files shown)


runtime-modules/storage/src/lib.rs
```diff
@@ -1224,6 +1227,14 @@ decl_storage! {
         /// "Distribution buckets per bag" number limit.
         pub DistributionBucketsPerBagLimit get (fn distribution_buckets_per_bag_limit): u64;
     }
+    add_extra_genesis {
+        build(|_| {
+            let module_account_id = StorageTreasury::<T>::module_account_id();
```
bedeho, line 1240: Def. add a big bold explainer comment here on what is going on and why.

mnaamani (approved):
LGTM

Issue #3420, opened 2022-03-15
Runtime: fix invalid storage access code in the content pallet.


shamil-gadelshin:
The content pallet contains invalid storage access code:

```
let video = VideoById::<T>::get(video_id);
let channel = ChannelById::<T>::get(video.in_channel);
```

This code could lead to the default object creation by the Substrate Framework. We need to use `ensure_*` methods instead with explicit errors on absent objects and add related tests. We have such code in multiple extrinsics, mostly related to comments/reactions on videos and channels.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201973399734040) by [Unito](https://www.unito.io)

mnaamani:
The code was correct, the only occurrence of:
`let video = VideoById::<T>::get(video_id);` 
Was preceded by
`let post = Self::ensure_post_exists(video_id, post_id)?;`
So it was still safe, and an explicit check for ensure_video_exists was not necessary because of the invariant that if a post is found then the video must exist.
And finally `let channel = ChannelById::<T>::get(video.in_channel);` was therefore also safe.

I did however update code to have some consistency and more clarity, in some cases at the expense of doing a redundant check.

Done in https://github.com/Joystream/joystream/pull/3541

PR #3541, opened 2022-04-01
Audit3: safer storage access in content pallet

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3420

Also made consistent error returned when trying to perform a Post related action on videos.


bedeho (changes requested):
This is actually not good, because we don't buld in a broken runtime detection into our error system, but what you are doing here, and in fact what was already happening before, is that we do throw an error for the case when a video exists, but not its channel, which is impossible. I suggest  `get_channel_from_video` only returns `Channel`, no failure case. Statically, this will be handled by the default constructing behaviour of storage maps when they key does not exist, but at runtime this should never happen.. unless runtime is broken.

mnaamani:
> This is actually not good, because we don't buld in a broken runtime detection into our error system, but what you are doing here, and in fact what was already happening before, is that we do throw an error for the case when a video exists, but not its channel, which is impossible. I suggest `get_channel_from_video` only returns `Channel`, no failure case. Statically, this will be handled by the default constructing behaviour of storage maps when they key does not exist, but at runtime this should never happen.. unless runtime is broken.

Fixed in [d821e85](https://github.com/Joystream/joystream/pull/3541/commits/d821e85e178b3f870e3967134e68ab6dda687f80)

bedeho (approved):
LGTM.

Issue #3259, opened 2022-02-17
Drop runtime level video categories


bedeho:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201847341142770) by [Unito](https://www.unito.io)

PR #3542, opened 2022-04-01
drop video categories from runtime

mnaamani:
Implements https://github.com/Joystream/joystream/issues/3259

chain-metadata and types not updated, simpler to do that all together once all runtime changes are completed.


bedeho (approved):
LGTM.

Issue #3437, opened 2022-03-16
Add payment swap logic to channel transfer feature


bedeho:
# Background
[Original issue](https://github.com/Joystream/joystream/issues/2816)

Initial work was done here https://github.com/Joystream/joystream/pull/3432, but did not include a payment based swap as part of logic, while state was added (`ChannelTransferStatus::PendingTransfer::price`).

# Proposal
When  a non-zero price is set, then accepting the transfer attempt requires that the new `channelOwner` provide such funds to complete the swap. 

* A member _buyer_ funds out of controller account, and member _owner_ receives such funds in the same account. 
* A curator group owner funds, and receives funds, in the budget of the content working group. Hence spending requires that the budget currently allows minting the required quantity of $JOY, receiving just means $JOY is burned and budget of group is increased.

To safely enable the latter features we will later add the following

* Proper curator group permissions: https://github.com/Joystream/joystream/issues/3436
* Worker level spending constraints in the WG module (not a _must_): https://github.com/Joystream/joystream/issues/3435

but we can still implement core feature without these supportive guardrails.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201980015963009) by [Unito](https://www.unito.io)

shamil-gadelshin:
Done

PR #3544, opened 2022-04-01
Audit3. Runtime: Add swap channel price to the content pallet.

shamil-gadelshin:
This PR implements paid channel transfer. It supports members and curator groups.

#### Changes
- integrated content working group's budget manager
- added balances' checks  and transfers for members and curator groups

[Original issue](https://github.com/Joystream/joystream/issues/3437)

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -28,7 +28,9 @@ pub use storage::{
 };
 
 pub use common::{
-    currency::GovernanceCurrency, membership::MembershipInfoProvider, working_group::WorkingGroup,
+    currency::GovernanceCurrency,
+    membership::MembershipInfoProvider,
+    working_group::{WorkingGroup, WorkingGroupBudgetHandler},
     MembershipTypes, StorageOwnership, Url,
 };
 use frame_support::{
@@ -52,6 +54,7 @@ use sp_runtime::{
     ModuleId,
 };
 use sp_std::{collections::btree_set::BTreeSet, vec::Vec};
+
 /// Module configuration trait for Content Directory Module
 pub trait Trait:
     frame_system::Trait
@@ -125,6 +128,9 @@ pub trait Trait:
         + Copy
         + MaybeSerializeDeserialize
         + PartialEq;
+
+    /// Content working group pallet integration.
+    type ContentWorkingGroup: common::working_group::WorkingGroupBudgetHandler<Self>;
 }
 
 decl_storage! {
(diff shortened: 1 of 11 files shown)


runtime-modules/content/src/lib.rs
```diff
@@ -2372,22 +2378,25 @@ decl_module! {
...
             }
 
+            let new_owner = actor_to_channel_owner::<T>(&actor)?;
+
             //
             // == MUTATION SAFE ==
             //
 
             if let ChannelTransferStatus::PendingTransfer(params) = channel.transfer_status {
+                if !params.transfer_params.is_free_of_charge() {
```
bedeho: You should check for `PendingTransfer` before mutation barrier, and return error if not, not silently succeed without side effect like now. You are actually doing this statically above, so better have that check leave a `params: TransferParameters` as a result, as I think you are doing this to do access it again.

runtime-modules/content/src/lib.rs
```diff
@@ -2781,6 +2790,87 @@ impl<T: Trait> Module<T> {
...
+        // Decrease (or slash) balance for the new owner
+        match new_owner {
+            ChannelOwner::Member(member_id) => {
+                let controller_account_id =
+                    T::MemberAuthenticator::controller_account_id(*member_id)?;
+
+                let _ = Balances::<T>::slash(&controller_account_id, price);
+            }
+            ChannelOwner::CuratorGroup(_) => {
+                let new_budget = T::ContentWorkingGroup::get_budget().saturating_sub(price);
```
bedeho, line 2857: Are we sure this budget is sufficient, what if it is empty or too low?

bedeho: Ah seeing now that you check this prior, can you add comments here which describe that sufficiency checks have been done prior, and also add `debug_asserts`? I know we only use debug asserts for invariant that are across extrinsic calls, but just feels so important to make sure these two parts of the codepath do not eventually drift apart in some weird scenario.

runtime-modules/content/src/lib.rs
```diff
@@ -2781,6 +2790,87 @@ impl<T: Trait> Module<T> {
...
+    fn ensure_sufficient_balance_for_channel_transfer(
+        owner: &ChannelOwner<T::MemberId, T::CuratorGroupId>,
+        transfer_cost: BalanceOf<T>,
+    ) -> DispatchResult {
+        let balance = match owner {
+            ChannelOwner::Member(member_id) => {
+                let controller_account_id =
+                    T::MemberAuthenticator::controller_account_id(*member_id)?;
+
+                Balances::<T>::usable_balance(&controller_account_id)
```
bedeho, line 2804: I assume this means that any funds that have been received as part of invitation would not count? as this would be a fund leak. If so, please add a comment to that effect reminding the reader.

bedeho (commented):
Nice work, only minor things.

runtime-modules/content/src/lib.rs
shamil-gadelshin: Correct. Added a comment.

runtime-modules/content/src/lib.rs
shamil-gadelshin: Good idea! Fixed.

runtime-modules/content/src/lib.rs
shamil-gadelshin: I added a comment. However, `debug_assert` burdens the code when it's ready and invariants are checked in tests.

bedeho: I think the utility of this is beyond the verification, its for clarity for reader, to highlight less visible invariants, but I rest my case ;)

bedeho (approved):

Issue #2655, opened 2021-09-14
NFT: destroy_nft


iorveth:
Introduce an ability to destroy nft
Extrinsic caller should be current nft owner ~& initial creator~.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201837557047920) by [Unito](https://www.unito.io)

bedeho:
I also just realized that the spec is a bit misleading, its not allowed that the channel owner can destroy an NFT for a video in their channel. Its only the genuine owner who can do it, so it is either when the NFT is owned by the channel owner, i.e. `NftOwner::ChannelOwner` or when its the explicit member owning the NFT, so `NftOwner::Member`.

mnaamani:
Completed in #3547

PR #3547, opened 2022-04-01
Audit3: destroy nft feature

mnaamani:
implements https://github.com/Joystream/joystream/issues/2655

- [x] Add destroy_nft extrinsic
- [x] Tests

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -1548,13 +1548,8 @@ decl_module! {
             // Ensure nft is already issued
             let nft = video.ensure_nft_is_issued::<T>()?;
 
-            // Ensure channel exists, retrieve channel owner
-            // let channel = Self::ensure_channel_exists(&video.in_channel)?;
-            // Question - Should NFT Destruction be a pausable feature?
-            // channel.ensure_feature_not_paused::<T>(PausableChannelFeature::VideoNftDestruction)?;
-
             // Authorize nft destruction
-            ensure_actor_authorized_to_destroy_nft::<T>(origin, &actor, &nft.owner, video.in_channel)?;
+            ensure_actor_authorized_to_manage_nft::<T>(origin, &actor, &nft.owner, video.in_channel)?;
 
             // Ensure there nft transactional status is set to idle.
             Self::ensure_nft_transactional_status_is_idle(&nft)?;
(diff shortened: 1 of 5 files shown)


runtime-modules/content/src/lib.rs
```diff
@@ -1535,6 +1535,43 @@ decl_module! {
...
+            // Ensure given video exists
+            let video = Self::ensure_video_exists(&video_id)?;
+
+            // Ensure nft is already issued
+            let nft = video.ensure_nft_is_issued::<T>()?;
+
+            // Ensure channel exists, retrieve channel owner
+            // let channel = Self::ensure_channel_exists(&video.in_channel)?;
+            // Question - Should NFT Destruction be a pausable feature?
+            // channel.ensure_feature_not_paused::<T>(PausableChannelFeature::VideoNftDestruction)?;
```
mnaamani: @bedeho Should we add an enum variant `PausableChannelFeature::VideoNftDestruction` to allow moderators to prevent nft destruction on channels?

bedeho: No I don't think so, you should be able to destroy assets that are really yours, this does not harm third parties in any obvious way.

Also critically, and related I think, look at this: https://github.com/Joystream/joystream/issues/2655#issuecomment-1086058293

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -299,6 +299,50 @@ pub fn ensure_actor_authorized_to_manage_nft<T: Trait>(
     Ok(())
 }
 
+/// Ensure actor can destroy nft
+pub fn ensure_actor_authorized_to_destroy_nft<T: Trait>(
```
bedeho: This seems virtually identical to `ensure_actor_authorized_to_manage_nft`, I think perhaps you made this because you thought you needed to block `NftOwner::Member`, but you actually don't :D

bedeho (changes requested):

runtime-modules/content/src/permissions/mod.rs
mnaamani: That right :) will just revert to using `ensure_actor_authorized_to_manage_nft` which has the correct permission check for this action.

bedeho (approved):
Nice job!

Issue #3342, opened 2022-03-03
NFT query node - auction field returning null


kdembler:
This is the data for example NFT (id=5, https://54.172.37.177.nip.io/orion/graphql)

```json
        "transactionalStatusUpdates": [
          {
            "transactionalStatus": {
              "__typename": "TransactionalStatusIdle",
              "dummy": null
            }
          },
          {
            "transactionalStatus": {
              "__typename": "TransactionalStatusBuyNow",
              "price": 1000
            }
          },
          {
            "transactionalStatus": {
              "__typename": "TransactionalStatusIdle",
              "dummy": null
            }
          },
          {
            "transactionalStatus": {
              "__typename": "TransactionalStatusAuction",
              "auction": null
            }
          },
          {
            "transactionalStatus": {
              "__typename": "TransactionalStatusIdle",
              "dummy": null
            }
          }
        ]
```

As you can see, all the `auction` fields are `null`, I would expect to get the auction object. Possibly related to #3312?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1201912378737416) by [Unito](https://www.unito.io)

kdembler:
Originally this issue was about past statuses, but turns out problem exists also for current status:
```graphql
fragment TransactionalStatus on OwnedNft {
  transactionalStatus {
    __typename
    ... on TransactionalStatusIdle {
      dummy
    }
    ... on TransactionalStatusBuyNow {
      price
    }
    ... on TransactionalStatusAuction {
      auction {
        auctionType {
          __typename
          ... on AuctionTypeEnglish {
            duration
            extensionPeriod
          }
          ... on AuctionTypeOpen {
            bidLockingTime
          }
        }
      }
    }
  }
}

query {
  videos(where: { channel: { id_eq: "2" } }) {
    nft {
      ...TransactionalStatus
    }
  }
}
```

ondratra:
This issue is known for quite some time and has it's own ticket in Hydra https://github.com/Joystream/hydra/issues/396.

PR #3548, opened 2022-04-01
query node - temporary fix for auction in transactional status being empty

ondratra:
Temporary workaround for https://github.com/Joystream/joystream/issues/3342 . 

Can be used in query like this:
```
videos {
    id
    nft {
      id
      transactionalStatus {
        __typename
        ... on TransactionalStatusAuction {
          dummy
          tmpAuctionId # this will be properly set
          auction { # this will be null
            id
          }
        }
      }
    }
  }
```


ondratra:
Unfortunately, this doesn't solve the issue.

PR #3552, opened 2022-04-01
decorated Open Auction Bid Accepted event with winer id

dobertRowneySr:
as in title

**ALERT**:
I was not able to properly generate metadata


bedeho (approved):
LGTM.

runtime-modules/content/src/lib.rs
```diff
@@ -2540,8 +2540,8 @@ decl_event!(
         AuctionBidCanceled(MemberId, VideoId),
         AuctionCanceled(ContentActor, VideoId),
         EnglishAuctionCompleted(MemberId, VideoId),
-        BidMadeCompletingAuction(MemberId, VideoId, Option<MemberId>),
```
kdembler, line 2543: I don't think this `Option` should be removed here?

dobertRowneySr: I was used an outdated version of `upstrea/rhodes`
Now it is fixed

PR #3558, opened 2022-04-04
Runtime. Refactor test fixtures for the storage pallet.

shamil-gadelshin:
Our test fixtures contain a lot of similar functions to simplify tests:

```rust
struct FooFixture{
   origin: RawOrigin<u64>,
   account_id: u64,
   member_id: u64,
}

impl FooFixture{
   pub fn default() -> Self{
       origin: RawOrigin::Signed(DEFAULT_ACCOUNT_ID),
       account_id: Default::default(),
       member_id: Default::default(),
   }
   pub fn with_origin(self, origin: RawOrigin<u64>) -> Self{
      Self {
         origin,
         ..self  
    }

    pub fn with_account_id(self, account_id: u64) -> Self{
    ...
    }

    pub fn with_member_id(self, member_id: u64) -> Self{
    ...
    }
    
    pub fn call(self) {
    ...
    }
}
```

Usage will look like:
```rust
let fixture = FooFixture::default()
    .with_account_id(account_id)
    .with_member_id(member_id);
```

These extra lines of codes could be avoided using two [derive macros](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macros):
- the `default` method could be simplified using the [derive_new macro](https://github.com/nrc/derive-new)
- `with_*` methods could be autogenerated using the newly created **derive_fixture** macro. The PR contain a dedicated crate for this proc_macro.

New code will look like this:
```rust
#[derive(Fixture, new)]
struct FooFixture{
   #[new(value = "RawOrigin::Signed(DEFAULT_ACCOUNT_ID)")]
   origin: RawOrigin<u64>,

   #[new(default)]
   account_id: u64,

   #[new(default)]
   member_id: u64,
}

impl FooFixture{
    pub fn call(self) {
    ...
    }
}
```

Usage will look like similar to the original version:
```rust
let fixture = FooFixture::new()
    .with_account_id(account_id)
    .with_member_id(member_id);
```

#### Comments
- Rust demands that **derive_macro** code should be placed in a separate crate from the using code.
- The further ideas to improve testing and benchmarking experience would be introducing separate crates for tests and benchmarks with the common code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,6 +1077,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive-new"
+version = "0.5.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3418329ca0ad70234b9735dc4ceed10af4df60eff9c8e7b06cb5e520d92c3535"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "derive_more"
 version = "0.99.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4205,6 +4216,7 @@ name = "pallet-storage"
 version = "4.0.1"
 dependencies = [
  "derive-fixture",
+ "derive-new",
  "frame-benchmarking",
  "frame-support",
  "frame-system",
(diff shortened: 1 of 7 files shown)


L3pereira (approved):
pallet-storage
Built, 
tested, 
clippy built, 
checked manually for mistakes

I think everything is ok,
This should be applied later to all tests in the runtime-modules, 
it will reduce a substantial amount of code.

Issue #3557, opened 2022-04-04
QN flatten transactional status auction relation


ondratra:
Due to a problem described in https://github.com/Joystream/joystream/issues/3342 will take some time to fix properly, we will create a temporary workaround.
Let's add an optional property `transactionalStatusAuction` to `OwnedNft` entity. It will be a relation to `Auction`, and it will duplicate a value from `TransactionalStatusAuction.auction` (it will be empty on all other transactional status).

In the mappings, it will likely be enough to change the implementation of `setNewNftTransactionalStatus`. Check the type of transactional status there, if it's an auction, set `transactionalStatusAuction` to relevant auction or null otherwise.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202070859330664) by [Unito](https://www.unito.io)

PR #3560, opened 2022-04-05
Rhodes QN NFT - flatten auction field

zeeshanakram3:
implements #3557


query-node/schemas/contentNft.graphql
```diff
@@ -72,6 +72,11 @@ type OwnedNft
   "NFT transactional status"
   transactionalStatus: TransactionalStatus!
```
kdembler: I think if we're going this way, we should flatten all of the variants and remove `transactionalStatus`, not only the auction one. As it is, you can use both `transactionalStatus.auction` and `transactionalStatusAuction` for accessing the same data which may be confusing

zeeshanakram3: I think if we just keep it this way then it will be easier to roll back this change when we make the fix in Hydra and `transactionalStatus.auction` starts working, i.e we only need to remove the `transactionalStatusAuction` field and make a few changes in mappings. 

> we should flatten all of the variants 

Also, I don't see why this is necessary since we don't have any problem with other variants apart from `transactionalStatusAuction`. You are right that having both `transactionalStatus.auction` and `transactionalStatusAuction` for accessing the same data may be confusing but would it create technical problems/challenges in Atlas that demand not to have redundant data?

ondratra (approved):
LGTM. I just want to point out that it will be best to squash-merge this PR so we can revert it easily after we fix the original issue in Hydra https://github.com/Joystream/joystream/issues/3342.

Issue #3436, opened 2022-03-16
Channel agent permissions


bedeho:
# Background
We already have collaborators, which always did seem to require more permissions than we started with, but we also have curators, and it as the value of a channel is getting quite high now (videos, nfts, creator tokens, channel treasurie), it's becoming more important to allow some control on what someone can and cannot do. This became very clear now that you can even swap an entire channel for $JOY.

Importantly, we already now have another form of permissions for curators

https://github.com/Joystream/joystream/issues/2760

however these are for when acting as a _moderator_, the permissions referred to here are _as agents_ for the owner (DAO), or when a collaborator is an agent for owner.

# Proposal 

* Channel permissions have the same form for both purposes: collaborators and curators.
* The exact set of actions we want to block/allow does not need to cover every possible action in a channel, perhaps only the most sensitive, and it does not need to be fully granular.
* Each collaborator has unique permissions on a given channel, set by owner. **If owner is curator group member, they cannot grant action which they themselves cannot do**.
* Each member of a curator group has unique permissions, but the same permissions then apply to all channels the group owns. Lead can set these permissions to anything.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201980009138324) by [Unito](https://www.unito.io)

PR #3561, opened 2022-04-05
Content directory: Channel agent permissions

Lezek123:
Implements https://github.com/Joystream/joystream/issues/3436

Some important, additional changes to note:
- Only **Lead** can now create curator channel and act as curator channel owner, since each curator group member (curator) now has individual, customizable permissions w.r.t. the curator group channel and is generally referred to as channel's **agent**
- There has been a few changes in extrinsic signatures w.r.t. the change mentioned above, for example, `create_channel` will now take `ChannelOwner` as argument (instead of `ContentActor`). Some extrinsics that require channel owner access will also no longer take `ContentActor` as argument (as it can be now easily derived from channel owner), for example: `channel_owner_remark`, `accept_channel_transfer`.
- I took a huge endeavor to refactor the runtime content tests in order to make it easier to test permissions and different actor contexts. The changes in this PR are mostly limited to `/content/src/tests/fixtures.rs` and `/content/src/tests/channels.rs`, even though the actual tests introduced relate to all content pallet features (including videos, nfts, video posts, channel transfers, channel payouts, metaprotocol etc.). This means there may be now some redundancy in tests between `channels.rs` and `nft.rs` / `posts.rs` / `videos.rs` / `merkle.rs` / `metaprotocol.rs` / `transfers.rs` etc. Because I already devoted unproportional amount of time for tests in this PR, I decided to leave it in the current state, possibly to be cleaned up later in another PR.


runtime-modules/content/src/types.rs
```diff
@@ -69,7 +125,7 @@ pub struct ChannelRecord<
     /// The videos under this channel
     pub num_videos: u64,
     /// collaborator set
-    pub collaborators: BTreeSet<MemberId>,
+    pub collaborators: BTreeMap<MemberId, ChannelAgentPermissions>,
```
L3pereira: The type is no longer a Set, maybe "map" in the comment would be better
/// collaborator set -> /// collaborator map

runtime-modules/content/src/types.rs
```diff
@@ -208,7 +273,7 @@ pub struct ChannelCreationParametersRecord<
     /// Metadata about the channel.
     pub meta: Option<Vec<u8>>,
     /// initial collaborator set
-    pub collaborators: BTreeSet<MemberId>,
+    pub collaborators: BTreeMap<MemberId, ChannelAgentPermissions>,
```
L3pereira: The type is no longer a Set, maybe "map" in the comment would be better
/// collaborator set -> /// collaborator map

runtime-modules/content/src/types.rs
```diff
@@ -238,7 +303,7 @@ pub struct ChannelUpdateParametersRecord<StorageAssets, DataObjectId: Ord, Membe
     /// assets to be removed from channel
     pub assets_to_remove: BTreeSet<DataObjectId>,
     /// collaborator set
-    pub collaborators: Option<BTreeSet<MemberId>>,
+    pub collaborators: Option<BTreeMap<MemberId, ChannelAgentPermissions>>,
```
L3pereira: The type is no longer a Set, maybe "map" in the comment would be better
/// collaborator set -> /// collaborator map

runtime-modules/content/src/types.rs
```diff
@@ -54,6 +57,59 @@ pub struct ChannelCategoryUpdateParameters {
...
+    AgentRemark,
+    // Allows updating channel transfer status through `update_channel_transfer_status` tx
+    TransferChannel,
+    // Allows claiming channel reward through `claim_channel_reward` tx
+    // TODO: or `claim_and_withdraw_channel_reward` tx (provided `WithdrawFromChannelBalance` permission is also granted)
+    ClaimChannelReward,
+    // TODO: Allows withdrawing channel balance through `withdraw_from_channel_balance` tx
+    // or `claim_and_withdraw_channel_reward` tx (provided `ClaimChannelReward` permission is also granted)
+    WithdrawFromChannelBalance,
+}
```
L3pereira: Comments should have 3 '/'
Ex. `//` Allows updating channel ... -> `///` Allows updating channel ...

runtime-modules/content/src/permissions/curator_group.rs
```diff
@@ -78,7 +78,7 @@ where
     T::ActorId: Ord,
 {
     /// Curators set, associated with a iven curator group
-    curators: BTreeSet<T::CuratorId>,
+    curators: BTreeMap<T::CuratorId, ChannelAgentPermissions>,
```
L3pereira: Curators set should be map, typo -> iven
/// Curators `map`, associated with a `given` curator group

runtime-modules/content/src/permissions/mod.rs
```diff
@@ -136,106 +136,273 @@ pub fn ensure_actor_auth_success<T: Trait>(
...
+    let opt_agent_permissions = ensure_actor_has_channel_permissions::<T>(
+        &sender,
+        &actor,
+        &channel,
+        &required_permissions,
+    )?;
+
+    if params.collaborators.is_some() && opt_agent_permissions.is_some() {
+        let agent_permissions = opt_agent_permissions.as_ref().unwrap();
+        let new_collaborators_set = params.collaborators.as_ref().unwrap();
```
L3pereira: Although params.collaborators.is_some() && opt_agent_permissions.is_some(), ensure that
unwrap doesn't panic, I think it's not idiomatic, 
maybe if we use a tuple with an `if let (Some(a), Some(b)) =` is better ([stackoverflow](https://stackoverflow.com/questions/64293804/how-to-write-multiple-condition-in-if-let-statement))
```rust
    if let (Some(agent_permissions), Some(new_collaborators_set)) = 
        (opt_agent_permissions.as_ref(), params.collaborators.as_ref()){
        ...
        ensure!(
            updated_collaborators
                .chain(added_collaborators)
                .all(|(member_id, new_permissions)| {
                    let old_permissions = channel
                        .collaborators
                        .get(&member_id)
                        .map_or(BTreeSet::new(), |v| v.clone());
                    old_permissions.is_subset(agent_permissions)
                        && new_permissions.is_subset(agent_permissions)
                }),
            Error::<T>::ChannelAgentInsufficientPermissions
        )
    }

    Ok(opt_agent_permissions)
```

L3pereira (changes requested):
I think most things are correct, just a few points in terms of code cosmetics.
The code builds, the tests pass, clippy and fmt are ok
I checked the logic, but It's a bit cumbersome to analyze and navigate through dependencies, without the rust-analyzer (still complaining about overflow and cyclic deps).
Since I'm not used to this part of the code I think my review about the logic is superficial, I think it's better for @ignazio-bovo  to take a deeper look.

runtime-modules/content/src/types.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

runtime-modules/content/src/types.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

runtime-modules/content/src/types.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

runtime-modules/content/src/types.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

runtime-modules/content/src/permissions/curator_group.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

runtime-modules/content/src/permissions/mod.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3561/commits/158987e062ec8c1a6e047e6e879904c925478f28

L3pereira (approved):
Everything seems ok

dobertRowneySr (approved):
LGTM great work on the test refactoring.
I just have a question:
in `delete_video` there is a call 
`Self::try_to_perform_video_deletion(&sender, channel_id, video_id, &video)?` before the mutation mark (which mutates storage pallet state + content pallet state) while for example in `delete_video_assets_as_moderator` there is a call to `Storage::delete_data_objects` + video mutation. 
what is the rationale behind?

Lezek123:
> in delete_video there is a call
Self::try_to_perform_video_deletion(&sender, channel_id, video_id, &video)? before the mutation mark (which mutates storage pallet state + content pallet state) while for example in delete_video_assets_as_moderator there is a call to Storage::delete_data_objects + video mutation.
what is the rationale behind?

`try_to_perform_video_deletion` is a common logic between `delete_video_as_moderator` and `delete_video`, that's why I factored it out to a separate method, while for `delete_video_assets_as_moderator` there is no equivalent method that reuses this logic.

Issue #3495, opened 2022-03-25
Directly burn channel account funds


bedeho:
When a channel account owned by the platform (i.e. a curator group), owns a channel which generates some kind of revenue in $JOY, then this should be recapturable by the DAO through a token burn. In order to not complicate permissions any further, and since its quite sensitive, and likely quite rare, lets make this a lead only action.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202024733703336) by [Unito](https://www.unito.io)

bedeho:
This could also be instead sent back to council budget.

shamil-gadelshin:
Done: https://github.com/Joystream/joystream/pull/3562

PR #3562, opened 2022-04-05
Audit3.Runtime. Claim channel reward for a council by a content WG leader.

shamil-gadelshin:
This PR introduces an extrinsic `claim_council_reward` to allow the content leader to transfer channel reward to the council budget.

[Original issue](https://github.com/Joystream/joystream/issues/3495)


dobertRowneySr (approved):
LGTM

runtime-modules/content/src/lib.rs
```diff
@@ -2348,6 +2353,39 @@ decl_module! {
...
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn claim_council_reward(
+            origin,
+            channel_id: T::ChannelId,
+        ) {
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+
+            ensure_actor_authorized_to_claim_council_reward::<T>(origin, &channel.owner)?;
+
+            ensure!(
```
bedeho: Don't we have en explicit ensure for this? Seems this will be done again and again, so a more succinct one liner which can be repeated would be nice?

runtime-modules/content/src/lib.rs
```diff
@@ -2348,6 +2353,39 @@ decl_module! {
...
+            // == MUTATION SAFE ==
+            //
+
+            let channel_account_id = ContentTreasury::<T>::account_for_channel(channel_id);
+            let reward: BalanceOf<T> = Balances::<T>::usable_balance(&channel_account_id);
+
+            if !reward.is_zero(){
+                let _ = Balances::<T>::slash(&channel_account_id, reward);
+
+                let budget = T::CouncilBudgetManager::get_budget();
```
bedeho: Seems it would be useful to just have a  `add_to_budget` in the `CouncilBudgetManager`, to avoid this reading and safe addition and setting in client code.

runtime-modules/content/src/lib.rs
```diff
@@ -2348,6 +2353,39 @@ decl_module! {
...
+            );
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            let channel_account_id = ContentTreasury::<T>::account_for_channel(channel_id);
+            let reward: BalanceOf<T> = Balances::<T>::usable_balance(&channel_account_id);
+
+            if !reward.is_zero(){
```
bedeho: Can we do this check, with error if `is_zero`, before we go into mutation safe zone and also trigger an event. I think the simple rule of no side effect => no event is useful for all extrinsics which are supposed to really have a side effect.

bedeho (changes requested):
Nice work, only minor tweaks.

runtime-modules/content/src/lib.rs
shamil-gadelshin: Sure. Fixed: https://github.com/Joystream/joystream/pull/3562/commits/84782b5ed856271539b530963a5689424e322a03

runtime-modules/content/src/lib.rs
shamil-gadelshin: Sure. Fixed: https://github.com/Joystream/joystream/pull/3562/commits/84782b5ed856271539b530963a5689424e322a03

runtime-modules/content/src/lib.rs
shamil-gadelshin: We will have this method as `increase_budget` in the `ephesus` branch.

bedeho (approved):
Excellent!

PR #3566, opened 2022-04-06
fix open auction withdraw bid logic

kdembler:
The problem case where we encountered this:
1. Place a bid on an open auction
2. Place a 2nd bid (replace previous one)
3. Now there are 2 bids (first cancelled, second not cancelled), all good
4. Withdraw a bid
5. The second bid is not cancelled (it should be)


ondratra (approved):
LGTM

PR #3570, opened 2022-04-06
fix updating topBid after changing bid

kdembler:
Further improvement from my last PR. This should fix this case:
1. User A bids for 120
2. User B bids for 150
3. At this point B's bid for 150 is `topBid`
4. B changes their bid to 100
5. `topBid` is not updated, even though it's cancelled now

With this PR, after step 4., `topBid` will be updated to 120 bid from A


ondratra (approved):
LGTM

Issue #3581, opened 2022-04-08
NFT Auction: non winner claim Eng Auction & Types/Runtime mismatch


dobertRowneySr:
* Claim english auction works when claims is made by non winner -> NFT can be stolen via `claim_won_english_auction`
* Inconsistencies between `types/content/index.ts` and `runtime-modules/content/src/nft/types.rs`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202099501508009) by [Unito](https://www.unito.io)

PR #3582, opened 2022-04-08
Rhodes fix english auction claim

dobertRowneySr:
fixes: #3581


bedeho (changes requested):
I don't think this is the right fix: we want anyone to be able to trigger the ownership transfer, its just that we don't want the transfer to be to whomever is actually doing the triggering, but to the actual winner.

runtime-modules/content/src/lib.rs
```diff
@@ -1851,6 +1851,12 @@ decl_module! {
             // Ensure top bid exists
             let top_bid = english_auction.ensure_top_bid_exists::<T>()?;
 
+            // ensure claiming member is top auction bidder
+            ensure!(
+                top_bid.bidder_id == member_id,
+                Error::<T>::MemberClaimingIsNotWinner
+            );
```
kdembler: I don't think that's actually the behaviour we want, this would impact already designed and implemented user experience. The auction should still be settle-able by anyone (or we can restrict it to winner&owner if needed). Instead, I think it's a matter of passing `top_bid.bidder_id` as the auction winner to `Self::complete_auction` instead of `member_id`.

kdembler: Ah completely missed Bedeho's comment, that's the same thing

dobertRowneySr:
last 2 tests in `content/src/tests/nft/claim_english_auctions` gives an idea for the desired behavior

runtime-modules/content/src/errors.rs
```diff
@@ -205,6 +205,9 @@ decl_error! {
         /// Auction whitelist has only one member
         WhitelistHasOnlyOneMember,
 
+        /// Fail to claim English auction by a non winner
+        MemberClaimingIsNotWinner,
```
bedeho: This should be removed.

runtime-modules/content/src/lib.rs
```diff
@@ -1850,6 +1850,8 @@ decl_module! {
 
             // Ensure top bid exists
             let top_bid = english_auction.ensure_top_bid_exists::<T>()?;
+            let top_bidder_id = top_bid.bidder_id;
+            let top_bidder_account_id = T::MemberAuthenticator::controller_account_id(top_bidder_id)?;
```
bedeho: So this is really a call which cannot fail unless the runtime is broken, but I also dont think we should have a panicable `ok` call either.

So why dont you do a `debug_assert(::is_ok)` call on the result, and add a big comment that this should never fail, since membership has been authenticated as real prior, and they are never deleted.

bedeho (changes requested):
As discussed

- rename to `settle_english_auction`, update event name also.
- caller needs not be a member, can be just raw account.

also see inline comments.

runtime-modules/content/src/errors.rs
dobertRowneySr: removed here : 766a142670f3113a33469f205a5a246205951687

runtime-modules/content/src/lib.rs
dobertRowneySr: added here: 7336a43021ccd917c153094a13d48535f1cca3b7

bedeho (approved):
LGTM.

Issue #3584, opened 2022-04-08
NFT Auction completed by exceeding buy now price does not slash buyer


dobertRowneySr:
As title suggests, this is due to non reserving funds before the auction completion function is called

PR #3585, opened 2022-04-08
Rhodes fix buy now complete auction correct slash

dobertRowneySr:
fixes: #3584


shamil-gadelshin (changes requested):
Shouldn't it contain the old bid unreserve code as well (similar to 'non-final' bids)?

shamil-gadelshin (approved):
LGTM

Issue #3568, opened 2022-04-06
NFT query node - add last sale to NFT entity


kdembler:
In Atlas, if the NFT is not on sale, we have a design that shows the last price of this NFT (basically the price of the last completed sale + date of that). To achieve that, let's add info on this to the `OwnedNft` entity. We can do it via an object, like so:

```
lastSale: { price: number; date: Date }
```

but if it's easier to flatten out, that works as well:
```
lastSalePrice: number;
lastSaleDate: Date;
```.



┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202088096404638) by [Unito](https://www.unito.io)

PR #3589, opened 2022-04-11
add lastPrice field to ownedNft

zeeshanakram3:
implements #3568

### DIFF
--- a/query-node/schemas/contentNft.graphql
+++ b/query-node/schemas/contentNft.graphql
@@ -89,6 +89,9 @@ type OwnedNft
 
   "NFT creator's channel"
   creatorChannel: Channel!
+
+  "NFT's last sale (if any)"
+  lastSale: LastSale
 }
 
 type TransactionalStatusUpdate @entity {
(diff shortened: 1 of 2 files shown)


kdembler (approved):
Schema looks good to me, let's get Ondrej to approve the code

query-node/schemas/contentNft.graphql
```diff
@@ -105,6 +108,18 @@ type TransactionalStatusUpdate @entity {
...
+
+"NFT's last sale variant"
+type LastSaleVariant @variant {
+  "Last sale price"
+  price: BigInt!
+
+  "Time at which NFT's last sale happened"
+  date: DateTime!
+}
+
```
ondratra, line 122: It really seems weird that we are using `variant`/`union` for only one variant, but after some thinking, it might I think it's okay to implement this as `variant` instead of `entity` to prevent some of the issues we are currently having with entity relations, filtering, etc.

ondratra (approved):
LGTM

Issue #2651, opened 2021-09-13
NFT minting limits


bedeho:
# Background 

https://github.com/Joystream/joystream/issues/2650

# Proposal 

- There is a global limit of how many NFTs can be issued, across all channels in the content directory, per day. Can be updated by a proposal.
- There is a global limit of how many NFTs can be issued, across all channels in the content directory, per week. Can be updated by a proposal
- There is a default limit of how many NFTs a given channel can mint, per day. Can be updated by lead.
- There is a default limit of how many NFTs a given channel can mint, per week. Can be updated by lead.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1201837410137762) by [Unito](https://www.unito.io)

PR #3609, opened 2022-04-14
Ephesus. Runtime. Add NFT-limits to the content pallet.

shamil-gadelshin:
This PR adds NFT-limits to the content pallet.

[Original Issue](https://github.com/Joystream/joystream/issues/2651)

#### Limit types
1. daily or weekly
1. global or channel

Unlike the original issue this PR allows setting channel NFT limits to leaders only. This will change after the [moderation action permissioning system](https://github.com/Joystream/joystream/issues/3610) is integrated.

## Comments
* There are four combinations of these limits like `global daily` limit. 
* Limits are imposed on the NFT-issuing extrinsics (like 'update_video'). 
* Global limits update requires `root` privileges and could be set using a new proposal.
* Channel limits update requires `curator` privileges and could be set using a dedicated extrinsic.

## Changes
*  content-pallet: `issue_nft`, `create_video` and `update_video` extrinsics are affected
* new default constants (exported) for NFT limits (DefaultGlobalDailyNftLimit, DefaultGlobalWeeklyNftLimit, DefaultChannelDailyNftLimit, DefaultChannelWeeklyNftLimit)
* default channel limits are set on the channel creation
* channel limits are removed on the channel deletion
* global limits are set on the genesis
* new limits could be set using the new extrinsic in the pallet-content: `update_nft_limit`
* new global limits could be using the new proposal: `update_nft_limit`
* tests for runtime, proposals-codex-pallet, and pallet-content crates
* benchmarking for proposals-codex-pallet

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202133891752150) by [Unito](https://www.unito.io)

### DIFF
--- a/runtime-modules/content/src/lib.rs
+++ b/runtime-modules/content/src/lib.rs
@@ -58,9 +58,6 @@ use sp_std::{collections::btree_set::BTreeSet, vec::Vec};
 // - consts
 // - proposals
 // - config
-// - set extrinsics
-// - enum type for limits
-// - shared map and config for nft limit enums??
 // -
 
 /// Module configuration trait for Content Directory Module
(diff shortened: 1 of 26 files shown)


bedeho (changes requested):
As agreed, 

- move channel specific limits & counters into channels
- move global limits out of maps and into top level storage fields

runtime-modules/content/src/lib.rs
```diff
@@ -811,17 +811,13 @@ decl_module! {
             }
 
             if nft_status.is_some() {
-                Self::check_nft_limits(&channel_id)?;
+                Self::check_nft_limits(&channel)?;
```
bedeho: Eeek! classic side-effect violation due to storage integration! what happens if upload suceeds but mintin limit is constrained! 😢 

We will kill this devil once and for all with a better interface, but for now the fix is easy, lift this check up, right after construction of `nft_status` makes most sense.

bedeho: I think this was the only one, but great if you can double check!

bedeho (changes requested):

runtime-modules/content/src/lib.rs
shamil-gadelshin: Yes, you are correct. I was confused by the `mutation safe` mark. Fixed here: https://github.com/Joystream/joystream/pull/3609/commits/4399cc53878de0e79618ba6e579535ae0050a99a

I didn't touch the mark though as agreed in previous PRs.

bedeho (approved):

dobertRowneySr (approved):

Issue #3597, opened 2022-04-12
Issue permissioned CRT with small list of members


bedeho:
Make it possible to launch a permissioned CRT with a small (i.e. constrained size `BTreeSet<MemberId>`) set of initial members, not only a merkle root like now. In this case, manually create account for each. In fact, lets allow specifying initial balance+vesting schedule as well for each, each having different one.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202115992837053) by [Unito](https://www.unito.io)

dobertRowneySr:
When creating a token issuer can specify permissioned mode with a merkle root that will contain all the whitelisted accounts.
Initial whitelist will specify at most `N` accounts for which merkle proof *won't be verified*. 
Now can exist accounts for which merkle verification succeeds and are not specified in the initial whitelist passed during issuance?
If so I immagine that such accounts will have to use `join_whitelist` to have their account created...

bedeho:
> Now can exist accounts for which merkle verification succeeds and are not specified in the initial whitelist passed during issuance?

No, here the way the account is created is not using Merkle proof. It is just directly created as a result of the issuer doing it. This is for scenarios where the issuer wants to just get started more easily with a small number of initial accounts.

So there is no use of `join_whitelist`. I think even this extrinsic name is a bit confusing, perhaps we should just have called it `create_account` or something, because there isn't exactly a specific list.

Lezek123:
I added the [`InitialAllocation` type](https://github.com/Lezek123/substrate-runtime-joystream/blob/creator-token-sales/runtime-modules/project-token/src/types.rs#L406) on my `creator-token-sales` branch, which is:
```rust
pub struct InitialAllocation<AddressId, Balance, VestingScheduleParams> {
    pub(crate) address: AddressId,
    pub(crate) amount: Balance,
    pub(crate) vesting_schedule: Option<VestingScheduleParams>,
}
```
And used inside `TokenIssuanceParameters`
```rust
pub struct TokenIssuanceParameters<Balance, Hash, InitialAllocation> {
    pub(crate) initial_allocation: InitialAllocation,
    // ...
```

It could be easily just changed to `initial_allocation: Vec<InitialAllocation>` to support this idea.

bedeho:
@Lezek123 Looks solid.

bedeho:
Captured in https://github.com/Joystream/joystream/pull/3618.

Issue #2362, opened 2021-04-26
Creator Tokens Requirements


bedeho:
# Background
Already assumes this has been done

https://github.com/Joystream/joystream/issues/3263

in particular the built in channel accounts for holding $JOY earned.

# Requirements
A given generic creator token is herein referred to as $CRT. Here is a super simple state machine diagram which captures the major states and transitions.

![creator!UNITO-UNDERSCORE!tokens](https://user-images.githubusercontent.com/437292/156215128-cbc30e66-4c82-45ac-a7b4-a1018b4b1c93.png)

## Issue $CRT
A channel owner, but not collaborator, can issue a fungible token controlled using normal Substrate accounts, if one does not already exist, with parameters:

* **Ticker:** Short identifying string which is unique across all other creator tokens in the system.
* **Creator Allocation:** An account where a given number of units of the initial supply, denoted `I` below, with optional cliff+linear vesting schedule.
* **Revenue Split %:** what %, `S` of channel revenue the token holders are entitled to.
* **Description:** a plain text description.
* **Initial Offer:** There are three possible initial offer states, and issuance allows for specifying initial parameters for these
  * _sale_: a token sale, possibly followed by the _AMM_ state, see for more information below
  * _AMM_: an automated market maker.
  * _idle_: an idle state, from which a sale can be initiated later.
* **Patronage Rate:** % used for calculating patronage to channel owner.
* **Transferability Policy:** Policy (see below).
* **Metadata:** can be used by apps to add richer semantics.

Let the actual dynamic supply of the token at any time be denoted by `Q`.

### A note on decimals

One unit of the ticker symbol of a token refers to `10^18` base units of the actual token. This is purely a UI concept, and it means all input and output fields of UIs should use this denomination, hence there can be up to  `18` decimal places. So if the user inputs `1,87` this means `18700000000` units as understood by the runtime, or if the runtime has unit balance `3427812313251` then UIs should display `0.00000342781`.

## General
* The token must be compatible with Polkadot ecosystem fungible token standard, and work with major Polkadot ecosystem wallets (Polkadotjs, Talisman, etc.). The most prominent standard I have found is `orml_tokens::MultiTokenCurrencyExtended`, from [ORML](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/currencies/src/lib.rs). It is for example used by AMMs like
  * [HydraDX](https://github.com/galacticcouncil/Basilisk-node/blob/master/pallets/xyk/src/lib.rs#L39),
[...]

PR #3618, opened 2022-04-15
Creator token sales

Lezek123:
Implements:
- Creator token sales functionality based on:
  - Initial spec: https://github.com/Joystream/joystream/issues/2362
  - https://github.com/Joystream/joystream/issues/3638 and related discussion in https://gist.github.com/Lezek123/442a9c6f88e028f6764b1990b45da3cf
- Account balances and vesting based on https://github.com/Joystream/joystream/issues/3633
- Initial allocation based on https://github.com/Joystream/joystream/issues/3597


dobertRowneySr (approved):
LGTM

runtime-modules/project-token/src/types.rs
```diff
@@ -76,42 +137,325 @@ pub enum TransferPolicy<Hash> {
...
+}
+
+impl<JoyBalance, Balance, BlockNumber, AccountId>
+    TokenSale<JoyBalance, Balance, BlockNumber, VestingScheduleParams<BlockNumber>, AccountId>
+where
+    BlockNumber: Saturating + Zero + Copy + Clone + PartialOrd,
+    Balance: Saturating + Clone + Copy + From<u32> + Unsigned + TryInto<u32> + TryInto<u64> + Ord,
+{
+    pub(crate) fn try_from_params<T: Trait>(
+        params: TokenSaleParamsOf<T>,
```
bedeho, line 307: Do we not need the duration > 0 or vesting schedule to be checked in some way?

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+        /// - `amount * sale.unit_price` JOY tokens are transfered from `sender`
+        ///   to `sale.tokens_source` account
+        /// - if new account created: `bloat_bond` transferred from `sender` to treasury
+        /// - buyer's `vesting_schedule` related to the current sale is updated/created
+        /// - some finished vesting schedule is removed from buyer's account_data in case
+        ///   number of buyer's vesting_schedules was == MaxVestingSchedulesPerAccountPerToken
+        ///   and a new schedule was added
+        /// - `token_data.last_sale.quantity_left` is decreased by `amount`
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn purchase_tokens_on_sale(
```
bedeho, line 304: I think we should add a platform fee for sales actually, like we do on NFT sales, where part of the spend has to be burned, and this % comes from a storage variable in the pallet.

Can you make a separate issue for such a feature?

runtime-modules/project-token/src/types.rs
```diff
@@ -1,44 +1,89 @@
...
 
-    /// Initial offering state
-    pub(crate) offering_state: OfferingState,
+    /// Total number of tokens issued
+    pub(crate) tokens_issued: Balance,
+
+    // TODO: Limit number of sales per token?
+    /// Number of sales initialized, also serves as unique identifier
+    /// of the current sale (`last_sale`) if any.
+    pub(crate) sales_initialized: TokenSaleId,
```
bedeho: 1) We have no plan to limit the total number of sales for a token, just to clarify that part.
2) Why does this live outside `TokenSale` ? is there a case where we want to know the id of a past non-existing sale? if so why?
3) Remember that its fine if sale ids are global, not specific to a given token.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+                    sale_id,
+                    &sender,
+                    amount,
+                    cap
+                )?;
+            }
+
+            // Ensure account exists if Permissioned token
+            if let TransferPolicy::Permissioned(_) = token_data.transfer_policy {
+                Self::ensure_account_data_exists(token_id, &sender)?;
```
bedeho: Don't we already have the truth value of this call in variable `account_exists` ?

runtime-modules/project-token/src/lib.rs
```diff
@@ -539,24 +773,43 @@ impl<T: Trait> Module<T> {
...
+
+        // Ensure token offering state is Idle
+        OfferingStateOf::<T>::ensure_idle_of::<T>(token_data)?;
+
+        // Ensure no unrecovered tokens remaining from the previous sale
+        ensure!(
+            token_data.last_sale_remaining_tokens().is_zero(),
+            Error::<T>::RemainingUnrecoveredTokensFromPreviousSale
+        );
+
```
bedeho, line 824: Are we sure we don't need to check anything else? e.g. 

- should we allow `cap_per_member:is_some()` even if token is not permissioned? It does not make sense in that case, and it can help app makers if you have some loud error handling for those sorts of cases.
- is it not a problem with `duration == 0`, these sorts of same block cases have a tendency to generate weird problems that are not easily caught, but that typically is if you are also doing some sort of `on_finalized` type processing at the same time. But best make sure this is safe. There is perhaps also just a case for allowing the platform to regulate minimum sale time, as people can do some really funny business where they do a sale for 1min, and they buy their own tokens at some very high price which causes inattentive third parties, in particular price aggregators or other tools, to show very large market cap.... not sure how big of a problem it is.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+                    AccountData::new_with_vesting_and_bond(
+                        VestingSource::Sale(sale_id),
+                        sale.get_vesting_schedule(amount),
+                        bloat_bond,
+                    )
+                );
+            }
+
+            TokenInfoById::<T>::mutate(token_id, |t| {
+                t.last_sale.as_mut().unwrap().quantity_left = sale.quantity_left.saturating_sub(amount);
```
bedeho: This `unwrap` is not ok, unwrap is in general never ok, and is not needed, w already know statically that the sale exists, so you need to change to take advantage of this.

Generally: never ever panic in the runtime, even if yout hink it can never happen. Panics mean entire side-effect of extrinsic is reversed, including fee payment, which means its a free transaction, which means anyone can kill your chain for free by just flooding such transactions.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+            let amount_to_recover = token_info.last_sale_remaining_tokens();
+            ensure!(
+                !amount_to_recover.is_zero(),
+                Error::<T>::NoTokensToRecover
+            );
+
+            // == MUTATION SAFE ==
+            AccountInfoByTokenAndAccount::<T>::mutate(
+                token_id,
+                &token_info.last_sale.unwrap().tokens_source,
```
bedeho: Same unwrap problem.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+                |ad| {
+                    ad.increase_amount_by(amount_to_recover);
+                },
+            );
+            TokenInfoById::<T>::mutate(token_id, |token_info| {
+                token_info.last_sale.as_mut().unwrap().quantity_left = <T as Trait>::Balance::zero();
```
bedeho: Same unwrap problem.

runtime-modules/project-token/src/types.rs
```diff
@@ -76,42 +137,325 @@ pub enum TransferPolicy<Hash> {
...
+    // Number of blocks before the linear vesting begins
+    pub(crate) blocks_before_cliff: BlockNumber,
+    // Initial, instantly vested amount once linear vesting begins (percentage of total amount)
+    pub(crate) cliff_amount_percentage: Permill,
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, Default)]
+pub struct VestingSchedule<BlockNumber, Balance> {
+    // Block at which the vesting begins
+    pub(crate) start_block: BlockNumber,
```
bedeho: Is this the block the cliff occurs? if so lets rename, right now its a bit unclear to me.

runtime-modules/project-token/src/types.rs
```diff
@@ -76,42 +137,325 @@ pub enum TransferPolicy<Hash> {
...
+    // Linear vesting duration
+    pub(crate) duration: BlockNumber,
```
bedeho: lets rename this to more explicitly be `linear_vesting_duration`.

runtime-modules/project-token/src/types.rs
```diff
@@ -76,42 +137,325 @@ pub enum TransferPolicy<Hash> {
...
+        })
+    }
+
+    pub(crate) fn get_vesting_schedule(
+        &self,
+        amount: Balance,
+    ) -> VestingSchedule<BlockNumber, Balance> {
+        self.vesting_schedule.as_ref().map_or(
+            // Default VestingSchedule when none specified (distribute all tokens right after sale ends)
+            VestingSchedule::<BlockNumber, Balance> {
```
bedeho, line 367: Why is this default impotent schedule needed? it should not be.

runtime-modules/project-token/src/types.rs
```diff
@@ -76,42 +137,325 @@ pub enum TransferPolicy<Hash> {
...
     fn default() -> Self {
         TransferPolicy::<Hash>::Permissionless
     }
 }
 
-/// The possible offering variants: This is a stub
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, Default)]
+pub struct VestingScheduleParams<BlockNumber> {
+    // Vesting duration
+    pub(crate) duration: BlockNumber,
```
bedeho: This notion of duration is unclear to me, is it counting the linear part? if so, lets rename.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+            // == MUTATION SAFE ==
+
+            <Joy::<T> as Currency<T::AccountId>>::transfer(
+                &sender,
+                &sale.tokens_source,
+                joy_amount,
+                ExistenceRequirement::KeepAlive
+            )?;
+
+            if account_exists {
```
bedeho, line 370: So this code fragment was _super complicated_. It took me a very long time (>30m) to become comfortable with it actually doing what it should, but I am still not entirely sure. **I strongly encourage you to review one more time.** e.g. are we sure everyone gets a vesting schedule that starts at the same time (end of sale), not aligned with time of purchase?

The main points are how the crediting of the purchaser is happening (which is kind of hidden) and how vesting schedules are getting stacked with each purchase. I don't know if there is a better way to make this codepath simpler, but in either case, I think you should add a long meaty explainer section in the comments to just explain what is about to happen, as it will help the reader develop a mental model faster, which reduces risks of making mistakes on future changes.

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,40 +234,209 @@ decl_module! {
...
+        ///
+        /// Postconditions:
+        /// - `token_data.last_sale.tokens_source` account balance is increased by
+        ///   `token_data.last_sale.quantity_left`
+        /// - `token_data.last_sale.quantity_left` is set to 0
+        #[weight = 10_000_000] // TODO: adjust weight
+        fn recover_unsold_tokens(origin, token_id: T::TokenId) -> DispatchResult {
+            ensure_signed(origin)?;
+            let token_info = Self::ensure_token_exists(token_id)?;
+            OfferingStateOf::<T>::ensure_idle_of::<T>(&token_info)?;
```
bedeho, line 422: It seems a bit akward to just keep the sale around forever, despite not being needed after a last recover, or being sold out. Wouldn't it make a simpler mental model to just require the sale to be cleaned up to recover any unsold $CRT, or ot have it done automatically when last$CRT is sold? 

Interested in your opinion here.

bedeho (changes requested):
Excellent work, really well done ;)

runtime-modules/project-token/src/types.rs
Lezek123: Sale with `duration == 0` is just a wasted tx, vesting with `duration == 0` is just a delayed payout (in case `blocks_before_cliff` are not also `== 0`, in which case, it's just an instant payout).

None of these cause any invalid state in the runtime etc., but I can add the checks just to be 100% safe.

runtime-modules/project-token/src/types.rs
Lezek123: > Why does this live outside `TokenSale`? is there a case where we want to know the id of a past non-existing sale? if so why?

It's basically `next_sale_id`, it could also live in `TokenSale` as `id` (in which case `next_sale_id = token_data.last_sale.map_or(0, |s| s.id + 1)`) or be a separate global storage value, as you mentioned.

To compare the alternative approaches with the current one:
- If it lives in `TokenSale`, you need to perform this `let next_sale_id = token_data.last_sale.map_or(0, |s| s.id + 1)` which is slightly more complex than `let next_sale_id = token_data.sales_initialized + 1`, the storage write/read cost is pretty much the same though
- if `next_sale_id` lives as a separate storage value it doesn't really solve any issue and just adds one more storage read + write operation during sale initialization. We still need to know the id of the current sale per token for the purpose of identifying whether a member already participated in this sale (`VestingSource`), so we still need to store it either as part of `TokenData` or `TokenSale`, so additional global `next_sale_id` just forces us to check and update an additional, separate storage value, that is not really needed.

runtime-modules/project-token/src/lib.rs
Lezek123: Right, well spotted!

runtime-modules/project-token/src/lib.rs
Lezek123: > should we allow cap_per_member:is_some() even if token is not permissioned? It does not make sense in that case, and it can help app makers if you have some loud error handling for those sorts of cases.

I thought `cap_per_member` should be independent from the token permission model. My reasoning as to why it makes some sense was that membership has a cost and potentially there could be some additional screening implemented for members, so it's not that trivial to just spawn an infinite amount of identities (as is the case with accounts), therefore there is **_some_** utility in having this cap, but probably not much unless there is some really solid protection agains sybil memberships.

runtime-modules/project-token/src/types.rs
Lezek123: It was supposed to serve as a way to delay the distribution of tokens until the sale is finished, accordingly to this comment:
> - all sale terms will be uniform across all partcipants: quantity cap, price, cliff, unlocking rate
> - **cashout only happens in the last block of the sale.**
> - **vesting and cliff only applies from that point on**

_Source: https://github.com/Joystream/joystream/issues/2362#issuecomment-1075221896_

Now, I didn't know that the unvested tokens would be eligible for staking at the time I implemented this, so now I can see a flaw in this approach which is that the tokens bought on sale can be immediately staked, which gives an advantage to those who bought earlier into the sale, however I'm not sure what the current expectations are w.r.t. that and whether something changed since the comment I quoted was published.

runtime-modules/project-token/src/types.rs
Lezek123: Yes, will do

runtime-modules/project-token/src/lib.rs
Lezek123: > are we sure everyone gets a vesting schedule that starts at the same time (end of sale), not aligned with time of purchase?

Yes. Every vesting schedule related to the sale is produced by `TokenSale::get_vesting_schedule`, which either:
- sets `VestingSchedule.start_block` to `self.start_block.saturating_add(self.duration)` (sale end block) or;
- sets `init_block` for `VestingSchedule::from_params` to `self.start_block.saturating_add(self.duration)` (sale end block), in which case `VestingSchedule.start_block = init_block.saturating_add(params.blocks_before_cliff)`, so `sale_end_block.saturating_add(params.blocks_before_cliff)`.

I think it will become much easier to parse if I just introduce `TokenSale::end_block()` method and rename `VestingSchedule::start_block` to something like `VestingSchedule::linear_vesting_start_block` or `VestingSchedule::cliff_block`, but I will add some more comments also.

runtime-modules/project-token/src/lib.rs
Lezek123: Sure, it can be cleaned up on `recover_unsold_tokens` / in case all tokens are bought, but just note that this would render my comment under https://github.com/Joystream/joystream/pull/3618#discussion_r871325710 obsolete and in my opinion would make the argument for keeping `next_sale_id` / `sales_initialized` field as part of `TokenData` even stronger.

runtime-modules/project-token/src/types.rs
Lezek123: Introduced `MinSaleDuration` const. In case of vesting, I there is a case for having `blocks_before_cliff == 0` as well as there is a case for having `duration == 0`, so I left it as is.

runtime-modules/project-token/src/lib.rs
Lezek123: Done https://github.com/Joystream/joystream/issues/3750

runtime-modules/project-token/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/lib.rs
Lezek123: `MinSaleDuration` introduced in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef (**edit:** changed to storage value in https://github.com/Joystream/joystream/pull/3618/commits/ccaf56158ec93ffcee24ecffe81109968858141b + introduced a separate non-zero check), please confirm whether I should disallow setting `cap_per_member` for Permissionless tokens.

runtime-modules/project-token/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/types.rs
Lezek123: Renamed to `linear_vesting_start_block` in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/types.rs
Lezek123: Renamed in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/types.rs
Lezek123: Renamed to `linear_vesting_duration` in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/lib.rs
Lezek123: I added more comments and did some renaming which should make this logic clearer.
Done in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/lib.rs
Lezek123: Introduced the cleanup on `recover_unsold_tokens` and in case `sale.quantity_left == 0`.
Done in https://github.com/Joystream/joystream/pull/3618/commits/d8e70c04967c6c8050a12b204be1f39974e955ef

runtime-modules/project-token/src/types.rs
Lezek123: `MinSaleDuration` changed to storage value in https://github.com/Joystream/joystream/pull/3618/commits/ccaf56158ec93ffcee24ecffe81109968858141b

runtime-modules/project-token/src/types.rs
bedeho: ah ok, lets rename to `next_sale_id`.

runtime-modules/project-token/src/lib.rs
bedeho: > therefore there is some utility in having this cap, but probably not much unless there is some really solid protection agains sybil memberships

Hmm ok, makes sense.

> MinSaleDuration introduced in

Great.

runtime-modules/project-token/src/types.rs
bedeho: > It was supposed to serve as a way to delay the distribution of tokens until the sale is finished

Ok, not totally clear how it does this, but may be a moot point now, see next remark.

> cashout only happens in the last block of the sale.

Oh, I had forgotten that, is there any benefit to this actually? This make UX much worse, in that people must first buy then cashout. What does this buy us? if nothing, then certainly lets drop.

> I didn't know that the unvested tokens would be eligible for staking at the time I implemented this, so now I can see a flaw in this approach an advantage to those who bought earlier into the sale

Anyone arbing a $CRT, whether via OTC, AMM, an external DEX/CEX, has to be aware of what the underlying $JOY pool attached to it, and dynamics around an upcoming/ongoing revenue split. Seems this is part of what people need to be aware of when transacting, and that for thisr eason that revenue splits must be slow to both start and end.

Do you think this is a big issue?

Lezek123: > Oh, I had forgotten that, is there any benefit to this actually?

I can't think of anything specific, besides the point that it kind of "evens the playing field" for all sale participants, but even that is not entirely true, since early participants still have the benefit of using the bought tokens immediately for staking, as noted before.

The benefit on the codebase side is that since all sale purchases are subject to **some** vesting, it allows us to sidestep the need for introducing another storage map like `sale_purchase_by_token_by_member` or some other, separate way to track the amount of tokens already purchased on the current sale by a given member (since we can just use `acccount.vesting_schedules` for this, thanks to indexing by `VestingSource`).

So thinking about this, in order to allow instant cashouts we'd need one of the following:
- `sale_purchase_by_member_by_token` which will be cleaned with `remove_prefix` upon sale completion. I'm guessing `remove_prefix` does not have a deterministic cost (?), so perhaps there's also a need to either introduce max. number participants per sale and use worst-case-scenario in benchmarking or force the caller of sale finalization extrinsic to provide the map size as extrinsic argument and calculate the cost based on that
- `last_sale_total_purchased_amount` as part of `account_data`, which will be a `(sale_id, amount)` pair and it just won't be cleaned up, only updated.

Perhpas I'm wrong about the cost of `remove_prefix`, but I couldn't find any relevant resources.

runtime-modules/project-token/src/types.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/3618/commits/805aa09ac8dba3c3427738a06866668006415aea

runtime-modules/project-token/src/types.rs
bedeho: # Questions

1. Is it the case that we need one of these two fixes in order to enforce the per member cap that applies? 
2. Is there any other reason?
3. When reading the code I forget now: why do we actually need to track the origin of a vesting schedule using `VestingSource` ? Determining which, if any, are no longer in effect, does not depend on this, but on how much they are locking at the current block. Is it so that we can keep updating the same vesting schedule from a single sale multiple times if the buyer purchases more, without having to add a new schedule?

# Remarks

- `remove_prefix` is deterministic, its just a function of `N`, which is controlled by the users, and thus may end up becoming impossible to ever perform within the weight constraint of a block.
- `last_sale_total_purchased_amount` is by far the best solution.

# Conclusion

- I'm finally seeing now why the first schedule was being made now. :D
- While the sale UX is worse, I also see that it is a bit more orderly to give everyone access to their purchased tokens at the same time. However, lets not forget that one can simulate delayed cashouts by manually specifying such a schedule, even if chain allows instant cashouts! so this solution is in a way the most general
- The instant cashouts require new storage variable `last_sale_total_purchased_amount`, while delayed cashouts is worse UX and also is a bit weird in introducing reusing the concept of vesting for two very distinct sort of activties: 
- **In my mind, its a really close call, I think you should decide what is best here, but I lean towards instant cashouts**.

PR #3619, opened 2022-04-15
merge bounty perpetual working period #3485 to audit3_bounty_delivered

L3pereira:
Major differences

`submit_oracle_judgment` receives a new parameter `rationale: Vec<u8>`, the weights, tests had to be updated.
contributor_remark, oracle_remark, entrant_remark, creator_remark accept a `msg: Vec<u8>` but they didn't have a weight function that took that into account.

Now, entries storage, has a double map, all calls to this storage were updated, to include a bounty_id.
```rust
        /// Work entry storage map.
        pub Entries get(fn entries): double_map
            hasher(blake2_128_concat) T::BountyId,
            hasher(blake2_128_concat) T::EntryId => Entry<T>;
```

### DIFF
--- a/runtime-modules/bounty/src/actors.rs
+++ b/runtime-modules/bounty/src/actors.rs
@@ -58,7 +58,7 @@ impl<T: Trait> BountyActorManager<T> {
         }
     }
 
-    // Validate balance is sufficient for the bounty
+    /// Validate balance is sufficient for the bounty
     pub(crate) fn validate_balance_sufficiency(
         &self,
         required_balance: BalanceOf<T>,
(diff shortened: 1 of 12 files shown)


bedeho:
Just to clarify, the differences you are summarising here is just stuff from the work we already did? you did not do any additional changes just for this PR ?

L3pereira:
> Just to clarify, the differences you are summarising here is just stuff from the work we already did? you did not do any additional changes just for this PR ?

I did not add anything new, these are the differences between the work we did and the `audi_bounty_deliver`. I included these, because `audi_bounty_deliver` is more recent than `bounty_updates`, 
The only fix I did was the weight functions of the remark functions, they accept a Vec<u8> and that wasn't being calculated.
Even if the extrinsic  fires only an event and no storage is made, the node has to spend computational power on that parameter

PR #3620, opened 2022-04-15
fix: withdraw bids from past auctions

zeeshanakram3:


query-node/mappings/src/content/nft.ts
```diff
@@ -797,31 +798,37 @@ export async function contentNft_AuctionBidCanceled({ event, store }: EventConte
...
+      'nft',
+      'nft.video',
+      'nft.ownerMember',
+      'nft.ownerCuratorGroup',
+      'auction',
+      'auction.topBid',
+      'auction.bids',
+      'auction.bids.bidder',
+    ],
+  })
```
kdembler, line 813: Will this load the right bid though? Seems it only checks for bidder and video id, I think there can be multiple bids that can match this condition, probably should also filter by not cancelled bids?

kdembler (approved):

Issue #3617, opened 2022-04-15
CRT Account Lifecycle


dobertRowneySr:
# Creation
Refers to the initialization and adding the `AccountData` record to pallet storage
## Permissioned Mode
Either one or another (see [this](https://github.com/Joystream/joystream/issues/3597#issuecomment-1099975963)).  
! permissioned accounts = all the accounts that are allowed to exists in `Permissioned` mode  
- **`(TO BE CONFIRMED)`**: CRT issuer (owner) will issue token in permissioned mode with a merkle `commit` holding the permissioned accounts (via `issue_token` **pallet method**).  users can then actively create their accounts with `join_whitelist` **extrinsic**, providing the merkle proof verification to assert that they are effectively part of the permissioned account set.  if they are successful at doing so they are expected to pay a **bloat bond** fees.
- CRT issuer (owner) can optionally pass a whitelist containing *all the accounts* during token issuance (via `issue_token` **pallet method**). **`(TO BE CONFIRMED)`**: No **bloat bond** is required in this case?
## Permissionless Mode
 - account `X` is created for CRT when existing account `Y` transfers $CRT to `X` (via `transfer` **extrinsic**). **In this case `Y` is expected to pay the bloat bond for** `X`. (And for all the `X`s in case of multi destination transfer)
 - buying into a sale into a new account
 - buying from the AMM into a new account (probably via a `bond` **extrinsic**)
In either cases an `AccountData` record is added to storage.
# Account Deletion
Refers to the `AccountData` record removal from storage. ([reference](https://github.com/Joystream/joystream/pull/3456#issuecomment-1099405766))
account deletion can happen only by a **user** initiated call to the `dust_account(origin, token_id, account_id)` **extrinsic**:
- if `origin` signer is `account_id` then deletion is successful regardless of the permission mode active or the residual balance in the account. In this case token supply has to be reduced by the residual balance deleted (effectively burning it)
- if `origin` signer is not `account_id` then first `AccountData` for account id has to be **empty** and the mode has to be `Permissionless`
# Bloat Bond Management
`AccountData` will contains also a `bloat_bond: ReserveBalance` field, containing the amount paid (in $JOYs) for the bloat bond.
The bloat bond itself can either be:
a. transferred to an internal module account, with the sole purpose of storing bloat bonds
b. burned from the user creating the account and later minted when user deletes his account

[...]

Issue #3382, opened 2022-03-10
Creator Token Workload


dobertRowneySr:
# Notes
- $CRT = generic creator token
- `creator` = anyone that is the owner of a channel, has the right to issue a own $CRT 
# Pallet components
## Supporting library
(token core implementation): I am in favor of using [assets](https://github.com/paritytech/substrate/tree/master/frame/assets):
**Dependencies**:
```
codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
sp-std = { version = "4.0.0", default-features = false, path = "../../primitives/std" }
# Needed for various traits. In our case, `OnFinalize`.
sp-runtime = { version = "6.0.0", default-features = false, path = "../../primitives/runtime" }
# Needed for type-safe access to storage DB.
frame-support = { version = "4.0.0-dev", default-features = false, path = "../support" }
# `system` module provides us with all sorts of useful stuff and macros depend on it being around.
frame-system = { version = "4.0.0-dev", default-features = false, path = "../system" }
frame-benchmarking = { version = "4.0.0-dev", default-features = false, path = "../benchmarking", optional = true }
```
This will provide us with a bunch of primitives such as : 
- `burn(token_id, account, amount)` [here](https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#428)
- `mint(token_id, account, amount)` [here](https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#383)
- `transfer(token_id, from, to, amount)` [here](https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#485)
##  CRT token core
- initial creator allocation, whose implementation details needs to be developed
- vesting schedule, same
- existential amount: smallest amount that an account can hold under which the account is "destroyed" and the balance is burned. Also we want add a custom `crt_token::burn` function that ensure this constraint is respected whenever balance of freshly burned accounts drop below the `existential_amount` threshold.
- `crt_token::{stake, unstake}` functions, that *I suppose* can be build on top of `freeze, thaw` [freeze][(https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#190](https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#626)), [thaw](https://docs.rs/pallet-assets/latest/src/pallet_assets/lib.rs.html#655)
[...]

PR #3632, opened 2022-04-19
Creator token revenue split

dobertRowneySr:
Tracking issue: https://github.com/Joystream/joystream/issues/3711

### References
- https://github.com/Joystream/joystream/issues/2362#issue-867837174
- https://github.com/Joystream/joystream/issues/3382#issuecomment-1079648599

### Extrinsics introduced
- `claim_revenue_split`
- `participate_to_split`
- `abandon_revenue_split`
### Pallet methods introduced
- `issue_revenue_split`
- `finalize_revenue_split`
### Notes
Module account and sub-accounts share the same balance, this means that bloat bonds and revenue split allocation are mixed up. That is why I have added a `dividend_payed` variable to keep track of what has been payed so far and compute leftover amount when split is being finalized
### Nice to have
- Refactor all tests using only the `Fixture` method.For testing new functionalities introduced with this PR I have only used an `execute_call` method. I am satisfied with that approach since it allows to clearly separate the Arrange/Act/Assert parts in the tests
- I have also introduced a `Timeline` data structure that could replace all the `(start, duration)` pairs used to identify a block period.
### CRT developement timeline next up:
- https://github.com/Joystream/joystream/issues/3617#issuecomment-1104766789
- https://github.com/Joystream/joystream/issues/3637
- https://github.com/Joystream/joystream/issues/3687


runtime-modules/project-token/src/events.rs
```diff
@@ -32,6 +45,36 @@ decl_event! {
...
 
+        /// Revenue Split issued
+        /// Params:
+        /// - token identifier
+        /// - start of the split
+        /// - duration of the split
+        /// - JOY allocated for the split
+        RevenueSplitIssued(TokenId, BlockNumber, BlockNumber, JoyBalance),
+
+        /// Revenue Split issued
```
Lezek123:
```suggestion
        /// Revenue Split finalized
```

runtime-modules/project-token/src/events.rs
```diff
@@ -32,6 +45,36 @@ decl_event! {
...
+        /// Params:
+        /// - token identifier
+        /// - recovery account for the leftover funds
+        /// - leftover funds
+        RevenueSplitFinalized(TokenId, AccountId, JoyBalance),
+
+        /// Revenue Split issued
```
Lezek123:
```suggestion
        /// User participated in split
```

runtime-modules/project-token/src/events.rs
```diff
@@ -32,6 +45,36 @@ decl_event! {
...
+        /// Params:
+        /// - token identifier
+        /// - user account
+        /// - user allocated reserved balance
+        UserParticipatedToSplit(TokenId, AccountId, Balance),
```
Lezek123:
```suggestion
        UserParticipatedInSplit(TokenId, AccountId, Balance),
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,47 +239,365 @@ decl_module! {
...
+            TokenInfoById::<T>::mutate(token_id, |token_info| {
+                token_info.last_sale.as_mut().unwrap().quantity_left = <T as Trait>::Balance::zero();
+            });
+
+            Self::deposit_event(RawEvent::UnsoldTokensRecovered(token_id, token_info.sales_initialized, amount_to_recover));
+
+            Ok(())
+        }
+
+        /// Participate to the *latest* token revenue split (if ongoing)
```
Lezek123:
```suggestion
        /// Participate in the *latest* token revenue split (if ongoing)
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,47 +239,365 @@ decl_module! {
...
+        /// - `token.split_status` must be active AND THEN current_block in
+        ///    [split.start, split.start + split_duration)
+        /// - `account.staking_status.is_none()`
+        /// - `account.amount` >= `amount`
+        ///
+        /// Postconditions:
+        /// - `account.staking_status` set to Some(..) with `amount` and `token.latest_split`
+        /// no-op if `amount.is_zero()`
+        #[weight = 10_000_000] // TODO: adjust weight
+        fn participate_to_split(
```
Lezek123:
```suggestion
        fn participate_in_split(
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,47 +239,365 @@ decl_module! {
...
+        /// Preconditions
+        /// - `token` must exist for `token_id`
+        /// - `account` must exist for `(token_id, sender)` with `origin` signed by `sender`
+        /// - `token.revenue_split` is active
+        /// - `token.revenue_split` has ended
+        /// - `account.staking_status.is_some()`
+        /// - `account.staking_status` is Some(..) with split_id == `token.latest_split`
+        /// - let `dividend = split_allocation * account.staked_amount / token.supply``
+        ///    then `treasury` must be ablet to transfer `dividend` amount of JOY.
+        ///    This conditions technically, should always be satisfied
```
Lezek123:
```suggestion
        ///    This condition technically, should always be satisfied
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -237,47 +239,365 @@ decl_module! {
...
+        ///
+        /// Postconditions
+        /// - `dividend` amount of JOYs transferred to `treasury_account` to `sender`
```
Lezek123:
```suggestion
        /// - `dividend` amount of JOYs transferred from `treasury_account` to `sender`
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -444,14 +810,118 @@ impl<T: Trait> PalletToken<T::AccountId, TransferPolicyOf<T>, TokenIssuanceParam
...
+    /// - `token` must exist for `token_id`
+    /// - `token` revenue split status must be inactive
+    /// - `start` >= System::block_number()
+    /// - `duration` must be >= `T::MinRevenueSplitDuration`
+    /// - specified `reserve_source` must be able to *transfer* `allocation` amount of JOY
+    ///
+    /// PostConditions
+    /// - `allocation` transferred from `reserve_source` to `treasury_account` for `token_id`
+    /// - `token.revenue_split` set to `Active(..)` with timeline [start, start + duration)
+    ///    and `token.revenue_split.allocation_left = allocation`
```
Lezek123: `token.revenue_split.allocation_left` doesn't exist

runtime-modules/project-token/src/lib.rs
Lezek123: > A token holder staking V number of tokens in such an account will, **upon staking**, receive (V/Q)*X*(S/100) in $JOY from the channel balance to any account of their liking.

_Source: https://github.com/Joystream/joystream/issues/2362#issue-867837174_

I think your current design where the revenue split participant needs to wait until the end of the revenue split period in order to recieve JOY is flawed. All JOY should be distributed **before** the revenue split ends (_upon staking_, as mentioned in the comment above), the only 2 things that should happen **after** it ends are:
- Unstaking
- Finalizing

If you only allow claiming the JOY after the revenue split ends then you would need to introduce another period during which `claim_revenue_split_amount` can be called, but `finalize_revenue_split` cannot and you would also need to make sure that no issuance changes happen during this `claiming period`, but this would be just an unnecessary complication. There's no need to have `participation announcing period` -> `claiming + unstaking period` -> `finalizing`, since you can combine them into `participation + claiming period` -> `unstaking + finalizing`

My suggestion therefore is:
- combine `participate_in_split` and `claim_revenue_split_amount` into one method (but without the unstaking!)
- in `abandon_revenue_split` only allow unstaking if account's staking_status does not refer to the current, ongoing split (so allow only when either current split has ended or `account_data.staking_status.split_id < current_split_id`)

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,807 @@
...
+        // account id balance increased by leftover joy amount
+        assert_eq!(
+            Joy::<Test>::usable_balance(DEFAULT_ACCOUNT_ID),
+            DEFAULT_SPLIT_ALLOCATION + ExistentialDeposit::get()
+        );
+    })
+}
+
+#[test]
+fn participate_to_split_failst_with_invalid_token_id() {
```
Lezek123:
```suggestion
fn participate_in_split_fails_with_invalid_token_id() {
```
And in general `participate_to_split` => `participate_in_split`

runtime-modules/project-token/src/lib.rs
```diff
@@ -444,14 +810,118 @@ impl<T: Trait> PalletToken<T::AccountId, TransferPolicyOf<T>, TokenIssuanceParam
...
+    /// - `token.latest_split` incremented by 1
+    /// no-op if allocation is 0
+    fn issue_revenue_split(
+        token_id: T::TokenId,
+        start: T::BlockNumber,
```
Lezek123: I think this should be optional, since most of the time the issuer will probably want to start at the current block and it's a bad interface if it needs to be provided manually in that case (I know the `content` pallet could also provide it, but that's unnecessary, additional storage read then).

runtime-modules/project-token/src/tests/fixtures.rs
```diff
@@ -0,0 +1,865 @@
...
+            assert_eq!(state_pre, state_post)
+        }
+
+        result
+    }
+}
+
+pub fn treasury_account_for(token_id: u64) -> AccountId {
+    TokenModuleId::get().into_sub_account(token_id)
+}
```
Lezek123: This will always return the same account, but that's something specific to tests, because `ModuleId` is 8 bytes and `u64` (which is mock `AccountId`) is also 8 bytes.

`ModuleId.into_sub_account(token_id)` basically just takes the account prefix value (4 bytes), joins it with `ModuleId` and `TokenId` (in that case) and converts back to `AccountId`, but since prefix + `ModuleId` are already > 8 bytes, the `TokenId` value is just ignored.

In production, the `AccountId` is 32 bytes, so `prefix + ModuleId + TokenId` will convert to different `AccountId` for different `TokenId` and our current test environment is not reflecting this properly.

runtime-modules/project-token/src/lib.rs
```diff
@@ -98,23 +99,25 @@ decl_storage! {
...
-        pub BloatBond get(fn bloat_bond) config(): ReserveBalanceOf<T>;
+        pub BloatBond get(fn bloat_bond) config(): JoyBalanceOf<T>;
     }
 
     add_extra_genesis {
         build(|_| {
-            // We deposit some initial balance to the pallet's module account on the genesis block
-            // to protect the account from being deleted ("dusted") on early stages of pallet's work
+            // We deposit some initial balance to the pallet's module account
+            // (and sub-accounts!!) on the genesis block to protect the account
```
Lezek123: Expanding on the previous comment:
The only reason why this works in the test environment is because of the `u64` mock for `AccountId`.
Normally, you'd need to initialize a sub-account for each `token_id` separately, which is just not feasible, so I suggest using a single module account for all purposes.

runtime-modules/project-token/src/events.rs
```diff
@@ -1,17 +1,30 @@
...
         TransferPolicy = TransferPolicyOf<T>,
-        TokenData = TokenDataOf<T>,
+        TokenIssuanceParameters = TokenIssuanceParametersOf<T>,
         ValidatedTransfers = Transfers<Validated<<T as frame_system::Trait>::AccountId>, <T as crate::Trait>::Balance>,
     {
+
+        /// Token amount is unreserved
+        /// Params:
+        /// - token identifier
+        /// - account tokens are unreserved from
```
Lezek123: > No other term is allowed: liquidity, reserved, reservability, free balance, usability... its all banished!

_Source: https://github.com/Joystream/joystream/issues/3633_

Based on this, please restrain from using terms like `reserved` / `unreserved`. I suggest using `unstaked` instead.

runtime-modules/project-token/src/events.rs
```diff
@@ -32,6 +45,36 @@ decl_event! {
...
+        /// - token identifier
+        /// - recovery account for the leftover funds
+        /// - leftover funds
+        RevenueSplitFinalized(TokenId, AccountId, JoyBalance),
+
+        /// Revenue Split issued
+        /// Params:
+        /// - token identifier
+        /// - user account
+        /// - user allocated reserved balance
```
Lezek123: > No other term is allowed: liquidity, reserved, reservability, free balance, usability... its all banished!

_Source: https://github.com/Joystream/joystream/issues/3633_

Based on this, please restrain from using terms like `reserved` / `unreserved`. I suggest using `unstaked` instead.

Lezek123 (changes requested):
Most of the comments are about typos, so to highlight the most important stuff:
- All commits coming from `audit3` need to be removed from this branch (I suggest `git rebase -i ephesus`)
-  There is an issue with how the claiming process was implemented. Users should be able to claim JOY **during** the ongoing revenue split, not after it ends (this is explained further in one of my comments).
- `ModuleId.into_sub_account` in the test environment will always return the same account, which is not the case in production. This is also explained in one of the comments. Because of that I suggested that you only use one module account (or at least don't create a separate one for each creator token, as this would get quite complex due to existential deposits etc.)

runtime-modules/project-token/src/tests/fixtures.rs
dobertRowneySr: I added all the required functionalities starting from: 723e8b634a3442d6bd08ac9bf666b9245889ead8

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I have used a single module account for everything now: 4f043139f9005f5e3e918c05c39624aef72dbe2a

runtime-modules/project-token/src/lib.rs
```diff
@@ -883,7 +883,7 @@ impl<T: Trait>
     /// - `token.revenue_split` has ended
     ///
     /// Postconditions
-    /// - `token.revenue_split.allocation_left` of JOYs transferred to `account_id`
+    /// - `token.revenue_split.allocation` of JOYs transferred to `account_id`
```
Lezek123:
```suggestion
    /// - `token.revenue_split.leftovers()` of JOYs transferred to `account_id`
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -432,8 +432,13 @@ decl_module! {
         ///    [split.start, split.start + split_duration)
         /// - `account.staking_status.is_none()`
         /// - `account.amount` >= `amount`
+        /// - let `dividend = split_allocation * account.staked_amount / token.supply``
+        ///    then `treasury` must be ablet to transfer `dividend` amount of JOY.
```
Lezek123:
```suggestion
        ///    then `treasury` must be able to transfer `dividend` amount of JOY.
```

runtime-modules/project-token/src/events.rs
```diff
@@ -57,15 +57,7 @@ decl_event! {
...
-        UserParticipatedInSplit(TokenId, AccountId, Balance),
-
-        /// User claimed revenue split
-        /// Params:
-        /// - token identifier
-        /// - user account
-        /// - Revenue Amount in JOY
-        /// - Revenue split Id
-        UserClaimedRevenueSplit(TokenId, AccountId, JoyBalance, RevenueSplitId),
+        UserParticipatedInSplit(TokenId, AccountId, Balance, RevenueSplitId),
```
Lezek123: Add number of JOY claimed

runtime-modules/project-token/src/tests/fixtures.rs
Lezek123: Great! Looks like this function can now be removed

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,687 @@
...
+                allocation: DEFAULT_SPLIT_ALLOCATION,
+                dividends_payed: DEFAULT_SPLIT_JOY_DIVIDEND,
+                ..
+            })
+        ));
+    })
+}
+
+#[test]
+fn abandon_revenue_splitd_fails_with_invalid_token_id() {
```
Lezek123:
```suggestion
fn abandon_revenue_split_fails_with_invalid_token_id() {
```

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,687 @@
...
+        let result = AbandonRevenueSplitFixture::default()
+            .with_token_id(2u64)
+            .execute_call();
+
+        assert_err!(result, Error::<Test>::TokenDoesNotExist);
+    })
+}
+
+#[test]
+fn abandon_revenue_splitd_fails_with_invalid_account_id() {
```
Lezek123:
```suggestion
fn abandon_revenue_split_fails_with_invalid_account_id() {
```

runtime-modules/project-token/src/tests/mock.rs
```diff
@@ -107,15 +121,213 @@ impl frame_system::Trait for Test {
...
+    pub const InitialStorageBucketsNumberForDynamicBag: u64 = 3;
+    pub const MaxRandomIterationNumber: u64 = 3;
+    pub const DefaultMemberDynamicBagNumberOfStorageBuckets: u64 = 3;
+    pub const DefaultChannelDynamicBagNumberOfStorageBuckets: u64 = 4;
+    pub const DistributionBucketsPerBagValueConstraint: storage::DistributionBucketsPerBagValueConstraint =
+    storage::StorageBucketsPerBagValueConstraint {min: 3, max_min_diff: 7};
+    pub const MaxDataObjectSize: u64 = 1_000_000_000;
+
+    // --------- project token pallet constants --------------
+    pub const MinRevenueSplitDuration: u64 = 10;
```
Lezek123: Those are defined on line `88`

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,687 @@
...
+        let result = FinalizeRevenueSplitFixture::default()
+            .with_token_id(2u64)
+            .execute_call();
+
+        assert_err!(result, Error::<Test>::TokenDoesNotExist);
+    })
+}
+
+#[test]
+fn finalize_split_ok_with_active_but_not_ended_revenue_split() {
```
Lezek123:
```suggestion
fn finalize_split_fails_with_active_but_not_ended_revenue_split() {
```

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,687 @@
...
+
+        assert!(matches!(
+            Token::token_info_by_id(1u64).revenue_split,
+            RevenueSplitState::Inactive,
+        ));
+    })
+}
+
+#[test]
+fn finalize_split_ok_with_leftover_joys_transferred_to_account() {
```
Lezek123, line 300: In order to make sure this works as expected I suggest adding at least one claim during the split

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -0,0 +1,687 @@
...
+        // account id balance increased by leftover joy amount
+        assert_eq!(
+            Joy::<Test>::usable_balance(DEFAULT_ACCOUNT_ID),
+            DEFAULT_SPLIT_ALLOCATION + ExistentialDeposit::get()
+        );
+    })
+}
+
+#[test]
+fn participate_in_split_failst_with_invalid_token_id() {
```
Lezek123:
```suggestion
fn participate_in_split_fails_with_invalid_token_id() {
```

runtime-modules/project-token/src/errors.rs
```diff
@@ -39,5 +42,71 @@ decl_error! {
...
+        RevenueSplitNotOngoing,
+
+        /// User already participating in the revenue split
+        UserAlreadyParticipating,
+
+        /// User does not posses enough balance to participate in the revenue split
+        InsufficientBalanceForSplitParticipation,
+
+        /// Attempt to claim revenue split by a non participant
+        UserNotAParticipantForTheSplit,
```
Lezek123: This is now only thrown during `abandon_revenue_split`, so the comment is no longer valid.
I suggest also renaming to `UserNotAParticipantInAnySplit`

runtime-modules/project-token/src/events.rs
```diff
@@ -32,6 +37,35 @@ decl_event! {
...
+        /// - token identifier
+        /// - recovery account for the leftover funds
+        /// - leftover funds
+        RevenueSplitFinalized(TokenId, AccountId, JoyBalance),
+
+        /// User partipated in a revenue split
+        /// Params:
+        /// - token identifier
+        /// - user account
+        /// - user allocated staked balance
```
Lezek123, line 70: Missing revenue split id

runtime-modules/project-token/src/lib.rs
```diff
@@ -444,14 +775,118 @@ impl<T: Trait> PalletToken<T::AccountId, TransferPolicyOf<T>, TokenIssuanceParam
...
+    /// Issue a revenue split for the token
+    /// Preconditions:
+    /// - `token` must exist for `token_id`
+    /// - `token` revenue split status must be inactive
+    /// - `start` >= System::block_number()
+    /// - `duration` must be >= `T::MinRevenueSplitDuration`
+    /// - specified `reserve_source` must be able to *transfer* `allocation` amount of JOY
+    ///
+    /// PostConditions
+    /// - `allocation` transferred from `reserve_source` to `treasury_account` for `token_id`
```
Lezek123:
```suggestion
    /// - `allocation` transferred from `reserve_source` to `treasury_account`
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -539,24 +973,64 @@ impl<T: Trait> Module<T> {
...
         });
     }
 
     pub(crate) fn current_block() -> T::BlockNumber {
         <frame_system::Module<T>>::block_number()
     }
 
+    /// Computes (staked_amount / supply) * allocation
+    /// Preconditions:
+    /// - supply > user_staked_amount > 0
```
Lezek123:
```suggestion
    /// - supply >= user_staked_amount > 0
```

Lezek123 (changes requested):
Looks good, just a few cosmetic changes requested.

I'd also suggest adding a few more test cases to check for:
- correct calculation of `transferrable` balance when account has both `unvested` and `staked` balance (ie. `amount = 2000, unvested = 1500, staked = 1000, expected_transferrable = 500` and `amount = 2000, unvested = 1000, staked = 1500, expected_transferrable = 500`)
- test the `no-op` case for `issue_revenue_split` with `allocation == 0` and `participate_in_split` with `amount == 0`

Lezek123 (commented):
Looks good, just a few `cargo clippy` issues remaining to fix.

Lezek123 (approved):
Perfect :ok_hand: Great job, just need to wait for https://github.com/Joystream/joystream/pull/3618 before merging

runtime-modules/project-token/src/types.rs
```diff
@@ -94,6 +86,138 @@ pub struct TokenData<Balance, Hash, BlockNumber, TokenSale> {
 
     /// Account counter
     pub(crate) accounts_number: u64,
+
+    /// Revenue Split state info
+    pub(crate) revenue_split: RevenueSplitState,
+
+    /// Latest Token Revenue split (active / inactive)
+    pub(crate) latest_revenue_split_id: RevenueSplitId,
```
bedeho: This name does not match the type, what if there was no split? If this field really is id of next split, then lets call it that.

runtime-modules/project-token/src/types.rs
```diff
@@ -94,6 +86,138 @@ pub struct TokenData<Balance, Hash, BlockNumber, TokenSale> {
...
+/// Revenue Split Information for an *Active* revenue split
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+pub struct RevenueSplitInfo<JoyBalance, BlockNumber> {
+    /// Original Allocation
+    pub(crate) allocation: JoyBalance,
+
+    /// Split timeline [start, start + duration)
+    pub(crate) timeline: Timeline<BlockNumber>,
+
+    /// Dividends payed out after staking period is over
```
bedeho, line 163: `dividends_claimed` feels a bit stronger

runtime-modules/project-token/src/lib.rs
```diff
@@ -439,6 +444,124 @@ decl_module! {
...
+            // it should not really be possible to have supply == 0 with staked amount > 0
+            debug_assert!(!token_info.total_supply.is_zero());
+            let dividend_amount = Self::compute_revenue_split_dividend(
+                amount,
+                token_info.total_supply,
+                split_info.allocation,
+            );
+
+            let treasury_account: T::AccountId = Self::module_treasury_account();
+            Self::ensure_can_transfer_joy(&treasury_account, dividend_amount)?;
```
bedeho: How could this fail? The caller has a claim on a specific amount of JOY from this split.

runtime-modules/project-token/src/lib.rs
```diff
@@ -439,6 +444,124 @@ decl_module! {
...
+        /// no-op if `amount.is_zero()`
+        #[weight = 10_000_000] // TODO: adjust weight
+        fn participate_in_split(
+            origin,
+            token_id: T::TokenId,
+            amount: TokenBalanceOf<T>,
+        ) -> DispatchResult {
+            let sender = ensure_signed(origin)?;
+
+            if amount.is_zero() { return Ok(()) }
```
bedeho: Seems like it should be an error really?

runtime-modules/project-token/src/lib.rs
```diff
@@ -439,6 +444,124 @@ decl_module! {
...
+        /// - `token` must exist for `token_id`
+        /// - `account` must exist for `(token_id, sender)` with `origin` signed by `sender`
+        /// - `account.staking status.is_some()'
+        /// - if `(account.staking_status.split_id == token.latest_revenue_split_id`
+        ///    AND `token.revenue_split` is active) THEN split staking period  must be ended
+        ///
+        /// Postconditions
+        /// - `account.staking_status` set to None
+        #[weight = 10_000_000] // TODO: adjust weight
+        fn abandon_revenue_split(origin, token_id: T::TokenId) -> DispatchResult {
```
bedeho: This name had me long confused, was reading impl and expecting this was about exiting and refunding your released $JOY.

I think at least call it something like `leave_inactive_revenue_split` or something to that effect.

bedeho (approved):
Overall looks solid, good job :)

runtime-modules/project-token/src/types.rs
dobertRowneySr: Here's the reason why is not called `next_*`. It is initialized at `0` and its only reason to exist is for checking the condition:
`split_staking_status` for a given user account is referring to the current split. An alternative name could be:
`last_active_revenue_split_id`

runtime-modules/project-token/src/lib.rs
dobertRowneySr: Ok yes I agree on this

runtime-modules/project-token/src/types.rs
dobertRowneySr: renamed

runtime-modules/project-token/src/lib.rs
dobertRowneySr: This is meant to allow a user to effectively **exit** a revenue split after the staking period is expired:
how about `exit_revenue_split`?

runtime-modules/project-token/src/lib.rs
dobertRowneySr: The codition `dividend_amount <= usable_balance(treasury_account)` should be a runtime invariant.
Should I just remove it or maybe add a `debug_assert`?

bedeho: Yes please.

runtime-modules/project-token/src/lib.rs
bedeho: hmm, still confusing imho., but its fine.

runtime-modules/project-token/src/types.rs
bedeho: So we concluded to go with `next_` and have check for whether user is unstaking from ongoing split by just doing `next_ - 1`.

bedeho:
merge conflict.

runtime-modules/project-token/src/lib.rs
```diff
@@ -534,7 +534,7 @@ decl_module! {
         /// - `token` must exist for `token_id`
         /// - `account` must exist for `(token_id, sender)` with `origin` signed by `sender`
         /// - `account.staking status.is_some()'
-        /// - if `(account.staking_status.split_id == token.latest_revenue_split_id`
+        /// - if `(account.staking_status.split_id == token.next_revenue_split_id`
```
Lezek123:
```suggestion
        /// - if `(account.staking_status.split_id == token.next_revenue_split_id - 1`
```

Lezek123 (changes requested):

Lezek123 (approved):

Issue #3650, opened 2022-04-22
Revising destination of proceeds in NFT transactions


bedeho:
# Background

In `Rhodes`, I reviewed the impact of NFT ownership, reward account status for the destination of funds in both an auction and buy now scenarios for the purpose of harmonizing 

## Auction

This table summarizes destination of revenue (not royalties) when an auction is completed.

| NftOwner vs. Channel Reward Account | Not | Set |
|---------------------------|-----|-----|
| `NftOwner::Member` | member controller account | member controller account |
| `NftOwner::ChannelOwner` who is `Member` | burned | reward account  |
| `NftOwner::ChannelOwner` who is `CuratorGroup` | burned | reward account |

## Buy Now

This table summarizes destination of revenue (not royalties) when buy now (not in an auction) is completed.

| NftOwner vs. Channel Reward Account | Not | Set |
|---------------------------|-----|-----|
| `NftOwner::Member` | follows rules of two rows below: BUG! | reward account: BUG! |
| `NftOwner::ChannelOwner` who is `Member` | member controller account | reward account  |
| `NftOwner::ChannelOwner` who is `CuratorGroup` | _impossible_ * | reward account |

_* its not possible for a curator group owned channel to sell an NFT through channel ownership when there is no reward account._ 

# Proposal

Update payouts to work as follows for both `rhodes`, `ephesus` and `audit3`. Obviously, in the latter two, the channel reward account not set variations automatically go away. The bugs have been isolated in a separate issue

https://github.com/Joystream/joystream/issues/3648

## Auction

This table summarizes what happens with revenue (not royalties) when an auction is completed.

| NftOwner vs. Channel Reward Account | Not | Set |
|---------------------------|-----|-----|
| `NftOwner::Member` | member controller account | member controller account |
| `NftOwner::ChannelOwner` who is `Member` | **member controller account** | reward account  |
| `NftOwner::ChannelOwner` who is `CuratorGroup` | burned | reward account |

## Buy Now

This table summarizes what happens with revenue (not royalties) buy now (not in an auction) is completed.

| NftOwner vs. Channel Reward Account | Not | Set |
|---------------------------|-----|-----|
| `NftOwner::Member` | **member controller account** | **member controller account** |
| `NftOwner::ChannelOwner` who is `Member` | member controller account | reward account  |
| `NftOwner::ChannelOwner` who is `CuratorGroup` | **burned** | reward account |

dobertRowneySr:
[...]

PR #3654, opened 2022-04-22
nft owner now gets paid in buy_now

### LINKED ISSUE #3648: Bug: Nft Buy Now send revenue into channel account, not nft owner
# Problem
The extrinsic `buy_nft`

* sends funds to built in channel account if present, rather than nft owner account
* sends funds to either member channel owner if account is not present.
* blocks entire purrchase if account is not present and channel is owned by curator group

see here

https://github.com/Joystream/joystream/blob/audit3/runtime-modules/content/src/lib.rs#L1944

# Fix
In all cases, send revenue to controller account of owning member.

# PRs
We need two separate PRs for this

* one fixing this for `rhodes`.
* one fixing this for `ephesus` first of all, then pulled into `audit3`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202209932947637) by [Unito](https://www.unito.io)

dobertRowneySr:
fixes #3648 
and also implement this #3650


dobertRowneySr:
I added unit tests to confirm that now all proceedings from nft auction, sale, offer are burned when nft is owned by a channel curator group

runtime-modules/content/src/nft/mod.rs
```diff
@@ -283,51 +283,54 @@ impl<T: Trait> Module<T> {
...
+                Self::complete_payment(
+                    in_channel,
+                    nft.creator_royalty,
+                    price.to_owned(),
+                    new_owner_account_id,
+                    old_owner_account_id,
+                    false,
+                );
+            } else {
+                let _ = Balances::<T>::slash(&new_owner_account_id, price.to_owned());
```
bedeho: Can you add a comment here about what the scenario is, why there is this flow of royalites, proceeds and fee.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -283,51 +283,54 @@ impl<T: Trait> Module<T> {
...
-                nft.creator_royalty,
-                *price,
-                new_owner_account_id,
-                Some(owner_account_id),
-                false,
-            );
-
-            nft.owner = NftOwner::Member(new_owner);
+            if old_owner_account_id.is_some() {
+                Self::complete_payment(
```
bedeho: Can you add a comment here about what the scenario is, why there is this flow of royalites, proceeds and fee.

runtime-modules/content/src/lib.rs
```diff
@@ -2045,14 +2045,14 @@ decl_module! {
             // Ensure new pending offer is available to proceed
             Self::ensure_new_pending_offer_available_to_proceed(&nft, &receiver_account_id)?;
 
-            let owner_account_id = Self::ensure_owner_account_id(video.in_channel, &nft)?;
+            let nft_owner_account = Self::ensure_owner_account_id(video.in_channel, &nft).ok();
```
bedeho, line 2049: Add comments explaining what `nft_owner_account` will hold in various cases. Both the name and ensure method is very complex semantics, so we should be more clear for `ephesus`, but for `rhodes` adding comments can be decent crutch.

runtime-modules/content/src/lib.rs
```diff
@@ -2120,15 +2120,20 @@ decl_module! {
             // Ensure given participant can buy nft now
             Self::ensure_can_buy_now(&nft, &participant_account_id, price_commit)?;
 
-            let channel = Self::channel_by_id(&video.in_channel);
-            let owner_account_id =  Self::ensure_reward_account(&channel)?;
+            let old_nft_owner_account_id = Self::ensure_owner_account_id(video.in_channel, &nft).ok();
```
bedeho, line 2125: Same here some sort of comment as a crutch.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -283,51 +283,54 @@ impl<T: Trait> Module<T> {
...
-                Self::complete_payment(
-                    in_channel,
-                    nft.creator_royalty,
-                    *price,
-                    new_owner_account_id,
-                    Some(owner_account_id),
-                    false,
-                );
+                if owner_account_id.is_some() {
+                    Self::complete_payment(
```
bedeho: Can you add a comment here about what the scenario is, why there is this flow of royalites, proceeds and fee.

runtime-modules/content/src/nft/mod.rs
```diff
@@ -283,51 +283,54 @@ impl<T: Trait> Module<T> {
...
+                        in_channel,
+                        nft.creator_royalty,
+                        *price,
+                        new_owner_account_id,
+                        owner_account_id,
+                        false,
+                    );
+                } else {
+                    let _ = Balances::<T>::slash(&new_owner_account_id, *price);
```
bedeho: Can you add a comment here about what the scenario is, why there is this flow of royalites, proceeds and fee.

bedeho (changes requested):
This looks correct w.r.t. bugs and new behaviors intended, but I am requesting some comments to just make the code more clear, as we are not going to refactor `Rhodes` code at this point.

# `complete_payment`

If reward account is not set, then royalties are burned, even if owner of channel is a member who has a controller account. If he is selling the NFT as a member in this case ,he would get proceeds, but no royalties. Also, if someone else i doing the selling, he would not get royalites. This was not desired behaviour, so instead just make `creator_royalty: Option<Royalty, AccountId>`  a param to `complete_payment` , and caller has to figure out what to inject, this also removes ugly lookup into `channel_by_id` in there.

dobertRowneySr:
> This looks correct w.r.t. bugs and new behaviors intended, but I am requesting some comments to just make the code more clear, as we are not going to refactor `Rhodes` code at this point.
> # `complete_payment`
> 
> If reward account is not set, then royalties are burned, even if owner of channel is a member who has a controller account. If he is selling the NFT as a member in this case ,he would get proceeds, but no royalties. Also, if someone else i doing the selling, he would not get royalites. This was not desired behaviour, so instead just make `creator_royalty: Option<Royalty, AccountId>` a param to `complete_payment` , and caller has to figure out what to inject, this also removes ugly lookup into `channel_by_id` in there.

To address all this I tweaked the `complete_payment`, it should serve as a base for [this](https://github.com/Joystream/joystream/issues/3657#event-6483847257)

dobertRowneySr:
Just one marginal question:
- https://github.com/Joystream/joystream/blob/140d08e9f2fe559995e37eb1300add8dbaeeafe4/runtime-modules/content/src/lib.rs#L1905 should this be treated the same as https://github.com/Joystream/joystream/blob/140d08e9f2fe559995e37eb1300add8dbaeeafe4/runtime-modules/content/src/lib.rs#L1870

I renamed the `make_bid_payment` to something easier to understand

runtime-modules/content/src/nft/mod.rs
```diff
@@ -339,68 +335,54 @@ impl<T: Trait> Module<T> {
 
     /// Complete payment, either auction related or buy now/offer
     pub(crate) fn complete_payment(
-        in_channel: T::ChannelId,
-        creator_royalty: Option<Royalty>,
+        _in_channel: T::ChannelId,
```
bedeho: Why `_in_channel` here?

runtime-modules/content/src/nft/mod.rs
```diff
@@ -339,68 +335,54 @@ impl<T: Trait> Module<T> {
...
+        creator_royalty: Option<Royalty<T::AccountId>>,
         amount: BalanceOf<T>,
         sender_account_id: T::AccountId,
         receiver_account_id: Option<T::AccountId>,
         // for auction related payments
-        is_auction: bool,
+        _is_auction: bool,
```
bedeho: Why is `_is_auction` here?

runtime-modules/content/src/nft/types.rs
```diff
@@ -4,7 +4,7 @@ use super::*;
 pub type NftMetadata = Vec<u8>;
 
 /// Owner royalty
-pub type Royalty = Perbill;
+pub type Royalty<AccountId> = (Perbill, AccountId);
```
bedeho: Yes, this is not correct. We do not want the issuer to lock in an account, we only want the `complete_auction` method to statically either get a bompined royalty % & account, or neither. So you should just have left this type alone and updates signature of `complete_payment` to be `(..., royalty_payment: Option<Royalty, AccountId>`.

bedeho (changes requested):
On the question of how to deal with checking controller accounts, please see how this is done in other places in the runtime.

runtime-modules/content/src/lib.rs
```diff
@@ -1867,11 +1866,8 @@ decl_module! {
             let top_bid = english_auction.ensure_top_bid_exists::<T>()?;
             let top_bidder_id = top_bid.bidder_id;
 
-            // THIS WILL NOT FAIL UNLESS THE RUNTIME IS BROKEN
-            let top_bidder_account_result =
-                T::MemberAuthenticator::controller_account_id(top_bidder_id);
-            debug_assert!(top_bidder_account_result.is_ok());
-            let top_bidder_account_id = top_bidder_account_result.unwrap_or_default();
+            // THIS WILL NOT FAIL UNLESS RUNTIME IS BROKEN
+            debug_assert!(T::MemberAuthenticator::controller_account_id(top_bidder_id).is_ok());
```
bedeho: Why are we doing this assert? seems we don't actually need to fetch the account any longer, so why even do this call?

runtime-modules/content/src/nft/mod.rs
```diff
@@ -482,4 +461,34 @@ impl<T: Trait> Module<T> {
...
     }
+
+    pub(crate) fn build_royalty_payment_for_channel(
+        channel_id: T::ChannelId,
+        creator_royalty: Option<Royalty>,
+    ) -> Option<(Royalty, T::AccountId)> {
+        // payment is none if there is no royalty
+        if let Some(royalty) = creator_royalty {
+            // payment is none if creator channel has been deleted
+            if let Ok(channel) = Self::ensure_channel_exists(&channel_id) {
```
bedeho: This is a false check, no one can call `build_royalty_payment_for_channel` this if channel has been deleted?? 
Perhaps you can isntead pass channel istelf as input or some other thing to sidestep.

bedeho (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: This was suggested here https://github.com/Joystream/joystream/pull/3582#discussion_r846608185
Current runtime does not allow membership deletion. 
fixed @ baaae9f90948eaaf156c0cafcc773d53f13a32b0

runtime-modules/content/src/nft/mod.rs
dobertRowneySr: fixed here 832bf2ebf543493f7aa93d45a2acf55648695761

Issue #3631, opened 2022-04-19
Benchmark CRT


bedeho:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202414696932014) by [Unito](https://www.unito.io)

PR #3671, opened 2022-04-26
Creator token pallet benchmarking

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3631

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476232479975) by [Unito](https://www.unito.io)


runtime-modules/project-token/src/lib.rs
```diff
@@ -327,7 +362,17 @@ decl_module! {
...
         /// - account for `member_id` created and added to pallet storage
         /// - `bloat_bond` transferred from sender to treasury account
-        #[weight = 10_000_000] // TODO: adjust weights
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (H)` where:
+        /// - `H` is the length of `proof.0`
+        ///   - `O(1)` - doesn't depend on the state or parameters
```
dobertRowneySr, line 372:
```suggestion
         ///  - DB:
        ///   - `O(1)` - doesn't depend on the state or parameters
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -177,7 +186,16 @@ decl_module! {
...
         /// - `outputs.beneficiary` tokens amount increased by `amount`
-        #[weight = 10_000_000] // TODO: adjust weight
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (T)` where:
+        /// - `T` is the length of `outputs`
+        /// - DB:
+        ///   - `O(T)` - from the the generated weights
```
dobertRowneySr, line 196:
```suggestion
        ///   - `O(T)` - event fields written to Storage
```

dobertRowneySr: This is due to the deposited event (containing the validated transfers) that is written to Storage.
https://docs.substrate.io/main-docs/build/events-errors/

dobertRowneySr (changes requested):
Looks Ok in terms of benchmarking and performance accounting.  
 There is just a minor and subtle issue with the statement:
`token.sale.is_some() (affects the size of TokenData):` The size of an option is
size of `T` + bytes necessary to align `T`, in any case. So it is not necessary to start
a sale during the benchmarks in order to increase the size of token data (which does not
depend whether Option is some or not).
The same idea holds for split staking status and also with revenue split state field
(which is a enum and the same rules hold).
Removing these part should speed up the weight generation process (in theory):

https://users.rust-lang.org/t/why-does-an-option-around-t-imply-a-size-of-std-t-2/32945/2

runtime-modules/project-token/src/lib.rs
Lezek123: I think that's actually due to token balances of the destinations being updated.
Events don't seem to affect `writes` and `reads` in the weight function.

Lezek123:
> Looks Ok in terms of benchmarking and performance accounting.
There is just a minor and subtle issue with the statement:
token.sale.is_some() (affects the size of TokenData): The size of an option is
size of T + bytes necessary to align T, in any case. So it is not necessary to start
a sale during the benchmarks in order to increase the size of token data (which does not
depend whether Option is some or not).
The same idea holds for split staking status and also with revenue split state field
(which is a enum and the same rules hold).

Great observation.
I actually noticed that the impact on the weight was almost non-existing and I was wondering why that's the case.
Thank you for the detailed explanation!

dobertRowneySr (approved):
LGTM

Issue #3537, opened 2022-03-31
Tune Rhodes runtime parameter values


bedeho:
We need to revisit and make sure that these make sense in light of what we have found, in particular its critical that invited members can do certain key things.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201899819496282/1202055944788386) by [Unito](https://www.unito.io)

bedeho:
This would be one known problem: https://github.com/Joystream/joystream/issues/3519#issuecomment-1083767929

bwhm:
From `runtime/src/lib.rs` unless noted otherwise.

## Validator
```
pallet_staking_reward_curve::build! {
    const REWARD_CURVE: PiecewiseLinear<'static> = curve!(
        min_inflation: 0_050_000,
        max_inflation: 0_750_000,
        ideal_stake: 0_300_000,
        falloff: 0_050_000,
        max_piece_count: 100,
        test_precision: 0_005_000,
    );
}
```
- `max_inflation` from  0_750_000 to 0_180_000

## Forum

```
parameter_types! {
    pub const MaxCategoryDepth: u64 = 6;
    pub const MaxSubcategories: u64 = 20;
    pub const MaxThreadsInCategory: u64 = 20;
    pub const MaxPostsInThread: u64 = 20;
    pub const MaxModeratorsForCategory: u64 = 20;
    pub const MaxCategories: u64 = 20;
    pub const MaxPollAlternativesNumber: u64 = 20;
    pub const ThreadDeposit: u64 = 30;
    pub const PostDeposit: u64 = 10;
    pub const ForumModuleId: ModuleId = ModuleId(*b"mo:forum"); // module : forum
    pub const PostLifeTime: BlockNumber = 3600;
}
```
->
- `MaxCategories` from 20 to 40
  - Shamil sceptical
- `MaxSubcategories` from 20 to 40
  - Not sure if it's actually doing anything

## Proposals
```
pub(crate) fn set_council_budget_increment_proposal() -> ProposalParameters<BlockNumber, Balance> {
    ProposalParameters {
        voting_period: 72000,
        grace_period: 100800,
        approval_quorum_percentage: 66,
        approval_threshold_percentage: 80,
        slashing_quorum_percentage: 60,
        slashing_threshold_percentage: 80,
        required_stake: Some(200_000),
        constitutionality: 1,
    }
}
```
->
- `grace_period` from 100800 to 14400


## NFTs
`node/src/chain_spec/content_config.rs`
[...]

PR #3678, opened 2022-04-27
Rhodes update runtime constants

dobertRowneySr:
updated according to #3537


mnaamani (commented):
Changes are correct. However the changes to the NFT config values, `max_starting_price` and `max_bid_step`, only have affect at genesis/new chain. We need to update the implementation of `on_runtime_upgrade()` in `runtime/src/runtime_api.rs` to set the storage values in the content pallet.

mnaamani (approved):
LGTM just waiting on CI checks to complete.

PR #3693, opened 2022-04-29
Check price of data object bloat bonds

### LINKED ISSUE #3607: Check price of data object bloat bonds
# Background
We are looking to add video state bloat bond in the content pallet. This is needed as video objects occupy storage space. https://github.com/Joystream/joystream/issues/3394

# Proposal

**Implement `funds_needed_for_upload(:UploadParameters)`:**

When creating a video, the extrinsic `create_video` will need to ensure that the creator has enough funds for the state bloat bonds. Since creating a video often involves also adding data objects to the storage pallet by calling `upload_data_objects(:UploadParameters)`, the `create_video` needs to be able to ask the storage pallet the amount of funds needed for the state bloat bonds of the new data objects through some new `funds_needed_for_upload(:UploadParameters)`. It may be that `funds_needed_for_upload` does not actually need all values in `UploadParameters`, in which case consider introducing some simpler type.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202127980784428) by [Unito](https://www.unito.io)

L3pereira:
`funds_needed_for_upload(params: UploadParameters<T>)` created in pallet storage
I think It's more advantageous to use `UploadParameters` as a parameter than to separate the fields into params.

This function calculates the storage fee and the net prize an actor has to pay if he wants to update X amount of data objects.
It only works on updates, so the net deletion prize is always positive (`NetDeletionPrize::<T>::Pos(b)`).

`net_prize = number of objects to update * DataObjectDeletionPrizeValue.`
`storage_fee = DataObjectPerMegabyteFee * round_up_to_nearest_integer(size of total objects in bytes / 1_048_576)`

`funds_needed_for_upload = net_prize + storage_fee`

- Closes #3607

### DIFF
--- a/runtime-modules/storage/src/lib.rs
+++ b/runtime-modules/storage/src/lib.rs
@@ -214,6 +214,14 @@ pub trait DataObjectStorage<T: Trait> {
     /// - balance of data size fee + total deletion prize is transferred from caller to treasury account
     fn upload_data_objects(params: UploadParameters<T>) -> DispatchResult;
 
+    /// Returns the funds needed to upload a num of objects
+    /// This is dependent on (data_obj_deletion_prize * num_of_objs_to_upload) plus a storage fee that depends on the
+    /// objs_total_size_in_bytes
+    fn funds_needed_for_upload(
+        num_of_objs_to_upload: usize,
+        objs_total_size_in_bytes: u64,
+    ) -> BalanceOf<T>;
+
     /// Validates moving objects parameters.
     /// Validates voucher usage for affected buckets.
     ///
@@ -3492,6 +3500,19 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
         DataObjectsById::<T>::iter_prefix(&bag_id)
             .map(|x| x.0)
             .collect()
+    }
+
+    fn funds_needed_for_upload(
+        num_of_objs_to_upload: usize,
+        objs_total_size_in_bytes: u64,
+    ) -> BalanceOf<T> {
+        let num_of_objs_to_upload = num_of_objs_to_upload.saturated_into();
+        let deletion_fee =
+            Self::data_object_deletion_prize_value().saturating_mul(num_of_objs_to_upload);
+
+        let storage_fee = Self::calculate_data_storage_fee(objs_total_size_in_bytes);
+
+        deletion_fee.saturating_add(storage_fee)
     }
 }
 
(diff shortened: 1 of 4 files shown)


runtime-modules/storage/src/lib.rs
```diff
@@ -3493,6 +3498,73 @@ impl<T: Trait> DataObjectStorage<T> for Module<T> {
...
+                Self::compute_net_prize(init_net_prize, creation_list.len(), 0)
+            }
+            _ => Default::default(),
+        };
+
+        // storage fee: zero if VoucherUpdate is default
+        let storage_fee = Self::calculate_data_storage_fee(new_voucher_update.objects_total_size);
+        //Only the case where the net deletion prize (NDP) is positive will be returned, as the
+        //function will be used to validate whether the actor has enough balance to update the assets.
+        let funds_needed_for_upload = match net_prize.add_balance(storage_fee) {
```
bedeho: Won't `net_prize` include old deletin prize? If so, wouldn't `funds_needed_for_upload` in fact be the new total cost, rather than cost of doing this additional upload, which is what we care about?

bedeho (changes requested):
This implementation is doing too much. You should not be checking and verifying lots of checks that are not relevant for fee estimation, those will be checked later as part of actual call to upload. Checking for existance of bag is OK, since we need contained informatino to do actual evaluation, but the other stuff is not.

runtime-modules/storage/src/lib.rs
L3pereira: When we call `upload_data_objects` this function will call `try_mutating_storage_state` with 
`BagOperationParams::Update(object_creation_list)`
then 

1. we will ensure bag exists and bag has a deletion prize, that  is some for dynamic bags or none for static ones.
2. Construct objects to upload  where deletion prize is the old one
3. we compute net prize init_value + (deletion_prize * num_obj_to_create)
4. We add net_prize to storage_fee

 So net prize include deletion prize in the calculation, but that is what `try_mutating_storage_state` is doing to do the ensure_funds validation 
I think the goal here is to separate funds validation from try_mutate_storage, right?

L3pereira:
Now we have a much simpler funds calculation function. This function doesn't fail.
I didn't check for bag existence, let's move that responsibility somewhere else, since the main goal of this function is calculating funds.
The signature is very simple 
```rust
    fn funds_needed_for_upload(
        num_of_objs_to_upload: usize,
        objs_total_size_in_bytes: u64,
    ) -> BalanceOf<T>;
``` 
I tested for overflow saturation, but I'm thinking now, these numbers are so huge, that no quantity of memory will ever amount to that.
This function is will roundup to the next greater integer when objs_total_size_in_bytes % 1048576 (1Mb) > 0

I also move to fixtures some helper functions, that were in the middle of the unit tests.

bedeho (approved):
This looks very simple and correct, originally, this is what I thought we would need, something very very simple. Why were you able to simplify so much only now?

L3pereira:
I thought your goal was to separate the `try_mutate` in such a way we would have validation and then the mutation safe parts for the content pallet, and the funds function would have those validations, But for now, you just want a calculator.

Issue #3683, opened 2022-04-28
commentCreatedEvents error


rafalpawlow:
The following query produces an error, probably because of nesting. We should be able to query this event by `channel` or `video`.

```
query GetCommentsNotifications($channelId: ID, $limit: Int!) {
  commentCreatedEvents (
    where: { comment: { video: { channel: { id_eq: $channelId } } } }
    limit: $limit
    orderBy: [createdAt_DESC]
  ) {
    comment {
      id
      createdAt
      video {
        id
        title
      }
    }
  }
}
```
![Zrzut ekranu 2022-04-28 o 10 38 37](https://user-images.githubusercontent.com/13706246/165732591-5edaebfd-3b7d-4b08-92f8-e9c1f34c7e2e.png)

kdembler:
@zeeshanakram3 Could we add `video` directly to this event type?

PR #3710, opened 2022-05-05
Comments and reactions flatten video property

zeeshanakram3:
Implements #3683


ondratra (commented):
LGTM. I don't approve because https://github.com/Joystream/joystream/pull/3340 should be merged first. After that, we can merge this PR as well.

kdembler:
Hey @zeeshanakram3, unfortunately, this doesn't solve the issue. That's partly my mistake because I suggested placing `video` on that event, but if you look at the original query Rafał posted in the issue, even with the video on the event, the query would be still too deep: `where: { video: { channel: { id_eq: $channelId } } }`. I think we will need to put the `channel` on the event as well, but maybe for clarity what this represents let's use slightly different name, maybe `videoChannel`?

kdembler (approved):

Issue #3599, opened 2022-04-13
Issuer account creation during transfers on permissioned token


bedeho:
It would likely be useful for the issuer to be able to make transfers which automatically create accounts for beneficiaries who do not have accounts, despite the transfer policy being permissioned. Such transfers should also be able to encumber the new funds by a some new vesting schedule. Making this a separate extrinsic may help keep the basic transfer clean and simple. Issuer pays state bloat bonds.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202121258609960) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/3719

PR #3719, opened 2022-05-06
Creator tokens: Issuer transfers

Lezek123:
Introduces `issuer_transfer` method in the creator token pallet, based on the specification in issue https://github.com/Joystream/joystream/issues/3599


bedeho (approved):
This was really well done, despite the inherent complexity. 

I think I would prefer a bit more inline comments and breaking up these very large methods (`do_transfer`) into a few smaller methods, even tho those smaller methods may not be called more than once, just to increase pace at which someone can build a good mental model of what they are reading, thereby also increasing likelihood of catching a mistake within a set period of time.

Given the complexity I think its very important @ignazio-bovo also does a careful review.

runtime-modules/project-token/src/lib.rs
```diff
@@ -909,27 +991,29 @@ impl<T: Trait> Module<T> {
...
+            Self::ensure_account_data_exists(token_id, dst),
+            is_issuer,
         ) {
-            (&TransferPolicyOf::<T>::Permissionless, Err(_)) => {
-                Ok(Validated::<_>::NonExisting(dst))
-            }
-            (&TransferPolicyOf::<T>::Permissionless, Ok(_)) => Ok(Validated::<_>::Existing(dst)),
-            (&TransferPolicyOf::<T>::Permissioned(_), Ok(_)) => Ok(Validated::<_>::Existing(dst)),
-            (&TransferPolicyOf::<T>::Permissioned(_), Err(e)) => Err(e),
+            (&TransferPolicyOf::<T>::Permissionless, Err(_), _) => Ok(None),
```
dobertRowneySr: This guard method has become very complex, if I am not wrong it can be splitted into:
1. ensure that `Permissioned` and `is_issuer == false` does not happen (this can be a stand-alone method that checks only this)
2. calling `ensure_account_data_exist(dst).ok()` which should give the same result as the ok case 

WDYT?

runtime-modules/project-token/src/lib.rs
```diff
@@ -942,15 +1026,36 @@ impl<T: Trait> Module<T> {
...
         transfer_policy: &TransferPolicyOf<T>,
-    ) -> Result<ValidatedTransfers<T>, DispatchError> {
+        is_issuer: bool,
+    ) -> Result<ValidatedTransfersOf<T>, DispatchError> {
         let result = transfers
+            .0
             .into_iter()
             .map(|(dst, payment)| {
-                Self::validate_destination(token_id, dst, transfer_policy).map(|res| (res, payment))
+                let dst_acc_data =
```
dobertRowneySr: Referring to the previous comment, I would say:
1. use `ensure_account_data_exist` here,
2. if necessary add the vesting schedule params after making sure that the extra VS can be added
3. return tagged `account_id` with `Existing/NonExistng` + payment params

dobertRowneySr (changes requested):
Logic is ok. I have suggested a few improvements that could increase readability.
Let me know what you think

runtime-modules/project-token/src/lib.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3719/commits/f3f6c258ae8c707a878ed568bef3c133b213b866, hope I got that right

runtime-modules/project-token/src/lib.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/3719/commits/f3f6c258ae8c707a878ed568bef3c133b213b866, hope I got that right

dobertRowneySr (approved):
LGTM

PR #3723, opened 2022-05-09
Distributor node: Fix mime-type detection

Lezek123:
Removed dependency on `read-chunk` module due to an issue of unknown exact cause:
```
{
   "@timestamp":"2022-05-09T10:03:09.043Z",
   "log.level":"error",
   "message":"Error while trying to detect object mimeType: fsP.open is not a function",
   "ecs":{
      "version":"1.6.0"
   },
   "error":{
      "type":"TypeError",
      "message":"fsP.open is not a function",
      "stack_trace":"TypeError: fsP.open is not a function\n    at module.exports (/joystream/node_modules/with-open-file/index.js:13:6)\n    at readChunk (/joystream/node_modules/read-chunk/index.js:11:9)\n    at ContentService.detectMimeType (/joystream/distributor-node/src/services/content/ContentService.ts:174:37)\n    at /joystream/distributor-node/src/services/content/ContentService.ts:284:37\n    at internal/util.js:435:14\n    at finish (internal/streams/pipeline.js:162:7)\n    at internal/util.js:435:14\n    at WriteStream.<anonymous> (internal/streams/pipeline.js:65:7)\n    at WriteStream.<anonymous> (internal/util.js:435:14)\n    at WriteStream.onclose (internal/streams/end-of-stream.js:123:14)"
   },
   "label":"ContentService",
   "objectId":"3",
   "objectPath":"/data/3"
}
```
Implemented custom `readFileChunk` method in the `ContentService` instead.

Since the detected `mime-type` is beeing cached in distributor-node's `/{directories.cacheState}`, in order to reload mime types, the node operator needs to execute a command like:
```
cat /path/to/cache.json | jq '.mimeTypeByObjectId |= []' > tmp.json && mv tmp.json /path/to/cache.json
```
And restart the node

### DIFF
--- a/distributor-node/src/services/content/ContentService.ts
+++ b/distributor-node/src/services/content/ContentService.ts
@@ -9,7 +9,7 @@ import { Readable, pipeline } from 'stream'
 import { NetworkingService } from '../networking'
 import { ContentHash } from '../crypto/ContentHash'
 import { PendingDownloadStatusType } from '../networking/PendingDownload'
-import { promisify } from 'util';
+import { promisify } from 'util'
 
 export const DEFAULT_CONTENT_TYPE = 'application/octet-stream'
 export const MIME_TYPE_DETECTION_CHUNK_SIZE = 4100
@@ -168,15 +168,15 @@ export class ContentService {
     return new FileContinousReadStream(this.path(objectId), options)
   }
 
-  public async readFileChunk(path: string, bytes: number) {
-    const open = promisify(fs.open);
-    const read = promisify(fs.read);
-    const close = promisify(fs.close);
-    const buff = Buffer.alloc(bytes);
-    const fd = await open(path, 'r');
-    await read(fd, buff, 0, bytes, null);
-    await close(fd);
-    return buff;
+  public async readFileChunk(path: string, bytes: number): Promise<Buffer> {
+    const open = promisify(fs.open)
+    const read = promisify(fs.read)
+    const close = promisify(fs.close)
+    const buff = Buffer.alloc(bytes)
+    const fd = await open(path, 'r')
+    await read(fd, buff, 0, bytes, null)
+    await close(fd)
+    return buff
   }
 
   public async detectMimeType(objectId: string): Promise<string> {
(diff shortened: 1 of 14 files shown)


distributor-node/src/services/content/ContentService.ts
```diff
@@ -168,10 +168,21 @@ export class ContentService {
     return new FileContinousReadStream(this.path(objectId), options)
   }
 
+  public async readFileChunk(path: string, bytes: number): Promise<Buffer> {
+    const open = promisify(fs.open)
+    const read = promisify(fs.read)
+    const close = promisify(fs.close)
+    const buff = Buffer.alloc(bytes)
```
mnaamani, line 172: I assume garbage collector will handle cleanup of allocated memory?

mnaamani: It looks very neat, but does it make sense to promisify the `fs` methods somewhere once instead of repeatedly in this function. Unless it causes any slow down its okay for now.

mnaamani:
The underlying package which read-chunk was using `https://www.npmjs.com/package/with-open-file` seems to be quite old last published 3y ago. `fsP` does seem to indicate it was also promisifying the `fs` module. So I guess it is good we are dropping it.

distributor-node/src/services/content/ContentService.ts
Lezek123: > I assume garbage collector will handle cleanup of allocated memory?

Yes

> It looks very neat, but does it make sense to promisify the fs methods somewhere once instead of repeatedly in this function.

Factored this code out in https://github.com/Joystream/joystream/pull/3723/commits/7fa075b52fc5a9e4cd6aa760a17f4c9c57679f53

Lezek123:
I bumped the version and exposed it on the `/api/v1/status` endpoint in order to easily verify which nodes have been upgraded.

mnaamani:
Great adding the version to the status page.

Confirmed to work now on safari!

Screen Shot 2022-05-11 at 12 19 40 AM

mnaamani (approved):
Playback works now in safari browser.
I wasn't able to test clearing cache however.

PR #3724, opened 2022-05-09
fix: set ownerMember in EnglishAuctionSettledEvent to previous owner

zeeshanakram3:
set `ownerMember` in `EnglishAuctionSettledEvent` to previous owner(not the auction winner)

### DIFF
--- a/query-node/mappings/src/content/nft.ts
+++ b/query-node/mappings/src/content/nft.ts
@@ -269,7 +269,11 @@ async function finishAuction(
   }
 
   // load video and auction
-  const { video, auction } = await getCurrentAuctionFromVideo(
+  const {
+    video,
+    auction,
+    nft: { ownerMember: previousOwnerMember, ownerCuratorGroup: previousOwnerCuratorGroup },
+  } = await getCurrentAuctionFromVideo(
     store,
     videoId.toString(),
     `Non-existing video's auction was completed`,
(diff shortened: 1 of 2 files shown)


query-node/mappings/src/content/nft.ts
```diff
@@ -865,14 +869,16 @@ export async function contentNft_EnglishAuctionSettled({ event, store }: EventCo
...
+    store,
+    videoId.toNumber(),
+    event.blockNumber
+  )
 
   if (winnerId.toString() !== winningBid.bidder.id) {
     return inconsistentState(`English auction winner haven't placed the top bid`, { videoId, winnerId })
   }
 
-  nft.ownerMember = winningBid.bidder
```
kdembler, line 874: Why was this removed? Won't this create a bug of owner not being updated?

zeeshanakram3: This expression was redundant. The new owner has already been set before this line, in `resetNftTransactionalStatusFromVideo` [function](https://github.com/Joystream/joystream/blob/6eea1608f610ddf4a5f5588eaf6d8ca3705468a3/query-node/mappings/src/content/nft.ts#L168) (which is called from `finishAuction` [function](https://github.com/Joystream/joystream/blob/6eea1608f610ddf4a5f5588eaf6d8ca3705468a3/query-node/mappings/src/content/nft.ts#L294))

kdembler: Okay, thanks for clarification

kdembler (approved):

Issue #3712, opened 2022-05-05
CRT: Cannot dust non empty accounts


dobertRowneySr:
according to this : https://github.com/Joystream/joystream/issues/3617#issuecomment-1104766789

##### Current behavior
owner can delete his own account even when it is not **empty**

##### Desired Behavior
only **empty** accounts can be dusted 

empty as #3633

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202251388617609) by [Unito](https://www.unito.io)

PR #3741, opened 2022-05-12
Creator tokens: Allow dusting only for empty accounts

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3712


dobertRowneySr (approved):
LGTM

PR #3746, opened 2022-05-12
Storage pallet refunds based on old data object prizes

### LINKED ISSUE #3698: Bug: Storage pallet refunds based on old data object prizes
In `compute_net_prize` it uses `data_object_deletion_prize_value` which returns _current_ prize,  not what the actual prize of the objects being deleted. This may be hard to fix, may require invasive change, but first we must verify that this is a real bug.
Lets do two things

1. Verify that this is real
1. Determine why tests did not catch this.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202210185299638) by [Unito](https://www.unito.io)

L3pereira:
- Closes #3698 

This solves the problem, the storage pallet has, 
We update a bag with a new object at T0 with a `DataObjectDeletionPrizeValue` (ex. 10)
At T1 we update `DataObjectDeletionPrizeValue` to 20
At T2 we delete the object created at T0
The deleting operation would withdraw 20 from the treasury, this behaviour can be used to steal funds.

This fix was a bit invasive, 
`ensure_sufficient_balance` has been changed to return all parts of the amounts to pay and receive 
```rust
        // check that user or treasury account have enough balance
        let (
            bag_deletion_prize_request,
            obj_deletion_prize_request,
            bag_deletion_prize_refund,
            obj_deletion_prize_refund,
            storage_fee,
        ) = Self::ensure_sufficient_balance(&bag_op, objects_to_update_voucher, &account_id)?;
```
This is used later to update `total_data_obj_deletion_prize_paid`, this variable has been added to the `Bag`
```rust
pub struct BagRecord<StorageBucketId: Ord, DistributionBucketId: Ord, Balance> {
    /// Associated storage buckets.
    pub stored_by: BTreeSet<StorageBucketId>,

    /// Associated distribution buckets.
    pub distributed_by: BTreeSet<DistributionBucketId>,

    /// Bag deletion prize (valid for dynamic bags).
    pub deletion_prize: Option<Balance>,

    /// Total object size for bag.
    pub objects_total_size: u64,

    /// Total object number for bag.
    pub objects_number: u64,

    /// Total amount of data object deletion prizes paid by members
    /// This amount decreases in case of refunds (delete operations)
    pub total_data_obj_deletion_prize_paid: Balance, <---------------------------------------
}
```
We use `total_data_obj_deletion_prize_paid` to delete the bag, instead of searching for all objects 
and sum all deletion_prizes, we just refund `total_data_obj_deletion_prize_paid` amount.
This variable is used also in the following functions, since we are decreasing objects from one bag and increasing the other.
 
```rust
fn move_data_objects(
        src_bag_id: BagId<T>,
        dest_bag_id: BagId<T>,
        objects: BTreeSet<T::DataObjectId>,
    ) -> DispatchResult;

fn change_storage_bucket_vouchers_for_bag(
        bag_id: &BagId<T>,
        bag: &Bag<T>,
        bag_change: &BagUpdate<T::Balance>,
        voucher_operation: OperationType,
    );
```

`compute_deletion_prizes` was splitted into two simpler functions 
```rust
    fn compute_obj_deletion_prize_request(num_obj_to_create: usize) -> BalanceOf<T> {
        let num_obj_to_create: BalanceOf<T> = num_obj_to_create.saturated_into();
        let obj_deletion_prize = Self::data_object_deletion_prize_value();

        num_obj_to_create.saturating_mul(obj_deletion_prize)
    }

    fn compute_obj_deletion_prize_refund(
        obj_to_delete: &BTreeMap<T::DataObjectId, BalanceOf<T>>,
    ) -> BalanceOf<T> {
        obj_to_delete
            .values()
            .cloned()
            .fold(Zero::zero(), |acc, val| acc.saturating_add(val))
    }
```
The deletion prizes were separated into different variables, to make it easier to calculate each situation.
Ex. When we transfer the funds, we calculate `bag_deletion_prize_request + obj_deletion_prize_request`
```rust
        let deletion_prize_request =
            bag_deletion_prize_request.saturating_add(obj_deletion_prize_request);
        if !deletion_prize_request.is_zero() {
            <StorageTreasury<T>>::deposit(&account_id, deletion_prize_request)?;
            //  slash: no-op if storage_fee = 0
            let _ = Balances::<T>::slash(&account_id, storage_fee);
        }
```

Later, we update `total_data_obj_deletion_prize_paid` by calculating
`total_data_obj_deletion_prize_paid + obj_deletion_prize_request - obj_deletion_prize_refund`

### DIFF
--- a/runtime-modules/storage/src/lib.rs
+++ b/runtime-modules/storage/src/lib.rs
@@ -189,7 +189,8 @@ pub trait WeightInfo {
 
 type WeightInfoStorage<T> = <T as Trait>::WeightInfo;
 
-type DeletionPrizeAndStorageFee<T> = (BalanceOf<T>, BalanceOf<T>, BalanceOf<T>);
+type DataObjAndDeletionPrizeAndObjSize<T> =
+    Result<(Vec<DataObject<BalanceOf<T>>>, BalanceOf<T>, u64), DispatchError>;
 
 /// Content ID length in bytes (46 bytes multihash).
 pub const CID_LENGTH: usize = 46;
@@ -953,14 +954,6 @@ struct VoucherUpdate {
 }
 
 impl VoucherUpdate {
-    fn add_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
-        list.iter().fold(self, |acc, obj| acc.add_object(obj.size))
-    }
-
-    fn sub_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
-        list.iter().fold(self, |acc, obj| acc.sub_object(obj.size))
-    }
-
     fn get_updated_voucher(&self, voucher: &Voucher, voucher_operation: OperationType) -> Voucher {
         let (objects_used, size_used) = match voucher_operation {
             OperationType::Increase => (
@@ -985,13 +978,6 @@ impl VoucherUpdate {
         Self {
             objects_number: self.objects_number.saturating_add(1),
             objects_total_size: self.objects_total_size.saturating_add(size),
-        }
-    }
-
-    fn sub_object(self, size: u64) -> Self {
-        Self {
-            objects_number: self.objects_number.saturating_sub(1),
-            objects_total_size: self.objects_total_size.saturating_sub(size),
         }
     }
 }
@@ -4283,25 +4269,6 @@ impl<T: Trait> Module<T> {
             .collect()
     }
 
-    fn compute_deletion_prizes(
-        deletion_prize_request: BalanceOf<T>,
-        deletion_prize_refund: BalanceOf<T>,
-        num_obj_to_create: usize,
-        num_obj_to_delete: usize,
-    ) -> (BalanceOf<T>, BalanceOf<T>) {
-        let obj_deletion_prize = Self::data_object_deletion_prize_value();
-        let num_obj_to_create: BalanceOf<T> = num_obj_to_create.saturated_into();
-        let num_obj_to_delete: BalanceOf<T> = num_obj_to_delete.saturated_into();
-
-        let deletion_prize_request: BalanceOf<T> = deletion_prize_request
-            .saturating_add(num_obj_to_create.saturating_mul(obj_deletion_prize));
-
-        let deletion_prize_refund: BalanceOf<T> = deletion_prize_refund
-            .saturating_add(num_obj_to_delete.saturating_mul(obj_deletion_prize));
-
-        (deletion_prize_request, deletion_prize_refund)
-    }
-
     /// Utility function that checks existence for bag deletion / update &
     /// verifies that dynamic bag does not exists in case of bag creation
     /// returning a new one
@@ -4344,19 +4311,27 @@ impl<T: Trait> Module<T> {
         } = Self::retrieve_dynamic_bag(&bag_op)?;
 
         // check and generate any objects to add/remove from storage
-        let object_creation_list = Self::construct_objects_to_upload(&bag_op)?;
-        let objects_removal_list = Self::construct_objects_to_remove(&bag_op)?;
+        let (object_creation_list, deletion_prize_request, upload_objs_size) =
+            Self::construct_objects_to_upload(&bag_op)?;
 
-        let objects_to_update_voucher = VoucherUpdate::default()
-            .add_objects_list(object_creation_list.as_slice())
-            .sub_objects_list(objects_removal_list.as_slice());
+        let (objects_removal_list, deletion_prize_refund, remove_objs_size) =
+            Self::construct_objects_to_remove(&bag_op, deletion_prize.unwrap_or_else(Zero::zero))?;
+
+        // storage fee: zero if VoucherUpdate is default
+        let storage_fee = Self::calculate_data_storage_fee(upload_objs_size);
+
+        let objects_number = objects_number
+            .saturating_add(object_creation_list.len() as u64)
+            .saturating_sub(objects_removal_list.len() as u64);
+
+        let objects_total_size = objects_total_size
+            .saturating_add(upload_objs_size)
+            .saturating_sub(remove_objs_size);
 
         let all_objects_with_update_voucher = VoucherUpdate {
             objects_total_size,
             objects_number,
-        }
-        .add_objects_list(object_creation_list.as_slice())
-        .sub_objects_list(objects_removal_list.as_slice());
+        };
 
         // new candidate storage buckets
         let new_storage_buckets = Self::update_storage_buckets(
@@ -4364,18 +4339,13 @@ impl<T: Trait> Module<T> {
             all_objects_with_update_voucher,
             &bag_op.params,
         )?;
+
         // new candidate distribution buckets
         let new_distribution_buckets =
             Self::update_distribution_buckets(&distributed_by, &bag_op.params);
 
         // check that user or treasury account have enough balance
-        let (deletion_prize_request, deletion_prize_refund, storage_fee) =
-            Self::ensure_sufficient_balance(
-                &bag_op,
-                objects_to_update_voucher,
-                deletion_prize,
-                &account_id,
-            )?;
+        Self::ensure_sufficient_balance(&account_id, deletion_prize_request, storage_fee)?;
 
         //
         // == MUTATION SAFE ==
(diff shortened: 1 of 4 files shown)


bedeho:
> This fix was a bit invasive,

Yes, this is concerning. Why is this bugfix so invasive? All you need to do is to simply remember the bloat fee for each data object, and use that during deletion. Why is this doing so much?

L3pereira:
> In `compute_net_prize` it uses `data_object_deletion_prize_value` which returns _current_ prize, not what the actual prize of the objects being deleted. **This may be hard to fix, may require invasive change, but first we must verify that this is a real bug**. Lets do two things

As you said in the [issue comments](https://github.com/Joystream/joystream/issues/3698#issue-1222652981), this bug is invasive and hard to fix. Changing how delete_prizes are calculated, can propagate the change to various places in the codebase. 
It's not just calling a global value in storage, the delete_prizes values are stored in each of the objects, I have to make sure that the delete prize value is available, avoiding in this way doing calls to the storage.
We also have the case when the bag is deleted, here I have to remember the sum of all objects, otherwise, I had to search all objects (expensive)

PS the next issue - removing the bag deletion_prize, will get this calculation a little bit more cleaner
Ex. this will return only three parameters since bag_deletion_prizes will be removed
```rust
        // check that user or treasury account have enough balance
        let (
            bag_deletion_prize_request,
            obj_deletion_prize_request,
            bag_deletion_prize_refund,
            obj_deletion_prize_refund,
            storage_fee,
        ) = Self::ensure_sufficient_balance(&bag_op, objects_to_update_voucher, &account_id)?;
```

bedeho:
> I have to make sure that the delete prize value is available, avoiding in this way doing calls to the storage.

What do you mean by this part?

L3pereira:
I realize that in case of bag deletion, we iterate all bag objects.
When we call `construct_objects_to_remove`
https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L4390
we iterate all objects 
https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L4527-L4529

But here, we are basically subtracting `objects_removal_list` to `objects_number` and `objects_total_size`, this will always be zero in case of bag deletion. 
`objects_removal_list` is not used anymore, so I guess we shouldn't iterate all objects.

https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L4393-L4398

https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L957-L959

What made me create a `total_data_obj_deletion_prize_paid` variable was this one, here we call a variable that accumulates the number of objects, so I thought we should do the same for deletion prizes 
https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L4557-L4561

Basically, If we iterate all objects, we can drop total_data_obj_deletion_prize_paid and the code base gets simpler but is less efficient

bedeho:
So I had forgotten that this actually does iteration of the bag. I think the idea then was that the extrinsic which uses this, e.g. in the content directory, will need to get a witness from the user attesting to the size of the bag about to be deleted, such that this can be captured by the weight function. We have not started benchmarking the contnet directory yet, so this will come up then.

> But here, we are basically subtracting objects_removal_list to objects_number and objects_total_size, this will always be zero in case of bag deletion.

Yes this is bad, but remember they will be iterated later any way in

`BagOperationParams::<T>::Delete => DataObjectsById::<T>::remove_prefix(&bag_op.bag_id),`

where all objects are deleted. It is quite wasteful to do it needlessly though

> Basically, If we iterate all objects, we can drop total_data_obj_deletion_prize_paid and the code base gets simpler but is less efficient

I think we cannot avoid iteration anyway if we are allowing deletion of non-empty bags, as I point out in the prior point, hence this variable appears to not actually save us in any way.

Am I mistaken?

# Remark

Ultimately  `try_mutating_storage_state` is just a very bad construction, likely with also very wasteful database access patterns. I think we should just focus on fixing the literal bugs in question, and then the whole implementation can be fixed/decomposed in steps later, as is planned.

L3pereira:
Since we were iterating the data  in case of bag removing, 3 times
https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L4527-L4529
https://github.com/Joystream/joystream/blob/92d70638d8b4de19a47fd7c5b291f0d132d5e291/runtime-modules/storage/src/lib.rs#L957-L959
https://github.com/Joystream/joystream/blob/b46d77564bdbffba04a2f896de0896ce83731e3f/runtime-modules/storage/src/lib.rs#L4413

I calculated the total size, number and prizes of objects to delete/update in the `construct_objects_to_upload` and `construct_objects_to_remove`. This avoids folding the collection in the voucher update.

I used try_fold to shortcut the iteration in case one item fails.
`bag_deletion_prize` is injected in the fold initial value, in case of creation or bag deletion.

```rust
    fn construct_objects_to_upload(op: &BagOperation<T>) -> DataObjAndDeletionPrizeAndObjSize<T> {
        match &op.params {
            BagOperationParams::<T>::Delete => Ok((Default::default(), Zero::zero(), Zero::zero())),
            BagOperationParams::<T>::Create(bag_deletion_prize, list, _, _) => {
                Self::construct_objects_from_list(list, *bag_deletion_prize)
            }
            BagOperationParams::<T>::Update(list, _) => {
                Self::construct_objects_from_list(list, Zero::zero())
            }
        }
    }

    fn construct_objects_from_list(
        list: &[DataObjectCreationParameters],
        bag_deletion_prize: BalanceOf<T>,
    ) -> DataObjAndDeletionPrizeAndObjSize<T> {
        let deletion_prize = Self::data_object_deletion_prize_value();
        list.iter()
            .map(|param| {
                Self::upload_data_objects_checks(param).and({
                    Ok(DataObject {
                        accepted: false,
                        deletion_prize,
                        size: param.size,
                        ipfs_content_id: param.ipfs_content_id.clone(),
                    })
                })
            })
            .try_fold(
                Ok((Vec::new(), bag_deletion_prize, 0)),
                |acc: DataObjAndDeletionPrizeAndObjSize<T>, obj| {
                    obj.map(|obj| {
                        acc.map(|(mut acc_obj, acc_deletion_prize_request, acc_size)| {
                            let acc_deletion_prize_request =
                                acc_deletion_prize_request.saturating_add(obj.deletion_prize);

                            let acc_size = acc_size.saturating_add(obj.size);

                            acc_obj.push(obj);

                            (acc_obj, acc_deletion_prize_request, acc_size)
                        })
                    })
                },
            )
            .and_then(|x| x) 
         //Try fold is operating on this structure
         //[Result<Item1>, Result<Item1>, Result<Item3>, ...]
         //and returns
         //Result<Result[Item1 + Item2 + Item3 + ...]>
         //Since and_then changes the type
         //(A[u] -> A[k]) we get rid of the outer result -> Result[Item1 + Item2 + Item3 + ...]
    }
```
```rust
    fn construct_objects_to_remove(
        op: &BagOperation<T>,
        bag_deletion_prize: BalanceOf<T>,
    ) -> DataObjAndDeletionPrizeAndObjSize<T> {
        match &op.params {
            BagOperationParams::<T>::Delete => Ok(DataObjectsById::<T>::iter_prefix(&op.bag_id)
                .map(|(_, obj)| obj)
                .fold(
                    (Vec::new(), bag_deletion_prize, 0),
                    |(mut acc_obj, deletion_prize_refund, acc_size), obj| {
                        let deletion_prize_refund =
                            deletion_prize_refund.saturating_add(obj.deletion_prize);

                        let acc_size = acc_size.saturating_add(obj.size);

                        acc_obj.push(obj);

                        (acc_obj, deletion_prize_refund, acc_size)
                    },
                )),
            BagOperationParams::<T>::Create(..) => Ok((Default::default(), Zero::zero(), 0)),
            BagOperationParams::<T>::Update(_, list) => list
                .iter()
                .try_fold(
                    Ok((Vec::new(), Zero::zero(), 0)),
                    |acc: DataObjAndDeletionPrizeAndObjSize<T>, id| {
                        Self::ensure_data_object_exists(&op.bag_id, id).map(|obj| {
                            acc.map(|(mut acc_obj, acc_deletion_prize_refund, acc_size)| {
                                let acc_deletion_prize_refund =
                                    acc_deletion_prize_refund.saturating_add(obj.deletion_prize);

                                let acc_size = acc_size.saturating_add(obj.size);

                                acc_obj.push(obj);

                                (acc_obj, acc_deletion_prize_refund, acc_size)
                            })
                        })
                    },
                )
                .and_then(|x| x),
        }
    }
```
And `ensure_sufficient_balance` got simpler by doing only the balance check
```rust
    fn ensure_sufficient_balance(
        account_id: &T::AccountId,
        deletion_prize_request: T::Balance,
        storage_fee: T::Balance,
    ) -> DispatchResult {
        ensure!(
            Balances::<T>::usable_balance(account_id)
                >= deletion_prize_request.saturating_add(storage_fee),
            Error::<T>::InsufficientBalance
        );

        Ok(())
    }
```

runtime-modules/storage/src/tests/mod.rs
```diff
@@ -6145,3 +6148,111 @@ fn funds_needed_for_upload_succeeds_overflow_sat() {
         assert_eq!(funds_needed_saturated, u64::MAX);
     });
 }
+
+#[test]
+fn new_obj_deletion_prize_have_no_effect_on_older_data_obj() {
```
dobertRowneySr: I feel that this test should be reworked:
0. first off use something like `delete_data_object_succeeds_with_original_obj_deletion_prize` as test name
1. Setup your scenario with one dynamic bag containing some > 1 (in order to make sure that all deletion prize loops used in calculation works as expected) objects @ a given deletion prize (Arrange part)
2. change deletion prize (Arrange part)
3. call `delete_data_object` (Act part)
4. assert variation in balance is proportional to the original deletion prize (Assert part)

runtime-modules/storage/src/lib.rs
```diff
@@ -4459,97 +4444,110 @@ impl<T: Trait> Module<T> {
...
+
+                            (acc_obj, acc_deletion_prize_request, acc_size)
+                        })
+                    })
+                },
+            )
+            .and_then(|x| x)
     }
 
     fn construct_objects_to_remove(
```
dobertRowneySr, line 4506: This function will probably change in a future refactoring since it has became quite complex.
For now please just add a doc line like describing the main units of work done

runtime-modules/storage/src/lib.rs
```diff
@@ -4459,97 +4444,110 @@ impl<T: Trait> Module<T> {
...
+                Self::construct_objects_from_list(list, *bag_deletion_prize)
+            }
+            BagOperationParams::<T>::Update(list, _) => {
+                Self::construct_objects_from_list(list, Zero::zero())
             }
-            BagOperationParams::<T>::Update(list, _) => Self::construct_objects_from_list(list),
         }
     }
 
     fn construct_objects_from_list(
```
dobertRowneySr, line 4464: This function will probably change in a future refactoring since it has became quite complex.
For now please just add a doc line like describing the main units of work done

runtime-modules/storage/src/lib.rs
```diff
@@ -953,13 +954,13 @@ struct VoucherUpdate {
 }
 
 impl VoucherUpdate {
-    fn add_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
-        list.iter().fold(self, |acc, obj| acc.add_object(obj.size))
-    }
+    // fn add_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
```
dobertRowneySr: Remove dead code if you think it's obsolete

runtime-modules/storage/src/lib.rs
```diff
@@ -953,13 +954,13 @@ struct VoucherUpdate {
...
+    //     list.iter().fold(self, |acc, obj| acc.add_object(obj.size))
+    // }
 
-    fn sub_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
-        list.iter().fold(self, |acc, obj| acc.sub_object(obj.size))
-    }
+    // fn sub_objects_list<Balance>(self, list: &[DataObject<Balance>]) -> Self {
```
dobertRowneySr: Remove dead code if you think it's obsolete

runtime-modules/storage/src/lib.rs
```diff
@@ -988,12 +989,12 @@ impl VoucherUpdate {
         }
     }
 
-    fn sub_object(self, size: u64) -> Self {
-        Self {
-            objects_number: self.objects_number.saturating_sub(1),
-            objects_total_size: self.objects_total_size.saturating_sub(size),
-        }
-    }
+    // fn sub_object(self, size: u64) -> Self {
```
dobertRowneySr: Remove dead code if you think it's obsolete

dobertRowneySr (changes requested):
Logic is ok, I just left some style-related comments.
I know it is not part of this PR but why did we remove the check on the expected object deletion prize fee.

runtime-modules/storage/src/tests/mod.rs
L3pereira: This test has been refactored, it has only one bag now with 5 objects.
deletion prize changes,
3 objects  are removed,
in the end we have only one assert.

runtime-modules/storage/src/lib.rs
L3pereira: Now these functions have a comment explaining the workflow better.
I added a function to simplify, 
```rust
    //Sums the accumulated prize and size to the new object in the iteration.
    fn calculate_acc_size_and_acc_prize(
        mut acc_obj: Vec<DataObject<BalanceOf<T>>>,
        acc_deletion_prize: BalanceOf<T>,
        acc_size: u64,
        obj: DataObject<BalanceOf<T>>,
    ) -> (Vec<DataObject<BalanceOf<T>>>, BalanceOf<T>, u64) {
        let acc_deletion_prize: T::Balance = acc_deletion_prize.saturating_add(obj.deletion_prize);

        let acc_size: u64 = acc_size.saturating_add(obj.size);

        acc_obj.push(obj);

        (acc_obj, acc_deletion_prize, acc_size)
    }
```
```rust
.try_fold(
        //As an initial value we set the fold with an empty vec, the bag deletion prize,
        //(it's zero in case of an update), and 0 size.
        Ok((Vec::new(), bag_deletion_prize, 0)),
        |acc: DataObjAndDeletionPrizeAndObjSize<T>, obj| {
               obj.map(|obj| {
                        acc.map(|(acc_obj, acc_deletion_prize_request, acc_size)| {
                            Self::calculate_acc_size_and_acc_prize( <-----------------------------------------------
                                acc_obj,
                                acc_deletion_prize_request,
                                acc_size,
                                obj,
                            )
                        })
                    })
                },
            )
       //Removes outer result
       //(Result<Result[Item1 + Item2 + Item3 + ...]>) -> Result[Item1 + Item2 + Item3 + ...]
       .and_then(|x| x)
```

L3pereira:
> I know it is not part of this PR but why did we remove the check on the expected object deletion prize fee.

I didn't find this issue, If you can create a test reproducing it, please, let me know

dobertRowneySr (commented):
Good job, LGTM

The check on `expected_data_object_deletion_prize` is here: https://github.com/Joystream/joystream/blob/34720962075ed5f6c01882a1348faf86ab445572/runtime-modules/storage/src/lib.rs#L3204
My bad

dobertRowneySr (approved):

Issue #3743, opened 2022-05-12
Project tokens are owned by members not raw accounts


bedeho:
# Background
https://github.com/Joystream/joystream/issues/2362#issuecomment-1088722206

# Proposal
Do that, but probably best to do _after_ both sales and revenue splits have been merged, but before AMM work is started.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202266620134532) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/3755

PR #3755, opened 2022-05-16
Creator tokens: Change account identifier from AccountId to MemberId

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3743


dobertRowneySr (approved):
LGTM

Issue #3637, opened 2022-04-20
Project Token Revenue Split Tweaks


bedeho:
As a result of product meeting we identified some needs

* All revenue splits must confront a universal lower bound on the time to start and duration, so as to give everyone fair access. These lower bounds can be stored as parameters in storage which will later be updatable with proposals, but not need to worry about those proposals now.
* ~It should be possible for the issuer to start a revenue split and stake their own $CRT holdings in one extrinsic.~
* It should be possible to stake for an ongoing revenue split by simultaneously unstaking for a prior one in one extrinsic.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202156232989995) by [Unito](https://www.unito.io)

dobertRowneySr:
>   All revenue splits must confront a universal lower bound on the time to start and duration, so as to give everyone fair access. These lower bounds can be stored as parameters in storage which will later be updatable with proposals, but not need to worry about those proposals now.

As of current implementation issuer can specify an optional `starting_block` and a duration, your request would be equiv. to:
- `starting_block - current_block >= LOWER_BOUND_ON_START`
- `duration >= LOWER_BOUND_ON_DURATION` (this is already implemented)

bedeho:
Yep ;)

But make these bounds storage variables, not constants, we are going to move away from all constants.

dobertRowneySr:
Ok, and when the optional starting block is **NOT SPECIFIED** should the constraint above be respected anyways, namely:
`starting_block := current_block + LOWER_BOUND_ON_START` in that case

bedeho:
Yep ;)

dobertRowneySr:
Now however we have another problem:
According to [specs](https://github.com/Joystream/joystream/issues/2362):
> During this period, any token holder (not member, token holder!) can claim part of the revenue split by staking the $CRT balance in an account. 

The tweaks would require the issuer to stake **immediately** in the block the revenue split is issued. However the first tweak require user to wait at least `LOWER_BOUND_ON_START` before allowing staking.
Should I break this rule for the token issuer?

bedeho:
> The tweaks would require the issuer to stake immediately in the block the revenue split is issued

Why? Why can't a $CRT staker stake any time during the split being live? so during `[starting_block, starting_block + duration]` ?

dobertRowneySr:
[...]

PR #3756, opened 2022-05-17
Creator token revenue split tweaks

dobertRowneySr:
Specs: #3637 

#### Changes
- `MinRevenueSplitDuration` is now a storage variable
- `MinRevenueSplitForewarning` storage variable added. Now whenever `issue_revenue_split` is called (either with or without optional starting block) at `current_block` it holds `revenue_split_starting_block >= current_block + MinRevenueSplitForewarning`. 
- User participating in a previous split is able to unstake automatically by using `participate_in_split`


runtime-modules/project-token/src/lib.rs
```diff
@@ -103,6 +100,11 @@ decl_storage! {
 
         /// Minimum duration of a token sale
         pub MinSaleDuration get(fn min_sale_duration) config(): T::BlockNumber;
+
+        /// Minimum revenue split duration constraint
+        pub MinRevenueSplitDuration get(fn min_revenue_split_duration) config(): T::BlockNumber;
+        /// Minimum revenue split forewarning constraint
+        pub MinRevenueSplitForewarning get(fn min_revenue_split_forewarning) config(): T::BlockNumber
```
Lezek123: I think `TimeToStart` would be much clearer of a name than `Forewarning`

Lezek123 (changes requested):
- Please add `issue_split_ok_with_custom_start_block` test case to test a scenario where the user provides `Some(start_block)`
- Not strictly related to this PR, but I think `issue_revenue_split` w/ `allocation == 0` should result in `DispatchError` (consistently w/ `participate_in_split` w/ `amount == 0`)
- Merge conflicts

runtime-modules/project-token/src/lib.rs
dobertRowneySr: fixed in 94d7272167f3cb56eebb85640dc8f66868a3ca28

Lezek123 (approved):
LGTM

Issue #3737, opened 2022-05-10
Revenue split - pausing feature


dobertRowneySr:
As discussed https://github.com/Joystream/joystream/issues/3687 , pausing any possible minting / burning activity during the split period is a feasible solution to avoid token supply value fluctuations

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202253783036680) by [Unito](https://www.unito.io)

PR #3757, opened 2022-05-17
Creator token revenue split pausing

dobertRowneySr:
Addresses #3737 

##### Changes
Extrinsics that can corrupt supply:
- `claim_patronage_credit`

It's now not possible to call one of the above extrinsics when revenue split state for a token is `Active`. 
Technically it won't cause a problem to modify supply value even when a revenue split is active but the `current_block` value is not in `[split_starting_block, split_starting_block + duration)`. But I think the above decision is simpler

### DIFF
--- a/runtime-modules/project-token/src/tests/patronage.rs
+++ b/runtime-modules/project-token/src/tests/patronage.rs
@@ -2,7 +2,7 @@
 use frame_support::{assert_noop, assert_ok};
 use sp_runtime::{Permill, Perquintill};
 
-use crate::tests::fixtures::default_upload_context;
+use crate::tests::fixtures::{default_upload_context, IssueRevenueSplitFixture};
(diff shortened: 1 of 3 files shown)


runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -274,6 +274,35 @@ fn decreasing_patronage_rate_fails_invalid_token() {
     })
 }
 
+#[test]
+fn claim_patronage_fails_with_ongoing_revenue_split() {
```
Lezek123:
```suggestion
fn claim_patronage_fails_with_active_revenue_split() {
```

Lezek123 (approved):
LGTM, just suggested a small naming change, since as I understand, `ongoing` means `active` + `current_block` in `[start_block, end_block)`

PR #3764, opened 2022-05-18
Creator token sales: Fix `update_upcoming_sale`

Lezek123:
This PR fixes `update_upcoming_sale` which was missing the validation of `new_duration` parameter.

### DIFF
--- a/runtime-modules/project-token/src/lib.rs
+++ b/runtime-modules/project-token/src/lib.rs
@@ -920,9 +920,13 @@ impl<T: Trait>
     /// Preconditions:
     /// - token by `token_id` exists
     /// - token offering is in UpcomingSale state
+    /// - If `new_duration.is_some()`:
+    ///   - `new_duration` >= `min_sale_duration` && `new_duration` > 0
+    /// - If `new_start_block.is_some()`:
+    ///   - `new_start_block` >= `current_block`
     ///
     /// Postconditions:
-    /// - token's sale `duration` and `start_block` is updated according to provided parameters
+    /// - token's sale `duration` and `start_block` updated according to provided parameters
     fn update_upcoming_sale(
         token_id: T::TokenId,
         new_start_block: Option<T::BlockNumber>,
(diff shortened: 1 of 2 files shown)


runtime-modules/project-token/src/lib.rs
```diff
@@ -920,25 +920,44 @@ impl<T: Trait>
...
         token_id: T::TokenId,
         new_start_block: Option<T::BlockNumber>,
         new_duration: Option<T::BlockNumber>,
     ) -> DispatchResult {
         let token_data = Self::ensure_token_exists(token_id)?;
         let sale = OfferingStateOf::<T>::ensure_upcoming_sale_of::<T>(&token_data)?;
+
+        // Validate sale duration
+        if let Some(duration) = new_duration {
+            ensure!(!duration.is_zero(), Error::<T>::SaleDurationIsZero);
```
dobertRowneySr, line 940: Isn't  `duration >= min_sale_duration` enough already? Since quite probably we will have a non zero min sale duration

dobertRowneySr (changes requested):
Looks good, I just left a question

runtime-modules/project-token/src/lib.rs
Lezek123: I wasn't sure what state will we start with, so I added this additional guard just in case (so that nothing unexpected happens if at any point we have `min_sale_duration == 0`)

dobertRowneySr: Ok, what do you think about this:
- the check goes away together with the error
- use a `debug_assert(duration != 0)` in order to signal that allowing `duration == 0` is symptom of a broken runtime, since nobody is able to participate in
?

Lezek123: I'm not sure if `debug_assert` is needed, but the alternative I see is:
- Set `min_sale_duration` to non-zero `on_runtime_upgrade` (similarly to https://github.com/Joystream/joystream/blob/master/runtime/src/runtime_api.rs#L86)
- Make sure `min_sale_duration` is also validated to be `> 0` when a proposal is made to change it

dobertRowneySr: Right proposals, I forgot about those. Let's leave your check on == 0 then for the moment

dobertRowneySr (approved):
LGTM

Issue #3734, opened 2022-05-10
Integrate project tokens pallet into content pallet


bedeho:
Do it on ephesus, but disable?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202252286387135) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/3771 (except the disabling part, which I suggest we do separately before the release)

PR #3771, opened 2022-05-19
Project token->Content pallet integration

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3734

A few notes about how the integration was done:

### Events:
The only event emitted by `content` pallet w.r.t. creator tokens is `CreatorTokenIssued`, since this is the only event needed in order for the query node to be able to connect a creator token with a given channel. All other events are emitted from `project_token` pallet (some missing events were added: `TokenSaleInitialized`, `UpcomingTokenSaleUpdated`, `TransferPolicyChangedToPermissionless`)

### Deriving member context:

`get_worker_member_id` method was added to `WorkingGroupAuthenticator`, allowing the `content` pallet to derive the `member_id` based on `ContentActor` context through a new function called `get_member_id_of_actor`:

| actor         | `get_member_id_of_actor`         |
|---------------|--------------------------------------------|
| `Member` | `member_id` of `Member`           |
| `Curator`  | `worker.member_id` of `Curator` |
| `Lead`      | `worker.member_id` of `Lead`     |

This function is used to determine:
- CRT source account for creator token sale (`init_creator_token_sale`)
- CRT patronage destination (`claim_creator_token_patronage_credit`)
- CRT source account for issuer transfer (`creator_token_issuer_transfer`)

### Token sale JOY earnings

A new `earnings_destination` field was added to `TokenSale`, which is `Option<AccountId>` and works as follows:
- If `earnings_destination.is_some()` - JOY earnings from the sale (minus platform fee) are sent to `earnings_destination` account
- If `earnings_destination.is_none()` - all JOY earnings from the sale are burned

The `content` pallet fills this field based on the channel owner:
- If `ChannelOwner` is `Member`: `earnings_destination` is `Some(member_controller_account)`
- If `ChannelOwner` is `CuratorGroup`: `earnings_destination` is `None`

### Curator channels

Following https://github.com/Joystream/joystream/issues/3784, there were a few other curator-channel-specific features implemented:
- `auto_finalize` field was added to `TokenSale`, which controls whether the sale is automatically removed when `quantity_left == 0`. Content pallet sets this field to `true` if `ChannelOwner` is `Member` and `false` if `ChannelOwner` is `CuratorGroup`.
- `recover_unsold_tokens` extrinsic was replaced with `finalize_token_sale` **method**, which is callable only by channel owner through `content` pallet and returns the total amount of JOY collected during the sale (`funds_collected`). Additionally, if `ChannelOwner` is `CuratorGroup`, the content pallet then adds `funds_collected` to the Council Budget
- `claim_creator_token_patronage_credit` was blocked for curator-owned channels

### Bloat bonds

Some code in `project_token` pallet was changed so that the `bloat_bond` is always paid by the sender of `content` extrinsic (in case the interaction with `project_token` pallet happens through `content` pallet). This means:
- Member controller account in case `ChannelOwner` is `Member`
- Curator's **role account** in case `ChannelOwner` is `Curator`

### Tests:

I limited the scope of the tests to errors / events / state changes coming directly from `content` pallet, as we have a separate set of tests for `project_token` pallet. This can be improved in the future, but I suggest it to be done outside of the scope of this PR, as it's already quite large.

### Types and `ProjectToken` trait:

`pallet_token`'s `trait.rs` and `types.rs` were exposed and some types were made public (`pub(crate) -> pub`), I also slightly modified the `ProjectToken` trait to make it easier for the `content` module to use it.


dobertRowneySr (approved):
LGTM, logic and test coverage look :ok_hand: Also I don't have any particular remarks about cosmetics.

PR #3781, opened 2022-05-23
Qn comments section tweaks

zeeshanakram3:
- Removes the meta protocol way of enabling/disabling the comments section
- Adds support for Enabling/Disabling comments section using `VideoMetadata`


kdembler (approved):
Works well but we should fix integration tests

Issue #3640, opened 2022-04-21
Allow burning all funds


bedeho:
Allow burning funds even if they are subject to vesting or staking. I don't think this should cause any trouble, but double check!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202159598893522) by [Unito](https://www.unito.io)

Lezek123:
No burning extrinsic is currently implemented, needs to be added

Lezek123:
Done in https://github.com/Joystream/joystream/pull/3782

PR #3782, opened 2022-05-23
Creator tokens: Burning

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3640

Burning priority:
- Unvested and staked tokens are burned first
- Transferrable tokens are burned only if no unvested / staked tokens remain

This is achieved by introducing `burned_amount` property to `vesting_schedule`. `vesting_schedule.locks(b)` is equal to `max(0, unvested_amount (according to original schedule) - burned_amount)`. Multiple test cases have been added to ilustrate this and make sure everything works as expected.

When burning funds, all vesting schedules that either already were or just became stale (ended up with `locks(current_block) == 0`) are removed.

It is possible to immediately `dust_account` once all funds are burned, even if some of those funds were still locked in a vesting schedule before (unvested). If some tokens were staked, they need to be unstaked first however (even though the staked amount is now `0`), this is to prevent `burn` extrinsic from directly affecting the `staking_status` of an account (since this is a more risky ingerention than just changing the staked `amount` to `0`)

### DIFF
--- a/runtime-modules/project-token/src/lib.rs
+++ b/runtime-modules/project-token/src/lib.rs
@@ -203,6 +203,74 @@ decl_module! {
             Ok(())
         }
 
+        /// Burn tokens from specified account
+        ///
+        /// Preconditions:
+        /// - `amount` is > 0
+        /// - origin signer is a controller account of `member_id` member
+        /// - token by `token_id` exists
+        /// - an account exists for `token_id` x `member_id`
+        /// - account's tokens amount is >= `amount`
+        /// - token supply can be modified (there is no active revenue split)
+        ///
+        /// Postconditions:
+        /// - starting with `unprocessed` beeing equal to `amount`, account's vesting schedules
+        ///   are iterated over and updated:
+        ///   - if a vesting scheduled has `vs.locked(current_block)` <= `unprocessed`:
+        ///     - `unprocessed` is set to `unprocessed - vs.locked(current_block)`
+        ///     - the vesting schedule is removed
+        ///   - if a vesting schedule has `vs.locked(current_block)` > `unprocessed`:
+        ///     - `unprocessed` is set to `0`
+        ///     - vesting schedule's `burned_amount` is updated to `vs.burned_amount + unprocessed`
+        /// - if the account has any `split_staking_status`, the `split_staking_status.amount`
+        ///   is reduced by `min(amount, split_staking_status.amount)`
+        /// - `account.amount` is reduced by `amount`
+        /// - token supply is reduced by `amount`
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn burn(origin, token_id: T::TokenId, member_id: T::MemberId, amount: TokenBalanceOf<T>) -> DispatchResult {
+            // Ensure burn amount is non-zero
+            ensure!(
+                !amount.is_zero(),
+                Error::<T>::BurnAmountIsZero
+            );
+
+            // Ensure sender is member's controller account
+            T::MemberOriginValidator::ensure_member_controller_account_origin(
+                origin,
+                member_id
+            )?;
+
+            // Ensure token exists
+            let token_info = Self::ensure_token_exists(token_id)?;
+
+            // Ensure token account data exists by `token_id` x `member_id`
+            let account_info = Self::ensure_account_data_exists(token_id, &member_id)?;
+
+            // Ensure burn amount doesn't exceed account's tokens amount
+            ensure!(
+                account_info.amount >= amount,
+                Error::<T>::BurnAmountGreaterThanAccountTokensAmount
+            );
+
+            // Ensure token supply can be modified
+            token_info.ensure_can_modify_supply::<T>()?;
+
+            // == MUTATION SAFE ==
+            let now = Self::current_block();
+
+            // Burn tokens from the account
+            AccountInfoByTokenAndMember::<T>::mutate(token_id, member_id, |account| {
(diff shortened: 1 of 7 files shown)


runtime-modules/project-token/src/lib.rs
```diff
@@ -203,6 +203,74 @@ decl_module! {
...
+        ///
+        /// Preconditions:
+        /// - `amount` is > 0
+        /// - origin signer is a controller account of `member_id` member
+        /// - token by `token_id` exists
+        /// - an account exists for `token_id` x `member_id`
+        /// - account's tokens amount is >= `amount`
+        /// - token supply can be modified (there is no active revenue split)
+        ///
+        /// Postconditions:
```
dobertRowneySr, line 216: I would use a simpler description like:
- if a schedule vesting amount is completely burned, then vesting schedule is removed
- else vesting schedule `burned_amount += amount`

runtime-modules/project-token/src/tests/canonical.rs
```diff
@@ -1062,3 +1062,406 @@ fn issue_token_ok_with_owner_accounts_data_added() {
...
+        .with_token(token_id, token_data)
+        .with_account(member_id, account_data)
+        .build();
+
+    build_test_externalities(config).execute_with(|| {
+        // Burn 300 tokens and re-fetch account_data
+        assert_ok!(Token::burn(origin!(account), token_id, member_id, 300));
+        let account_data = Token::ensure_account_data_exists(token_id, &member_id).unwrap();
+        // Expect transferrable amount at linear_vesting_start_block still equal to 300 (`cliff_amount`)
+        assert_eq!(
```
dobertRowneySr, line 1425: I felt (by going through this) that the test is too complex, since you are basically testing the following units of behavior:
- whether  vesting schedule accounting logic &`transferrable` behaves correctly as blocks pass (which should already be covered in the `sales`- tests) -> first 5 assertions
- you are checking that `transferrable` amount is left untouched after the burn since only vesting amount has been modified (and you have already tested this in `burn_ok_with_vesting_and_staked_tokens_burned_first`)
- then you are testing that `transferrable` amount accounting is correct
- then you are testing that a schedule with a fully burned vesting amount is gone -> ok!

dobertRowneySr (changes requested):
great job!
I would add the following test cases:
- burn ok with just staking
- burn ok with no vesting and no staking
- burn ok with vesting period over amount fully burned (which you are already testing in the `...should_work_as_expected` test)
- burn ok with vesting period non over amount fully burned

runtime-modules/project-token/src/tests/canonical.rs
dobertRowneySr: WDYT?

Lezek123: > - whether vesting schedule accounting logic &transferrable behaves correctly as blocks pass (which should already be covered in the sales- tests) -> first 5 assertions

This is different from the vesting schedule tests in `sale.rs`, because it checks the behavior of a schedule that is "partially burned" (while the tests inside `sale.rs` only test the vesting schedule that has `burend_amount == 0`)

> you are checking that transferrable amount is left untouched after the burn since only vesting amount has been modified (and you have already tested this in burn_ok_with_vesting_and_staked_tokens_burned_first)

`burn_ok_with_vesting_and_staked_tokens_burned_first` is different, because in that case vesting_schedules are removed completely. Those tests check the behavior of a "partially burned" vesting schedule throughout different stages.

Lezek123:
> I would add the following test cases:

> - burn ok with just staking

Ok

> - burn ok with no vesting and no staking

This is already covered by basic cases:
- `burn_ok`
- `burn_ok_with_account_tokens_amount_decreased`
- `burn_ok_with_token_supply_decreased` 

> - burn ok with vesting period over amount fully burned (which you are already testing in the ...should_work_as_expected test)

I'm not sure what you mean by `with vesting period over amount fully burned`.
Vesting schedules are also fully burned (and removed) as part of `burn_ok_with_vesting_and_staked_tokens_burned_first`

> - burn ok with vesting period non over amount fully burned

Same as above, I don't understand what exactly should I test here, could you rephrase / elaborate?

dobertRowneySr:
> Same as above, I don't understand what exactly should I test here, could you rephrase / elaborate?

Yes, so I was referring to the edge case when the internal `burn` methods finds a ended vesting schedule.
But you are already checking cases when a vesting schedule is removed (with both `locks() != 0` and `locks() == 0`) during burning. So forget about the those remarks.

runtime-modules/project-token/src/tests/canonical.rs
```diff
@@ -1062,3 +1062,433 @@ fn issue_token_ok_with_owner_accounts_data_added() {
...
+        );
+        assert_eq!(
+            first_vesting_schedule.non_burned_amount(),
+            1000 - burn1_amount - burn2_amount
+        );
+    })
+}
+
+#[test]
+fn burn_ok_with_partially_burned_vesting_schedule_working_as_expected() {
```
dobertRowneySr:
```suggestion
fn burn_ok_with_partially_burned_vesting_schedule_amounts_working_as_expected() {
```

dobertRowneySr: Or something similar

dobertRowneySr (changes requested):
LGTM, just cosmetics left

dobertRowneySr (approved):

Issue #3559, opened 2022-04-05
Bug: Release vote stake not working (for very specific scenario)


mochet:
Two users have reported this issue so far.
This bug only seems to impact users who match a very specific set of criteria. Please see 1st comment on issue for examples showing this to be the case.
* Both users voted during `cycle id: 0`
* Both users revealed votes for unsuccessful candidates from `cycle id: 0`
* Both users attempted to release the stake during `cycle id: 1`
* Both users were trying to release stake for votes for candidates who were unsuccessful in `cycle id: 0` but were successfully elected in `cycle id: 1`
* Both user's votes still exist within `chain state` > `referendum.votes`
* Both user's vote accounts show `balances.locks` with only locks for `id: voting` as existing
* Both users attempted to use `extrinsic` > `referendum.releaseVoteStake` and got the error `referendum.UnstakingForbidden` (details say: `Unstaking has been forbidden for the user (at least for now`)
* Doing the extrinsic via Pioneer/Polkadot.js makes no difference, results in the same failure.
* This goes against what is described in the handbook.

Addresses/links:
* `5CUefaByXT5B2uxaeuAPiDrGDMVEHd8Ds52bz9XAi2xXp4y1`
	* voted for 2182 - ACTIVE CM in `cycle ID: 1`
	* Error: https://polkadot.js.org/apps/#/explorer/query/0x43f463c83e48bcdfac0ae900156e76f26052f0c517b5e7624d1f7fcfe34fd8b3
* `5D34bqj2H1vAgHaABts6Qn4nqPCKXKKfEthA2KPQtNzDaA5j`
	* voted for 644 - ACTIVE CM in `cycle ID: 1`
	* Error: https://polkadot.js.org/apps/#/explorer/query/0xed74cfa3276e63fed5c941d1d3491a729bafb86e9cd8ea07a79f5413e6319f4c
* `5GBpm9ghjsgRPjvbzjjnb3pCJarMUWLsXHnjNynKd71mRQb8`
	* voted for 306 - ACTIVE CM in `cycle ID: 1`
	* Error: I don't have a link for this one

mochet:
Further investigation:
* `referendum.votes` contains `50 votes`
* Query Node shows `28 votes` were released successfully

This shows an example of an example of a successful release similar to the problem scenarios. They voted in the first election for someone who was successful in the second election.
```
            "id": "00000B",
[...]

PR #3783, opened 2022-05-24
Fix Vote Stake Unlocking

mnaamani:
Fix for https://github.com/Joystream/joystream/issues/3559

- Fix implementation of `council::can_unlock_vote_stake()` to match the rules described in the handbook
- Adjust run_full_council_cycle to support interrupt to allow testing vote unlocks during council idle stage

### DIFF
--- a/runtime-modules/council/src/lib.rs
+++ b/runtime-modules/council/src/lib.rs
@@ -1184,7 +1184,7 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
     fn can_unlock_vote_stake(vote: &CastVoteOf<T>) -> Result<(), Error<T>> {
         let current_voting_cycle_id = AnnouncementPeriodNr::get();
 
-        // If the vote is for an election prior to the last concluded..
+        // If the vote was cast before the latest Announcing stage...
         if vote.cycle_id != current_voting_cycle_id {
             // ..it is always recoverable.
             return Ok(());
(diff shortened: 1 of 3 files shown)


runtime-modules/council/src/lib.rs
```diff
@@ -1184,41 +1184,32 @@ impl<T: Trait> ReferendumConnection<T> for Module<T> {
     fn can_unlock_vote_stake(vote: &CastVoteOf<T>) -> Result<(), Error<T>> {
         let current_voting_cycle_id = AnnouncementPeriodNr::get();
 
-        // allow release for very old votes
-        if current_voting_cycle_id > vote.cycle_id + 1 {
+        // If the vote is for an election prior to the last concluded..
```
Lezek123:
```suggestion
        // If the vote was cast before the latest Announcing stage...
```
The handbook should also be updated accordingly

Lezek123 (changes requested):
I think it would be a good idea to add tests for the following scenarios:
- `run_council_cycle_with_interrupt` at `CouncilCycleInterrupt::AfterVoting` - expect no voter is able to unstake
- `run_council_cycle_with_interrupt` at `CouncilCycleInterrupt::AfterRevealing` - expect no voter is able to unstake
- a scenario similar to `council_vote_for_winner_stakes_longer`, but with second cycle interrupted at `CouncilCycleInterrupt::BeforeCandidatesAnnounce`

bedeho:
I very much agree with the idea of adding explicit tests here.

mnaamani:
> I very much agree with the idea of adding explicit tests here.

Agreed, its the best way to describe the rules

Lezek123 (approved):
LGTM

Issue #3612, opened 2022-04-15
Enhance NFT minting limits


bedeho:
# Background
https://github.com/Joystream/joystream/pull/3609

# Proposal
* Update to be mergable into `audit3`, not `ephesus`.
* Allow unlimited mode for limits, meaning there is no active constraint.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202133808413661) by [Unito](https://www.unito.io)

dobertRowneySr:
#### Background
The record containing all the current nft limits `(global, channel) x (weekly_limit, daily_limit)` is stored in the `content` pallet on chain state.
this is used by the[ `check_nft_limit` ](https://github.com/Joystream/joystream/blob/396df125e0290a2a693e0f398db6d6ac601846fd/runtime-modules/content/src/lib.rs#L2619)helper in the following extrinsics:
- [`create_video`](https://github.com/Joystream/joystream/blob/396df125e0290a2a693e0f398db6d6ac601846fd/runtime-modules/content/src/lib.rs#L848)
- [`update_video`](https://github.com/Joystream/joystream/blob/396df125e0290a2a693e0f398db6d6ac601846fd/runtime-modules/content/src/lib.rs#L939)
- [`issue_nft`](https://github.com/Joystream/joystream/blob/396df125e0290a2a693e0f398db6d6ac601846fd/runtime-modules/content/src/lib.rs#L1234)

Effectively controlling the nft issuance rate.

#### Question
1. Does this feature aims to drop `check_nft_limit`? (effectively removing any issuance rate constraints)
2. What is the purpose of the NFT limit related code now? Should it be also removed?

bedeho:
Not removal, it means that it should be possible to both have unconstrained and constrained minting as two separate modes in the system. Meaning a lead can change this constraint in-flight.

bedeho:
Done in https://github.com/Joystream/joystream/pull/3809

PR #3809, opened 2022-05-30
Carthage enhance nft limits

dobertRowneySr:
addresses #3612

#### Changes
* [x] added `NftLimitsEnabled` variable to storage, used in the `check_nft_limits` helper to enforce nft limits constraints, the chainspec value is `true`
* [x] added extrinsic `toggle_nft_limits` 
* [x] added unit tests for above extrinsic behavior
* [x] added unit test for verifying nft limits behavior is enabled/disabled
* [x] added chainspec configuration for the above storage variable
* [x] only root origin can change such value
* [x] rebased on `carthage-substratev3`
* [x] generate types via the new mechanism

#### ToDo (Not part of this PR)
- [ ] introduce proposal to allow the council to toggle the value

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476302162799) by [Unito](https://www.unito.io)


Lezek123 (commented):
I'm not 100% sure how the permissions should work w.r.t. setting and updating the nft issuance limits, but currently they seem quite incosistent, ie.:
| action                                         | allowed actors                                                                               |
|------------------------------------------------|----------------------------------------------------------------------------------------------|
| update channel's NFT limit                     | Content Lead or any Curator w/ `ContentModerationAction::UpdateChannelNftLimits` permissions |
| update global NFT limit                        | Council                                                                                      |
| enable/disable all limits (global & channel)   | Content Lead                                                                                 |

The main point is - before this PR only the **council** could manage the **global NFT limits**, but now we're giving access to the **lead** to disable/enable them any time, effectively making them able to sidestep any limit enforced by the **coucil**.

Is that really what we want?
Pinging @bedeho to confirm

bedeho:
Very well done spotting this inconsistency @Lezek123 , lets just make it `Root` origin, and in the future there can be a proposal added for this.

node/src/chain_spec/content_config.rs
```diff
@@ -28,6 +28,7 @@ pub fn production_config() -> ContentConfig {
         platform_fee_percentage: Perbill::from_percent(1),
         auction_starts_at_max_delta: DAYS * 30,
         max_auction_whitelist_length: 100,
+        nft_limits_enabled: true,
```
Lezek123, line 30: Just an observation: We can make use this feature instead of commenting out the checks in https://github.com/Joystream/joystream/pull/3914 once this PR is merged

runtime-modules/content/src/lib.rs
```diff
@@ -235,7 +235,11 @@ decl_storage! {
 
         /// Global weekly NFT limit.
         pub GlobalWeeklyNftLimit get(fn global_weekly_nft_limit):
-            LimitPerPeriod<T::BlockNumber>;
+        LimitPerPeriod<T::BlockNumber>;
```
Lezek123, line 236: The formatting is now inconsistent:
```suggestion
            LimitPerPeriod<T::BlockNumber>;
```

runtime-modules/content/src/lib.rs
```diff
@@ -235,7 +235,11 @@ decl_storage! {
...
+
+        /// NFT limits enabled or not
+        /// Can be updated in flight by Lead
```
```suggestion
        /// Can be updated by the Council
```

runtime-modules/content/src/lib.rs
```diff
@@ -2304,6 +2308,23 @@ decl_module! {
             Self::deposit_event(RawEvent::NftBought(video_id, participant_id));
         }
 
+        /// Only Lead can toggle nft issuance limits constraints
```
```suggestion
        /// Only Council can toggle nft issuance limits constraints
```

runtime-modules/content/src/lib.rs
```diff
@@ -2304,6 +2308,23 @@ decl_module! {
...
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn toggle_nft_limits(
```
Lezek123, line 2359: This could result in unexpected behavior if there are multiple proposals created at a similar time to toggle the value. I think it should be explicit whether the proposal wishes to set the value to `true` or `false`, so the function signature should be `set_nft_limits_enabled(origin, enabled: bool)`. You also don't need to read the old value in that case, so it's more optimal.

runtime-modules/content/src/lib.rs
```diff
@@ -2304,6 +2308,23 @@ decl_module! {
...
+            let _ = ensure_root(origin)?;
+
+            //
+            // == MUTATION SAFE ==
+            //
+
+            let old_v = NftLimitsEnabled::get();
+            NftLimitsEnabled::put(!old_v);
+
+            Self::deposit_event(RawEvent::LeadToggledNftLimitsEnabled(!old_v));
```
Lezek123: Obviously the event name should be changed

runtime-modules/content/src/tests/nft/update_nft_limit.rs
```diff
@@ -165,3 +166,30 @@ fn default_channel_nft_limits_set_successfully() {
         );
     })
 }
+
+#[test]
+fn toggle_nft_limits_ok_with_event_deposited_and_status_changed() {
```
Lezek123, line 171: Probably worth adding at least one invalid origin test case (for example: lead)

Lezek123 (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: I added `by the Council` but only `root` can toggle the variable at the moment

runtime-modules/content/src/lib.rs
dobertRowneySr: I added the suggested comment but only `root` can toggle the variable at the moment

runtime-modules/content/src/tests/nft/update_nft_limit.rs
dobertRowneySr: added in e84b6b9d6f9d9021c4fa7ff6d58700270c65ad6e

runtime-modules/content/src/tests/nft/issue_nft.rs
```diff
@@ -369,14 +372,24 @@ fn issue_nft_channel_weekly_limit_works_as_expected() {
...
+            false,
         );
     })
 }
 
+#[test]
+fn issue_nft_ok_with_limits_not_enforced() {
+    with_default_mock_builder(|| {
+        // chainspec value = true, setting to false
+        test_helper_for_nft_limit_works_as_expected(NftLimitId::GlobalDaily, Ok(()), true);
```
Lezek123, line 386: Oh, I overlooked this before (although it's not a big issue), but it seems like you're only testing `GlobalDaily` limit here. There should probably a case for each type of limit (or alternatively you could just loop over them)

Lezek123 (approved):
Chain metadata conflict should be resolved + I left a small test improvement suggestion, but it's not really that important

dobertRowneySr:
I fixed the test missing test case. Merging...

PR #3821, opened 2022-06-01
introduce_channel_state_bloat_bond

### LINKED ISSUE #3511: Protect channel account dusting: introduce channel state bloat bond
# Background
So channel state bloat is indirectly deterred by the fact that there is a storage bag state bloat bond which is put up, and all channels have 1-1 relationship with lifetime of a dynamic bag. However, there is still the problem of the channel account risking getting dusted, as outlined here https://github.com/Joystream/joystream/issues/3497, for this reason we may as well introduce a state bloat bond, as it is functionally a solution to the state bloat bond. Moreover, dynamic bag bloat bonds are going away: https://github.com/Joystream/joystream/issues/3580

# Dependency
https://github.com/Joystream/joystream/issues/3580

# Proposal
Add state bloat bond, paid by creator at point of channel creation, and returned to whomever deletes channel. Fund with membership of caller, even if curator group or lead. 

An alternative solution which could have served the same purpose would be to just mint this upon channel creation, however, as we are removing bloat bonds for dynamic bags (https://github.com/Joystream/joystream/issues/3580), a genuine user funded bloat bond is now needed to solve both state protection and channel account dusting.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202127937423752) by [Unito](https://www.unito.io)

L3pereira:
* Closes #3511

When creating a channel, the sender gets slashed channel state bloat bond
When deleting a channel sender gets paid the channel state bloat bond amount
 Uses the `funds_needed_for_upload` implemented here #3693 to verify the funds needed before charging 
the channel state bloat bond

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202478219211283) by [Unito](https://www.unito.io)


runtime-modules/content/src/lib.rs
```diff
@@ -455,6 +474,8 @@ decl_module! {
             // == MUTATION SAFE ==
             //
 
+            let _ = Balances::<T>::slash(&sender, channel_state_bloat_bond);
+
```
bedeho, line 494: Do we not credit the bloat bond to some treasury account? I don't actually remember, what do we do for bounties or in the storage pallet?

L3pereira: Yes, we credited to StorageTreasury, in bounties we had a Council Budget, that's why I got confused about using content workgroup budget account instead of simply the sender's account. Since, for example, channel transfer uses the budget or the sender based on which type is the owner.
I implemented with an imbalance, and always used the sender's account, the reason I didn't use the storage treasury, was I thought we wanted to move this logic away from the storage pallet
In my opinion, we should use the following (I implemented this way at first, but then I step back since it was not clear to me if I should use it)
Lead -> Senders account
Memeber - > Senders account
Curator Group -> Content working group budget
We should also break the functions that transfer funds so they do not fail, as we did in bounties

runtime-modules/content/src/tests/channels.rs
```diff
@@ -75,11 +82,24 @@ fn successful_channel_creation_with_storage_upload_and_member_context() {
     with_default_mock_builder(|| {
         run_to_block(1);
         create_initial_storage_buckets_helper();
-        increase_account_balance_helper(DEFAULT_MEMBER_ACCOUNT_ID, INITIAL_BALANCE);
+        //1 channel SBB + 1 data obj SBB * 10 objs = 11 SBB
+        increase_account_balance_helper(DEFAULT_MEMBER_ACCOUNT_ID, 11);
+
+        let channel_state_bloat_bond = 1;
```
dobertRowneySr, line 88: Why not using `DEFAULT_CHANNEL_STATE_BLOAT_BOND` same thing also for `DATA_STATE_BLOAT_BOND`?

runtime-modules/content/src/lib.rs
```diff
@@ -1185,6 +1213,21 @@ decl_module! {
...
 
+        /// Updates channel state bloat bond value.
+        /// Only lead can upload this value
+        #[weight = 10_000_000] // TODO: adjust weight
+        pub fn update_channel_state_bloat_bond(
+            origin,
+            new_channel_state_bloat_bond: BalanceOf<T>,
+        ) {
+            let sender = ensure_signed(origin)?;
+            ensure_authorized_to_update_channel_state_bloat_bond::<T>(&sender)?;
```
dobertRowneySr, line 1266:
```suggestion
            ensure_authorized_to_update_channel_state_bloat_bond::<T>(&sender)?;
            // == MUTATION SAFE ==
```

dobertRowneySr (changes requested):
Cosmetics, otherwise LGTM. Good job!

dobertRowneySr (approved):
LGTM

dobertRowneySr (approved):

dobertRowneySr:
- build node & runtime :heavy_check_mark: 
- types and metadata :heavy_check_mark: 
- cargo test :heavy_check_mark: 
- there are some problem with benchmarks: 
```
---- tests::council_discard_remaining_rewards_on_depose stdout ----
thread 'tests::council_discard_remaining_rewards_on_depose' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::council_many_cycle_rewards stdout ----
thread 'tests::council_many_cycle_rewards' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17

---- tests::council_rewards_are_paid stdout ----
thread 'tests::council_rewards_are_paid' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17

---- tests::vote_stake_locks_after_new_election_starts stdout ----
thread 'tests::vote_stake_locks_after_new_election_starts' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17
```
These should be fixed see https://github.com/mnaamani/joystream/pull/13

Issue #3325, opened 2022-02-28
ensure_member_controller_account_origin mock, Incomplete implementation 


L3pereira:
@shamil-gadelshin 
I found that the function implementation of ```ensure_member_controller_account_origin``` used in tests, is not correct and is not reflecting the real outputs that this function has.
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/tests/mocks.rs#L209-L222

In reality we validate the origin eg. in the ```cancel_bounty(origin, creator: BountyActor<MemberId<T>>, bounty_id: T::BountyId)``` with 
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/lib.rs#L777-L780
Inside ```ensure_bounty_actor_manager``` we have the Membership implementation 
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/actors.rs#L35

If we analyze what the above function does, we can see we will return UnsignedOrigin in case the origin is not a signed account. 
Then we validate the existence of an actor_id and his controller account (bellow), if the validation fails we return MemberProfileNotFound or ControllerAccountRequired.
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/membership/src/lib.rs#L1102-L1130

This means that in tests the following function returns only Ok or BadOrigin error, but in reality the code will return one of the following errors 
ControllerAccountRequired
UnsignedOrigin
MemberProfileNotFound
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/tests/mocks.rs#L210 

The test below I think is not correct because, in reality, the error should be ControllerAccountRequired or MemberProfileNotFound.
a ```NotBountyActor``` means that the ```ensure_bounty_actor_manager``` returned Ok(()) which is not true 
https://github.com/Joystream/joystream/blob/1c37a713511cfaf35b9577069f78827492324f2c/runtime-modules/bounty/src/tests/mod.rs#L611-L624

I propose the following implementation for tests
```rust
    fn ensure_member_controller_account_origin(
        origin: Origin,
        member_id: u64,
    ) -> Result<u128, DispatchError> {
        
         let signed_account_id = frame_system::ensure_signed(origin)?;

[...]

PR #3840, opened 2022-06-03
Audit Ref 22-05-982-REP v1.1 - INFO_15_17_18 LOW_20_21_22

L3pereira:
Fixes some of the issues in the audit report #3840

Bounty module
**create_bounty**
`INFO_15 Unsafe increment of the bounty_count`

**fund_bounty**
LOW_20 Unsafe arithmetic in funding_period_expired

**announce_work_entry**
`LOW_21 Unsafe increment of the entry_count`
`INFO_18 Unsafe increment in increment_active_work_entry_counter`

**terminate_bounty**
`INFO_17 Unnecessary work in the get_terminate_bounty_actor function`

**submit_oracle_judgment**
`LOW_22 Missing event in case of reject`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202478162269313) by [Unito](https://www.unito.io)


L3pereira:
@bedeho I have to change this storagemap `Entries`
```rust
        /// Work entry storage map.
        pub Entries get(fn entries): double_map
            hasher(blake2_128_concat) T::BountyId,
            hasher(blake2_128_concat) T::EntryId => Entry<T>;
```
https://github.com/L3pereira/joystream/blob/91b5bdcc49e78d269e933e82fcdcc79002e86f81/runtime-modules/bounty/src/lib.rs#L567-L569

into this 
```rust
        /// Work entry storage map.
        pub Entries get(fn entries): double_map
            hasher(blake2_128_concat) T::BountyId,
            hasher(blake2_128_concat) T::EntryId => Option<Entry<T>>; <-------
```
because we have to implement std::default::Default in `EntryRecord`
```rust
#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
pub struct EntryRecord<AccountId, MemberId, BlockNumber> {
    /// Work entrant member ID.
    pub member_id: MemberId,

    /// Account ID for staking lock.
    pub staking_account_id: AccountId,

    /// Work entry submission block.
    pub submitted_at: BlockNumber,

    /// Signifies that an entry has at least one submitted work.
    pub work_submitted: bool,
}
```
https://github.com/L3pereira/joystream/blob/91b5bdcc49e78d269e933e82fcdcc79002e86f81/runtime-modules/bounty/src/lib.rs#L449-L463

The problem is, in this substrate version `AccountId` doesn't implement `std::default::Default`,
so I've to implement an `Option<Entry<T>>` because we've no default value for `AccountId`.
for example instead of verifying this way 
```rust
    // Verifies work entry existence and retrieves an entry from the storage.
    fn ensure_work_entry_exists(
        bounty_id: &T::BountyId,
        entry_id: &T::EntryId,
    ) -> Result<Entry<T>, DispatchError> {
        ensure!(
            <Entries<T>>::contains_key(bounty_id, entry_id),
            Error::<T>::WorkEntryDoesntExist
        );

        let entry = Self::entries(bounty_id, entry_id);

        Ok(entry)
    }
```
We can use something like this
```rust
    // Verifies work entry existence and retrieves an entry from the storage.
    fn ensure_work_entry_exists(
        bounty_id: &T::BountyId,
        entry_id: &T::EntryId,
    ) -> Result<Entry<T>, DispatchError> {
        match Self::entries(bounty_id, entry_id) {
            Some(entry) => Ok(entry),
            None => Err(Error::<T>::WorkEntryDoesntExist)
        }
    }
```

L3pereira:
- As described above there was a change in the Entries storage map (I'm open to a better solution)
- Constants in runtime lib have higher values due to (`currency::DOLLARS`), this makes the bounty benchmarks throw errors while generating weights, because of this, some initial values and asserts were modified.

runtime-modules/bounty/src/tests/mocks.rs
```diff
@@ -35,17 +24,31 @@ frame_support::construct_runtime!(
         NodeBlock = Block,
         UncheckedExtrinsic = UncheckedExtrinsic,
     {
-        System: frame_system::{Module, Call, Config, Storage, Event<T>},
-        Membership: membership::{Module, Call, Storage, Event<T>},
-        Balances: balances::{Module, Call, Storage, Config<T>, Event<T>},
-        Bounty: bounty::{Module, Call, Storage, Event<T>},
-        Referendum: referendum::<Instance1>::{Module, Call, Storage, Event<T>},
-        Council: council::{Module, Call, Storage, Event<T>},
+        System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
```
Lezek123:
```suggestion
        System: frame_system,
```

runtime-modules/bounty/src/tests/mocks.rs
```diff
@@ -35,17 +24,31 @@ frame_support::construct_runtime!(
...
+        Membership: membership::{Pallet, Call, Storage, Event<T>},
+        Balances: balances::{Pallet, Call, Storage, Config<T>, Event<T>},
```
Lezek123:
```suggestion
        Balances: balances,
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -617,32 +678,75 @@ decl_event! {
...
+
+        /// Work entry stake slashed.
+        /// Params:
+        /// - bounty ID
+        /// - entry ID
+        /// - stake account
+        WorkEntrantStakeSlashed(
+            BountyId,
+            EntryId,
+            AccountId),
```
Lezek123: I think this event is missing the slashed amount (which may vary)

runtime-modules/bounty/src/lib.rs
```diff
@@ -818,149 +905,213 @@ decl_module! {
...
-            let current_bounty_stage = Self::get_bounty_stage(&bounty);
 
-            Self::ensure_bounty_stage(
-                current_bounty_stage,
-                BountyStage::Funding { has_contributions: false }
+            let terminate_bounty_actor = BountyActorManager::<T>::ensure_bounty_actor_manager(
+                origin,
+                bounty.creation_params.creator.clone(),
             )?;
+            let bounty_creator_manager = Self::ensure_creator_actor_manager(&bounty)?;
```
Lezek123: Isn't the `let bounty_creator_manager = Self::ensure_creator_actor_manager(&bounty)?;` redundant here?
It looks like it returns the same value as the call above, ie.:
```rust
            let terminate_bounty_actor = BountyActorManager::<T>::ensure_bounty_actor_manager(
                origin,
            )?;
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1114,93 +1285,91 @@ decl_module! {
...
+            // Update entry
+            <Entries<T>>::mutate(bounty_id, entry_id, |entry| {
+                *entry = entry.clone().map(|x| {
+                    Entry::<T> {
+                        member_id: x.member_id,
+                        staking_account_id: x.staking_account_id,
+                        submitted_at: x.submitted_at,
+                        work_submitted: true,
+                    }
+                });
```
Lezek123:
```suggestion
                entry.as_mut().map(|e| { e.work_submitted = true; });
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1212,103 +1381,196 @@ decl_module! {
...
-                if matches!(*work_entry_judgment, OracleWorkEntryJudgment::Winner{ .. }) {
-                    <Entries<T>>::mutate(bounty_id, entry_id, |entry| {
-                        entry.oracle_judgment_result = Some(*work_entry_judgment);
-                    });
-                } else {
-                    let entry = Self::entries(bounty_id, entry_id);
+                match *work_entry_judgment{
+                    OracleWorkEntryJudgment::Winner{ reward } => {
+
+                        let entry = Self::ensure_work_entry_exists(&bounty_id, entry_id)?;
```
Lezek123: This should be moved above `match` to avoid redundancy

runtime-modules/bounty/src/lib.rs
```diff
@@ -1212,103 +1381,196 @@ decl_module! {
...
+                        // Fire an event.
+                        Self::deposit_event(RawEvent::WorkEntrantFundsWithdrawn(bounty_id, *entry_id, entry.member_id));
 
-                    Self::deposit_event(RawEvent::WorkEntrySlashed(bounty_id, *entry_id));
+                    },
+                    OracleWorkEntryJudgment::Rejected{
+                        slashing_share,
+                        ..
+                    } => {
+                        let entry = Self::ensure_work_entry_exists(&bounty_id, entry_id)?;
```
Lezek123: Same as above

runtime-modules/bounty/src/lib.rs
```diff
@@ -1317,11 +1579,12 @@ decl_module! {
         /// # <weight>
         ///
         /// ## weight
-        /// `O (1)`
+        /// `O (N)`
+        /// - `N` is msg
```
Lezek123:
```suggestion
        /// - `N` is `msg` length
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1346,11 +1609,12 @@ decl_module! {
...
```
Lezek123:
```suggestion
        /// - `N` is `msg` length
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1382,11 +1646,12 @@ decl_module! {
...
```
Lezek123:
```suggestion
        /// - `N` is `msg` length
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1412,11 +1681,12 @@ decl_module! {
...
```
Lezek123:
```suggestion
        /// - `N` is `msg` length
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1519,20 +1821,6 @@ impl<T: Config> Module<T> {
...
-                    Error::<T>::ClosedContractMemberNotFound
-                );
-            }
-        }
-
-        if let BountyActor::Member(member_id) = params.oracle {
-            ensure!(
-                T::Membership::controller_account_id(member_id).is_ok(),
-                Error::<T>::InvalidOracleMemberId
-            );
```
Lezek123: Those checks were added in `carthage` during https://github.com/Joystream/joystream/issues/4074, so they should be kept (along with the related tests)

runtime-modules/bounty/src/lib.rs
```diff
@@ -1587,52 +1940,59 @@ impl<T: Config> Module<T> {
...
-            let _ = balances::Pallet::<T>::slash(&bounty_account_id, all);
-        }
+        Self::deposit_event(RawEvent::CreatorStateBloatBondWithdrawn(
+            *bounty_id,
+            bounty.creation_params.creator.clone(),
+            T::CreatorStateBloatBondAmount::get(),
+        ));
+
+        <Bounties<T>>::remove(bounty_id);
+        <BountyContributions<T>>::remove_prefix(bounty_id, None);
```
Lezek123: Isn't `remove_prefix` here an issue, since:
1. Expected map size is not provided as a parameter
2. Number of entries in this map doesn't seem constrained

So essentially we don't know the number of iterations this will result in and the weight does not account for that in any way.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1737,18 +2102,23 @@ impl<T: Config> Module<T> {
...
-        );
+        // ensure!(
+        //     <Entries<T>>::contains_key(bounty_id, entry_id),
+        //     Error::<T>::WorkEntryDoesntExist
+        // );
+
+        // let entry = Self::entries(bounty_id, entry_id);
 
-        let entry = Self::entries(bounty_id, entry_id);
+        // Ok(entry)
```
Lezek123: Remove the commented-out code

runtime-modules/bounty/src/tests/fixtures.rs
```diff
@@ -696,8 +568,8 @@ impl SubmitWorkFixture {
 
         let new_entry = Bounty::entries(self.bounty_id, self.entry_id);
 
-        if actual_result.is_ok() {
-            assert!(new_entry.work_submitted);
+        if actual_result.is_ok() && new_entry.is_some() {
```
Lezek123:
```suggestion
        if actual_result.is_ok() {
```
`actual_result.is_ok()` already implies that `new_entry.is_some()`

runtime-modules/bounty/src/lib.rs
```diff
@@ -54,68 +69,50 @@ mod stages;
...
-use common::council::CouncilBudgetManager;
-use common::membership::{
-    MemberId, MemberOriginValidator, MembershipInfoProvider, StakingAccountValidator,
-};
 use staking_handler::StakingHandler;
 
 /// Main pallet-bounty trait.
 pub trait Config:
-    frame_system::Config + balances::Config + common::membership::MembershipTypes
+    frame_system::Config + balances::Config + common::membership::MembershipTypes + TypeInfo
```
Lezek123: I think you should use this: https://docs.rs/scale-info/latest/scale_info/#scale_infoskip_type_params (instead of adding `TypeInfo` to `Config`)

runtime-modules/bounty/src/lib.rs
```diff
@@ -397,13 +411,12 @@ pub type Entry<T> = EntryRecord<
...
     <T as common::membership::MembershipTypes>::MemberId,
     <T as frame_system::Config>::BlockNumber,
-    BalanceOf<T>,
 >;
 
 /// Work entry.
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug)]
-pub struct EntryRecord<AccountId, MemberId, BlockNumber, Balance> {
+// #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
Lezek123: Why was this line commented-out?

runtime-modules/bounty/src/lib.rs
```diff
@@ -1114,93 +1285,91 @@ decl_module! {
...
+        /// (not surpassing the total fund amount)
         /// # <weight>
         ///
         /// ## weight
-        /// `O (N)`
-        /// - `N` is the work_data length,
+        /// `O (N x M + Z)`
+        /// - `N` is judgment,
+        /// - `M` is action_justification (inside OracleJudgment),
+        /// - `Z` is rationale
```
Lezek123:
```suggestion
        /// `O (N + M + Z)`
        /// - `N` is number of judgment entries,
        /// - `M` is the sum of all action_justification lengths (inside OracleJudgment),
        /// - `Z` is rationale length
```

I think you should also adjust the benchmarks to reflect this, because currently you're passing the sum of the length of all `action_justification`s as a parameter to the weight function, but the benchmarks effectively multiply the impact of this parameter by `MAX_WORK_ENTRIES` (because there you create one `action_justification` of length `k` per each entry), which means the user will be overpaying a lot.

runtime-modules/bounty/src/lib.rs
```diff
@@ -1971,4 +2270,43 @@ impl<T: Config> Module<T> {
...
+                rationale,
+            ),
+        )
+        .max(
+            WeightInfoBounty::<T>::submit_oracle_judgment_by_member_all_rejected(
+                collection_length,
+                justification_length,
+                rationale,
+            ),
+        )
```
Lezek123: The order of the parameters is wrong.
In `submit_oracle_judgment_by_member_all_rejected` and `submit_oracle_judgment_by_council_all_rejected` the parameters `i`, `j`, `k` are the judgment entries count, rationale length and justification length respectively.
Here you're passing `justification_length` as `j` and `rationale` as `k` instead.

Lezek123 (changes requested):
1. Update `chain-metadata.json` and the types
2. The doc-comment over `BountyCreatorRemarked` incorrectly states `/// Bounty entrant made a message remark`
3. `generate-weights.sh` - I suppose you can now re-enable the bounty module
4. There's a serious issue with `validate_judgment`, because `reward_sum_from_judgment` is vulnerable to being purposefully overflown just to pass the `reward_sum_from_judgment == bounty.total_funding` check. Any unpaid reward will then be ignored, becasue `transfer_funds_from_bounty_account` is infallible (actually I think you should also make it fallible again, because now we have the default `transactional` protection)
5. Remark tests are missing cases where `BountyActor::Member(id)` is, for example, a valid creator (in case of `creator_remark`), but the signer is not the member controller account of this member. I think that's the case for all extrinsics that rely on `BountyActor` checking, because the mocks don't even differentiate between valid and invalid controller account of a member (`ensure_member_controller_account_origin` never fails)

runtime-modules/bounty/src/lib.rs
L3pereira: Could you be more clear please, about what you want me to do in benchmarks?

The weight should be the N x M + Z in case of all rejected and N + Z in case of all successful

runtime-modules/bounty/src/lib.rs
L3pereira: We will have to implement another map to map a bountyId to the contributions count

L3pereira: @bedeho

L3pereira:
@Lezek123 
4. There's a serious issue with `validate_judgment`, because `reward_sum_from_judgment` is vulnerable to being purposefully overflown just to pass the `reward_sum_from_judgment == bounty.total_funding` check. Any unpaid reward will then be ignored, becasue `transfer_funds_from_bounty_account` is infallible (actually I think you should also make it fallible again, because now we have the default `transactional` protection)

Could you write a test where you exploit this vulnerability, I don't see how I can overflow the `reward_sum_from_judgment` and pass `reward_sum_from_judgment == bounty.total_funding`

L3pereira:
5. Remark tests are missing cases where BountyActor::Member(id) is, for example, a valid creator (in case of creator_remark), but the signer is not the member controller account of this member. I think that's the case for all extrinsics that rely on BountyActor checking, because the mocks don't even differentiate between valid and invalid controller account of a member (ensure_member_controller_account_origin never fails)

I opened an [issue](https://github.com/Joystream/joystream/issues/3325#issuecomment-1057113676) to implement `ensure_member_controller_account_origin`, but doesn't make sense to implement it, you would be just testing a mock function.
So Issue was closed

runtime-modules/bounty/src/lib.rs
L3pereira: I guess this is not a problem, since this is called in `remove_bounty` and this function must be called after 
`has_no_contributions_and_no_work_entries`, so there should be no contributions at this stage.
In fact, remove_prefix is not needed, I tested and after removing all contributions the keys were also removed.
I also added more checks in the unit tests, to ensure there are no contributions

bedeho:
The problem is that you are doing addition which can overflow here 

[https://github.com/L3pereira/joystream/blob/audit_report_ref_22-05-982-REP_v1.1_%233824/runtime-modules/bounty/src/lib.rs#L2183](https://github.com/L3pereira/joystream/blob/audit_report_ref_22-05-982-REP_v1.1_#3824/runtime-modules/bounty/src/lib.rs#L2183)

So the oracle can provide a `judgment` where the rewards cause `reward_sum_from_judgment` to overflow, but the final value still matches the `bounty.total_funding`, this is trivial. This means the final check `reward_sum_from_judgment == bounty.total_funding` will incorrectly pass. The good thing is that when reward payouts happen, it is constrained by the balance of the bounty module account in this transfer
[https://github.com/L3pereira/joystream/blob/audit_report_ref_22-05-982-REP_v1.1_%233824/runtime-modules/bounty/src/lib.rs#L1407](https://github.com/L3pereira/joystream/blob/audit_report_ref_22-05-982-REP_v1.1_#3824/runtime-modules/bounty/src/lib.rs#L1407), however, it should still not happen, and in the future if the bounty module is changed without someone being aware of this, it could cause very bad unintended consequences.

Even if it’s not clear to you how exploitable this is at this point, lets just have it fixed, as it is more correct.

runtime-modules/bounty/src/lib.rs
L3pereira: Clippy suggested instead
```rust
            // Update entry
            <Entries<T>>::mutate(bounty_id, entry_id, |entry| {
                if let Some(e) = entry.as_mut() { e.work_submitted = true; };
            });
```

runtime-modules/bounty/src/lib.rs
Lezek123: So in `submit_oracle_judgment_by_council_all_rejected` and `submit_oracle_judgment_by_member_all_rejected` you have a parameter `k` and in `submit_oracle_judgment_weight` you have `justification_length` which represents the same parameter.

In `submit_oracle_judgment_weight`, the `justification_length` is **the sum of all `action_justification.len()` in the `judgment_map`**

In `submit_oracle_judgment_by_council_all_rejected` and `submit_oracle_judgment_by_member_all_rejected` benchmarks, `k` is the length of **each `action_justification` in the `judgement` map (so the sum of all `action_justification.len()`s is `k * i`)**

So `k` and `justification_length` mean 2 different things, while they should mean the same one.
One way to solve this is to make `k` in `submit_oracle_judgment_by_council_all_rejected` and `submit_oracle_judgment_by_member_all_rejected` actually represent the sum of all `action_justification.len()`s, for example:

```diff
        let judgment = (0..i).into_iter().map(|it|{
            let entry_id = announce_entry_and_submit_work::<T>(&bounty_id, it);
            let judgment = OracleWorkEntryJudgment::Rejected {
                slashing_share: Perbill::from_percent(50),
-               action_justification: action_justification.clone(),
+               action_justification: match it == 0 {
+                   true => vec![0u8].repeat((k / i + k % i) as usize),
+                   false => vec![0u8].repeat((k / i) as usize)
+               },
            };
            (entry_id, judgment)
        }).collect::<BTreeMap<_, _>>();
```

Another issue here is I don't think `all_rejected` and `all_winners` actually represent good cases for benchmarking, because:
- `all_winners` doesn't take into account `justification_length`
- `all_rejected` doesn't take into account the actual number of winners

So imagine a scenario where there is 1 `Rejected` entry with a huge `justification_length` (for example, 3 megabytes) and 100 `Winner` entires in the `judgment` (so total of 101 entries). In that case:
- `all_winners` weight function may give an underestimated value, because it won't take into account the huge impact that `justification_length` of the single `Rejected` entry would have on the weight
- `all_rejected` weight function may also give an underestimated value, because it won't take into account the huge impact that processing 100 `Winner`s would have (I'm assuming processing a `Winner` is of higher complexity, because it involves a transaction, while processing rejected entry involves only slashing)

Therefore **I think that the number of `Winner`/`Rejected` entries should also be the weight function parameters.**

Lezek123:
> doesn't make sense to implement it, you would be just testing a mock function.

You would be testing at least whether it's actually being called when expected or not. Right now the tests would pass regardless of whether, for example, the `ensure_bounty_actor_manager` function actually calls `ensure_member_controller_account_origin` or not. I think it makes sense to have tests like that (we have such tests for content pallet for example).

runtime-modules/bounty/src/tests/mod.rs
```diff
@@ -1049,6 +1062,27 @@ fn terminate_bounty_fails_with_invalid_origin() {
...
+#[test]
+fn terminate_bounty_fails_with_invalid_member_controller_account() {
+    build_test_externalities().execute_with(|| {
+        let account_id = 1;
+        let initial_balance = 500;
+
+        increase_account_balance(&account_id, initial_balance);
+        set_council_budget(initial_balance);
+
+        // Created by council - try to cancel with bad origin
```
Lezek123:
```suggestion
        // Created by member - try to cancel with bad origin
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -1343,7 +1343,7 @@ decl_module! {
         /// # <weight>
         ///
         /// ## weight
-        /// `O (N x M + Z)`
+        /// `O (N + M + Z)`
         /// - `N` is number of judgment entries,
         /// - `M` is the sum of all action_justification lengths (inside OracleJudgment),
         /// - `Z` is rationale length
```
Lezek123: The comment is outdated, as there are now 4 parameters

runtime-modules/bounty/src/tests/mod.rs
```diff
@@ -4821,6 +4972,38 @@ fn switch_oracle_invalid_origin() {
     });
 }
 
+#[test]
+fn switch_oracle_fails_with_invalid_new_oracle_member_id() {
```
Lezek123, line 5008: Missing `switch_oracle_fails_with_invalid_member_controller_account` case

runtime-modules/storage/src/weights.rs
```diff
@@ -18,7 +18,7 @@
 //! Autogenerated weights for storage
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-07-28, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2022-08-23, STEPS: `5`, REPEAT: 2, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
Lezek123: Please revert `runtime-modules/storage/src/weights.rs` to the previous state, there's no reason to update this file in this PR

Lezek123 (commented):
There's also small conflict to fix, otherwise looks good :+1:

runtime-modules/bounty/src/tests/mod.rs
L3pereira: You cannot test for `ensure_member_controller_account_origin` in this case. Since in `ensure_switch_oracle_actors` doesn't stop execution (no ?) 

If ensure_bounty_actor_manager returns an error, It will be `BadOrigin` from `ensure_root` or `MemberProfileNotFound` from `controller_account_id`
```rust
        let switcher_actor_manager =
            BountyActorManager::<T>::ensure_bounty_actor_manager(origin.clone(), current_oracle);

        let switcher = match switcher_actor_manager {
            Ok(oracle_switcher_manager) => oracle_switcher_manager.get_bounty_actor(),
            Err(_) => {
                ensure_root(origin)?;
                BountyActor::Council
            }
        };

        //Check if new oracle is a member
        BountyActorManager::<T>::get_bounty_actor_manager(new_oracle)?;
        Ok(switcher)
```

this is tested in `switch_oracle_fails_with_invalid_new_oracle_member_id`

Lezek123: What I meant was to test the case where the caller is not a valid member controller of the **current** orcale, ie. this specific path:
`switch_oracle` => `ensure_switch_oracle_actors` => `ensure_bounty_actor_manager(origin.clone(), current_oracle)` => `ensure_member_controller_account_origin`

L3pereira: [switch_oracle_fails_with_invalid_member_controller_of_current_oracle](https://github.com/Joystream/joystream/pull/3840/commits/d10800b7ab7f228b4a5a9bfe701e6af6e5f849a2)

Lezek123 (approved):
LGTM

PR #3843, opened 2022-06-03
add reactionsAndRepliesCount field to Comment entity

### LINKED ISSUE #3839: Add `reactionsAndRepliesCount` to `Comment` entity
Add `reactionsAndRepliesCount` field to `Comment` entity. It would hold the sum of both `reactionsCount ` & `repliesCount `.

zeeshanakram3:
resolve #3839

### DIFF
--- a/query-node/mappings/src/content/commentAndReaction.ts
+++ b/query-node/mappings/src/content/commentAndReaction.ts
@@ -160,7 +160,7 @@ async function getOrCreateCommentReactionsCountByReactionId(
   return reactionsCountByReactionId
 }
 
-export function setReactionsCount(
+export function setReactionsAndRepliesCount(
   event: SubstrateEvent,
   entity: Video | Comment,
   reactionsCountByReactionId: VideoReactionsCountByReactionType | CommentReactionsCountByReactionId,
@@ -172,6 +172,7 @@ export function setReactionsCount(
     reactionsCountByReactionId.updatedAt = eventTime
 
     ++entity.reactionsCount
+    if (entity instanceof Comment) ++entity.reactionsAndRepliesCount
     entity.updatedAt = eventTime
     return
   }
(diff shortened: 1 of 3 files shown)


query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -172,6 +172,7 @@ export function setReactionsCount(
     reactionsCountByReactionId.updatedAt = eventTime
 
     ++entity.reactionsCount
+    if (entity instanceof Comment) ++entity.reactionsAndRepliesCount
```
ondratra: Please don't use one-line `if`s. While it is a valid construct, it makes the code harder to read. 
```suggestion
    if (entity instanceof Comment) {
      ++entity.reactionsAndRepliesCount
    }
```

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -180,6 +181,7 @@ export function setReactionsCount(
   reactionsCountByReactionId.updatedAt = eventTime
 
   --entity.reactionsCount
+  if (entity instanceof Comment) --entity.reactionsAndRepliesCount
```
ondratra:
```suggestion
  if (entity instanceof Comment) {
    --entity.reactionsAndRepliesCount
  }
```

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -160,7 +160,7 @@ async function getOrCreateCommentReactionsCountByReactionId(
   return reactionsCountByReactionId
 }
 
-export function setReactionsCount(
+export function setReactionsAndRepliesCount(
```
ondratra, line 163: While it's not the scope of this PR, I've noticed that this function has a separate section for increment and decrement that are almost identical. Let's unite the code for both cases in the following fashion:
```
const change = (operation == 'INCREMENT') ? 1 : -1
...
entity.reactionsCountByReactionId += change
entity.reactionsCount += change
...
```

ondratra (changes requested):

ondratra (approved):
LGTM

PR #3796, opened 2022-05-26
Creator token transfers: Validate dst member_id

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3791

PR #3871, opened 2022-06-07
Carthage revenue split rate

dobertRowneySr:
#### Changes
* added revenue split rate to token data struct in the `project_token` pallet
* `revenue_split_allocation = (channel_allocation - existential_deposit) * revenue_split_rate`
* there where two tests not related to the revenue split failing, one was due to parameters being passed in the wrong order to the method, the other was due to underflow in `next_sale_id - 1` subtraction. I changed the init value for `next_sale_id` from `0` to `1`
#### To Do
* establish min and max bounds for the revenue split percentage in accordance with https://github.com/Joystream/atlas/issues/2728#issuecomment-1138615962 point 5

#### Notes
* rebased on `carthage-substratev3`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476245561157) by [Unito](https://www.unito.io)


runtime-modules/project-token/src/tests/canonical.rs
```diff
@@ -954,9 +955,10 @@ fn issue_token_ok_with_token_info_added() {
                     rate,
                 },
                 sale: None,
-                next_sale_id: 0,
+                next_sale_id: 1,
```
Lezek123: Should no longer be needed after https://github.com/Joystream/joystream/pull/3928/commits/36eaa385575f14bd529baa1ddc71452f0ed857b4

runtime-modules/project-token/src/lib.rs
```diff
@@ -1086,17 +1086,19 @@ impl<T: Config>
...
+        let token_info = Self::ensure_token_exists(token_id)?;
+        token_info.revenue_split.ensure_inactive::<T>()?;
+
+        let allocation_amount = token_info
+            .revenue_split_rate
+            .mul_floor(nominal_allocation_amount);
         ensure!(
             !allocation_amount.is_zero(),
             Error::<T>::CannotIssueSplitWithZeroAllocationAmount,
         );
```
Lezek123: Since now only part of the channel balance amount is used for the revenue split, while the remaining part is supposed to go to the creator, you need to add the transfer logic for this second part (which goes directly to the creator).

Note that now it's also possible to issue a token with `revenue_split_rate` equal to `0`, which effectively blocks any attempt to withdraw funds from the channel, because issuing a revenue split is the only way to do this when there's a creator token issued. I think we should just disallow issuing a creator token with `revenue_split_rate == 0` (or possibly even have a non-zero `MinRevenueSplitRate`).

The current implementation is also causing the `content` pallet creator token tests to fail with `CannotIssueSplitWithZeroAllocationAmount`:

```
failures:
    tests::creator_tokens::finalize_revenue_split::successful_finalize_curator_channel_revenue_split_by_curator
    tests::creator_tokens::finalize_revenue_split::successful_finalize_curator_channel_revenue_split_by_lead
    tests::creator_tokens::finalize_revenue_split::successful_finalize_member_channel_revenue_split_by_collaborator
    tests::creator_tokens::finalize_revenue_split::successful_finalize_member_channel_revenue_split_by_owner
    tests::creator_tokens::issue_revenue_split::successful_issue_curator_channel_revenue_split_by_curator
    tests::creator_tokens::issue_revenue_split::successful_issue_curator_channel_revenue_split_by_lead
    tests::creator_tokens::issue_revenue_split::successful_issue_member_channel_revenue_split_by_collaborator
    tests::creator_tokens::issue_revenue_split::successful_issue_member_channel_revenue_split_by_owner
```

runtime-modules/project-token/src/tests/split.rs
```diff
@@ -296,7 +301,7 @@ fn finalize_split_ok_with_event_deposit() {
         last_event_eq!(RawEvent::RevenueSplitFinalized(
             1u64,
             member!(1).1,
-            DEFAULT_SPLIT_ALLOCATION
+            DEFAULT_SPLIT_ALLOCATION_RATE * DEFAULT_SPLIT_ALLOCATION,
```
Lezek123: The term "split allocation" now became kind of ambiguous, because it's used both to mean the channel balance (minus existential deposit) at the time of the split issuance (the `full_amount`) and the amount actually to be split between token holders (`full_amount * split_allocation_rate`). I think we should come up with some 2 separate terms for those and use them consistently.

runtime-modules/project-token/src/tests/transfer.rs
```diff
@@ -1025,6 +1025,33 @@ fn issuer_transfer_fails_with_non_existing_source() {
     })
 }
 
+#[test]
+fn issuer_transfer_fails_with_non_existing_dst_member() {
```
Lezek123, line 1052: Hmm, interesting, this looks like it's actually coming from https://github.com/Joystream/joystream/pull/3796, but some of the changes are still missing (for example, `transfer_fails_with_non_existing_dst_member` test case)

runtime-modules/project-token/src/types.rs
```diff
@@ -1200,10 +1206,12 @@ where
             sale: None,
             transfer_policy: params.transfer_policy.into(),
             patronage_info,
-            next_sale_id: 0,
+            next_sale_id: 1,
```
Lezek123: Should no longer be needed after https://github.com/Joystream/joystream/pull/3928/commits/36eaa385575f14bd529baa1ddc71452f0ed857b4

Lezek123 (changes requested):

Lezek123:
After addressing the comments, please merge carthage and re-generate `chain-metadata.json` and the types

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I fixed all tests in the `content` pallet
I agree with disallowing to issue a token with `0` `revenue_split_rate` I added a guard and a test for that.
I could not understand the first part, I also added a test which (if I understood it properly) should test your first assertion and it work as expected.
As of now the creator retains in his account `(100% - revenue_split_rate) * allocation`
Can you rephrase maybe?

runtime-modules/project-token/src/tests/transfer.rs
dobertRowneySr: I added back all the changes in that PR: f5d8b50a3ce09be67cdb721e88f790f3b6be8d3d

runtime-modules/project-token/src/tests/split.rs
dobertRowneySr: I used `allocation` for the actual gross allocation and `revenue` for `allocation * rate`

dobertRowneySr: 38a7c6e3a1f0cbb2df2e044be22357bea656e207

Lezek123: 1. I think `revenue` for gross amount and `allocation` for `revenue * rate` actually makes more sense. My reasoning here being: `revenue` is 100% of the channel revenue, to be split between creator and token holders, `allocation` is the amount that's allocated to token holders only.
2. Make sure to change it in the tests/comments/mocks/traits etc. also

runtime-modules/project-token/src/lib.rs
Lezek123: > I could not understand the first part, I also added a test which (if I understood it properly) should test your first assertion and it work as expected.
As of now the creator retains in his account (100% - revenue_split_rate) * allocation
Can you rephrase maybe?

The amount for the split comes from channel account balance, so if you have, let's say, 10% `split_rate` and you issue a revenue split, 90% will remain in the channel account and the creator won't be able to access it. What we want is to allocate 10% for the split and **transfer** the rest to the creator (member controller account in case of member channel and probably council budget in case of curator channel), so this is what is missing. It should be added in `content` pallet.

Lezek123: See: https://github.com/Joystream/joystream/blob/carthage/runtime-modules/content/src/lib.rs#L2692

runtime-modules/project-token/src/tests/split.rs
dobertRowneySr: fixed here : e6d8a3004abc21147f33f3080a196b8db571ed44

runtime-modules/project-token/src/lib.rs
dobertRowneySr: done also added respective tests: 4738b9ec3e1158d2ae366494eda7ad333eb6d2c4, e826da33caa420bd9aac570cd2eea9fb06d2ca1b

runtime-modules/project-token/src/tests/fixtures.rs
```diff
@@ -761,7 +762,7 @@ impl IssueRevenueSplitFixture {
             start: None,
             duration: BlockNumber::from(DEFAULT_SPLIT_DURATION),
             allocation_source: member!(1).1,
-            allocation: Balance::from(DEFAULT_SPLIT_ALLOCATION),
+            allocation: Balance::from(DEFAULT_SPLIT_REVENUE),
```
Lezek123: The allocation/revenue naming is still inconsistent here (the field name is `allocation`, but you're assigning `revenue`).
This is also the case in some tests (`burn_fails_with_active_revenue_split`, `issue_revenue_split_ok_with_allocation_leftovers_retained_by_issuer`)

Lezek123 (changes requested):
Looks good, just some naming inconsistencies left

Lezek123:
Also - `cargo fmt` produces some differences in output again (perhaps you could use some formatting plugin for the IDE to make sure the files always stay correctly formatted)

ignazio-bovo:
- :ballot_box_with_check: runtime & node build
- :ballot_box_with_check: metadata generated correctly
- :ballot_box_with_check: types generated correctly
- :x: cargo test successful

I got these
```
---- tests::council_discard_remaining_rewards_on_depose stdout ----
thread 'tests::council_discard_remaining_rewards_on_depose' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::council_many_cycle_rewards stdout ----
thread 'tests::council_many_cycle_rewards' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17

---- tests::council_rewards_are_paid stdout ----
thread 'tests::council_rewards_are_paid' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17

---- tests::vote_stake_locks_after_new_election_starts stdout ----
thread 'tests::vote_stake_locks_after_new_election_starts' panicked at 'attempt to subtract with overflow', runtime-modules/council/src/mock.rs:1496:17

```

Lezek123 (commented):
LGTM, but the chain metadata and types seem out of date

Lezek123 (approved):
Actually it's ok, I was just pulling metadata from a wrong node

PR #3814, opened 2022-05-31
Audit3: Merge Ephesus

Lezek123:
- [x] Conflict resolutions:
	- [x] storage tests (new fixture macros introduced on `ephesus` vs a lot of functionality changes in `audit3`)
	- [x] council/working-group budget refactorization (`ephesus`)
	- [x] new proposals beeing introduced ([`UpdateGlobalNftLimitsProposal`](https://github.com/Joystream/joystream/pull/3706) on `audit3`, [`UpdateChannelPayoutsProposal`](https://github.com/Joystream/joystream/pull/3444) on `ephesus`)
	- [x] NFT limits (`audit3`)
	- [x] NFT adjustments from `rhodes` (`rhodes` => `ephesus`)
	- [x] [Channel payouts/chashouts finalization](https://github.com/Joystream/joystream/pull/3444) (`ephesus`)
	- [x] removal of video posts feature (`audit3`)
	- [x] the introduction on creator tokens (`ephesus`)
	- [x] channel permissions model changes (+ related changes in tests) (`audit3`)
	- [x] proposals configuration rework (`olympia` => `ephesus`)
	- [x] storage interface changes (`audit3`)
- [x] `ChannelAgenetPermissions` integration w/ creator tokens functionality
- [x] Content mocks - changed `AccountId` to `u128` to fix some of the tests (differentiate between content module treasury account vs channel accounts)
- [x] Channel transfers - [establish which features should be blocked](https://github.com/Joystream/joystream/issues/3428) and what other constraints to be applied (ie. w.r.t. creator tokens)
- [x] Creator tokens and channel cashouts - `PausableChannelFeature` integration
- [ ] `claim_council_reward` and `withdraw_from_channel` unification (both should not be possible when creator token is issued, should require the same permissions etc.)
- [x] Prevent channel deletion if creator token is issued (moderators can still delete?)
- [x] Update `@joystream/types`

PR #3890, opened 2022-06-13
Audit3 merge ephesus ignazio

dobertRowneySr:
merges `ephesus` into `audit`
forked off Leszek's https://github.com/Joystream/joystream/pull/3814


bedeho:
I think in retrospect what we should have done was first a clean merge into audit3 which just focused on getting it up to speed with whatever ephesus was based on, and then a separate ephesus merge, because this is not feasible to really review well.

Issue #3688, opened 2022-04-29
HIGH_1: NFT- complete payment business logic vulnerability


dobertRowneySr:
# Complete Payment function current behavior
`complete_payment` function in `content/src/nft/mod.rs`:

```rust
    pub(crate) fn complete_payment(
        royalty_payment: Option<(Royalty, T::AccountId)>,
        amount: BalanceOf<T>,
        sender_account_id: T::AccountId,
        receiver_account_id: Option<T::AccountId>,
    );
```

 has the following postconditions (`royalty_fee = royalty_payment.0 * amount`, `platform_fee = platform_fee_percentage * amount`):

*  `sender_account_id` slashed by `amount`
*  if `royalty_payment.is_none()`  `receiver_account_id` is credited with `amount - platform_fee`
*  else if `royalty_fee + platform_fee <= amount` (or equivalently `royalty_payment.0 + platform_fee_percentage <= 100%` then `receiver_account_id` credited with `amount - royalty_fee - platform_fee`
* else (case: `royalty_fee + platform_fee > amount`) `receiver_account_id` credited with `amount - platform_fee`
* `royalty_payment.1` (creator account) is credited with `royalty` if `royalty_payment.is_some()`

## Attack Vector
1.  set up 2 accounts: one **creator** one **buyer**
1.  issue a NFT with the **creator** account and NFT royalty s.t. :
    `platform_percentage + royalty > 100%`
1.  Sell NFT at `amount`
1.  Buy NFT with the **buyer** account

## Result
*   `amount` is first slashed from **creator** (in order to be transferred to the treasury account)
*   **creator** is granted with `royalty_fee `
*   **buyer**  is granted with:  `amount - platform_fee` 

Malicious user gets `[amount - platform_fee] + [royalty_fee]` :  `[buyer_account] + [creator_account]` by using just `amount` JOY,  registering `royalty_fee - platform_fee` free profit

## Recommendations
We investigated this issue and tried to determine under which circumstances it could be exploited. As the fees and royalties are percentages, the issue should only arise when `platform_fee_percentage + creator_royalty > 1`
[...]

Issue #3657, opened 2022-04-23
Simplify `complete_payment`


bedeho:
# Background
The crate method `complete_payment` is a utility method which attempts to settle value transfers for NFTs, see here 

https://github.com/Joystream/joystream/blob/rhodes/runtime-modules/content/src/nft/mod.rs#L353

It accepts `is_auction` as a way to determine whether to deduct funds from the supplied `sender_account_id`, or from the built in content treasury which holds NFT bid funds.

# Proposal
Make following improvements, and try to bring them to `ephesus` as well, with suitable changes.

## Simplify where to send funds
Just drop `is_auction` and require that caller provides the correct value of  `sender_account_id` , which really should be renamed now to `source_account_id`. So if it's an auction, the caller provides `ContentTreasury::<T>::module_account_id()`, otherwise it provides normal account`.

## Disentangle royalties, channel and owner
This entire section is a bit of a mess because it is coupling how much to pay with the payment execution, which leads to repetition and codepaths with complex semantics. The worst code small is this statement: `// Should always be Some(_) at this stage, because of previously made check.`

Now how to do this will be impacted by presence of channel accounts, which arguably will make things simpler, but it means one has to do one version for `Rhodes` and one for `Ephesus`. 

```Rust
if let Some(creator_royalty) = creator_royalty {
            let royalty = creator_royalty * amount;

            // Deposit amount, exluding royalty and platform fee into receiver account
            match receiver_account_id {
                Some(receiver_account_id) if amount > royalty + auction_fee => {
                    let _ = Balances::<T>::deposit_creating(
                        &receiver_account_id,
                        amount - royalty - auction_fee,
                    );
                }
                Some(receiver_account_id) => {
                    let _ =
                        Balances::<T>::deposit_creating(&receiver_account_id, amount - auction_fee);
                }
                _ => (),
            };

            // Should always be Some(_) at this stage, because of previously made check.
            if let Some(creator_account_id) = Self::channel_by_id(in_channel).reward_account {
                // Deposit royalty into creator account
                let _ = Balances::<T>::deposit_creating(&creator_account_id, royalty);
            }
        } else if let Some(receiver_account_id) = receiver_account_id {
            // Deposit amount, exluding auction fee into receiver account
            let _ = Balances::<T>::deposit_creating(&receiver_account_id, amount - auction_fee);
        }
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202209987416696) by [Unito](https://www.unito.io)

PR #3906, opened 2022-06-15
Ephesus  fix nft payments

dobertRowneySr:
Addresses:
- #3688 
- #3657 

#### Changes
- implemented royalty fee accounting proposed in #3688 
- `complete_payment` helper function cleaned up
- add test for case described in the security report


bedeho:
No longer needed.

Issue #3882, opened 2022-06-08
Freeze runtime features


bedeho:
# Background
As part of having a very focused low risk `Carthage` and mainnet scope, we should seek to restrict as many of the newly implement features post `rhodes` to have much of any impact on the QN or downstream Atlas/QN/Pioneer. This can easily be done by selectively disabling extrinsics that hold key code paths for triggering such features. This excellent summary of changes in the content directory was very useful in coming up with this list.

https://github.com/Joystream/joystream/issues/3874#issuecomment-1149869411

# Proposal
Freeze, by which I mean to just return `Err` as first action, the following. We don't even bother with weights being wrong, as long as they are not 0.

## Content Directory
* Initiating a channel transfer, this should block the following events and thus sidestep need for schema support, 
mappings or Atlas having to check this condition as part of all sorts of actions that are not possible in transit,
  * `UpdateChannelTransferStatus(ChannelId, ContentActor, ChannelTransferStatus)`
  * `ChannelTransferAccepted(ChannelId, TransferParameters)`
* **Turn off actual minting limit validation on issuance of NFTs:** `check_nft_limits`
* Everything having to do with updating or setting minting limits, this should block the following events and thus sidestep need for schema support, mappings or Atlas having to check this condition as part of issuing NFTs.
  * `GlobalNftLimitUpdated(NftLimitPeriod, u64)`
  * `ChannelNftLimitUpdated(ContentActor, NftLimitPeriod, ChannelId, u64)`
* Destroying NFTs: `NftDestroyed(ContentActor, VideoId)`
* Moderators pausing channel features.
  * `ChannelPausedFeaturesUpdatedByModerator`
* Updating the privilege level of a channel and also permissions of a curator group
  * `CuratorGroupPermissionsUpdated`
  * `ChannelPrivilegeLevelUpdated` 

## Bounties V2
Block bounty creation, which blocks everything else.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202416807025146) by [Unito](https://www.unito.io)

Lezek123:
> Updating the privilege level of a channel and also permissions of a curator group
> CuratorGroupPermissionsUpdated
> ChannelPrivilegeLevelUpdated

Curator group permissions can also be initially set via `create_curator_group` extrinsic, so there seems to be very little benefit in disabling just the `update_curator_group_permissions` extrinsic. Should we also modify `create_curator_group`?

PR #3914, opened 2022-06-16
Carthage freeze runtime

dobertRowneySr:
Addresses: #3882 

#### Changes
- I have used a `todo!()` macro for blocking the extrinsics and also `#[ignore]`d the relevant tests. 
- I have also added `TODO`s for future references

##### Content Directory
* [x] UpdateChannelTransferStatus(ChannelId, ContentActor, ChannelTransferStatus)
* [x] ChannelTransferAccepted(ChannelId, TransferParameters)
* [x] check_nft_limits
* [x] Everything having to do with updating or setting minting limits
* [x] GlobalNftLimitUpdated(NftLimitPeriod, u64)
* [x] ChannelNftLimitUpdated(ContentActor, NftLimitPeriod, ChannelId, u64)
* [x] Destroying NFTs: NftDestroyed(ContentActor, VideoId)
* [x] Moderators pausing channel features.
* [x] ChannelPausedFeaturesUpdatedByModerator
* [x] CuratorGroupPermissionsUpdated
* [x] ChannelPrivilegeLevelUpdated
##### Proposals
* [x] Disable Update Global Nft limits proposal
##### Bounties V2 
* [x] Block bounty creation (which blocks everything else)

#### Notes
* creator group is created with _immutable_ permissions set specified at `create_curator_group`. It can no longer be modified since we are disabling `update_curator_group_permissions`
* Rebased on latest `carthage` with types generated

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476251168310) by [Unito](https://www.unito.io)


mnaamani (changes requested):
We should definitely not `panic!` in the runtime, and the `todo!` macro [panics](https://doc.rust-lang.org/std/macro.todo.html)

Perhaps a safer strategy is to immediately return `Ok(())` ? clippy will probably complain about unreachable code, so we can tell it to ignore for those transactions.

Now for the tests that are `#[ignore]`d you still commented out code, I assume that is to still be able to test by passing `--ignored` flag to cargo test?

I was going to suggest using cargo features but that will just complicate the build process..

Maybe a cleaner approach is to have a duplicate test function with the modification and keep original one ignored.

But not sure if we can do something similar for mocks/fixtures.

Finally I see changes in bounty pallet, have we already merged bounty v2 pallet into carthage?

runtime-modules/bounty/src/lib.rs
```diff
@@ -775,6 +775,7 @@ decl_module! {
         /// # </weight>
         #[weight = Module::<T>::create_bounty_weight(&params, &metadata)]
         pub fn create_bounty(origin, params: BountyCreationParameters<T>, metadata: Vec<u8>) {
+            todo!();
```
Lezek123: Now with the bounty module disabled in the runtime, I don't think we need this anymore

runtime-modules/content/src/lib.rs
```diff
@@ -492,7 +494,7 @@ decl_module! {
                 collaborators: params.collaborators.clone(),
                 cumulative_reward_claimed: BalanceOf::<T>::zero(),
                 transfer_status: ChannelTransferStatus::NoActiveTransfer,
-                privilege_level: Zero::zero(),
+                privilege_level: Zero::zero(), // start with lowes privilege level
```
```suggestion
                privilege_level: Zero::zero(),
```
The default `privilege_level` (`0`) is not necessarily the lowest, by itself it's just a meaningless number, it's up to the content working group leader to define how given level corresponds to curator groups' permissions.

runtime-modules/content/src/lib.rs
```diff
@@ -314,29 +314,31 @@ decl_module! {
...
+            _permissions_by_level: ModerationPermissionsByLevel<T>
         ) {
-            let sender = ensure_signed(origin)?;
-            // Ensure given origin is lead
-            ensure_lead_auth_success::<T>(&sender)?;
-            // Ensure curator group under provided curator_group_id already exist
-            Self::ensure_curator_group_under_given_id_exists(&curator_group_id)?;
-            // Ensure permissions_by_level map max. allowed size is not exceeded
-            Self::ensure_permissions_by_level_map_size_not_exceeded(&permissions_by_level)?;
+            todo!()
```
Lezek123: We should use a custom macro that doesn't panic the runtime

runtime-modules/content/src/lib.rs
```diff
@@ -2351,72 +2359,77 @@ decl_module! {
...
+
+            // Self::deposit_event(
+            //     RawEvent::UpdateChannelTransferStatus(channel_id, actor, new_transfer_status)
+            // );
         }
 
         /// Accepts channel transfer.
         /// `commitment_params` is required to prevent changing the transfer conditions.
         #[weight = 10_000_000] // TODO: adjust weight
         pub fn accept_channel_transfer(
```
Lezek123: I don't think there's a need to disable this one. If you disable `update_channel_transfer_status` then `accept_channel_transfer` will always throw an error, so there is no way `ChannelTransferAccepted` event will be emitted

runtime-modules/content/src/lib.rs
```diff
@@ -3187,7 +3204,7 @@ impl<T: Config> Module<T> {
             .collect::<BTreeSet<_>>()
     }
 
-    fn ensure_sufficient_balance_for_channel_transfer(
+    fn _ensure_sufficient_balance_for_channel_transfer(
```
Lezek123: I noticed that using `_` prefix for some functions forces you to comment-out some other functions, like `validate_channel_transfer_acceptance` completely (because they are calling the now-prefixed function by it's original name).

This seems to add some unnecessary complexity. In order to minimize the amount of changes (like commenting-out etc.) I think just going for:
```
    #[allow(dead_code)] // TODO: Remove the `allow` attribute after carthage
    fn some_no_longer_called_function(
```
Will allow you to temporarily silence the clippy warnings for no longer used functions and you won't need to prefix/comment-out any additional ones

runtime-modules/content/src/lib.rs
```diff
@@ -3340,21 +3360,23 @@ impl<T: Config> Module<T> {
...
-                    channel.daily_nft_limit.limit = limit;
-                });
-            }
-            NftLimitId::ChannelWeekly(channel_id) => {
-                ChannelById::<T>::mutate(channel_id, |channel| {
-                    channel.weekly_nft_limit.limit = limit;
-                });
-            }
-        }
+    pub(crate) fn set_nft_limit(_limit_id: NftLimitId<T::ChannelId>, _limit: u64) {
```
Lezek123: This function is not used due to other related parts of the code being commented out, so I think you can just mark it with
```rust
    #[allow(dead_code)] // TODO: Remove the `allow` attribute after carthage
    pub(crate) fn set_nft_limit(limit_id: NftLimitId<T::ChannelId>, limit: u64) {
```
and leave the body intact

runtime-modules/content/src/lib.rs
```diff
@@ -3475,6 +3497,14 @@ decl_event!(
...
             TransferParameters<<T as common::MembershipTypes>::MemberId, BalanceOf<T>>,
         AccountId = <T as frame_system::Config>::AccountId,
+        // TODO: enable after Carthage
+        // ChannelTransferStatus = ChannelTransferStatus<
+        //     <T as common::MembershipTypes>::MemberId,
+        //     <T as ContentActorAuthenticator>::CuratorGroupId,
+        //     BalanceOf<T>,
+        // >,
+        // TransferParameters =
+        //     TransferParameters<<T as common::MembershipTypes>::MemberId, BalanceOf<T>>,
```
Lezek123: This is a commented-out duplicate of the code that's just 2 lines above it, I think it's safe to just remove it

runtime-modules/content/src/tests/channels.rs
```diff
@@ -300,6 +300,7 @@ fn unsuccessful_channel_update_with_invalid_channel_id() {
     })
 }
 
+#[ignore]
```
Lezek123: Missing
```
// TODO: enable after...
```
comment

runtime-modules/content/src/tests/channels.rs
```diff
@@ -2531,6 +2556,8 @@ fn unsuccessful_nft_owner_remark_by_curator_agent_without_permissions() {
     })
 }
 
+// TODO: enable after enabling u
```
Lezek123: Broken comment

runtime-modules/content/src/tests/channels.rs
```diff
@@ -2804,6 +2835,8 @@ fn unsuccessful_nft_owner_remark_by_collaborator_without_permissions() {
...
```
Lezek123: Broken comment

runtime-modules/content/src/tests/channels.rs
```diff
@@ -2820,6 +2853,8 @@ fn unsuccessful_nft_destruction_by_collaborator_without_permissions() {
     })
 }
 
+// TODO: enable after enabling destroy_nft
+#[ignore]
 #[test]
 fn succesfull_nft_management_actions_by_collaborator() {
```
Lezek123: I think it's better to just slightly modify `SuccessfulNftManagementFlow`:
1. Add another `CreateVideoFixture` before `UpdateVideoFixture`
2. Add `with_video_id(3)` to `UpdateVideoFixture`

After that the flow should be runnable.
It would be a shame to sacriface the entrie test case only because of `destroy_nft`

runtime-modules/content/src/tests/channels.rs
```diff
@@ -2837,6 +2872,8 @@ fn succesfull_nft_management_actions_by_collaborator() {
     })
 }
 
+// TODO: enable after enabling destroy_nft
+#[ignore]
 #[test]
 fn succesfull_nft_management_actions_by_owner_member() {
```
Lezek123: Same as above

runtime-modules/content/src/tests/channels.rs
```diff
@@ -2940,6 +2977,7 @@ fn successful_channel_agent_remark_by_owner_member() {
 
 // Curator channels
 
+#[ignore]
```
Lezek123: Missing:
```
// TODO: enable after
```
comments (also for all tests below related to the transfers)

runtime-modules/content/src/tests/creator_tokens/claim_patronage_credit.rs
```diff
@@ -81,6 +81,7 @@ fn successful_claim_member_channel_creator_token_patronage_credit_by_owner() {
     })
 }
 
+#[ignore]
```
Lezek123: Missing:
```
// TODO: enable after...
```
comments (also the case for most of the similar tests related to channel transfers blocked features)

dobertRowneySr:
- I replaced `todo!()` with an error message where it was possible.
- disabled proposals are commented out
- merged current carthage with metadata & types generated

runtime-modules/content/src/tests/fixtures.rs
```diff
@@ -4405,7 +4407,7 @@ impl SuccessfulChannelCollaboratorsManagementFlow {
 }
 
 // helper functions
-pub fn assert_group_has_permissions_for_actions(
+pub fn __assert_group_has_permissions_for_actions(
```
Lezek123: A suggested before:
```suggestion
#[allow(dead_code)] // TODO: Remove the `allow` attribute after carthage
pub fn assert_group_has_permissions_for_actions(
```

runtime-modules/content/src/tests/nft/destroy_nft.rs
```diff
@@ -4,6 +4,7 @@ use crate::tests::mock::*;
 use crate::*;
 use frame_support::{assert_err, assert_noop, assert_ok};
 
+#[ignore]
```
Lezek123: Missing:
```
// TODO: enable after...
```
comments (also for all other `destroy_nft`-related tests)

runtime-modules/content/src/tests/nft/destroy_nft.rs
```diff
@@ -36,14 +37,20 @@ fn destroy_nft() {
...
-            ContentActor::Member(DEFAULT_MEMBER_ID),
-            video_id,
-        ));
+        // assert_event(
+        //     MetaEvent::content(RawEvent::NftDestroyed(
+        //         ContentActor::Member(DEFAULT_MEMBER_ID),
+        //         video_id,
+        //     )),
+        //     number_of_events_before_call + 1,
+        // );
```
Lezek123: The code was changed here (`last_event_eq` => `assert_event`).
I suppose `last_event_eq` is the newer approach?

runtime-modules/content/src/tests/nft/issue_nft.rs
```diff
@@ -267,6 +267,7 @@ fn issue_nft_fails_with_invalid_open_auction_parameters() {
     })
 }
 
+#[ignore]
```
Lezek123: Missing:
```
// TODO: enable after...
```
comments (also for all other nft_limits-related tests)

runtime-modules/content/src/types.rs
```diff
@@ -313,7 +313,7 @@ pub struct TransferParameters<MemberId: Ord, Balance: Zero> {
 
 impl<MemberId: Ord, Balance: Zero> TransferParameters<MemberId, Balance> {
     // Defines whether the transfer is free.
-    pub(crate) fn is_free_of_charge(&self) -> bool {
+    pub(crate) fn _is_free_of_charge(&self) -> bool {
```
Lezek123: Avoid prefixing as suggested before

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -251,10 +251,11 @@ pub trait Config:
...
     type VetoProposalProposalParameters: Get<ProposalParameters<Self::BlockNumber, BalanceOf<Self>>>;
 
+    // TODO: enable after Carthage
     /// `Update Nft limit` proposal parameters
-    type UpdateGlobalNftLimitProposalParameters: Get<
-        ProposalParameters<Self::BlockNumber, BalanceOf<Self>>,
-    >;
+    // type UpdateGlobalNftLimitProposalParameters: Get<
+    //     ProposalParameters<Self::BlockNumber, BalanceOf<Self>>,
+    // >;
```
Lezek123: As discussed with Bedeho, I think we should go for validation failing approach instead to minimalize commenting-out

bedeho:
> I replaced todo!() with an error message where it was possible.

Does this mean you still have some left?

runtime-modules/content/src/tests/curators.rs
```diff
@@ -71,93 +71,94 @@ fn curator_group_management() {
...
+        // // Check CuratorGroupPermissionsUpdated event
+        // assert_eq!(
+        //     System::events().last().unwrap().event,
+        //     MetaEvent::content(RawEvent::CuratorGroupPermissionsUpdated(
+        //         curator_group_id,
+        //         permissions.clone()
+        //     ))
+        // );
 
         // Validate permissions
```
Lezek123: Comments should also be commented-out:
```suggestion
        // // Validate permissions
```

runtime-modules/content/src/tests/curators.rs
```diff
@@ -71,93 +71,94 @@ fn curator_group_management() {
...
-        let group = Content::curator_group_by_id(curator_group_id);
-        assert_eq!(group.get_permissions_by_level().len(), 3);
+        // let group = Content::curator_group_by_id(curator_group_id);
+        // assert_eq!(group.get_permissions_by_level().len(), 3);
         // Iterate over privilege levels from 0 to 3
         // (3 will be a "non-existent map entry" case)
```
Lezek123: Comments should also be commented-out:
```suggestion
        // // Iterate over privilege levels from 0 to 3
        // // (3 will be a "non-existent map entry" case)
```

runtime-modules/content/src/tests/curators.rs
```diff
@@ -71,93 +71,94 @@ fn curator_group_management() {
...
-            System::events().last().unwrap().event,
-            Event::Content(RawEvent::CuratorGroupPermissionsUpdated(
-                curator_group_id,
-                permissions.clone()
-            ))
-        );
+        // // Check CuratorGroupPermissionsUpdated event
+        // assert_eq!(
+        //     System::events().last().unwrap().event,
+        //     MetaEvent::content(RawEvent::CuratorGroupPermissionsUpdated(
```
Lezek123: The code was changed here.
Original version:
```rust
            Event::Content(RawEvent::CuratorGroupPermissionsUpdated(
```
Commented-out version:
```rust
            MetaEvent::content(RawEvent::CuratorGroupPermissionsUpdated(
```

runtime-modules/content/src/tests/curators.rs
```diff
@@ -1,7 +1,7 @@
 #![cfg(test)]
 
 use std::collections::BTreeMap;
-use std::iter::FromIterator;
+//use std::iter::FromIterator;
```
Lezek123: I think it would be a good idea to use:
```
TODO: enable after...
```
Also above the commented-out `use` statements

dobertRowneySr:
> > I replaced todo!() with an error message where it was possible.
> 
> Does this mean you still have some left?

No there are no todos any more, the only case that I could not replace in with a `Error` is on `set_nft_limits` internal
helper function which is now a no-op since it doesn't return anything

runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -1933,72 +1933,74 @@ fn create_set_referral_cut_common_checks_succeed() {
...
-            },
-            proposal_parameters:
-                <Test as crate::Config>::UpdateGlobalNftLimitProposalParameters::get(),
-        };
-        proposal_fixture.check_all();
-    });
-}
+// TODO: enable after Carthage
+// #[test]
+// fn create_update_global_nft_limit_proposal_common_checks_succeed() {
```
Lezek123: Use the `#[ignore]` approach instead of commenting-out for consistency.
Also the commented-out code seems to have some differences (`Trait` instead of `Config`, `ProposalCodex` instead of `ProposalsCodex` etc.)

Lezek123 (changes requested):

runtime-modules/content/src/lib.rs
dobertRowneySr: I replaced all `todo!()` with an `Error::<T>::FeatureNotImplemented`

runtime-modules/proposals/codex/src/lib.rs
dobertRowneySr: As I discussed with Mokhtar we should simply comment out Proposals system. The extra effort to add enabling / disabling logic is not worth anymore, since it is probably the last time we are disabling proposals.

runtime-modules/proposals/codex/src/tests/mod.rs
dobertRowneySr: I need to comment-out this one if I want to comment out proposals, see previous comment

runtime-modules/proposals/codex/src/lib.rs
Lezek123: Correct me if I'm wrong @ignazio-bovo @mnaamani, but I think it would be just a simple change in `ensure_details_checks` from:
```rust
            ProposalDetails::UpdateGlobalNftLimit(..) => {
                // Note: No checks for this proposal for now
            }
```
To:
```rust
            ProposalDetails::UpdateGlobalNftLimit(..) => {
                 ensure!(false, Error::<T>::FeatureNotImplemented) // TODO: Enable after Carthage
            }
```

runtime-modules/content/src/lib.rs
```diff
@@ -3518,9 +3540,15 @@ decl_event!(
             <T as ContentActorAuthenticator>::CuratorGroupId,
             BalanceOf<T>,
         >,
+        AccountId = <T as frame_system::Config>::AccountId,
+        // TODO: enable after Carthage
+        // ChannelTransferStatus = ChannelTransferStatus<
+        //     <T as common::MembershipTypes>::MemberId,
+        //     <T as ContentActorAuthenticator>::CuratorGroupId,
+        //     BalanceOf<T>,
+        // >,
```
Lezek123: There's no point in adding this comment, since there's already an uncommented version of this code above (line 3538)

runtime-modules/content/src/lib.rs
```diff
@@ -3387,21 +3407,23 @@ impl<T: Config> Module<T> {
...
-            }
-            NftLimitId::ChannelWeekly(channel_id) => {
-                ChannelById::<T>::mutate(channel_id, |channel| {
-                    channel.weekly_nft_limit.limit = limit;
-                });
-            }
-        }
+    #[allow(dead_code)] // TODO: Remove the `allow` attribute after Carthage
+    pub(crate) fn set_nft_limit(_limit_id: NftLimitId<T::ChannelId>, _limit: u64) {
+        // TODO: enable after Carthage
```
Lezek123: As mentioned during the previous review, this function is never used now, so you can uncomment the body in order to limit the amount of commented-out code

runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -1933,72 +1934,71 @@ fn create_set_referral_cut_common_checks_succeed() {
...
+//                 )
+//             },
+//             successful_call: || {
+//                 ProposalCodex::create_proposal(
+//                     RawOrigin::Signed(1).into(),
+//                     general_proposal_parameters.clone(),
+//                     proposal_details.clone(),
+//                 )
+//             },
+//         };
```
Lezek123: The commented-out version is now missing:
```rust
            proposal_parameters:
                <Test as crate::Config>::UpdateGlobalNftLimitProposalParameters::get(),
```

runtime-modules/bounty/src/lib.rs
```diff
@@ -775,41 +778,43 @@ decl_module! {
...
-                milestone: created_bounty_milestone,
-                active_work_entry_count: 0,
-            };
-
-            <Bounties<T>>::insert(bounty_id, bounty);
-            BountyCount::mutate(|count| {
-                *count = next_bounty_count_value
-            });
-            Self::deposit_event(RawEvent::BountyCreated(bounty_id, params, metadata));
+            return Err(Error::<T>::FeatureNotImplemented.into());
```
Lezek123: As mentioned in the previous review - I'm not sure why would you need to touch the `bounty` module, which is excluded from the `construct_runtime` in `runtime/src/lib.rs`

runtime-modules/content/src/lib.rs
```diff
@@ -3338,40 +3357,41 @@ impl<T: Config> Module<T> {
...
+        //     &channel.weekly_nft_counter,
+        //     Error::<T>::ChannelNftWeeklyLimitExceeded,
+        // )?;
 
         Ok(())
     }
 
     // Checks generic NFT-limit.
-    fn check_generic_nft_limit(
+    fn _check_generic_nft_limit(
```
Lezek123: Avoid using `_` the prefix, as mentioned in the previous review

runtime-modules/content/src/tests/channels.rs
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/channels.rs
```diff
@@ -3110,6 +3151,7 @@ fn claim_council_reward_succeeded() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/creator_tokens/claim_patronage_credit.rs
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/creator_tokens/init_token_sale.rs
```diff
@@ -120,6 +120,7 @@ fn successful_init_curator_channel_creator_token_sale_by_lead() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/creator_tokens/issue_revenue_split.rs
```diff
@@ -149,6 +149,7 @@ fn successful_issue_curator_channel_revenue_split_by_lead() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/creator_tokens/make_permissionless.rs
```diff
@@ -146,6 +146,7 @@ fn successful_make_curator_channel_creator_token_permissionless_by_lead() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/creator_tokens/reduce_patronage_rate.rs
```diff
@@ -119,6 +119,7 @@ fn successful_reduce_curator_channel_creator_token_patronage_rate_by_lead() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/curators.rs
Lezek123: Previous review comment still not addressed

runtime-modules/content/src/tests/fixtures.rs
```diff
@@ -3085,13 +3085,14 @@ impl AcceptChannelTransferFixture {
...
-                ))
-            );
+            // TODO: enable after Carthage
+            // assert_eq!(
+            //     System::events().last().unwrap().event,
+            //     MetaEvent::Content(RawEvent::ChannelTransferAccepted(
+            //         self.channel_id,
+            //         self.params.clone()
+            //     ))
+            // );
```
Lezek123: There's no need to comment this out

runtime-modules/content/src/tests/merkle.rs
```diff
@@ -389,6 +389,7 @@ fn unsuccessful_channel_balance_withdrawal_when_amount_exceeds_balance_minus_exi
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/metaprotocol.rs
```diff
@@ -362,6 +362,7 @@ fn unsuccessful_nft_owner_with_invalid_acount() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/metaprotocol.rs
```diff
@@ -385,6 +386,7 @@ fn unsuccessful_nft_rewark_during_transfer() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/metaprotocol.rs
```diff
@@ -406,6 +408,7 @@ fn unsuccessful_channel_owner_remark_during_transfer() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/cancel_buy_now.rs
```diff
@@ -216,6 +216,7 @@ fn cancel_buy_now_not_in_auction_state() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/cancel_nft_auction.rs
```diff
@@ -284,6 +284,7 @@ fn cancel_nft_auction_english_auction_with_bids() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/cancel_nft_auction.rs
```diff
@@ -306,6 +307,7 @@ fn cancel_open_auction_fails_during_channel_transfer() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/cancel_offer.rs
```diff
@@ -217,6 +217,7 @@ fn cancel_offer_not_in_auction_state() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/offer_nft.rs
```diff
@@ -216,6 +216,7 @@ fn offer_nft_transactional_status_is_not_idle() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/pick_open_auction_winner.rs
```diff
@@ -590,6 +590,7 @@ fn pick_open_auction_ok_with_channel_owner_correctly_credited() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/sling_nft_back.rs
```diff
@@ -211,6 +211,7 @@ fn sling_nft_back_transactional_status_is_not_idle() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/start_nft_auction.rs
```diff
@@ -579,6 +579,7 @@ fn start_eng_auction_fails_with_invalid_forward_starting() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/start_nft_auction.rs
```diff
@@ -601,6 +602,7 @@ fn start_open_auction_fails_during_channel_transfer() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/update_buy_now.rs
```diff
@@ -231,6 +231,7 @@ fn update_buy_now_price_not_in_auction_state() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/update_nft_limit.rs
```diff
@@ -7,6 +7,7 @@ use frame_system::RawOrigin;
 use sp_std::collections::btree_map::BTreeMap;
 use sp_std::iter::FromIterator;
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/update_nft_limit.rs
```diff
@@ -33,6 +34,7 @@ fn update_nft_limits_works_as_expected_for_global_limits() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/nft/update_nft_limit.rs
```diff
@@ -142,6 +144,7 @@ fn update_channel_nft_limit_test_helper(
 }
 
 // channel creation tests
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -68,6 +68,7 @@ fn successful_video_creation_by_member() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -486,6 +487,7 @@ fn unsuccessful_video_creation_due_to_bucket_having_insufficient_objects_number_
...
```
Lezek123: I think this test shouldn't actually be ignored

runtime-modules/content/src/tests/videos.rs
```diff
@@ -635,6 +637,7 @@ fn successful_video_update_by_member_with_assets_removal() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1049,6 +1053,7 @@ fn successful_video_deletion_by_member_with_assets_removal() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -954,6 +957,7 @@ fn unsuccessful_video_update_due_to_bucket_having_insufficient_objects_number_le
...
```
Lezek123: I think this test shouldn't actually be ignored

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1315,6 +1321,7 @@ fn unsuccessful_video_update_with_nft_issuance_when_nft_already_issued() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1325,6 +1332,7 @@ fn create_video_failed_with_exceeded_global_daily_nft_limits() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1335,6 +1343,7 @@ fn create_video_failed_with_exceeded_global_weekly_nft_limits() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1345,6 +1354,8 @@ fn create_video_failed_with_exceeded_channel_daily_nft_limits() {
         );
     })
 }
+
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1382,6 +1393,7 @@ fn nft_test_helper_for_exceeded_limit_on_creating_video(
         .call_and_assert(Err(expected_error.into()));
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1392,6 +1404,7 @@ fn update_video_failed_with_exceeded_global_daily_nft_limits() {
     })
 }
 
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1401,6 +1414,8 @@ fn update_video_failed_with_exceeded_global_weekly_nft_limits() {
         );
     })
 }
+
+#[ignore]
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/content/src/tests/videos.rs
```diff
@@ -1411,6 +1426,8 @@ fn update_video_failed_with_exceeded_channel_daily_nft_limits() {
...
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/proposals/codex/src/benchmarking.rs
```diff
@@ -8,8 +8,8 @@ use balances::Pallet as Balances;
 use codec::Decode;
 use common::working_group::WorkingGroup;
 use common::BalanceKind;
-use content::NftLimitPeriod;
-use frame_benchmarking::{account, benchmarks};
+//use content::NftLimitPeriod;
```
Lezek123: `// TODO: enable after...` still missing here

runtime-modules/proposals/codex/src/types.rs
```diff
@@ -10,7 +10,7 @@ use common::working_group::WorkingGroup;
 use common::BalanceKind;
 use common::FundingRequestParameters;
 
-use content::NftLimitPeriod;
+//use content::NftLimitPeriod;
```
Lezek123: `// TODO: enable after...` still missing here

Lezek123 (changes requested):

runtime-modules/proposals/codex/src/lib.rs
Lezek123: Tell me what do you think about this. Just by looking at the amount of code that needs to be commented-out in order to disable this single proposal, I think it's a way simpler and cleaner approach to make use of `ensure_details_checks`

runtime-modules/content/src/tests/curators.rs
```diff
@@ -139,7 +139,7 @@ fn curator_group_management() {
         // // Check CuratorGroupPermissionsUpdated event
         // assert_eq!(
         //     System::events().last().unwrap().event,
-        //     MetaEvent::content(RawEvent::CuratorGroupPermissionsUpdated(
+        //     MetaEvent::Content(RawEvent::CuratorGroupPermissionsUpdated(
```
Lezek123:
```suggestion
        //     Event::Content(RawEvent::CuratorGroupPermissionsUpdated(
```

runtime-modules/content/src/tests/nft/cancel_nft_auction.rs
Lezek123:
```suggestion
// TODO: Enable after Carthage
#[ignore]
```

runtime-modules/content/src/tests/nft/cancel_nft_auction.rs
Lezek123:
```suggestion
// TODO: Enable after Carthage
#[ignore]
```

runtime-modules/content/src/tests/videos.rs
```diff
@@ -486,6 +488,8 @@ fn unsuccessful_video_creation_due_to_bucket_having_insufficient_objects_number_
     })
 }
 
+// TODO: Enable after Carthage
+#[ignore]
 #[test]
 fn unsuccessful_video_creation_with_max_object_size_limits_exceeded() {
```
Lezek123:
```suggestion
#[test]
fn unsuccessful_video_creation_with_max_object_size_limits_exceeded() {
```

runtime-modules/content/src/tests/videos.rs
```diff
@@ -954,6 +960,8 @@ fn unsuccessful_video_update_due_to_bucket_having_insufficient_objects_number_le
     })
 }
 
+// TODO: Enable after Carthage
+#[ignore]
 #[test]
 fn unsuccessful_video_update_with_max_object_size_limits_exceeded() {
```
Lezek123:
```suggestion
#[test]
fn unsuccessful_video_update_with_max_object_size_limits_exceeded() {
```

runtime-modules/content/src/lib.rs
```diff
@@ -3367,41 +3385,40 @@ impl<T: Config> Module<T> {
...
+        //     &channel.weekly_nft_limit,
+        //     &channel.weekly_nft_counter,
+        //     Error::<T>::ChannelNftWeeklyLimitExceeded,
+        // )?;
         Ok(())
     }
 
     // Checks generic NFT-limit.
-    fn check_generic_nft_limit(
+    fn _check_generic_nft_limit(
```
```suggestion
    // Checks generic NFT-limit.
    #[allow(dead_code)] // TODO: Remove the `allow` attribute after Carthage
    fn check_generic_nft_limit(
```

Lezek123 (changes requested):

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -231,10 +231,11 @@ pub trait Config:
     /// `Veto Proposal` proposal parameters
     type VetoProposalProposalParameters: Get<ProposalParameters<Self::BlockNumber, BalanceOf<Self>>>;
 
-    /// `Update Nft limit` proposal parameters
-    type UpdateGlobalNftLimitProposalParameters: Get<
-        ProposalParameters<Self::BlockNumber, BalanceOf<Self>>,
-    >;
+    // TODO: enable after Carthage
+    // `Update Nft limit` proposal parameters
```
Lezek123:
```suggestion
    // /// `Update Nft limit` proposal parameters
```

runtime-modules/content/src/lib.rs
```diff
@@ -3368,40 +3386,40 @@ impl<T: Config> Module<T> {
...
-            )?;
-        }
+    fn check_nft_limits(_channel: &Channel<T>) -> DispatchResult {
+        // TODO: enable after Carthage
+        // // Global daily limit.
+        // Self::check_generic_nft_limit(
+        //     &Self::global_daily_nft_limit(),
+        //     &Self::global_daily_nft_counter(),
+        //     Error::<T>::GlobalNftDailyLimitExceeded,
+        // )?;
```
Lezek123: The commented-out code is outdated. Missing:
```rust
 if Self::nft_limits_enabled() {
```

runtime-modules/proposals/codex/src/benchmarking.rs
```diff
@@ -951,12 +953,13 @@ mod tests {
...
-    fn test_update_global_nft_limit_proposal() {
-        initial_test_ext().execute_with(|| {
-            assert_ok!(ProposalsCodex::test_benchmark_create_proposal_update_global_nft_limit());
-        })
-    }
+    // TODO: enable after Carthage
+    // #[test]
+    // fn test_update_global_nft_limit_proposal() {
+    //     initial_test_ext().execute_with(|| {
+    //         assert_ok!(test_benchmark_create_proposal_update_global_nft_limit::<Test>());
```
Lezek123: The commented-out code is outdated
```suggestion
    //         assert_ok!(ProposalsCodex::test_benchmark_create_proposal_update_global_nft_limit());
```
(that's also the case for `test_update_channel_payouts_proposal`)

runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -1931,72 +1932,74 @@ fn create_set_referral_cut_common_checks_succeed() {
...
-                <Test as crate::Config>::UpdateGlobalNftLimitProposalParameters::get(),
-        };
-        proposal_fixture.check_all();
-    });
-}
+//         let proposal_fixture = ProposalTestFixture {
+//             general_proposal_parameters: general_proposal_parameters.clone(),
+//             proposal_details: proposal_details.clone(),
+//             insufficient_rights_call: || {
+//                 ProposalCodex::create_proposal(
```
Lezek123: The commented-out code here is outdated (ie. `ProposalCodex` instead of `ProposalsCodex` etc.)

Lezek123 (changes requested):
1. `cargo fmt` needed (the CI check is failing because of it: https://github.com/Joystream/joystream/runs/7228248731?check_suite_focus=true)
2. Although this was introduced in one of the previous PRs, if I uncomment `create_proposal_update_channel_payouts` in `runtime-modules/proposals/codex/src/benchmarking.rs` I'm getting some errors:
    ```
    error: expected one of `:`, `;`, `=`, `@`, or `|`, found keyword `in`
       --> /home/leszek/projects/joystream/joystream-ws-4/runtime-modules/proposals/codex/src/benchmarking.rs:744:15
        |
    744 |         let t in ...;
        |               ^^ expected one of `:`, `;`, `=`, `@`, or `|`

    error[E0599]: no function or associated item named `default` found for associated type `<T as frame_system::Config>::AccountId` in the current scope
       --> /home/leszek/projects/joystream/joystream-ws-4/runtime-modules/proposals/codex/src/benchmarking.rs:753:45
        |
    753 |             uploader_account: T::AccountId::default(),
        |                                             ^^^^^^^ function or associated item not found in `<T as frame_system::Config>::AccountId`
  
    error[E0599]: no function or associated item named `from_iter` found for struct `frame_benchmarking::Vec<_, _>` in the current scope
       --> /home/leszek/projects/joystream/joystream-ws-4/runtime-modules/proposals/codex/src/benchmarking.rs:756:39
        |
    756 |                 ipfs_content_id: Vec::from_iter((0..i).map(|v| u8::MAX))
        |                                       ^^^^^^^^^ function or associated item not found in `frame_benchmarking::Vec<_, _>`

    ```

3. The commented-out code in `runtime-modules/proposals/codex/src/tests/mod.rs` also seems outdated and generated some errors when uncommented (just pasting a few of those):

    ```
    2081 |                 <Test as crate::Trait>::UpdateChannelPayoutsProposalParameters::get(),
         |                                 ^^^^^ could not find `Trait` in the crate root
    
    error[E0433]: failed to resolve: use of undeclared type `ProposalCodex`
        --> runtime-modules/proposals/codex/src/tests/mod.rs:2109:13
         |
    2109 |             ProposalCodex::create_proposal(
         |             ^^^^^^^^^^^^^ use of undeclared type `ProposalCodex`
    
    error[E0599]: no function or associated item named `from_iter` found for struct `frame_benchmarking::Vec<_, _>` in the current scope
        --> runtime-modules/proposals/codex/src/tests/mod.rs:2037:47
         |
    2037 |                         ipfs_content_id: Vec::from_iter((0..46).map(|_| u8::MAX)),
         |                                               ^^^^^^^^^ function or associated item not found in `frame_benchmarking::Vec<_, _>`
    ```

5. I already asked this in the previous review, but seems like my comment got lost, so let me bring that up again @ignazio-bovo @mnaamani : There's a significant amount of code commented out that has to do with proposals, proposals_integration, proposals benchmarking etc. The commented-out code has a tendency to become outdated very quickly (as can be observed in this PR). What's the advantage of commenting out this code over just modifying `ensure_details_check` in `/runtime-modules/proposals/codex/src/lib.rs` with something like:
```rust
            ProposalDetails::UpdateGlobalNftLimit(..) => {
                // TODO: Enable after Carthage
                ensure!(
                    false
                    Error::<T>::ProposalNotImplemented
                );
            }
```

dobertRowneySr:
As suggested by Mokhtar I added an extra trait in order to enable disable proposals in mock and runtime setup.
--> No longer commented-out code

runtime-modules/content/src/lib.rs
```diff
@@ -2513,6 +2515,10 @@ decl_module! {
             nft_limit_period: NftLimitPeriod,
             limit: u64,
         ) {
+            if !Self::nft_limits_enabled() {
+                return Ok(());
+            }
```
Lezek123: It seems risky because:
1. It returns `Ok`, which could be misleading from the user perspective, because it will look like everything went ok with the proposal execution, but the limits still didn't get updated for some mysterious reason
2. It doesnt have `//TODO` comment to suggest it's only a temporary solution
3. I think there's a valid use-case for updating the limits while they are still disabled and only enabling them after they're already set to correct values

I think it's not actually necessary anyway, since the proposal is going to be disabled

runtime-modules/content/src/lib.rs
```diff
@@ -2538,6 +2544,10 @@ decl_module! {
             channel_id: T::ChannelId,
             limit: u64,
         ) {
+            if !Self::nft_limits_enabled() {
+                return Ok(());
+            }
```
Lezek123: Same issue as in the comment above

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -438,6 +441,11 @@ decl_module! {
             general_proposal_parameters: GeneralProposalParameters<T>,
             proposal_details: ProposalDetailsOf<T>,
         ) {
+            // no op in case proposal is not enabled
+            if !T::ProposalsEnabled::is_proposal_enabled(&proposal_details) {
+                return Ok(());
+            }
```
Lezek123: I think in that case it should return and error like `ProposalDisabled` / `ProposalNotImplemented`.
Since you created a configurable trait, you can just enable all proposals conditionally if the runtime is built with `--features runtime-benchmarks` to avoid breaking the benchmarks

runtime-modules/content/src/lib.rs
```diff
@@ -2427,6 +2435,8 @@ decl_module! {
             actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
             transfer_params: InitTransferParametersOf<T>,
         ) {
+            return Err(Error::<T>::FeatureNotImplemented.into());
```
Lezek123: Missing `TODO` comment

Also - other extrinsics (like `destroy_nft`, `toggle_nft_limits`) are commented-out, which gave me an idea:

Why don't we just disable all extrinsics by only using:
```rust
        #[allow(unused_variables)] // TODO: Enable after carthage and remove the `allow` attribute
        pub fn initialize_channel_transfer(
```
Instead of commenting out the body of the method and prefixing the parameters with `_`.

What do you think?

runtime-modules/proposals/codex/src/tests/mod.rs
```diff
@@ -1931,72 +1932,74 @@ fn create_set_referral_cut_common_checks_succeed() {
...
-        let general_proposal_parameters_no_staking = GeneralProposalParameters::<Test> {
-            member_id: 1,
-            title: b"title".to_vec(),
-            description: b"body".to_vec(),
-            staking_account_id: None,
-            exact_execution_block: None,
-        };
+// TODO: enable after Carthage
+// #[test]
+// fn create_update_global_nft_limit_proposal_common_checks_succeed() {
```
Lezek123: Those tests can be now re-enabled I think

runtime/src/tests/proposals_integration/mod.rs
```diff
@@ -1071,41 +1071,42 @@ fn proposal_reactivation_succeeds() {
...
-
-        assert_eq!(
-            content::GlobalDailyNftLimit::<Runtime>::get().limit,
-            new_limit
-        );
-    });
-}
+// TODO: Enable after Carthage
+// #[test]
+// fn update_global_nft_limit_proposal_succeeds() {
```
Lezek123: This can be now re-enabled

Lezek123 (changes requested):

runtime-modules/bounty/src/lib.rs
```diff
@@ -623,6 +623,9 @@ decl_event! {
 decl_error! {
     /// Bounty pallet predefined errors
     pub enum Error for Module<T: Config> {
+        /// Feature not implemented error
+        FeatureNotImplemented,
```
mnaamani: This new error variant is not used as far as I can see in the bounty pallet.

runtime-modules/content/src/lib.rs
mnaamani: Even better approach would be filtering calls in runtime config, like https://github.com/Joystream/joystream/pull/3986
We can use cargo `features` to have different filtering for production and tests/benchmarking

mnaamani (changes requested):
I like that we implemented the proposal selector approach, much cleaner this way.
I suggest to also use the base call filter like we are planning for carthage launch to disable the extrinsics. This way you can have less commented out code and you do not have to ignore so many tests. https://github.com/Joystream/joystream/pull/3986

runtime-modules/content/src/lib.rs
dobertRowneySr: ~I will proceed with the recommended approach after #3986 is merged, so that I can leverage the provided trait
For the scope of this PR I used Leszek's suggestion~.  
*Update*  
For disabled content calls I have used filtered calls in runtime config as done by Mokhtar in #3986

dobertRowneySr:
I have used Mokhtar's filtered call approach #3986 to disable `content` extrinsics.
I added conditional compilation options for getting a normal (unfrozen) runtime for benchmarks

Lezek123 (changes requested):
1. In `runtime/src/lib.rs` the `BaseCallFilter` was still set to `frame_support::traits::Everything`. After changing it to `LockedDownBaseFilter` I could confirm that the calls are indeed filtered out (throwing `system.CallFiltered` error)
1. I noticed that the tests are not really affected by the implementation of `BaseCallFilter` in `runtime/src/lib.rs`, so it was possible to just remove all `#[ignore]` parameters and they are still passing.
1. I also realized we no longer need `ProposalsEnabled` feature, as we can just use `BaseCallFilter` for this purpose.
1. It was also possible to now remove `#[allow(unused_variables)]` and `#[allow(dead_code)]`.
1. It was possible to uncomment some code that previously needed to be commented out.
1. `update_channel_nft_limit`, `set_channel_paused_features_as_moderator` and `initialize_channel_transfer` were missing in disabled calls. I also added `update_global_nft_limit` (even though it's executable only by root) for completeness.

**I applied all the fixes mentioned above in my sub-PR to save some time (since I already did all those changes locally for the purpose of testing): https://github.com/ignazio-bovo/joystream/pull/7**

Tests I've done locally:
- [x] `cargo build --release --features runtime-benchmarks` + generated all weights for Joystream pallets
- [x] `cargo clippy --release --features runtime-benchmarks`
- [x] `cargo clippy --release`
- [x] `cargo test --release --features runtime-benchmarks`
- [x] `cargo test --release`
- [x] `./scripts/run-dev-chain.sh` + metadata & types update + tested some of the disabled calls (`update_curator_group_permissions`, `create_proposal` with `ProposalDetails::UpdateChannelPayouts` and `create_proposal` with `ProposalDetails::UpdateGlobalNftLimit`)

Lezek123 (approved):
LGTM!

Issue #3948, opened 2022-06-24
Fix Build Issues Post Substrate v3 update


mnaamani:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202500458737184) by [Unito](https://www.unito.io)

mnaamani:
- Fix All Clippy warnings - `BUILD_DUMMY_WASM_BINARY=1 cargo +nightly-2022-05-11 clippy --release --all -- -D warnings`
- Address this https://github.com/Joystream/joystream/pull/3928#discussion_r902987924
- Look at ignored tests check with `cargo test -- --ignored`
- Update commented fee/currency tests in `runtime/src/constants.rs`
- Benchmark build issues `cargo build --release --features runtime-benchmarks` and tests `cargo test --release --features runtime-benchmarks`
- A couple of more tests failing with overflow/underflow run test without `--release` arg to observe: `cargo test`

mnaamani:
Selection of nominations is not correct in chainspec construction https://github.com/Joystream/joystream/blob/8d3a0b53ef508fc8cde08f325acf475d62889570/node/src/chain_spec/mod.rs#L196 

which results in:

```
---- chain_spec::tests::test_create_development_chain_spec stdout ----
thread 'chain_spec::tests::test_create_development_chain_spec' panicked at 'Expected Ok(_). Got Err(
    Module(
        ModuleError {
            index: 8,
            error: [
                4,
                0,
                0,
                0,
            ],
            message: Some(
                "EmptyTargets",
            ),
        },
    ),
)', /Users/mokhtar/.cargo/git/checkouts/substrate-7e08433d4c370a21/6cbe177/frame/staking/src/pallet/mod.rs:592:17
```

Issue #3923, opened 2022-06-20
Fix Referendum Benchmarks


mnaamani:
Following https://github.com/Joystream/joystream/pull/3797 Tests were updated but not the benchmark tests for referendum.
Needs some effort, there may also be a chance that the voting fixture used in the benchmarks is not voting for the expected options.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202472618939242) by [Unito](https://www.unito.io)

mnaamani:
Fixed in https://github.com/Joystream/joystream/pull/3970

PR #3970, opened 2022-06-27
Review: Fix benchmarks and weights generation

mnaamani:
Address Benchmark issues from https://github.com/Joystream/joystream/issues/3948 and https://github.com/Joystream/joystream/issues/3923
And fix integer overflow in council test.

All these should pass:
```sh
cargo test
cargo test --release --features runtime-benchmarks
cargo build --release --features runtime-benchmarks
./scripts/generate-weights.sh 1 1
# after weights are generated just check them out
# git checkout .
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202509058340833) by [Unito](https://www.unito.io)


mnaamani:
All pallet benchmarks are correctly being executed against the main runtime (not mock runtime used in tests).
I've re-organized where the weights.rs for each pallet is now stored.

I haven't been able to resolve issue with generating storage info to see more detailed information about storage accessed by the weight functions.

Pending fix is https://github.com/Joystream/joystream/issues/3923 the referendum benchmark test is only failing for the mock runtime for some reason.

mnaamani:
> All pallet benchmarks are correctly being executed against the main runtime (not mock runtime used in tests). I've re-organized where the weights.rs for each pallet is now stored.
> 
> I haven't been able to resolve issue with generating storage info to see more detailed information about storage accessed by the weight functions.
> 
> Pending fix is #3923 the referendum benchmark test is only failing for the mock runtime for some reason.

Finally resolved, it was a real pain to debug https://github.com/Joystream/joystream/pull/3970/commits/78553b278ca2c63a612e8aa9dc8012e60b103921

runtime-modules/proposals/discussion/src/lib.rs
```diff
@@ -229,7 +229,10 @@ decl_module! {
         /// - DB:
         ///    - O(1) doesn't depend on the state or parameters
         /// # </weight>
-        #[weight = WeightInfoDiscussion::<T>::add_post(0, text.len().saturated_into())]
+        #[weight = WeightInfoDiscussion::<T>::add_post(
+            T::MaxWhiteListSize::get() as u32,
```
dobertRowneySr: I did not understand where this part comes from:
The `MaxWhiteListSize` scenario can only be reached in `change_thread_mode` however here only `ensure_thread_mode` is used.

dobertRowneySr: And also here the `text` parameter is passed by move. So all the operation on it should be O(1).
Hence I am not getting the `text.len()` weight complexity parameter

dobertRowneySr (changes requested):
Looks ok, I have just left a comment about weights generation

runtime-modules/proposals/discussion/src/lib.rs
mnaamani: Looks like this was a mistake, the first arg should be `T::MaxCategoryDepth::get()`

Here is the benchmark code snippet for `add_post()`

```
add_post {

        let forum_user_id = 0;
        let caller_id =
            insert_a_leader::<T>(forum_user_id);

        let i in 1 .. (T::MaxCategoryDepth::get() + 1) as u32;

        let j in 0 .. MAX_BYTES;
...
```

mnaamani: > And also here the `text` parameter is passed by move. So all the operation on it should be O(1). Hence I am not getting the `text.len()` weight complexity parameter

My guess is the subsequent call to `add_new_post()` which computes a hash over the text, so I presume that is `O(text.len())` ? The weight component from weights.rs looks like 
`.saturating_add((2_000 as Weight).saturating_mul(j as Weight))`

mnaamani: Lol sorry I was looking at `forum::add_post()`

mnaamani: So you are right about that, the actual value (`i`) is not even used by the weight function:
```rust
fn add_post(_i: u32, j: u32, ) -> Weight {
		(74_579_000 as Weight)
			// Standard Error: 0
			.saturating_add((1_000 as Weight).saturating_mul(j as Weight))
			.saturating_add(T::DbWeight::get().reads(6 as Weight))
			.saturating_add(T::DbWeight::get().writes(4 as Weight))
	}
```

mnaamani: The benchmark should probably be re-written to remove that value as a "component".

As for the text len weight component is very tiny `.saturating_add((1_000 as Weight).saturating_mul(j as Weight))` but still `O(n)` not `O(1)`. I suppose there is still computation involved to decode the vec and copy it into the event.

mnaamani: Is this the correct fix @ignazio-bovo 81c89b8a6fdd3e7b65047c05dce22711c5005c3b

dobertRowneySr (approved):
LGTM

Issue #3901, opened 2022-06-15
Remove on-chain content categories


bedeho:
Being replaced by metaprotocol.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202453994918752) by [Unito](https://www.unito.io)

PR #3973, opened 2022-06-27
Carthage  remove content categories

dobertRowneySr:
Addresses #3901 

#### Changes
* [x] removed categories-related types and extrinsics from content runtime
* [x] rebased on latest carthage
* [x] generated types

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202526467047384) by [Unito](https://www.unito.io)


mnaamani (commented):
Looks good, only suggestion is to not check-in the yarn.lock file at this stage, its only changing warthog a dependency of the query-node. 

The types package checks are failing with error:
```
ERROR: No custom definitions found: /home/runner/work/joystream/joystream/types/src/augment/definitions.ts is not
```

I get same issue locally. Not sure why.. @Lezek123

mnaamani:
@ignazio-bovo Leszek advised the issue with failing types checks is the auto-generated code is not correctly formatted.
He will add a `yarn format` step to the build. Perhaps you can run `yarn format` in `types/` folder and check-in the updated files again.

mnaamani:
Also from Leszek: "There should be no `types/augment/augment-types.ts` file anymore. It was incorrectly checked-in again."

PR #3978, opened 2022-06-30
Fix initial nominators setup at genesis

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3948#issuecomment-1170858038

Added two initial nominators to dev chain.

Test with

```bash
cargo test --release -p joystream-node
```

you should see the development chain spec construction test pass

```bash
test chain_spec::tests::test_create_development_chain_spec ... ok
```

Started dev chain and I can see two nominators:

Screen Shot 2022-06-30 at 10 59 59 AM

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202537343383647) by [Unito](https://www.unito.io)

### DIFF
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -193,7 +193,7 @@ pub fn testnet_genesis(
         .chain(initial_nominators.iter().map(|x| {
             use rand::{seq::SliceRandom, Rng};
             let limit = (MaxNominations::get() as usize).min(initial_authorities.len());
-            let count = rng.gen::<usize>() % limit;
+            let count = (rng.gen::<usize>() % limit).max(1); // at least one nomination
             let nominations = initial_authorities
                 .as_slice()
                 .choose_multiple(&mut rng, count)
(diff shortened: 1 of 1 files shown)


node/src/chain_spec/mod.rs
```diff
@@ -193,7 +193,7 @@ pub fn testnet_genesis(
         .chain(initial_nominators.iter().map(|x| {
             use rand::{seq::SliceRandom, Rng};
             let limit = (MaxNominations::get() as usize).min(initial_authorities.len());
-            let count = rng.gen::<usize>() % limit;
+            let count = (rng.gen::<usize>() % limit).max(1); // at least one nomination
```
Lezek123: How about:
```suggestion
            let count = (rng.gen::<usize>() % limit).saturating_add(1); // at least one nomination
```
The issue with the current approach is that `count` never reaches `initial_authorities.len()`, for example: with 2 initial authorities and 50 nominators, none of the nominates more than 1 authority. With 3 authorities and 50 nominators, none of them nominates more than 2 etc. This is because `rng.gen::<usize>() % limit` will never be equal to `limit`, but I don't think there is any reason it shouldn't be

Lezek123 (changes requested):

node/src/chain_spec/mod.rs
mnaamani: Thats a great observation. I agree there isn't a reason why a nominator can't or shouldn't nominate all the initial validators. I think as long as we have something that works I'm happy. I doubt we are going to start a chain with a pre-defined set of nominators, let a lone randomly backing validators, but its good to have this functioning in anycase.

Lezek123 (approved):
LGTM

Issue #3979, opened 2022-06-30
size_of::<Call> more than 208 bytes


mnaamani:
We have a test in the runtime that checks the size of the `Call` type and it is reporting that is is larger than 208

```
---- runtime_api::tests::call_size stdout ----
thread 'runtime_api::tests::call_size' panicked at 'size of Call 352 is more than 208 bytes: some calls
have too big arguments, use Box to reduce the size of Call. If the limit is too strong, maybe consider
increase the limit to 300.', runtime/src/runtime_api.rs:466:9
```

This test is taken from substrate repo. Should we be concerned or is it just something we have to accept. The test does suggest increasing the limit and using `Box` to reduce it's size. What does the latter mean exactly?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202531880923166) by [Unito](https://www.unito.io)

mnaamani:
I've asked for some clarity from https://substrate.stackexchange.com/questions/3566/what-should-be-the-maximum-size-of-call-for-a-runtime

bedeho:
Let us try to 
- get specific numbers on how big the `Call` value was that caused problems for Parity
- produce list of sizes for extrinsics across runtime

PR #3981, opened 2022-06-30
update size_of::<Call> test to allow upto 352

mnaamani:
One way to fix https://github.com/Joystream/joystream/issues/3979
But I think we should still look into this issue more deeply.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202536077867207) by [Unito](https://www.unito.io)


bedeho:
Lets try to pin down what the max safe number is for `Call`. Our assumption is that for the processing of a block, there will be exactly one `Call` instance on the stack of a validator simultanously, hence we must ensure that this does not overrun the stack size of the default/desired node WASM execution environment. So we need to pin down

- how many `Calls` can there be at most for a given set of node/runtime configurations and a given `Call` size.
- how is the stack size.

Unless we have this pinned down, we will not be able to know if any candidate reduction of `Call` is actually potent, and we also will miss the possibility that we are already fine at `352  bytes`+our node. We may also find its cheaper to adjust our node confiuguration, rather than runtime, to make it all safe in the end, at least at this time.

Issue #3944, opened 2022-06-22
Update faucet


bedeho:
No longer use invitation system as it depends on DAO activity, just pay for memberships straight up. Try to keep API the same.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202487333255650) by [Unito](https://www.unito.io)

bedeho:
While the invitation system has the constraint of meaning that the invitor has to have a certain number of invites, which is contingent on DAO authorities replenishing them over tim (which then becomes a source of failure when you run out), it still has the virtue of imposing transaction fee only locking, which dramatically lowers the risk of being Sybilled, which is the main obstacle of running a faucet. Perhaps then the best thing to do is to separate these two: allow somehow to invite or create a membership for someone where you endow them with some initial source of funds that are subject to an invitation lock, or some lock with a similar effect. This seems to require a runtime change, and also change to faucet.

mnaamani:
Implemented in #3985 #4098

PR #3985, opened 2022-07-02
gift_membership feature

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3944

Add new extrinsic `gift_membership` in runtime membership pallet. Allows any account with sufficient funds to buy a membership for someone. This gets around `invite_member()` constraints of working group: no need to setup working group lead, no need to set a budget and providing sufficient invites to a lead and then transferring to a dedicated member account to be used by the inviter/faucet. The Gifter pays the membership fee and can optionally provide new member with some funds that can be locked for only transaction payments.

- [x] Implementation complete
- [x] Benchmarks
- [x] Unit tests 
- [x] Update chain-metadata.json and gen types
- [x] Update Query Node mappings for membership
~- [ ] Update playground startup script to setup faucet with this new approach.~
~- [ ] Update faucet implementation.~

The faucet service api will be unchanged. We just need to update the implementation to use new call.


bedeho:
# Feedback

I think this initial approach needlessly introduces a Sudo privilidge for a problem that does not really need it, as faucet operators should expect to have to pay out of their own pocket, see here https://github.com/Joystream/joystream/issues/3944#issuecomment-1170893440.

I think the simplest solution here would just be to update `buy_membership`, and introduce a parameter field to `BuyMembershipParameters` like this for example

```Rust
struct BuyMembershipParameters {
/* snip */
gift: NewMembershipGift
}

struct NewMembershipGift {

 // Caller pays, even if account exists, and applies lock if  Some(x), but only if x <= credit_controller_account_balance,
// None means no lock.
 credit_controller_account_balance: Balance
 apply_controller_account_invitation_lock_of_size: Option<Balance> 
 
 // Caller pays, even if account exists, and applies lock if  Some(x), but only if x <= credit_root_account_balance
 // None means no lock.
 credit_root_account_balance: Balance
 apply_root_account_invitation_lock_of_size: Option<Balance>
}
```

mnaamani:
Ok I think I mis-understood the original proposal. I would suggest a small change, instead of modifying the buy_membership extrinsic, and BuyMembershipParameters type, to add a new extrinsic specifically for this purpose ie.

```rust
struct NewMembershipGift {
 // Caller pays, even if account exists, and applies lock if  Some(x), but only if x <= credit_controller_account_balance,
// None means no lock.
 credit_controller_account_balance: Balance
 apply_controller_account_invitation_lock_of_size: Option<Balance> 
 
 // Caller pays, even if account exists, and applies lock if  Some(x), but only if x <= credit_root_account_balance
 // None means no lock.
 credit_root_account_balance: Balance
 apply_root_account_invitation_lock_of_size: Option<Balance>
}

fn gift_membership(origin, params: BuyMembershipParameters, gift: NewMembershipGift) {
    // re-using of BuyMembershipParameters, we just don't do anything with referrer_id field
}
```

Avoid's refactoring/modifying types, tests benchmarks.

bedeho:
I guess that should be fine, but please review whether  to keep  using `BuyMembershipParameters`, it may have stuff in there that is not relevant to gifting, also, it may just be more parsimonious to just make a new separate `GiftMembershipParamters` which is more focused.

mnaamani:
> I guess that should be fine, but please review whether to keep using `BuyMembershipParameters`, it may have stuff in there that is not relevant to gifting, also, it may just be more parsimonious to just make a new separate `GiftMembershipParamters` which is more focused.

Noted. Indeed as I pointed out there is the `referrer_id` field in `BuyMembershipParameters` which doesn't make sense for a gift, since the gifter shouldn't get a referral cut ;)

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+        ) {
+            let gifter = ensure_signed(origin)?;
+
+            let handle_hash = Self::get_handle_hash(
+                &params.handle,
+            )?;
+
+            // Check that gifter has sufficient funds
+            let membership_fee = Self::membership_price();
+            let gifter_free_balance = balances::Pallet::<T>::usable_balance(&gifter);
```
dobertRowneySr:
```suggestion
            let gifter_usable_balance = balances::Pallet::<T>::usable_balance(&gifter);
```

dobertRowneySr: I would recommend using  the word usable, since  the free balance for an account is lock-agnostic, in Substrate parlance

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+            let total_credit = params
+                .credit_controller_account
+                .saturating_add(params.credit_root_account)
+                .saturating_add(membership_fee);
+
+            ensure!(
```
dobertRowneySr, line 834: There is a subtle issue with the balance accounting:
In the mutation section you are transfering (hence locks are relevant):
- `credit_root_account`: `gifter` -> `root_account`
- `credit_controller_account`:`gifter` -> `controller_account`

Then you are slashing `membership_fee` from `gifter` (slashing is lock-agnostic). So I will recommend
`total_credit = credit_root_account + credit_controller_account` then checking separately:
`can_slash(gifter, membership_fee + total_credit)` and `gifter_usable_balance >= total_credit`

dobertRowneySr: Here's a easy to read reference: https://substrate.stackexchange.com/questions/24/what-is-the-difference-between-free-balance-reserved-balance-locked-balance-v

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+                balances::Pallet::<T>::can_slash(&gifter, membership_fee),
+                Error::<T>::InsufficientBalanceToGift
+            );
+
+            ensure!(
```
dobertRowneySr, line 839: See my prev. comment

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+                gifter_free_balance >= total_credit,
+                Error::<T>::InsufficientBalanceToGift
+            );
+            ensure!(
```
dobertRowneySr: See my prev. comment

runtime-modules/membership/src/lib.rs
dobertRowneySr: And also should we allow the `gifter` to be dropped in case the balance gets below the existential deposit limit?

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+                // Lock credited balance. Allow only transaction payments.
+                if let Some(locked_balance) = params.apply_controller_account_invitation_lock {
+                    T::InvitedMemberStakingHandler::lock_with_reasons(
+                        &params.controller_account,
+                        locked_balance,
+                        WithdrawReasons::except(WithdrawReasons::TRANSACTION_PAYMENT)
+                    );
+                };
+            } else {
+                let locked_balance: BalanceOf<T> = params.apply_controller_account_invitation_lock.map_or(Zero::zero(), |b| b)
```
dobertRowneySr:
```suggestion
                let locked_balance: BalanceOf<T> = params.apply_controller_account_invitation_lock.unwrap_or(Zero::zero())
```

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+                    .saturating_add(params.apply_root_account_invitation_lock.map_or(Zero::zero(), |b| b))
```
dobertRowneySr:
```suggestion
                    .saturating_add(params.apply_root_account_invitation_lock.unwrap_or(Zero::zero()))
```

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+                    .saturated_into();
+
+                if locked_balance != Zero::zero() {
```
dobertRowneySr:
```suggestion
                if !locked_balance.is_zero() {
```

runtime-modules/membership/src/lib.rs
```diff
@@ -741,6 +784,153 @@ decl_module! {
...
+            );
+
+            // Ensure the entire fee was slashed. This should not fail
+            // since we checked for sufficient usable balance.
+            ensure!(
+                balance_not_slashed == Zero::zero(),
+                Error::<T>::InsufficientBalanceToGift
+            );
+
+            if params.root_account != params.controller_account {
```
dobertRowneySr, line 910: This is, technically speaking, not necessary, since locks overlay each other and don't stack. 
So adding the another lock with the same amount won't change the way locks are accounted for

runtime-modules/membership/src/tests/mod.rs
```diff
@@ -3,18 +3,109 @@
...
+            params.credit_controller_account
+        );
+
+        // usable-balance
+        assert_eq!(
+            Balances::usable_balance(profile.root_account),
+            params.credit_root_account
+                - params
+                    .apply_root_account_invitation_lock
+                    .map_or(Zero::zero(), |b| b)
```
dobertRowneySr:
```suggestion
                    .unwrap_or(Zero::zero())
```

runtime-modules/membership/src/tests/mod.rs
```diff
@@ -3,18 +3,109 @@
...
+        );
+        assert_eq!(
+            Balances::usable_balance(profile.controller_account),
+            params.credit_controller_account
+                - params
+                    .apply_controller_account_invitation_lock
+                    .map_or(Zero::zero(), |b| b)
```
dobertRowneySr:
```suggestion
                    .unwrap_or(Zero::zero())
```

runtime-modules/membership/src/lib.rs
```diff
@@ -182,6 +186,35 @@ pub struct BuyMembershipParameters<AccountId, MemberId> {
...
+    pub root_account: AccountId,
+
+    /// New member controller account.
+    pub controller_account: AccountId,
+
+    /// New member handle.
+    pub handle: Option<Vec<u8>>,
+
+    /// Metadata concerning new member.
+    pub metadata: Vec<u8>,
```
dobertRowneySr, line 201: This could potentially open up doors for unbounded input type attacks, shouldn't we use [BoundedVec](https://paritytech.github.io/substrate/master/frame_support/struct.BoundedVec.html)

runtime-modules/membership/src/lib.rs
```diff
@@ -182,6 +186,35 @@ pub struct BuyMembershipParameters<AccountId, MemberId> {
...
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, TypeInfo)]
+pub struct GiftMembershipParameters<AccountId, Balance> {
+    /// New member root account.
+    pub root_account: AccountId,
+
+    /// New member controller account.
+    pub controller_account: AccountId,
+
+    /// New member handle.
+    pub handle: Option<Vec<u8>>,
```
dobertRowneySr, line 198: This could potentially open up doors for unbounded input type attacks, shouldn't we use [BoundedVec](https://paritytech.github.io/substrate/master/frame_support/struct.BoundedVec.html)

dobertRowneySr (changes requested):
1. There are some issues with gifter balance accounting
2. There could be vulnerabilities in passing unbounded `metadata` and `handle` bytes
3. I would add more granularity to the tests (this is a bit arbitrary so use your own judment as well):
 - `gift_membership_fails_with_controller_lock_greater_than_controller_credit_amount`
 -  `gift_membership_fails_with_root_lock_greater_than_root_credit_amount`
 -  `gift_membership_succeeds_with_membership_fee_locked_out` (basically you could add a vesting schedule for a total amount ` >= membership_fee + total_credit`, vest `total_credit` and leave `membership_fee` locked out
 - `gift_membership_succeeds_with_same_root_and_controller`

runtime-modules/membership/src/lib.rs
mnaamani: Excellent catch here with the proper checks.
For allowing death, maybe I will re-consider. Perhaps it can be a bool argument to the dispatch call, so caller can decide?

runtime-modules/membership/src/lib.rs
mnaamani: Correct. Here it was just an optimization so if we already checked once, we don't need to check a second time if the accounts are the same.

runtime-modules/membership/src/lib.rs
mnaamani: That is a very good point. The `BuyMembershipParameters` and `InviteMembershipParameters` have the same problem. As a matter of fact this may be the case for other calls in more than one pallet.

Now I see the benchmarks use an upper bound `const MAX_BYTES: u32 = 16384;` for inputs like `Vec<u8>`. I think if that would fill a block and the weight function of the call uses that as a parameter, then the attacker will be paying a large fees for the attack. Is that sufficient to mitigate the attack.

Although the handle and metadata are not stored in state, It probably still make sense to test query-node if it can handle large values.

runtime-modules/membership/src/lib.rs
mnaamani: clippy eventually recommended `unwrap_or_else` over `unwrap_or`.
So I ended up with `.unwrap_or_else(Zero::zero)`

see: https://stackoverflow.com/a/56726618/19639605

runtime-modules/membership/src/lib.rs
dobertRowneySr: This issue about `Vec<u8>` is present all over the code base, maybe it is worth to open an issue in order to find a definite solution (provided that it is not already present)
I will take care of the issue. Let's leave this part like this for the moment

runtime-modules/membership/src/lib.rs
dobertRowneySr: I think it will be better to make sure the `gifter` has enough cash to cover everything including ex. dep.
So that the `gift_membership` extr. can only used in one single (simple) way. But besides that I don't have any other 
counter arguments

dobertRowneySr (approved):
LGTM

Issue #3934, opened 2022-06-21
Filter initial extrinsics for Carthage rollout


bedeho:
# Background
While launching `Carthage` and mainnet, we wish to filter out any non-essential user actions, so as to avoid disrupting #3903 and to make relaunches more graceful, if required in a failure scenario.

# Proposal
Using `BaseFilter` we drop all extrinsics unless they are for pallets

* `sudo`: so as to allow bootstrapping by Jsgensis and next runtime update
* `staking`: so as to allow user stake tokens and declare their intent to validate or nominate. **Double check that this actually works and is the correct pallet to filter for this purpose, and nothing else is needed.**

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202481649471457) by [Unito](https://www.unito.io)

mnaamani:
Implementation in #3986

PR #3986, opened 2022-07-03
BaseCallFilter to limit calls on launch

mnaamani:
Implement https://github.com/Joystream/joystream/issues/3934

There are more calls enabled than was suggested in the proposal, but that is necessary for validators to function properly as they submit inherent extrinsics. like the timestamp for example.

I confirmed that the Utiility.batch call applies filter as well so there is no way to circumvent filter with that call.

While testing that `SudoAs` can still make calls to the filtered calls, I was getting a BadOrigin error. Initially I thought it was because the address lookup was failing, because we have configured runtime Address type to be an MultiAddress. Turns out this was just an issue from within  polkadot-js/app itself. Tested with a script and works as expected.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202545114021400) by [Unito](https://www.unito.io)


mnaamani:
@bwhm Can a validator stash and controller account be the same? I don't remember off the top of my head.

bedeho:
Make compile time inclusion conditional on benchmarking being done.

mnaamani:
> Make compile time inclusion conditional on benchmarking being done.

Done in https://github.com/Joystream/joystream/pull/3986/commits/4d3fe340209368c3a46a0bedbe7fb3df2459a91f

dobertRowneySr (approved):
LGTM

PR #3992, opened 2022-07-05
Forum bug fixes

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3111 according to specification in https://github.com/Joystream/joystream/issues/3111#issuecomment-1081717432, namely:
> 1. block the ability to delete a thread if `Thread::number_of_posts > 0`.
> 1. update `ensure_can_moderate_post` so that it does not accept the possibility that the thread does not exist due change in 1, thereby you can always authenticate whether a given moderator genuinely is allowed to moderate a given post.
> 1. make sure we never do this `<PostById<T>>::get(thread_id, post_id)`, which should not be needed

Updates tests and benchmarks accordingly.

Also addresses https://github.com/Joystream/joystream/issues/3994

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202557710456426) by [Unito](https://www.unito.io)


runtime-modules/forum/src/lib.rs
```diff
@@ -2342,4 +2348,12 @@ impl<T: Config> Module<T> {
             Err(Error::<T>::DataMigrationNotDone)
         }
     }
+
+    fn ensure_empty_thread(thread: &ThreadOf<T>) -> DispatchResult {
+        ensure!(
+            thread.number_of_posts == 0u64,
```
dobertRowneySr:
```suggestion
            thread.number_of_posts.is_zero(),
```
In case we decide to change the type

runtime-modules/forum/src/tests.rs
```diff
@@ -98,6 +98,30 @@ fn update_category_membership_category_moderator_does_not_exist() {
     });
 }
 
+#[test]
+// test an attempt to add moderator to a category when provided moderator_id is not an existing worker id
+fn update_category_membership_moderator_worker_does_not_exist() {
```
dobertRowneySr:
```suggestion
fn update_category_membership_fails_when_moderator_worker_does_not_exist() {
```

dobertRowneySr (changes requested):
Good job, just small fixes needed

dobertRowneySr (commented):
LGTM, merging

Issue #3461, opened 2022-03-20
Query node - ensure all Promises are resolved


ondratra:
When a developer mistakenly omits `await` keyword where `Promise` is expected to be resolved, a bug that's not easy to spot can occur.

    // assume we have some db adapter and next id generator
    const db = ...
    const getNextId: () => number = ... 
        
    async function save(objectToBeSaved: {name: string}) {
      objectToBeSaved.id = nextId() 
      await db.save(objectTobeSaved)
    }
        
    const myBook = { name: 'My nice book' }
    save(myBook) // `await` is mistakenly missing -> promise is not resolved on this line
    // await save(myBook) // that's what actually should be here
        
    console.log(myBook.id) // prints `undefined` instead of expected numberAs discussed in https://github.com/Joystream/joystream/pull/3460, there is a TS lint directive that ensures all promises are handled, and no "fork" happens unexpectedly. https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md Ensure this directive is used in both QN codebase as well as Hydra codebase. Adapt the code if needed, and the described error can't happen.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202453843320849) by [Unito](https://www.unito.io)

PR #3921, opened 2022-06-20
build QN for Carthage runtime

zeeshanakram3:
implements #3895 

Includes 

* Updated mappings
* Updated types package (6cd14d44132c2ac5e60bbc4534b23713a8ba0dc1)
* Updated CLI commands related to channels/videos creation
* Integration tests adjustments

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202476251966520) by [Unito](https://www.unito.io)

PR #3998, opened 2022-07-06
Carthage lint improvements

ondratra:
Fixes https://github.com/Joystream/joystream/issues/3461. Builds on top of https://github.com/Joystream/joystream/pull/3921 and expects it to be merged first. Replaces https://github.com/Joystream/joystream/pull/3977.

Adds linting of all typescript files in monorepo to pre-push hook. The linting checks can also be invoked by running `yarn lint` in the root directory.

Changes in code were done in:
- [package.json](https://github.com/Joystream/joystream/pull/3998/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)
- [queryTemplates.ts](https://github.com/Joystream/joystream/pull/3998/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)
- new script for linting workspaces [scripts/lint-typescript.sh](https://github.com/Joystream/joystream/pull/3998/files#diff-8396cc566661aa39fed1a013118af4d9af3319f17fa43d02cad103285a50014d)
- lint setting update [devops/eslint-config/index.js](https://github.com/Joystream/joystream/pull/3998/files#diff-868664046b6ded144d52d664f9ad6e916542616210c8592fd13c93546c224ec2)
- husky pre-push hook changed to include linting workspaces [devops/git-hooks/pre-push](https://github.com/Joystream/joystream/pull/3998/files#diff-c45dce71c224d3b549c475993aa33ed666c39a3fb1ebc28c64474b539c844f94)

These files need review - the rest files are only formatted/linted due to new lint settings.

The CI tests failing are to be expected because Carthage doesn't have all tests passing yet.


mnaamani (changes requested):
Really nice effort on this Andy.  Just one small fix needed is to re run `yarn build` in the types packages and check the files in, to revert the eslint autofix already applied to the augment/* files. My guess is you added them to .eslintignore after they were linted?

Nice catching quite a few missed `await`s, hopefully we will not miss any in the future with this linting and it will certainly fix a few hidden bugs. We should probably add the lint step in our CI checks too.

ondratra:
> Really nice effort on this Andy. Just one small fix needed is to re run `yarn build` in the types packages and check the files in, to revert the eslint autofix already applied to the augment/* files. My guess is you added them to .eslintignore after they were linted?

It turned out to be a more complicated problem. I've noticed that changes in `src/augment/` folder were all just reordering existing lines - and all those lines were containing a letter(s) `ch`. And then it stroke me. In Czech and Slovak languages, we have a character `ch` in our alphabets that sits between `h` and `i`. Since my OS is using a Czech locale, something inside of the types generating script is using this locale to sort things during code generation. And the results vary between my local setup when `ch` is one character with a special place in the alphabet, and English-like setups where `ch` are two characters together `c` + `h`. That creates these weird changes that I've also noticed a couple of times before during query node build.

Unfortunately, there seems to be no way how to force the locale (`C`, `en_US.UTF8`, etc.) in the monorepo project. NPM, Yarn, Volta nor Cargo support such a thing. So I will have to be careful when generating code sensitive to this problem and use `LC_ALL=C yarn workspace @joystream/types build` to prevent this problem.

Changes in `types/src/augment` are reverted now - it's fixed now.

> Nice catching quite a few missed `await`s, hopefully we will not miss any in the future with this linting and it will certainly fix a few hidden bugs. We should probably add the lint step in our CI checks too.

I've created an issue so we can handle that as a separate effort. https://github.com/Joystream/joystream/issues/4024

Issue #3838, opened 2022-06-02
INFO_14: `accept_incoming_offer` dangerous race condition


Lezek123:
Quote:
> The API of `accept_incoming_offer` extrinsic should be changed to match the API of
`accept_channel_transfer`, both for coherence and prevent future updates to open the possi-
bility of a race condition. If the `offer_nft` function did not check that the status is Idle or if a
function allowed to update the offer, the `accept_incoming_offer` extrinsic would be vulnerable
to a race condition (updating the offer in the same block would make the other user accept a
different price than expected).

The missed-out fact here is that there is **already** a race condition and a serious vulnerability due to the fact that one can execute `cancel_offer` and then `offer_nft` again (with a different price) within the same block or even within the same transaction (using a batch call). Without any `witness` parameter, this puts the accepting user at a serious risk of beeing front-runned by an attacker to accept an offer at a higher-than-expected price.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202414586783951) by [Unito](https://www.unito.io)

bedeho:
How is this issue related to this: https://github.com/Joystream/joystream/issues/3856

Lezek123:
Addressed by https://github.com/Joystream/joystream/pull/4000

PR #4000, opened 2022-07-06
INFO_14: accept_nft_offer - add witness parameter

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3838

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202561216119407) by [Unito](https://www.unito.io)


runtime-modules/content/src/lib.rs
```diff
@@ -2225,6 +2225,7 @@ decl_module! {
         pub fn accept_incoming_offer(
             origin,
             video_id: T::VideoId,
+            witness_price: Option<<T as balances::Config>::Balance>
```
dobertRowneySr, line 2228: in `buy_nft` I have used the term `price_commit` with the same semantics. If you think that `witness_price` is better please change that one also, so there is no confusion (since both params serve the same purpose)

runtime-modules/content/src/nft/mod.rs
```diff
@@ -265,10 +265,14 @@ impl<T: Config> Module<T> {
     pub(crate) fn ensure_new_pending_offer_available_to_proceed(
         nft: &Nft<T>,
         participant_account_id: &T::AccountId,
+        witness_price: Option<<T as balances::Config>::Balance>,
```
dobertRowneySr, line 271: in `buy_nft` (and related) I have used the term `price_commit` with the same semantics. If you think that witness_price is better please change that one also, so there is no confusion (since both params serve the same purpose)

runtime-modules/content/src/nft/mod.rs
```diff
@@ -265,10 +265,14 @@ impl<T: Config> Module<T> {
...
     ) -> DispatchResult {
         if let TransactionalStatus::<T>::InitiatedOfferToMember(member_id, price) =
             &nft.transactional_status
         {
+            // Validate witness price
+            ensure!(*price == witness_price, Error::<T>::NftOfferPriceChanged);
```
dobertRowneySr: In the similar case for buy now I have used `InvalidBuyNowPriceProvided` error.
What do you think about renaming to `InvalidOfferPriceChange` ?

dobertRowneySr (changes requested):
Nothing major, just cosmetics

dobertRowneySr (approved):
LGTM

Issue #3854, opened 2022-06-06
LOW_13:Check in `ensure_constraints_on_bid_amount should` be `>=`


dobertRowneySr:
The `ensure_constraints_on_bid_amount` https://github.com/Joystream/joystream/blob/eb735d038dae707303f892bd62da5c57834fc5a1/runtime-modules/content/src/nft/types.rs#L192 
is checking the condition `amount > buy_now which`
is not consistent with the rest of the code in the pallet. Moreover, if the top bid is too close to
buy_now_price, the extrinsic will fail if a user sends a bid of `buy_now_price`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202414676680676) by [Unito](https://www.unito.io)

PR #4010, opened 2022-07-08
Carthage  buy now condition edge case

dobertRowneySr:
Addresses https://github.com/Joystream/joystream/issues/3854

##### Changes
* Change bid comparison to buy now to >=
* Add test for edge case

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202926796721538) by [Unito](https://www.unito.io)


Lezek123 (changes requested):
1. Needs update from Carthage
2. We now have `make_bid_ok_with_english_auction_completion_with_bid_below_min_step`, but there is no such test for `open_auction`
3. I think `make_bid_ok_with_english_auction_completion_and_total_balance_slashed` should still use `DEFAULT_BUY_NOW_PRICE + 10` for the bid (just like the `make_bid_ok_with_open_auction_completion_and_total_balance_slashed`), as it tests the behavior when `bid > buy_now`, which is a specific case that is now missing for english auction.

dobertRowneySr:
> 2. We now have `make_bid_ok_with_english_auction_completion_with_bid_below_min_step`, but there is no such test for `open_auction`

The logic problem was present in only english auction (for which the bid step constraint is enforced), the tests highlights this particular instance
- `min_bid_step = 20`
- `buy_now_price = 100`
- `first_offer = 90`

Now if another user steps in and makes another offer at `second_offer = 100  == buy_now_price` it throws since
the check against `buy_now_price` has only a `<` sign and the guard for the `min_bid_step` constraint will fail also.

Open auction have no concept of `bid_step`.

> 3. I think `make_bid_ok_with_english_auction_completion_and_total_balance_slashed` should still use `DEFAULT_BUY_NOW_PRICE + 10` for the bid (just like the `make_bid_ok_with_open_auction_completion_and_total_balance_slashed`), as it tests the behavior when `bid > buy_now`, which is a specific case that is now missing for english auction.

:+1:

runtime-modules/content/src/tests/nft/make_bid.rs
```diff
@@ -788,27 +788,56 @@ fn make_bid_ok_with_open_auction_completion_and_no_reserve_balance_left_for_bidd
...
-        );
-    })
+#[test]
+fn make_bid_ok_with_english_auction_completion_and_total_balance_slashed() {
+    pub const BID_OFFER: u64 = DEFAULT_BUY_NOW_PRICE + 10;
+    ExtBuilder::default()
+        .build_with_balances(vec![(SECOND_MEMBER_ACCOUNT_ID, BIDDER_BALANCE)])
+        .execute_with(|| {
+            ContentTest::default().with_video_nft().setup();
+            StartEnglishAuctionFixture::default().call_and_assert(Ok(()));
```
Lezek123: This function should test auction completion scenario (as the name suggests), but the auction doesn't complete, since there is no `buy_now_price`

```suggestion
            StartEnglishAuctionFixture::default()
                .with_buy_now_price(DEFAULT_BUY_NOW_PRICE)
                .call_and_assert(Ok(()));
```

runtime-modules/content/src/tests/nft/make_bid.rs
```diff
@@ -788,27 +788,56 @@ fn make_bid_ok_with_open_auction_completion_and_no_reserve_balance_left_for_bidd
...
+            assert_ok!(Content::make_english_auction_bid(
+                Origin::signed(SECOND_MEMBER_ACCOUNT_ID),
+                SECOND_MEMBER_ID,
+                VideoId::one(),
+                BID_OFFER,
+            ));
+
+            assert_eq!(
+                Balances::<Test>::usable_balance(&SECOND_MEMBER_ACCOUNT_ID),
+                BIDDER_BALANCE - BID_OFFER,
```
Lezek123: We expect the bidder balance to be decreased by no more than the "buy now" price
```suggestion
                BIDDER_BALANCE - DEFAULT_BUY_NOW_PRICE,
```

Lezek123 (changes requested):
Also needs update from `carthage`

Lezek123 (approved):
LGTM, just needs a conflict resolution

Issue #3221, opened 2022-02-12
Benchmark content pallet


bedeho:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202329185040794) by [Unito](https://www.unito.io)

PR #4011, opened 2022-07-08
Content Pallet enchmarks

L3pereira:
* https://github.com/Joystream/joystream/issues/3221
 
Implementation of create_channel extrinsic benchmark

Note:

The runtime constraints, for example

`StorageBucketsPerBagValueConstraint`, can be represented in a complex variable, 
`let c in 0 .. T::StorageBucketsPerBagValueConstraint::get().max();`
the problem is that this constraint is a range and it has a min and max, so we should write 
`let c in T::StorageBucketsPerBagValueConstraint::get().min .. T::StorageBucketsPerBagValueConstraint::get().max();`.
Unfortunately this is not allowed, so a range rescale function was added.
`let c in 0 .. 100; ` the complex variable is always in between  0 and 100
and then we rescale this range [0;100] to [`T::StorageBucketsPerBagValueConstraint::get().min`;`T::StorageBucketsPerBagValueConstraint::get().max()`]

```rust
fn rescale_constraints<T: Config>(new_min: u64, new_max: u64, value: u32) -> Option<u32> {
    let old_min = 0f64;
    let old_max = 100f64;
    let value = value as f64;
    let new_min: f64 = new_min as f64;
    let new_max: f64 = new_max as f64;

    if value < old_min || value > old_max || new_max < new_min {
        return None;
    }

    let new_value: u32 =
        ((new_max - new_min) / (old_max - old_min) * (value - old_max) + new_max) as u32;

    if new_value < new_min as u32 || new_value > new_max as u32 {
        return None;
    }

    Some(new_value)
}
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202577004325353) by [Unito](https://www.unito.io)


runtime-modules/content/src/benchmarks/channels.rs
```diff
@@ -0,0 +1,197 @@
...
+                    f.into());
+                    id.saturated_into()
+            })
+            .collect::<BTreeSet<_>>();
+
+        let distribution_families =
+            create_distribution_bucket_with_family_adjusted::<T>(
+                distribution_wg_lead_account_id.clone(),
+                max_distribution_bucket_family_number,
+                rescaled_e);
```
Lezek123: The `e` parameter ends up unused by the weight function:
```rust
fn channel_creation_with_channel_bag(a: u32, b: u32, c: u32, _e: u32, f: u32, ) -> Weight {
```
And the weight is way overestimated:
```rust
// ...
			.saturating_add(T::DbWeight::get().reads(32 as Weight))
			.saturating_add(T::DbWeight::get().writes(25 as Weight))
```

Because you're using `max_distribution_bucket_family_number` as a 2nd parameter here, which means there are always **at least 20** distribution buckets assigned to the channel regardless of `e`, since `create_distribution_bucket_with_family_adjusted` always assigns **at least 1** bucket to each family:

```rust
    let db_per_family = bucket_number / bucket_family_number;
    let db_per_family = match db_per_family {
        0 => 1,
        _ => db_per_family,
    };
```

runtime/src/lib.rs
```diff
@@ -898,16 +899,16 @@ impl common::StorageOwnership for Runtime {
...
     pub const MaxNumberOfPendingInvitationsPerDistributionBucket: u64 = 20; //TODO: adjust value
     pub const StorageModuleId: PalletId = PalletId(*b"mstorage"); // module storage
     pub const StorageBucketsPerBagValueConstraint: storage::StorageBucketsPerBagValueConstraint =
-        storage::StorageBucketsPerBagValueConstraint {min: 5, max_min_diff: 15}; //TODO: adjust value
+        storage::StorageBucketsPerBagValueConstraint {min: 1, max_min_diff: 19}; //TODO: adjust value
     pub const DefaultMemberDynamicBagNumberOfStorageBuckets: u64 = 5; //TODO: adjust value
     pub const DefaultChannelDynamicBagNumberOfStorageBuckets: u64 = 5; //TODO: adjust value
     pub const DistributionBucketsPerBagValueConstraint: storage::DistributionBucketsPerBagValueConstraint =
-        storage::DistributionBucketsPerBagValueConstraint {min: 1, max_min_diff: 100}; //TODO: adjust value
+        storage::DistributionBucketsPerBagValueConstraint {min: 1, max_min_diff: 19}; //TODO: adjust value
```
Lezek123: I don't think those parameters should be changed

runtime-modules/content/src/benchmarks/channels.rs
```diff
@@ -0,0 +1,197 @@
...
+              T::AccountId: CreateAccountId,
+              T: Config ,
+    }
+
+    channel_creation_with_channel_bag {
+
+        let a in 1 .. MAX_BYTES; //max metadata
+        let b in 1 .. MAX_COLABORATOR_IDS as u32; //max colaborators
+        let c in 0 .. 100; //StorageBucketsPerBagValueConstraint
+        let e in 0 .. 100; // DistributionBucketsPerBagValueConstraint
```
Lezek123: This re-scaling approach doesn't seem correct, because it will produce invalid weights.

When passing parameters to the weight functions you're using the real number of buckets:
```rust
        //storage_buckets
        let c = params.storage_buckets.len() as u32;
        //distribution_buckets
        let e = params.distribution_buckets.len() as u32;
```

But here in the benchmarks this `c` and `e` is "skewed", because the actual number of buckets used for benchmarking depends on `rescaled_c` and `rescaled_e`, which are rescaled from an arbitrary `0..100` range.

It seems to be actually possible to use:
```rust
let c in (T::StorageBucketsPerBagValueConstraint::get().min as u32) .. (T::StorageBucketsPerBagValueConstraint::get().max() as u32);
```

So I suggest to just go with that

runtime-modules/content/src/weights.rs
```diff
@@ -0,0 +1,89 @@
...
+			.saturating_add((8_365_000 as Weight).saturating_mul(f as Weight))
+			.saturating_add(T::DbWeight::get().reads(32 as Weight))
+			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(b as Weight)))
+			.saturating_add(T::DbWeight::get().writes(25 as Weight))
+			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(f as Weight)))
+	}
+}
+
+// Default implementation for tests
+impl WeightInfo for () {
```
Lezek123, line 90: Causes:
```
conflicting implementations of trait `weights::WeightInfo` for type `()`
conflicting implementation for `()`rustc[E0119](https://doc.rust-lang.org/error-index.html#E0119)
mock.rs(425, 1): first implementation here
```

The `WeightInfo` impl should be removed from `mock.rs`

runtime-modules/content/src/lib.rs
```diff
@@ -3589,6 +3599,33 @@ impl<T: Config> Module<T> {
...
+    // Calculates weight for channel_creation_weight extrinsic.
+    fn channel_creation_weight(params: &ChannelCreationParameters<T>) -> Weight {
+        // new_meta.
+        let a = params.meta.as_ref().map_or(0, |v| v.len()) as u32;
+
+        //collaborators
+        let b = params
+            .collaborators
+            .iter()
+            .fold(0, |acc, (_, permissions)| acc + permissions.len() + 1) as u32;
```
Lezek123: In the benchmark code `b` represents the number of collaborators, but here you're passing the number of collaborators + the number of permissions that each collaborator has

Lezek123 (changes requested):

runtime-modules/content/src/lib.rs
L3pereira: I'm trying to represent the worst case scenario.
Collaborators are composed by a collection inside a collection
`pub collaborators: BTreeMap<MemberId, ChannelAgentPermissions>`
iterating that will make a node to process the ChannelAgentPermissions for each collaborator.
Am I mistaken?

runtime-modules/content/src/benchmarks/channels.rs
L3pereira: But for this extrinsic the distribution bucket family number is already defined (is not an input), it should not be taken as a variable right? 
That's why I chose the worst case scenario for that variable (max_distribution_bucket_family_number).

Lezek123: The number of families is not an input, but the number of buckets is part of the input (`let e = params.distribution_buckets.len() as u32;`)
What your script does is it always creates 20 families with at least 1 bucket each and then it uses all those buckets as part of `create_channel` input regardless of `e`:
```rust
        // This always creates at least 20 families (with 1 bucket per family)
        let distribution_families =
            create_distribution_bucket_with_family_adjusted::<T>(
                distribution_wg_lead_account_id.clone(),
                max_distribution_bucket_family_number,
                rescaled_e);

        // This creates a policy where each dynamic bag needs to be assigned to basically all existing buckets
        let distribution_buckets_families_policy =
            distribution_families.clone().into_iter()
            .map(|(family_id, buckets)|
                (family_id, buckets.len() as u32))
            .collect::<BTreeMap<_,_>>();

        // This collects ids of all existing buckets (which is always >= 20) 
        let distribution_buckets =
            distribution_families.into_iter()
            .flat_map(|(family_id, buckets)| buckets)
            .collect::<BTreeSet<_>>();
```
Even if you choose to create `max_number_of_families` (although I'm not sure if it's actually necessary), you could still use only `e` buckets, but you're always using all of them.

runtime-modules/content/src/lib.rs
Lezek123: Collaborator is just one entry in the map, regardless of `ChannelAgentPermissions` size.
Your benchmarks already use fixed `CHANNEL_AGENT_PERMISSIONS` constant which should represent a worst-case-scenario per collaborator. So `b` is the number of entries in `collaborators` map (`collaborators.len()`) and for each iteration of `b`, the worst case scenario for the size of `ChannelAgentPermissions` is assumed by the benchmarks.

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -491,3 +444,98 @@ pub fn create_data_object_candidates_helper(
         })
         .collect()
 }
+
+#[allow(clippy::too_many_arguments)]
+fn generate_channel_creation_params<T>(
+    dyn_bag_type: DynamicBagType,
```
Lezek123: This is always going to be `DynamicBagType::Channel`, so I don't think you need to take it as a parameter

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+        account::<Self>("default", id.try_into().unwrap(), SEED)
+    }
+}
+
+fn get_permissions(i: u64) -> ChannelAgentPermissions {
+    iter::repeat(())
+        .take(i.saturated_into())
+        .map(|_| CHANNEL_AGENT_PERMISSIONS[i as usize])
+        .collect::<BTreeSet<_>>()
+}
```
Lezek123: It appears to me that this function always returns a set of `len() == 1`, because you just create an iterator over `i` elements, where each element is `CHANNEL_AGENT_PERMISSIONS[i as usize]` (where `i` is constant) and collect it into a `BTreeSet`, which will remove all duplicates, so you will just end up with a single `CHANNEL_AGENT_PERMISSIONS[i as usize]` element in the set.

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+where
+    T: Config
+        + storage::Config
+        + membership::Config
+        + balances::Config
+        + working_group::Config<StorageWorkingGroupInstance>
+        + working_group::Config<DistributionWorkingGroupInstance>,
+    T::AccountId: CreateAccountId,
+{
+    let chan_perm = CHANNEL_AGENT_PERMISSIONS.len() as u32 - 1;
```
Lezek123: Why not just:
```suggestion
    let chan_perm = CHANNEL_AGENT_PERMISSIONS.len() as u32;
```
I think it makes sense for `get_permissions` function to take the length as a parameter (as opposed to index of the last element)

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+    number: u64,
+    size: u64,
+) -> Vec<DataObjectCreationParameters> {
+    let range = (starting_ipfs_index as u64)..((starting_ipfs_index as u64) + number);
+
+    range
+        .into_iter()
+        .map(|_| DataObjectCreationParameters {
+            size,
+            ipfs_content_id: vec![1u8],
```
Lezek123, line 433: This shouldn't be possible. `ipfs_content_id` of any length other than 46 bytes should not be accepted by the runtime. If this is not the case, it means that this important check is missing. This is bad, because it means `ipfs_content_id` is unbound and can be of any arbitrary length.

I created an issue to address that: https://github.com/Joystream/joystream/issues/4022

runtime-modules/content/src/tests/mock.rs
```diff
@@ -465,10 +459,160 @@ impl common::council::CouncilBudgetManager<u128, u64> for CouncilBudgetManager {
...
+        ensure!(
+            Self::is_member_controller_account(&member_id, &account_id),
+            DispatchError::BadOrigin
+        );
+        Ok(account_id.into())
+    }
+
+    fn is_member_controller_account(_member_id: &u64, _account_id: &u128) -> bool {
+        return true;
+    }
```
Lezek123: The solution that should work for both testing and benchmarking context:
```suggestion
    fn is_member_controller_account(member_id: &u64, account_id: &u128) -> bool {
        return Membership::is_member_controller_account(member_id, account_id)
            || TestMemberships::is_member_controller_account(member_id, account_id);
    }
```

Lezek123 (changes requested):

runtime-modules/content/src/benchmarks/mod.rs
L3pereira: I replaced by 
```rust
    let permissions = BTreeSet::from_iter(CHANNEL_AGENT_PERMISSIONS);
```
and removed the get_permissions

L3pereira:
I removed metadata from weights since it's not being stored

Lezek123:
> I removed metadata from weights since it's not being stored

It is emitted as an event data, however, so I think it may have some effect

L3pereira:
But when I generated the weights with metadata, the benchmark engine didn't use the variable

Lezek123 (approved):
Ok, LGTM!

PR #4001, opened 2022-07-06
INFO_2: Add missing `update_data_object_state_bloat_bond` benchmark

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3837

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202560490398844) by [Unito](https://www.unito.io)

PR #3980, opened 2022-06-30
Fix all linter/clippy warnings

mnaamani:
Based on https://github.com/Joystream/joystream/pull/3970

Fix all clippy warnings, and re-enable aggressive linter checks before successful build.

I left a few unused imports in some benchmarking.rs files where code was commented out temporarily, to be re-enabled after carthage release.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202537343542077) by [Unito](https://www.unito.io)

PR #4015, opened 2022-07-11
Fix clippy warnings in storage `weights.rs`

Lezek123:
Fix clippy warnings introduced in https://github.com/Joystream/joystream/pull/4001 which was merged after https://github.com/Joystream/joystream/pull/3980


runtime-modules/storage/src/weights.rs
```diff
@@ -37,6 +37,7 @@
 #![cfg_attr(rustfmt, rustfmt_skip)]
 #![allow(unused_parens)]
 #![allow(unused_imports)]
+#![allow(unused_variables)]
```
dobertRowneySr, line 40: If you add this directive you can safely discard adding the underscore to `i`, `j`

dobertRowneySr (approved):
LGTM, I just left a small note.

Issue #3911, opened 2022-06-16
Fix runtime bug


bedeho:
# Background
https://github.com/Joystream/community-repo/issues/768#issuecomment-1143690560

# Proposal
WIP

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202458930431782) by [Unito](https://www.unito.io)

Issue #3228, opened 2022-02-14
Query node: Remove the need for bootstrapping values


Lezek123:
### Memberships
* `defaultInviteCount` _(initial value: 5)_ - this value is used to populate `Membership.inviteCount` during member creation. Once updated in the runtime, the update is reflected in `MembershipSystemSnapshot.defaultInviteCount`. **Alternative solution:** Include the value in `MembershipBought` and `MemberInvited` events on the runtime side and remove `MembershipSystemSnapshot` entity.
* `membershipPrice` _(initial value: 100)_ - only used in `MembershipSystemSnapshot` entity for consistency. **Alternative solution:** Remove `MembershipSystemSnapshot` entity, Pioneer should fetch the price directly from the chain api (`api.query.members.membershipPrice`)
* `referralCut` _(initial value: 0)_ - only used in `MembershipSystemSnapshot` entity for consistency. **Alternative solution:** Remove `MembershipSystemSnapshot` entity, Pioneer should fetch the value directly from the chain api if needed (`api.query.members.referralCut`)
* `invitedInitialBalance` _(initial value: 100)_ - only used in `MembershipSystemSnapshot` entity for consistency. **Alternative solution:** Remove `MembershipSystemSnapshot` entity, Pioneer should fetch the value directly from the chain api if needed (`api.query.members.initialInvitationBalance`)

### Storage system
* `blacklist` _(initial value: [])_ - only used in `StorageSystemParameters` entity. **Alternative solution:** Remove `StorageSystemParameters` entity, apps should use `api.query.storage.blacklist` directly if needed.
* `storageBucketsPerBagLimit` _(initial value: 0)_ - only used in `StorageSystemParameters` entity. **Alternative solution:** Remove `StorageSystemParameters` entity, apps should use `api.query.storage.storageBucketsPerBagLimit` directly if needed.
* `distributionBucketsPerBagLimit` _(initial value: 0)_ - only used in `StorageSystemParameters` entity. **Alternative solution:** Remove `StorageSystemParameters` entity, apps should use `api.query.storage.distributionBucketsPerBagLimit` directly if needed.
* `uploadingBlocked` _(initial value: false)_ - only used in `StorageSystemParameters` entity. **Alternative solution:** Remove `StorageSystemParameters` entity, apps should use `api.query.storage.uploadingBlocked` directly if needed.
[...]

PR #4016, opened 2022-07-11
Council & Referendum: Store "endsAt" block of each stage

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3911


Lezek123:
QN bootstrapping may also need to be adjusted here (see: https://github.com/Joystream/joystream/issues/3228)

mnaamani (changes requested):
Great fix.
I think you just missed updating the mock genesis config proposal discussion and engine that is why the benchmarks ci check was failing. I fixed and confirmed they all pass now.
Opened PR to your branch https://github.com/Lezek123/substrate-runtime-joystream/pull/13

Obviously update from carthage is needed to fix merge conflicts and that should also get other integration tests to pass.

Issue #3951, opened 2022-06-25
MEDIUM_5: `check_buckets_for_overflow` internal function


bedeho:
# Finding
This function performs a raw addition for checking both the total object number and the total object size limits on a VoucherUpdate structure.
It is used by validation functions which are called directly or indirectly by extrinsics. It could allow to easily bypass limit checks of the storage providers, which might render some of them unresponsive by filling their storage space.

# Analysis
This can be fixed by the following

```rust

    // Iterates through buckets. Verifies voucher parameters to fit the new limits:
    // objects number and total objects size.
    fn check_buckets_for_overflow(
        bucket_ids: &BTreeSet<T::StorageBucketId>,
        voucher_update: &VoucherUpdate,
    ) -> DispatchResult {
        for bucket_id in bucket_ids.iter() {

            let bucket = Self::storage_bucket_by_id(bucket_id);

            let objs_voucher_sum = Self::get_checked_add(
                       &voucher_update.objects_number, 
                       &bucket.voucher.objects_used)?;

            // Total object number limit is not exceeded.
            ensure!(
                objs_voucher_sum <= bucket.voucher.objects_limit,
                Error::<T>::StorageBucketObjectNumberLimitReached
            );

            let objs_voucher_size_sum = Self::get_checked_add(
                       &voucher_update.objects_total_size, 
                       &bucket.voucher.size_used)?;

            // Total object size limit is not exceeded.
            ensure!(
                objs_voucher_size_sum <= bucket.voucher.size_limit,
                Error::<T>::StorageBucketObjectSizeLimitReached
            );
        }

        Ok(())
    }

fn get_checked_add<U>(x: &U, y: &U) -> Result<U, DispatchError>
   where U: CheckedAdd{
      x.checked_add(y).ok_or(Error::<T>::AddOperationOverflow.into())
}
```

# Conclusion
?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202504810299957) by [Unito](https://www.unito.io)

PR #4017, opened 2022-07-12
Carthage  fix check bucket overflow vulnerability

dobertRowneySr:
fixes https://github.com/Joystream/joystream/issues/3951

I followed the suggestion by adding using `checked_add` method. Which I think it should be the default from now on instead of the usual `saturating_add`.

I added a (very complex) test to highlight the vulnerability


Lezek123 (approved):
LGTM, but needs update from Carthage

Lezek123 (changes requested):
Types were not updated

Lezek123 (approved):
LGTM

PR #4025, opened 2022-07-13
Carthage currency tuning

mnaamani:
- [x] Adjusted values for DOLLARS, CENTS, MILLICENT based on initial issuance and estimated USD market cap value.
- [x] Set non zero existential deposit and non zero byte fee for transactions.
- [x] Fixed runtime integration tests to take into account existential deposit.
- [x] Adjusting some rutime config stakes/deposits to fix balances:transfer error 'ExistentialDeposit' ie. when transferring less than ED.
- [x] Fixing integration tests again after tweaks
- [x] Fixing network-integration tests - some constants are hard-coded, read from chain instead, increase starting balances for worker accounts to cover fees
- [x] Fixing benchmarks 
   - [x] utility, council 
   - [x] project tokens
- [x] Made storage fees configurable at genesis
- [x] Updated faucet initial config in new chain setup scenario for playgrounds

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202627063658779) by [Unito](https://www.unito.io)


runtime/src/constants.rs
```diff
@@ -152,59 +149,83 @@ lazy_static! {
...
-
+    /// The base currency unit.
+    pub const HAPI: Balance = 1;
+    /// One JOY equals 10 Billion HAPIs. Hence we use 10 decimal places in
+    /// currency representation.
+    pub const JOY: Balance = 10_000_000_000 * HAPI;
+    /// The planned initial token issuance of JOY at genesis.
+    const TOTAL_JOY_ISSUANCE_AT_GENESIS: Balance = 1_000_000_000 * JOY;
+    /// A Hypothetical valuation of total issued JOY tokens in USD.
+    const ESTIMATED_JOY_USD_MCAP: Balance = 100_000_000;
```
bedeho: Shouldn't this be 60M?

runtime/src/constants.rs
```diff
@@ -152,59 +149,83 @@ lazy_static! {
...
 
 pub mod currency {
     use super::Balance;
 
-    pub const JOYS: Balance = 250_000_000;
-    pub const DOLLARS: Balance = JOYS / 12500; // 20_000
-    pub const CENTS: Balance = DOLLARS / 100; // 200
-
+    /// The base currency unit.
+    pub const HAPI: Balance = 1;
```
bedeho: Do we actually need this one?

runtime/src/constants.rs
```diff
@@ -152,59 +149,83 @@ lazy_static! {
...
+    /// One JOY equals 10 Billion HAPIs. Hence we use 10 decimal places in
+    /// currency representation.
+    pub const JOY: Balance = 10_000_000_000 * HAPI;
```
bedeho: This name feels a bit opaque, wouldn't something like `BASE_UNITS_PER_JOY` or `HAPI_PER_JOY` be more clear?

runtime/src/constants.rs
```diff
@@ -152,59 +149,83 @@ lazy_static! {
...
+    /// The planned initial token issuance of JOY at genesis.
+    const TOTAL_JOY_ISSUANCE_AT_GENESIS: Balance = 1_000_000_000 * JOY;
```
bedeho: This name does not reflect value, name should be `HAPI_ISSUANCE_AT...`, right?

runtime/src/constants.rs
```diff
@@ -152,59 +149,83 @@ lazy_static! {
...
+    /// A Hypothetical valuation of total issued JOY tokens in USD.
+    const ESTIMATED_JOY_USD_MCAP: Balance = 100_000_000;
+    // Constants used to derive balance configurations of pallets more human readable
+
+    /// Balance estimated worth one USD.
+    pub const DOLLARS: Balance = TOTAL_JOY_ISSUANCE_AT_GENESIS / ESTIMATED_JOY_USD_MCAP;
```
bedeho: - Is this division going to be safe/functional as people start varying those other variables? in partiuclar the mcap?
- Can we find a more direct name? like `HAPIs_IN_A_DOLLAR` or something?

bedeho (changes requested):

runtime/src/constants.rs
mnaamani: Not really.

runtime/src/constants.rs
mnaamani: Yes that would be more accurate, I only picked 100M to have nice round numbers. Will change.

runtime/src/constants.rs
mnaamani: The short name was to make it easy to read where it is used, but if we will use USD,CENTS instead then I don't mind giving it a more clear name.

runtime/src/constants.rs
mnaamani: since I will drop the HAPI constant I will just rename this to `BASE_UNIT_ISSUANCE_AT_GENESIS`

runtime/src/constants.rs
mnaamani: regarding the division, that was another reason I picked multiple of 10.

But parsing this I realize now your original suggestion that we might not need all these DOLLAR, CENT constants make more sense.

It might make more sense to just stick with JOY, or BASE_UNIT, then smaller fractions of those, MILLI_JOY, MICRO_JOY and we should need to have a mental model of how much value 1 JOY might represent. 

This also means these constants will never need to be changed since we don't need to update the total issuance constant or the USD mcap..

mnaamani: .. But I see the benefit in keeping, as long as the the MCAP is "regulary" updated to match real world value, it will keep fees and stakes and bonds more in tune with real world values.

mnaamani: used saturating_div and saturating_mul for safety

mnaamani:
@kdembler has rightly pointed out that the content and storage fees configured through:

```
storage.dataObjectPerMegabyteFee
storage.dataObjectStateBloatBondValue
content.videoStateBloatBondValue
content.channelStateBloatBondValue
```

Are all still configured (at genesis) to default of zero.
They are currently not configurable at genesis, so we need to have the leads call the 'update' extrinsics for each value to give them non-zero values.

Will add to the network-integration tests and setupNewChain scenario.

bedeho:
> They are currently not configurable at genesis

They should be, everything should be, made issue: https://github.com/Joystream/joystream/issues/4077

bedeho:
What is the status of this discussion?
https://github.com/Joystream/joystream/pull/4025#discussion_r921066702

You re still using DOLLARS right now.

I think

1. We should keep dollar denominations, this is indeed simpler to reason about, and that does imply doing updates in the future. It think in the future we can even add a new storage parameter value for this, so it can be changed without a new runtime upgrade per say.
2.  We should have better name than `DOLLARS` or `CENTS`, because it took much longer than needed to figure out what those values meant.

What is your view here?

runtime/src/lib.rs
```diff
@@ -1030,8 +1032,8 @@ parameter_types! {
     pub const MaxModeratorsForCategory: u64 = 20;
     pub const MaxCategories: u64 = 40;
     pub const MaxPollAlternativesNumber: u64 = 20;
-    pub const ThreadDeposit: u64 = 30;
-    pub const PostDeposit: u64 = 10;
+    pub const ThreadDeposit: Balance = 25 * currency::CENTS;
```
bedeho, line 1028: I think one more thing we should try to include in this PR is using the `deposit` function for all bloat bonds, as this is what it seems to be for. I still am note sure exactly how it should be used.

mnaamani:
> What is the status of this discussion? [#4025 (comment)](https://github.com/Joystream/joystream/pull/4025#discussion_r921066702)
> 
I'm using saturating div and mul instead for safety.


> You re still using DOLLARS right now.
> 
> I think
> 
>     1. We should keep dollar denominations, this is indeed simpler to reason about, and that does imply doing updates in the future. It think in the future we can even add a new storage parameter value for this, so it can be changed without a new runtime upgrade per say.
> 
>     2. We should have better name than `DOLLARS` or `CENTS`, because it took much longer than needed to figure out what those values meant.
> 
> 
> What is your view here?

One approach I saw used in acala, is to give these constants a better name as you suggested so 
`BASE_UNITS_IN_ONE_DOLLAR` etc..
and add helper functions that are used in other places to actually set balance amounts eg.

```
fn dollars(d: u32) -> Balance {
     BASE_UNITS_IN_ONE_DOLLAR.saturating_mul(d)
}

fn cents(c: u32) -> Balance {
     BASE_UNITS_IN_ONE_DOLLAR.saturating_div(100).saturating_mul(c)
}
```

runtime/src/lib.rs
mnaamani: Here is an example of how it is used in substrate to configure deposits for the multi-sig pallet:

```
parameter_types! {
	// One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
	pub const DepositBase: Balance = deposit(1, 88);
	// Additional storage item size of 32 bytes.
	pub const DepositFactor: Balance = deposit(0, 32);
}

impl pallet_multisig::Config for Runtime {
	type Event = Event;
	type Call = Call;
	type Currency = Balances;
	type DepositBase = DepositBase;
	type DepositFactor = DepositFactor;
	type MaxSignatories = ConstU16<100>;
	type WeightInfo = pallet_multisig::weights::SubstrateWeight<Runtime>;
}
```

how it is used to store the call that will be later invoked when all signers sign:

```rust
/// Place a call's encoded data in storage, reserving funds as appropriate.
	///
	/// We store `data` here because storing `call` would result in needing another `.encode`.
	///
	/// Returns a `bool` indicating whether the data did end up being stored.
	fn store_call_and_reserve(
		who: T::AccountId,
		hash: &[u8; 32],
		data: OpaqueCall<T>,
		other_deposit: BalanceOf<T>,
	) -> DispatchResult {
		ensure!(!Calls::<T>::contains_key(hash), Error::<T>::AlreadyStored);
		let deposit = other_deposit +
			T::DepositBase::get() +
			T::DepositFactor::get() *
				BalanceOf::<T>::from(((data.encoded_len() + 31) / 32) as u32);
		T::Currency::reserve(&who, deposit)?;
		Calls::<T>::insert(&hash, (data, who, deposit));
		Ok(())
	}
```

Yes I think it does make sense to use it in a similar way for bloat binds like post deposits, thread deposits and proposals discussions posts, as well as the bloat bonds for storage pallet.

runtime-modules/project-token/src/benchmarking.rs
```diff
@@ -17,24 +17,28 @@ use storage::BagId;
...
-const DEFAULT_SPLIT_REVENUE: u32 = 1_000_000;
+const DEFAULT_SPLIT_REVENUE: u64 = 8_000_000;
 const DEFAULT_REVENUE_SPLIT_RATE: Permill = Permill::from_percent(50);
-const DEFAULT_SPLIT_ALLOCATION: u32 = 500_000; // DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE
-const DEFAULT_SPLIT_PARTICIPATION: u32 = 300_000_000;
-const DEFAULT_SPLIT_PAYOUT: u32 = 150_000;
+const DEFAULT_SPLIT_ALLOCATION: u64 = 4_000_000; // DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE
+const DEFAULT_SPLIT_PAYOUT: u64 = 2_000_000; // = (DEFAULT_SPLIT_PARTICIPATION / DEFAULT_TOKEN_ISSUANCE) * DEFAULT_SPLIT_ALLOCATION
+                                             // const DEFAULT_SPLIT_PARTICIPATION: u64 =
+                                             //     DEFAULT_SPLIT_PAYOUT * DEFAULT_TOKEN_ISSUANCE * DEFAULT_SPLIT_ALLOCATION;
```
Lezek123: This comment can be removed now

node/src/chain_spec/content_config.rs
```diff
@@ -19,16 +19,18 @@ pub fn production_config() -> ContentConfig {
...
         max_creator_royalty: Perbill::from_percent(50),
-        min_bid_step: 1,
+        min_bid_step: ExistentialDeposit::get(),
         max_bid_step: 200_000_000_000_000_000,
         platform_fee_percentage: Perbill::from_percent(1),
         auction_starts_at_max_delta: DAYS * 30,
         max_auction_whitelist_length: 100,
         nft_limits_enabled: false,
+        channel_state_bloat_bond_value: 10 * currency::CENTS,
+        video_state_bloat_bond_value: 2 * currency::CENTS,
```
Lezek123, line 33: Perhaps we should also introduce `project_token` pallet config here with some initial value for `bloat_bond`, `min_sale_duration`, `min_revenue_split_duration`, `min_revenue_split_time_to_start` and `sale_platform_fee`

runtime/src/constants.rs
```diff
@@ -152,59 +149,79 @@ lazy_static! {
 
 // Change it when changing the currency constants!
 parameter_types! {
-    pub const ExistentialDeposit: u128 = 1;
+    // TODO: Update invited member initial balance in membership to be greater than this!
+    // and any other relevant places in runtime/pallets.
```
Lezek123: I think that's already done?

runtime/src/lib.rs
```diff
@@ -1090,11 +1089,11 @@ parameter_types! {
     // This should be more costly than `apply_on_opening` fee with the current configuration
     // the base cost of `apply_on_opening` in tokens is 193. And has a very slight slope
     // with the lenght with the length of rationale, with 2000 stake we are probably safe.
```
Lezek123: This part of the comment is outdated now

runtime/src/lib.rs
```diff
@@ -1090,11 +1089,11 @@ parameter_types! {
...
-    pub const MinimumApplicationStake: Balance = 2000;
+    pub const MinimumApplicationStake: Balance = 20 * currency::DOLLARS;
     // This should be more costly than `add_opening` fee with the current configuration
     // the base cost of `add_opening` in tokens is 81. And has a very slight slope
     // with the lenght with the length of rationale, with 2000 stake we are probably safe.
```
Lezek123: This part of the comment is outdated now

runtime/src/tests/fee_tests.rs
```diff
@@ -4,7 +4,8 @@ use crate::Runtime;
 use crate::MAXIMUM_BLOCK_WEIGHT;
 use pallet_transaction_payment::Pallet as TransactionPayment;
 
-// TODO: Fix Fee Tests
+// Polkadot runtime doesn't have an equivalent to this test anymore.
+// Should we fix this?
```
Lezek123: It looks pretty straightforward to fix and I think that's a useful test

runtime/src/tests/locks.rs
```diff
@@ -12,8 +12,8 @@ use staking_handler::StakingHandler;
 fn compatible_stakes_check_passed_successfully() {
     initial_test_ext().execute_with(|| {
         let account_id = account_from_member_id(0);
-        let total_amout = 10000;
-        let stake_amount = 100;
+        let stake_amount = currency::DOLLARS * 100;
+        let total_amout = stake_amount * 100;
```
Lezek123: `total_amout` typo (there are a few of other occurrences of it in this file)

runtime/src/tests/proposals_integration/mod.rs
```diff
@@ -466,7 +471,7 @@ fn set_membership_leader(lead_account_id: AccountId32, lead_id: u64) {
         working_group::OpeningType::Leader,
         StakePolicy {
             stake_amount:
-                <Runtime as working_group::Config<MembershipWorkingGroupInstance>>::MinimumApplicationStake::get(
+                <Runtime as working_group::Config<MembershipWorkingGroupInstance>>::LeaderOpeningStake::get(
```
Lezek123: I think `MinimumApplicationStake` was more correct here.
`LeaderOpeningStake` is the amount that lead needs to stake in order to create a new opening. It can potentially be lower than `MinimumApplicationStake`

runtime/src/tests/proposals_integration/mod.rs
```diff
@@ -481,7 +486,7 @@ fn set_membership_leader(lead_account_id: AccountId32, lead_id: u64) {
         reward_account_id: lead_account_id.clone(),
         description: vec![0u8],
         stake_parameters: StakeParameters {
-            stake: <Runtime as working_group::Config<MembershipWorkingGroupInstance>>::MinimumApplicationStake::get() as
+            stake: <Runtime as working_group::Config<MembershipWorkingGroupInstance>>::LeaderOpeningStake::get() as
```
Lezek123: Same as above

runtime/src/tests/proposals_integration/mod.rs
```diff
@@ -512,7 +517,7 @@ where
 
                 let min_stake = <Runtime as working_group::Config<
                     MembershipWorkingGroupInstance,
-                >>::MinimumApplicationStake::get();
+                >>::LeaderOpeningStake::get();
```
Lezek123: Same as above

runtime/src/tests/proposals_integration/mod.rs
```diff
@@ -523,6 +528,7 @@ where
             }
         }
 
+        // are we missing one proposal when getting max stake?
```
Lezek123: This was fixed in this PR I think?

runtime/src/tests/proposals_integration/working_group_proposals.rs
```diff
@@ -749,9 +760,10 @@ fn run_create_decrease_group_leader_stake_proposal_execution_succeeds<
     initial_test_ext().execute_with(|| {
         let member_id: u64 = create_new_members(1)[0];
         let account_id: [u8; 32] = account_from_member_id(member_id).into();
-        let stake_amount: Balance = 10_000;
+        let stake_amount = <Runtime as working_group::Config<
+            MembershipWorkingGroupInstance,>>::LeaderOpeningStake::get();
```
Lezek123: `MinimumApplicationStake` would be more correct as mentioned in the other comments above

runtime/src/tests/proposals_integration/working_group_proposals.rs
```diff
@@ -929,20 +944,24 @@ fn run_create_slash_group_leader_stake_proposal_execution_succeeds<
...
```
Lezek123: `MinimumApplicationStake` would be more correct as mentioned in the other comments above

runtime/src/tests/proposals_integration/working_group_proposals.rs
```diff
@@ -1520,14 +1531,15 @@ fn run_create_terminate_group_leader_role_proposal_execution_succeeds<
     initial_test_ext().execute_with(|| {
         let member_id: u64 = create_new_members(1)[0];
         let account_id: [u8; 32] = account_from_member_id(member_id).into();
-        let stake_amount = 100_000_u128;
+        let stake_amount = <Runtime as working_group::Config<
+            MembershipWorkingGroupInstance,>>::LeaderOpeningStake::get();
```
Lezek123: `MinimumApplicationStake` would be more correct as mentioned in the other comments above

runtime/src/tests/proposals_integration/working_group_proposals.rs
```diff
@@ -1695,14 +1707,15 @@ fn run_create_terminate_group_leader_role_proposal_with_slashing_execution_succe
...
```
Lezek123: `MinimumApplicationStake` would be more correct as mentioned in the other comments above

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -37,7 +36,10 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
     await new FixtureRunner(addStakingAccountsFixture).run()
 
     const votersStakingAccounts = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
-    await api.treasuryTransferBalanceToAccounts(votersStakingAccounts, voteStake.addn(MINIMUM_STAKING_ACCOUNT_BALANCE))
+    await api.treasuryTransferBalanceToAccounts(
+      votersStakingAccounts,
+      voteStake.add(api.consts.referendum.minimumStake)
```
Lezek123: `voteStake` is already `=== api.consts.referendum.minimumStake`
I think the `MINIMUM_STAKING_ACCOUNT_BALANCE` addition was just a wrong assumption I made at some point about locks not "stacking" on top of each other. I tried changing this line to:
```suggestion
      voteStake
```
And the test is still passing

tests/network-tests/src/fixtures/council/NotEnoughCandidatesWithVotesFixture.ts
```diff
@@ -25,7 +23,7 @@ export class NotEnoughCandidatesWithVotesFixture extends BaseQueryNodeFixture {
     const votersStakingAccounts = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
     await this.api.treasuryTransferBalanceToAccounts(
       votersStakingAccounts,
-      voteStake.addn(MINIMUM_STAKING_ACCOUNT_BALANCE)
+      voteStake.add(this.api.consts.referendum.minimumStake)
```
Lezek123: Same issue as in the comment above

tests/network-tests/src/fixtures/membership/BuyMembershipWithInsufficienFundsFixture.ts
```diff
@@ -37,7 +37,7 @@ export class BuyMembershipWithInsufficienFundsFixture extends BaseFixture {
 
     assert.isBelow(
       balance.toNumber(),
-      membershipFee.add(membershipTransactionFee).toNumber(),
+      membershipFee.add(membershipTransactionFee).add(this.api.consts.balances.existentialDeposit).toNumber(),
```
Lezek123, line 40: Looks like the existential deposit is not currently taken into account when buying membership (so the previous assertion seemed correct)

tests/network-tests/src/flows/membership/membershipSystem.ts
```diff
@@ -10,20 +10,19 @@ export default async function membershipSystem({ api, query, env }: FlowProps):
   debug('Started')
   api.enableDebugTxLogs()
 
+  // testing updates, keeping final membership price significantly larger
+  // than buy membership tx fee, or BuyMembershipWithInsufficienFundsFixture would fail.
```
Lezek123, line 14: Possibly the reason it failed was the existential deposit addition in `BuyMembershipWithInsufficienFundsFixture.ts`

tests/network-tests/src/flows/membership/membershipSystem.ts
```diff
@@ -10,20 +10,19 @@ export default async function membershipSystem({ api, query, env }: FlowProps):
...
   const updates = [
     {
       defaultInviteCount: 10,
-      membershipPrice: new BN(1000),
+      membershipPrice: new BN(100_000_000_000),
       referralCut: 5,
       invitedInitialBalance: new BN(500),
     },
     // BigInt above Int32 case:
```
Lezek123: This comment is now invalidly placed (should probably be above the first case)

tests/network-tests/src/flows/membership/membershipSystem.ts
```diff
@@ -10,20 +10,19 @@ export default async function membershipSystem({ api, query, env }: FlowProps):
...
 
+  // testing updates, keeping final membership price significantly larger
+  // than buy membership tx fee, or BuyMembershipWithInsufficienFundsFixture would fail.
   const updates = [
     {
       defaultInviteCount: 10,
-      membershipPrice: new BN(1000),
+      membershipPrice: new BN(100_000_000_000),
       referralCut: 5,
       invitedInitialBalance: new BN(500),
```
Lezek123: Those low `invitedInitialBalance`'s are not causing any issues, because we're initializing all accounts with `ExistentialDeposit` for the tests, but it may be safer to increase them too

node/src/chain_spec/content_config.rs
```diff
@@ -19,16 +19,18 @@ pub fn production_config() -> ContentConfig {
         max_auction_extension_period: HOURS * 6,
         min_bid_lock_duration: MINUTES * 1,
         max_bid_lock_duration: HOURS * 24,
-        min_starting_price: 1,
+        min_starting_price: ExistentialDeposit::get(),
         max_starting_price: 200_000_000_000_000_000,
         min_creator_royalty: Perbill::from_percent(1),
         max_creator_royalty: Perbill::from_percent(50),
-        min_bid_step: 1,
+        min_bid_step: ExistentialDeposit::get(),
```
Lezek123, line 26: Since we're changing values here to more realistic ones, I think it's worth changing `min_cashout_allowed` and `max_cashout_allowed` too

runtime/src/lib.rs
```diff
@@ -794,16 +793,16 @@ parameter_types! {
...
     // council parameteres
     pub const MinNumberOfExtraCandidates: u64 = 1;
     pub const AnnouncingPeriodDuration: BlockNumber = 14400;
     pub const IdlePeriodDuration: BlockNumber = 57600;
     pub const CouncilSize: u64 = 5;
-    pub const MinCandidateStake: u64 = 11000;
+    pub const MinCandidateStake: Balance = 100 * currency::DOLLARS;
     pub const ElectedMemberRewardPeriod: BlockNumber = 14400;
-    pub const DefaultBudgetIncrement: u64 = 5000000;
+    pub const DefaultBudgetIncrement: Balance = 100 * currency::DOLLARS;
```
Lezek123: This value seems unused, it's initialized by genesis config instead (which is where I think it should be changed along with `councilor_reward`)

Lezek123 (changes requested):
I wasn't commenting on any of the values, since I assumed it's not within the scope of this PR to define the final ones.

node/src/chain_spec/content_config.rs
mnaamani: Good idea, I created an issue https://github.com/Joystream/joystream/issues/4095 we can work on it in a separate PR.

runtime/src/tests/fee_tests.rs
mnaamani: fixed in e98e2bd3c2450266fe756151c442861c62e386be and dabca1dee23fd29a962cc718b4eb1014d3c97b7b

runtime/src/tests/proposals_integration/mod.rs
mnaamani: yes it was a note to self, will remove

tests/network-tests/src/flows/membership/membershipSystem.ts
mnaamani: It was actually failing if i remember correctly because the last update case had the membership price lower than existential deposit.

tests/network-tests/src/fixtures/membership/BuyMembershipWithInsufficienFundsFixture.ts
mnaamani: I'm adding it because the account is initialized with ed. Are you suggesting removing this addition?

Or is your point that the implementation of buy_membership() is flawed?

runtime/src/lib.rs
mnaamani: Yes indeed.

tests/network-tests/src/fixtures/membership/BuyMembershipWithInsufficienFundsFixture.ts
Lezek123: All I'm saying is that the if the account has a balance that is >= `membershipFee + membershipTransactionFee`, but `< membershipFee + membershipTransactionFee + existentialDeposit` the tx will still pass, so I don't think the assertion is correctly checking the balance before testing for expected failure.

You could say the implementation is a little flawed too, because if someone has a balance `== membershipFee + membershipTransactionFee` then he will only pay `membershipFee + membershipTransactionFee - existentialDeposit`, so slightly less than someone who has a higher balance, because slashing will always leave at least `existentialDeposit` funds in the account and since in the runtime we'll be trying to perform a slash that would result in balance == 0, the `existentialDeposit` part of the slash will just be ignored (returned as `not_slashed`, which is a second value in the tuple retruned by `balances::slash`)

Lezek123: That property of `balances::slash` may actually have not been taken into account in some other places in the runtime too (for example, when buying nfts), I'll create a separate issue about it

Lezek123 (approved):
LGTM

tests/network-tests/src/fixtures/membership/BuyMembershipWithInsufficienFundsFixture.ts
mnaamani: Issue mentioned above https://github.com/Joystream/joystream/issues/4105

Issue #4029, opened 2022-07-14
ChannelCreated - expose channel module account


Lezek123:
Expose channel module account in `ChannelCreated` event to allow query node to map it and frontend applications to easily retrieve it.

Although the channel accounts can be "calculated" on the frontend with a script like:

```typescript
import { createType } from '@joystream/types'
import { u8aConcat, u8aFixLength } from '@polkadot/util'

const typeId = 'modl'
const moduleId = 'mContent'
const moduleAccType = createType('Bytes', 'CHANNEL')
const channelId = createType('u64', 1)

createType(
  'AccountId',
  u8aFixLength(u8aConcat(typeId, moduleId, moduleAccType.toU8a(false), channelId.toU8a()), 8 * 32, true)
)
```

This is quite complex and sensitive to:

* changes in substrate
* change of content module id in the runtime (currently `mContent`)
* change of `into_sub_account` parameters in the runtime

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202604817497772) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4030

PR #4030, opened 2022-07-14
ChannelCreated event - add channel account

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4029

**Scope:**
- updated `ChannelCreated` event in the runtime
- updated types & metadata
- updated mappings
- updated network tests


tests/network-tests/src/flows/clis/createChannel.ts
```diff
@@ -48,6 +50,22 @@ export default async function createChannel({ api, query }: FlowProps): Promise<
...
+          'modl',
+          'mContent',
+          createType('Bytes', 'CHANNEL').toU8a(false),
+          createType('u64', channelId).toU8a()
+        ),
+        32 * 8,
+        true
+      )
+    )
+    .toString()
```
ondratra, line 67: Just curious how you found out what the parameters to `u8aConcat` should be? I mean, great work because I don't know where I would start to look for them.

ondratra (approved):

tests/network-tests/src/flows/clis/createChannel.ts
Lezek123: Just following the implementation of `ModuleId::into_sub_account_truncating` in rust and then basically some trial and error and hex comparisons in order to get the address that matches the one generated by the runtime. I had some idea how different types are usually encoded, so fortunately it didn't take that long.

PR #4034, opened 2022-07-15
Carthage playground

mnaamani:
Some fixes for setup and linting scripts.
- On ubuntu 22.04 cannot install python package, you have to be more specific (python2 or python3)
- Factored out the cargo-checks into its own script because we also run ts linting in the git-hook, and in some CI workflows we don't bother installing yarn dependencies so it was [breaking joystream-node workflow](https://github.com/Joystream/joystream/runs/7345051271?check_suite_focus=true#step:6:3526)
- Deleted a script which I had checked-in at some point accidentally.

### DIFF
--- a/devops/git-hooks/pre-push
+++ b/devops/git-hooks/pre-push
@@ -4,19 +4,6 @@ set -e
 SCRIPT_PATH="$(dirname "${BASH_SOURCE[0]}")"
 cd $SCRIPT_PATH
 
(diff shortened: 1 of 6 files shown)


Lezek123 (approved):
LGTM
Tested with both `ubuntu:20.04` and `ubuntu:22.04` docker images, works as expected

PR #4036, opened 2022-07-15
Content benchmarking: Curator groups

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4037


runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,226 @@
...
+        assert!(group == CuratorGroup::create(true, &permissions_by_level));
+        assert_last_event::<T>(Event::<T>::CuratorGroupCreated(group_id).into());
+    }
+
+    update_curator_group_permissions {
+        let a in 0 .. (T::MaxKeysPerCuratorGroupPermissionsByLevelMap::get() as u32);
+
+        let lead_account = insert_content_leader::<T>();
+        let group_id = setup_worst_case_curator_group_with_curators::<T>(
+            T::MaxNumberOfCuratorsPerGroup::get()
```
L3pereira: I think It's better to add this, as you cannot create more curators than the min of these two limits (workers and curators), this can differ later in runtime and fail the test.
```suggestion
min(<T as working_group::Config<ContentWorkingGroupInstance>>::MaxWorkerNumberLimit::get(),
        T::MaxNumberOfCuratorsPerGroup::get()),
```

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,226 @@
...
+        assert_last_event::<T>(Event::<T>::CuratorGroupPermissionsUpdated(
+            group_id,
+            permissions_by_level
+        ).into());
+    }
+
+    set_curator_group_status {
+        let lead_account = insert_content_leader::<T>();
+        let group_id = setup_worst_case_curator_group_with_curators::<T>(
+            T::MaxNumberOfCuratorsPerGroup::get()
```
L3pereira: I think It's better to add this, as you cannot create more curators than the min of these two limits (workers and curators), this can differ later in runtime and fail the test.
```suggestion
min(<T as working_group::Config<ContentWorkingGroupInstance>>::MaxWorkerNumberLimit::get(),
        T::MaxNumberOfCuratorsPerGroup::get()),
```

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,226 @@
...
+        let group = Pallet::<T>::curator_group_by_id(group_id);
+        assert!(!group.is_active());
+        assert_last_event::<T>(Event::<T>::CuratorGroupStatusSet(group_id, false).into());
+    }
+
+    add_curator_to_group {
+        let lead_account = insert_content_leader::<T>();
+        let permissions = worst_case_channel_agent_permissions();
+        let group_id = setup_worst_case_curator_group_with_curators::<T>(
+            T::MaxNumberOfCuratorsPerGroup::get() - 1
```
L3pereira: I think It's better to add this, as you cannot create more curators than the min of these two limits (workers and curators), this can differ later in runtime and fail the test.
```suggestion
min(<T as working_group::Config<ContentWorkingGroupInstance>>::MaxWorkerNumberLimit::get(),
        T::MaxNumberOfCuratorsPerGroup::get()),
```

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,226 @@
...
+    verify {
+        let group = Pallet::<T>::curator_group_by_id(group_id);
+        assert_eq!(group.get_curators().get(&curator_id), Some(&permissions));
+        assert_last_event::<T>(Event::<T>::CuratorAdded(group_id, curator_id, permissions).into());
+    }
+
+    remove_curator_from_group {
+        let lead_account = insert_content_leader::<T>();
+        let group_id = setup_worst_case_curator_group_with_curators::<T>(
+            T::MaxNumberOfCuratorsPerGroup::get()
```
L3pereira: I think It's better to add this, as you cannot create more curators than the min of these two limits (workers and curators), this can differ later in runtime and fail the test.
```suggestion
min(<T as working_group::Config<ContentWorkingGroupInstance>>::MaxWorkerNumberLimit::get(),
        T::MaxNumberOfCuratorsPerGroup::get()),
```

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,226 @@
...
+    insert_distribution_leader, insert_storage_leader, member_funded_account,
+    setup_worst_case_curator_group_with_curators, worst_case_channel_agent_permissions,
+    worst_case_content_moderation_actions_set, CreateAccountId, RuntimeConfig, CURATOR_IDS,
+    DEFAULT_MEMBER_ID,
+};
+
+benchmarks! {
+    where_clause {
+        where
+            T: RuntimeConfig,
```
L3pereira, line 35: We should use Config, is set this way in the rest of the project
```suggestion
            T: Config,
```

L3pereira (changes requested):
@Lezek123 solve the conflicts first, please.

runtime-modules/content/src/benchmarks/benchmarking.rs
Lezek123: This is just a simplification, so instead of writing:
```rust
              T: balances::Config,
              T: membership::Config,
              T: storage::Config,
              T: working_group::Config<StorageWorkingGroupInstance>,
              T: working_group::Config<DistributionWorkingGroupInstance>,
              T: working_group::Config<ContentWorkingGroupInstance>,
              T: Config 
```
In many places, we can just use this `T: RuntimeConfig` "alias trait" (in which we define all the traits that are needed).
It seems to work just fine, do you see any issues with this approach?

L3pereira (approved):
LGTM

Issue #4049, opened 2022-07-18
Content benchmarking: Channel CUD Group


L3pereira:
- #4018

Extrinsics to benchmark

- create_channel
- update_channel
- delete_channel

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202621184029243) by [Unito](https://www.unito.io)

PR #4061, opened 2022-07-18
Channel cud group benchmarking

L3pereira:
* #4049 

Extrinsics to benchmark

* create_channel
* update_channel
* delete_channel

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202627117867209) by [Unito](https://www.unito.io)


runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,264 @@
...
+
+        let assets_to_remove =
+            Storage::<T>::get_next_data_object_ids(c as usize);
+
+        let params = generate_channel_creation_params::<T>(
+            storage_wg_lead_account_id,
+            distribution_wg_lead_account_id,
+            max_obj_size.saturating_mul((2 * MAX_OBJ_NUMBER).into()),
+            (2 * MAX_OBJ_NUMBER).into(),
+            a,
```
Lezek123: You can use the maximum allowed number of collaborators here as a worst-case-scenario starting point

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,264 @@
...
+            T::StorageBucketsPerBagValueConstraint::get().max() as u32,
+            T::DistributionBucketsPerBagValueConstraint::get().max() as u32,
+            c,
```
Lezek123: You can use max allowed number of assets here as a worst-case-scenario starting point.

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,264 @@
...
+            max_obj_size,
+        );
+
+        Pallet::<T>::create_channel(
+            sender.into(),
+            channel_owner,
+            params.clone()).unwrap();
+
+        let permissions = BTreeSet::from_iter([
```
Lezek123: I'm not sure how the size of this set will affect the weight, but probably worth using either all possible variants or all minus one (in order to make them different from the current permissions)

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -0,0 +1,264 @@
...
+        .map(|(member_id, _)|{
+            (member_id, permissions.clone())
+        }).collect::<BTreeMap<_, _>>();
+
+        let assets_to_upload = StorageAssets::<T> {
+                expected_data_size_fee:
+                    Storage::<T>::data_object_per_mega_byte_fee(),
+                object_creation_list: create_data_object_candidates_helper(
+                    1,
+                    (b - 1).into(),
```
Lezek123: Why are you using `(b-1)` instead of `b`?

Lezek123 (changes requested):
I think for `update_channel` (that may be the case for `delete_channel` also) we need 2-3 different benchmark functions depending on whether the actor performing the update is channel owner or collaborator/curator.

For example: executing channel update as curator will take a more complex path, because it will need to validate whether the curator has all the required channel agent permissions to perform the update, which is especially relevant when the curator is updating `channel.collaborators` (because then we need to additionally compare them with the existing set, extract all the changes and for each of them validate whether the curator is allowed to make this change)

Edit: Actually, another issue is that when the channel is updated/deleted by a curator, we also need to read a curator group from the runtime state, which is potentially a large storage object (with `MaxCuratorsPerGroup` curators, each having a full set of permissions)

Edit2: See https://github.com/Joystream/joystream/pull/4061#issuecomment-1189135679

Lezek123:
> I think for update_channel (that may be the case for delete_channel also) we need 2-3 different benchmark functions depending on whether the actor performing the update is channel owner or collaborator/curator.

Actually since we cannot distinguish between member channel owner and member channel collaborator (as both are `ContentActor::Member`) without reading the storage unless we change the content pallet dramatically, I suggest we just always use `ContentActor::Curator` for benchmarking, since it's the worst-case-scenario actor.

bedeho:
> since it's the worst-case-scenario actor.

Why is it the worst? is it some artefact having to do with checking permissions or something? if so, difference is very small, no?

Lezek123:
> Why is it the worst? is it some artefact having to do with checking permissions or something? if so, difference is very small, no?

As I mentioned in the first comment, because authenticating a curator requires loading a curator group, which is potentially a very big storage object, so that would have a noticeable effect on the weight.

Additionally, when `update_channel` is being done by curator/collaborator there is also additional code path triggered which extracts all changes in `channel.collaborators` and then iterates over them and verifies that the actor has sufficient permissions to perform them (although this is probably less significant).

runtime-modules/content/src/benchmarks/benchmarking.rs
Lezek123: The previous review comment is still not addressed. I suggest using a set like:
```suggestion
        let permissions: ChannelAgentPermissions = worst_case_channel_agent_permissions().into_iter().skip(1).collect();
```

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -542,3 +640,180 @@ where
...
+        )?;
+    }
+    Ok(group_id)
+}
+
+#[allow(clippy::type_complexity)]
+fn setup_worst_case_scenario_curator_channel<T>(
+    objects_num: u32,
+    bucket_objs_size_limit: u64,
+    bucket_objs_number_limit: u64,
```
Lezek123: I don't think it's necessary to have those as parameters, it adds complexity and doesn't really affect the weight.
I think it's best to just use fixed limits which are high enough to work for all benchmarking scenarios, like: https://github.com/Lezek123/substrate-runtime-joystream/blob/videos-benchmarking/runtime-modules/content/src/benchmarks/mod.rs#L557

Lezek123 (changes requested):
1. One thing I realized while working on video benchmarking is that you need to setup some non-zero bloat bonds (see: https://github.com/Lezek123/substrate-runtime-joystream/blob/videos-benchmarking/runtime-modules/content/src/benchmarks/mod.rs#L795) & data fee in order to account for the worst case scenario.

    Methods like `balances::Pallet::<T>::slash()` or `balance::Pallet::<T>::transfer()` usually skip the execution of any costly operations if the amount is `== 0`, so when all bloat bonds and fees are set to `0` you will get an understated weight.

2. Even though the metadata size does not have a big impact on the weight I think we should still use metadata with `Some` value, as this could trigger some additional code paths, for example:
    ```
        if params.new_meta.is_some() {
            required_permissions.push(ChannelActionPermission::UpdateChannelMetadata);
        }
    ```
    Plus in general using some large value in the metadata filed would give us more confidence that the weight is not understated
3. This may be done once you introduce storage bag witness parameter, but I think it would be very useful to split channel update benchmarking into two cases: `channel_update_with_assets` and `channel_update_without_assets`. In the second case the runtime should just ignore the value of `channel_bag_witness` and skip all the logic related to verifying/performing any bag operation. For videos I even split the update method into 4 separate scenarios. as you can see here: https://github.com/Lezek123/substrate-runtime-joystream/blob/videos-benchmarking/runtime-modules/content/src/benchmarks/benchmarking.rs#L285

L3pereira:
- #4064 implemented

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -183,7 +190,75 @@ benchmarks! {
...
+    channel_update_without_assets {
+
+        let a in 1 .. T::MaxNumberOfCollaboratorsPerChannel::get(); //max colaborators
+
+        let b in (T::StorageBucketsPerBagValueConstraint::get().min as u32) ..
+         (T::StorageBucketsPerBagValueConstraint::get().max() as u32);
+
+        let c in
+            (T::DistributionBucketsPerBagValueConstraint::get().min as u32) ..
+            (T::DistributionBucketsPerBagValueConstraint::get().max() as u32);
```
Lezek123, line 202: For `channel_update_without_assets` and `channel_delete_without_assets` we need some modifications in the content pallet, so that the weights don't depend on storage/distribution buckets number in that case.

Basically you need to put everything related to data object uploading/removing in:
```rust
if params.assets_to_upload.is_some() || !params.assets_to_remove.is_empty() {
  // ...
}
```
Especially the `upload_and_delete_data_objects` call

You can also make `channel_bag_witness` optional and only check it if `params.assets_to_upload.is_some() || !params.assets_to_remove.is_empty()`

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -183,7 +190,75 @@ benchmarks! {
         );
     }
 
-    delete_channel {
+    channel_update_without_assets {
```
Lezek123, line 193: I think you should also introduce `let d in 1 .. MAX_BYTES_METADATA;` benchmark parameter here for consistency

Lezek123 (changes requested):

runtime-modules/content/src/benchmarks/benchmarking.rs
L3pereira: do we need this "if" in delete_channel? 
The `num_objects_to_delete` is used here

https://github.com/Joystream/joystream/blob/12cd1645806644280b15269c3afe51f82730f77c/runtime-modules/content/src/lib.rs#L730

and is validating only if the number of the objects in the bag, matches the number we want to delete

https://github.com/Joystream/joystream/blob/12cd1645806644280b15269c3afe51f82730f77c/runtime-modules/content/src/lib.rs#L3288-L3291

should we always remove the bag? even if there are no objects?
I need to write a test where the objects are 0 but the fixture is validating at the end, if the bag exists

Lezek123: > should we always remove the bag? even if there are no objects?

Oh, right, I kept having `delete_video` on my mind, where there was no need to remove the bag.

Here we need to remove the entire channel bag, so we always need to update all storage/distribution buckets anyway. This means there's really no point in splitting `delete_channel` to "with_assets" and "without_assets" case. Sorry for a bad suggestion.

Lezek123 (approved):
Perfect

PR #4070, opened 2022-07-20
Carthage playground updates

mnaamani:
Some minor deployment tweaks

- Increase aws instance to startup timeout
~- Drop manual install of docker on aws instance, basic version is sufficient~
- Replaced docker install step from https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository
- Rename config.json (containing all the endpoint urls) to network-config.json
- Remove `--log` arg for joystream-node service (incorrect for new version of substrate node)

Running a test deployment from this branch here: https://github.com/Joystream/joystream/runs/7426877999?check_suite_focus=true - FAILED

trying again.. https://github.com/Joystream/joystream/runs/7429906467?check_suite_focus=true


devops/aws/roles/rpc/tasks/main.yml
```diff
@@ -109,7 +109,7 @@
 - name: Create config.json to serve as Caddy endpoint
   copy:
     content: '{{ all_services | from_yaml | to_json }}'
-    dest: '/home/ubuntu/config.json'
+    dest: '/home/ubuntu/network-config.json'
```
Lezek123, line 112: Should this be also changed in `/devops/aws/deploy-playground-playbook.yml` and `/devops/aws/templates/Playground-Caddyfile.j2`?

mnaamani: Yeah to have same setup, although the playbook and the caddy template it is using are consistent.
This changed turned out to not be needed at all infact but I will keep, and update the files you pointed out as it is clearer.

mnaamani:
Unfortunately the deployment is still failing, I will have to debug again.
I think there is still some point in this script https://github.com/Joystream/joystream/pull/4070/files#diff-967ebb663b6cec0cc6e8ce9df7c48fd6121403bac33aa533088cae4eeb755787R122 that is either stalling or asking for user-input interactively.

devops/aws/create-ec2-instance.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
Lezek123: This just looks like a leftover test script, I don't think it should be committed?

Lezek123 (changes requested):

devops/aws/create-ec2-instance.sh
mnaamani: yes I thought I had removed it already :)

mnaamani: removed

Lezek123 (approved):
LGTM

Issue #4064, opened 2022-07-19
Add number of storage/distribution buckets as a parameter to content extrinsics that modify channel bag


Lezek123:
Currently `update_channel` and `delete_channel` extrinsics have very high weights (see: https://github.com/Joystream/joystream/pull/4061). The reason for this is that each update of a channel bag propagates into all storage/distribution buckets assigned to this bag.

This give us, for example, a `delete_channel` weight function like this:

```
	fn delete_channel(a: u32, ) -> Weight {
		(1_774_761_000 as Weight)
			// Standard Error: 33_000
			.saturating_add((11_427_000 as Weight).saturating_mul(a as Weight))
			.saturating_add(T::DbWeight::get().reads(127 as Weight))
			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
			.saturating_add(T::DbWeight::get().writes(123 as Weight))
			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(a as Weight)))
	}
```

_(notice 127 reads and 123 writes regardless of the provided parameters)_

Currently we don't provide number of storage/distribution buckets as a parameter for `update_channel` / `delete_channel` , so we have to use the worst case scenario (which is 100 distribution buckets and 20 storage buckets).
Adding those parameters should mitigate this problem.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202630930638977) by [Unito](https://www.unito.io)

Lezek123:
Other extrinsics that are affected by this issue:
- `create_video`
- `update_video`
- `delete_video`
- `delete_video_assets_as_moderator`
- `delete_video_as_moderator`
- `delete_channel_assets_as_moderator`

Lezek123:
Reference implementation for `create_video`: https://github.com/Joystream/joystream/pull/4073/commits/726a47a5cc6744b20c236d4f722cb2792e877ec3

Issue #4052, opened 2022-07-18
Content benchmarking: Video CUD Group


L3pereira:
- #4018

Extrinsics to benchmark

* create_video
* update_video
* delete_video

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202621091608300) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4073

PR #4073, opened 2022-07-21
Videos benchmarking

Lezek123:
Initial work on https://github.com/Joystream/joystream/issues/4052 and https://github.com/Joystream/joystream/issues/4064

Based on top of https://github.com/Joystream/joystream/pull/4036


runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -399,6 +404,376 @@ benchmarks! {
...
+    // Worst case scenario: initial state - EnglishAuction
+    create_video_with_nft {
+        let a in 1..T::MaxNumberOfAssetsPerVideo::get();
+        let b in
+            (T::StorageBucketsPerBagValueConstraint::get().min as u32)
+            ..(T::StorageBucketsPerBagValueConstraint::get().max() as u32);
+        let c in 2..MAX_AUCTION_WHITELIST_LENGTH;
+        let d in 1..MAX_BYTES_METADATA;
+
+        let (curator_account_id, actor, channel_id, params) = prepare_worst_case_scenario_video_creation_parameters::<T>(
```
dobertRowneySr, line 454: You should set the channel pausable features to all but `VideoCreation` in order to maximize the 
complexity of the call `channel.ensure_feature_not_paused::<T>(PausableChannelFeature::VideoCreation)?;` inside `create_video`

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -399,6 +404,376 @@ benchmarks! {
...
+        }
+        assert_last_event::<T>(Event::<T>::VideoUpdated(
+            actor,
+            video_id,
+            params,
+            BTreeSet::new()
+        ).into());
+    }
+
+    update_video_with_assets_with_nft {
```
dobertRowneySr, line 673: same here: the worst pre-existing state for this extrinsics is when the channel pausable features are all set except the necessary ones

dobertRowneySr (changes requested):
- worst case scenario for pausable channels is not set up
- I assume that it is still the case that having a curator channel is worst (weight wise) than 
   member channel for verifying the permissions (as I noticed in my previous benchmarks PR)

dobertRowneySr (changes requested):
you can find in the current `carthage-content-benchmarking` a helper that set all channel pausable features except desired

dobertRowneySr (approved):
LGTM

Issue #3910, opened 2022-06-16
Explore: Multisig support


bedeho:
# Background
It would be useful to allow `Sudo` and other key user accounts to be controllable with multisig.

# Proposal
Review what it involves introducing multisig support, likely with the use of the `multisig` FRAME pallet, and in particular confirm how invasive the changes are in the runtime and node to allow use of it for example for `Sudo` and user authentication currently guarded by `ensure_signed`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202458082542259) by [Unito](https://www.unito.io)

Lezek123:
`multisig` pallet along with the `sudo` testing scripts introduced in https://github.com/Joystream/joystream/pull/4076

PR #4076, opened 2022-07-22
Multisig pallet implementation + sudo multisig support in api-scripts

Lezek123:
Related issue: https://github.com/Joystream/joystream/issues/3910

- Implemented `pallet-multisig` in the runtime (for `DepositBase` and `DepositFactor` params I followed the Substrate implementation)
- Added an example test script in `utils/api-scripts.sh` to ilustrate how to set multisig sudo key and then use multisig to sign sudo transactions: https://github.com/Lezek123/substrate-runtime-joystream/blob/carthage-multisig-support/utils/api-scripts/test-sudo-multisig.sh (it requires a node to be running first, for example, started via `./scripts/run-dev-chain.sh`)


runtime/src/lib.rs
```diff
@@ -1437,6 +1437,25 @@ impl pallet_vesting::Config for Runtime {
...
+}
+
+impl pallet_multisig::Config for Runtime {
+    type Event = Event;
+    type Call = Call;
+    type Currency = Balances;
+    type DepositBase = DepositBase;
+    type DepositFactor = DepositFactor;
+    type MaxSignatories = MaxSignatories;
+    type WeightInfo = pallet_multisig::weights::SubstrateWeight<Runtime>;
```
mnaamani: perhaps use the benchmarks we will compute instead:
```suggestion
    type WeightInfo = weights::pallet_multisig::SubstrateWeight<Runtime>;
```

utils/api-scripts/src/helpers/extrinsics.ts
```diff
@@ -40,17 +41,122 @@ export class ExtrinsicsHelper {
...
+    signers: KeyringPair[],
+    keys: string[],
+    tx: SubmittableExtrinsic<'promise'>,
+    errorMsg?: string
+  ): Promise<ISubmittableResult> {
+    if (!signers.length || !keys.length) {
+      throw new Error('sendAndCheckMulisig: signers.length and keys.length must be > 0')
+    }
+    const otherKeys = (s: KeyringPair) =>
+      sortAddresses(
```
mnaamani, line 81: I was going to ask why sorting is needed then I saw the pallet implementation which checks `ensure_sorted_and_insert`. And now I realize the sorting is to ensure the multi-address derived from the accounts depend on the order :)

mnaamani (changes requested):
Great stuff, tested it out works as advertised.
The updated test scripts where very helpful also as an example of how to use it.
Left just change request to update the weights for the pallet to come from the generated ones for consistency.

Needs a merge conflict fix for the types gen.

mnaamani (approved):

Issue #4080, opened 2022-07-25
`try_mutating_storage_state` refactorization plan


Lezek123:
Original issue: https://github.com/Joystream/joystream/issues/3412

## `try_mutating_storage_state` method

Currently we have 3 separate actions that can be executed using a single `try_mutating_storage_state` method:
```rust
// Create dynamic bag (and upload initial data objects)
try_mutating_storage_state(
    account_id,
    BagOperation::<T> {
        bag_id,
        params: BagOperationParams::<T>::Create(
            objects_to_upload,
            storage_buckets,
            distribution_buckets,
        ),
    },
)

// Update bag (add and/or remove data objects)
try_mutating_storage_state(
  account_id,
  BagOperation::<T> {
    bag_id,
    params: BagOperationParams::<T>::Update(objects_to_upload, objects_to_remove),
  },
)

// Remove bag (and all data objects)
try_mutating_storage_state(
    account_id,
    BagOperation::<T> {
        bag_id,
        params: BagOperationParams::<T>::Delete,
    },
)
```

I suggest removing the `BageOperationParams` enum and `BagOperation` struct and using 3 separate methods instead:
```rust
// Create bag (and upload initial data objects)
try_performing_bag_creation(account, data_objects, storage_buckets, distribution_buckets)

// Update bag (add and/or remove data objects)
try_performing_bag_update(account, objects_to_upload, objects_to_remove)

// Remove bag (and all data objects)
try_performing_bag_removal()
```

All logic specific to a given `BagOperation` should be moved to one of those corresponding methods.

## Other methods that depend on `BagOperation`

There are other functions that execute different logic depending on provided `BagOperation`, which should also be refactored:

### `fn retrieve_dynamic_bag(bag_op: &BagOperation<T>)`
This function creates a new dynamic bag (if already exists - throws an error) and returns it in case if `bag_op.params` is of `BagOperationParams::Create` type.
In all other cases it just returns `Self::ensure_bag_exists(&bag_op.bag_id)`.

**Possible refactorization:** Change it to `new_dynamic_bag(storage_buckets, distribution_buckets)` and only leave the part that's inside:
```rust
if let BagOperationParams::<T>::Create...
```
For other purposes `Self::ensure_bag_exists` can be just used directly.

### `fn construct_objects_to_upload(op: &BagOperation<T>)`
In case of `BagOperationParams::Create` and `BagOperationParams::Update` it just returns `Self::construct_objects_from_list(assets_to_upload)`.
[...]

PR #4082, opened 2022-07-26
Storage pallet: `try_mutating_storage_state` refactorization

Lezek123:
Refactorization according to https://github.com/Joystream/joystream/issues/4080

**Also:**
- Dropped `BagObjectsChanged` event since it was not used by the QN and this data can be obtained from other events
- Added set of uploaded objects ids to `DataObjectsUpdated` event
- Refactored `construct_objects_to_remove` (not mentioned in the original issue)
- Small adjustments to CLI and QN mappings

**Issues fixed:**
- It's no longer possible to sidestep storage buckets per bag policy by providing some `storage_buckets` that don't exist or don't have enough capacity to store the new data objects (during dynamic bag creation)
- Storage bucket voucher updates were fixed (this bug was described in https://github.com/Joystream/joystream/issues/4080)
- No redundant read/write operations for distribution buckets (only during bag creation / removal)
- `get_next_data_object_ids` method is no longer needed. `create_dynamic_bag`, `upload_data_objects` and `upload_and_delete_data_objects` now return the set of created data object ids as part of the `Result`


dobertRowneySr (commented):
LGTM overall, but how about benchmarks?

dobertRowneySr (approved):
LGTM

Issue #3689, opened 2022-04-29
Update CLI to work in Carthage


bedeho:
# Background
We will both be updating the query node and runtime in substantial ways.

# Proposal
Update CLI to work in Carthage, with the priority of just maintaining functionality in the already existing feature set, and perhaps **even disabling things which are not critical for integration testing code or usage**, but which is expensive to fix. Fees will be non-zero, covering both size of extrinsic parameters, and also computational costs. This may warrant even some redesigns to clarify how choices impact fees, but this should be left out of scope in this work, that can be a future enhancement. Work should be done in two stages

1. Map out scope of changes neeeded to be done, write up reply here and share with release team.
1. Do the work.

Note: Ask bwhm about what truly absolute minimal scope is.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202202568644150) by [Unito](https://www.unito.io)

zeeshanakram3:
Addressed in #4084

PR #4084, opened 2022-07-27
Carthage cli rework

zeeshanakram3:
implements https://github.com/Joystream/joystream/issues/3689


cli/src/Types.ts
```diff
@@ -210,6 +211,15 @@ export type ChannelCreationInputParameters = Omit<IChannelMetadata, 'coverPhoto'
...
 }
 
+export type ModerationPermissionsByLevelInputParameters = {
+  channelPreviledgeLevel: number
+  permissions: (
+    | Exclude<ContentModerationAction['type'], 'ChangeChannelFeatureStatus' | 'DeleteVideoAssets'>
+    | { 'DeleteVideoAssets': boolean }
+    | { 'ChangeChannelFeatureStatus': ContentModerationAction['asChangeChannelFeatureStatus']['type'] }
+  )[]
+}
```
Lezek123: There's a small inconsistency between this type and the actual json input.
The type defines non-value enum variants as simple strings, for example:
```
permissions: 'DeleteVideo'
```
While the json schema only accepts objects will `null` values:
```
permissions: { DeleteVideo: null }
```

Ideally that would be caught by `JsonSchema<ModerationPermissionsByLevelInputParameters>` type constraint in `/cli/src/schemas/ContentDirectory.ts`, but I guess `JsonSchema<T>` is not sophisticated enough to understand such complex unions.

cli/README.md
```diff
@@ -98,28 +98,28 @@ When using the CLI for the first time there are a few common steps you might wan
...
 - [`joystream-cli content:createChannel`](#joystream-cli-contentcreatechannel)
-- [`joystream-cli content:createChannelCategory`](#joystream-cli-contentcreatechannelcategory)
 - [`joystream-cli content:createCuratorGroup`](#joystream-cli-contentcreatecuratorgroup)
 - [`joystream-cli content:createVideo`](#joystream-cli-contentcreatevideo)
 - [`joystream-cli content:createVideoCategory`](#joystream-cli-contentcreatevideocategory)
 - [`joystream-cli content:curatorGroup ID`](#joystream-cli-contentcuratorgroup-id)
 - [`joystream-cli content:curatorGroups`](#joystream-cli-contentcuratorgroups)
 - [`joystream-cli content:deleteChannel`](#joystream-cli-contentdeletechannel)
-- [`joystream-cli content:deleteChannelCategory CHANNELCATEGORYID`](#joystream-cli-contentdeletechannelcategory-channelcategoryid)
+- [`joystream-cli content:deleteChannelAsModerator`](#joystream-cli-contentdeletechannelasmoderator)
```
Lezek123, line 108: The readme seems outdated, there are quite a few diffs when I'm regenerating it with `yarn workspace @joystream/cli oclif-dev readme`

cli/src/commands/content/createCuratorGroup.ts
```diff
@@ -1,18 +1,52 @@
...
   static flags = {
+    status: flags.enum({
+      required: true,
+      description: `Status of newly created Curator Group: (${CURATOR_GROUP_CREATION_CONTEXTS.join('/')})`,
+      options: [...CURATOR_GROUP_CREATION_CONTEXTS],
+    }),
+    permissions: flags.string({
+      char: 'p',
+      required: true,
+      description: `Path to JSON file containing moderation permissions by channel previledge to use as input`,
```
Lezek123:
```suggestion
      description: `Path to JSON file containing moderation permissions by channel privilege level to use as input`,
```

cli/src/commands/content/deleteChannelAsModerator.ts
```diff
@@ -0,0 +1,107 @@
...
+    }),
+    rationale: flags.string({
+      char: 'r',
+      required: true,
+      description: 'Reason of deleting the channel by moderator',
+    }),
+    ...ContentDirectoryCommandBase.flags,
+  }
+
+  async getDataObjectsInfoFromQueryNode(channelId: number): Promise<[string, BN][]> {
```
Lezek123: Perhaps worth factoring out those common methods to `ContentDirectoryCommandBase` (for example), but it's not a high priority

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+    assetIds: flags.build({
+      parse: (value: string) => {
+        const arr = value.split(',').map((v) => {
+          if (!/^-?\d+$/.test(v)) {
+            throw new Error(`Expected comma-separated integers, but received: ${value}`)
+          }
+          return v
+        })
+        return arr
+      },
```
Lezek123: I think in all places we still use this overcomplicated solution we can just replace it with:
```typescript
flags.integer({
      required: true,
      multiple: true,
      // ...
    }),
```

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+    }),
+    rationale: flags.string({
+      char: 'r',
+      required: true,
+      description: 'Reason for removing the channel assets by moderator',
+    }),
+    ...ContentDirectoryCommandBase.flags,
+  }
+
+  async getDataObjectsInfo(videoId: number, assetIds: string[]): Promise<number[]> {
```
Lezek123: This should be `channelId`, not `videoId`

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+    const dataObjects = (await this.getQNApi().dataObjectsByChannelId(videoId.toString())).map((o) => o.id)
```
Lezek123: This set will include objects that belong to videos in the channel too, but `deleteChannelAssetsAsModerator` will only allow you to delete objects that are direct channel assets.

This means you need to filter those objects, so that only those with `'channel' in o.type` remain

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  async run(): Promise<void> {
+    const {
+      flags: { channelId, assetIds, rationale },
+    } = this.parse(DeleteChannelAssetsAsModeratorCommand)
+    // Context
+    const [actor, address] = await this.getCuratorContext()
+
+    const dataObjects = await this.getDataObjectsInfo(channelId, assetIds)
+    const stateBloatBond = dataObjects.length * (await this.getApi().dataObjectStateBloatBond())
```
Lezek123: This is not a valid way to compute bloat bond. Each data object may have a different bloat bond, because the value may change during the lifetime of the channel. You should instead sum `o.stateBloatBond` of all the objects

cli/src/commands/content/deleteVideoAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+import BN from 'bn.js'
+import chalk from 'chalk'
+import { formatBalance } from '@polkadot/util'
+import { createType } from '@joystream/types'
+import ExitCodes from '../../ExitCodes'
+
+export default class DeleteVideoAsModeratorCommand extends ContentDirectoryCommandBase {
+  static description = 'Delete the video and optionally all associated data objects.'
+
+  protected requiresQueryNode = true
```
Lezek123, line 12: Technically you could remove this requirement in this and other similar commands, because now we have `video.data_objects` and `channel.date_objects` stored in the chain state, but that's not a high priority

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+    }),
+    ...ContentDirectoryCommandBase.flags,
+  }
+
+  async getDataObjectsInfo(videoId: number, assetIds: string[]): Promise<number[]> {
+    const dataObjects = (await this.getQNApi().dataObjectsByVideoId(videoId.toString())).map((o) => o.id)
+
+    return assetIds.map((id) => {
+      if (!dataObjects.includes(id)) {
+        this.error(`Data object ${id} is not associated eith video ${videoId}`, {
```
Lezek123:
```suggestion
        this.error(`Data object ${id} is not associated with video ${videoId}`, {
```

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  async run(): Promise<void> {
+    const {
+      flags: { videoId, assetIds, rationale },
+    } = this.parse(DeleteVideoAssetsAsModeratorCommand)
+    // Context
+    const [actor, address] = await this.getCuratorContext()
+
+    const dataObjects = await this.getDataObjectsInfo(videoId, assetIds)
+    const stateBloatBond = dataObjects.length * (await this.getApi().dataObjectStateBloatBond())
```
Lezek123: Incorrect way of calculating bloat bond as mentioned in one of the previous comments

cli/src/commands/content/updateCuratorGroupPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+      required: false,
+      description: 'ID of the Curator Group',
+    },
+  ]
+
+  static flags = {
+    permissions: flags.string({
+      char: 'p',
+      required: true,
+      description: `Path to JSON file containing moderation permissions by channel previledge to use as input`,
```
Lezek123:
```suggestion
      description: `Path to JSON file containing moderation permissions by channel privilege level to use as input`,
```

cli/src/commands/content/updateCuratorGroupPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+        ])
+      )
+    )
+
+    await this.sendAndFollowNamedTx(keypair, 'content', 'updateCuratorGroupPermissions', [
+      id,
+      moderationPermissionsByLevel,
+    ])
+
+    console.log(chalk.green(`New group successfully created!`))
```
Lezek123: This should say something along the lines of _permissions successfully updated_

cli/src/helpers/InputOutput.ts
```diff
@@ -33,7 +33,7 @@ export async function getInputJson<T>(inputPath: string, schema?: unknown): Prom
     throw new CLIError(`JSON parsing failed for file: ${inputPath}`, { exit: ExitCodes.InvalidInput })
   }
   if (schema) {
-    await validateInput(jsonObj, schema)
+    await validateInput(jsonObj.length ? jsonObj[0] : jsonObj, schema)
```
Lezek123: You can define the schema root as an array instead.
The problem with this approach is that it will only validate the first value in the array, while all the other values may still be invalid:
```
[
  {
    "channelPreviledgeLevel": 0,
    "permissions": [
      { "HideVideo": null },
      { "ChangeChannelFeatureStatus": "CreatorTokenIssuance" },
      { "DeleteVideoAssets": true }
    ]
  },
  "I can put whatever here and it will still pass..."
]
```

cli/src/schemas/ContentDirectory.ts
```diff
@@ -41,7 +42,6 @@ export const ChannelCreationInputSchema: JsonSchema<ChannelCreationInputParamete
             items: {
               type: 'string',
               enum: [
-                'UpdateChannelMetadata',
```
Lezek123, line 44: Why was this removed?
I think it's best to just use:
```typescript
              enum: keysOf<PalletContentChannelActionPermission, 'PalletContentChannelActionPermission'>(
                'PalletContentChannelActionPermission'
              ),
```

cli/src/schemas/ContentDirectory.ts
```diff
@@ -72,6 +72,51 @@ export const ChannelCreationInputSchema: JsonSchema<ChannelCreationInputParamete
   },
 }
 
+export const ModerationPermissionsByLevelInputSchema: JsonSchema<ModerationPermissionsByLevelInputParameters> = {
+  type: 'object',
+  additionalProperties: false,
+  properties: {
+    channelPreviledgeLevel: { type: 'integer' },
```
Lezek123:
```suggestion
    channelPrivilegeLevel: { type: 'integer' },
```

cli/src/schemas/ContentDirectory.ts
```diff
@@ -72,6 +72,51 @@ export const ChannelCreationInputSchema: JsonSchema<ChannelCreationInputParamete
...
+    permissions: {
+      type: 'array',
+      items: {
+        type: 'object',
+        enum: [
+          { DeleteVideo: null },
```
Lezek123: Probably worth iterating over `keysOf('PalletContentPermissionsCuratorGroupContentModerationAction')` and `keysOf('PalletContentPausableChannelFeature')` (in case of `ChangeChannelFeatureStatus`) here also

cli/src/commands/content/createCuratorGroup.ts
```diff
@@ -1,18 +1,52 @@
...
-  static aliases = ['createCuratorGroup']
   static flags = {
+    status: flags.enum({
+      required: true,
+      description: `Status of newly created Curator Group: (${CURATOR_GROUP_CREATION_CONTEXTS.join('/')})`,
+      options: [...CURATOR_GROUP_CREATION_CONTEXTS],
+    }),
+    permissions: flags.string({
+      char: 'p',
+      required: true,
```
Lezek123: It should be possible to create curator group with empty moderation permissions, this is currently not possible because this flag is required and the empty array does not pass the json schema validation

cli/src/commands/content/updateVideoStateBloatBond.ts
```diff
@@ -0,0 +1,33 @@
...
+  ]
+
+  async run(): Promise<void> {
+    const { value } = this.parse(UpdateVideoStateBloatBondCommand).args
+
+    // Context
+    const lead = await this.getRequiredLeadContext()
+    const keypair = await this.getDecodedPair(lead.roleAccount)
+
+    this.jsonPrettyPrint(JSON.stringify({ newChannelStateBloatBond: value }))
```
Lezek123:
```suggestion
    this.jsonPrettyPrint(JSON.stringify({ newVideoStateBloatBond: value }))
```

cli/src/commands/content/addCuratorToGroup.ts
```diff
@@ -17,13 +20,32 @@ export default class AddCuratorToGroupCommand extends ContentDirectoryCommandBas
...
+          }
+          return v as ChannelActionPermission['type']
+        })
+        return arr
+      },
+    })({
+      char: 'p',
+      description: `Comma separated list of permissions to associate with the curator`,
+      required: true,
+    }),
```
Lezek123:
```suggestion
    permissions: flags.enum({
      options: keysOf<ChannelActionPermission, 'PalletContentChannelActionPermission'>(
        'PalletContentChannelActionPermission'
      ),
      multiple: true,
      char: 'p',
      description: `Comma separated list of permissions to associate with the curator`,
      required: false,
    }),
```

And later on you can use:
```typescript
      createType(
        'BTreeSet<PalletContentChannelActionPermission>',
        typeof permissions === 'string' ? [permissions] : permissions
      ),
```

Lezek123 (changes requested):
1. In `updateChannel` there is still an old condition:
```typescript
    if (collaborators !== undefined && !this.isChannelOwner(channel, actor)) {
      this.error("Only channel owner is allowed to update channel's collaborators!", { exit: ExitCodes.AccessDenied })
    }
```
This is no longer the case, because the collaborator may have `ManageChannelCollaborators` permissions, in which case he will be blocked by the CLI from performing an action which he is allowed to perform. You could also look for other checks like that to make sure curators/collaborators are not blocked from actions which they were previously not allowed to perform.

2. I think I already pointed it out in one of the previous PRs, but there are a few issues with the current implementation of `selectStorageBucketsForNewChannel` and `selectDistributionBucketsForNewChannel` in `Api.ts`:
    1. Buckets are picked deterministically using `.slice` which results in some buckets being picked basically all the time, while others are never being picked.
    2. You should probably favor storage buckets with most available space, otherwise you may pick buckets which are already full or close to being full.
3. You may want to update display commands like `content:curatorGroups` / `content:curatorGroup` etc. to include useful new information, like group/curator permissions for example. It's not a high priority, but it should be quite straightforward to do.
4. Channel/video state bloat bond should be displayed when creating/deleting a channel/video (probably worth adding a summary of all costs associated with channel/video creation too)

cli/src/commands/content/addCuratorToGroup.ts
zeeshanakram3: addressed in [6adbb26](https://github.com/Joystream/joystream/pull/4084/commits/6adbb26c9931578c08b9e5edb8d163f55175ada0)

Though `multiple` option does not seem to work inside `flags.enum`, as the flag generated is always of type `T` instead of `T[]`, but `flags.string` or `flags.integer` with `multiple` works as intended

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
zeeshanakram3: Addressed in [7b05821](https://github.com/Joystream/joystream/pull/4084/commits/7b05821d3ac84d116436d12a7c17b4e93a17db0b)

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
zeeshanakram3: Addressed in [7b05821](https://github.com/Joystream/joystream/pull/4084/commits/7b05821d3ac84d116436d12a7c17b4e93a17db0b)

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
zeeshanakram3: Seems `this.getQNApi().dataObjectsByChannelId(videoId.toString()))` only includes data objects that are direct channel assets. The underlying [query](https://github.com/zeeshanakram3/joystream/blob/6adbb26c9931578c08b9e5edb8d163f55175ada0/cli/src/graphql/queries/storage.graphql#L75) filters & returns only direct channel assets.

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
zeeshanakram3: Addressed in [7b05821](https://github.com/Joystream/joystream/pull/4084/commits/7b05821d3ac84d116436d12a7c17b4e93a17db0b)

cli/src/commands/content/deleteChannelAsModerator.ts
zeeshanakram3: Addressed in [882efe9](https://github.com/Joystream/joystream/pull/4084/commits/882efe90a343439a11ef78973e6b31fa855f44ff)

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
zeeshanakram3: Addressed in [7b05821](https://github.com/Joystream/joystream/pull/4084/commits/7b05821d3ac84d116436d12a7c17b4e93a17db0b)

cli/README.md
zeeshanakram3: Addressed in [f0446f4](https://github.com/Joystream/joystream/pull/4084/commits/f0446f469ba25d548142dfd43f042e96e4d55792)

cli/src/Types.ts
zeeshanakram3: Addressed in [968db20](https://github.com/Joystream/joystream/pull/4084/commits/968db206d806dd26aa424635446ccac84ca03157)

cli/src/commands/content/deleteChannel.ts
```diff
@@ -78,11 +60,15 @@ export default class DeleteChannelCommand extends ContentDirectoryCommandBase {
           exit: ExitCodes.InvalidInput,
         })
       }
-      const stateBloatBond = dataObjectsInfo.reduce((sum, [, bloatBond]) => sum.add(bloatBond), new BN(0))
+      const videosStateBloatBond = videosInfo.reduce((sum, [, bloatBond]) => sum.add(bloatBond), new BN(0))
```
Lezek123: Only channel state bloat bond is recovered when deleting a channel. Channel cannot be deleted if it has any outstanding videos.

cli/src/commands/content/deleteChannelAsModerator.ts
```diff
@@ -0,0 +1,81 @@
...
+      ? await this.getDataObjectsInfoFromQueryNode(channelId)
+      : await this.getDataObjectsInfoFromChain(channelId)
+
+    if (dataObjectsInfo.length) {
+      if (!force) {
+        this.error(`Cannot remove associated data objects unless ${chalk.magentaBright('--force')} flag is used`, {
+          exit: ExitCodes.InvalidInput,
+        })
+      }
+      const videosStateBloatBond = videosInfo.reduce((sum, [, bloatBond]) => sum.add(bloatBond), new BN(0))
```
Lezek123: Only channel state bloat bond is recovered when deleting a channel. Channel cannot be deleted if it has any outstanding videos.

query-node/schemas/content.graphql
```diff
@@ -149,6 +149,9 @@ type Video @entity {
   "Video media asset"
   media: StorageDataObject
 
+  "Video media asset"
+  videoStateBloatBond: BigInt!
```
Lezek123: Invalid property description

cli/src/Api.ts
```diff
@@ -509,20 +509,22 @@ export default class Api {
     const distributionBucketIds = []
 
     for (const { id, buckets } of families || []) {
-      const bucketsCountPolicy = distributionBucketFamiliesPolicy.get(id as unknown as DistributionBucketFamilyId)
-      if (bucketsCountPolicy && bucketsCountPolicy.toNumber() < buckets.length) {
-        throw new CLIError(`Distribution buckets policy constraint unsatifified. Not enough distribution buckets exist`)
+      const bucketsCountPolicy = distributionBucketFamiliesPolicy
+        .get(id as unknown as DistributionBucketFamilyId)
+        ?.toNumber()
```
Lezek123: Currently it always returns `undefined`
```suggestion
      const bucketsCountPolicy = [...distributionBucketFamiliesPolicy]
        .find(([familyId]) => familyId.toString() === id)?.[1]
        .toNumber()
```

cli/src/graphql/queries/storage.graphql
```diff
@@ -18,7 +18,7 @@ query getStorageNodesInfoByBagId($bagId: ID) {
 }
 
 query getStorageBuckets {
-  storageBuckets(where: { acceptingNewBags_eq: true }) {
+  storageBuckets(where: { acceptingNewBags_eq: true }, orderBy: [dataObjectsSize_ASC]) {
```
Lezek123: Storage buckets may have different limits, so you need calculate `dataObjectsSizeLimit - dataObjectsSize` to get the available size and `dataObjectCountLimit - dataObjectsCount` to get the available number of objects. This cannot be done with a single query currently, so you need to do some post-processing on the results (either filtering out the buckets that are already unable to store the assets to be uploaded or assigning some kind of priority for buckets based on available size and available number of objects)

cli/src/commands/content/channels.ts
```diff
@@ -18,6 +18,10 @@ export default class ChannelsCommand extends ContentDirectoryCommandBase {
           'ID': id.toString(),
           'Owner': JSON.stringify(c.owner.toJSON()),
           'Collaborators': c.collaborators.size,
+          'ChannelStateBloatBond': c.channelStateBloatBond.toString(),
```
Lezek123:
```suggestion
          'ChannelStateBloatBond': formatBalance(channelStateBloatBond),
```
It's more readable this way

cli/src/commands/content/channel.ts
```diff
@@ -28,6 +39,9 @@ export default class ChannelCommand extends ContentDirectoryCommandBase {
       displayCollapsedRow({
         'ID': channelId.toString(),
         'Owner': JSON.stringify(channel.owner.toJSON()),
+        'ChannelStateBloatBond': channel.channelStateBloatBond.toString(),
```
Lezek123:
```suggestion
        'ChannelStateBloatBond': formatBalance(channel.channelStateBloatBond),
```

cli/src/commands/content/curatorGroup.ts
```diff
@@ -36,5 +36,11 @@ export default class CuratorGroupCommand extends ContentDirectoryCommandBase {
         )
         .join(', ')
     )
+    displayHeader(`Group Permissions (${[...permissionsByLevel].length})`)
+    this.log(
+      [...permissionsByLevel]
+        .map(([level, permissions]) => chalk.magentaBright(`Privilege Level: ${level}; (Permissions: ${permissions})`))
+        .join(', ')
```
Lezek123: This needs some better visual distinction, for example:
```suggestion
        .join('\n\n')
```

Perhaps you could also display `ChannelAgentPermissions` of each group member (curator)?

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  protected requiresQueryNode = true
+
+  static flags = {
+    channelId: flags.integer({
+      char: 'c',
+      required: true,
+      description: 'ID of the Channel',
+    }),
+    assetIds: flags.string({
```
Lezek123:
```suggestion
    assetIds: flags.integer({
```

cli/src/commands/working-groups/apply.ts
```diff
@@ -112,18 +107,18 @@ export default class WorkingGroupsApply extends WorkingGroupsCommandBase {
...
-          opening_id: openingId,
-          role_account_id: roleAccount,
-          reward_account_id: rewardAccount,
-          stake_parameters: stakeParams,
+        createType('PalletWorkingGroupApplyOnOpeningParams', {
+          memberId: memberContext.id,
+          openingId: openingId,
+          roleAccountId: roleAccount,
+          rewardAccountId: rewardAccount,
+          stakeParameters: stakeParams!,
```
Lezek123:
```suggestion
          stakeParameters: stakeParams,
```

cli/src/commands/content/deleteVideoAsModerator.ts
```diff
@@ -0,0 +1,85 @@
...
+    const [actor, address] = await this.getCuratorContext()
+
+    const dataObjectsInfo = await this.getDataObjectsInfo(videoId)
+    if (dataObjectsInfo.length) {
+      if (!force) {
+        this.error(`Cannot remove associated data objects unless ${chalk.magentaBright('--force')} flag is used`, {
+          exit: ExitCodes.InvalidInput,
+        })
+      }
+      const { videoStateBloatBond } = await this.getApi().videoById(videoId)
```
Lezek123: It's not a huge issue, but moving `await this.getApi().videoById(videoId)` more to the top would be better, because it would make sense to make sure the video exists before performing any other checks

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  async run(): Promise<void> {
+    const {
+      flags: { videoId, assetIds, rationale },
+    } = this.parse(DeleteVideoAssetsAsModeratorCommand)
+    // Context
+    const [actor, address] = await this.getCuratorContext()
+
+    const dataObjectsInfo = await this.getDataObjectsInfo(videoId, assetIds)
```
Lezek123, line 70: Here it probably also makes sense to check if the video even exists before executing any other logic

cli/src/commands/content/deleteChannelAssetsAsModerator.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  async run(): Promise<void> {
+    const {
+      flags: { channelId, assetIds, rationale },
+    } = this.parse(DeleteChannelAssetsAsModeratorCommand)
+    // Context
+    const [actor, address] = await this.getCuratorContext()
+
+    const dataObjectsInfo = await this.getDataObjectsInfo(channelId, assetIds)
```
Lezek123, line 67: Make sure the channel exists first

cli/src/commands/content/videos.ts
```diff
@@ -31,6 +31,8 @@ export default class VideosCommand extends ContentDirectoryCommandBase {
         videos.map(([id, v]) => ({
           'ID': id.toString(),
           'InChannel': v.inChannel.toString(),
+          'VideoStateBloatBond': v.videoStateBloatBond.toString(),
```
Lezek123:
```suggestion
          'VideoStateBloatBond': formatBalance(v.videoStateBloatBond),
```

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
```diff
@@ -0,0 +1,78 @@
...
+      })
+    })
+  }
+
+  async run(): Promise<void> {
+    const {
+      flags: { videoId, assetIds, rationale },
+    } = this.parse(DeleteVideoAssetsAsModeratorCommand)
+    // Context
+    const [actor, address] = await this.getCuratorContext()
```
Lezek123: It would be a nice feature to allow `Lead` context for any `AsModerator` command, so that the lead doesn't need to create a group and add himself to it in order to perform those actions.

cli/src/commands/content/updateChannel.ts
```diff
@@ -92,8 +79,18 @@ export default class UpdateChannelCommand extends UploadCommandBase {
     const meta = asValidatedMetadata(ChannelMetadata, channelInput)
     const { collaborators, coverPhotoPath, avatarPhotoPath } = channelInput
 
-    if (collaborators !== undefined && !this.isChannelOwner(channel, actor)) {
-      this.error("Only channel owner is allowed to update channel's collaborators!", { exit: ExitCodes.AccessDenied })
+    if (
+      collaborators !== undefined &&
+      !this.isChannelOwner(channel, actor) &&
+      !this.isCollaboratorWithRequiredPermission(channel, actor, 'ManageChannelCollaborators')
```
Lezek123: I think it makes sense to just generally rework the context & permissions in the CLI, because:
- For curator channels we now need to introduce `Curator` context (while only `Owner` / `Collaborator` is currently available)
- For both `Curator` and `Collaborator` context we need to verify `ChannelAgentPermissions`. In case of `updateChannel` there are 3 different permissions that need to be taken into account: `UpdateChannelMetadata` (in case metadata is changed), `ManageNonVideoChannelAssets` (in case assets are updated), `ManageChannelCollaborators` (in case collaborators are updated). For `ManageChannelCollaborators` ideally you'd also check whether the affected collaborators have lower permissions than the Curator/Collaborator performing the update (see the comments in https://github.com/Joystream/joystream/blob/carthage/runtime-modules/content/src/types.rs#L134). For other actions like `createVideo` / `updateVideo` / `deleteVideo` etc. you also need to check different permissions depending on the parameters.

Lezek123 (changes requested):
1. Could you also fix https://github.com/Joystream/joystream/issues/4118 as part of this PR? It seems like the issue is caused by invalid typings for `entiresByIds` in `Api.ts`:
  ```typescript
      const applicationEntries = await this.entriesByIds<ApplicationId, Application>(
        this.workingGroupApiQuery(group).applicationById
      )
  ```
You can remove `<ApplicationId, Application>` part as it's no longer needed and it seems to override the types that come from api augmentations.
There are a few other places where similar problem occurs, for example:
```typescript
  async allMembers(): Promise<[MemberId, Membership][]> {
    return this.entriesByIds<MemberId, Membership>(this._api.query.members.membershipById)
  }
```
(the runtime api was changed, it now returns `Option<Membership>`, but the types here are also incorrectly overriden).
Basically please remove the generic types from any call to `entriesByIds` in `Api.ts`

2. Was it intentional to skip commands for actions like hiding video/channel, updating channel paused features etc.?

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
Lezek123: Also perhaps in case of curator you could add a check whether curator has the right moderation permissions? (based on group's `ModerationPermissionsByLevel` and channel's privilege level)

zeeshanakram3:
> 1. Could you also fix #4118 as part of this PR? 

Sure

> 2. Was it intentional to skip commands for actions like hiding video/channel, updating channel paused features, etc.?

I overlooked hiding video/channel extrinsics, will add them now. Since updating channel paused features, updating curator group permissions, and updating channel privilege level actions were disabled in #3914, I intentionally skipped their CLI support
@Lezek123

cli/src/commands/content/updateChannel.ts
zeeshanakram3: Addressed in [965e1d1](https://github.com/Joystream/joystream/pull/4084/commits/965e1d14433b2f9d5ab9cea37292026a3a2a8304)

1. Isn't for [CuratorGroup](https://github.com/zeeshanakram3/joystream/blob/965e1d14433b2f9d5ab9cea37292026a3a2a8304/cli/src/base/ContentDirectoryCommandBase.ts#L97) owned channels both `Lead` & `Curator` contexts are already captured by `Owner` context? So not sure if introducing `Curator` context is necessary.
2. Since support for NFT actions doesn't exist in CLI, including in `createVideo`/`updateVideo` as these commands just set `autoIssueNft`  to _null_. Hence, `ManageVideoNfts` permission check is missing from both these commands as of now

cli/src/graphql/queries/storage.graphql
zeeshanakram3: Addressed in [36c5bab](https://github.com/Joystream/joystream/pull/4084/commits/36c5babdffde9824ed39d4b13ad2e5f30afdfa2f)

Now we fetch all buckets that are accepting the bags, sorting those based on most available size & most available count, and then pick first `n` buckets based on policy. Does this approach works?

cli/src/commands/content/setChannelVisibilityAsModerator.ts
```diff
@@ -0,0 +1,61 @@
...
+    channelId: flags.integer({
+      char: 'c',
+      required: true,
+      description: 'ID of the channel',
+    }),
+    status: flags.enum({
+      char: 's',
+      options: [...CHANNEL_VISIBILITY_CONTEXTS],
+      description: 'The visibility status of the channel',
+      required: false,
```
Lezek123:
```suggestion
      required: true,
```

cli/src/commands/content/updateChannel.ts
```diff
@@ -118,11 +106,31 @@ export default class UpdateChannelCommand extends UploadCommandBase {
       assetIndices.avatarPhotoPath
     )
 
+    // Ensure actor is authorized to perform channel update
+    const requiredPermissions: ChannelActionPermission['type'][] = []
+    if (collaborators) {
+      requiredPermissions.push('ManageChannelCollaborators')
+    }
+    if (assetsToUpload) {
```
Lezek123:
```suggestion
    if (assetsToUpload || assetsToRemove.length) {
```

cli/src/commands/content/updateChannel.ts
```diff
@@ -118,11 +106,31 @@ export default class UpdateChannelCommand extends UploadCommandBase {
...
+        }
+      )
+    }
+
     const channelUpdateParameters = createType('PalletContentChannelUpdateParametersRecord', {
       expectedDataObjectStateBloatBond,
       assetsToUpload,
       assetsToRemove,
-      newMeta: metadataToBytes(ChannelMetadata, meta),
+      newMeta: serializedMeta,
```
Lezek123:
```suggestion
      newMeta: serializedMeta.length ? serializedMeta : null,
```
Runtime won't check `UpdateChannelMetadata` permissions only if provided metadata `isNone`
(the same fix is needed in `updateVideo.ts`)

cli/src/commands/content/updateVideo.ts
```diff
@@ -87,14 +91,33 @@ export default class UpdateVideoCommand extends UploadCommandBase {
...
     const assetsToUpload = await this.prepareAssetsForExtrinsic(resolvedAssets)
     const assetsToRemove = await this.getAssetsToRemove(
       videoId,
       assetIndices.videoPath,
       assetIndices.thumbnailPhotoPath
     )
+
+    // Ensure actor is authorized to perform video update
+    const requiredPermissions: ChannelActionPermission['type'][] = []
+    if (assetsToUpload) {
```
Lezek123:
```suggestion
    if (assetsToUpload || assetsToRemove.length) {
```

cli/src/commands/content/curatorGroup.ts
```diff
@@ -40,7 +40,15 @@ export default class CuratorGroupCommand extends ContentDirectoryCommandBase {
     this.log(
       [...permissionsByLevel]
         .map(([level, permissions]) => chalk.magentaBright(`Privilege Level: ${level}; (Permissions: ${permissions})`))
-        .join(', ')
+        .join('\n\n')
+    )
+    displayHeader(`Curator by Permissions (${[...permissionsByLevel].length})`)
```
Lezek123:
```suggestion
    displayHeader(`Permissions by Curator`)
```

cli/src/graphql/queries/storage.graphql
Lezek123: It works except for the edge case when an operator has a lot of size available, but the `dataObjectCountLimit` is already reached. But it's not a big deal, I guess in this case the lead should either set `acceptingNewBags` to `false` for this bucket or increase the objects limit.

cli/src/commands/content/updateChannel.ts
Lezek123: 1. Oh, I see. The naming is a little confusing now, since `Curator` cannot act as an owner anymore, only `Lead` can. So my suggestion would be to have 3 possible contexts: `Owner` (member or lead depending on channel type), `Curator` (only for curator channels), `Collaborator` (for all channels).
2. Makes sense

cli/src/commands/content/updateChannel.ts
```diff
@@ -112,21 +96,44 @@ export default class UpdateChannelCommand extends UploadCommandBase {
...
+    }
+    if (assetsToUpload) {
+      requiredPermissions.push('ManageNonVideoChannelAssets')
+    }
+    if (serializedMeta.length) {
+      requiredPermissions.push('UpdateChannelMetadata')
+    }
+    if (!(await this.hasRequiredChannelAgentPermissions(actor, channel, requiredPermissions))) {
+      this.error(
+        `Only channelOwner or collaborator with ${requiredPermissions} permission can update channel ${channelId}!`,
```
Lezek123:
```suggestion
        `Only channelOwner or collaborator with ${requiredPermissions} permission can perform this update!`,
```
Since permissions are specific to update params

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -109,6 +121,75 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
...
+        const curatorChannelAgentPermissions = curators.get(actor.asCurator[1])
+        return !!(
+          curatorChannelAgentPermissions &&
+          requiredPermissions.every((requiredPermission) =>
+            [...curatorChannelAgentPermissions].find((p) => p[`is${requiredPermission}`])
+          )
+        )
+      }
+      // Collaborator context
+      const collaboratorChannelAgentPermissions = channel.collaborators.get(actor.asMember)
```
Lezek123: The `.get` method of `BTreeMap`s is not really useful, because it compares instances (like `u64`) by reference, not by the actual value. I usually workaround this with something like:
```suggestion
      const collaboratorChannelAgentPermissions = [...channel.collaborators].find(([k]) => k.eq(actor.asMember))?.[1]
```
But you can also choose to override `BTreeMap.prototype.get` to be safer.
I may create a PR to `@polkadot/api` to fix this when I have some spare time.

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -109,6 +121,75 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
...
+    // CASE: CuratorGroup owned channel
+    if (channel.owner.isCuratorGroup) {
+      // Lead context
+      if (actor.isLead) {
+        return true
+      }
+      // Curator context
+      if (actor.isCurator && actor.asCurator[0].eq(channel.owner.asCuratorGroup)) {
+        const { curators } = await this.getCuratorGroup(actor.asCurator[0].toNumber())
+        const curatorChannelAgentPermissions = curators.get(actor.asCurator[1])
```
Lezek123: Same issue as with the collaborators map:
```suggestion
        const curatorChannelAgentPermissions = [...curators].find(([k]) => k.eq(actor.asCurator[1]))?.[1]
```

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -109,6 +121,75 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
...
+    }
+
+    // CASE: Member owned channel
+    if (channel.owner.isMember) {
+      // Owner context
+      if (actor.isMember) {
+        return actor.asMember.eq(channel.owner.asMember)
+      }
+      // Collaborator context
+      const collaboratorChannelAgentPermissions = channel.collaborators.get(actor.asMember)
```
Lezek123: Same issue with the `BTreeMap.get`:
```suggestion
      const collaboratorChannelAgentPermissions = [...channel.collaborators].find(([k]) => k.eq(actor.asMember))?.[1]
```

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -109,6 +121,75 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
...
+    channelPrivilegeLevel: ChannelPrivilegeLevel,
+    permission: CreateInterface<ContentModerationAction>
+  ): Promise<boolean> {
+    if (actor.isLead) {
+      return true
+    }
+
+    const permissionRequired = createType('PalletContentPermissionsCuratorGroupContentModerationAction', permission)
+    const { permissionsByLevel } = await this.getCuratorGroup(actor.asCurator[0].toNumber())
+    const permissionsForLevel = permissionsByLevel.get(channelPrivilegeLevel)
```
Lezek123: Same issue with `BTreeMap.get`:
```suggestion
    const permissionsForLevel = [...permissionsByLevel].find(([k]) => k.eq(channelPrivilegeLevel))?.[1]
```

cli/src/commands/content/updateChannel.ts
```diff
@@ -112,21 +96,44 @@ export default class UpdateChannelCommand extends UploadCommandBase {
...
       expectedDataObjectStateBloatBond,
-      collaborators: [],
-      assetsToRemove: createType('BTreeSet<u64>', assetsToRemove),
-      newMeta: metadataToBytes(ChannelMetadata, meta),
+      assetsToUpload,
+      assetsToRemove,
+      newMeta: serializedMeta,
+      collaborators: new Map(
+        collaborators?.map(({ memberId, channelAgentPermissions }) => [memberId, channelAgentPermissions])
+      ),
```
Lezek123: Same issue as with `newMeta`:
```suggestion
      collaborators: collaborators?.length
        ? new Map(collaborators?.map(({ memberId, channelAgentPermissions }) => [memberId, channelAgentPermissions]))
        : null,
```

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -109,6 +121,75 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
...
+        )
+      )
+    }
+
+    // CASE: Member owned channel
+    if (channel.owner.isMember) {
+      // Owner context
+      if (actor.isMember) {
+        return actor.asMember.eq(channel.owner.asMember)
+      }
```
Lezek123: Collaborator is also represented as `ContentActor::Member`, so for member channels those conditions essentially disallow any action by collaborator

cli/src/commands/content/deleteChannel.ts
```diff
@@ -23,66 +24,51 @@ export default class DeleteChannelCommand extends ContentDirectoryCommandBase {
...
+      ? await this.getDataObjectsInfoFromQueryNode(channelId)
+      : await this.getDataObjectsInfoFromChain(channelId)
+
+    // Ensure actor is authorized to perform channel deletion
+    const requiredPermissions: ChannelActionPermission['type'][] = dataObjectsInfo.length
+      ? ['DeleteChannel', 'ManageNonVideoChannelAssets']
+      : ['DeleteChannel']
+    if (!(await this.hasRequiredChannelAgentPermissions(actor, channel, requiredPermissions))) {
+      this.error(
+        `Only channelOwner or collaborator with ${requiredPermissions} permission can delete channel ${channelId}!`,
```
Lezek123:
```suggestion
        `Only channelOwner or collaborator with ${requiredPermissions} permissions can perform this deletion!`,
```

cli/src/commands/content/deleteVideo.ts
```diff
@@ -40,26 +41,42 @@ export default class DeleteVideoCommand extends ContentDirectoryCommandBase {
...
 
     const dataObjectsInfo = await this.getDataObjectsInfo(videoId)
+
+    // Ensure actor is authorized to perform video deletion
+    const requiredPermissions: ChannelActionPermission['type'][] = dataObjectsInfo.length
+      ? ['DeleteVideo', 'ManageVideoAssets']
+      : ['DeleteVideo']
+    if (!(await this.hasRequiredChannelAgentPermissions(actor, channel, requiredPermissions))) {
+      this.error(
+        `Only channel owner or collaborator with ${requiredPermissions} permissions can delete video ${videoId}!`,
```
Lezek123:
```suggestion
        `Only channel owner or collaborator with ${requiredPermissions} permissions can perform this deletion!`,
```

cli/src/commands/content/deleteVideoAssetsAsModerator.ts
```diff
@@ -0,0 +1,92 @@
...
+    // ensure video exists
+    const { inChannel, nftStatus } = await this.getApi().videoById(videoId)
+    const { privilegeLevel } = await this.getApi().channelById(inChannel)
+
+    // Ensure moderator has required permission
+    if (
+      !(await this.isModeratorWithRequiredPermission(actor, privilegeLevel, { DeleteVideoAssets: nftStatus.isSome }))
+    ) {
+      this.error(
+        `Only content lead or curator with "DeleteVideoAssets" permission can delete video ${videoId} assets!`,
```
Lezek123: Please add a differentiation between `DeleteVideoAssets(true)` and `DeleteVideoAssets(false)`

cli/src/QueryNodeApi.ts
```diff
@@ -156,18 +171,35 @@ export default class QueryNodeApi {
...
+      { count: countQueryResult.totalCount },
       'storageBuckets'
     )
+
+    // sorting buckets based on available size, if two buckets have same
+    // available size then sort the two based on available dataObjects count
+    return buckets.sort(
+      (x, y) =>
+        y.dataObjectsSizeLimit - y.dataObjectsSize - (x.dataObjectsSizeLimit - x.dataObjectsSize) ||
+        y.dataObjectCountLimit - y.dataObjectsCount - (x.dataObjectCountLimit - x.dataObjectsCount)
```
Lezek123: `dataObjectsSize`, `dataObjectsSizeLimit`, `dataObjectsCount` and  `dataObjectCountLimit` have `BigNumber` type in the QN, so the api returns them as `string`. Using arithmetic operators on strings may give some unexpected results, you should convert them to `BN`s first, ie:
```suggestion
        new BN(y.dataObjectsSizeLimit)
          .sub(new BN(y.dataObjectsSize))
          .cmp(new BN(x.dataObjectsSizeLimit).sub(new BN(x.dataObjectsSize))) ||
        new BN(y.dataObjectCountLimit)
          .sub(y.dataObjectsCount)
          .cmp(new BN(x.dataObjectCountLimit).sub(new BN(x.dataObjectsCount)))
```

Lezek123 (changes requested):
Besides the comments, I also noticed we're still using `memberId` for uploads, which is no longer required, created a small PR into your branch to remove those: https://github.com/zeeshanakram3/joystream/pull/2

cli/src/base/ContentDirectoryCommandBase.ts
zeeshanakram3: Addressed in [ed79fd3](https://github.com/Joystream/joystream/pull/4084/commits/ed79fd35ad0f0f9a5e3e34c174974a8c167925aa)

cli/src/base/ContentDirectoryCommandBase.ts
```diff
@@ -211,6 +206,16 @@ export default abstract class ContentDirectoryCommandBase extends WorkingGroupCo
       // continue
     }
 
+    try {
+      if (channel.owner.isCuratorGroup) {
+        const owner = await this.getCuratorContext(channel.owner.asCuratorGroup)
+        this.log('Derived context: Channel curator')
+        return owner
```
Lezek123:
```suggestion
        const curator = await this.getCuratorContext(channel.owner.asCuratorGroup)
        this.log('Derived context: Curator')
        return curator
```

Lezek123 (approved):
LGTM, just a small naming issue

Issue #4087, opened 2022-07-28
Extract Subtitles from Playlists PR


dmtrjsg:
Title says it all

PR #4092, opened 2022-07-29
Qn carthage video subtitles

zeeshanakram3:
implements #4087

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202703110812154) by [Unito](https://www.unito.io)


query-node/manifest.yml
```diff
@@ -722,11 +722,11 @@ mappings:
...
       handler: content_ChannelAssetsRemoved
     - event: content.VideoCreated
-      handler: content_VideoCreated
+      handler: content_ContentCreated
     - event: content.VideoUpdated
-      handler: content_VideoUpdated
+      handler: content_ContentUpdated
     - event: content.VideoDeleted
-      handler: content_VideoDeleted
+      handler: content_ContentDeleted
```
ondratra, line 730: Why is the handler renamed? It will be easier to navigate between runtime and mappings if handler names are the same as event names.

query-node/mappings/src/content/utils.ts
```diff
@@ -129,6 +138,33 @@ async function processVideoAssets(
...
+      }
+      const dataObjectType = new DataObjectTypeConstructor()
+      dataObjectType.subtitleId = subtitle.id
+      dataObjectType.videoId = subtitle.video.id
+      asset.type = dataObjectType
+      subtitle[schemaFieldName] = asset
+      await store.save<StorageDataObject>(asset)
+    }
+  }
+}
```
ondratra, line 165: Let's reduce the amount of nested `if`s by introducint helper function(s).

query-node/mappings/src/content/utils.ts
```diff
@@ -465,6 +506,45 @@ async function processLanguage(
...
+  const { store } = ctx
+
+  const subtitlesToRemove = await store.getMany(VideoSubtitle, { where: { video: { id: video.id } } })
+
+  console.log('before subtitlesToRemove: ', subtitlesToRemove)
+
+  for (const subtitleMeta of subtitlesMeta) {
+    const subtitleId = `${video.id}-${subtitleMeta.type}-${subtitleMeta.language}`
+
+    _.remove(subtitlesToRemove, (sub) => sub.id === subtitleId)
```
ondratra, line 517: Let's not intruduce `lodash` only to use `remove` function once. We can use vanilla JS `Array.filter` instead.

tests/network-tests/src/cli/joystream.ts
```diff
@@ -185,18 +186,14 @@ export class JoystreamCLI extends CLI {
...
-      jsonFile,
-      videoId.toString(),
-    ])
+    const { stderr, exitCode } = await this.run('content:updateVideo', ['--input', jsonFile, videoId.toString()])
 
     if (exitCode && !this.isErrorDueToNoStorage(exitCode)) {
-      // ignore warnings
-      throw new Error(`Unexpected CLI failure on creating video category: "${stderr}"`)
+      // ignore warningscreating
+      throw new Error(`Unexpected CLI failure on  updating video: "${stderr}"`)
```
ondratra: Typo

ondratra (changes requested):
I've suggested a couple of tweaks. Plus, some tests are failing and need to be fixed.

query-node/mappings/src/content/utils.ts
```diff
@@ -219,8 +219,8 @@ export async function processVideoMetadata(
   }
 
   // prepare subtitles if needed
-  if (isSet(meta.subtitles)) {
-    await processVideoSubtitles(ctx, video, assets, meta.subtitles)
+  if (isSet(meta.subtitles) || isSet(meta.clearSubtitles)) {
+    await processVideoSubtitles(ctx, video, assets, meta.subtitles || [], meta.clearSubtitles)
```
kdembler: I think `clearSubtitles` should take priority if defined next to regular subtitles? Maybe we do
```ts
const subtitles = meta.clearSubtitles ? [] : meta.subtitles
if (isSet(subtitles)) {
  // ...
}
```

kdembler (approved):
Tested this a bunch and everything seems to be working properly 🎉

Issue #4068, opened 2022-07-20
Content benchmark: Channel payouts & withdrawals


dobertRowneySr:
* `update_channel_payouts`
* `claim_channel_reward`
* `withdraw_from_channel_balance`
* `claim_and_withdraw_channel_reward`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202635643506464) by [Unito](https://www.unito.io)

PR #4067, opened 2022-07-20
Channel transfers benchmarking

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4038

Based on https://github.com/Joystream/joystream/pull/4036 and https://github.com/Joystream/joystream/pull/4062

Also introduces some useful utility methods:
- `setup_worst_case_scenario_channel`
- `setup_worst_case_scenario_curator_channel`
- `worst_case_scenario_collaborators`

I suggest always using `setup_worst_case_scenario_curator_channel` when possible, because content extrinsics executed with `ContentActor::Curator` context are most expensive.

Here's a comparison of the cost of a single extrinsic when executed with different contexts:

```rust
	fn initialize_channel_transfer_as_owner(a: u32, ) -> Weight {
		(77_277_000 as Weight)
			// Standard Error: 54_000
			.saturating_add((5_492_000 as Weight).saturating_mul(a as Weight))
			.saturating_add(T::DbWeight::get().reads(3 as Weight))
			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
	}

	fn initialize_channel_transfer_as_collaborator(a: u32, ) -> Weight {
		(75_567_000 as Weight)
			// Standard Error: 105_000
			.saturating_add((6_314_000 as Weight).saturating_mul(a as Weight))
			.saturating_add(T::DbWeight::get().reads(3 as Weight))
			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
	}

	fn initialize_channel_transfer_as_curator(a: u32, ) -> Weight {
		(236_663_000 as Weight)
			// Standard Error: 180_000
			.saturating_add((4_784_000 as Weight).saturating_mul(a as Weight))
			.saturating_add(T::DbWeight::get().reads(4 as Weight))
			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
        }
```

@L3pereira @ignazio-bovo

PR #4096, opened 2022-07-29
Carthage content benchmark  channel payouts and withdrawals

dobertRowneySr:
Reference: https://github.com/Joystream/joystream/issues/4068

uses code from #4067 so it's better to merge after that PR.

- added benchmarks
- updated weight info traits
- generated weights
- added short rationale for worst case estimation

I am aware of the conflicts, which I will resolve just before merging


runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -937,6 +939,134 @@ where
...
+    .map(|&x| x)
+    .collect::<BTreeSet<_>>()
+}
+
+pub fn create_pull_payments_with_reward<T: Config>(
+    payments_number: u32,
+    cumulative_reward_earned: BalanceOf<T>,
+) -> Vec<PullPayment<T>> {
+    let mut payments = Vec::new();
+    for i in 1..payments_number {
```
Lezek123: `1..payments_number` range will only cover `payments_number - 1` numbers, for example, if `payments_number` is `2` the range will only contain `[1]`, if the `payments_number` is `4` is will be `[1, 2, 3]` etc.
You should use `1..=payments_number` instead to make the range inclusive

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,167 @@ benchmarks! {
...
+
+        Pallet::<T>::set_channel_paused_features_as_moderator(
+            origin.clone().into(),
+            crate::ContentActor::Lead,
+            channel_id,
+            super::all_channel_pausable_features_except(BTreeSet::from_iter(vec![
+                crate::PausableChannelFeature::CreatorCashout,
+            ])),
+            b"reason".to_vec(),
+        ).unwrap();
```
Lezek123: This is a nice feature, I think it's worth factoring out to another utility function like `set_channel_paused_features_all_except(features)` so that it can be used in other benchmarks too (without too much boilerplate)

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,167 @@ benchmarks! {
...
+            ..Default::default()
+        })?;
+
+        let actor = crate::ContentActor::Lead;
+        let item = payments[0].clone();
+        T::CouncilBudgetManager::set_budget(cumulative_reward_claimed + T::ExistentialDeposit::get());
+    }: _ (origin, actor, proof, item)
+        verify {
+            let cashout = item
+                 .cumulative_reward_earned - cumulative_reward_claimed;
```
Lezek123: Unused variable

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,167 @@ benchmarks! {
...
```
Lezek123: Unused variable

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -937,6 +939,134 @@ where
...
+        payments.push(PullPayment::<T> {
+            channel_id: (i as u64).into(),
+            cumulative_reward_earned,
+            reason: T::Hashing::hash_of(&b"reason".to_vec()),
+        });
+    }
+    payments
+}
+
+pub fn generate_merkle_root_helper<T: Config, E: Encode>(collection: &[E]) -> Vec<T::Hash> {
```
Lezek123: We already have those utility functions defined in a lot of places, even within the content pallet itself they're in both `src/benchmarks` and `src/tests`. Perhaps we should just move them to `common` pallet?

runtime-modules/content/src/benchmarks/payouts.rs
```diff
@@ -0,0 +1,49 @@
+#![cfg(feature = "runtime-benchmarks")]
```
Lezek123: This looks like a leftover file

runtime-modules/content/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 #![allow(clippy::unused_unit)]
 
 #[cfg(test)]
-mod tests;
+pub mod tests;
```
Lezek123: I suspect you initially wanted to import the merkle tree utilities from `tests`, but then you copied them to `benchmarks/mod.rs`, so making this module public is no longer needed? Same would apply to `fixtures.rs` etc.
As I suggested before, I think it's best to just move those merkle tree utils to `common` pallet so that we don't need to define them in so many places.

runtime-modules/content/src/lib.rs
```diff
@@ -1343,7 +1344,16 @@ decl_module! {
...
-        #[weight = 10_000_000] // TODO: adjust Weight
+        /// Update channel payouts
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)` where:
+        /// - DB:
+        /// - O(1)
+        /// # </weight>//
```
```suggestion
        /// # </weight>
```

runtime-modules/content/src/lib.rs
```diff
@@ -1403,7 +1413,17 @@ decl_module! {
...
+        /// Claim reward in JOY from channel account
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (H)` where:
+        /// - `H` is the lenght of the provided merkle `proof`
+        /// - DB:
+        ///    - O(1)
+        /// # </weight>//
```
```suggestion
        /// # </weight>
```

runtime-modules/content/src/lib.rs
```diff
@@ -1426,7 +1446,16 @@ decl_module! {
...
-        #[weight = 10_000_000] // TODO: adjust Weight
+        /// Claim reward in JOY from channel account
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)`
+        /// - DB:
+        ///    - O(1)
+        /// # </weight>//
```
```suggestion
        /// # </weight>
```

runtime-modules/content/src/lib.rs
```diff
@@ -1517,7 +1546,17 @@ decl_module! {
...
+        /// Claim and withdraw reward in JOY from channel account
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (H)` where:
+        /// - `H` is the lenght of the provided merkle `proof`
+        /// - DB:
+        ///    - O(1)
+        /// # </weight>//
```
```suggestion
        /// # </weight>
```

scripts/generate-weights.sh
```diff
@@ -44,7 +44,7 @@ joystream_pallet_benchmark() {
       --chain=dev \
       --steps=$STEPS \
       --repeat=$REPEAT \
-      --execution=wasm \
+      --execution=native \
```
Lezek123: Looks like those changes in this file have been accidently commited (this one and the uncommenting of `joystream_pallet_benchmark bounty bounty`)

Lezek123 (changes requested):
1. The runtime fails to compile: https://github.com/Joystream/joystream/runs/7694107015?check_suite_focus=true
2. The formatting seems off in a lot of places in `benchmarking.rs` because `cargo fmt` doesn't work inside macros. Unfortunately I don't think there's any other way to fix this than manually, ie.:
    1. Making sure lines don't exceed 100 characters
    2. Fixing the indentation
    3. Consistent spacing, for example, changing `let i= 1u32;` to `let i = 1u32;`
3. I'm not sure if in case of channel withdrawals the curator channels are indeed the worst case scenario, because for member channel the execution path is very different, ie. `balances::transfer` is triggered instead of `balances::slash` + ` T::CouncilBudgetManager::increase_budget`. Did you compare those two cases?

dobertRowneySr:
I got these results for `withdraw_channel_balance`
##### With curator channel
``` rust
		(49_206_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(5 as Weight))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
```
##### With member channel
```rust
		(74_204_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(4 as Weight))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
```
The base weight is higher on members, I guess due to the transfer involved,
so I kept the latter (but I left the original benchmark commented, in case you want to verify on your machine)

runtime-modules/content/src/tests/mod.rs
```diff
@@ -3,7 +3,7 @@
 mod channels;
 mod creator_tokens;
 mod curators;
-mod fixtures;
+pub(crate) mod fixtures;
```
Lezek123: This change can be rolled-back

runtime-modules/content/src/tests/fixtures.rs
```diff
@@ -4889,7 +4889,7 @@ fn index_path_helper(len: usize, index: usize) -> Vec<IndexItem> {
     path
 }
 
-pub fn generate_merkle_root_helper<E: Encode>(collection: &[E]) -> Vec<HashOutput> {
+pub fn generate_merkle_root_helper<T: Config, E: Encode>(collection: &[E]) -> Vec<T::Hash> {
```
Lezek123: Those helper functions can now be imported from `common` I think

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -783,6 +788,75 @@ where
     Ok(channel_id)
 }
 
+fn setup_worst_case_scenario_member_channel<T: Config>(
```
Lezek123, line 785: Most of the code in that function is just a duplicate of `setup_worst_case_scenario_channel` (which can be used to setup both a member and a curator channel), while some other parts are just not needed (like `setup_worst_case_curator_group_with_curators`)

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,181 @@ benchmarks! {
...
+    //         assert_eq!(
+    //             Balances::<T>::usable_balance(ContentTreasury::<T>::account_for_channel(channel_id)),
+    //             T::ExistentialDeposit::get(),
+    //         );
+    //     }
+
+    withdraw_from_channel_balance {
+        let (channel_id, member_id, member_account_id, lead_account_id) =
+            setup_worst_case_scenario_member_channel::<T>(
+                0,
```
Lezek123:
```suggestion
                T::MaxNumberOfAssetsPerChannel::get(),
```

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,181 @@ benchmarks! {
...
+    // - channel has all feature paused except the necessary for the extr. to succeed to maximize permission validation complexity
+    claim_channel_reward {
+        let h in 1 .. MAX_MERKLE_PROOF_HASHES;
+
+        let cumulative_reward_claimed: BalanceOf<T> = 100u32.into();
+        let payments = create_pull_payments_with_reward::<T>(2u32.pow(h), cumulative_reward_claimed);
+        let commitment = generate_merkle_root_helper::<T, _>(&payments).pop().unwrap();
+        let proof = build_merkle_path_helper::<T, _>(&payments, 0);
+        let (channel_id, group_id, lead_account_id, _, _) =
+            setup_worst_case_scenario_curator_channel::<T>(0,T::StorageBucketsPerBagValueConstraint::get().min as u32,T::DistributionBucketsPerBagValueConstraint::get().min as u32, false)?;
```
Lezek123: 1. This line is way to long
2. You should use `T::MaxNumberOfAssetsPerChannel::get()` as the number of assets
3. You can use `setup_worst_case_scenario_curator_channel_all_max(false)` instead to solve both issues

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,181 @@ benchmarks! {
...
+    // - channel has all feature paused except the necessary for the extr. to succeed to maximize permission validation complexity
+    claim_and_withdraw_channel_reward {
+        let h in 1 .. MAX_MERKLE_PROOF_HASHES;
+
+        let cumulative_reward_claimed: BalanceOf<T> = 100u32.into();
+        let payments = create_pull_payments_with_reward::<T>(2u32.pow(h), cumulative_reward_claimed);
+        let commitment = generate_merkle_root_helper::<T, _>(&payments).pop().unwrap();
+        let proof = build_merkle_path_helper::<T, _>(&payments, 0);
+        let (channel_id, group_id, lead_account_id, _, _) =
+            setup_worst_case_scenario_curator_channel::<T>(0,T::StorageBucketsPerBagValueConstraint::get().min as u32,T::DistributionBucketsPerBagValueConstraint::get().min as u32, false)?;
```
Lezek123: Same as the previous comment - I suggest using `setup_worst_case_scenario_curator_channel_all_max(false)` instead

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -518,6 +509,181 @@ benchmarks! {
...
+            assert_eq!(
+                Pallet::<T>::commitment(),
+                hash,
+            );
+        }
+
+    // WORST CASE SCENARIO:
+    // - curator channel belonging to a group with max number curator and max curator permissions
+    // - channel has all feature paused except the necessary for the extr. to succeed to maximize permission validation complexity
+    // withdraw_from_channel_balance {
```
Lezek123: So personally I'm getting:
```rust
        // Curator
	fn withdraw_from_channel_balance_curator() -> Weight {
		(59_104_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(5 as Weight))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
	}

        // Member
	fn withdraw_from_channel_balance() -> Weight {
		(56_962_000 as Weight)
			.saturating_add(T::DbWeight::get().reads(4 as Weight))
			.saturating_add(T::DbWeight::get().writes(2 as Weight))
	}
```

Which means curator channel is actually a worse case.
I think the difference you experienced may caused by using `--execution=wasm` in one case and `--execution=native` in the other.

In any case, I think it's worth to keep both those benchmarks, so I suggest:
1. Uncomment this function and rename to `withdraw_from_channel_balance_member_channel`
2. Rename the other one to `withdraw_from_channel_balance_curator_channel`
3. Keep tests for both under `pub mod tests`
4. In the content pallet - either use `#[weight = WeightInfoContent::<T>::withdraw_from_channel_balance_member_channel().max(WeightInfoContent::<T>::withdraw_from_channel_balance_curator_channel())]` or just choose the worse case depending on the results of weight generation.

The same may make sense for `claim_and_withdraw` also

runtime/src/weights/frame_system.rs
```diff
@@ -18,7 +18,7 @@
 //! Autogenerated weights for frame_system
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-06-27, STEPS: `1`, REPEAT: 1, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2022-08-12, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
Lezek123: We don't need to update those weights now, so I suggest removing any changes inside `runtime/src/weights` from this PR for clarity.

scripts/generate-weights.sh
```diff
@@ -16,7 +16,7 @@ substrate_pallet_benchmark() {
...
```
Lezek123: Please rollback to `wasm`

scripts/generate-weights.sh
Lezek123: Please rollback to `wasm`

scripts/generate-weights.sh
```diff
@@ -90,7 +90,7 @@ joystream_pallet_benchmark referendum referendum
 joystream_pallet_benchmark forum forum
 joystream_pallet_benchmark membership membership
 # Disabled until we merge new bounty module
-# joystream_pallet_benchmark bounty bounty
+joystream_pallet_benchmark bounty bounty
```
Lezek123: This is causing the CI checks to fail, please rollback this change.

runtime-modules/content/src/lib.rs
```diff
@@ -1426,7 +1446,16 @@ decl_module! {
             Ok(())
         }
 
-        #[weight = 10_000_000] // TODO: adjust Weight
+        /// Claim reward in JOY from channel account
```
```suggestion
        /// Withdraw JOY from channel account
```

Lezek123 (changes requested):

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -684,6 +684,47 @@ benchmarks! {
                 cumulative_reward_claimed + T::ExistentialDeposit::get(),
             );
         }
+
+    // Worst case scenario:
+    // - curator channel belonging to a member with max number of collaborators and max agent permissions
```
Lezek123:
```suggestion
    // - channel belonging to a member with max number of collaborators and max agent permissions
```

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -800,11 +800,9 @@ where
     T: RuntimeConfig,
     T::AccountId: CreateAccountId,
 {
-
     let (_, lead_account_id) = insert_content_leader::<T>();
 
-    let _ =
-        setup_worst_case_curator_group_with_curators::<T>(max_curators_per_group::<T>())?;
+    let _ = setup_worst_case_curator_group_with_curators::<T>(max_curators_per_group::<T>())?;
```
Lezek123: You don't need to setup a curator group for member channel

scripts/generate-weights.sh
Lezek123: Still not fixed

scripts/generate-weights.sh
Lezek123: Still not fixed

runtime/src/weights/frame_system.rs
Lezek123: Previous review comment not addressed

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -60,6 +56,13 @@ const fn gen_array_u128<const N: usize>(init: u128) -> [u128; N] {
     res
 }
 
+pub const MEMBER_IDS_INIT: u128 = 500;
+pub const MAX_MEMBER_IDS: usize = 100;
+
+pub const MEMBER_IDS: [u128; MAX_MEMBER_IDS] = gen_array_u128::<MAX_MEMBER_IDS>(MEMBER_IDS_INIT);
+
+pub const DEFAULT_MEMBER_ID: u128 = MEMBER_IDS[0];
```
Lezek123: Except `DEFAULT_MEMBER_ID` those constants now don't serve any purpose

runtime-modules/content/src/lib.rs
```diff
@@ -1426,7 +1446,16 @@ decl_module! {
...
+        /// Withdraw JOY from channel account
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)`
+        /// - DB:
+        ///    - O(1)
+        /// # </weight>
+        #[weight = WeightInfoContent::<T>::withdraw_from_member_channel_balance().max(WeightInfoContent::<T>::withdraw_from_curator_channel_balance())]
```
Lezek123: Respecting max. line length:
```suggestion
        #[weight = WeightInfoContent::<T>::withdraw_from_member_channel_balance()
            .max(WeightInfoContent::<T>::withdraw_from_curator_channel_balance())]
```

runtime-modules/content/src/lib.rs
```diff
@@ -1517,7 +1546,17 @@ decl_module! {
...
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (H)` where:
+        /// - `H` is the lenght of the provided merkle `proof`
+        /// - DB:
+        ///    - O(1)
+        /// # </weight>
+        #[weight = WeightInfoContent::<T>::claim_and_withdraw_member_channel_reward(proof.len() as u32).max(WeightInfoContent::<T>::claim_and_withdraw_curator_channel_reward(proof.len() as u32))]
```
Lezek123: Respecting max. line length:
```suggestion
        #[weight =
            WeightInfoContent::<T>::claim_and_withdraw_member_channel_reward(proof.len() as u32)
                .max(WeightInfoContent::<T>::claim_and_withdraw_curator_channel_reward(
                    proof.len() as u32
                ))]
```

Lezek123 (changes requested):

Lezek123 (approved):
LGTM

PR #4102, opened 2022-08-01
Carthage content benchmarking: Update from carthage

Lezek123:
1. Merged current `carthage`
2. Some updates related to no longer existing `storage::Module::get_next_data_object_ids()` function
3. Updated existing content benchmarks in https://github.com/Joystream/joystream/pull/4102/commits/e8718e5a6e734db14118adc9cad1ec34fc710ccf: Distribution buckets no longer need to be taken into account when updating a channel (after https://github.com/Joystream/joystream/pull/4082)


L3pereira (changes requested):

runtime-modules/content/src/errors.rs
```diff
@@ -29,9 +29,16 @@ decl_error! {
         /// Channel bag witness parameters don't match the current runtime state
         InvalidChannelBagWitnessProvided,
 
-        /// Channel bag witness must be provided when channel/video assets are being updated
+        /// Channel bag witness must be provided when channel is being deleted
         MissingChannelBagWitness,
```
L3pereira: I guess this should be removed since is not being used.

L3pereira (changes requested):

L3pereira (approved):
LGTM

Issue #4109, opened 2022-08-01
Query node crashes when deleting channel assets as moderator


Lezek123:
```
processor                | Content has been disconnected from Channel { channelId: '3', dataObjectId: '4' }
processor                | Content has been disconnected from Channel { channelId: '3', dataObjectId: '5' }
processor                | Channel assets have been removed { ids: [ <BN: 4>, <BN: 5> ] }
processor                | ERROR: Stopping the proccessor due to errors: {
processor                |   "query": "INSERT INTO \"channel_assets_deleted_by_moderator_event\"(\"id\", \"created_at\", \"created_by_id\", \"updated_at\", \"updated_by_id\", \"deleted_at\", \"deleted_by_id\", \"version\", \"in_extrinsic\", \"in_block\", \"network\", \"index_in_block\", \"actor\", \"channel_id\", \"asset_ids\", \"rationale\") VALUES ($1, $2, $3, $4, DEFAULT, DEFAULT, DEFAULT, $5, $6, $7, $8, $9, DEFAULT, $10, $11, $12) RETURNING \"version\"",
processor                |   "parameters": [
processor                |     "OLYMPIA-5255-6",
processor                |     "2022-08-01T18:01:22.001Z",
processor                |     "-",
processor                |     "2022-08-01T18:01:22.001Z",
processor                |     1,
processor                |     "0x07af6e63dc4b7176ddbfc9c70ed6ec51806cad59c646b884e0a0b097c4a2b718",
processor                |     5255,
processor                |     "OLYMPIA",
processor                |     6,
processor                |     3,
processor                |     [
processor                |       4,
processor                |       5
processor                |     ],
processor                |     "Just testing"
processor                |   ],
processor                |   "driverError": {
processor                |     "length": 409,
processor                |     "name": "error",
processor                |     "severity": "ERROR",
processor                |     "code": "23502",
processor                |     "detail": "Failing row contains (OLYMPIA-5255-6, 2022-08-01 18:01:22.001, -, 2022-08-01 18:01:22.001, null, null, null, 1, 0x07af6e63dc4b7176ddbfc9c70ed6ec51806cad59c646b884e0a0b097c4a2b7..., 5255, OLYMPIA, 6, null, 3, {4,5}, Just testing).",
processor                |     "schema": "public",
processor                |     "table": "channel_assets_deleted_by_moderator_event",
processor                |     "column": "actor",
processor                |     "file": "execMain.c",
processor                |     "line": "1948",
[...]

Issue #4108, opened 2022-08-01
QN: StorageSystemParameters.dataObjectStateBloatBondValue should be BigInt


Lezek123:
When updating the data object state bloat bond to a value greater than Int32 max:

```
processor                | ERROR: Stopping the proccessor due to errors: {
processor                |   "query": "UPDATE \"storage_system_parameters\" SET \"updated_at\" = $1, \"blacklist\" = $2, \"data_object_state_bloat_bond_value\" = $3, \"version\" = \"version\" + 1 WHERE \"id\" IN ($4) RETURNING \"version\"",
processor                |   "parameters": [
processor                |     "2022-08-01T17:14:51.000Z",
processor                |     [],
processor                |     10000000000,
processor                |     "1"
processor                |   ],
processor                |   "driverError": {
processor                |     "length": 112,
processor                |     "name": "error",
processor                |     "severity": "ERROR",
processor                |     "code": "22003",
processor                |     "file": "numutils.c",
processor                |     "line": "250",
processor                |     "routine": "pg_strtoint32"
processor                |   },
processor                |   "length": 112,
processor                |   "severity": "ERROR",
processor                |   "code": "22003",
processor                |   "file": "numutils.c",
processor                |   "line": "250",
processor                |   "routine": "pg_strtoint32"
processor                | } name: QueryFailedError, message: value "10000000000" is out of range for type integer, stack: QueryFailedError: value "10000000000" is out of range for type integer
processor                |     at QueryFailedError.TypeORMError [as constructor] (/joystream/node_modules/typeorm/error/TypeORMError.js:9:28)
processor                |     at new QueryFailedError (/joystream/node_modules/typeorm/error/QueryFailedError.js:13:28)
processor                |     at PostgresQueryRunner.<anonymous> (/joystream/node_modules/typeorm/driver/postgres/PostgresQueryRunner.js:294:31)
processor                |     at step (/joystream/node_modules/tslib/tslib.js:144:27)
processor                |     at Object.throw (/joystream/node_modules/tslib/tslib.js:125:57)
processor                |     at rejected (/joystream/node_modules/tslib/tslib.js:116:69)
processor                |     at processTicksAndRejections (internal/process/task_queues.js:95:5)
processor                | ERROR: {} name: Error, message: QueryFailedError: value "10000000000" is out of range for type integer, stack: Error: QueryFailedError: value "10000000000" is out of range for type integer
[...]

PR #4110, opened 2022-08-01
Carthage Query Node: Fixes for processor-crashing issues

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4108 and https://github.com/Joystream/joystream/issues/4109


zeeshanakram3 (approved):
LGTM, thanks for the fixes.

Issue #3955, opened 2022-06-25
Remove `sudo_create_dynamic_bag`


bedeho:
SSIA.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202504906935885) by [Unito](https://www.unito.io)

mnaamani:
Dropped in https://github.com/Joystream/joystream/pull/4111

Issue #3953, opened 2022-06-25
Remove `sudo_upload_data_objects`


bedeho:
SSIA.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202504811248305) by [Unito](https://www.unito.io)

mnaamani:
Addressed in https://github.com/Joystream/joystream/pull/4111

PR #4111, opened 2022-08-02
Drop sudo test dispatchables

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/3953 - Instead of dropping `sudo_upload_data_objects`, I removed dependency of the benchmarks on it, and made it a no-op in production builds. It is useful when testing storage node and distributor nodes, there is are `dev` cli commands that use this extrinsic. I avoided dropping it entirely depending on the cargo feature so as to not have a scenario where there would be two different metadata/types depending on the build.

Fixes https://github.com/Joystream/joystream/issues/3955 - dropped `sudo_create_dynamic_bag` it is not used.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202705202634323) by [Unito](https://www.unito.io)


runtime-modules/storage/src/lib.rs
```diff
@@ -2964,31 +2964,25 @@ decl_module! {
...
 
             //
             // == MUTATION SAFE ==
             //
 
+            #[cfg(any(
+                feature = "staging_runtime",
+                feature = "testing_runtime",
+                feature = "playground_runtime"
+            ))]
```
Lezek123: This way the extrinsic can still be executed in the production runtime, it even returns `Ok()`, but it's just a no-op.

I think it can be a little confusing. What do you think about using call filter instead? This way we also wouldn't need to expose those configuration flags into the storage pallet, we could keep all the configuration logic in the `runtime/lib.rs`

mnaamani: The call filter is certainly cleaner, and it did occurred to me, I was just concerned we might forget about this particular extrinsic when we remove all the filters at some point in the future. At which point the filter would be there only to block this extrinsic..

Lezek123: We have separate functions like `filter_non_essential`, `filter_content_and_proposals` so if we introduce another one like `filter_sudo_calls` I think the semantics will be quite clear.

I'll leave it up to you to decide, but I think in any case we should return `Err` in the "disabled" version.

mnaamani: Okay I think I will go with your suggestion, when I do it through the filter the extrinsic will fail with Err `CallFiltered`

mnaamani:
After realizing call filter is bypassed for sudo dispatches, reverted to using features to do conditional invocation in sudo_upload_data_objects, and returning a new error type CallDisabled in production build.

runtime-modules/storage/src/lib.rs
```diff
@@ -2964,29 +2967,22 @@ decl_module! {
             );
         }
 
-        // ===== Sudo actions (development mode) =====
-
         /// Upload new data objects. Development mode.
         #[weight = 10_000_000]
+        #[allow(dead_code)]
+        #[allow(unused_variables)]
```
Lezek123: Those are not needed anymore
```suggestion
```

Lezek123 (approved):

Issue #3866, opened 2022-06-07
Channel agent permissions


dmtrjsg:
Full description and accurate representation of scope will be provided by ondratra

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202862000428983) by [Unito](https://www.unito.io)

bedeho:
Note https://github.com/Joystream/joystream/issues/3874#issuecomment-1151030225

ondratra:
A description of this issue can be seen in the original issue for the runtime https://github.com/Joystream/joystream/issues/3436 and the implementing PR https://github.com/Joystream/joystream/pull/3561.

dmtrjsg:
@bedeho @ondratra is there a PR corresponding to this issue?

ondratra:
> @bedeho @ondratra is there a PR corresponding to this issue?

https://github.com/Joystream/joystream/pull/4150

PR #4150, opened 2022-08-10
qn - channel agent permissions I

ondratra:
Fixes https://github.com/Joystream/joystream/issues/3866.

Brings QN part of changes to channel agent permissions done in https://github.com/Joystream/joystream/pull/3561. Related PRs that brought some pieces needed for this feature are https://github.com/Joystream/joystream/pull/4030 and https://github.com/Joystream/joystream/pull/3726 (updated events in QN, etc.).

In the current state, there is a missing Hydra/Warthog feature - array of enums - that is blocking this PR and is not yet overcome. Also, the code is not covered by integration tests yet.


ondratra:
Due to limitations described in https://github.com/Joystream/hydra/issues/507, the schema has been rewritten, and there is a `ChannelAgentPermissions` and `CuratorAgentPermissions` for each permission granted (previously, all permissions granted to the member for a given channel/curatorGroup were represented in a single record).

It is also worth noting that some seemingly redundant data were introduced in the schema. For example, channel's `collaborators`
and `collaboratorsPermissions` will have similar records (`collaboratorsPermissions` will have some extra info). This is desired state and it will extend existing querying options without removing any. 

In a perfect world, where all Hydra issues are solved, only `collaboratorsPermissions` would be needed, but with current nested filtering limitations and other issues, it's better to have it this way.

query-node/mappings/src/content/channel.ts
```diff
@@ -332,3 +345,54 @@ export async function content_ChannelAgentRemarked(ctx: EventContext & StoreCont
     await updateMetaprotocolTransactionStatus(store, metaprotocolTxIdentifier, statusErrored, e)
   }
 }
+
+export async function ChannelTransferAccepted({ store, event }: EventContext & StoreContext): Promise<void> {
```
zeeshanakram3: Extrinsics related to channel transfers are disabled right now, so don't think this mapping is needed.

query-node/schemas/content.graphql
```diff
@@ -365,3 +368,42 @@ type Comment @entity {
...
+  channel: Channel!
+
+  "Related member"
+  member: Membership!
+
+  # unfortunately permissions can't be modeled as ist of enums
+  # `permissions: [ChannelActionPermission!]!`
+  # rework it after this feature is available https://github.com/Joystream/hydra/issues/507
+  "List of member's permissions"
+  permission: ChannelActionPermission!
```
zeeshanakram3: Since hydra doesn't support a list of enums, so for a missing feature I think we don't need to complicate the schema design. We can have `permissions` as a list of Strings, e.g. `permissions: [String!]!`. Though this way, we won't have Db enforced checks, I think we can mitigate this with better checks in the mappings.

query-node/schemas/contentNft.graphql
```diff
@@ -263,3 +269,17 @@ type Bid @entity {
...
+  curatorGroup: CuratorGroup!
+
+  "Related member"
+  curator: Curator!
+
+  # unfortunately permissions can't be modeled as list of enums
+  # `permissions: [ChannelActionPermission!]!`
+  # rework it after this feature is available https://github.com/Joystream/hydra/issues/507
+  "List of member's permissions"
+  permission: ChannelActionPermission!
```
zeeshanakram3: Same as suggested above. We can have permissions as a list of Strings, e.g., permissions: [String!]!.

query-node/schemas/content.graphql
```diff
@@ -70,6 +70,9 @@ type Channel @entity {
   "List of channel collaborators (members)"
   collaborators: [Membership!]
```
zeeshanakram3: Also, I think we no longer need a separate `collaborators` field, collaborator members are already part of `ChannelAgentPermissions` entity; also better to rename `ChannelAgentPermissions` to `Collaborators`.

zeeshanakram3 (commented):
Looks good overall, please address the following points.

query-node/mappings/src/content/channel.ts
ondratra: Are you sure about this? I haven't found any place in runtime code that suggests that in `carthage` branch.

Anyway, I've removed the event. It can be reintroduced easily by reverting https://github.com/Joystream/joystream/pull/4150/commits/f0fbb1c01426555ca0446228f1ae62aa0b0402ec.

query-node/schemas/content.graphql
ondratra: Done. No extra checks are needed because QN only displays data - all the checks are done in the runtime.

query-node/schemas/contentNft.graphql
ondratra: Done

query-node/schemas/content.graphql
ondratra: Done. I just hope it doesn't reduce querying possibilities. In case it does, the redundancy can be reintroduce by reverting https://github.com/Joystream/joystream/pull/4150/commits/84f912d597294a8531fd9a9538c81c95966ba91b.

ondratra:
Points from the previous review have all been addressed. One of the tests is failing due to some unrelated connection issue but succeeds when run locally.

query-node/schemas/contentNft.graphql
```diff
@@ -263,3 +264,17 @@ type Bid @entity {
 }
 
 # TODO entity for (cancelable) offers; will be needed to see history of offers
+
+type CuratorAgentPermissions @entity {
```
zeeshanakram3, line 268: Can't this entity be named just `Curator`, similar to how `Collaborator` is named. Also in [Curator](https://github.com/Joystream/joystream/pull/4150/files#diff-ddf70c18d6730ebb157f4fa1dc7839a954e2807eb69de6ab907903a10e2ca9cdR44) the field `curatorGroups` seems confusing as its a list of `CuratorAgentPermissions`. Is it possible to remove [this](https://github.com/Joystream/joystream/pull/4150/files#diff-ddf70c18d6730ebb157f4fa1dc7839a954e2807eb69de6ab907903a10e2ca9cdR44) type, and use `CuratorAgentPermissions` renamed as `Curator` instead

zeeshanakram3 (commented):
LGTM. Would it be possible for you to add some integration tests, so we are confident about merging it? Thanks

query-node/mappings/src/content/channel.ts
zeeshanakram3: Yes, some extrinsics were disabled in #3914 including channel transfers actions

query-node/schemas/contentNft.graphql
ondratra: I don't think that's possible because Curator can be in multiple curator groups and can be a ContentActor as well.

ondratra:
Integration tests have been added. 1 test failed due to unrelated connection issues.

query-node/mappings/src/content/channel.ts
```diff
@@ -276,6 +281,38 @@ export async function content_ChannelAgentRemarked(ctx: EventContext & StoreCont
...
+    where: { channel: { id: channel.id.toString() } },
+  })
+  for (const agentPermissions of collaborators) {
+    await store.remove(agentPermissions)
+  }
+
+  // create new records for privledged members
+  for (const [memberId, permissions] of Array.from(collaboratorsPermissions)) {
+    const permissionsArray = Array.from(permissions)
+    if (!permissionsArray.length) {
```
zeeshanakram3: Just curious why this check is there. Since you can create on-chain collaborators without any permissions, too, so won't this check create inconsistency b/w runtime and QN state.

query-node/mappings/src/content/curatorGroup.ts
```diff
@@ -116,19 +120,49 @@ export async function content_CuratorRemoved({ store, event }: EventContext & St
...
+      curatorGroup: { id: curatorGroup.id.toString() },
+      curator: { id: curator.id.toString() },
+    },
+  })
+  for (const agentPermissions of existingAgentPermissions) {
+    await store.remove(agentPermissions)
+  }
+
+  const permissionsArray = Array.from(permissions || [])
+  if (!permissionsArray.length) {
```
zeeshanakram3: As pointed above,  won't this check create inconsistency b/w runtime and QN state

query-node/mappings/src/content/curatorGroup.ts
```diff
@@ -116,19 +120,49 @@ export async function content_CuratorRemoved({ store, event }: EventContext & St
...
-  // ensure curator group exists
-  if (curatorIndex < 0) {
-    return inconsistentState('Non-associated curator removal from curator group requested', curatorId)
-  }
-
-  // update curator group
-  curatorGroup.curators.splice(curatorIndex, 1)
-
   // save curator group
   await store.save<CuratorGroup>(curatorGroup)
```
zeeshanakram3: Don't think we need to save `CuratorGroup` here?

query-node/mappings/src/content/channel.ts
ondratra: A relic from the previous implementation that's no longer needed. Removed.

query-node/mappings/src/content/curatorGroup.ts
ondratra: A relic from the previous implementation that's no longer needed. Removed.

query-node/mappings/src/content/curatorGroup.ts
ondratra: A relic from the previous implementation that's no longer needed. Removed.

ondratra:
All remaining issues have been addressed.

zeeshanakram3 (approved):
LGTM

ondratra:
The recent conflict with the carthage branch has been resolved. @mnaamani please merge this PR asap before another conflict arises.

Issue #4162, opened 2022-08-11
Plan: Fixing existential deposit, bloat bonds, locking and slashing issues


Lezek123:
#### Related issues:
* https://github.com/Joystream/joystream/issues/3760
  * bloat bond returning case
  * `MUTATION SAFE` issue no longer applies, as we have protection for that on a framework level now
* https://github.com/Joystream/joystream/issues/4002 (partially)
  * the part that applies to bloat bonds
* https://github.com/Joystream/joystream/issues/3653 (fully)
* https://github.com/Joystream/joystream/issues/4033 (fully)
* https://github.com/Joystream/joystream/issues/3509 (fully)

#### Fixing plan:
1. [x] Never use slashing for bloat bonds. Make sure all bloat bonds are always transferred to the corresponding module account and recoverable (also through transfer) from that account.
1. [x] Make sure all static module accounts are initialized with existential deposit at genesis
    * [x] storage treasury
    * [x] content treasury **_(fixed)_**
    * [x] project token treasury
1. [x] Make sure all dynamic bloat bonds are higher than existential deposit (enforce this if possible), because:
  * for dynamic module accounts (for example: channel module accunts), the bloat bond should simultaneously serve as an existential deposit:
      * [x] channel accounts
      * [x] thread accounts (forum)
      * <s>thread accounts (proposals)</s> (those should be replaced with a single `proposals_discussion` module account, as there are no thread deposits and there's no logic in place that would make this replacement infeasible)
  * <s>transaction to recover a bloat bond may fail otherwise (with `Error::ExistentialDeposit`)</s> (this is not really a breaking issue)
1. [x] Make sure dynamic module accounts (for example: channel accounts), use transfer with `ExistentialRequirement::AllowDeath` when returning the bloat bond (ie. on channel deletion)
1. [x] Update tests and benchmarks

#### Related issues:
* https://github.com/Joystream/joystream/issues/3414 (fully)
* https://github.com/Joystream/joystream/issues/3569 (fully)
  * makes sure locked funds used to pay bloat bonds are not recoverable as usable balance
* https://github.com/Joystream/joystream/issues/4126 (fully)

#### Fixing plan:
1. [x] Implement a new mechanism for checking and paying transactions costs
    <s>**_Draft_**</s> **_EXPIRED, see: https://github.com/Joystream/joystream/issues/4162#issuecomment-1213451412)_**
    <s>
    ```rust
    // Describes a single cost (for example: a bloat bond) associated with an extrinsic
[...]

PR #4164, opened 2022-08-11
Payment fixes: Bloat bonds, existential deposits, locks

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4162 and https://github.com/Joystream/joystream/issues/4208

Currently expected locks behavior: https://gist.github.com/Lezek123/7eb89937fcbffabefe73fede0b336943

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202814403191323) by [Unito](https://www.unito.io)


node/src/chain_spec/mod.rs
```diff
@@ -265,6 +265,7 @@ pub fn testnet_genesis(
         storage: storage_cfg,
         referendum: council_config::create_referendum_config(),
         project_token: Default::default(),
+        proposals_discussion: Default::default(),
```
bedeho, line 268: Why was this needed now?

runtime/src/lib.rs
```diff
@@ -1079,7 +1079,7 @@ parameter_types! {
     pub const MaxPostsInThread: u64 = 20;
     pub const MaxModeratorsForCategory: u64 = 20;
     pub const MaxCategories: u64 = 40;
-    pub const ThreadDeposit: Balance = 25 * currency::CENTS;
+    pub const ThreadDeposit: Balance = ExistentialDeposit::get() + 25 * currency::CENTS; // Must be higher than ExistentialDeposit!
```
dobertRowneySr, line 1082: Thread deposit should serve the same function as a bloat bond, correct? 
That is why we are requiring it to be `>=` than the `ExistentialDeposit`

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Ensure all given costs can be covered by a given account
+pub fn ensure_can_cover_costs<T: frame_system::Config + balances::Config>(
+    account_id: &T::AccountId,
+    costs: &[Cost<T::Balance>],
+) -> Result<T::Balance, DispatchError> {
+    let total_cost = costs
+        .iter()
+        .fold(Some(T::Balance::zero()), |acc, cost| {
```
dobertRowneySr: check `try_fold` for this: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold

runtime-modules/common/src/bloat_bond.rs
```diff
@@ -0,0 +1,65 @@
...
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+use sp_runtime::DispatchError;
+use sp_std::fmt::Debug;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, TypeInfo)]
+pub struct RepayableBloatBond<AccountId, Balance> {
+    /// Account where the bloat bond will be repaid.
+    account: Option<AccountId>, // We're using Option to sidestep the Default impl issue
```
dobertRowneySr: I have been thinking about this, instead of using like `Option<AccountId>` can't you just use the newtype pattern (https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html) like:
``` rust
pub struct DefaultAccountId(AccountId);
impl Default for DefaultAccountId { 
    fn default() -> Self {
        Self ( AccountId::zero() )
    }
}
impl Into<AccountId> for DefaultAccount {
   fn into(self) -> AccountId {
       self.0
   }
}
```

dobertRowneySr: This should help avoiding all `Option`s calls for accessing the account value, and will allow you to `#derive[Default]` on `RepayableBloatBond`

dobertRowneySr: I think using `0x0` as default is fine. It is a standard way to check that an account is not valid in solidity programming

dobertRowneySr: also how about calling it `payer` or something similar that way I know it is the account that bloates the state

bedeho: This was made `Option` to provide `RepayableBloatBond` natural default constructor, but as mentioned in discussion, we should consistently apply the pattern of wrapping the value type `V` in any storage map `K -> V` in an `Option` instead, so the map becomes `K -> Option<V>`, and not altering internal representation of `V`. This has benefit of
- a single fix which does not distort representation of `V`, even if there are multiple `AccountId` involved in type hierarchy of `V`. 
- is what we are doing other places to solve this, was @mnaamani approach when converting to Substrate v3
- is more aligned with new query format for storage definitions in the new macro format.
- is just generally safer

Now, I do think we should actually keep this  field as a wrapped for a different reason: we should only set the account here if the user is paying for the bond using funds in an account that are locked by an invite lock, i.e. the balance of the account will be less than the lock amount of an invite lock on that account. This complicates the logic a bit, but also seems more reasonable, as we don't need to restrict where funds flow back if they initially were not invite-locked.

If we go with this, then
- update field comment
- rename field to `repayment_restricted_to: Option<AccountId>`.

runtime-modules/content/src/lib.rs
```diff
@@ -3195,6 +3234,12 @@ impl<T: Config> Module<T> {
             channel.num_videos = channel.num_videos.saturating_sub(1)
         });
 
+        // Repay video state bloat bond
+        let channel_account = ContentTreasury::<T>::account_for_channel(channel_id);
+        video
+            .video_state_bloat_bond
+            .repay::<T>(&channel_account, false)?;
```
bedeho: I am seeing lots of fallible calls post the mutation safe barrier. I don't think we should have this, even though the transaction layer protects us. A best effort attempt to very clearly distinguish preconditions from side-effects is extremely valuable for clarity, and the transactional layer should really only help us with

- the tough few cases where you cannot separate checks and side-effects when doing calls across pallets
- a last line of defense against bugs

runtime-modules/content/src/tests/mock.rs
```diff
@@ -815,7 +833,10 @@ impl MemberOriginValidator<Origin, u64, u128> for TestMemberships {
 
     fn is_member_controller_account(member_id: &u64, account_id: &u128) -> bool {
         match *member_id {
-            DEFAULT_MEMBER_ID => *account_id == DEFAULT_MEMBER_ACCOUNT_ID,
+            DEFAULT_MEMBER_ID => {
```
bedeho, line 823: Why is this being done?

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,198 @@
...
+        .max()
+        .unwrap_or_else(Zero::zero);
+    let usable_funds = free_balance.saturating_sub(unusable_free_balance);
+
+    ensure!(
+        usable_funds >= cost.amount,
+        DispatchError::Other("Insufficient balance to cover the cost")
+    );
+
+    Ok(())
```
dobertRowneySr: how about having this function returning `usable_funds - cost.amount`?

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,198 @@
...
+        remaining_free_balance >= total_cost,
+        "ensure_can_cover_costs: Insufficient balance"
+    );
+
+    ensure!(
+        total_balance >= total_cost.saturating_add(T::ExistentialDeposit::get()),
+        "ensure_can_cover_costs: Insufficient balance (KeepAlive)"
+    );
+
+    for cost in costs {
```
dobertRowneySr: and then here having a `try_fold`
with 
- init `accumulator = balances::Pallet::<T>::free_balance(account_id) - T::ExistentialDeposit::get())`
-  iteration`accumulator = ensure_cost_can_be_covered_with::<T>(cost, accumulator, &locks)?`

dobertRowneySr: so that the first two `ensure!` guards should not be needed

bedeho (changes requested):
# Costly flexibility

The approach taken to allowing uses to restrict funding locks really has optimised for flexibility, for example, it allows

1. Multiple locks to be avoided.
2. Distinct set of locks to be avoided for distinct costs in the same pallet.
3. Distinct set of locks to be avoided for same costs across pallets.
4. Extrinsics are able to make unconstrained queries about how costs should be covered, allowing distinct calls over time or across extrinsics in the same pallet.
5. Same issue as 4, but for payment.

I list these, not as possible security or correctness risks, although that is also an issue, but to highlight that we have modelled an order of magnitude more flexibility than we need, and it is showing itsel in how complex the changes are. This flexbility could be worth it if we anticipated future complex evolution in how such restrictions should work, and we don't want to redo everything later, but we don't, yet we are paying a big price here.

I believe the current and long term restriction is dead simple: invitation locks can fund nearly all extrinsics, except a handful in content dir (and later bounties). I tink we would do better by just modelling that.

I want us to resolve this issue before I do any further review, as it's fundamental.

# Failing calls 

 See general remark about fallible calls past _mutation safe_ marks.

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,198 @@
...
+    ensure_cost_can_be_covered_with::<T>(&cost, free_balance, &locks)?;
+
+    // KeepAlive check
+    ensure!(
+        total_balance.saturating_sub(cost.amount) >= T::ExistentialDeposit::get(),
+        DispatchError::Other("pay_cost: KeepAlive violated")
+    );
+
+    // Calculate payer's new free balance
+    let new_free_balance = free_balance
```
dobertRowneySr: this should not be necessary after a `ensure_cost_can_be_covered_with`, correct?

node/src/chain_spec/mod.rs
Lezek123: To trigger `add_extra_genesis` in the proposals discussion pallet.
Previously we had 1 account per proposals discussion thread, I changed it to one global module account, which needs to be initialized with existential deposit on genesis.

runtime-modules/content/src/lib.rs
Lezek123: > a last line of defense against bugs

That's how I understand it in this case. If we make this call infallible we're essentially resigning from this last line of defense. This call is never supposed to fail, because the channel account should always have enough funds to repay the bloat bond, but just in case there is some unexpected state I think it's a good idea to keep this call fallible.

runtime-modules/content/src/tests/mock.rs
Lezek123: If you mean why I allowed 2 accounts to act as controller of `DEFAULT_MEMBER_ID` member: It's because this way I can test if the bloat bond is returned to the correct account when I delete a channel/video. For example: I can sign `create_channel` with `DEFAULT_MEMBER_ACCOUNT_ID` and then delete channel using `DEFAULT_MEMBER_ALT_ACCOUNT_ID`.

Lezek123:
> Distinct set of locks to be avoided for distinct costs in the same pallet.

There are valid use cases for this currently, for example, when purchasing tokens on a creator token sale one needs to pay for the purchased tokens + optionally the account bloat bond (if the account does not yet exist):
- Bloat bond is coverable with invitation-locked funds
- The cost of tokens purchased is only coverable with `usable_balance`

I think in that case this feature is very useful to have.

bedeho:
> I think in that case this feature is very useful to have.

So we have two alternatives here as I see it

1. **Current:**
  - Bloat bond is coverable with invitation-locked funds
  - The cost of tokens purchased is only coverable with usable_balance
2. **Alternative:**
  - token purchasing only coverable by usable_balance
 
The question is now how are these different in terms of functionality? If someone can buy under 2, the same user could buy under 1, however, the converse is not true, but exactly in which cases? It seems those cases are precisely when **someone has enough usable $JOY to pay for the $CRTs, but they could not afford the bloat cost, but they do have those funds under an invite lock**.

How plausible is this?

I think very implausible, because
a) the relative size of the two costs ($CRT cost in $JOY vs bloat bond) is likely orders of magnitude difference, we are after all taking about sub 10c costs for bloat bonds in general, otherwise they make all sorts of publishing prohibitive.
b) how much one buys of the $CRT can always be adjusted down in these cases to allow the user to afford the bloat bond, and this would only in fact not be possible in cases where they are trying to buy a quantity of $CRT so small that its priced lower than the bloat bond.

So to me, I think the only difference between the two is one which is of very limited utility. Even independently of the broader complexity it generates across the runtime to have the flexibility to accommodate this, just the added complexity of locally having this nuance seems not worth it.

dobertRowneySr:
The logic seems correct,  altought I'd be more comfortable taking another stab, maybe during the next iteration.
I just added a couple of restyling suggestions
As @bedeho said these are big changes and the choices you made allow for a very flexible cost management.
However since the costs, locks and bonds are pretty much set at this stage of developement. I would go for a more specific approach involving checking only the invitation / vesting locks only.
I guess we can have a chat about this approach, during the next infra call

dobertRowneySr (changes requested):
see my previous comment

runtime-modules/common/src/bloat_bond.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/de270fe393a313ba6705053b5c8e0a7e1209b5ec

runtime/src/lib.rs
Lezek123: Yes, it serves as a bloat bond and as an existential deposit for forum thread account (as there is one account per thread)

runtime-modules/common/src/costs.rs
Lezek123: This part of the code was removed following Bedeho's suggestion to simplify this functionality

runtime-modules/common/src/bloat_bond.rs
Lezek123: This was now changed with accordance to https://github.com/Joystream/joystream/pull/4164#discussion_r949872545

runtime-modules/common/src/costs.rs
Lezek123: I changed it to return the "consumed locked balance" to support the new functionality introduced in https://github.com/Joystream/joystream/pull/4164/commits/de270fe393a313ba6705053b5c8e0a7e1209b5ec

runtime-modules/common/src/costs.rs
Lezek123: This part of the code was removed following Bedeho's suggestion to simplify this functionality

runtime-modules/common/src/costs.rs
Lezek123: It's not really necessary, but I think it's a good practice to use `checked_sub` just in case some of the other logic changes and introduces a bug

runtime-modules/common/src/bloat_bond.rs
```diff
@@ -0,0 +1,74 @@
...
+};
+use scale_info::TypeInfo;
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+use sp_std::fmt::Debug;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, TypeInfo)]
+pub struct RepayableBloatBond<AccountId, Balance> {
+    /// Account where the bloat should be repaid (if restricted).
```
bedeho, line 15: Lets be even more specific: this restriction will only come when an account was used as origin with an invitation lock where the lock amount would have blocked the transaction if it was any other type of lock.

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+    traits::{
+        fungible::{Inspect, Mutate},
+        tokens::WithdrawConsequence,
+        Currency, Get, Imbalance, SignedImbalance, StoredMap,
+    },
+};
+use sp_arithmetic::traits::{CheckedSub, Saturating};
+use sp_runtime::{traits::Zero, DispatchError, DispatchResult};
+
+pub fn has_sufficient_balance_for_payment<T: frame_system::Config + balances::Config>(
```
bedeho, line 16: I think a very generous descriptive comment is warranted here, trying to unpack what a "payment" is semantically, because we mean something quite specific (bloat bonds, inclusion fees, ...), however it's a very open ended term that can refer to anything out of context.

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+    account: &T::AccountId,
+    amount: T::Balance,
+) -> bool {
+    balances::Pallet::<T>::can_withdraw(account, amount) == WithdrawConsequence::Success
+}
+
+pub fn has_sufficient_balance_for_fees<T: frame_system::Config + balances::Config>(
```
bedeho, line 28: A similar comment is useful here I think, "fees" here does not have a narrow meaning, but will easily be interpreted that way.

bedeho, line 28: As agreed, this will be simplified in this issue in a future PR: https://github.com/Joystream/joystream/issues/4201

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+        .unwrap_or_else(Zero::zero);
+    let usable_funds = free_balance.saturating_sub(unusable_free_balance);
+
+    usable_funds >= amount && total_balance >= amount.saturating_add(T::ExistentialDeposit::get())
+}
+
+// Pay a certain fee associated with a transaction (for example: a bloat bond, data size fee)
+// Fees are assumed to be payable w/ invitation-locked balance.
+// Returns the part of amount that was covered with invitation-locked balance (possibly 0).
+pub fn pay_fee<T: frame_system::Config + balances::Config>(
```
bedeho, line 62: I think this implementation is too defensive, where end up rechecking in assumptions in new restrictive ways, so much so that the opacity seems like a bigger risk than trying to catch any mistakes in usage. It is also not clear why we should go to such extreme lengths for this one case when we are in many places moving or locking funds without redoing this sort of exercise. I don't think it would be maintainable.

My understanding is this is all equivalent to just doing a slashing of the `payer`  and depositing into the `reciever: Some(x)`. For good measure we could preceed this with `debug_assert!(has_sufficient_balance_for_fees())` to catch a bad call in dev mode only

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        // Check deposit result
+        ensure!(
+            deposit_imbalance.peek() == amount,
+            DispatchError::Other("pay_cost: Unexpected deposit imbalance")
+        );
+    }
+
+    // locked_balance_used = min(account.frozen(Reasons::All).saturating_sub(new_free_balance), amount)
+    let locked_balance_used = account
```
dobertRowneySr: - `locked = account.misc_frozen.max(account.fee_frozen)` represent the locked free balance amount
-  `usable_balance = free - locked`, usable balance

Shouldn't be just
`locked_balance_used = amount - usable` if `amount > usable` and `0` otherwise ?

dobertRowneySr: in case we want to deplete usable balance first Am I missing something?

bedeho: This remark is a bit of a moot point if we are not keeping this implementation, but I will still ask: 

This was hard to penetrate, I _think_ it works in all cases, including when balance already starts out lower than invite lock, but not totally sure, but why is it `Reasons:All` and not fee? because only invite lock can pay this fee, so either they are not present, or smaller than invite lock, or just smaller than what the spending eats into.

runtime-modules/content/src/lib.rs
```diff
@@ -3373,35 +3379,47 @@ impl<T: Config> Module<T> {
...
-            ChannelOwner::CuratorGroup(_) => {
-                // The budget is sufficient. It was checked previously in functions:
-                // validate_channel_transfer_acceptance() ->
-                // ensure_sufficient_balance_for_channel_transfer()
-                let new_budget = T::ContentWorkingGroup::get_budget().saturating_sub(price);
+            (ChannelOwner::Member(old_owner_id), ChannelOwner::CuratorGroup(_)) => {
+                let old_owner_controller_acc =
+                    T::MemberAuthenticator::controller_account_id(*old_owner_id)?;
+                // Decrease working group budget.
+                let new_budget = T::ContentWorkingGroup::get_budget()
```
bedeho: Is there any room for making this interaction with the `ContentWorkingGroup` more client friendly? Its very plausible that other callers could want to do this from diferent places in the runtime, and it would be better to not reimplement and maintain the same budget interaction all those places. It would also avoid this leaking of very low level details into client code.

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+        .saturating_sub(new_free_balance)
+        .min(amount);
+
+    Ok(locked_balance_used)
+}
+
+pub fn burn_from_usable<T: frame_system::Config + balances::Config>(
+    account: &T::AccountId,
+    amount: T::Balance,
+    allow_death: bool,
```
bedeho: I'm not really understanding the logic in when `allow_death` is set, the pattern I see is summarized below. Seems to me that we should just always allow it at the core level, but that the relevant account will always have the sufficient balance prefunded to block it when it should not happen.

# Allow Death

- curator group paying for channel from user.
- pay into council budget
- moderate thread

# Don't Allow Death

- pay for NFT
- buying membership
- council withdraws channel balance to budget
- moderate post

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
+use crate::locks::InvitedMemberLockId;
```
bedeho, line 1: A general remark about all methods in this utility module: would it not be better to return proper success types, rather than literal strings, because this will help with making top level extrinsics tests assert the proper type safe error in failure cases, without needing to retype the same literal both places.

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    Ok(locked_balance_used)
+}
+
+pub fn burn_from_usable<T: frame_system::Config + balances::Config>(
+    account: &T::AccountId,
+    amount: T::Balance,
+    allow_death: bool,
+) -> DispatchResult {
+    let burned = balances::Pallet::<T>::burn_from(account, amount)?;
```
bedeho: - Is the distinction between `burn_from` and `slash` that it can't touch reserved funds?

runtime-modules/common/src/costs.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    ensure!(
```
bedeho: I don't understand why the two ensures are done here.

dobertRowneySr (changes requested):
I left just one extra question

runtime-modules/common/src/costs.rs
Lezek123: You're right, provided that we calculate this before executing the payment (not after as it's currently done).
Since it's much clearer I'll follow your approach.

runtime-modules/content/src/lib.rs
bedeho: Ok, I think I am wrong here, in the context of transactional rollbacks this is actually quite valuable in the real word, I think I just now am worried about the inconsistency, as we have not followed this rule so far, but better late than never to start this convention.

I guess my question is then: is there any way we can make it super obvious to the reader that this call should not fail, its qualitatively different to the calls before the mutation safe barrier? I thinks this is easy to miss? What about something like

```Rust
let paid_channel_bloat_bond_result = <thing that should not fail>;

// Cannot fail
debug_assert!(paid_channel_bloat_bond_result.is_ok());

let channel_account = paid_channel_bloat_bond_result?
```

It's a bit verbose, but feels like it really reminds the reader that this is different, is loud when somethign happens in debug mode, and enjoys the rollback last line of defence in production. Could probably be made very pretty with some wrapper macro which can be used to do this pattern over and over again, like `do_infallible_call!()`, so it is just a one liner.

mnaamani: I think this is a good suggestion. There are cases where we might also want to assert that the result details of a call are what we expect. Example with `balances::slash()` here https://github.com/Joystream/joystream/blob/carthage/runtime-modules/membership/src/lib.rs#L881

And to catch these debug assertions, we should probably change our CI checks to not use `--release` flag when running the cargo unit tests - https://github.com/Joystream/joystream/blob/carthage/scripts/cargo-checks.sh#L23

runtime-modules/content/src/lib.rs
```diff
@@ -765,15 +779,12 @@ decl_module! {
             Self::ensure_channel_bag_can_be_dropped(channel_id, num_objects_to_delete)?;
 
             // try to remove the channel
-            Self::try_to_perform_channel_deletion(sender.clone(), channel_id)?;
+            Self::try_to_perform_channel_deletion(&sender, channel_id, channel)?;
```
bedeho: Could we perhaps add some comment about bloat bond being permanently gone, not being returned to anyone, just to remind the reader?

runtime-modules/content/src/lib.rs
```diff
@@ -1117,9 +1130,6 @@ decl_module! {
             // == MUTATION SAFE ==
             //
 
-            //rewards the sender a state bloat bond amount for the work to delete the video.
```
bedeho, line 1120: Same here, can we make a reward about bloat bond not being returned.

runtime-modules/content/src/lib.rs
```diff
@@ -3299,49 +3331,23 @@ impl<T: Config> Module<T> {
...
-        );
-        Ok(())
-    }
-
     fn ensure_sufficient_balance_for_channel_transfer(
         owner: &ChannelOwner<T::MemberId, T::CuratorGroupId>,
         transfer_cost: BalanceOf<T>,
     ) -> DispatchResult {
-        let balance = match owner {
+        let funds_are_sufficient = match owner {
```
bedeho: `is_funding_sufficient`

runtime-modules/common/src/costs.rs
Lezek123: > Seems to me that we should just always allow it at the core level, but that the relevant account will always have the sufficient balance prefunded to block it when it should not happen.

There are 2 separate cases when I'm providing `false` as `allow_death`:
1. If the account is never expected to die anyway (that's the `moderate_post` example)
1. If the account could die if we didn't explicitly ensure that the balance is sufficient for it not to die (pay for NFT, buy membership etc.)

In the ideal world it's never actually required for `burn_from_usable` to check this, because we're either assuming the account cannot die (case `1.`) or we already checked the balance to ensure that it won't (case `2.`), so this is just another example of an "extra safety" check.

> curator group paying for channel from user.

When curator group is paying there is no room for allowing/disallowing account death, because the funds are just substracted from the ContentWorkingGroup budget and deposited into previous owner's account. Perhpas you meant member paying for curator group channel?

runtime-modules/common/src/costs.rs
Lezek123: It also doesn't touch any locked funds

Lezek123: It also fails in case of any issues (like account cannot be reaped due to remaining `consumers` etc.), while `slash` is infallable and may in some cases slash lower amount then requested (like in the remaining `consumers` case - it would then fallback to slashing as much as possible while leaving at least `ExistentialDeposit`)

runtime-modules/common/src/bloat_bond.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: Addressed in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: If you mean error types, there are no more `DispatchError`s here after https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: Removed `allow_death` parameter in https://github.com/Joystream/joystream/pull/4164/commits/fe0a624904917b1747c5ec5ae2c63cdfe5e3fc42

runtime-modules/common/src/costs.rs
Lezek123: You're right, it wouldn't really matter whether it's `Reasons:All` or `Reasons:Fee`, because `has_sufficient_balance_for_payment` check that's done before will prevent eating into any other lock than the invitation lock.
This was simplified on Ignazio's request however: https://github.com/Joystream/joystream/pull/4164/commits/f095684787b5b46b3c360a3d6cb34ea033a4a345

runtime-modules/council/src/lib.rs
```diff
@@ -899,17 +899,18 @@ decl_module! {
...
             //
             // == MUTATION SAFE ==
             //
 
-            Mutations::<T>::set_budget(budget.saturating_add(amount));
+            // Account is allowed to die when funding the council
+            burn_from_usable::<T>(&account_id, amount)?;
 
-            let _ = Balances::<T>::slash(&account_id, amount);
+            Mutations::<T>::set_budget(budget.saturating_add(amount));
```
bedeho: Perhaps there should be an `add_to_budget` which does the saturation internally, so user does not need to be careful and remember.

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/58e2b7070a851da1e234a3666a49ab7ccd48d230

runtime-modules/forum/src/lib.rs
```diff
@@ -1021,7 +1019,7 @@ decl_module! {
             // == MUTATION SAFE ==
             //
 
-            Self::slash_thread_account(thread_id, thread.cleanup_pay_off);
+            Self::slash_thread_account(thread_id, thread.cleanup_pay_off.amount)?;
```
bedeho, line 1022: Why is `::repay` not suitable here, like you do for videos and channels?

runtime-modules/forum/src/lib.rs
```diff
@@ -1224,7 +1218,7 @@ decl_module! {
             // == MUTATION SAFE ==
             //
 
-            Self::slash_thread_account(thread_id, post.cleanup_pay_off);
+            Self::slash_thread_account(thread_id, post.cleanup_pay_off.amount)?;
```
bedeho, line 1221: Same here?

runtime-modules/forum/src/lib.rs
bedeho: I guess because we don't want fees to go back when there is moderation?

runtime-modules/project-token/src/lib.rs
```diff
@@ -1661,39 +1666,42 @@ impl<T: Config> Module<T> {
...
             ensure!(
-                src_usable_balance >= T::JoyExistentialDeposit::get().saturating_add(total_amount),
+                has_sufficient_balance_for_payment::<T>(src, total_amount),
                 Error::<T>::InsufficientJoyBalance
             );
             for (dst, amount) in destinations {
                 if !amount.is_zero() {
                     ensure!(
-                        Joy::<T>::free_balance(*dst).saturating_add(*amount)
+                        Joy::<T>::total_balance(*dst).saturating_add(*amount)
```
bedeho: Why are we doing this check only here, when this sort of consideration is disregarded everywhere else?

runtime-modules/project-token/src/lib.rs
```diff
@@ -1711,23 +1719,34 @@ impl<T: Config> Module<T> {
     pub(crate) fn perform_initial_allocation(
         token_id: T::TokenId,
         targets: &BTreeMap<T::MemberId, TokenAllocationOf<T>>,
-        bloat_bond: JoyBalanceOf<T>,
-    ) {
+        repayable_bloat_bonds: Vec<RepayableBloatBondOf<T>>,
```
bedeho: Why is this vector separated from the `targets` , would it not be safer to just have this as something like 

`targets: &BTreeMap<T::MemberId, (TokenAllocationOf<T>, RepayableBloatBondOf<T>)>`

this would remove need for `DispatchError::Other("repayable_bloat_bonds: Index out of bounds"),`

runtime-modules/project-token/src/lib.rs
bedeho: Lets drop this since it is a one off artefact from the past.

runtime-modules/storage/src/lib.rs
```diff
@@ -4404,36 +4394,55 @@ impl<T: Config> Module<T> {
...
+    fn prepare_data_objects_to_insert(
+        objects: Vec<DataObjectOf<T>>,
+        bloat_bonds: Vec<RepayableBloatBondOf<T>>,
+    ) -> Result<Vec<DataObjectOf<T>>, DispatchError> {
+        let mut i = 0u64;
+        objects
+            .iter()
+            .map(|o| {
+                let state_bloat_bond = bloat_bonds
+                    .get(i as usize)
```
bedeho: Same issue here with the uncoordinate bloat bond and object lists,

runtime-modules/working-group/src/lib.rs
```diff
@@ -1192,17 +1193,18 @@ decl_module! {
...
             //
             // == MUTATION SAFE ==
             //
 
-            Self::set_working_group_budget(wg_budget.saturating_add(amount));
+            // An account is allowed to die when funding working group budget
+            burn_from_usable::<T>(&account_id, amount)?;
 
-            let _ = Balances::<T>::slash(&account_id, amount);
+            Self::set_working_group_budget(wg_budget.saturating_add(amount));
```
bedeho: Same point here about budget interaction.

bedeho (changes requested):
As a general observation I would say that things are getting more complicated than needed because the implementation is very defensive. I would much rather have our internal implementation code be very crisp and clean, with lots of extra dev-asserts to remind the reader of what should be true in all case, rather than having the reader second guessin what is actually going on, and thus making it harder to catch possible new or existing bugs.

Note: I have not reviewed storage+working groups, will have to submit a new one for that.

runtime-modules/content/src/lib.rs
Lezek123: Currently it's still returned, I planned to fix this as part of https://github.com/Joystream/joystream/issues/4208, but I can do it within the scope of this PR, since it's a small change

runtime-modules/forum/src/lib.rs
Lezek123: Yes

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/content/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/council/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/forum/src/lib.rs
Lezek123: Yes

runtime-modules/project-token/src/lib.rs
Lezek123: Removed in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/project-token/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/storage/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

runtime-modules/working-group/src/lib.rs
Lezek123: Fixed in https://github.com/Joystream/joystream/pull/4164/commits/deaf7f35d2f97f1068c51df5364962d755b043b2

bedeho (approved):
Excellent work, light speed turnaround, thank you 🥇

PR #3862, opened 2022-06-07
Qn map metadata for external resources for members

thesan:
Closes #2813

PR #4178, opened 2022-08-16
QN Prevent duplicated membership external resources entry

### LINKED ISSUE #2813: Metadata for external resources for members
# Background

Often it is useful to be able to identify how to connect directly with a channel or a member, outside of Joystream, for purposes of collaboration, ideation, business propositions, research, etc. Currently we do not have a metadata standard for this.

# Proposal

For both members and channels, add decorations to the metadata standard for each entity, to support adding one or more of the following

- Email
- Handle on one among a standardized, and easily extensible, set of services: the following seem sufficient for starters
  - Twitter 
  - Telegram
  - Discord
  - Facebook
  - YouTube
  - Matrix
  - IRC
  - Wechat
  - WhatsApp
- Raw link

Obviously only one value per category.

thesan:
This is the same branch which was merged on `master` in #3862 but with 3 additional commits to fix the duplicated entry problem.

Closes #2813
Unblock https://github.com/Joystream/pioneer/pull/3186 and https://github.com/Joystream/pioneer/pull/3356


query-node/schemas/membership.graphql
```diff
@@ -10,6 +10,26 @@ type AvatarUri @variant {
...
+  WECHAT
+  WHATSAPP
+}
+
+type MembershipExternalResource @entity {
+  type: MembershipExternalResourceType!
+  value: String!
+  memberMetadata: MemberMetadata!
+}
+
```
zeeshanakram3: This schema definition looks fine. However, I think if we had custom `ID` input of some unique definition such as concatenation of `memberId` & `resourceType`, then we won't have to deal with comprehensive checks to avoid duplicate records. With some unique ID space, Postgres would ensure no duplicate records (i.e. a member with multiple records of the same type).
e.g.
```graphql
type MembershipExternalResource @entity {
 "Unique combination of {memberId}-{resourceType}"
  id: ID!
  type: MembershipExternalResourceType!
  value: String!
  memberMetadata: MemberMetadata!
}
```

query-node/mappings/src/membership.ts
```diff
@@ -88,6 +92,77 @@ async function getOrCreateMembershipSnapshot({ store, event }: EventContext & St
...
+
+  for (const prevResource of memberMetadata.externalResources ?? []) {
+    await store.remove(prevResource)
+  }
+
+  const externalResources = (from ?? txExtResources?.flatMap(asMembershipExternalResource) ?? []).map(
+    ({ type, value }) => new MembershipExternalResource({ type, value, memberMetadata })
+  )
+  for (const resource of externalResources) {
+    await store.save<MembershipExternalResource>(new MembershipExternalResource(resource))
```
zeeshanakram3: Why is here constructor being called agian, since resources were just constructed at [line](https://github.com/Joystream/joystream/pull/4178/files#diff-9af242945476363410a798a87c9d22b687aeeea210b9b1bbc4d0697fde116b04R110)

zeeshanakram3 (commented):
Looks good overall.
It seems like graphql generated code in  `tests/network-tests/src/graphql/generated` is not update. Can you please check that by running `yarn generate:all` in `tests/network-tests` and committing any newly generated code? Maybe it's because of some upstream Carthage code you pulled in. Thanks

query-node/mappings/src/membership.ts
thesan: Right I forgot to remove this one!

query-node/schemas/membership.graphql
thesan: Good idea thanks. I implemented this but with the `MemberMetadata.Id` instead of `memberId` because for a given member multiple events refers to different `MemberMetadata` with potentially different external resources (as well as different values for other fields too)

thesan:
@zeeshanakram3 thanks for the review, I addressed the CR and regenerated the integration tests queries. It looks like these had not been updated in a while.

query-node/mappings/src/membership.ts
```diff
@@ -226,13 +302,9 @@ export async function members_MembershipGifted({ store, event }: EventContext &
     controllerAccount: member.controllerAccount,
     rootAccount: member.rootAccount,
     handle: member.handle,
-    metadata: new MemberMetadata({
-      ...member.metadata,
-      id: undefined,
-    }),
+    metadata: await saveMembershipMetadata(store, member),
```
zeeshanakram3, line 305: Now I see one problem that predates your PR, here in the event you are creating a metadata record using `await saveMembershipMetadata(store, member)` and then setting the event's metadata property to the record.

But we already have a member's metadata record created by the `createNewMemberFromParams` function at [line](https://github.com/Joystream/joystream/pull/4178/files#diff-9af242945476363410a798a87c9d22b687aeeea210b9b1bbc4d0697fde116b04L221). So for every member, we are creating two same metadata records in the table. What you can do is just use metadata property from the member object.

```suggestion
    metadata: member.metadata,
```

Also, this problem exists in other mappings too, such as `members_MembershipBought`, `members_MemberInvited`, can you please check those mappings too? Thanks

thesan: Yes I think this was on purpose. As you mentioned even before this PR the membership mappings were creating one metadata record for the `Membership` and on more for the `MemberXxxEvent`. This way when `MemberProfileUpdated` is dispatched: `Membership.metadata` is mutated but all the previous `MemberXxxEvent`s are not.

I'm not sure how important it is to keep an accurate history of the membership metadata events. But because I saw this as a common pattern in the mappings, I tried to follow it here too. Although if you think this is too costly I could reset the branch to [this commit](https://github.com/Joystream/joystream/pull/4178/commits/3c8f6e2044385a7177a16bbe9b53ce8466cbf4b6) (and link every `MembershipExternalResource` to a `Membership`) which significantly simplifies the code. I don't think it would actually affect Pioneer ATM (I'm not sure about Atlas). 

WDYT ?

zeeshanakram3: Thanks for the clarification, I didn't have the full context of why we had duplicate metadata records, your description sounds perfectly fine and I think we should have history of metadata records and try not to do any change at this stage. Everything else is fine, so approving the PR.

zeeshanakram3 (approved):

PR #4181, opened 2022-08-16
Carthage metaprotocol content categories

kdembler:
Forked off of Zeeshan's PR: 4122

- removed all notions of channel categories
- anyone can create video categories
- there is no video categories update/delete
- new videos can have unknown video categories
- updated CLI and tests to support above points


zeeshanakram3 (approved):
LGTM

Issue #4189, opened 2022-08-17
Staking/Nomination lock does not behave as rivalrous


Lezek123:
While doing final checks as part of https://github.com/Joystream/joystream/issues/4162 I noticed that it's possible to use the same account (and therefore the same funds) to stake for both some rivalrous purpose (for example: working group application/role) and validation/nomination at the same time.

I suppose this should not be possible, because:

* Validation/nomination lock is **not** part of `NON_RIVALROUS_LOCKS`
* It's mentioned as **_Rivalrous_** in the handbook: https://joystream.gitbook.io/testnet-workspace/system/accounts-and-staking#locks-1
* It makes sense for it to be rivalrous, as nomination/validation stake is subject to slashing (and so are other rivalrous stakes)

Despite that I have been able to setup an _Alice_ account with those 3 stakes simultaneously:

* staking account candidate stake (`boundsta`, _Non-Rivalrous_)
* forum working group role (lead) stake (`wg-forum`, **_Rivalrous!_**)
* nomination stake (`staking`, **_Rivalrous!_**)

![image](https://user-images.githubusercontent.com/19978260/185195851-b6cf7d6b-2ed4-4107-ae6a-1a8533a524e1.png)

Notice that **this is only possible if I create `wg-forum` stake before `staking` stake**, because otherwise the application would fail on `ensure!(T::StakingHandler::is_account_free_of_conflicting_stakes, /* ... */)`. There is no such check when staking as validator/nominator however, as this logic lives in `pallet_staking` which is outside of our control.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202815846267250) by [Unito](https://www.unito.io)

bedeho:
>  to stake for both some rivalrous purpose (for example: working group application/role) and validation/nomination at the same time.

I guess this first off begs the question: why did we not find out about this sooner? It certainly means we don't have QA for this? Can we add this @dmtrjsg ?

We also need to add integration tests to for this clearly, so adding this issue https://github.com/Joystream/joystream/issues/4190.

>  There is no such check when staking as validator/nominator however, as this logic lives in pallet_staking which is outside of our control.

Damn, this is _so bad_. I can't even compute how we did not think of that, nor how two audits missed this basic point 🤯 .  Seems there are two options
1. Fork `pallet_staking`, and only tweak the place it checks to use our lock type aware conflict chcker.
[...]

PR #4203, opened 2022-08-23
Make Validation/Nomination Staking Locks act as rivalrous (hack)

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4189 by implementing suggestion to "override" free_balance() for the runtime balances pallet instance used by the staking pallet.

* [x] Implementation
* [x] Tests
* [x] Staking pallet benchmarks failure work-arounds

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202852705059891) by [Unito](https://www.unito.io)


runtime/src/lib.rs
```diff
@@ -1581,3 +1582,146 @@ construct_runtime!(
...
+    fn free_balance(who: &AccountId) -> Self::Balance {
+        let existing_locks = Balances::locks(who);
+        let existing_lock_ids: Vec<LockIdentifier> =
+            existing_locks.iter().map(|lock| lock.id).collect();
+
+        if Runtime::are_locks_conflicting(&STAKING_LOCK_ID, existing_lock_ids.as_slice()) {
+            0
+        } else {
+            Balances::free_balance(who)
+        }
```
Lezek123, line 1634: Because `STAKING_LOCK_ID` is rivalrous this will prevent someone from increasing the bond, because after initially bonding there will be already a "conflicting" lock in the account when calling `bond_extra`.

To solve this you could:
- return `Balances::free_balance(who)` also in case there is already a `STAKING_LOCK_ID` lock assigned to the account (but not any other rivalrous lock)
- add a test to make sure this works as expected (`bond` + `bond_extra` scenario)

dobertRowneySr (approved):
lgtm

Lezek123 (changes requested):
I was going through alternative solutions that would actually make the `bond` call fail and therefore make us avoid any additional confusion among the users that try to bond with an already existing rivalrous lock. I found one that is pretty simple and looks promising to me.

There is one fallable call in `staking::bond()` method which seems like we can safely make fail for a given caller (stash) account:

```rust
frame_system::Pallet::<T>::inc_consumers(&stash).map_err(|_| Error::<T>::BadState)?;
```

As I discovered, there are 2 failing conditions for `Account::inc_consumers`:
- `a.providers == 0` - we probably don't want trigger this one
- `a.consumers >= T::MaxConsumers::max_consumers()` - this one seems safe and easy to trigger

So my suggestion is:
- When a new rivalrous lock is being set via `StakingHandler`, we increase account's `consumers` value by `T::MaxConsumers::max_consumers()` (this way we make sure the account never goes below `T::MaxConsumers::max_consumers()` while the rivalrous lock is in place and it also makes it trivial to go back to the "actual" value once the lock is removed)
- When a rivalrous lock is being removed, we decrease account's `consumers` by `T::MaxConsumers::max_consumers()`, which brings it back to the real value

And that's about it. By applying this fix we prevent `staking::bond()` from ever succeeding in context of an account that has a rivalrous lock. This seems to me like a much simpler and safer approach than the `BalancesProxy` and I couldn't find any bad side-effects that this approach would have.

Let me know what you think.

bedeho:
> I couldn't find any bad side-effects that this approach would have

This is the main thing here, I don't really understand this fix, but certainly we are here making a change to very public state which lots of code may be using, or may use in the future in some random fix. The proxy solution did exactly one thing, locally: make staking pallet think there was 0 funds if there was a rivalrous lock. Noone else is affected by that. I'm not even 100% sure we fully grasp the consequences of that, say for example if the account can be used it in some other way with the staking pallet where the balance > 0 is not a requirement, and then at a later time the balance has gone up in a way the staking pallet expects to know (say if it depends on some on_balance_change hook or something). But for this fix I just have no way to gauge what level of confidence we should ascribe to your inability to find a sideeffect right now.

runtime/src/lib.rs
```diff
@@ -1581,3 +1582,135 @@ construct_runtime!(
         DistributionWorkingGroup: working_group::<Instance9>::{Pallet, Call, Storage, Event<T>},
     }
 );
+
+// Proxy into the Balances pallet, used to modify behavior of free_balance()
+// MUST ONLY BE USED BY THE STAKING PALLET
```
bedeho, line 1587: A longer explanation of _why_ this is being added makes sense, also,nice if this all could be in a different file, this file is getting out of hand :D.

runtime/src/lib.rs
```diff
@@ -1581,3 +1582,146 @@ construct_runtime!(
...
```
bedeho, line 1587: A longer explanation of why this is being added makes sense, also,nice if this all could be in a different file, this file is getting out of hand :D.

mnaamani:
Replaced by https://github.com/Joystream/joystream/pull/4213

PR #4212, opened 2022-08-25
Carthage fix non editable post bug

### LINKED ISSUE #4175: Forum: Adding non-editable post can prevent a thread from ever being removed
1. Adding a post increases `thread.number_of_posts` even if the post being added is non-editable (is not preserved in the state): https://github.com/Joystream/joystream/blob/carthage/runtime-modules/forum/src/lib.rs#L1406
1. Removing/moderating a post is only possible if it exists in the state (https://github.com/Joystream/joystream/blob/carthage/runtime-modules/forum/src/lib.rs#L1491, https://github.com/Joystream/joystream/blob/carthage/runtime-modules/forum/src/lib.rs#L1477)
1. Removing/moderating a thread is only possible if `thread.number_of_posts == 0` (https://github.com/Joystream/joystream/blob/carthage/runtime-modules/forum/src/lib.rs#L1633, https://github.com/Joystream/joystream/blob/carthage/runtime-modules/forum/src/lib.rs#L1655)

This implies:

* If a post is created in non-editbale initial state in a thread, then that thread cannot be removed

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202801705812740) by [Unito](https://www.unito.io)

dobertRowneySr:
fixes #4175 

Adding a non editable post won't increase the `num_post` counter in a thread

#### Changes
- added test for bug
- fix


runtime-modules/forum/src/lib.rs
```diff
@@ -1400,12 +1400,12 @@ impl<T: Config> Module<T> {
...
             <PostById<T>>::insert(thread_id, new_post_id, new_post);
-        }
-
-        let mut thread = <ThreadById<T>>::get(category_id, thread_id);
-        thread.number_of_posts = thread.number_of_posts.saturating_add(1);
 
-        <ThreadById<T>>::mutate(category_id, thread_id, |value| *value = thread);
+            <ThreadById<T>>::mutate(category_id, thread_id, |thread| {
+                // non editable post should leave the counter untouched
+                thread.number_of_posts = thread.number_of_posts.saturating_add(1);
```
Lezek123: The filed should be renamed as suggested in https://github.com/Joystream/joystream/issues/4175#issuecomment-1215436306. Perhaps `number_of_mutable_posts`?

runtime-modules/forum/src/tests.rs
```diff
@@ -2588,3 +2588,50 @@ fn storage_limit_checks() {
         }
     });
 }
+
+#[test]
+fn add_non_editable_post_should_not_increase_thread_number_of_posts() {
+    pub const INITIAL_BALANCE: u64 = 10_000_000;
```
Lezek123: Please make it `ed() + thread_deposit + post_deposit`, as those parameters may change

Lezek123 (changes requested):
https://github.com/Joystream/joystream/runs/8019618731?check_suite_focus=true is failing on `cargo-fmt` check

runtime-modules/forum/src/lib.rs
dobertRowneySr: renamed it to `number_of_editable_posts`

Lezek123 (changes requested):
Looks ok, but I think you missed my previous comment about `cargo fmt`.
CI is still failing: https://github.com/Joystream/joystream/runs/8135755862?check_suite_focus=true

Lezek123 (approved):

PR #4213, opened 2022-08-25
SignedExtension to Control Access to Staking::Bond() call

mnaamani:
Alternative fix for https://github.com/Joystream/joystream/issues/4189 described in https://github.com/Joystream/joystream/issues/4189#issuecomment-1227824357

* [x] Implementation
* [x] Integration test for this fix/hack
* [x] Update to 'run-network-tests' github workflow run multiple scenarios with matrix strategy
* [x] Set initial min validator/nominator bond to be non-zero

Important note: Since we have not adding any implementation client side for the signed extension, polkadot api is still able to handle it since the extension doesn't add anything extra to the transaction payload. But we do see this warning when connecting:

```
REGISTRY: Unknown signed extensions CheckCallAllowed found, treating them as no-effect
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202872818383930) by [Unito](https://www.unito.io)


runtime/src/signed_extensions/check_call_allowed.rs
```diff
@@ -0,0 +1,148 @@
...
+    // Checks to see if the account has any rivalrous locks that would conflict
+    // with the FRAME staking pallet lock
+    fn has_no_conflicting_locks(who: &AccountId) -> bool {
+        let existing_locks = Balances::locks(who);
+
+        if existing_locks.len() == 0 {
+            return true;
+        }
+
+        if existing_locks.len() == 1 && existing_locks[0].id == STAKING_LOCK_ID {
```
Lezek123, line 56: The `bond` should always fail if there is already an existing bond, so I think this check is not really needed.

runtime/src/signed_extensions/check_call_allowed.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    // Recursively checks calls until one invalid bonding call is detected and returns false,
+    // otherwise returns true.
+    fn has_no_invalid_bonding_calls(who: &AccountId, calls: Vec<Call>) -> bool {
+        let mut all_calls_valid = true;
+
+        for call in calls.into_iter() {
+            all_calls_valid = all_calls_valid
+                && match call {
```
Lezek123: This looks overcomplicated. This method returns as soon as it finds an invalid call, so `all_calls_valid` should always be `true` here. It also appears to me we don't need a mutable `all_calls_valid` variable at all. We can just:
- check if the iteration contains invalid call
- if it does - return false
- if the loop finished - return true (as this means there were no invalid calls

runtime/src/signed_extensions/check_call_allowed.rs
```diff
@@ -0,0 +1,148 @@
...
+        _len: usize,
+    ) -> TransactionValidity {
+        if Self::has_no_invalid_bonding_calls(who, vec![call.clone()]) {
+            Ok(ValidTransaction {
+                priority: 0,
+                requires: vec![],
+                provides: vec![],
+                longevity: TransactionLongevity::max_value(),
+                propagate: true,
+            })
```
Lezek123: It seems to be the same as:
```suggestion
            Ok(Default::default())
```
(which is returned by `frame_system::CheckWeight` for example)

runtime/src/tests/locks.rs
```diff
@@ -64,3 +64,48 @@ fn incompatible_stakes_check_passed_successfully() {
         );
     });
 }
+
+#[test]
+fn incompatible_stakes_for_staking_pallet() {
```
Lezek123, line 71: Would it be possible to test the SignedExtension behavior in this test? (ie. that it actually blocks the transaction?)

Lezek123 (changes requested):
**Additional note:** I think that since we're adding this extension we can no longer rely on `BaseExtrinsicWeight` from  `frame_support::weights`. We need our own benchmarks that will take into account the new signed extension.

runtime/src/signed_extensions/check_call_allowed.rs
```diff
@@ -0,0 +1,148 @@
...
+                        Self::has_no_conflicting_locks(who)
+                    }
+
+                    // should we prevent Sudo from bypassing these checks?
+                    // Call::Sudo(pallet_sudo::Call::<Runtime>::sudo { call }) => ...
+                    // Call::Sudo(pallet_sudo::Call::<Runtime>::sudo_as { who, call }) => ...
+                    // Call::Sudo(pallet_sudo::Call::<Runtime>::sudo_unchecked_weight { call }) => ...
+                    // Call::Utility(substrate_utility::Call::<Runtime>::batch_all { calls }) => ...
+                    // Call::Utility(substrate_utility::Call::<Runtime>::force_batch { calls }) => ...
+                    // Call::Utility(substrate_utility::Call::<Runtime>::dispatch_as { calls }) => ...
```
Lezek123: The list should also include `multisig.asMulti` and `multisig,asMultiThreshold1`

runtime/src/signed_extensions/check_call_allowed.rs
mnaamani: You are correct. The only reason I added this and the simpler prior check is to make this function as fast as possible. So to avoid need to always call `Runtime::are_locks_conflicting`. But it is also to not consider it an invalid transaction when account is already bonded, so user will get a proper Dispatch Error detailing reason why bond() call fails.

runtime/src/tests/locks.rs
mnaamani: It should be possible to write a test that instantiates the extension, and we invoke pre_dispatch() and validate(), but for a more realistic test we would have to maybe do it in the node package, instantiate a transaction pool and test by injecting extrinsics into it. It might make sense to add, but I went for integration test instead. WDYT

runtime/src/signed_extensions/check_call_allowed.rs
mnaamani: Indeed

runtime/src/signed_extensions/check_call_allowed.rs
mnaamani: will add

mnaamani:
> **Additional note:** I think that since we're adding this extension we can no longer rely on `BaseExtrinsicWeight` from `frame_support::weights`. We need our own benchmarks that will take into account the new signed extension.

You are correct. Added issue https://github.com/Joystream/joystream/issues/4233

mnaamani:
I completed the checks for remaining call paths, although we have decided that this approach has become too complex.
We will drop this solution and go for modifying the staking pallet directly as the safest option.

PR #4235, opened 2022-08-30
No system.ExtrinsicSuccess event processing

mnaamani:
Since bounty module and bounty mappings are not enabled, scheduler is not needed, which was running after processing each block. Disabled postBlockHooks and since we don't need to run scheduled on every block, disabled processing system.ExtrinsicSuccess event.
This may help improve processing speed.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202893971416614) by [Unito](https://www.unito.io)


Lezek123 (approved):
LGTM, I've run some benchmarks and the results were following:
- Indexing speed still seems to be the main contributor to sync time, at least when I'm using a remote rpc. It took me **~10 minutes to sync ~5000 blocks.**
- Using a remote indexer, I can sync the processor w/ the current staging network (~10500 blocks) in a matter of **a few seconds** on this branch (no scheduler)
- Using a remote indexer it takes **about a minute** to sync the processor w/ the current staging network (~10500 blocks) using the `carthage` branch

This means that the speed of procesing itself has greatly improved (~20x or more), however when indexing from scratch there's possibly no improvement (this would have to be confirmed using a local rpc node however)

zeeshanakram3 (approved):

mnaamani:
> This means that the speed of procesing itself has greatly improved (~20x or more), 

Great that we got back the fast processing. This was the main bottleneck when re-processing.
Now the question is, even without the postBlockHook/scheduler, on mainnet, if we assume there will be at least one tx per block, will we loose this performance gain? In other words is there something specific about the postBlockHook that is slowing down the processor or was there some slow code/memory leak in the scheduler itself?

> however when indexing from scratch there's possibly no improvement (this would have to be confirmed using a local rpc node however)

This is to be expected, since the processing speed is bound by how fast the indexer can fetch new blocks.

PR #4248, opened 2022-09-02
CLI fixes: selecting distribution buckets & subtitile assets indices

Lezek123:
Fixes for Carthage CLI:

* Fixed handling for distribution buckets selection when some families are not included in the policy
* Fixed indices of subtitle assets in case other assets (like video thumbnail, video media or previous subtitles' assets) are not provided

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202914194106559) by [Unito](https://www.unito.io)

### DIFF
--- a/cli/src/commands/content/createVideo.ts
+++ b/cli/src/commands/content/createVideo.ts
@@ -77,18 +77,20 @@ export default class CreateVideoCommand extends UploadCommandBase {
     meta.thumbnailPhoto = videoAssetIndices.thumbnailPhotoPath
 
     // Subtitle assets
-    const resolvedSubtitleAssets = await Promise.all(
-      (subtitles || []).map(async (subtitleInputParameters, i) => {
-        const { subtitleAssetPath } = subtitleInputParameters
-        const [[resolvedAsset], assetIndices] = await this.resolveAndValidateAssets({ subtitleAssetPath }, input)
-        // Set assets indices in the metadata
-        if (meta.subtitles) {
-          meta.subtitles[i].newAsset =
-            assetIndices.subtitleAssetPath || 0 + Object.entries(videoAssetIndices).length + i
-        }
-        return resolvedAsset
-      })
-    )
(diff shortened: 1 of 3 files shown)


cli/src/commands/content/createVideo.ts
```diff
@@ -77,18 +77,20 @@ export default class CreateVideoCommand extends UploadCommandBase {
...
+          const { subtitleAssetPath } = subtitleInputParameters
+          const [[resolvedAsset]] = await this.resolveAndValidateAssets({ subtitleAssetPath }, input)
+          // Set assets indices in the metadata
+          if (meta.subtitles && resolvedAsset) {
+            meta.subtitles[i].newAsset = subtitleAssetIndex++
+          }
+          return resolvedAsset
+        })
+      )
+    ).filter((r) => r)
```
zeeshanakram3, line 93: Why is this `filter` applied here, seems like a redundant operation since the output list would be the same as the input? Also, the same expressions exist in `updateVideo.ts` too

Lezek123: Some subtitles may not have assets provided, for example an input can be like:
```
  "subtitles": [
    { "type": "subtitle", "language": "en", "mimeType": "SRT", "subtitleAssetPath": "./subtitle-en-1.srt" },
    { "type": "subtitle", "language": "pl", "mimeType": "SRT" },
    { "type": "subtitle", "language": "fr", "mimeType": "SRT", "subtitleAssetPath": "./subtitle-fr-1.srt" }
  ]
```
In this case you'll have 3 instances of subtitle metadata, but only 2 assets (so input will be of length 3, but output will be of length 2)

Lezek123: But perhaps it would be better to disallow providing subtitles without assets

zeeshanakram3: Oh, Ok, thanks for the clarification.

zeeshanakram3 (approved):

Issue #4197, opened 2022-08-22
Disable CRT


bedeho:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202839042442243) by [Unito](https://www.unito.io)

PR #4249, opened 2022-09-02
Disable creator token issuance

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4197

Blocking `content::issue_creator_token` is sufficient because:
- All `pallet_token` extrinsics are protected by `ensure_token_exists`
- All other CRT-related extrinsics in content pallet are protected by `channel.ensure_creator_token_issued`


dobertRowneySr (approved):
LGTM, just need to be updated

Issue #4107, opened 2022-08-01
Colossus: Add a command to update data object state bloat bond


Lezek123:
SSIA

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202701158426596) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4250

PR #4250, opened 2022-09-02
Colossus: Add a command to update data object bloat bond

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4107


mnaamani (approved):
Tested and works as expected

PR #4251, opened 2022-09-04
joy denominated initial balances

mnaamani:
The initial balances and vesting config we are producing for carthage and mainnet will have balances denominated in JOY, not the base unit of `Balance` type, so we are making this adjustment to the initial-balances deserialization code.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202926705682328) by [Unito](https://www.unito.io)


dobertRowneySr (approved):
LGTM, just one small question
Are we calling the base unit nJOY, correct?

mnaamani:
Thanks for reviewing, we may not end up using this and opt to update the scripts producing the initial balances file instead. So no merge now, I'll convert to draft.

Issue #4124, opened 2022-08-05
Weight Functions lack storage metadata


mnaamani:
### Problem
Weight functions generated are missing the names of storage values being written to and read from. Eg.

```rust
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bc445c60ee96be8ba74dcbcf1be5859b72] (r:1 w:0)
	// Storage: unknown [0x26aa394eea5630e07c48ae0c9558cef7b99d880ec681799c0cf30e8886371da9] (r:1 w:1)
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bc6b8a2f06065bb1b4d3395eb9f6a6ce60] (r:1 w:1)
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bc13020dc69e85870ac7b4c755bb8753c2] (r:1 w:1)
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bc0d3094b474c99662ab2c5f2e2f8c27b6] (r:1 w:0)
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bc4560c4f4db6b0f2ccd9ba01ebd9f4757] (r:1 w:1)
	// Storage: unknown [0x2ce461329fdf4be12bce01afc0af09bccaf1c595cf48e2f63ea3f59e0a8ea0f4] (r:1 w:0)
	fn buy_membership_with_referrer(i: u32, _j: u32, ) -> Weight {
		(45_799_000 as Weight)
			// Standard Error: 0
			.saturating_add((1_000 as Weight).saturating_mul(i as Weight))
			.saturating_add(T::DbWeight::get().reads(7 as Weight))
			.saturating_add(T::DbWeight::get().writes(4 as Weight))
	}
```

To get the name of the storage key, instead of `unknown`/`raw key value` we need to return the metadata in
`fn benchmark_metadata()` here:

https://github.com/Joystream/joystream/blob/9c90d21fcfbcd223c76973404f6f65086fb26735/runtime/src/runtime_api.rs#L340

```rust
.
.
use frame_support::traits::StorageInfoTrait;
let storage_info = AllPalletsWithSystem::storage_info();
(list, storage_info)
```

However we get a compiler error if we try
> "function or associated item cannot be called on ... due to unsatisfied trait bounds":

[...]

Issue #4041, opened 2022-07-15
No unbound collections in stored objects


Lezek123:
1. Make sure we don't store any collections that have an unbound number of items/entries (either directly or as part of a struct / other stored type)
1. Make sure we always use worst-case-scenario sized objects for all extrinsics that load/modify those objects (in the runtime state) in the benchmarks 

The problem with storing unbound collections is that the size of the object stored by the runtime (for example `Channel`, `Video`, `CuratorGroup` etc.) has a significant impact on all operations performed on this object.

For example: the cost of `set_curator_group_status` for an empty curator group (no `curators` and no `perrmisions_by_level`) is ~4 times lower (not counting the cost of `read` and `write` operations) than the cost of updating a wrost-case-scenario sized curator group (`curators.len() == T::MaxNumberOfCuratorsPerGroup`, `permissions_by_level.len() == T::MaxKeysPerCuratorGroupPermissionsByLevelMap`)

This means that when benchmarking extrinsics that load/modify given object in the runtime state we should always use an object with a worst-case-scenario possible size for a given type. When an object contains an unbound collection this is either not possible or very hard to estimate.

Some examples of unbound collections we're currently storing:

* `Channel.collaborators`
* `Channel.data_objects`
* `Video.data_objects`

Some examples of benchmarking where we don't use worst-case-scenario size when the extrinsic is operating on an object that contains a collection:

* `Storage::accept_pending_data_objects` - the bag is being loaded from the runtime state, but the `bag.distributed_by` collection is not pre-populated with the worst-case-scenario number of items in the benchmarks
* ...

Those are just a few examples I've been able to catch by quickly glancing at the code, there is probably a lot more of those.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202613285606107) by [Unito](https://www.unito.io)

Lezek123:
https://github.com/Joystream/joystream/issues/4046#issuecomment-1199179759

Issue #4268, opened 2022-09-07
The comedy of starting referenda


bedeho:
In trying to configure `referendum::Config::MaxWinnerTargetCount`, I was warned by @mnaamani  that simply setting it to the council size would result in a problem if `<Runtime as council::Config>::MinNumberOfExtraCandidates > 0`. I could not reproduce this, but it did lead me to the following questions

1. Why do we have two functions for starting a referendum, where one is not used

```Rust
/// referendum/src/lib.rs

    // TODO: Disabled during the 'Olympia - Sumer' merge. Refactor on the next pallet change.
    #[allow(clippy::result_unit_err)]
    /// Start a new referendum.
    fn start_referendum(
        origin: Origin,
        extra_winning_target_count: u64,
        cycle_id: u64,
    ) -> Result<(), ()>;

    /// Start referendum independent of the current state.
    /// If an election is running before calling this function, it will be discontinued without
    /// any winners selected.
    /// If it is called with a bigger winning target count greated than the max allowed the max
    /// will be used
    fn force_start(extra_winning_target_count: u64, cycle_id: u64);
```

Former is not used, and should really be removed.

2. Why do we have _forceful_ referendum starts. Seems like this shuold never happen unless referendum & council pallets are incompatibly configured.
3. Why does council actually ask for one less winner than required!

```Rust
    // Change the council stage from the announcing to the election stage.
    fn finalize_announcing_period(candidates_count: u64) {
        let extra_winning_target_count = T::CouncilSize::get() - 1;

        // start referendum
        T::Referendum::force_start(extra_winning_target_count, AnnouncementPeriodNr::get());
```

4. and then, comically, refendum pallet adds an extra winner!

```Rust
    // Start referendum independent of the current state.
    // If an election is running before calling this function, it will be discontinued without any
    // winners selected.
    fn force_start(extra_winning_target_count: u64, cycle_id: u64) {
        let winning_target_count = extra_winning_target_count + 1;

        // If a greater than the max allowed target count is used the max will be used in its place
        let winning_target_count = winning_target_count.min(T::MaxWinnerTargetCount::get());

```

So this clearly needs an examination, and verifying whether there really ever is an issue like mnaamani was concerned about.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202940672873861) by [Unito](https://www.unito.io)

Lezek123:
[...]

PR #4276, opened 2022-09-09
Weight functions lack storage metadata: Implement `generate_storage_info`

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4124 and the first part of https://github.com/Joystream/joystream/issues/4041


runtime/src/lib.rs
```diff
@@ -1006,29 +1008,30 @@ impl common::StorageOwnership for Runtime {
...
 }
 
 // Staging/testing storage parameters
 #[cfg(any(feature = "staging_runtime", feature = "testing_runtime"))]
 parameter_types! {
-    pub const StorageBucketsPerBagValueConstraint: storage::StorageBucketsPerBagValueConstraint =
-        storage::StorageBucketsPerBagValueConstraint {min: 1, max_min_diff: 15};
-    pub const DefaultMemberDynamicBagNumberOfStorageBuckets: u64 = 1;
-    pub const DefaultChannelDynamicBagNumberOfStorageBuckets: u64 = 1;
+    pub const MinStorageBucketsPerBag: u32 = 1;
```
mnaamani, line 1031: Does it make sense for `MinStorageBucketsPerBag` to ever be 0? If not should we have an assertion that it is always gt 0?

runtime-modules/council/src/lib.rs
```diff
@@ -345,6 +346,20 @@ decl_storage! {
         /// Councilor reward per block
         pub CouncilorReward get(fn councilor_reward) config(): Balance<T>;
     }
+
+    add_extra_genesis {
+        config(council_members): Vec<CouncilMemberOf<T>>;
+
+        build(|s| {
```
mnaamani: We have always had this config, but I wonder is it safe to have initial set of council members without anything at stake? Also since we no longer set initial members at genesis, does it really make sense to have initial council members?

runtime-modules/council/src/lib.rs
```diff
@@ -314,13 +315,13 @@ pub trait ReferendumConnection<T: Config> {
...
 
-decl_storage! {
+decl_storage! { generate_storage_info
     trait Store for Module<T: Config> as Council {
         /// Current council voting stage
         pub Stage get(fn stage) config(): CouncilStageUpdate<T::BlockNumber>;
 
         /// Current council members
-        pub CouncilMembers get(fn council_members) config(): Vec<CouncilMemberOf<T>>;
+        pub CouncilMembers get(fn council_members): WeakBoundedVec<CouncilMemberOf<T>, T::CouncilSize>;
```
mnaamani, line 324: Since we are updating this, does it make sense to use a `BoundedBTreeSet` instead of a `WeakBoundedVec` since members should be unique.

runtime-modules/forum/src/lib.rs
```diff
@@ -126,6 +138,9 @@ pub trait Config:
     /// Maximum depth for nested categories
     type MaxCategoryDepth: Get<u64>;
 
+    /// Maximum number of category's stickied threads
+    type MaxStickiedThreads: Get<u32>;
+
```
mnaamani, line 143: Should `type MaxCategoryDepth: Get<u64>;` also be changed to `u32` ?

mnaamani (commented):
Wow this refactor job turned out to be bigger than expected, but certainly worth it.
I just had a few minor points otherwise its good to go.

runtime-modules/council/src/lib.rs
Lezek123: Unfortunately there is no `WeakBoundedBTreeSet`. The advantage of `WeakBoundedVec` is that it has an unfallable `froce_from` method, which can be used during `on_initialize` (more specifically in `elect_new_council` method). `BoundedBTreeSet` only has a fallable `try_from`, which returns `Result`, but since we cannot fail in `on_initialize`, we'd have to resort to something like `debug_assert` + `unwrap_or` perhaps...

runtime-modules/forum/src/lib.rs
Lezek123: There are other limits also, ie.:
```rust
    pub const MaxCategoryDepth: u64 = 6;
    pub const MaxSubcategories: u64 = 40;
    pub const MaxThreadsInCategory: u64 = 20;
```

But since they are not used to limit the size of a bounded collection I haven't changed their type to limit the amount of changes in this PR.

runtime-modules/council/src/lib.rs
Lezek123: Removed in https://github.com/Joystream/joystream/pull/4276/commits/0367c3ec05cf294025d1f8fc5da7ebdbc67b360f

runtime/src/lib.rs
Lezek123: Added the assertions in https://github.com/Joystream/joystream/pull/4276/commits/00ea10ea249a4bb3323b4bfa094de2d9512b4211

runtime-modules/council/src/lib.rs
mnaamani: I see. Can we get rid of use of `try_from`? I assume we are we using it to convert from the Vec of results from the referendum pallet? If that means more refactoring in referendum pallet, I think we can live with this.

mnaamani: Should we make an issue for it to tackle this in future?

mnaamani (approved):

runtime-modules/council/src/lib.rs
Lezek123: > I see. Can we get rid of use of try_from? I assume we are we using it to convert from the Vec of results from the referendum pallet? If that means more refactoring in referendum pallet, I think we can live with this.

I think that would mean referendum's `IntermediateWinners` and council's `CouncilMembers` would need to be the same type (ie. have the same bound), so referendum's `MaxWinnerTargetCount` and council's `CouncilSize` would need to be unified. I'll have that in mind when investigating https://github.com/Joystream/joystream/issues/4268.

Issue #4272, opened 2022-09-08
Colossus: check worker ID on startup


kdembler:
### Problem
With the current Colossus, it's possible to provide wrong worker ID (not associated with you on-chain) and start the node. AFAIK, there's no messaging letting you know that you are using the wrong worker ID. The only indication would be uploads failing if your misconfigured worker ID is not assigned to correct storage buckets. We have experienced this already after one storage worker has changed to lead position and their worker ID has changed

### Solution
Add a startup check to Collosus that verifies that the key you provided is associated with the worker ID you have provided. If that's not the case, the node should exit with an error, making it clear that something is wrong

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202945940410669) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4283

PR #4283, opened 2022-09-12
Colossus: Verify workerId to transactorKey compatibility

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4272

Note that the check is only failing if:
- Worker id is invalid
- There is an active bucket assigned to the provided worker, but its transactor key doesn't match the key configured in colossus

There is however a case where there are no active buckets assigned to the worker, in which case the `server` command will only emit a warning, but will not exit. This is because in that case:
- We cannot verify transactor key, since it's not yet set
- We still want to allow the storage worker to start syncing before he accepts a bucket invitation, so that (s)he can already be fully synced when starting to act as bucket operator.

### DIFF
--- a/storage-node/src/commands/server.ts
+++ b/storage-node/src/commands/server.ts
@@ -12,6 +12,9 @@ import path from 'path'
 import { promisify } from 'util'
 import ExitCodes from './../command-base/ExitCodes'
 import fs from 'fs'
+import { getStorageBucketIdsByWorkerId } from '../services/sync/storageObligations'
+import { PalletStorageStorageBucketRecord } from '@polkadot/types/lookup'
+import { Option } from '@polkadot/types-codec'
 const fsPromises = fs.promises
 
 /**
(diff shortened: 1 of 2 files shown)


kdembler (approved):
Looks good to me

storage-node/src/commands/server.ts
```diff
@@ -152,10 +155,13 @@ Supported values: warn, error, debug, info. Default:debug`,
     }
 
     const storageProviderAccount = this.getAccount(flags)
+    const workerId = flags.worker
+    if (!(await verifyWorkerId(api, flags.queryNodeEndpoint, workerId, storageProviderAccount.address))) {
+      this.exit(ExitCodes.InvalidWorkerId)
```
mnaamani, line 160: It might be more convenient for the storage node operator if instead of exiting, we can loop trying to recheck valid keys and worker id. It allows the node operator to start their node before being assigned buckets. And as soon as they are assigned buckets and accept invite the server api will come up.

mnaamani (approved):
Works as expected. Left a suggestion for alternate behavior that could be more practical for operator.

storage-node/src/commands/server.ts
Lezek123: It is possible now to start the node without any buckets assigned, this will only emit a warning: https://github.com/Joystream/joystream/pull/4283/files#diff-c2831ec0c824e9dbc4d17c21757e3ec83fc0a97495066219cb69f350414dbb9eR288

Issue #4281, opened 2022-09-12
`ChannelVisibilitySetByModerator` event is not indexed by QN


kdembler:
`ChannelVisibilitySetByModerator` event can be emitted by calling `content.setChannelVisibilityAsModerator` extrinsic as seen here: https://github.com/Joystream/joystream/blob/carthage/runtime-modules/content/src/lib.rs#L815

This event is not included in the QN manifest and therefore QN state is not updated after the extrinsic is called.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202962265154876) by [Unito](https://www.unito.io)

PR #4284, opened 2022-09-12
add mapping for ChannelVisibilitySetByModerator event

zeeshanakram3:
implements #4281


query-node/mappings/src/content/channel.ts
```diff
@@ -44,6 +45,7 @@ import {
 import { BTreeMap, BTreeSet, u64 } from '@polkadot/types'
 // Joystream types
 import { PalletContentChannelActionPermission } from '@polkadot/types/lookup'
+import { getAllManagers } from '../derivedPropertiesManager/applications'
```
kdembler: Seems this import isn't used?

kdembler (approved):

Issue #3802, opened 2022-05-26
Launch with PoA validator set which can be switchted to NoPoS


bedeho:
# Background
As part of `Carthage` Jsgensis will operate `PoA` validators as part of the first stage of roleout, and then transition to `NoPoS`, as described here https://github.com/Joystream/joystream/issues/3792.

# Proposal
Update runtime so as to allow it to start in PoA mode, and then through a `Root` origin extrinsic transition to `NoPoS`. Initial validator set is set in genesis config.

## Resources
* https://www.gautamdhameja.com/resilient-poa-network-substrate/
* https://github.com/gautamdhameja/substrate-poa
* https://docs.substrate.io/tutorials/v3/private-network/

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202346732309222) by [Unito](https://www.unito.io)

mnaamani:
Additional Resource: https://substrate.stackexchange.com/questions/996/how-to-upgrade-consensus-engine-of-a-live-chain-from-poa-to-pos/1038#1038

dobertRowneySr:
@mnaamani 
I am not 100% sure about the role of these pallets in our runtime. Can you please provide a small description about those market with a "?".
I have used a reference substrate node template https://github.com/substrate-developer-hub/substrate-node-template
## Frame pallets are used in JS runtime and not in the node template
- Aura, tied to block validation by authority
## Frame pallets are used in JS runtime and not in the node template
- Utility substrate_utility, various and unrelated to block production/finalization
- ? BabeNPoS block production
- ? Authorship
- ? ElectionProviderMultiPhase, is this used in council election
- ? Staking: handles validator staking
- Session: handle a period of time that has a constant set of validators, it particularly takes care of maintaining the validator set
- ? Historical
- ? AuthorityDiscovery: retrieve the current set of authorities,
- ImOnline: keeps track of the active validators,
-  Offences: Tracks reported offences (i.e. slash stacked amount from validators)
- ? BagsList
- Vesting, related to the vesting of funds
- Multisig, multisig

My plan is to 
- [ ] remove all of those related to PoS from `runtime/src/lib.rs` and keep Aura only for the PoA version
- [ ] adjust chainspec if necessary so that Aura can have its validator set at genesis
- [ ] figure out a way to transfer validator set during migration 
- [ ] for the Npos I guess we should use the current runtime pallets configuration

dobertRowneySr:
So far I know that having: pallets
[...]

Issue #4286, opened 2022-09-12
Launching with restricted NPoS


mnaamani:
## Background
As an alternative option to starting chain with `PoA` configuration which would require a lot of code changes, we can start the chain with `NPoS` but with a very constrained configuration.

### Objectives
1. Start chain with an initial set of validators, without ability for any participants to join this set.
1. The initial set of validators must receive zero rewards. (nominators too)

### Validators and nominators config.
Can be achieved in two ways:

A. By ensuring the runtime configuration has the `BaseCallFilter` filter enabled on the system pallet that prevents any dispatches on the staking pallet. This will prevent `bond()` and `validate()` and `nominate()` calls which is how validators join the "waiting list" of validators to be selected in future sessions, and how participants become nominators.

B. By setting the minimum validator and nominator bond to an impractically high value, `Balance::max_value()`, either after launch by making a sudo call to `staking::set_staking_configs()`, or by modifying the [genesis_build](https://github.com/Joystream/substrate/blob/31f4f7654cf8e27e0d73ffeb1c11b2df2d4fbffe/frame/staking/src/pallet/mod.rs#L557) step in the staking pallet to allow bonding first of the initial validators (with minimal stake), and as last step to increase the min bond to max value.

I would opt for A since it is simpler to implement (already implemented in-fact) and doesn't require making further changes to the staking pallet.

### Zero Rewards config.
* The only source of rewards for validators currently in our runtime configuration are the era payouts and the transaction tips.
We can simply configure the staking `EraPayout` config to an empty tuple `()` [in place of the inflation curve](https://github.com/Joystream/joystream/blob/97b311f071f6f76b5de78da281d88188466c63ce/runtime/src/lib.rs#L568) which would return zero balance as rewards.
* Modify the transaction payment pallet configuration to burn all fees and tips.
* Prevent gaining points during era https://github.com/Joystream/joystream/issues/4286#issuecomment-1246668079

The EraPayout trait
https://github.com/Joystream/substrate/blob/31f4f7654cf8e27e0d73ffeb1c11b2df2d4fbffe/frame/staking/src/lib.rs#L768

## Risks
Adding more validators or removing to the set will force us to re-enable the staking pallet calls in the filter. If we need to do this before the right time, it can be problematic. We may also face this problem if validator keys are compromised and need to be changed.

### Disadvantage
As pointed out in a recent call, launching this way prevents new participants from preparing ahead of time for the switch to NPoS.

[...]

PR #4289, opened 2022-09-12
Carthage poa

dobertRowneySr:
implements #3802 proposal

### Notes
- network tests needs to be reviewed as the consensus logic has changed

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -466,11 +466,6 @@ impl<R: OnUnbalanced<NegativeImbalance>> OnUnbalanced<NegativeImbalance> for Dea
     }
 }
 
-pub struct BurnAllTxFees();
-impl OnUnbalanced<NegativeImbalance> for BurnAllTxFees {
-    fn on_unbalanceds<B>(mut _fees_then_tips: impl Iterator<Item = NegativeImbalance>) {}
-}
-
 parameter_types! {
     pub const TransactionByteFee: Balance = 2 * currency::MILLICENTS; // TODO: adjust value
     /// This value increases the priority of `Operational` transactions by adding
@@ -479,7 +474,7 @@ parameter_types! {
 }
 
 impl pallet_transaction_payment::Config for Runtime {
-    type OnChargeTransaction = CurrencyAdapter<Balances, BurnAllTxFees>;
+    type OnChargeTransaction = CurrencyAdapter<Balances, DealWithFees<Author>>;
     type OperationalFeeMultiplier = OperationalFeeMultiplier;
     type WeightToFee = constants::fees::WeightToFee;
     type LengthToFee = ConstantMultiplier<Balance, TransactionByteFee>;
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -466,6 +466,11 @@ impl<R: OnUnbalanced<NegativeImbalance>> OnUnbalanced<NegativeImbalance> for Dea
     }
 }
 
+pub struct BurnAllTxFees();
+impl OnUnbalanced<NegativeImbalance> for BurnAllTxFees {
+    fn on_unbalanceds<B>(mut _fees_then_tips: impl Iterator<Item = NegativeImbalance>) {}
+}
+
 parameter_types! {
     pub const TransactionByteFee: Balance = 2 * currency::MILLICENTS; // TODO: adjust value
     /// This value increases the priority of `Operational` transactions by adding
@@ -474,7 +479,7 @@ parameter_types! {
 }
 
 impl pallet_transaction_payment::Config for Runtime {
-    type OnChargeTransaction = CurrencyAdapter<Balances, DealWithFees<Author>>;
+    type OnChargeTransaction = CurrencyAdapter<Balances, BurnAllTxFees>;
     type OperationalFeeMultiplier = OperationalFeeMultiplier;
     type WeightToFee = constants::fees::WeightToFee;
     type LengthToFee = ConstantMultiplier<Balance, TransactionByteFee>;
--- a/runtime/src/tests/handle_fees.rs
+++ b/runtime/src/tests/handle_fees.rs
@@ -1,8 +1,9 @@
 use super::{
     account_from_member_id, increase_total_balance_issuance_using_account_id, initial_test_ext,
 };
-use crate::{currency, AccountId, Balances, DealWithFees, NegativeImbalance};
+use crate::{currency, AccountId, Balances, BurnAllTxFees, DealWithFees, NegativeImbalance};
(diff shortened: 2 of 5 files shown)


mnaamani:
I think we have to consider actually dropping the NPoS related pallets. Since the staking pallet still is part of the runtime, the validators are still accumulating points and can be payed rewards. If this is the only problem we can just configure zero rewards as described in https://github.com/Joystream/joystream/issues/4286

dobertRowneySr:
> I think we have to consider actually dropping the NPoS related pallets. Since the staking pallet still is part of the runtime, the validators are still accumulating points and can be payed rewards. If this is the only problem we can just configure zero rewards as described in #4286

Implemented, see my reply on that issue about how I went about burning all the tx fees

mnaamani:
> > I think we have to consider actually dropping the NPoS related pallets. Since the staking pallet still is part of the runtime, the validators are still accumulating points and can be payed rewards. If this is the only problem we can just configure zero rewards as described in #4286
> 
> Implemented, see my reply on that issue about how I went about burning all the tx fees

Yep that looks correct. A neater way (instead of commenting out code) might be to have an additional type in place of 
`struct DealWithFees` say `struct BurnAllFeesAndTips`. This way cargo test will also not need to be commented for DealWithFees, just adding one more test for `BurnAllFeesAndTips` in https://github.com/Joystream/joystream/blob/carthage/runtime/src/tests/handle_fees.rs

mnaamani:
needs cargo-fmt

runtime/src/lib.rs
```diff
@@ -466,6 +466,11 @@ impl<R: OnUnbalanced<NegativeImbalance>> OnUnbalanced<NegativeImbalance> for Dea
     }
 }
 
+pub struct DealWithFeesPoA<R>(PhantomData<R>);
```
mnaamani: `DealWithFeesPoA` doesn't seem to be the right name, how about `BurnAllTxFees` and lets add a test similar to the one for `DealWithFees`. Arguably that should also be renamed to reflect how it handles fees (now that we have two implementations)

mnaamani: Also since for this impl we are not sending fees to anywhere it doesn't really need to be generic. So it can be simply `pub struct BurnAllTxFees {}` and then

```
    type OnChargeTransaction = CurrencyAdapter<Balances, BurnAllTxFees>;
```

mnaamani (changes requested):
I think you forgot to import `BurnAllTxFees` in the tests:

```
 error[E0433]: failed to resolve: use of undeclared type `BurnAllTxFees`
  --> runtime/src/tests/handle_fees.rs:83:9
   |
83 |         BurnAllTxFees::<Recipient>::on_unbalanceds(vec![fees, tips].into_iter());
   |         ^^^^^^^^^^^^^ not found in this scope
   |
help: consider importing this struct
   |
1  | use crate::BurnAllTxFees;
```

mnaamani:
clippy fix is not enough, the test is missing import of `BurnAllTxFees`

runtime/src/tests/handle_fees.rs
```diff
@@ -47,3 +48,47 @@ fn block_author_only_receives_tips() {
...
+
+        assert_eq!(fees.peek(), fees_amount,);
+        assert_eq!(tips.peek(), tips_amount,);
+
+        struct Recipient;
+        impl OnUnbalanced<NegativeImbalance> for Recipient {
+            fn on_nonzero_unbalanced(amount: NegativeImbalance) {
+                Balances::resolve_creating(&author(), amount);
+            }
+        }
```
mnaamani: struct Recipient and its impl can be removed here.

runtime/src/tests/handle_fees.rs
```diff
@@ -47,3 +48,47 @@ fn block_author_only_receives_tips() {
         );
     });
 }
+
+#[test]
+fn block_author_receives_no_tips() {
+    initial_test_ext().execute_with(|| {
+        fn author() -> AccountId {
```
mnaamani: you can drop this author it is not used.

runtime/src/tests/handle_fees.rs
```diff
@@ -47,3 +48,47 @@ fn block_author_only_receives_tips() {
         );
     });
 }
+
+#[test]
+fn block_author_receives_no_tips() {
```
mnaamani:
```suggestion
fn all_tx_fees_are_burned() {
```

runtime/src/tests/handle_fees.rs
dobertRowneySr: It is user later to make sure that author free balances is 0

dobertRowneySr:
#### How forcing new era works
- independently of the session index we are in a new era is started immediately (together with the respective election cycle)
- forcing is then set to `NotForcing` allowing the triggering of new eras only at multiple of `SessionsPerEra`

Lezek123 (changes requested):
I tested this solution, but unfortunately it looks like the era points and rewards are still accumulated by the initial validator set. This is because, contrary to our assumption, the `ActiveEra` is not actually `None` when the chain is started, but it gets initialized to era `0` regardless of the value of `Forcing`: https://github.com/paritytech/substrate/blob/master/frame/staking/src/pallet/impls.rs#L305

On the screen below you can see the initial value of `ActiveEra` (this is while still in POA):
image

After the `staking.forceNewEra` extrinsic is called by sudo, the `ErasValidatorReward` entry for era `0` is populated based on, among others, the era length (which means the longer the initial era, the higher are the rewards accumulated):
image

In era `1`, which started at around block `160` a new validator set was selected, so the initial validator set was no longer producing blocks. The era length in my configuration is `60`, so the era `1` was much shorter than era `0`, but the new validator set had much higher stakes, which caused the validator rewards to increase significantly. You can notice however, that even with those increased rewards, the total reward for era `1` is lower that the reward for era `0` (because the era `0` was much longer):
image

And this is the `Payouts` view during era `2` (Alice and Bob were the POA validators, while the other 2 keys are the current ones):
image

**Possible solution: Reward curve?**

I was thinking about how can this issue be addressed and while writing this summary an idea came to my mind. When the validator set was re-elected at era `1` the validators that were elected had much higher stakes than the initial authorities, which made the rewards go up, as the total stake became closer to "ideal_stake".

What if we were to set the initial authorities' stakes to `0` and construct a reward curve in such a way that when `total_stake == 0`, the inflation rate is also `0`?

This would allow us to still not need to do a runtime upgrade, because basically any validator set except the inital POA set would always have `total_stake > 0`.

I haven't tried this solution, but seems like something worth investigating.

dobertRowneySr:
closing in favor of https://github.com/Joystream/joystream/pull/4315

Issue #4266, opened 2022-09-07
Fee profile command to CLI


bedeho:
# Background
There are a variety of scenarios where it would be useful to be able to guague how a new node/runtime would fare in terms of the cost of certain key usage patterns, they are

* when tweaking some parameter or some core business logic associated with some action, e.g. in development mode.
* when needing to determine the appropriate state bloat bond for key actions, since they must be no less than cleanup transaction fees: _this is already needed for Carthage_
* when wanting to do simulations how the system would work overall, and these costs are important inputs.

# Proposal
Add a new command to the CLI, e.g. call it `fee_profile`, which can ask a node what the cost of a suite of standard extrinsics are at that time. Here are some examples

* Buy a membership
* Make a channel + 2 assets
* Delete a channel.
* Make a video + 2 assets
* Delete a video
* Submit a comment on a video
* Make a , non-editable, forum thread with post of max length.
* Delete a forum thread
 
Now importantly, there are some free variables in all of these which will impact cost, the CLI can either have some hard coded assumptions about this (e.g. how long is a normal but long video title), or they can be imported in a separate JSON file so as to not require updating the CLI itself if you want to play with different values. The script should for each action list some key information, like the 
 
* inclusion fee: useful to have in $USD (if input price is provided to command), hapi & $JOY
* full fee: (including whatever extra deposit etc. is neede)
* total extrinsic size
 
It may also want to output certain chain state values of relevance to these costs, such as the current fee multiplier, but lets keep it minimal for now. Having output in some structure format, like JSON or something, would be beneficial for other downstream tools.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202938088239534) by [Unito](https://www.unito.io)

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4299

PR #4299, opened 2022-09-16
CLI - fee profile commands

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4266
Currently support all extrinsics mentioned in the issue.

Example:
image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202994460839783) by [Unito](https://www.unito.io)

### DIFF
--- a/cli/src/commands/fee-profile/addForumPost.ts
+++ b/cli/src/commands/fee-profile/addForumPost.ts
@@ -23,7 +23,7 @@ export default class FeeProfileAddForumPost extends FeeProfileCommandBase {
   }
 
   async run(): Promise<void> {
-    const api = await this.getOriginalApi()
+    const api = this.getOriginalApi()
     const { postLen, editable = false } = this.parse(FeeProfileAddForumPost).flags
     const { postDeposit } = api.consts.forum
 
--- a/cli/src/commands/fee-profile/addForumPost.ts
+++ b/cli/src/commands/fee-profile/addForumPost.ts
@@ -1,0 +1,45 @@
+import FeeProfileCommandBase from '../../base/FeeProfileCommandBase'
+import { flags } from '@oclif/command'
+import _ from 'lodash'
+import { formatBalance } from '@polkadot/util'
+import chalk from 'chalk'
+
+const DEFAULT_POST_LENGTH = 200
+
+export default class FeeProfileAddForumPost extends FeeProfileCommandBase {
+  static description = 'Create fee profile of forum.add_post extrinsic.'
+
+  static flags = {
+    postLen: flags.integer({
+      char: 'p',
+      default: DEFAULT_POST_LENGTH,
+      description: 'Post length to use for estimating tx fee',
+    }),
+    editable: flags.boolean({
+      char: 'e',
+      description: 'If specified - `editable` parameter is set to true when estimating the costs',
+    }),
+    ...super.flags,
+  }
+
+  async run(): Promise<void> {
+    const api = await this.getOriginalApi()
+    const { postLen, editable = false } = this.parse(FeeProfileAddForumPost).flags
+    const { postDeposit } = api.consts.forum
+
+    this.log(`Post deposit: ${chalk.cyanBright(formatBalance(postDeposit))}`)
+    this.log('Parameters:')
+    this.jsonPrettyPrint(JSON.stringify({ postLen, editable }))
+
+    const tx = api.tx.forum.addPost(0, 0, 0, _.repeat('x', postLen), editable)
+    const txFee = await this.getApi().estimateFee(this.pairs.alice, tx)
+    this.profile(
+      editable
+        ? {
+            txFee,
+            postDeposit,
+          }
+        : { txFee }
+    )
+  }
+}
--- a/cli/src/commands/fee-profile/buyMembership.ts
+++ b/cli/src/commands/fee-profile/buyMembership.ts
@@ -11,6 +11,8 @@ const DEFAULT_HANDLE_LENGTH = 10
 const DEFAULT_NAME_LENGTH = 10
 const DEFAULT_ABOUT_LENGTH = 0
 const DEFAULT_AVATAR_URI_LENGTH = 25
+const DEFAULT_EXTERNAL_RESOURCES_COUNT = 1
+const DEFAULT_EXTERNAL_RESOURCE_LENGTH = 25
 
 export default class FeeProfileBuyMembershipCommand extends FeeProfileCommandBase {
   static description = 'Create fee profile of members.buy_membership extrinsic.'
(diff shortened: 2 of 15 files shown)


cli/src/commands/fee-profile/createChannel.ts
```diff
@@ -0,0 +1,168 @@
...
+      avatarSize: inputAvatarSize,
+      coverSize: inputCoverSize,
+      noAvatar,
+      noCover,
+      titleLen,
+      descriptionLen,
+    } = this.parse(FeeProfileCreateChannel).flags
+    const dataSizeFee = await api.query.storage.dataObjectPerMegabyteFee()
+    const dataObjectBloatBond = await api.query.storage.dataObjectStateBloatBondValue()
+    const channelBloatBond = await api.query.content.channelStateBloatBondValue()
```
zeeshanakram3: We have helper functions defined in [cli/src/Api.ts](https://github.com/Lezek123/substrate-runtime-joystream/blob/6b1a74b9803428f196c0079bd69d8a3b0a0279ca/cli/src/Api.ts#L478) for bloat bond values. Maybe we can use that.

cli/src/commands/fee-profile/addForumPost.ts
```diff
@@ -0,0 +1,45 @@
...
+    }),
+    editable: flags.boolean({
+      char: 'e',
+      description: 'If specified - `editable` parameter is set to true when estimating the costs',
+    }),
+    ...super.flags,
+  }
+
+  async run(): Promise<void> {
+    const api = await this.getOriginalApi()
```
zeeshanakram3: Here `await` is unnecessary. Also, in all the fee-profile commands, the same expression with await is used. Can you please have a look at all these instances
```suggestion
    const api = this.getOriginalApi()
```

cli/src/commands/fee-profile/deleteVideo.ts
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    const tx = api.tx.content.deleteVideo({ Member: 0 }, 0, assetsNum, createType('Option<u32>', storageBucketsNum))
+    const txFee = await this.getApi().estimateFee(this.pairs.alice, tx)
+    const costs = {
+      txFee,
+    }
+    const returns = {
+      dataObjectsBloatBond: dataObjectBloatBond.muln(assetsNum),
+      videoBloatBond,
```
zeeshanakram3, line 53: Wouldn't using boat bond values directly from chain state can show incorrect `JOY` returns, since the bloat bond values can change after the video was created?

cli/src/commands/fee-profile/deleteChannel.ts
```diff
@@ -0,0 +1,79 @@
...
+      }),
+      assetsNum
+    )
+    const txFee = await this.getApi().estimateFee(this.pairs.alice, tx)
+    const costs = {
+      txFee,
+    }
+    const returns = {
+      dataObjectsBloatBond: dataObjectBloatBond.muln(assetsNum),
+      channelBloatBond,
```
zeeshanakram3, line 75: Wouldn't using boat bond values directly from chain state can show incorrect `JOY` returns, since the bloat bond values can change after the video was created?

cli/src/commands/fee-profile/buyMembership.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    this.log(`Membership price: ${chalk.cyanBright(formatBalance(membershipPrice))}`)
+    this.log('Parameters:')
+    this.jsonPrettyPrint(JSON.stringify({ handleLength, nameLength, aboutLength, avatarUriLength }))
+
+    const metadata: IMembershipMetadata = {
+      name: _.repeat('x', nameLength),
+      about: _.repeat('x', aboutLength),
+      avatarUri: _.repeat('x', avatarUriLength),
+    }
```
zeeshanakram3: Should the estimated fee also include the metadata of the external resource i.e., estimating fee in terms of no. of external resources & length of each external resource URL

cli/src/commands/fee-profile/deleteForumThread.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    this.log(`Thread deposit: ${chalk.cyanBright(formatBalance(threadDeposit))}`)
+
+    const tx = api.tx.forum.deleteThread(0, 0, 0, true)
+    const txFee = await this.getApi().estimateFee(this.pairs.alice, tx)
+    const costs = {
+      txFee,
+    }
+    const returns = {
+      threadDeposit,
```
zeeshanakram3, line 24: Same here. Wouldn't using thread deposit value directly from the chain state can show incorrect `JOY` returns, since the bloat bond values can change after the video was created?

cli/src/commands/fee-profile/deleteForumPost.ts
```diff
@@ -0,0 +1,48 @@
...
+        new Map([[createType('PalletForumExtendedPostIdObject', { categoryId: 0, threadId: 0, postId: 0 }), true]])
+      ),
+      _.repeat('x', rationaleLen)
+    )
+    const txFee = await this.getApi().estimateFee(this.pairs.alice, tx)
+    const costs = {
+      txFee,
+    }
+    const returns = {
+      postDeposit,
```
zeeshanakram3, line 44: Same here. Wouldn't using post deposit value directly from the chain state can show incorrect `JOY` returns, since the bloat bond values can change after the video was created?

cli/src/commands/fee-profile/deleteVideo.ts
Lezek123: This command does not estimate the cost of deleting a given, existing video, but just _some_ video.
The intention behind those commands is that they can be used to verify how given runtime parameters (like _WeightToFee_ function, bloat bond values etc.) affect the costs of some common extrinsics (see issue https://github.com/Joystream/joystream/issues/4266).
I can make the bloat bond confiugrable via `--videoBloatBond` flag, but I'm not sure if that's necessary given this purpose (so assuming the commands will mostly be ran against the local/dev chain anyway, where the user has full control over bloat bond values).

cli/src/commands/fee-profile/buyMembership.ts
Lezek123: Oh, I didn't realize those were added. Yes, it makes sense, I'll add support for this.

cli/src/commands/fee-profile/deleteForumThread.ts
Lezek123: Thread deposit is a constant, so it won't change unless there's a runtime upgrade. Also: see my first answer w.r.t. bloat bonds.

cli/src/commands/fee-profile/deleteVideo.ts
zeeshanakram3: Yeah, this implementation makes sense if commands are being used during local development

zeeshanakram3 (approved):
LGTM

Issue #2832, opened 2021-11-15
JS Website: Redesign Community Page


mochet:
Currently, the `community` link on the Joystream homepage only links to our Discord server which is undesirable for many, many reasons (detailed more here: https://github.com/Joystream/joystream-org/issues/342)

The page should be heavily designed in order to focus on explaining much better what our actually `community` is, and also include unofficial/community chat channels/servers as well as the official project ones.

A very rough mock up what this could look like:
![image](https://user-images.githubusercontent.com/50830161/141820213-58d53c2b-cf1e-4ca4-abae-4f77cff8d3b5.png)

bedeho:
Transferring this to the website repo.

PR #4309, opened 2022-09-25
fixed wrong leader for thread_moderated event

mkbeefcake:
Fixed the issue 

[USERSNAP] Thread moderated by me is displayed in forum activities panel as moderated by svasilenko (the previous lead) #2832
https://github.com/Joystream/pioneer/issues/2832

### DIFF
--- a/query-node/mappings/src/workingGroups.ts
+++ b/query-node/mappings/src/workingGroups.ts
@@ -509,6 +509,12 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
                 application.id
               )
             }
+
+            if (opening.type === WorkingGroupOpeningType.LEADER) {
+              // setting isLead of existing leader to false.
+              await removeIsLeadFromGroup(store, group.id)
+            }
+
             const worker = new Worker({
               id: `${group.name}-${workerRuntimeId.toString()}`,
               runtimeId: workerRuntimeId.toNumber(),
@@ -548,6 +554,23 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
     leaderSetEvent.worker = hiredWorkers[0]
     await store.save<LeaderSetEvent>(leaderSetEvent)
   }
+}
+
+async function removeIsLeadFromGroup(store: DatabaseManager, groupId: string) {
+  const groupWorkers = await store.getMany(Worker, {
+    where: {
+      group: { id: groupId },
+      isLead: true
+    },
+    relations: ['group']
+  })
+
+  await Promise.all(
+    groupWorkers.map((worker) => {
+      worker.isLead = false
+      return store.save<Worker>(worker)
+    })
+  )
 }
 
 export async function workingGroups_OpeningCanceled({ store, event }: EventContext & StoreContext): Promise<void> {
--- a/query-node/mappings/src/workingGroups.ts
+++ b/query-node/mappings/src/workingGroups.ts
@@ -484,6 +484,11 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
 
   await store.save<OpeningFilledEvent>(openingFilledEvent)
 
+  // Remove previous lead if necessary
+  if (opening.type === WorkingGroupOpeningType.LEADER && acceptedApplicationIds.length > 0) {
+    await removeIsLeadFromGroup(store, group.id)
+  }
+
   // Update applications and create new workers
   const hiredWorkers = (
     await Promise.all(
@@ -508,11 +513,6 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
                 'Fatal: No worker id found by accepted application when handling OpeningFilled event!',
                 application.id
               )
-            }
-
-            if (opening.type === WorkingGroupOpeningType.LEADER) {
-              // setting isLead of existing leader to false.
-              await removeIsLeadFromGroup(store, group.id)
             }
 
             const worker = new Worker({
--- a/query-node/mappings/src/workingGroups.ts
+++ b/query-node/mappings/src/workingGroups.ts
@@ -560,9 +560,9 @@ async function removeIsLeadFromGroup(store: DatabaseManager, groupId: string) {
   const groupWorkers = await store.getMany(Worker, {
     where: {
       group: { id: groupId },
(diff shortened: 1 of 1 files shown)


query-node/mappings/src/workingGroups.ts
```diff
@@ -534,6 +550,12 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
                 `Fatal: No worker id found by accepted application id ${application.id} when handling OpeningFilled event!`
               )
             }
+
+            if (opening.type === WorkingGroupOpeningType.LEADER) {
+              // setting isLead of other workers to false.
+              await removeIsLeadFromGroup(store, 'forumWorkingGroup')
```
singulart: This hard-coded value has to be replaced with the value of the _current_ group.

query-node/mappings/src/forum.ts
```diff
@@ -111,10 +112,20 @@ async function getPollAlternative(store: DatabaseManager, threadId: string, inde
...
+  })
+
+  if (!workingGroup) {
+    throw new Error(`Forum Working Group not found!`)
+  }
+  
   const worker = await store.get(Worker, {
     where: {
       group: { id: 'forumWorkingGroup' },
       ...(actor.isLead ? { isLead: true } : { runtimeId: actor.asModerator.toNumber() }),
```
singulart: I think there should be one `? :` operator,  not two.. 

Essentially, if `actor.isLead == true`, use the `leader.id` from the `workingGroup`, or else use moderator Id

singulart (changes requested):

query-node/mappings/src/workingGroups.ts
```diff
@@ -473,6 +473,22 @@ export async function workingGroups_AppliedOnOpening({ store, event }: EventCont
...
+    where: {
+      group: { id: groupId }
+    },
+    relations: ['group']
+  })
+
+  await Promise.all(
+    groupWorkers.map((worker) => {
+      worker.isLead = false
+      return store.save<Worker>(worker)
```
singulart: Such a loop will cause many SQL statements. For large working groups this may become a performance issue. 

I would suggest to look for the possibility to do only one SQL statement: 

`update WG_WORKERS set isLead = 0 where group_id=?1`

query-node/mappings/src/workingGroups.ts
mkbeefcake: I confused this with other function. i will fix it.

query-node/mappings/src/forum.ts
mkbeefcake: image
In https://github.com/Joystream/pioneer/issues/2832, I figured out two leaders in same group, so i tried to add new filter for it one more.

query-node/mappings/src/workingGroups.ts
mkbeefcake: I didn't find out the SQL running way, so added filter option; isLead: true, so it will return one record and only execute one SQL statement.

singulart: Ok, from my perspective this is enough. Thank you ;)

query-node/mappings/src/forum.ts
singulart: Yes, I understand. But you don't need two filters. Only one is enough: 

`...(actor.isLead ? { id: workingGroup.leader?.id } : { runtimeId: actor.asModerator.toNumber() })`

singulart (changes requested):

singulart (approved):

singulart:
LGTM. @dmtrjsg How to deal with these tickets? Who from JSG can review them?

dmtrjsg:
@singulart for all QN work also pls feel free to tag @thesan 👍

query-node/mappings/src/forum.ts
```diff
@@ -83,10 +84,19 @@ async function getPost(store: DatabaseManager, postId: string, relations?: 'thre
...
+  if (!workingGroup) {
+    throw new Error(`Forum Working Group not found!`)
+  }
+  
   const worker = await store.get(Worker, {
     where: {
       group: { id: 'forumWorkingGroup' },
-      ...(actor.isLead ? { isLead: true } : { runtimeId: actor.asModerator.toNumber() }),
+      ...(actor.isLead ? { id: workingGroup.leader?.id } : { runtimeId: actor.asModerator.toNumber() }),
     },
```
thesan: Deploying new mappings usually means remapping the QN. In this case `isLead` will have been fixed by the changes in  `query-node/mappings/src/workingGroups.ts`. So the way I understand it, this changes (in `query-node/mappings/src/forum.ts`) are not necessary. Or am I missing something ?

query-node/mappings/src/workingGroups.ts
```diff
@@ -509,6 +526,12 @@ export async function workingGroups_OpeningFilled({ store, event }: EventContext
                 application.id
               )
             }
+
+            if (opening.type === WorkingGroupOpeningType.LEADER) {
+              // setting isLead of existing leader to false.
+              await removeIsLeadFromGroup(store, group.id)
```
thesan: It's a detail but on line currently `566` there is:
```ts
// Update working group and LeaderSetEvent if necessary
if (opening.type === WorkingGroupOpeningType.LEADER && hiredWorkers.length) {
  group.leader = hiredWorkers[0]
  await store.save<WorkingGroup>(group)
  ...
```

IMO this:
```ts
if (opening.type === WorkingGroupOpeningType.LEADER) {
  // setting isLead of existing leader to false.
  await removeIsLeadFromGroup(store, group.id)
}
```
Should be called just before line `566` instead of in the `hiredWorkers` map. This way would be clearer that it's only called once.

thesan (changes requested):
Sorry for the delayed review. I kept on postponing it due to other task linked to the `carthage` release

thesan (changes requested):
Nice the last remaining thing is to run `yarn workspace query-node-root lint --fix` and push the changes again

thesan (approved):
Looks great ! Thank you for adding all the changes :pray:

zeeshanakram3 (approved):
LGTM

Issue #3036, opened 2022-01-13
Research Docker features for caching some build artefacts locally.


shamil-gadelshin:
The issue from the retrospective call: https://miro.com/app/board/uXjVOZwmt_k=/

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201703380142370/1201703600462154) by [Unito](https://www.unito.io)

mnaamani:
Some useful resources to consider:
https://fenollp.github.io/faster-rust-builds-docker_host
https://www.lpalmieri.com/posts/fast-rust-docker-builds/
https://benjamincongdon.me/blog/2019/12/04/Fast-Rust-Docker-Builds-with-cargo-vendor/
https://www.youtube.com/watch?v=xuqolj01D7M

mnaamani:
also https://github.com/Joystream/joystream/issues/3035#issuecomment-1027639018

mnaamani:
Explored the multiple tools and picked `cargo-chef` it works, and requires only changes in the `Dockerfile`.
Implemented in #4310

Issue #2853, opened 2021-11-23
The apps.Dockerfile tries to build ignored package


jodator:
While trying to build docker images form the `olympia` branch I find that:

*   The `apps.Dockerfile` is running the [`yarn build:packages` script](https://github.com/Joystream/joystream/blob/olympia/apps.Dockerfile#L15)
*   This runs the [`build-npm-packages.sh`](https://github.com/Joystream/joystream/blob/olympia/build-npm-packages.sh) which builds `@joystream/cli` among others

Unfortunatelly, the `cli` is not pushed to docker workspace as it is ignored: https://github.com/Joystream/joystream/blob/olympia/.dockerignore#L7

PR #4310, opened 2022-09-26
Faster Docker Builds with cargo-chef

mnaamani:
Fixes
https://github.com/Joystream/joystream/issues/3036
https://github.com/Joystream/joystream/issues/2853

Note: For the joystream-node.Dockerfile I'm no longer copying in the entire git repo (including .git/) so during the build step the custom build script cannot get the git commit hash to append to the node version string output of `--version`. So we get "unknown" instead of commit hash:

```
docker run joystream/node:test --version
node 7.3.0-unknown
```

We can set env variable manually: `SUBSTRATE_CLI_GIT_COMMIT_HASH` if we wish, but wondering if it makes any sense to add it at all.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203052029537863) by [Unito](https://www.unito.io)


mnaamani:
I realized I made some changes the `joystream-node-docker-dev.yml` workflow which should have been done for the `run-network-tests.yml` workflow instead. I'll revert and update the latter.

joystream-node.Dockerfile
```diff
@@ -1,25 +1,50 @@
...
+COPY runtime-modules ./runtime-modules
+RUN cargo chef prepare --recipe-path /joystream/recipe.json
+
+FROM rust AS cacher
+LABEL description="Cargo chef cook dependencies"
+WORKDIR /joystream
+COPY --from=planner /joystream/recipe.json /joystream/recipe.json
+ENV WASM_BUILD_TOOLCHAIN=nightly-2022-05-11
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path /joystream/recipe.json
```
dobertRowneySr, line 28: Is the `recipe.json` set up manually, if so should we vc it?

dobertRowneySr (approved):
LGTM, I just left a small question

joystream-node.Dockerfile
mnaamani: the recipie.json file is produced in the previous docker layer "planner" with the command `cargo chef prepare ..` It is copied into this layer with the `COPY` command on line 25.

Issue #3745, opened 2022-05-12
No panics in runtime/node codebase


bedeho:
Check that no panics exist in the code, regardless of whether they will happen or not, they should not be there.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202266872476717) by [Unito](https://www.unito.io)

PR #4313, opened 2022-09-26
No panics in the runtime: Add clippy rules & fixes

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/3745
Adds clippy rules and macro overrides to try to prevent any code that can cause the runtime to `panic` from being part of the production build

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203063681899246) by [Unito](https://www.unito.io)


mnaamani (approved):
Despite the size of the PR I thoroughly enjoyed going through it. Some nice tricks in there especially with scoping some types where we wanted to allow arithmetic, its too bad it affects the type names but the benefits here outweigh the inconvenience which pioneer and atlas might need to take on to update.

Issue #4269, opened 2022-09-07
Add test for PoA to NPos migration


dobertRowneySr:
with reference to #3802 an integration test must be added in order to verify that the validator selection algorithm has actually changed after runtime upgrade.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202941348565919) by [Unito](https://www.unito.io)

dobertRowneySr:
Draft testing plan based off https://github.com/Joystream/joystream/issues/3802#issuecomment-1250871903
## PoA 
- [x]  calling `staking_pallet::payout_stakers` gives 0 payout to stakers' stash account
- [x] account calling `staking_pallet::{bond, nominate, valid}` should be ok
- [x] `pallet_session::Pallet::<T>::validators` should always be equal to the original (genesis) authorities
- [x] `pallet_staking::force_era()` should be `ForceNone` in PoA period
- [x]  authorities do not get tips from transactions

## NPoS switch
- [x] `pallet_staking::force_era()` should be `ForceNew`
- [x] accounts can become `validators` and they can be elected (the next era)
- [x]  `pallet_session::Pallet::<T>::validators`  can change (the next era)
- [x]  calling `staking_pallet::payout_stakers` gives correct `amount >= 0` to stakers 
- [x] active era index is updated
- [x]  authorities do get tips from transactions

dobertRowneySr:
## Testing Plan PoA section

### Definitions
- PoA is active <-> `assert.isTrue(api.query.staking.forceEra().isForceNonoe)`
- NPoS is active <-> `assert.isFalse(api.query.staking.forceEra().isForceNone)`

#### 1. Current Era is None
##### Arrangement:
- PoA is active
- `x > 0` validator authorities set at genesis
##### Action
- 
##### Assertions:
- currentEra is None
#### 2. Calling payout stakers gives zero to validators
##### Arrangement:
- PoA is active
- `x > 0` validator authorities set at genesis
- create`n > 0` validators backed by `m > 0` nominators, besides original authorities
##### Action:
- a (valid) account claims payouts for all of the above validator and all the authorities stash account with `era > 0` (any positive number)
##### Assertions:
- validators `free_balance` is left unchanged by the claim
- authorities `free_balance` is left unchanged by the claim

#### 3. Validator set is left unchanged
##### Arrangement:
- PoA is active
- `n > 0` validator authorities set at genesis
##### Action:
- run `k > 0` blocks ahead
##### Assertions:
[...]

PR #4315, opened 2022-09-27
Carthage poa npos tests

dobertRowneySr:
implements #4269  (testing plan comment)
I did not use `chai-as-promised` package since it wasn't strictly necessary


runtime/src/tests/handle_fees.rs
```diff
@@ -3,6 +3,7 @@ use super::{
 };
 use crate::{currency, AccountId, Balances, DealWithFees, NegativeImbalance};
 use frame_support::traits::{Currency, Imbalance, OnUnbalanced};
+use sp_runtime::traits::Zero;
```
mnaamani: clippy says drop this import.

tests/network-tests/src/scenarios/carthagePoA.ts
```diff
@@ -0,0 +1,19 @@
+import bondingSucceedsInPoA from 'src/flows/carthage/bondingSucceedsInPoA'
```
mnaamani: change `src/` in import path, to `../` should be  `../flows/carthage/bondingSucceedsInPoA` 

I'm getting `Error: Cannot find module 'src/flows/carthage/bondingSucceedsInPoA'` when trying to run the scenario.

tests/network-tests/src/flows/carthage/constantAuthorities.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const forceEra = await api.getForceEra()
+  assert(forceEra.isForceNone)
+
+  // babe authorities
+  const currentAuthorities = await api.getBabeAuthorities()
+  const nextAuthorities = await api.getNextBabeAuthorities()
+  expect(nextAuthorities).to.be.deep.equal(currentAuthorities)
+
+  // make sure that next Session keys are none
```
mnaamani: the comment seems misleading, don't we mean next session keys are the same as current authority session keys?

tests/network-tests/src/scenarios/carthagePoA.ts
mnaamani: Same issues in other scenario.

For the flows change `src/` to `../../`

Although this problem doesn't show up when running `yarn build` and `yarn lint`, it's caused by the way we run the actual scenario with `node -r ts-node/register  src/scenario-name.ts` which is a bit annoying.

tests/network-tests/src/scenarios/carthageNPoSSwitch.ts
```diff
@@ -0,0 +1,12 @@
+import changingAuthorities from 'src/flows/carthage/changingAuthorities'
+import claimingPayoutsEnabled from 'src/flows/carthage/claimingPayoutEnabled'
+import setForceNewEra from 'src/flows/carthage/setForceNewEra'
+import currentEraIsSomeInPoS from 'src/flows/carthage/currentEraIsSomeInPoS'
+import { scenario } from '../Scenario'
+
+scenario('Carthage NPoS switch', async ({ job }) => {
+  job('set force era to force new', setForceNewEra)
+  job('current era is none', currentEraIsSomeInPoS)
```
mnaamani: jobs in general will run in parallel. In this scenario it makes more sense for job running setForNewEra flow to be the dependency for the remaining jobs. This can be done with the `after` or `require` methods like so:

```
const forceNewEraJob = job('set force era to force new', setForceNewEra)
// will run after forceNewEraJob only if it is successful.
job('current era is none', currentEraIsSomeInPoS).requires(forceNewEraJob)
.
.
```

mnaamani (changes requested):
Left a few change requests..

Made local changes to path to run the carthagePoA scenario to just test first two assertions and getting test for current era being none failing:

```
$ node -r ts-node/register --unhandled-rejections=strict src/scenarios/carthagePoA.ts
  integration-tests:api-factory Connecting to chain, attempt 1.. +0ms
Runtime Version: 11.2.0
  integration-tests:scenario Carthage PoA setup +0ms
  integration-tests:job:authority set is constant Running +0ms
  integration-tests:flow: constant Authorities in PoA started +0ms
  integration-tests:job:current era is none Running +0ms
  integration-tests:flow: validator-set started +0ms
  integration-tests:job:current era is none flow 0 failed: +10ms
AssertionError: Unspecified AssertionError
    at currentEraIsNoneInPoA (/Users/mokhtar/joystream/joystream/tests/network-tests/src/flows/carthage/currentEraIsNoneInPoA.ts:14:9)
    at async /Users/mokhtar/joystream/joystream/tests/network-tests/src/Job.ts:103:11 {
  showDiff: false,
  actual: null,
  expected: undefined
}
  integration-tests:job:current era is none [Failed] +7ms
  integration-tests:job:authority set is constant [Succeeded] +22ms
Job Results:
authority set is constant: Succeeded
current era is none: Failed
Error: Scenario Failed
```

Are tests passing for you?

Also its important to note that the tests are being run with the "TESTING" runtime profile, so no calls are actually filtered in the scenario.

tests/network-tests/src/scenarios/carthagePoA.ts
dobertRowneySr: fixed in 810027b7697f1e024fd708c813d0e187490788e6

mnaamani:
It seems even when setting the staking pallet config Forcing to ForceNone, `try_trigger_new_era()` in `staking::new_session()` is being invoked (by session pallet?)

https://github.com/Joystream/substrate/blob/joystream-carthage/frame/staking/src/pallet/impls.rs#L292

This is why testing for current_ers isNone is failing?

tests/network-tests/src/scenarios/carthageNPoSSwitch.ts
dobertRowneySr: I put all the PoA / NPoS assertions in one job

tests/network-tests/src/flows/carthage/constantAuthorities.ts
ignazio-bovo: Technically the rule is : next session keys None -> keep the current keys set as the next one

tests/network-tests/src/Api.ts
```diff
@@ -311,6 +315,22 @@ export class Api {
     return this.api.consts.babe.expectedBlockTime
   }
 
+  public async getBabeAuthorities(): Promise<string[]> {
+    return await this.api.query.babe.authorities().then((key) => key.map(([k]) => k.toString()))
```
mnaamani, line 319:
```suggestion
    return (await this.api.query.babe.authorities()).map(key => key.toString())
```
.. is easier on the eyes :)

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,77 @@
...
+import { BondingSucceedsFixture } from '../../fixtures/staking/BondingSucceedsFixture'
+import { ValidatingSucceedsFixture } from '../../fixtures/staking/ValidatingSucceedsFixture'
+import { NominatingSucceedsFixture } from '../../fixtures/staking/NominatingSucceedsFixture'
+
+export default async function carthagePoAAssertions({ api, query, env }: FlowProps): Promise<void> {
+  const debug = extendDebug('flow: constant Authorities in PoA')
+  debug('started')
+  api.enableDebugTxLogs()
+
+  const bondAmount = new BN(1000000000000000)
```
mnaamani: Perhaps instead of this constant we should take into account the min_validator_bond and min_nominator_bond values, so the test doesn't break if we change those values.
`api.query.staking.minValidatorBond` and `api.query.minNominatorBond` I think.

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,77 @@
...
+  const forceEra = await api.getForceEra()
+  assert(forceEra.isForceNone)
+
+  // 1. Authorities are constant
+  // 1.a. babe authorities are constant
+  const currentAuthorities = await api.getBabeAuthorities()
+  const nextAuthorities = await api.getNextBabeAuthorities()
+  expect(nextAuthorities).to.be.deep.equal(currentAuthorities)
+
+  // 1.b. Next session keys are none
```
mnaamani: The comment doesn't match what is being asserted.
Do we mean, "next authorities are equal to current session authorities" ?

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,77 @@
...
+  const activeEra = await api.getActiveEra()
+  if (activeEra.isSome) {
+    const { index } = activeEra.unwrap()
+    const nextEraIndex = index.addn(1)
+    const nextSessionIndex = await api.getErasStartSessionIndex(nextEraIndex as u32)
+    assert.equal(index.toNumber(), 0)
+    assert(nextSessionIndex.isNone)
+  }
+
+  // 3. Bonding succeds
```
mnaamani: Of course we want to be testing that bonding works in frozen stage, where balance transfers are not allowed.
Now the testing runtime doesn't filter any calls so this works. Perhaps we want to setup initial balances for the validator and nominator accounts and run the test in the production runtime?

tests/network-tests/src/Api.ts
mnaamani:
```suggestion
    return (await this.api.query.babe.authorities()).map(([k]) => k.toString())
```

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+    validatorAccount
+  )
+  const candidationFixture = new FixtureRunner(validatorCandidatingSucceedsFixture)
+  await candidationFixture.run()
+
+  // 5. Nominating succeds:
+  const nominatorCandidatingSucceedsFixture = new NominatingSucceedsFixture(api, [validatorAccount], nominatorAccount)
+  const nominationFixture = new FixtureRunner(nominatorCandidatingSucceedsFixture)
+  await nominationFixture.run()
```
mnaamani, line 56: We should also check that validation succeeds, not just bonding

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // Switch to NPoS
+  const setForceEraForcingNewFixture = new SetForceEraForcingNewFixture(api)
+  const fixtureRunner = new FixtureRunner(setForceEraForcingNewFixture)
+  await fixtureRunner.run()
+
+  // And wait until new era happened: about 10 minutes
+  let forceEraStatus = await api.getForceEra()
+  while (forceEraStatus.isForceNew) {
+    sleep(sleepTimeSeconds * 1000)
```
mnaamani:
```suggestion
    await sleep(sleepTimeSeconds * 1000)
```

tests/network-tests/src/scenarios/carthageNPoSSwitch.ts
```diff
@@ -0,0 +1,10 @@
+// import changingAuthorities from 'src/flows/carthage/changingAuthorities'
+// import claimingPayoutsEnabled from 'src/flows/carthage/claimingPayoutEnabled'
+// import setForceNewEra from 'src/flows/carthage/setForceNewEra'
+// import currentEraIsSomeInPoS from 'src/flows/carthage/currentEraIsSomeInPoS'
+import { scenario } from '../Scenario'
+import switchToNPoS from '../flows/carthage/switchToNPoS'
+
+scenario('Carthage NPoS switch', async ({ job }) => {
+  job('switch to npos checks', switchToNPoS)
```
mnaamani: shouldn't this scenario setup the validators and nominators with running the carthagePoAAssertions before doing the switch? Or will there be an issue with block production after switch if we don't run a second validator?

mnaamani (changes requested):
Good progress so far. I think the assertions are correct, but in the switch to NPoS I think we want to the test to be a bit more comprehensive.

mnaamani:
I just had an observation, this happens both on carthage branch and the current PoA PR, that the initial validators in simple dev chain setup "Alice", although they are correctly producing and finalizing blocks, and they appear in the validators list on the overview page, in the accounts page it doesn't show correctly and we must run validate action again to show up in accounts as validator:

Screen Shot 2022-10-06 at 10 35 32 AM

Screen Shot 2022-10-06 at 10 34 52 AM

Screen Shot 2022-10-06 at 10 33 10 AM

Screen Shot 2022-10-06 at 10 36 58 AM

But it looks like it was only to set additional validator preferences:

Screen Shot 2022-10-06 at 10 40 06 AM

Although it does look a bit confusing.. cc: @bwhm

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
dobertRowneySr: Replaced with : "next session keys (queued keys) are the same as the current sesssion"

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
dobertRowneySr: https://crates.parity.io/pallet_staking/pallet/enum.Call.html#variant.validate
Haven't I checked that at assertion 4?

runtime/src/lib.rs
```diff
@@ -537,6 +538,28 @@ pallet_staking_reward_curve::build! {
...
 }
 
+pub struct NoInflationIfNoEras;
+impl EraPayout<Balance> for NoInflationIfNoEras {
+    fn era_payout(
+        total_staked: Balance,
+        total_issuance: Balance,
+        era_duration_millis: u64,
+    ) -> (Balance, Balance) {
+        let era_index = pallet_staking::Pallet::<Runtime>::current_era().unwrap_or_else(Zero::zero);
```
Lezek123: Based on those definitions of `ActiveEra` and `CurrentEra`:
```rust
	/// The current era index.
	///
	/// This is the latest planned era, depending on how the Session pallet queues the validator
	/// set, it might be active or not.
	#[pallet::storage]
	#[pallet::getter(fn current_era)]
	pub type CurrentEra<T> = StorageValue<_, EraIndex>;

	/// The active era information, it holds index and start.
	///
	/// The active era is the era being currently rewarded. Validator set of this era must be
	/// equal to [`SessionInterface::validators`].
	#[pallet::storage]
	#[pallet::getter(fn active_era)]
	pub type ActiveEra<T> = StorageValue<_, ActiveEraInfo>;
```
And the fact that `end_era` (which tiggers `era_payout`) takes `active_era` as argument, it seems like you should actually query `active_era` here.

I tested this locally and `current_era` changes before `active_era`, causing validators to still recieve rewards for era `0`.

tests/network-tests/.eslintrc.js
```diff
@@ -8,5 +8,6 @@ module.exports = {
     'new-cap': 'off',
     // Disabled because of the false positive bug: https://github.com/eslint/eslint/issues/11899
     'require-atomic-updates': 'off',
+    '@typescript-eslint/no-floating-promises': 'off',
```
Lezek123, line 11: I think that's a good rule to have enabled. If there really needs to have a "floating" promise somewhere, it can be marked with `void`

tests/network-tests/src/fixtures/staking/ValidatingSucceedsFixture.ts
```diff
@@ -0,0 +1,21 @@
+import { Api } from '../../Api'
+import { BaseFixture } from '../../Fixture'
+
+export class ValidatingSucceedsFixture extends BaseFixture {
+  protected preferences: any
```
Lezek123: Please try to avoid using `any` type. You can use `createType` utility to construct `ValidatorPrefs` in a typesafe way.

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,78 @@
...
+  const sessionAuthorities = await api.getSessionAuthorities()
+  const queuedKeys = await api.getQueuedKeys()
+  expect(queuedKeys).to.be.deep.equal(sessionAuthorities)
+
+  // 2. Next Era starting session index is none
+  const activeEra = await api.getActiveEra()
+  if (activeEra.isSome) {
+    const { index } = activeEra.unwrap()
+    const nextEraIndex = index.addn(1)
+    const nextSessionIndex = await api.getErasStartSessionIndex(nextEraIndex as u32)
```
Lezek123: `nextSessionIndex` is a little confusing name. Perhaps `nextEraStartSessionIndex`?

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
```diff
@@ -0,0 +1,78 @@
...
+  // 1. Authorities are constant
+  // 1.a. babe authorities are constant
+  const currentAuthorities = await api.getBabeAuthorities()
+  const nextAuthorities = await api.getNextBabeAuthorities()
+  expect(nextAuthorities).to.be.deep.equal(currentAuthorities)
+
+  // 1.b. Queued keys (for next session) and current session keys are the same
+  const sessionAuthorities = await api.getSessionAuthorities()
+  const queuedKeys = await api.getQueuedKeys()
+  expect(queuedKeys).to.be.deep.equal(sessionAuthorities)
```
Lezek123: I think similar checks should also be done **after** some other accounts have already called bond&validate / bond&nominate. So for example:
- get current authorities / sessions authorities _(maybe also imOnline.keys?)_
- use some new accounts to bond & validate/nominate (new validators should have higher stakes than the current ones)
- wait for at least `era_duration` blocks
- make sure the authorities stay unchanged, queuedKeys are equal to current authorities and nextAuthorities are also equal to current authorities

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+  const setForceEraForcingNewFixture = new SetForceEraForcingNewFixture(api)
+  const fixtureRunner = new FixtureRunner(setForceEraForcingNewFixture)
+  await fixtureRunner.run()
+
+  // And wait until new era happened: about 10 minutes
+  let forceEraStatus = await api.getForceEra()
+  while (forceEraStatus.isForceNew) {
+    await sleep(sleepTimeSeconds * 1000)
+    forceEraStatus = await api.getForceEra()
+  }
```
Lezek123: Based on my observations during testing what happens once you call `forceNewEra` is:
- first `staking.forceEra` is set to `ForceNew`
- once the current session ends, `staking.currentEra` is set to `1`, but `activeEra.index` remains `0` (era `1` is just "planned" at this point)
- only after another full session the rewards for era `0` are paid and `activeEra.index` changes to `1`

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // ----------- ASSERT ----------------
+
+  // 1. Nex Era Starting session index is Some
+  const { index } = (await api.getActiveEra()).unwrap()
+  const nextEraStartingSessionIndex = await api.getErasStartSessionIndex(index.addn(1) as u32)
+  assert(nextEraStartingSessionIndex.isSome, 'ext era starting session index is not some')
```
Lezek123:
```suggestion
assert(nextEraStartingSessionIndex.isSome, 'Next era starting session index is not set')
```

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // 2. Check that genesis authorities claim for era 0 is 0
```
Lezek123: So based on my comment about `forceNewEra`: the rewards for era `0` are not yet set at this point, because `activeEra.index` is still `0`. You should wait until `activeEra.index` is `1` to properly test this.

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+  const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+  const getBalances = async (accounts: string[]) => Promise.all(accounts.map((account) => api.getBalance(account)))
+
+  // constants
+  const sleepTimeSeconds = 10
+
+  const authorities = await api.getSessionAuthorities()
+  const authoritiesStash = (await Promise.all(authorities.map((account) => api.getBonded(account)))).map((x) =>
+    x.unwrap().toString()
+  )
```
Lezek123: - `await api.getSessionAuthorities()` already returns the stash account
- `staking.bonded` is a map from stash account to controller account, not the other way around

So I think this should be just:
```suggestion
  const authoritiesStash = await api.getSessionAuthorities()
```

tests/network-tests/src/flows/carthage/switchToNPoS.ts
```diff
@@ -0,0 +1,69 @@
...
+      const fixtureRunner = new FixtureRunner(claimingPayoutStakersSucceedsFixture)
+      fixtureRunner.run()
+    })
+  )
+  const currentBalances = await getBalances(authoritiesStash)
+  assert.deepEqual(previousBalances, currentBalances)
+
+  // 3. Election rounds have happened
+  const electionRounds = await api.getElectionRounds()
+  assert.isAbove(electionRounds.toNumber(), 0)
```
Lezek123: I feel like this test should have more steps, like checking whether the first era rewards work as expected etc.

.github/workflows/run-network-tests.yml
```diff
@@ -149,7 +149,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        scenario: ['full', 'setupNewChain', 'setupNewChainMultiStorage', 'bonding']
+        scenario: ['full', 'setupNewChain', 'setupNewChainMultiStorage', 'bonding', 'carthagePoA', 'carthageNPoSSwitch']
```
Lezek123: `carthagePoA` scenario doesn't exist

Lezek123 (changes requested):

tests/network-tests/src/flows/carthage/switchToNPoS.ts
dobertRowneySr: I used active era as a reference

runtime/src/lib.rs
dobertRowneySr: used active era instead

tests/network-tests/src/flows/carthage/switchToNPoS.ts
dobertRowneySr: I have updated the waiting condition to check for active era index > 0

tests/network-tests/src/flows/carthage/carthagePoAAssertions.ts
dobertRowneySr: I used the same runtime computation for the `era_duration` locks

Lezek123 (changes requested):
I fixed a few bugs in the tests and added some extra assertions, please take a look: https://github.com/ignazio-bovo/joystream/pull/8

runtime/src/constants.rs
```diff
@@ -37,7 +37,11 @@ pub const SECS_PER_BLOCK: Moment = MILLISECS_PER_BLOCK / 1000;
 pub const BONDING_DURATION: u32 = 4 * 28; // 4 * 28 eras = 28 days (since 1 era = 6h)
 pub const SLASH_DEFER_DURATION: u32 = BONDING_DURATION - 1;
 
+#[cfg(feature = "testing_runtime")] // 30 seconds sessions for faster tests
+pub const EPOCH_DURATION_IN_BLOCKS: BlockNumber = MINUTES / 2;
+#[cfg(not(feature = "testing_runtime"))]
```
Lezek123: The feature has been renamed in https://github.com/Joystream/joystream/pull/4364, so this code no longer works
```suggestion
#[cfg(feature = "testing-runtime")] // 30 seconds sessions for faster tests
pub const EPOCH_DURATION_IN_BLOCKS: BlockNumber = MINUTES / 2;
#[cfg(not(feature = "testing-runtime"))]
```

Lezek123 (changes requested):

Lezek123 (approved):

mnaamani (approved):

PR #4316, opened 2022-09-27
Increase WASM heap size to 1 GB

Lezek123:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203062945353999) by [Unito](https://www.unito.io)


bin/node/src/service.rs
```diff
@@ -163,7 +164,7 @@ pub fn new_partial(
 
     let executor = NativeElseWasmExecutor::<ExecutorDispatch>::new(
         config.wasm_method,
-        config.default_heap_pages,
+        config.default_heap_pages.or(Some(DEFAULT_HEAP_PAGES)),
```
mnaamani: I see that the substrate cli is where the config is coming from and seems by default it is passes `default_heap_pages` as `None`, so this works. Should we for guaranteeing we always set our own value here just do:

```suggestion
        Some(DEFAULT_HEAP_PAGES),
```

..in-case we update substrate and cli is updated to pass Some value which is less than what we want it to be?

mnaamani (changes requested):
Looks good, made a suggestion.
Perhaps we can also set the default in `bin/inspect/src/command.rs` so the inspect command can also benefit.

bin/node/src/service.rs
Lezek123: Both suggestions implemented in https://github.com/Joystream/joystream/pull/4316/commits/1830eff3c33741ab1351d0387682e9f7b26b4db7

mnaamani (approved):

PR #4317, opened 2022-09-27
Worker remarked post moderation

### LINKED ISSUE #3993: Query Node: Allow moderation of posts with `PostStatusLocked` via metaprotocol
Query node should support moderation of posts with `PostStatusLocked` (posts that no longer exist in the runtime, but were deleted with `hide` flag set to `false`) through a new metaprotocol message sent via forum working group's `worker_remark` extrinsic.

The query node will have to:

* check whether the moderator that sent the remark has access to moderate the category that the post currently belong to in the database (`post->thread->category`). Note that you also need to recursively check `category->parent`, because if the moderator has access to the parent category, it should also be allowed to moderate posts under its child categories
* update post status to `PostStatusModerated` if the moderator is allowed to moderate the post

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202554996938921) by [Unito](https://www.unito.io)

thesan:
Closes #3993

### DIFF
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -59,7 +59,7 @@ message WorkingGroupMetadataAction {
 }
 
 message ModeratePost {
-  optional uint64 post_id = 1;
+  required uint64 post_id = 1;
   optional string rationale = 2;
 }
 
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -57,3 +57,14 @@ message WorkingGroupMetadataAction {
     RemoveUpcomingOpening remove_upcoming_opening = 3;
   }
 }
+
+message ModeratePost {
+  optional uint64 post_id = 1;
+  optional string rationale = 2;
+}
+
+message RemarkMetadataAction {
+  oneof action {
+    ModeratePost moderate_post = 1;
+  }
+}
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -60,7 +60,7 @@ message WorkingGroupMetadataAction {
 
 message ModeratePost {
   required uint64 post_id = 1;
-  optional string rationale = 2;
+  required string rationale = 2;
 }
 
 message RemarkMetadataAction {
--- a/query-node/mappings/src/forum.ts
+++ b/query-node/mappings/src/forum.ts
@@ -1,13 +1,14 @@
 /*
 eslint-disable @typescript-eslint/naming-convention
 */
-import { EventContext, StoreContext, DatabaseManager } from '@joystream/hydra-common'
+import { EventContext, StoreContext, DatabaseManager, SubstrateEvent } from '@joystream/hydra-common'
 import {
   bytesToString,
   deserializeMetadata,
   genericEventFields,
   getWorker,
   inconsistentState,
+  invalidMetadata,
   perpareString,
   toNumber,
 } from './common'
@@ -80,6 +81,56 @@ async function getPost(store: DatabaseManager, postId: string, relations?: 'thre
   }
 
   return post
+}
+
+type ModerationType = 'runtime' | 'leadRemark' | 'workerRemark'
+
+async function canModerate(store: DatabaseManager, actorId: string, categoryId: string): Promise<boolean> {
+  const category = await getCategory(store, categoryId, ['moderators'])
+  return (
+    category.moderators.some(({ id }) => id === actorId) ||
+    (category.parentId ? canModerate(store, actorId, category.parentId) : false)
+  )
+}
+
+export async function moderatePost(
+  store: DatabaseManager,
+  event: SubstrateEvent,
+  type: ModerationType,
+  postId: string,
+  actor: Worker,
+  rationale: string
+): Promise<void> {
+  const post = await getPost(store, postId, ['thread'])
+
+  if (type !== 'runtime' && post.status.isTypeOf !== 'PostStatusLocked') {
+    return invalidMetadata(
+      `Trying to moderate forum post ${post.id}, with status ${post.status.isTypeOf}. Only locked forum post can be moderated with remark`
+    )
+  }
+  if (type === 'workerRemark' && !(await canModerate(store, actor.id, post.thread.categoryId))) {
+    return invalidMetadata(`The worker ${actor.id} can not moderate the post: ${post.id}`)
+  }
+
+  const postModeratedEvent = new PostModeratedEvent({
+    ...genericEventFields(event),
+    actor,
+    post,
+    rationale,
+  })
+
+  await store.save<PostModeratedEvent>(postModeratedEvent)
+
+  const newStatus = new PostStatusModerated()
+  newStatus.postModeratedEventId = postModeratedEvent.id
+
+  post.status = newStatus
+  post.isVisible = false
+  await store.save<ForumPost>(post)
+
+  const { thread } = post
+  --thread.visiblePostsCount
+  await store.save<ForumThread>(thread)
 }
 
 async function getActorWorker(store: DatabaseManager, actor: PrivilegedActor): Promise<Worker> {
@@ -463,27 +514,7 @@ export async function forum_CategoryMembershipOfModeratorUpdated({
 export async function forum_PostModerated({ event, store }: EventContext & StoreContext): Promise<void> {
   const [postId, rationaleBytes, privilegedActor] = new Forum.PostModeratedEvent(event).params
   const actorWorker = await getActorWorker(store, privilegedActor)
-  const post = await getPost(store, postId.toString(), ['thread'])
-
-  const postModeratedEvent = new PostModeratedEvent({
-    ...genericEventFields(event),
-    actor: actorWorker,
-    post,
-    rationale: bytesToString(rationaleBytes),
-  })
-
-  await store.save<PostModeratedEvent>(postModeratedEvent)
-
-  const newStatus = new PostStatusModerated()
-  newStatus.postModeratedEventId = postModeratedEvent.id
-
-  post.status = newStatus
-  post.isVisible = false
-  await store.save<ForumPost>(post)
-
-  const { thread } = post
-  --thread.visiblePostsCount
-  await store.save<ForumThread>(thread)
+  await moderatePost(store, event, 'runtime', postId.toString(), actorWorker, bytesToString(rationaleBytes))
 }
 
 export async function forum_PostTextUpdated({ event, store }: EventContext & StoreContext): Promise<void> {
--- a/query-node/mappings/src/forum.ts
+++ b/query-node/mappings/src/forum.ts
@@ -74,7 +74,7 @@ async function getThread(store: DatabaseManager, threadId: string): Promise<Foru
   return thread
 }
 
-async function getPost(store: DatabaseManager, postId: string, relations?: 'thread'[]): Promise<ForumPost> {
+async function getPost(store: DatabaseManager, postId: string, relations?: string[]): Promise<ForumPost> {
   const post = await store.get(ForumPost, { where: { id: postId }, relations })
   if (!post) {
     throw new Error(`Forum post not found by id: ${postId.toString()}`)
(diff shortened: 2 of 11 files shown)


metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -57,3 +57,14 @@ message WorkingGroupMetadataAction {
     RemoveUpcomingOpening remove_upcoming_opening = 3;
   }
 }
+
+message ModeratePost {
+  optional uint64 post_id = 1;
+  optional string rationale = 2;
```
zeeshanakram3: Any particular reason why these fields are marked `optional` and not `required`. If these fields were mandatory then you would not have to check the possibility of undefined in the [mappings](https://github.com/thesan/joystream/blob/cf06f473330110c157e5c18bc2886f3a00428b58/query-node/mappings/src/workingGroups.ts#L664-L665)

query-node/schemas/forum.graphql
```diff
@@ -20,6 +20,7 @@ type ForumCategory @entity {
   id: ID!
 
   "Parent category (if none - this is a root category)"
+  parentId: ID
```
zeeshanakram3: Is this field added for doing nested filtering, if yes please mark this field with the [comment](https://github.com/thesan/joystream/blob/cf06f473330110c157e5c18bc2886f3a00428b58/query-node/schemas/contentEvents.graphql#L24), it will help to remove all the redundant fields once nested filtering is available.

query-node/schemas/forum.graphql
```diff
@@ -75,6 +76,7 @@ type ForumThread @entity {
   author: Membership!
 
   "Category the thread belongs to"
+  categoryId: ID!
```
zeeshanakram3: Is this field added for doing nested filtering, if yes please mark this field with the [comment](https://github.com/thesan/joystream/blob/cf06f473330110c157e5c18bc2886f3a00428b58/query-node/schemas/contentEvents.graphql#L24), it will help to remove all the redundant fields once nested filtering is available.

metadata-protobuf/proto/WorkingGroups.proto
thesan: Right, at least making `post_id` required makes more sense. But since AFAIK the runtime doesn't check `metadata`, the QN should still check whether they are defined

query-node/schemas/forum.graphql
thesan: I think I got confused here :zany_face:. This is only used in the mappings [canModerate](https://github.com/Joystream/joystream/blob/cf06f473330110c157e5c18bc2886f3a00428b58/query-node/mappings/src/forum.ts#L89) so there is no need to add this to the schema

query-node/schemas/forum.graphql
thesan: Same thing here...

thesan:
@zeeshanakram3 thank you for your review I addressed your CRs

query-node/schemas/forum.graphql
thesan: Never mind these needs to be exposed in schema apparently... my understanding of hydra is very limited!
So I just added nicer comments in the schema but solving nested filtering won't affect these.

zeeshanakram3: I am not sure why that is necessary, in fact this is redundant. When `ForumCategory` record is stored in DB, for the field `parent: ForumCategory`, only relation to the `ForumCategory` table is stored in the DB in the form of renamed `parent_Id` field, as you can see in the Postgres schema definition below.

```
CREATE TABLE forum_category (
    ....
    parent_id character varying REFERENCES forum_category(id),
   ....
);
```

query-node/schemas/forum.graphql
zeeshanakram3: Please see https://github.com/Joystream/joystream/pull/4317#discussion_r999076753

query-node/mappings/src/workingGroups.ts
```diff
@@ -648,6 +652,50 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
+
+  const metadata = deserializeMetadata(RemarkMetadataAction, metadataByte)
+  if (metadata?.moderatePost) {
+    if (group.name !== 'forumWorkingGroup') {
+      return invalidMetadata(`The ${group.name} is incompatible with the remarked moderatePost`)
+    }
+    const { postId, rationale } = metadata?.moderatePost ?? {}
+    if (!postId) {
+      return invalidMetadata('No valid forum post id was provided')
+    }
```
zeeshanakram3: This validation is not needed anymore, since in protobuf schema, the `post_id` field is now required (which means validation will be done by the protobuf serialization logic).

query-node/mappings/src/workingGroups.ts
```diff
@@ -648,6 +652,50 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
```
zeeshanakram3: This validation is not needed anymore, since in protobuf schema, the post_id field is now required (which means validation will be done by the protobuf serialization logic).

metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -57,3 +57,14 @@ message WorkingGroupMetadataAction {
     RemoveUpcomingOpening remove_upcoming_opening = 3;
   }
 }
+
+message ModeratePost {
+  required uint64 post_id = 1;
+  optional string rationale = 2;
```
zeeshanakram3: I think `rationale` should also be `required`, since in `PostModeratedEvent` [entity](https://github.com/thesan/joystream/blob/096ae678145fb249bf024956abdb3be82d1fb9d9/query-node/schemas/forumEvents.graphql#L288) it is defined as non-nullable field. So maybe it's consistent to have `rationale` in protobuf schema as required too.

thesan: Right! I missed that

query-node/schemas/forum.graphql
thesan: That's what I was thinking last Friday but it looks like the `ForumCategory` type in `query-node/generated/graphql-server/dist/src/modules/forum-category/forum-category.model.d.ts` is generated from the schema.

Here are the type errors: https://github.com/Joystream/joystream/actions/runs/3252154043/jobs/5338037876

The CI reported on the commit with this changes: https://github.com/Joystream/joystream/pull/4317/commits/ba8a8e75013e8220efe4c6f6ea256a4216325218#diff-0edeabbbad7f1c259beb22e445370523d09b886b184534cdd8e8bd1ffdb5c448

Or is there another way to access these ids ?

zeeshanakram3: > Or is there another way to access these ids?

I believe you are talking about this piece of code, if yes

```ts
async function canModerate(store: DatabaseManager, actorId: string, categoryId: string): Promise<boolean> {
  const category = await getCategory(store, categoryId, ['moderators'])
  return (
    category.moderators.some(({ id }) => id === actorId) ||
    (category.parentId ? canModerate(store, actorId, category.parentId) : false)
  )
}
```
Then you can actually get the id using parent relation, i.e., `parent.id`

zeeshanakram3 (approved):
LGTM

zeeshanakram3 (approved):

zeeshanakram3 (approved):

tests/network-tests/src/fixtures/forum/RemarkModeratePostsFixture.ts
```diff
@@ -0,0 +1,96 @@
...
+      const qEvent = this.findMatchingQueryNodeEvent(e, qEvents)
+      Utils.assert(qPost.status.__typename === 'PostStatusModerated', 'Invalid post status')
+      Utils.assert(qPost.status.postModeratedEvent, 'Query node: Missing PostModeratedEvent ref')
+      assert.equal(qPost.status.postModeratedEvent.id, qEvent.id)
+      assert.equal(qPost.isVisible, false)
+    })
+  }
+
+  protected assertQueryNodeEventIsValid(qEvent: PostModeratedEventFieldsFragment, i: number): void {
+    const { postId, asWorker, rationale } = this.moderations[i]
```
Lezek123, line 75: Since there will be no `qEvent` if `expectFailure === true`, I think this should be:
```suggestion
    const { postId, asWorker, rationale } = this.moderations[i].filter((m) => !m.expectFailure)
```

tests/network-tests/src/flows/forum/posts.ts
```diff
@@ -144,6 +161,21 @@ export default async function threads({ api, query }: FlowProps): Promise<void>
   // Run compound query node checks
   await Promise.all([updatePostsTextRunner.runQueryNodeChecks(), deletePostsRunner.runQueryNodeChecks()])
 
+  // Moderate removed posts
+  const postModerations = [
```
Lezek123, line 189: I think it would be useful to have other test cases here like:
- Trying to moderate as worker w/o category permissions
- Trying to moderate as non-forum worker
- Optionally: Trying to moderate as worker with permissions only to `category.parent.parent`

metadata-protobuf/proto/WorkingGroups.proto
Lezek123: We used to have all fields set to `optional` as it was considered best practice for protobuf messages (in terms of backwards compatibility) and starting from `proto3` it is no longer even possible to mark a field as `required`.

Lezek123:
During my local tests I didn't encounter any issues, everything seems to work as expected :+1:

Lezek123 (approved):
Approving, since the changes I requested in the previous review are non-critical and my local tests passed.

tests/network-tests/src/fixtures/forum/RemarkModeratePostsFixture.ts
thesan: I think I did this at first but then (if I remember correctly) it would failed because `i` comes from [tests/network-tests/src/Fixture.ts](https://github.com/Joystream/joystream/pull/4317/files/b04b048f623422d36171d54bbf5da00d7a80b9aa#diff-a69963794564bd61b084b85a260e2642dc7faa6a6d6bee805dfbb6e426306e64) which includes `expectFailure` but then skip them. so e.g: if `this.moderations[1]` is `expectFailure` then `assertQueryNodeEventIsValid` will be called with `i: 0, 2, ...` but not `1`

tests/network-tests/src/flows/forum/posts.ts
thesan: You're right, I will try to quickly add some of these tomorrow

thesan:
> Approving, since the changes I requested in the previous review are non-critical and my local tests passed.

Thanks

tests/network-tests/src/fixtures/forum/RemarkModeratePostsFixture.ts
Lezek123: You're right, my bad :+1:

query-node/mappings/src/workingGroups.ts
```diff
@@ -671,6 +675,42 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
+
+  const metadata = deserializeMetadata(RemarkMetadataAction, metadataByte)
+  if (metadata?.moderatePost) {
+    if (group.name !== 'forumWorkingGroup') {
+      return invalidMetadata(`The ${group.name} is incompatible with the remarked moderatePost`)
+    }
+    const { postId, rationale } = metadata.moderatePost
+    const actor = await getWorkingGroupLead(store, group.name)
+
+    await moderatePost(store, event, 'leadRemark', postId, actor, rationale)
```
mnaamani, line 690: if a lead provides an invalid postId and moderatePost throws, will this not kill the processor? Since the postId is not validated by the runtime because it is in the metadata..
expanding:
`moderatePost()` calls `getPost()` and then checks details of post if it is locked, it is getPost which will actually throw if post is not found. So moderate should wrap that invocation with try/catch and return invalidMetadata()

query-node/mappings/src/forum.ts
```diff
@@ -82,6 +83,56 @@ async function getPost(store: DatabaseManager, postId: string, relations?: 'thre
...
+
+export async function moderatePost(
+  store: DatabaseManager,
+  event: SubstrateEvent,
+  type: ModerationType,
+  postId: string,
+  actor: Worker,
+  rationale: string
+): Promise<void> {
+  const post = await getPost(store, postId, ['thread', 'thread.category'])
```
mnaamani: ref my other https://github.com/Joystream/joystream/pull/4317#discussion_r1059862100. call to getPost here can throw if postId is not found. So wrap this with try/catch and return `invalidMetadata()` on error.

mnaamani (changes requested):
Small bug fix needed. It might make more sense to wrap the call to moderatePost from the workingGroups handlers rather than in the getPost which, since the moderatePost in the forum module is called from the postModerated event handler where the postId is "valid" since it came from the runtime.

mnaamani (approved):

PR #4318, opened 2022-09-27
Add (sudo) Membership::create_member() dispatchable

mnaamani:
Renamed and Refactored create_founding_member to also work for creating regular member with sudo.
This will be used in bootstrapping members on chain launch. I explicitly didn't rename the event FoundingMemberCreated to minimize refactoring in query-node and integration tests.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203060548584919) by [Unito](https://www.unito.io)

PR #4319, opened 2022-09-28
Member migration-scripts

mnaamani:
Member migration scripts updated for carthage bootstrapping. Builds on https://github.com/Joystream/joystream/pull/4318

Howto use:
Export Membership from olympia:

```bash
yarn migration-scripts olympia-carthage:createMembershipsSnapshot \
    -o snapshot.json --queryNodeUri=https://query.joystream.org/graphql
```

Prepare a founding members list, JSON array of member ids, in a file: `fm.json`
Not this is required, so even if you don't want any FM, you must still provide the argument, and empty array in the file. (Just a precaution to avoid missing this step when migrating)

```json
[
   20,
   6,
   11,
   100
]
```

Inject members into new chain:

```bash
yarn migration-scripts olympia-carthage:migrateMembers \
     --snapshotFilePath snapshot.json --foundingMembersFilePath fm.json
```

For additional args check help:

```bash
yarn migration-scripts olympia-carthage:createMembershipsSnapshot --help
```

```bash
yarn migration-scripts olympia-carthage:migrateMembers --ehlp
```

While testing I observed one issue, I can't produce the correct hash in terminal with `b2sum` tool.

eg. for handle "bwhm0" hash on command line:

```bash
echo "bwhm0" | b2sum -l 256
cac4f8d06aef942c3eca94287134de89e9e96fde8820c1cc014f7f970c927096
```

But on chain it is `0x07f3f19cd189d7c438cfb0085b5e14a37f2748c1d6aee19880a448b8ce81ddd2`
This is not migration issue, the hash is same in chain where we exported from, but just wondering what am I missing? cc: @Lezek123 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203062615978274) by [Unito](https://www.unito.io)


runtime-modules/membership/src/lib.rs
```diff
@@ -397,7 +400,8 @@ decl_event! {
         StakingAccountRemoved(AccountId, MemberId),
         StakingAccountConfirmed(AccountId, MemberId),
         MemberRemarked(MemberId, Vec<u8>),
-        FoundingMemberCreated(MemberId, CreateFoundingMemberParameters, u32),
+        FoundingMemberCreated(MemberId, CreateMemberParameters, u32),
+        MemberCreated(MemberId, CreateMemberParameters, u32),
```
Lezek123: I think it would be nice to get rid of the redundancy here, as it would simplify a lot of the code. `MemberCreated` event already includes `CreateMemberParameters` which include `is_founding_member` property, so there's no need to have a separate `FoundingMemberCreated` event.

runtime-modules/membership/src/tests/fixtures.rs
```diff
@@ -857,3 +858,72 @@ impl CreateFoundingMemberFixture {
         }
     }
 }
+
+pub struct CreateMemberFixture {
```
Lezek123: I think `CreateFoundingMemberFixture` and `CreateMemberFixture` should be just one fixture with customizable `is_founding_member` parameter

tests/network-tests/src/fixtures/membership/utils.ts
```diff
@@ -13,10 +13,14 @@ type MemberCreationParams = {
...
   externalResources?: MembershipMetadata.IExternalResource[] | null
   metadata: Bytes
+  is_founding_member: boolean
 }
 
 // Common code for Membership fixtures
-export function generateParamsFromAccountId(accountId: string): MemberCreationParams {
+export function generateParamsFromAccountId(
+  accountId: string,
+  isFoundingMember: boolean = false // eslint-disable-line @typescript-eslint/no-inferrable-types
```
Lezek123: Is there really a need to disable this rule?
```suggestion
  isFoundingMember = false
```

tests/network-tests/src/flows/membership/creatingMembers.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default async function creatingMembers({ api, query, env }: FlowProps): Promise<void> {
+  const debug = extendDebug('flow:creating-founding-members')
+  debug('Started')
+  api.enableDebugTxLogs()
+
+  const N: number = +env.MEMBERSHIP_CREATION_N!
+  assert(N > 0)
+
+  // Assert membership can be bought if sufficient funds are available
```
Lezek123: Invalid comment (similar mistake was made by me in `tests/network-tests/src/flows/membership/creatingFoundingMembers.ts`)

tests/network-tests/src/scenarios/full.ts
```diff
@@ -49,7 +49,7 @@ scenario('Full', async ({ job, env }) => {
   // All other jobs should be executed after coreJob
 
   // Membership:
-  job('creating members', creatingMemberships).after(coreJob)
+  job('buying members', buyingMemberships).after(coreJob)
```
Lezek123, line 54: Just noticed that `full` scenario is missing the new `creatingFoundingMembers`/`creatingMembers` flows

utils/migration-scripts/src/olympia-carthage/BaseMigration.ts
```diff
@@ -0,0 +1,204 @@
...
+   * the result of any of the previous calls associated with that entity migration.
+   * (This means, for example, that regardless of whether balnces.transferKeepAlive failed and interrupted the batch or balnces.transferKeepAlive
+   * succeeded, but sudo.sudoAs failed - in both cases the migration is considered failed and should be fully re-executed on
+   * the next script run)
+   */
+  protected extractFailedMigrations<T extends { id: string }>(
+    result: SubmittableResult,
+    batch: T[],
+    callsPerEntity = 1,
+    usesSudoAs = true,
```
Lezek123: I think this parameter is no longer needed?

utils/migration-scripts/src/olympia-carthage/olympia-query-node/queries/queries.graphql
```diff
@@ -0,0 +1,40 @@
...
+    name
+    about
+    # We will only support avatarUri to bootstrap, since there will be no storage
+    # infra to upload data objects..
+    avatar {
+      ... on AvatarUri {
+        avatarUri
+      }
+    }
+    # Was this properly implemented on olympia? Is there any valuable info there?
```
Lezek123: It looks like there is only 1 member who has those set currenly (`4148`) according to https://query.joystream.org/graphql

utils/migration-scripts/src/olympia-carthage/olympia-query-node/queries/queries.graphql
```diff
@@ -0,0 +1,40 @@
...
+    # externalResources {
+    #   value
+    # }
+  }
+  controllerAccount
+  rootAccount
+  # isFoundingMember # Is this even populated?
```
Lezek123: It's just `false` for all members, it was added just as kind of a "reminder" that we will need it in the future

utils/migration-scripts/src/commands/olympia-carthage/migrateMembers.ts
```diff
@@ -0,0 +1,57 @@
...
+    snapshotFilePath: flags.string({
+      required: true,
+      description: 'Path to olympia memberships snapshot (json)',
+    }),
+    foundingMembersFilePath: flags.string({
+      required: true,
+      description: 'Path to list of founding member ids (json)',
+    }),
+    wsProviderEndpointUri: flags.string({
+      description: 'WS provider endpoint uri (Olympia)',
```
Lezek123:
```suggestion
      description: 'WS provider endpoint uri (Carthage)',
```

Lezek123 (changes requested):
Looks good, added a few small suggestions.

There is also a small issue with some handles that have non-standard characters in them. For some reason they seem to become hex-encoded once migrated:
somethingwrong

somethingwrong2

utils/migration-scripts/src/olympia-carthage/olympia-query-node/queries/queries.graphql
mnaamani: Yes I assumed so, that is why I commented out in the query.

utils/migration-scripts/src/commands/olympia-carthage/migrateMembers.ts
mnaamani: will update

utils/migration-scripts/src/olympia-carthage/olympia-query-node/queries/queries.graphql
mnaamani: Yes I think the reason not many members have set this info is because the feature was disabled from pioneer probably because it was not working as expected. Only one external resource can be set as apposed to an array.

runtime-modules/membership/src/lib.rs
mnaamani: I started with that initially, and when I realized how many files I had to touch just to handle renamed event name I reverted. But the simplification is worth it I guess.

runtime-modules/membership/src/tests/fixtures.rs
mnaamani: Yes, I avoided also just to limit refactoring scope, but sure why not :)

tests/network-tests/src/scenarios/full.ts
mnaamani: will add

utils/migration-scripts/src/olympia-carthage/BaseMigration.ts
mnaamani: correct, will remove

utils/migration-scripts/src/olympia-carthage/olympia-query-node/queries/queries.graphql
mnaamani: yep confirmed by pioneer team.

mnaamani:
Applied all requested changes.
Only issue left is the one you raised - "There is also a small issue with some handles that have non-standard characters in them. For some reason they seem to become hex-encoded once migrated:" I'll try to look into it.

mnaamani:
> There is also a small issue with some handles that have non-standard characters in them. For some reason they seem to become hex-encoded once migrated: somethingwrong
> 
> somethingwrong2

Caused by: "https://github.com/Joystream/joystream/commit/0053f34308b981a36417cd5116b132a75728a45e. Handle used to have a type Text when we were still using our custom type definitions, but now that the type defs come from the chain metadata, it became Bytes. If you use Bytes.toHuman() the method tries to guess whether it's a human-readable string or just some binary blob. I guess in case it contains some non-standard characters, it decides it's probably a binary blob and returns a hex string instead. One way to solve it would be to use bytesToString(handle) instead of handle.toHuman()  in the QN membership mappings" - @Lezek123

Lezek123 (approved):
The script works as expected, however I noticed that there are some issues w/ the query node if a high `--batchSize` is used.

When I used the value of `1000` I got following error in the processor:
```
INFO: Last block: 10 	: 187 blocks behind
INFO: Starting the event queue
INFO: Shutting down...
ERROR: {
  "response": {
    "errors": [
      {
        "message": "Invalid string length",
        "extensions": {
          "code": "INTERNAL_SERVER_ERROR"
        }
      }
    ],
    "status": 500,
    "headers": {}
  },
```

It happens when the processor sends a query to the indexer to retrieve the latest events. It uses a limit of `2000` on the query results and this is probably too much, which causes `JSON.stringify` to fail. Requires further investigation as a separate issue.

Issue #3880, opened 2022-06-08
Scope of Work in Channel Payouts Feature (Ephesus Release)


zeeshanakram3:
This issue tracks the scope of all work (including completed + remaining) for Channel Payouts Feature (Ephesus Release)

#### Scope
* [x] Renaming stuff (change command names from `creator*` to `channel*`)
* [x] Add CLI support for Channel payouts proposal
* [x] `content:creatorPayoutPayloadHeader` should both take remote URL and file path for fetching header
* [x]  Update Channel Payouts CLI commands docs 

## Query-Node
* #### Tracking PR
* https://github.com/Joystream/joystream/issues/3879

#### Scope (requires updating the old PR )
* [x] Renaming stuff (change function names from `creator*` to `channel*`); also apply the similar change to `metadata-protobuf` schemas
* [x] Update Channel Payouts utility function's usage docs 

#### Tracking PR
* https://github.com/Joystream/joystream/pull/3182

#### Tracking issue
* https://github.com/Joystream/joystream/issues/3828

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202453881030481) by [Unito](https://www.unito.io)

bedeho:
Thanks for this superb summary, hopefully this will make it reasonably easy to pick up the work again.

PR #3879, opened 2022-06-08
initial work on channel payouts mappings & proposals

zeeshanakram3:
implements 

* #3264 
* #3613 
* #3614
* #3615

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202417814816108) by [Unito](https://www.unito.io)

PR #4334, opened 2022-10-02
Channel payouts: Mappings, CLI, Integration-tests, TS lib

zeeshanakram3:
supersedes #3879 & #3182

Tracking issue #3880

- Channel Payouts Payload [format](https://github.com/Joystream/joystream/issues/3163#issuecomment-1029854088) 
- Channel Payouts [Protobuf schema](https://github.com/zeeshanakram3/joystream/blob/39d657e8db01b184f9978d972e01df591e04ae80/metadata-protobuf/proto/ChannelPayouts.proto)


joystreamjs/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "private": true,
+  "name": "joystreamjs",
```
kdembler: I think we should follow the current naming scheme and use `@joystream/*`. Maybe it's a single package called `@joystream/js` and then the rest is imported from its submodules? Not sure

joystreamjs/content/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "importHelpers": true,
+    "module": "commonjs",
```
kdembler: We should either make this use ES modules or be a hybrid CJS/ESM package with both types of import available. ESM is gonna replace CJS pretty quickly IMO and a lot of new tooling requires ESM. Atlas uses only ESM packages so if we are forced to use CJS dependency, we need to jump through hoops to make it properly work with our setup (we do that with the types package)

joystreamjs/content/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "noUnusedLocals": true,
+    "baseUrl": ".",
+    "paths": {
+      "@polkadot/types/augment": [
+        "../types/augment-codec/augment-types.ts"
+      ],
+      "@polkadot/api/augment": [
+        "../types/augment-codec/augment-api.ts"
+      ]
+    },
```
kdembler: I think this shouldn't be needed with the updated types library? 
```ts
// load type augments
import '@joystream/types'
```
In one of the source files should load the augments automatically

joystreamjs/utils/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@joystreamjs/utils",
+  "description": "common joystream utilities",
+  "version": "1.0.0",
+  "author": "Jsgenesis",
+  "bugs": "https://github.com/Joystream/joystream/issues",
+  "dependencies": {
+    "@joystream/metadata-protobuf": "^2.0.0",
+    "@joystream/types": "^0.19.3",
```
kdembler: These are a bit outdated, let's use the newest versions available at the time:
```
    "@joystream/metadata-protobuf": "2.5.0",
    "@joystream/types": "0.20.3",
```
Also I don't think we should minor version range selector (`^`) since both of those libraries can have breaking API changes between minor versions

kdembler: It may also be a good argument for having a single package with submodules so that we don't need to manage dependencies for all of them separately (in `content` you use different versions)

joystreamjs/utils/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@polkadot/types": "5.9.1"
```
kdembler: This should already be in the dependency tree of `@joystream/types`

joystreamjs/content/src/channelPayouts.ts
```diff
@@ -0,0 +1,283 @@
+import { ChannelPayoutsMetadata, IChannelPayoutsMetadata } from '@joystream/metadata-protobuf'
+import { createType } from '@joystream/types'
+import { asValidatedMetadata, readBytesFromFile, ReadFileContext } from '@joystreamjs/utils'
+import { ChannelPayoutProof } from '@joystreamjs/utils/typings/ChannelPayoutsPayload.schema'
+import { ChannelPayoutsVector } from '@joystreamjs/utils/typings/ChannelPayoutsVector.schema'
+import { u8aToHex } from '@polkadot/util'
+import { blake2AsU8a } from '@polkadot/util-crypto'
+import BN from 'bn.js'
+import Long from 'long'
```
kdembler: I was wondering, why do we use both `Long` and `BN`? I thought they are used for the same thing?

cli/src/commands/content/claimChannelReward.ts
```diff
@@ -0,0 +1,90 @@
...
+    })
+
+    const merkleBranch: ProofElement[] = []
+    payoutProof.merkleBranch.forEach((m) => {
+      const proofElement = createType('PalletCommonMerkleTreeProofElementRecord', {
+        hash_: u8aToHex(Buffer.from(m.hash, 'hex')),
+        side: m.side ? { Right: null } : { Left: null },
+      })
+      merkleBranch.push(proofElement)
+    })
```
kdembler: This will need to be replicated by all users of the library right? If yes, then maybe this functionality should be in the library itself?

zeeshanakram3: Sure. I will add this code block to the lib itself

joystreamjs/content/tsconfig.json
zeeshanakram3: You are right, the originally this file was created way before the Carthage release and I forgot to update it

joystreamjs/package.json
zeeshanakram3: Sure, sounds good.

cli/src/commands/content/claimChannelReward.ts
zeeshanakram3: Addressed in [e2fba88](https://github.com/Joystream/joystream/pull/4334/commits/e2fba88bc2668673431cf4abc81eed6e1eea19f6)

joystreamjs/utils/package.json
zeeshanakram3: Addressed in [e2fba88](https://github.com/Joystream/joystream/pull/4334/commits/e2fba88bc2668673431cf4abc81eed6e1eea19f6)

joystreamjs/utils/package.json
zeeshanakram3: Addressed in [e2fba88](https://github.com/Joystream/joystream/pull/4334/commits/e2fba88bc2668673431cf4abc81eed6e1eea19f6)

joystreamjs/content/tsconfig.json
zeeshanakram3: Addressed in [e2fba88](https://github.com/Joystream/joystream/pull/4334/commits/e2fba88bc2668673431cf4abc81eed6e1eea19f6)

joystreamjs/package.json
zeeshanakram3: Addressed in [e2fba88](https://github.com/Joystream/joystream/pull/4334/commits/e2fba88bc2668673431cf4abc81eed6e1eea19f6)

Now `@joystream/js` is a single package, having submodule exports like `@joystream/js/content` & `@joystream/js/utils`

joystreamjs/content/tsconfig.json
zeeshanakram3: I tried converting `@joystream/js` to ES module, but unfortunately it didn't work quite well with other monorepo packages whom its a dependency, e.g. CLI. Since, `@joystream/cli` is a commonJS module while packing it I was getting weird non-existing module resolution errors

Will have to look more into this otherwise, I will try to make this package hybrid

joystreamjs/content/src/channelPayouts.ts
zeeshanakram3: They aren't really the same thing, As you know, JS numbers are 53 bits long (not big enough), `Long` provides some more bits for integer representation, it represents numbers 64 bits 2's complements integer, and the protobuf JS library decodes `uint64` fields to `Long` types by default. Lastly, `BN` is used to represent arbitrary large numbers, and it works really well with the PolkadotJS as `u64` and `u128` in javascript exist in BN representation

joystreamjs/content/tsconfig.json
kdembler: @zeeshanakram3 any update on this issue? If doing hybrid is super hard, we can probably live with CJS but it would be much easier for us to have ESM available

cli/README.md
```diff
@@ -110,6 +110,7 @@ When using the CLI for the first time there are a few common steps you might wan
 - [`joystream-cli content:deleteVideo`](#joystream-cli-contentdeletevideo)
 - [`joystream-cli content:deleteVideoAsModerator`](#joystream-cli-contentdeletevideoasmoderator)
 - [`joystream-cli content:deleteVideoAssetsAsModerator`](#joystream-cli-contentdeletevideoassetsasmoderator)
+- [`joystream-cli content:getPayoutsOnchainCommitment`](#joystream-cli-contentgetpayoutsonchaincommitment)
```
kdembler, line 119: @zeeshanakram3 It seems more commands should be documented here? It misses generating the payload file, etc

zeeshanakram3: Yor are right. I have updated the docs

kdembler: Hmm, it seems that some formatting got updated but the README is still missing documentation on new commands? Only `content:getPayoutsOnchainCommitment` is available, shouldn't README also include commands for generating the payload or sending the payouts proposal?

joystreamjs/content/tsconfig.json
zeeshanakram3: > we can probably live with CJS but it would be much easier for us to have ESM available

 Yeah you are right about the benefit, but I think the cost is that after conversion to ESM, the packages(`cli`/`network-tests`) that use `@joystream/js` as dependency will have to change their importing/module-resolution behaviour.
 Anyways, I did some preliminary work but didn't fully pursue it. So please let me if this is absolutely required, I will start working on it.

cli/README.md
zeeshanakram3: Yeah, my bad. The CLI documentation in the readme is autogenerated by running `yarn workspace @joystream/cli oclif-dev readme` using the file name and commands description from the build artifacts. However, previously before running this command, I  forgot to build the cli package, so the old build artifacts were used.  :rofl:

joystreamjs/content/tsconfig.json
zeeshanakram3: Addressed in [9158a05](https://github.com/Joystream/joystream/pull/4334/commits/9158a05f44af39fa371da9052a5579c0ad18d880)

mnaamani:
@zeeshanakram3 as discussed I'm updating ephesus branch here https://github.com/Joystream/joystream/pull/4430 Once that is merged you can re-target this PR to ephesus

kdembler (approved):
Thanks for making the package hybrid! We verified that it works properly with Atlas now, without the need for patching which is great. LGTM from me and @drillprop 

There's some failing test though, not sure what that is about

mnaamani:
> There's some failing test though, not sure what that is about

I'll re-run them, but a proper fix for this occasional failure is in carthage branch now

kdembler:
btw, @mnaamani @zeeshanakram3, Atlas team will need those new libraries published so that we can consume them and start developing. That's new version of `@joystream/metadata-protobuf` and new library `@joystream/js`

mnaamani:
@zeeshanakram3 I retargeted the PR to ephesus branch. Perhaps you can rebase PR on ephesus just to be sure its all good.

metadata-protobuf/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/metadata-protobuf",
-  "version": "2.5.0",
+  "version": "2.6.0",
```
mnaamani, line 3: Lets bump this to `v3.0.0` to allow for version updates on carthage/mainnet in the `2.x.x` range.
And of course update the dependent packages.

mnaamani (changes requested):
Please update PR from ephesus branch and bump metadata-protobuf package version to 3.0.0

metadata-protobuf/package.json
zeeshanakram3: Sure. btw moving forward, wouldn't it be a better solution to have something like `2.6.0-beta` for pre-production /testing releases instead of reserving some range for a specific release, i.e., `2.x.x` range for carthage/mainnet. This way, devs can be more open to following semantic versioning guidelines properly, i.e., bumping major versions too in case of breaking changes/new features

mnaamani: ok makes sense.

zeeshanakram3: Sorry, but I haven't made the version bump change yet; I was just confirming if it's fine to do it this way before making the change. So will you reopen this PR, Or should I create a separate PR including the version change, i.e., setting `@joystream/metadata-protobuf` version to `2.6.0-beta`
@mnaamani

PR #4336, opened 2022-10-04
Enable call size test

mnaamani:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203132769408579) by [Unito](https://www.unito.io)


runtime/src/runtime_api.rs
```diff
@@ -459,16 +459,54 @@ mod tests {
             .fold(0, |acc, x| acc.checked_add(*x).unwrap());
     }
 
-    #[ignore]
     #[test]
     fn call_size() {
-        let size = core::mem::size_of::<Call>();
+        // Target limit - substrate frame pallets only Call size = 208
+        const LIMIT: usize = 120;
```
Lezek123: Why is `120` a target limit, given that so many of our pallets are already above it?
Shouldn't we use, for example, a number that we know is safe given our current wasm heap size?

runtime/src/runtime_api.rs
```diff
@@ -459,16 +459,54 @@ mod tests {
             .fold(0, |acc, x| acc.checked_add(*x).unwrap());
     }
 
-    #[ignore]
     #[test]
     fn call_size() {
```
Lezek123, line 463: `bounty::Call` is missing in the list
Perhaps we should iterate over `AllPalletsWithSystem` if possible or at least add a sanity `core::mem::size_of::<Call>();` check of a "global" `Call` size in case someone forgets to add some new pallet here once created?

runtime/src/runtime_api.rs
mnaamani: target limit was probably a bad choice of name. The intent was to just try to have as many pallets below the "clippy" recommended max of ~ 200 I think, but as you rightly pointed out, we should aim for something that is safe give the wasm heap size.

runtime/src/runtime_api.rs
mnaamani: Good catch.  This test was also originally simply a check on the `Call` of the runtime which would cover all the pallets, so it might be wise to just keep it that way?

I may have over complicated this test as a way to enumerate the pallets and look for which pallets were problematic. Perhaps a separate bin package that just outputs the individual pallet Call sizes would be better.

But for now now I'll update the test by iterating over `AllPalletsWithSystem` as suggested.

runtime/src/runtime_api.rs
mnaamani: Do you have a figure I can plug in, based on your past "tests"?

Lezek123: I was able to overflow a 512 MiB wasm memory limit with `880,000` calls and 128 MiB limit with `220,000` calls of our current call size (~`360 bytes`)
`880,000 * 360 / 1024 / 1024 = 302 MiB`
So I think we should aim for `max_possible_batched_calls * call_size_bytes  <= 1 / 2 * wasm_heap_limit_bytes`.
Since our current `wasm_heap_limit` is `1 GiB` and `max_possible_batched_calls` is ~`1,250,000` (due to block size limit) we get:
```
1,250,000 / 1024 / 1024 * x <= 512
x = ~429
```

So I think we shouldn't go much higher than `400` to be absolutely safe.

runtime/src/runtime_api.rs
mnaamani: I couldn't find a way to iterate over `AllPalletsWithSystem`, its a tuple of `Pallet`. I ended up just reverting the test to a simple check on the runtime Call size, and added a small utility to dump the pallet sizes. It outputs:

```sh
$ ./target/release/call-sizes 
referendum: 64
forum: 72
constitution: 24
joystream_utility: 32
project_token: 80
proposals_discussion: 56
bounty: 176
storage: 128
working_group: 176
membership: 224
proposals_engine: 40
proposals_codex: 336
proposals_discussion: 56
content: 352
council: 96
----
Runtime: 368
```

bin/utils/call-sizes/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    println!(
+        "project_token: {}",
+        core::mem::size_of::<project_token::Call<Runtime>>()
+    );
+
+    println!(
+        "proposals_discussion: {}",
+        core::mem::size_of::<proposals_discussion::Call<Runtime>>()
+    );
```
Lezek123: It's a duplicate
```suggestion
```

runtime/src/runtime_api.rs
Lezek123: Hmm, for some reason I'm getting different results now:
```
referendum: 56
forum: 72
constitution: 24
joystream_utility: 32
project_token: 72
proposals_discussion: 56
bounty: 168
storage: 120
working_group: 160
membership: 200
proposals_engine: 40
proposals_codex: 296
proposals_discussion: 56
content: 304
council: 96
----
Runtime: 312
```

Tested on commit [`3a5329c93dbd49d4a2bb02b83bca593ade51f8fc`](https://github.com/Joystream/joystream/pull/4336/commits/3a5329c93dbd49d4a2bb02b83bca593ade51f8fc)

mnaamani:
Strange you are getting different result, now I did update from carthage branch.
Let me push update with suggested duplicate fix and we can run again both from latest commit and compare notes?

runtime/src/runtime_api.rs
mnaamani: Brilliant, I'll update the test.

mnaamani:
Still getting same from: [a515981](https://github.com/Joystream/joystream/pull/4336/commits/a515981ce462e7f887e5ead5380f940997172a2d)
```
referendum: 64
forum: 72
constitution: 24
joystream_utility: 32
project_token: 80
bounty: 176
storage: 128
working_group: 176
membership: 224
proposals_engine: 40
proposals_codex: 336
proposals_discussion: 56
content: 352
council: 96
----
Runtime: 368
```

Lezek123 (commented):
Because the call size is different depending on the processor architecture I think we should have CI checks that run `cargo test` on both AMD64 and ARM64 machines to be sure that the call size doesn't exceed the safe on either of those. Otherwise we're at risk of chain split between ARM and AMD architecture nodes.

mnaamani:
> Because the call size is different depending on the processor architecture I think we should have CI checks that run `cargo test` on both AMD64 and ARM64 machines to be sure that the call size doesn't exceed the safe on either of those. Otherwise we're at risk of chain split between ARM and AMD architecture nodes.

I'll update the github flows.

mnaamani:
> I'll update the github flows.

done in https://github.com/Joystream/joystream/pull/4336/commits/9ccc9590bb951112926c97618053eb7c8e38be94

.github/workflows/joystream-node.yml
```diff
@@ -51,7 +50,10 @@ jobs:
   # Build and test joystream-node with all runtime profiles
   runtime_profiles:
     name: Runtime Profiles
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
Lezek123: It looks like `macos-latest` is still using `x86-64` VM (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), so I'm not sure if there's a difference.

Also it looks like GH doesn't offer any github-hosted ARM instances, we would need to use a self-hosted one.

Unless there's some easy solution I'm missing perhaps it's not worth tackling now then?

mnaamani: Last time I checked macos-latest was infact using M1 processor, so arm64. I remember testing it in the create-release github flow and left a comment to that effect there: https://github.com/Joystream/joystream/blob/carthage/.github/workflows/create-release.yml#L18

And yes you are right, I didn't find a github-hosted linux arm64 instance, so for building the docker arm64 images we deploy AWS instance with arm64 arch `t4g` ec2 instance type: https://github.com/Joystream/joystream/blob/carthage/.github/workflows/joystream-node-docker.yml#L61

mnaamani: I think the github docs might be out of date, I'm pretty sure binary created on macos-latest didn't work on my old mac laptop with x86_64, but worked fine on M1..

mnaamani: I'll execute call-sizes in the flow to confirm. 86a4213d247c240f4e00b89bc0daf6b5886c7614

mnaamani: Weird both returned size `312` bytes. But https://github.com/github/roadmap/issues/528 suggests latest runners use arm64..

Lezek123: Following the roadmap I stumbled upon this post: https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware/, which explains that although the actions runner supports M1 and ARM64, GitHub doesn't provide a hosted runner with this architecture, we'd need to set up our own self-hosted one:
> Now the [Actions runner](https://github.com/actions/runner) supports M1 and the ARM64 runtime meaning developers can run it on **their own M1 or M2 hardware**.

And:

> For additional information on how to set up a **self-hosted macOS ARM64 runner,** please refer to our [documentation](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-default-labels-to-route-jobs).

mnaamani: I guess support still doesn't exist. https://github.com/actions/runner-images/issues/2187

mnaamani: Ah thanks for looking into this. I suppose we can look into adding our own runner at a later date.

Issue #4275, opened 2022-09-09
Trivial forum fixes


bedeho:
1. Drop `MaxPostsInThread` in runtime config, appears to not be used. (also, should'nt clippy warn about this?
1. Same with `MaxThreadsInCategory`.
1. Tweak this comment which appears to be incorectly describing automated cleanup as being a reason for constraints. Doubl check!

```Rust
/// Upper bounds for storage maps and double maps. Needed to prevent potential block exhaustion during deletion, etc.
/// MaxSubcategories, and MaxCategories should be reasonably small because when the category is deleted
/// all of it's subcategories with their threads and posts will be iterated over and deleted.
```

1. Rename `MaxCategories` to `MaxTotalCategories`, as people have been confused about meaning => https://github.com/Joystream/joystream/issues/3676
1. Rename `MaxSubcategories` to `MaxDirectSubcategoriesInCategory` for same reason.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202951933347568) by [Unito](https://www.unito.io)

dobertRowneySr:
From my understanding the concern is whether deletion is `O(1)` or not, and in the negative case we might want to bound it by `O(MaxDirectSubcategoriesInCategory)`. I am investigating into this (but I am pretty sure that the concern is well grounded) and constant deletion is not possible.

This answer :https://substrate.stackexchange.com/questions/5257/storage-double-map-key-deletion/5258#5258 confirm that is effectively so

PR #4341, opened 2022-10-05
Carthage/mainnet runtime parameters

Lezek123:
Implements runtime parameters according to https://github.com/Joystream/joystream/issues/4300 (proposal + discussion).

**Monthly budget estimation for council:** `$660,000`

### Notable changes
* Reward curve's `max_piece_count` parameter was set to `40` instead of `100`, as the value of `100` was causing one of the tests to fail with `Generated reward curve approximation is invalid: has more points than specified`.
* `RuntimeUpgradeWasmProposalMaxLength` was set to 3 MB instead of 3.5 MB, as `mega_bytes` macro doesn't support float values

## Missing parameter values
Those parameter values were not included in the original proposal:

### `content`
```rust
    pub const MaxNumberOfAssetsPerChannel: MaxNumber = 10;
    pub const MaxNumberOfAssetsPerVideo: MaxNumber = 20;
    pub const MaxNumberOfCollaboratorsPerChannel: MaxNumber = 10;
```

### `storage`
```rust
 pub const MaxNumberOfOperatorsPerDistributionBucket: u32 = 20;
```

### `forum`
Note that the proposal assumes https://github.com/Joystream/joystream/issues/4275 has been addressed, but it's not the case yet, so I left some of the values untouched, ie.:

```rust
    pub const MaxSubcategories: u64 = 40;
    pub const MaxThreadsInCategory: u64 = 20;
```

There is also one new value that was introduced recently:

```rust
pub const MaxStickiedThreads: u32 = 20;
```

## Current fees (examples)
_**Disclaimer:** I was using the **current** weights, obviously the fees will be impacted once we re-generate weights using a proper setup._

### `addForumPost`
addForumPost

### `addVideoComment`
addVideoComment

### `buyMembership`
buyMembership

### `createChannel`
createChannel

### `createForumThread`
createForumThread

### `createVideo`
createVideo

### `deleteChannel`
deleteChannel

### `deleteForumPost`
deleteForumPost

### `deleteForumThread`
deleteForumThread

### `deleteVideo`
deleteVideo

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203133169476109) by [Unito](https://www.unito.io)


runtime-modules/membership/src/lib.rs
```diff
@@ -141,9 +141,10 @@ pub trait Config:
 
     /// Stake needed to candidate as staking account.
     type CandidateStake: Get<BalanceOf<Self>>;
-}
 
-pub(crate) const DEFAULT_MEMBER_INVITES_COUNT: u32 = 5;
+    /// Default number of invites a paid membership recieves.
+    type DefaultMemberInvitesCount: Get<u32>;
```
bedeho, line 146: Great that you introduced this, we really need to get rid of all of those constants eventually.

runtime/src/lib.rs
```diff
@@ -1079,11 +1181,18 @@ impl common::membership::MembershipTypes for Runtime {
...
 
 parameter_types! {
-    pub const DefaultMembershipPrice: Balance = 100 * currency::CENTS;
+    pub const DefaultMembershipPrice: Balance = dollars!(1);
     pub const ReferralCutMaximumPercent: u8 = 50;
-    pub const DefaultInitialInvitationBalance: Balance = 100 * currency::CENTS;
-    // The candidate stake should be more than the transaction fee
-    pub const CandidateStake: Balance = 200 * currency::CENTS;
+    pub const DefaultInitialInvitationBalance: Balance = cents!(50);
+    pub const DefaultMemberInvitesCount: u32 = 2;
```
bedeho, line 1200: I see I forgot to comment on this in any of my replies explicitly, I guess I was just assuming reducing `DefaultInitialInvitationBalance` from $5 would be sufficient to address the problem you raised. This looks low to me, but let's see how much of a barrier it ends up becoming.

runtime/src/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    weights::{GetDispatchInfo, WeightToFee},
+    StorageHasher,
+};
+pub use sp_runtime::Perbill;
+use sp_std::mem::size_of;
+
+/// Compute total fee for executing a call
+pub fn compute_fee(call: Call) -> Balance {
+    let xt = UncheckedExtrinsic::new_unsigned(call);
+    let length = xt.encode().len() as u64 + 102; // We're adding 102 bytes, as this is the length of a signature
```
bedeho: Can we just make `102` a named constant with this standalone description?

runtime/src/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    },
+    traits::StorageInfoTrait,
+    weights::{GetDispatchInfo, WeightToFee},
+    StorageHasher,
+};
+pub use sp_runtime::Perbill;
+use sp_std::mem::size_of;
+
+/// Compute total fee for executing a call
+pub fn compute_fee(call: Call) -> Balance {
```
bedeho, line 22: This implementation looks right, but how did you arrive at this? did you extract the core logic of some other routines?

runtime/src/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    )
+}
+
+/// Compute stake while being sensitive to cleanup transaction fee value
+pub fn stake_with_cleanup(stake: Balance, max_cleanup_inclusion_fee: Balance) -> Balance {
+    stake.max(max_cleanup_inclusion_fee.saturating_add(StorageDepositCleanupProfit::get()))
+}
+
+/// Compute size of a double map entry using fixed-size portion of the value type
+pub fn double_map_entry_fixed_byte_size<Map, Hasher1, Key1, Hasher2, Key2, Value>() -> u32
```
bedeho, line 76: This was really nice work on these two routines 🥇

runtime/src/lib.rs
```diff
@@ -195,6 +198,12 @@ parameter_types! {
 
 const_assert!(NORMAL_DISPATCH_RATIO.deconstruct() >= AVERAGE_ON_INITIALIZE_RATIO.deconstruct());
 
+// Bloat-bond related global parameters:
+parameter_types! {
+    pub const BloatBondFeePerByte: Balance = currency::MILLICENTS;
```
bedeho: Some descriptions would really help to explain role of each of these two values.

runtime/src/lib.rs
```diff
@@ -588,8 +612,8 @@ parameter_types! {
     pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_BLOCKS / 4;
 
     // signed config
-    pub const SignedRewardBase: Balance = currency::DOLLARS; // TODO: adjust value
-    pub const SignedDepositBase: Balance = currency::DOLLARS; // TODO: adjust value
+    pub const SignedRewardBase: Balance = dollars!(1); // TODO: adjust value
```
bedeho, line 621: Should these still be having `TODO` remark?

runtime/src/constants.rs
```diff
@@ -187,19 +182,19 @@ mod tests {
     // This function tests that the fee for `MAXIMUM_BLOCK_WEIGHT` of weight is correct
     fn full_block_fee_is_correct() {
         println!("Base: {}", ExtrinsicBaseWeight::get());
-        // A full block should cost between 10 and 100 DOLLARS.
+        // A full block should cost between 2 and 10 DOLLARS.
```
bedeho: - Where did you get these numbers 2 and 10.
- Can we define them as values with a comment that explains where the value comes from? Will make it easier to maintain later.

runtime/src/constants.rs
```diff
@@ -102,9 +102,8 @@ pub mod fees {
     impl WeightToFeePolynomial for WeightToFee {
         type Balance = Balance;
         fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-            // in Polkadot, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
-            let p = super::currency::CENTS;
-            let q = 10 * Balance::from(ExtrinsicBaseWeight::get());
+            let p = 2 * super::currency::CENTS;
+            let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
```
bedeho: Is this equivalent to just making q=50*x, and dropping the p=2* ?

runtime/src/lib.rs
```diff
@@ -1010,29 +1085,56 @@ impl common::StorageOwnership for Runtime {
...
-    pub const MaxDataObjectSize: u64 = 10 * 1024 * 1024 * 1024; // 10 GB
+    pub const MinDistributionBucketsPerBag: u32 = 1;
+    pub const MaxDistributionBucketsPerBag: u32 = 51;
+    pub const MaxDataObjectSize: u64 = giga_bytes!(30);
     pub const MaxNumberOfOperatorsPerDistributionBucket: u32 = 20; // TODO: adjust value
+
+    // Data object bloat bond related:
+    // To calculate the cost of removing a data object we substract the cost of removing a video
+    // w/ 1 asset from a cost of removing a video w/ 2 assets
+    pub DataObjectCleanupTxFee: Balance = compute_fee(
```
bedeho, line 1112: This is clever, seems like it should work.

runtime/src/lib.rs
```diff
@@ -1010,29 +1085,56 @@ impl common::StorageOwnership for Runtime {
...
-    pub const MaxNumberOfPendingInvitationsPerDistributionBucket: u32 = 20; //TODO: adjust value
+    pub const BlacklistSizeLimit: u64 = 1_000;
+    pub const MaxNumberOfPendingInvitationsPerDistributionBucket: u32 = 20;
     pub const StorageModuleId: PalletId = PalletId(*b"mstorage"); // module storage
-    pub const MinDistributionBucketsPerBag: u32 = 1; //TODO: adjust value
-    pub const MaxDistributionBucketsPerBag: u32 = 100; //TODO: adjust value
-    pub const MaxDataObjectSize: u64 = 10 * 1024 * 1024 * 1024; // 10 GB
+    pub const MinDistributionBucketsPerBag: u32 = 1;
+    pub const MaxDistributionBucketsPerBag: u32 = 51;
+    pub const MaxDataObjectSize: u64 = giga_bytes!(30);
```
bedeho: Can we double this to 60?

runtime/src/lib.rs
```diff
@@ -1096,20 +1205,53 @@ impl membership::Config for Runtime {
...
+    type DefaultMemberInvitesCount = DefaultMemberInvitesCount;
 }
 
 parameter_types! {
     pub const MaxCategoryDepth: u64 = 6;
-    pub const MaxSubcategories: u64 = 40;
-    pub const MaxThreadsInCategory: u64 = 20;
-    pub const MaxPostsInThread: u64 = 20;
-    pub const MaxModeratorsForCategory: u64 = 20;
+    pub const MaxSubcategories: u64 = 40; // TODO: adjust
```
bedeho, line 1226: Why are there TODOs here?

bedeho (approved):
Fantastic job!

- Fees are a tad high, but it looks acceptable, thanks again for making this tool, it really makes a tremendous end-of-the-day sanity check.
- The automated fees were just fantastic. I'm slightly nervous about them not being `const` but I think it's worth it, I also secretly don't quite believe this actually breaks anything despite Parity saying so :D. 
- I think it would be great if we could try to isolate a few of these literal values, in particular tests, into named constants, so there is some separate space for comments and just more maintainable.

runtime/src/utils.rs
Lezek123: So initially I was thinking about using `TransactionPayment::compute_fee` approach as described in https://github.com/Joystream/joystream/issues/4300, but it had a few major drawbacks:
- The main issue was that it was reading a storage value (fee multiplier) which meant the result would change depending on it. This was very risky, because the runtime constants shouldn't change.
- It didn't work for genesis config values, only for the pallet's `Config` trait parameters

So I implemented the logic of `TransactionPayment::compute_fee` in this function while disregarding the fee multiplier (so assuming it's `1`). The formula for inclusion fee is quite simple as it's just:
```
inclusion_fee = base_fee + length_fee + [targeted_fee_adjustment * weight_fee];
```
_Source: https://docs.rs/pallet-transaction-payment/3.0.0/pallet_transaction_payment/_

So if we disregard  `targeted_fee_adjustment` (fee multiplier), it's:
```
inclusion_fee = base_fee + length_fee + weight_fee;
```

runtime/src/lib.rs
Lezek123: Yes, the values of those are still to be determined, the first ones were proposed here: https://github.com/Joystream/joystream/issues/4300#issuecomment-1268954991

runtime/src/constants.rs
Lezek123: Knowing that based on our assumptions it should cost `> $30,000 /day` to DDoS the chain, but we were also taking a margin for up to 4-5x that, we end up with more or less this price range for a full block.

With $2 per block it's `$28,800 / day`
With $10 per block it's `$144,000 / day`

I'll make this range a little tighter and add comments.

runtime/src/constants.rs
Lezek123: Yes

runtime/src/lib.rs
Lezek123: Sure

runtime/src/lib.rs
Lezek123: It's explained in the PR description. It's because https://github.com/Joystream/joystream/issues/4275 is not yet implemented, but the https://github.com/Joystream/joystream/issues/4300 proposal assumes it is

runtime/src/lib.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4341/commits/2adcc04e3f6aa5b615f9dda4696fe93e5aac93ee

runtime/src/constants.rs
Lezek123: I replaced it with 1/50 in https://github.com/Joystream/joystream/pull/4341/commits/2adcc04e3f6aa5b615f9dda4696fe93e5aac93ee

runtime/src/constants.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4341/commits/2adcc04e3f6aa5b615f9dda4696fe93e5aac93ee

runtime/src/lib.rs
Lezek123: Added in https://github.com/Joystream/joystream/pull/4341/commits/2adcc04e3f6aa5b615f9dda4696fe93e5aac93ee

runtime/src/utils.rs
Lezek123: Thank you

runtime/src/utils.rs
Lezek123: Done in https://github.com/Joystream/joystream/pull/4341/commits/2adcc04e3f6aa5b615f9dda4696fe93e5aac93ee

Lezek123:
> Fees are a tad high, but it looks acceptable

The considerable part of the cost of channel/video creation is a result of `StorageDepositCleanupProfit` being set to `1 cent`, which means all bloat bonds (including data object bloat bond) are 1 cent higher than they really need to be.
For example, when creating a video with 3 assets (thumbnail + media + subtitles), the additional 4 cents are a result of `StorageDepositCleanupProfit` alone (`video_bloat_bond + 3 * data_object_bloat_bond`).

When it comes to inclusion fees, most of the cost is weight-driven. Notice in case of channels and videos we're using a worst case scenario number of storage/distribution buckets, the actual numbers may be much lower.

> I'm slightly nervous about them not being const

I think this is not a problem as long as the function that computes them doesn't depend on anything that may change over time. Currently the only way to make those functions (like `compute_fee` / `compute_single_bloat_bond` etc.) return different results would be to modify the runtime code. So they are constant in a sense, they are just not defined as constant functions.

runtime/src/lib.rs
mnaamani: 4275 is now implemented by Ignazio in https://github.com/Joystream/joystream/pull/4352

mnaamani (approved):

Issue #4298, opened 2022-09-16
Include abstention votes


bedeho:
Currentl works like this

```
    approval_quorum_bar
    This bar indicates the value : ( votes_count - abstentions) /council_size
    
    approval threshold bar
    This bar indicates the value : approves / (votes_count - abstention)
    
    slash quorum bar
    This bar indicates the value : (votes_count - abstentions) /council_size
    
    slash threshold bar
    This bar indicates the value : slashes / (votes_count - abstention)
```

Change to

```
    approval_quorum_bar
    This bar indicates the value : votes_count/council_size
    
    approval threshold bar
    This bar indicates the value : approves/votes_count
    
    slash quorum bar
    This bar indicates the value : votes_count/council_size
    
    slash threshold bar
    This bar indicates the value : slashes/votes_count
```

NB: requires more discussion

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203058397951197) by [Unito](https://www.unito.io)

dmtrjsg:
This topic originated from the bugfix on Pioneer:
- https://github.com/Joystream/pioneer/issues/2715

While reviewing the PR with the team I personally was surprised that Abstain is not calculated towards Quorum, while considered to still count as vote.  Quick desktop research indicated that a common mechanics would be for Abstain vote to count towards Quorum, and that is in fact is one of the purposes of such a vote.

<img width="789" alt="Screenshot 2022-09-16 at 17 50 34" src="https://user-images.githubusercontent.com/93924525/190654686-e3aba5f2-bb4d-4797-895f-a34f24f98e33.png">

Extrapolating this to a generic user expectations of evaluating likelihood of proposal likelihood to be rejected/ accepted for cases when Abstain votes are casted, `the hypothesis is that most generic users would expect Abstain to count towards quorum.`

The implication of such a decision on voting dynamics, is that council members who vote with "Abstain" would increase the likelihood of proposals to be executed for proposals where Approval Quorum > Approval Threshold. 

NB: Currently runtime proposal consts are set up with vast majority of Approval Quorum < Approval Threshold, so there will be no effect on changing the quorum calculation logic without changing Proposal Consts values for Threshold and/ or Quorum.

[...]

Issue #2953, opened 2021-12-20
Change effect of abstention


bedeho:
# Background
There is considerable dissatisfaction with the fact that abstentions count towards quorum, but do not count towards the threshold, hence being effectively counted as a reject in terms of effect. See here

https://github.com/Joystream/joystream/issues/2589

There is even a outside PR to fix this, but I am not sure if it is the right fix

https://github.com/Joystream/joystream/pull/2952

# Proposal
Abstention is supposed to be there as a way to allow a councilor to be able to signal awareness of a proposal, unlike if they just don't vote, but not being able or willing to take a position on the question. To me, it seems like either this should be treated as today, which means it counts towards quorum, and it counts as a reject, or, it just does not count towards the quorum at all, and thus also not threshold. Since people are unhappy about the former, seems the latter would be better.

I don't believe this change has any effect on anything outside the runtime, be it CLI, query node, Pioneer, etc. Some documentation needs to be updated though.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201709380879215/1201710200291876) by [Unito](https://www.unito.io)

shamil-gadelshin:
Wouldn't `lowering the threshold for troubling proposal types` be a safer and easier option for dealing with the problem? Those values are just runtime constants that could be changed back and forth during experiments (runtime upgrade is required though).

As an alternative, some changes variants could lead to a single `Approve` that will decide the outcome of a major proposal. Thanks @singulart for highlighting it in the [PR test](https://github.com/singulart/joystream/blob/169ae70d097a388db2c354346dbac68d2597d021/runtime-modules/proposals/engine/src/types/mod.rs#L589). Do we want this behavior?

bedeho:
> Wouldn't lowering the threshold for troubling proposal types be a safer and easier option for dealing with the problem?

I don't think this has anything to do with some proposal types being problematic, so I am not sure exactly what your understanding of the problem is.

> As an alternative, some changes variants could lead to a single Approve that will decide the outcome of a major proposal.  

This would not be the case if we did as I suggested above 

> or, it just does not count towards the quorum at all, and thus also not threshold

shamil-gadelshin:
I referred to such quotes from the original issue: `voter participation requirements up to 93.75%` . We could change the required voter participation percent by changing the threshold value.

Let's introduce an example:
Council size = 20
Approval quorum = 51%
Approval threshold = 51%

Voting results:
- Abstain = 19
- Approve = 1

[...]

PR #4345, opened 2022-10-07
Abstension votes included in qurom and threshold calculations

mnaamani:
Reverting of https://github.com/Joystream/joystream/issues/2953
Implements https://github.com/Joystream/joystream/issues/4298

As far as I can tell, no mappings changes are needed in Query Node.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203132768592827) by [Unito](https://www.unito.io)


runtime-modules/proposals/engine/src/tests/mod.rs
```diff
@@ -2714,123 +2708,3 @@ fn proposal_status_resolution_is_slashing_achievable_works_correctly() {
     assert!(slashing_threshold_achievable_resolution.is_slashing_threshold_achievable());
     assert!(!slashing_threshold_achievable_resolution.is_rejection_imminent());
 }
-
-#[test]
-fn proposal_status_resolution_slashing_threshold_works_correctly_with_abstentions() {
```
dobertRowneySr, line 2719: Maybe it's worth leaving these test since they provide explicit evidence that abstentions are effectively accounted for. WDYT?

dobertRowneySr (changes requested):
I only left a couple of small suggestions, but the PR LGTM even now.
I would change the name from `remaining_votes` to `abstentions

runtime-modules/proposals/engine/src/tests/mod.rs
mnaamani: Well those tests were were passing a value for `abstentions` field in `ProposalStatusResolution`. This field was dropped. So it doesn't make sense to  add those tests again.

mnaamani: But it could make sense to add some more tests of this type https://github.com/Joystream/joystream/blob/carthage/runtime-modules/proposals/engine/src/tests/mod.rs#L532 to confirm Abtention behavior is as expected..

mnaamani:
> I would change the name from `remaining_votes` to `abstentions

Do you mean in these statements?

```
let remaining_votes_count = self.total_voters_count.saturating_sub(self.votes_count);
```

Abstentions are counted in self.votes_count, which comes from `VotingResults::votes_number()`.

runtime-modules/proposals/engine/src/tests/mod.rs
dobertRowneySr: What I was suggesting was:
Have a test with `abstentions > 0` and having the correct quorum ratio

dobertRowneySr (approved):
LGTM

Issue #4344, opened 2022-10-06
Verify setting sudo key to zero disables sudo ext.


dobertRowneySr:
### Background
Originally we were planning to do a runtime upgrade and drop the pallet
instead I suggested we might be able to disable it without doing a runtime upgrade
by setting the sudo key to Zero.

### Goal
Verify that

> signed extensions checks on transactions that test their validity prevent any extrinsic signed with account `Zero`, so setting sudo key to `Zero` should block all sudo extrinsics

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203115672496257) by [Unito](https://www.unito.io)

mnaamani:
My original proposition was to set the sudo key (address) to zero as a means to disable sudo access. By setting the account to zero, any signed extrinsic with the corresponding private key would not pass the transaction validation check because we have a `SignedExtension` called `CheckNonZeroSender`. 

The reason it turns out for the need of `CheckNonZeroSender` is that the corresponding private key is easily derived (publicly known).
 - https://crypto.stackexchange.com/questions/49994/find-ecdsa-privkey-to-pubkey-0 
 - https://substrate.stackexchange.com/questions/982/why-does-the-all-0-public-key-have-a-known-private-key-in-sr25519-and-ed25519
 - https://substrate.stackexchange.com/questions/5212/api-js-how-to-set-an-accountid32-to-zero

Note also that in newer substrate version we updated to recently the Default trait implementation of AccountId was dropped, so the "zero" address could not be easily constructed, and use of address = 0 is discouraged. 

So although the signed extension would actually prevent a sudo dispatch via extrinsic, it is not worth taking the risk that
For example a combination of accounts could be found to produce a multisig address of zero..

PR #4356, opened 2022-10-10
Carthage zero sudo key

dobertRowneySr:
addresses #4344 

Sudo pallet appears not to be disabled by a zero key


dobertRowneySr:
Closing:
setting sudo pub key is actually dangerous because of this fact https://substrate.stackexchange.com/questions/982/why-does-the-all-0-public-key-have-a-known-private-key-in-sr25519-and-ed25519

Issue #4343, opened 2022-10-06
Prepare `Supervised` stage runtime


bedeho:
As per this plan 

https://github.com/Joystream/joystream/issues/3792

However, remember to not remove all filters, e.g. bounty, crt and other proposals and frozen features.

```Rust
// Stage 2: Filter out only a subset of calls on content pallet, some specific proposals
// and the bounty creation call.
#[cfg(not(feature = "runtime-benchmarks"))]
fn filter_stage_2(call: &<Runtime as frame_system::Config>::Call) -> bool {
    // TODO: adjust after Carthage
    match call {
        Call::Content(content::Call::<Runtime>::destroy_nft { .. }) => false,
        Call::Content(content::Call::<Runtime>::toggle_nft_limits { .. }) => false,
        Call::Content(content::Call::<Runtime>::update_curator_group_permissions { .. }) => false,
        Call::Content(content::Call::<Runtime>::update_channel_privilege_level { .. }) => false,
        Call::Content(content::Call::<Runtime>::update_channel_nft_limit { .. }) => false,
        Call::Content(content::Call::<Runtime>::set_channel_paused_features_as_moderator {
            ..
        }) => false,
        Call::Content(content::Call::<Runtime>::initialize_channel_transfer { .. }) => false,
        Call::Content(content::Call::<Runtime>::issue_creator_token { .. }) => false,
        Call::Bounty(bounty::Call::<Runtime>::create_bounty { .. }) => false,
        Call::ProposalsCodex(proposals_codex::Call::<Runtime>::create_proposal {
            general_proposal_parameters: _,
            proposal_details,
        }) => !matches!(
            proposal_details,
            proposals_codex::ProposalDetails::UpdateGlobalNftLimit(..)
        ),
        _ => true,
    }
}
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203114471756139) by [Unito](https://www.unito.io)

bedeho:
Make new branch `carthage_supervised`

bedeho:
Into new branch `carthage_supervised`.

PR #4370, opened 2022-10-12
Carthage supervised stage

dobertRowneySr:
use filter stage 2 https://github.com/Joystream/joystream/issues/4343

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203152438491542) by [Unito](https://www.unito.io)


mnaamani:
Please bump the spec version from 3 to 4, for `pub const VERSION` in `runtime/src/lib.rs` and version in runtime and node Cargo.toml (just the minor version).

Also update branch from carthage (will need to do this a couple times until carthage is finalized)

Issue #4368, opened 2022-10-12
Prepare `Liberated` runtime 


bedeho:
Make new branch `carthage_liberated`  where `sudo_pallet` is removed, based on `carthage_supervised` branch.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203149176225725) by [Unito](https://www.unito.io)

PR #4371, opened 2022-10-12
Carthage liberated drop sudo

dobertRowneySr:
drops sudo pallet #4368 

Should we burn remaining sudo balance during migration?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203152437212564) by [Unito](https://www.unito.io)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2850,7 +2850,6 @@ dependencies = [
  "pallet-staking-handler",
  "pallet-staking-reward-curve",
  "pallet-storage",
- "pallet-sudo",
  "pallet-timestamp",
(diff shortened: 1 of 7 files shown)


mnaamani:
Similarly to supervised runtime, please bump the spec version from 3 to 5, for `pub const VERSION` in `runtime/src/lib.rs` and version in runtime and node Cargo.toml (just the minor version).

Also update branch from supervised (will need to do this a couple times until carthage is finalized)

mnaamani:
> Should we burn remaining sudo balance during migration?

Good idea, @bwhm ?

Issue #4369, opened 2022-10-12
Core Weight Generation Error


mnaamani:
Running weight generation script on branch https://github.com/Joystream/joystream/tree/carthage-weights-gen (which just combines all the yet unmerged PRs for carthage) I get an error on the reference hardware machine AWS `c6id.2xlarge` EC2 instance type ->  3rd Generation Intel Xeon Scalable processors (code named Ice Lake) https://aws.amazon.com/ec2/instance-types/c6i/

```
uname -m
x86_64

Generating storage weights
Storage weights generated successfully
It took 1 seconds
Generating core weights
Error: Client(ApplyExtrinsicFailed(Validity(TransactionValidityError::Invalid(InvalidTransaction::BadProof))))
Core weights generated successfully
It took 2 seconds
Generating weights for frame_system...
.
.
```

The weight generation is successful however. Is this something to be concerned about?

I cannot reproduce the same error on Mac M1, or other AMD64 machines.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203150191418395) by [Unito](https://www.unito.io)

Lezek123:
No issue on Ubuntu 20.04.4 LTS, Intel Core i5-10400F:
```bash
$ uname -m
x86_64

$ ./scripts/generate-weights.sh 1 1
Generating storage weights
Storage weights generated successfully
It took 2 seconds
Generating core weights
Core weights generated successfully
It took 424 seconds
# ...
```

Lezek123:
However I noticed that this benchmark took very long time to execute, as compared to values shown in the issue description, ie.:

**My local run:**
```
Core weights generated successfully
It took 424 seconds
```

**AWS run (issue desc.):**
```
Core weights generated successfully
It took 2 seconds
```

mnaamani:
> However I noticed that this benchmark took very long time to execute, as compared to values shown in the issue

Great observation, I presume this means the generated weights "block_weights.rs" and "extrinsics_weights.rs" are not reliable. I will generate on Google Cloud ref machine in this case on, AMD Zen3, or newer (EPYC or Ryzen) and report back.

mnaamani:
On the AMD processor the error has gone away, I have generated weights and pushed them 5e68a62bbf2503c7c0d427079cb7aa8429565793

I also ran full integration tests on the amazon machine to confirm no issues appeared.

PR #4374, opened 2022-10-14
Carthage Updated Weights

mnaamani:
New weights generated (from commit 674974ab35a0fa8e3758f68414c096be8bf80ee2) , switched from Amazon machine to Google Cloud (see https://github.com/Joystream/joystream/issues/4369) 

The block execution weight has changed so there is a test failing now:

```
---- constants::tests::block_weight_fill_cost_per_day_is_correct stdout ----
weight per block: 1500000000000, block cost: 185¢, cost/day: $26703
thread 'constants::tests::block_weight_fill_cost_per_day_is_correct' panicked at 'assertion failed: day_of_full_blocks_cost >= BLOCK_WEIGHT_FILL_MIN_DAILY_COST', runtime/src/constants.rs:209:9
```

```rust
    // before
    pub const BlockExecutionWeight: Weight = 4_870_839 * WEIGHT_PER_NANOS;
    // new weight
    pub const BlockExecutionWeight: Weight = 8_056_202 * WEIGHT_PER_NANOS;

    // before
    pub const ExtrinsicBaseWeight: Weight = 70_231 * WEIGHT_PER_NANOS;
    // new weight
    pub const ExtrinsicBaseWeight: Weight = 161_778 * WEIGHT_PER_NANOS;

```

How flexible are we to change the expected block fees here, or should we run a second time to figure why the block weight changed so much?
https://github.com/Joystream/joystream/blob/64f13d24d039b71fdbff77fdf44f2d8818b46d20/runtime/src/constants.rs#L190

This is not final, will need to run again as there are some updates on carthage still not included here.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203164486975560) by [Unito](https://www.unito.io)


mnaamani:
After discussing with @Lezek123 , we decided to try weight generation again with recommended optimization for the node, namely disabling SMT (hyper-threading) and using NVMe interface for disk access. I configured two local SSDs on google instance and combined into single disk to get better performance. Generating the weights now, will update the PR once ready.

mnaamani:
> After discussing with @Lezek123 , we decided to try weight generation again with recommended optimization for the node, namely disabling SMT (hyper-threading) and using NVMe interface for disk access. I configured two local SSDs on google instance and combined into single disk to get better performance. Generating the weights now, will update the PR once ready.

Happy to report new generated weights do not break our fee range assumptions.

runtime-modules/bounty/src/weights.rs
```diff
@@ -210,18 +211,20 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Bounty EntryCount (r:1 w:1)
 	// Storage: Bounty Entries (r:0 w:1)
-	fn announce_work_entry(i: u32, _j: u32, ) -> Weight {
-		(445_662_000 as Weight)
+	fn announce_work_entry(i: u32, j: u32, ) -> Weight {
+		(140_991_000 as Weight)
 			// Standard Error: 0
 			.saturating_add((1_000 as Weight).saturating_mul(i as Weight))
+			// Standard Error: 592_000
+			.saturating_add((165_000 as Weight).saturating_mul(j as Weight))
```
Lezek123: I just noticed that those parameters are passed in the reverse order in `runtime-modules/bounty/src/lib.rs`, ie.:
- `i` is the metadata length, but `T::ClosedContractSizeLimit::get().saturated_into()` gets passed here
- `j` is the closed contract size limit, but `work_description.len().saturated_into()` gets passed here

runtime-modules/bounty/src/weights.rs
```diff
@@ -233,16 +236,12 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-			// Standard Error: 0
-			.saturating_add((1_000 as Weight).saturating_mul(j as Weight))
-			// Standard Error: 0
-			.saturating_add((1_000 as Weight).saturating_mul(k as Weight))
-			// Standard Error: 2_619_000
-			.saturating_add((54_133_000 as Weight).saturating_mul(w as Weight))
-			// Standard Error: 2_619_000
-			.saturating_add((79_574_000 as Weight).saturating_mul(r as Weight))
+	fn submit_oracle_judgment_by_council(_j: u32, _k: u32, w: u32, r: u32, ) -> Weight {
+		(11_663_510_000 as Weight)
```
Lezek123: Too bad that benchmarks didn't figure out the weight dependency on `j` and `k`, causing the base weight to be extremely overestimated here.

I think the reason for this is that `j` and `k` represent just a single metadata byte, which by itself has a negligible effect on weight (<= 1000), but with `3,000,000` bytes and `1,000` weight/byte, this is already `3,000,000,000` weight.

The best solution I can think of is to use bigger unit when benchmarking the impact of the metadata on weight. For example, 1 KB instead of 1B. In this case when we pass the metadata length to weight function we should do `div_ceil(1000)` and in benchmarking when generating the metadata `Vec` we should use, for example: `vec![0xff].repeat(1000 * i)`

runtime-modules/content/src/weights.rs
```diff
@@ -170,21 +171,15 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Storage NextDataObjectId (r:1 w:1)
 	fn channel_update_with_assets(a: u32, b: u32, c: u32, d: u32, e: u32, ) -> Weight {
-		(8_694_000 as Weight)
-			// Standard Error: 8_858_000
-			.saturating_add((4_998_000 as Weight).saturating_mul(a as Weight))
-			// Standard Error: 8_858_000
-			.saturating_add((62_585_000 as Weight).saturating_mul(b as Weight))
-			// Standard Error: 8_858_000
-			.saturating_add((32_089_000 as Weight).saturating_mul(c as Weight))
+		(971_918_000 as Weight)
```
Lezek123: Here it seems that the weight impact of `a` and `b` (not counting db operations) has been absorbed by the base weight, similarly to how this has been the case in bounty's `submit_oracle_judgment_by_council` weight function.

Strangely enough `a` and `b` seemed to have quite substantial impact on weight before (`5,000,000 * a` and `62,500,000 * b`), so this cannot be explained as just a simple rounding error.

There are 2 possible explainations: either the previous weight function was incorrect or this one is. If the latter is the case, I'm wondering what can we do to avoid these kind of errors, perhaps tweaking the `--external-repeat` value would help?

First I think we should try out different `steps` / `repeats` / `external-repeat` values for this specific function and see how they affect the weight function

runtime-modules/content/src/weights.rs
```diff
@@ -199,9 +194,7 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: Content CuratorGroupById (r:1 w:0)
 	// Storage: Membership MembershipById (r:1 w:0)
 	fn channel_update_without_assets(a: u32, b: u32, ) -> Weight {
-		(138_469_000 as Weight)
-			// Standard Error: 13_594_000
-			.saturating_add((16_303_000 as Weight).saturating_mul(a as Weight))
+		(510_448_000 as Weight)
```
Lezek123: `a` weight impact absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -258,12 +251,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Storage DataObjectsById (r:1 w:1)
 	// Storage: Storage DataObjectPerMegabyteFee (r:1 w:0)
-	// Storage: Storage StorageBucketById (r:20 w:20)
+	// Storage: Storage StorageBucketById (r:13 w:13)
 	// Storage: System Account (r:2 w:2)
 	fn delete_channel_assets_as_moderator(a: u32, b: u32, c: u32, ) -> Weight {
-		(968_325_000 as Weight)
-			// Standard Error: 6_304_000
-			.saturating_add((3_758_000 as Weight).saturating_mul(b as Weight))
+		(2_061_461_000 as Weight)
```
Lezek123: Weight impact of `b` absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -278,17 +269,13 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: System Account (r:3 w:3)
 	fn delete_channel_as_moderator(a: u32, b: u32, c: u32, d: u32, ) -> Weight {
-		(410_411_000 as Weight)
-			// Standard Error: 11_142_000
-			.saturating_add((23_122_000 as Weight).saturating_mul(a as Weight))
-			// Standard Error: 5_889_000
-			.saturating_add((11_908_000 as Weight).saturating_mul(b as Weight))
-			// Standard Error: 961_000
-			.saturating_add((8_864_000 as Weight).saturating_mul(c as Weight))
+		(1_605_250_000 as Weight)
```
Lezek123: The impact of `a` and `b` absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -437,16 +424,14 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Storage NextDataObjectId (r:1 w:1)
 	// Storage: Storage DataObjectsById (r:0 w:1)
 	// Storage: Content VideoById (r:0 w:1)
 	fn create_video_without_nft(a: u32, b: u32, c: u32, ) -> Weight {
-		(340_721_000 as Weight)
-			// Standard Error: 3_592_000
-			.saturating_add((21_337_000 as Weight).saturating_mul(a as Weight))
-			// Standard Error: 4_786_000
-			.saturating_add((3_485_000 as Weight).saturating_mul(b as Weight))
+		(1_502_789_000 as Weight)
```
Lezek123: The impact of `a` absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -488,13 +473,11 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Content VideoById (r:0 w:1)
 	fn create_video_with_nft(a: u32, b: u32, c: u32, d: u32, ) -> Weight {
-		(556_551_000 as Weight)
-			// Standard Error: 3_833_000
-			.saturating_add((20_807_000 as Weight).saturating_mul(a as Weight))
-			// Standard Error: 5_132_000
-			.saturating_add((7_113_000 as Weight).saturating_mul(b as Weight))
-			// Standard Error: 852_000
-			.saturating_add((3_117_000 as Weight).saturating_mul(c as Weight))
+		(613_162_000 as Weight)
```
Lezek123: The impact of `b` absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -567,12 +550,12 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-		(388_056_000 as Weight)
-			// Standard Error: 397_000
-			.saturating_add((4_768_000 as Weight).saturating_mul(a as Weight))
+		(247_814_000 as Weight)
+			// Standard Error: 931_000
+			.saturating_add((9_010_000 as Weight).saturating_mul(a as Weight))
 			// Standard Error: 0
 			.saturating_add((1_000 as Weight).saturating_mul(b as Weight))
-			.saturating_add(T::DbWeight::get().reads(21 as Weight))
+			.saturating_add(T::DbWeight::get().reads(22 as Weight))
```
Lezek123: Strange, the number of `reads` changed, but there is no change in the generated comment.
May be worth investigating.

runtime-modules/content/src/weights.rs
```diff
@@ -599,23 +582,19 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-		(646_407_000 as Weight)
-			// Standard Error: 3_371_000
-			.saturating_add((18_908_000 as Weight).saturating_mul(a as Weight))
-			// Standard Error: 3_371_000
-			.saturating_add((24_593_000 as Weight).saturating_mul(b as Weight))
-			// Standard Error: 4_552_000
-			.saturating_add((3_434_000 as Weight).saturating_mul(c as Weight))
-			// Standard Error: 764_000
-			.saturating_add((4_221_000 as Weight).saturating_mul(d as Weight))
+		(1_983_336_000 as Weight)
```
Lezek123: Impact of `b` and `c` absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -961,12 +941,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: Content GlobalWeeklyNftLimit (r:1 w:0)
 	// Storage: Content GlobalWeeklyNftCounter (r:1 w:1)
 	fn issue_nft(w: u32, b: u32, ) -> Weight {
-		(275_659_000 as Weight)
-			// Standard Error: 453_000
-			.saturating_add((5_519_000 as Weight).saturating_mul(w as Weight))
+		(1_102_450_000 as Weight)
```
Lezek123: The impact of `w` parameter absorbed into base weight

runtime-modules/content/src/weights.rs
```diff
@@ -722,15 +701,16 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: Token BloatBond (r:1 w:0)
 	// Storage: Token SymbolsUsed (r:1 w:1)
 	// Storage: Membership MembershipById (r:1 w:0)
+	// Storage: System Account (r:2 w:2)
```
Lezek123, line 732: This suggests this benchmark is not written properly, as the final result depends on the genesis config parameters. Now that we have non-zero bloat bond, those `2` reads and `2` writes were added, which means the bloat bond is probably not setup to a non-zero value during the benchmark setup step.

runtime-modules/content/src/weights.rs
```diff
@@ -961,12 +941,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Content GlobalWeeklyNftCounter (r:1 w:1)
 	fn issue_nft(w: u32, b: u32, ) -> Weight {
-		(275_659_000 as Weight)
-			// Standard Error: 453_000
-			.saturating_add((5_519_000 as Weight).saturating_mul(w as Weight))
+		(1_102_450_000 as Weight)
 			// Standard Error: 0
 			.saturating_add((1_000 as Weight).saturating_mul(b as Weight))
-			.saturating_add(T::DbWeight::get().reads(20 as Weight))
+			.saturating_add(T::DbWeight::get().reads(21 as Weight))
```
Lezek123: Number of reads unexpectedly changed

runtime-modules/forum/src/weights.rs
```diff
@@ -371,10 +374,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	fn delete_posts(i: u32, _j: u32, k: u32, ) -> Weight {
 		(0 as Weight)
-			// Standard Error: 31_668_000
-			.saturating_add((1_246_434_000 as Weight).saturating_mul(i as Weight))
-			// Standard Error: 341_000
-			.saturating_add((49_955_000 as Weight).saturating_mul(k as Weight))
+			// Standard Error: 5_830_000
+			.saturating_add((1_730_178_000 as Weight).saturating_mul(i as Weight))
+			// Standard Error: 36_000
+			.saturating_add((116_990_000 as Weight).saturating_mul(k as Weight))
```
Lezek123: Impact of `j` probably absorbed by either `i` or `k`

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -83,12 +84,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_signal(i: u32, t: u32, _d: u32, ) -> Weight {
-		(91_315_000 as Weight)
+	fn create_proposal_signal(i: u32, _t: u32, _d: u32, ) -> Weight {
+		(531_108_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -103,14 +102,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_runtime_upgrade(i: u32, t: u32, d: u32, ) -> Weight {
-		(0 as Weight)
+	fn create_proposal_runtime_upgrade(i: u32, _t: u32, _d: u32, ) -> Weight {
+		(478_699_000 as Weight)
```
Lezek123: The impact of `t` and `d` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -212,10 +207,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_fill_working_group_lead_opening(t: u32, d: u32, ) -> Weight {
-		(102_741_000 as Weight)
-			// Standard Error: 38_000
-			.saturating_add((63_000 as Weight).saturating_mul(t as Weight))
+	fn create_proposal_fill_working_group_lead_opening(_t: u32, d: u32, ) -> Weight {
+		(204_069_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -251,10 +244,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_decrease_working_group_lead_stake(t: u32, d: u32, ) -> Weight {
-		(94_265_000 as Weight)
-			// Standard Error: 31_000
-			.saturating_add((17_000 as Weight).saturating_mul(t as Weight))
+	fn create_proposal_decrease_working_group_lead_stake(_t: u32, d: u32, ) -> Weight {
+		(183_322_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -312,10 +305,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_terminate_working_group_lead(t: u32, d: u32, ) -> Weight {
-		(96_395_000 as Weight)
-			// Standard Error: 29_000
-			.saturating_add((1_000 as Weight).saturating_mul(t as Weight))
+	fn create_proposal_terminate_working_group_lead(_t: u32, d: u32, ) -> Weight {
+		(183_844_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -375,10 +366,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_set_membership_price(t: u32, d: u32, ) -> Weight {
-		(88_195_000 as Weight)
-			// Standard Error: 28_000
-			.saturating_add((31_000 as Weight).saturating_mul(t as Weight))
+	fn create_proposal_set_membership_price(_t: u32, d: u32, ) -> Weight {
+		(173_736_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -411,10 +404,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: ProposalDiscussion ThreadById (r:0 w:1)
 	// Storage: ProposalEngine Proposals (r:0 w:1)
 	// Storage: ProposalEngine DispatchableCallCode (r:0 w:1)
-	fn create_proposal_set_councilor_reward(t: u32, d: u32, ) -> Weight {
-		(90_065_000 as Weight)
-			// Standard Error: 28_000
-			.saturating_add((6_000 as Weight).saturating_mul(t as Weight))
+	fn create_proposal_set_councilor_reward(_t: u32, d: u32, ) -> Weight {
+		(174_635_000 as Weight)
```
Lezek123: The impact of `t` probably absorbed into base weight

runtime-modules/referendum/src/weights.rs
```diff
@@ -60,24 +61,16 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Council AnnouncementPeriodNr (r:1 w:1)
 	// Storage: Council Stage (r:0 w:1)
-	// Storage: Council Candidates (r:5 w:5)
-	// Storage: Balances Locks (r:5 w:5)
-	// Storage: System Account (r:5 w:5)
-	// Storage: Council CouncilorReward (r:1 w:0)
-	// Storage: Council Budget (r:1 w:1)
-	// Storage: Council CouncilMembers (r:1 w:1)
-	// Storage: ProposalEngine Proposals (r:1 w:0)
-	// Storage: Council NextRewardPayments (r:0 w:1)
```
Lezek123, line 70: Could be worth investigating why so many of the db reads/writes are no longer executed

runtime/src/tests/fee_tests.rs
```diff
@@ -30,8 +30,7 @@ fn runtime_upgrade_total_fee_is_correct() {
 
         assert!(x.max(y) - x.min(y) <= 1, "Inaccurate fee estimation");
 
-        // between 10 and 20 DOLLARS
-        assert!(y.gt(&DOLLARS.saturating_mul(10)));
-        assert!(y.lt(&DOLLARS.saturating_mul(20)));
+        assert!(y.lt(&DOLLARS.saturating_mul(80)));
+        // assert!(y.gt(&DOLLARS.saturating_mul(77)));
```
Lezek123, line 34: Does it fail with the previous range of $10-$20?

runtime/src/weights/pallet_im_online.rs
```diff
@@ -46,18 +48,18 @@ pub use pallet_im_online::weights::WeightInfo;
...
+		(171_083_000 as Weight)
+			// Standard Error: 0
+			.saturating_add((45_000 as Weight).saturating_mul(k as Weight))
 			// Standard Error: 4_000
-			.saturating_add((112_000 as Weight).saturating_mul(k as Weight))
-			// Standard Error: 43_000
-			.saturating_add((591_000 as Weight).saturating_mul(e as Weight))
-			.saturating_add(T::DbWeight::get().reads(5 as Weight))
+			.saturating_add((898_000 as Weight).saturating_mul(e as Weight))
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
```
Lezek123, line 62: Number of reads changed unexpectedly

mnaamani:
@Lezek123 The issues you pointed out re weight functions in the content pallet, could it be caused by a problem related to this:

```
failures:

---- benchmarks::benchmarking::tests::buy_nft stdout ----
thread 'benchmarks::benchmarking::tests::buy_nft' panicked at 'attempt to shift left with overflow', runtime-modules/content/src/benchmarks/mod.rs:225:13

---- benchmarks::benchmarking::tests::accept_channel_transfer_member_to_member stdout ----
thread 'benchmarks::benchmarking::tests::accept_channel_transfer_member_to_member' panicked at 'attempt to shift left with overflow', runtime-modules/content/src/benchmarks/mod.rs:225:13

---- benchmarks::benchmarking::tests::accept_channel_transfer_member_to_curator stdout ----
thread 'benchmarks::benchmarking::tests::accept_channel_transfer_member_to_curator' panicked at 'attempt to shift left with overflow', runtime-modules/content/src/benchmarks/mod.rs:225:13
```

The benchmark tests fail when run in debug build: `cargo test --features runtime-benchmarks`

```
failures:
    benchmarks::benchmarking::tests::accept_channel_transfer_curator_to_curator
    benchmarks::benchmarking::tests::accept_channel_transfer_member_to_curator
    benchmarks::benchmarking::tests::accept_channel_transfer_member_to_member
    benchmarks::benchmarking::tests::accept_incoming_offer
    benchmarks::benchmarking::tests::add_curator_to_group
    benchmarks::benchmarking::tests::buy_nft
    benchmarks::benchmarking::tests::cancel_buy_now
    benchmarks::benchmarking::tests::cancel_channel_transfer
    benchmarks::benchmarking::tests::cancel_english_auction
    benchmarks::benchmarking::tests::cancel_offer
    benchmarks::benchmarking::tests::cancel_open_auction
    benchmarks::benchmarking::tests::cancel_open_auction_bid
    benchmarks::benchmarking::tests::channel_agent_remark
    benchmarks::benchmarking::tests::channel_owner_remark
    benchmarks::benchmarking::tests::channel_update_with_assets
    benchmarks::benchmarking::tests::channel_update_without_assets
    benchmarks::benchmarking::tests::claim_channel_and_withdraw_curator_channel_reward
    benchmarks::benchmarking::tests::claim_channel_and_withdraw_member_channel_reward
    benchmarks::benchmarking::tests::claim_channel_reward
    benchmarks::benchmarking::tests::claim_creator_token_patronage_credit
    benchmarks::benchmarking::tests::create_channel
    benchmarks::benchmarking::tests::create_curator_group
    benchmarks::benchmarking::tests::create_video_with_nft
    benchmarks::benchmarking::tests::create_video_without_nft
    benchmarks::benchmarking::tests::creator_token_issuer_transfer
    benchmarks::benchmarking::tests::deissue_creator_token
    benchmarks::benchmarking::tests::delete_channel
    benchmarks::benchmarking::tests::delete_channel_as_moderator
    benchmarks::benchmarking::tests::delete_channel_assets_as_moderator
    benchmarks::benchmarking::tests::delete_video_as_moderator_with_assets
    benchmarks::benchmarking::tests::delete_video_as_moderator_without_assets
    benchmarks::benchmarking::tests::delete_video_assets_as_moderator
    benchmarks::benchmarking::tests::delete_video_with_assets
    benchmarks::benchmarking::tests::delete_video_without_assets
    benchmarks::benchmarking::tests::destroy_nft
    benchmarks::benchmarking::tests::finalize_creator_token_sale
    benchmarks::benchmarking::tests::finalize_revenue_split
    benchmarks::benchmarking::tests::init_creator_token_sale
    benchmarks::benchmarking::tests::initialize_channel_transfer
    benchmarks::benchmarking::tests::issue_creator_token
    benchmarks::benchmarking::tests::issue_nft
    benchmarks::benchmarking::tests::issue_revenue_split
    benchmarks::benchmarking::tests::issue_revenue_split_as_collaborator
    benchmarks::benchmarking::tests::make_creator_token_permissionless
    benchmarks::benchmarking::tests::make_english_auction_bid
    benchmarks::benchmarking::tests::make_open_auction_bid
    benchmarks::benchmarking::tests::nft_owner_remark
    benchmarks::benchmarking::tests::offer_nft
    benchmarks::benchmarking::tests::pick_open_auction_winner
    benchmarks::benchmarking::tests::reduce_creator_token_patronage_rate_to
    benchmarks::benchmarking::tests::remove_curator_from_group
    benchmarks::benchmarking::tests::sell_nft
    benchmarks::benchmarking::tests::set_channel_paused_features_as_moderator
    benchmarks::benchmarking::tests::set_channel_visibility_as_moderator
    benchmarks::benchmarking::tests::set_curator_group_status
    benchmarks::benchmarking::tests::set_video_visibility_as_moderator
    benchmarks::benchmarking::tests::settle_english_auction
    benchmarks::benchmarking::tests::sling_nft_back
    benchmarks::benchmarking::tests::start_english_auction
    benchmarks::benchmarking::tests::start_open_auction
    benchmarks::benchmarking::tests::update_buy_now_price
    benchmarks::benchmarking::tests::update_channel_nft_limit
    benchmarks::benchmarking::tests::update_channel_payouts
    benchmarks::benchmarking::tests::update_channel_privilege_level
    benchmarks::benchmarking::tests::update_channel_state_bloat_bond
    benchmarks::benchmarking::tests::update_curator_group_permissions
    benchmarks::benchmarking::tests::update_upcoming_creator_token_sale
    benchmarks::benchmarking::tests::update_video_state_bloat_bond
    benchmarks::benchmarking::tests::update_video_with_assets_with_nft
    benchmarks::benchmarking::tests::update_video_with_assets_without_nft
    benchmarks::benchmarking::tests::update_video_without_assets_with_nft
    benchmarks::benchmarking::tests::update_video_without_assets_without_nft
    benchmarks::benchmarking::tests::withdraw_from_curator_channel_balance
    benchmarks::benchmarking::tests::withdraw_from_member_channel_balance
```

The problematic code:
```rust
fn get_byte(num: u64, byte_number: u8) -> u8 {
    ((num & (0xff << (8 * byte_number))) >> (8 * byte_number)) as u8
}
```

Lezek123 (commented):
Most of the issues I identified belong to one of those 2 categories:
- New weight functions are missing some parameters and their impact on weight seems "absorbed" by either the constant component of the weight function or by some other parameters (for example: `content.channel_update_with_assets`)
- Number of db reads/writes changed unexpectedly, which means benchmark results may depend on genesis config parameters (`content.issue_creator_token`)

The first issue is especially hard to solve, because the results of running the weight generation script are very inconsistent, regardless of the number of steps/repeats/external-repeats used.

I was using `channel_update_with_assets` as an example and tried many different parameters locally. For each configuration I generated the weight functions a few times, but in no case was I able to get consistent results (in terms of which parameters the weight depends on) for more than 3 runs in a row. Some configurations I tried include:
- 50 steps, 20 repeats, 4x external-repeat (first 3 runs gave consistent results in this case, but the 4th one was different)
- 10 steps, 50 repeats, 5x external-repeat (first 2 runs already produced inconsistent results)
- 20 steps, 100 repeats, 1x external-repeat (same as above)

runtime/src/tests/fee_tests.rs
mnaamani: I can't remember when I changed this, probably in first weight generation run. But in general yes I find I have to change this test when the core weights change.

mnaamani:
> * Number of db reads/writes changed unexpectedly, which means benchmark results may depend on genesis config parameters (`content.issue_creator_token`)

This should be easy to resolve, as we discussed briefly earlier we can run the benchmarks for the production chain. (Will need a small change to node to allow a chain similar to `dev`, eg. `joystream-node --chain benchmarks` ) But of course it would be best to fix the benchmarks so they are not dependant on the initial values at genesis?

In the most recent weight generations I was running benchmar command and passing `--base-path` arg to select volume through NVMe, is it possible between multiple runs there was some state left over that affects the results? I know for sure when you run the dev chain with `joystream-node --dev` the chain directory is always cleared before the node starts. So maybe it is not relevant.

Lezek123:
Follow-up w.r.t. cases like `channel_update_with_assets`, where some parameters were omitted:

After a few more local tests I reached following conclusions:
- a margin of error of a few percent of worst-case-scenario extrinsic parameters seems almost unavoidable, I guess there is not much we can do about it other than update to a newer version of Substrate where the benchmarking framework may be possibly improved
- I was able to get most consistent results after lowering **steps/repeats to `10/5`** and adding **`--external-repeat` between 10 and 20**. When `--external-repeat x` flag is used, a new `StateMachine` instance is created for each of those `x` repeats. Then inside the `StateMachine` executor there is `--repeat` number of "internal" repeats. This happens for each step, so with parameters like: `--external-repeat 10 --steps 10 --repeat 5`, we get something like:
_Warning, pseudocode ahead!_
```rust
for step in 1..10 {
  // ...
  for external_repeat in 1..10 {
    StateMachine::create({ internal_repeats: 5, ... }).execute()
    // ...
  }
}

runtime-modules/bounty/src/weights.rs
Lezek123: Created a PR with a fix: https://github.com/Joystream/joystream/pull/4389

mnaamani:
Forum and Membership pallets updated benchmarks [acebda3](https://github.com/Joystream/joystream/pull/4374/commits/acebda3a53a9a03f3a1ca306f81b635d143f5baa) following fixes from https://github.com/Joystream/joystream/pull/4408

Re-ran benchmarks for election_provider_multi_phase, `d` component is now used again in `feasibility_check` but for `submit_unsigned` the `t` component was dropped. [0073cff](https://github.com/Joystream/joystream/pull/4374/commits/0073cfffb730600d157cc256577e40a318b2a345)

Honestly it doesn't seem worth spending more effort on this until we have a more reliable benchmarking framework in place (when we update to new version of substrate?)

bedeho:
> Honestly it doesn't seem worth spending more effort on this until we have a more reliable benchmarking framework in place 

When you say this, what ongoing problem are you then referring to?

mnaamani:
> When you say this, what ongoing problem are you then referring to?

Basically running the benchmark fixed one function, but another one now looks wrong in that one of the inputs/parameters is no longer used.

Lezek123 (approved):
LGTM

> Re-ran benchmarks for election_provider_multi_phase, d component is now used again in feasibility_check but for submit_unsigned the t component was dropped. [0073cff](https://github.com/Joystream/joystream/pull/4374/commits/0073cfffb730600d157cc256577e40a318b2a345)

I guess we can just re-use the previous weight function for `submit_unsigned`, which includes `t`, since they were both generated on the same machine?

mnaamani:
> LGTM
> 
> > Re-ran benchmarks for election_provider_multi_phase, d component is now used again in feasibility_check but for submit_unsigned the t component was dropped. [0073cff](https://github.com/Joystream/joystream/pull/4374/commits/0073cfffb730600d157cc256577e40a318b2a345)
> 
> I guess we can just re-use the previous weight function for `submit_unsigned`, which includes `t`, since they were both generated on the same machine?

Glad you mentioned that. I did suggest it to bedeho, in call earlier in the week, so we will go with that.
Updated in [a8512f4](https://github.com/Joystream/joystream/pull/4374/commits/a8512f4b521f34bb95d3e90412e212f7d4cdd929)

PR #4383, opened 2022-10-17
Colossus: verify data object during file upload

### LINKED ISSUE #4293: Colossus: explicitly handle nonexistent asset case
## Problem
If I try to upload asset (identified by `(bagId, dataObjectId)` pair) that doesn't exist (at least one of the IDs is wrong), Colossus will actually not let me know. Instead, as seen below, the upload will fail at asset size verification, telling me that there's size mismatch, when the nature of error is actually quite different.

https://github.com/Joystream/joystream/blob/97b311f071f6f76b5de78da281d88188466c63ce/storage-node/src/services/webApi/controllers/filesApi.ts#L272-L279

## Proposal
We should explicitly handle the case where `dataObject` wasn't found and provide meaningful error message to help with debugging.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203011377347786) by [Unito](https://www.unito.io)

zeeshanakram3:
fixes #4293

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203197951314120) by [Unito](https://www.unito.io)

### DIFF
--- a/storage-node/src/services/webApi/controllers/filesApi.ts
+++ b/storage-node/src/services/webApi/controllers/filesApi.ts
@@ -271,6 +271,10 @@ async function verifyDataObjectInfo(
 ): Promise<boolean> {
   const dataObject = await api.query.storage.dataObjectsById(bagId, dataObjectId)
 
+  if (dataObject.isEmpty) {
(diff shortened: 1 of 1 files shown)


Lezek123 (approved):
LGTM

PR #4385, opened 2022-10-18
"prod-test" chainspec

mnaamani:
- Add option to run a local "dev" chain using the production config settings at genesis instead of the test settings.

`joystream-node --chain prod-test --alice`

Unlike `joystream-node --dev` you will need to purge-chain between runs or use `--temp` flag:

`joystream-node --chain prod-test --alice --temp`

- Update generate weights script to use this chain config


Lezek123 (commented):
1. Some benchmarks were broken by this change so I fixed them in https://github.com/Joystream/joystream/pull/4386/commits/cb9f9b4b556a64bea005e25135ceed6c6821bf85
2. I think using `prod` chain name can be a little confusing, since it's not actually a production chain, it's just a chain with production config for benchmarking purposes

Lezek123:
> Unlike joystream-node --dev you will need to purge-chain between runs:
> `joystream-node purge-chain --chain prod`

It's also possible to use `--tmp` flag when running the chain, in this case the data directory is always cleaned

mnaamani:
> > Unlike joystream-node --dev you will need to purge-chain between runs:
> > `joystream-node purge-chain --chain prod`
> 
> It's also possible to use `--tmp` flag when running the chain, in this case the data directory is always cleaned

I'll have to test if the `--tmp` flag also works when using the "benchmark" command.

mnaamani:
> I'll have to test if the `--tmp` flag also works when using the "benchmark" command.

It's only available when running a full chain, not when running the benchmark command.

I renamed "prod" to "prod-test"

Lezek123 (approved):
LGTM

PR #4388, opened 2022-10-19
Carthage gateway rename

dobertRowneySr:
rename "Gateway" to "App"

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203409669373135) by [Unito](https://www.unito.io)


runtime-modules/common/src/locks.rs
```diff
@@ -14,7 +14,7 @@ parameter_types! {
     pub const BoundStakingAccountLockId: LockIdentifier = *b"boundsta";
     pub const BountyLockId: LockIdentifier = *b"bounty  ";
     pub const OperationsWorkingGroupAlphaLockId: LockIdentifier = *b"wg-opera";
-    pub const GatewayWorkingGroupLockId: LockIdentifier = *b"wg-gatew";
+    pub const AppWorkingGroupLockId: LockIdentifier = *b"wg-gatew";
```
mnaamani: should we also perhaps change the lock id itself to "wg-apps"

mnaamani (changes requested):
I suggested to update the lockid name for the working group.
Otherwise changes look good. We will need to update the types package version since runtime typenames changed. After https://github.com/Joystream/joystream/pull/4456 is merged as well.

PR needs to be updated from mainnet branch.
copying @thesan to be aware of upcoming changes.

mnaamani (approved):

PR #4389, opened 2022-10-19
Bounty weight function: Fix parameters passed in the wrong order

Lezek123:
Fixes an issue identified in https://github.com/Joystream/joystream/pull/4374#discussion_r997088342


mnaamani (approved):

mnaamani:
The ordering has been applied with correct `to_kb()` conversion as well for the work_description.len on carthage.

PR #4398, opened 2022-10-21
Carthage weights - Oct 21

mnaamani:
- Final weights generated.
- Added babe,grandpa,bags_list and election_provider_multi_phase benchmakrs
- Removed unused `pallet-elections-phragmen` from dependencies.


runtime-modules/forum/src/weights.rs
```diff
@@ -135,49 +136,51 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-			.saturating_add((10_415_000 as Weight).saturating_mul(i as Weight))
-			// Standard Error: 29_000
-			.saturating_add((2_732_000 as Weight).saturating_mul(j as Weight))
-			.saturating_add(T::DbWeight::get().reads(2 as Weight))
+		(28_127_000 as Weight)
+			// Standard Error: 33_000
+			.saturating_add((4_443_000 as Weight).saturating_mul(i as Weight))
+			// Standard Error: 0
+			.saturating_add((1_241_000 as Weight).saturating_mul(j as Weight))
+			.saturating_add(T::DbWeight::get().reads(1 as Weight))
```
Lezek123, line 144: I'm wondering why the number of reads changed for:
- `update_category_title_lead` (2 => 1)
- `update_category_title_moderator` (2 => 1)
- `update_category_description_lead` (2 => 1)
- `update_category_description_moderator` (2 => 1)

Yet the comments above didn't change.

Actually following the extrinsic logic, the following reads should always be executed when `i === 1`:
- `ForumWorkingGroup.CurrentLead` (only in case caller is lead)
- `ForumWorkingGroup.WorkerById`
- `Forum.CategoryById`
- `Forum.CategoryByModerator` (only in case caller is moderator)

Then for each additional category in `category_tree_path` there should be:
- 1 more `Forum.CategoryById` read 
- one more `Forum.CategoryByModerator` read (in case the caller is moderator)

So in `update_category_title_lead` the number of reads should be `2 + i` (same as before), while in `update_category_title_moderator` it should be `2 + 2*i` (the reason it was `2 + i` before is because the benchmarks seem incorrectly written).

I'll investigate it further.

runtime-modules/membership/src/weights.rs
```diff
@@ -92,53 +93,55 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Membership NextMemberId (r:1 w:1)
 	// Storage: Membership ReferralCut (r:1 w:0)
-	fn buy_membership_with_referrer(_i: u32, j: u32, ) -> Weight {
-		(181_305_000 as Weight)
-			// Standard Error: 84_000
-			.saturating_add((1_002_000 as Weight).saturating_mul(j as Weight))
+	fn buy_membership_with_referrer(i: u32, j: u32, ) -> Weight {
+		(62_234_000 as Weight)
+			// Standard Error: 2_000
+			.saturating_add((102_000 as Weight).saturating_mul(i as Weight))
```
Lezek123, line 99: This is too low due to my mistake, in `buy_membership_with_referrer` I updated:
```rust
let handle = handle_from_id::<T>(i * 1000);
```
But I didn't notice this is just referrer's handle. The referee handle is defined further below:
```rust
let second_handle = handle_from_id::<T>(i + 1);
```
It should be changed to:
```rust
let second_handle = handle_from_id::<T>(i * 1000 + 1);
```

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -338,14 +339,12 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-			.saturating_add((492_000 as Weight).saturating_mul(t as Weight))
-			// Standard Error: 52_000
-			.saturating_add((594_000 as Weight).saturating_mul(d as Weight))
+	fn create_proposal_amend_constitution(i: u32, t: u32, _d: u32, ) -> Weight {
+		(121_461_000 as Weight)
+			// Standard Error: 8_000
+			.saturating_add((1_323_000 as Weight).saturating_mul(i as Weight))
+			// Standard Error: 66_000
+			.saturating_add((632_000 as Weight).saturating_mul(t as Weight))
 			.saturating_add(T::DbWeight::get().reads(7 as Weight))
```
Lezek123, line 348: The dependency on `d` was removed. Probably it was just some interference during weight function generation.

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -191,13 +192,13 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-			.saturating_add((1_542_000 as Weight).saturating_mul(i as Weight))
-			// Standard Error: 62_000
-			.saturating_add((504_000 as Weight).saturating_mul(t as Weight))
-			// Standard Error: 62_000
-			.saturating_add((618_000 as Weight).saturating_mul(d as Weight))
+		(112_039_000 as Weight)
+			// Standard Error: 5_000
+			.saturating_add((1_387_000 as Weight).saturating_mul(i as Weight))
+			// Standard Error: 44_000
+			.saturating_add((1_000 as Weight).saturating_mul(t as Weight))
```
Lezek123, line 199: This looks wrong, but probably it was just some interference during weight function generation.

runtime/src/weights/pallet_babe.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fn report_equivocation(validator_count: u32) -> Weight {
+        // we take the validator set count from the membership proof to
+        // calculate the weight but we set a floor of 100 validators.
+        let validator_count = validator_count.max(100) as u64;
+
+        // worst case we are considering is that the given offender
+        // is backed by 1000 nominators
+        const MAX_NOMINATORS: u64 = 1000;
```
Lezek123, line 45: Where does this number come from?

runtime/src/weights/pallet_election_provider_multi_phase.rs
```diff
@@ -0,0 +1,162 @@
...
+			// Standard Error: 95_000
+			.saturating_add((2_907_000 as Weight).saturating_mul(d as Weight))
+			.saturating_add(T::DbWeight::get().reads(7 as Weight))
+			.saturating_add(T::DbWeight::get().writes(1 as Weight))
+	}
+	// Storage: ElectionProviderMultiPhase Round (r:1 w:0)
+	// Storage: ElectionProviderMultiPhase DesiredTargets (r:1 w:0)
+	// Storage: ElectionProviderMultiPhase MinimumUntrustedScore (r:1 w:0)
+	// Storage: ElectionProviderMultiPhase Snapshot (r:1 w:0)
+	fn feasibility_check(v: u32, t: u32, a: u32, _d: u32, ) -> Weight {
```
Lezek123, line 152: It's unfortunate that the dependency on `d` was lost. Polkadot's weights look very different:
```rust
	fn feasibility_check(v: u32, t: u32, a: u32, d: u32, ) -> Weight {
		Weight::from_ref_time(0 as u64)
			// Standard Error: 4_000
			.saturating_add(Weight::from_ref_time(789_000 as u64).saturating_mul(v as u64))
			// Standard Error: 8_000
			.saturating_add(Weight::from_ref_time(41_000 as u64).saturating_mul(t as u64))
			// Standard Error: 13_000
			.saturating_add(Weight::from_ref_time(5_713_000 as u64).saturating_mul(a as u64))
			// Standard Error: 20_000
			.saturating_add(Weight::from_ref_time(1_311_000 as u64).saturating_mul(d as u64))
			.saturating_add(T::DbWeight::get().reads(4 as u64))
```

runtime-modules/forum/src/weights.rs
Lezek123: Created an issue: https://github.com/Joystream/joystream/issues/4403

runtime/src/weights/pallet_babe.rs
mnaamani: I tried to set this higher than default because 200 seemed too small. But its not clear what the best value should be and all substrate runtimes seem to use these defaults.. I found one general reference about babe/grandpa weights here https://substrate.stackexchange.com/questions/1007/benchmarking-generates-files-with-possibly-missing-or-unnecessary-trait-impls/1015#1015

mnaamani:
As agreed we will merge this and apply fixes to the base branch

PR #4400, opened 2022-10-24
Carthage liberated updated

mnaamani:
https://github.com/Joystream/joystream/pull/4371 Updated from carthage

It is expected that integration tests would fail because sudo pallet is dropped. Created an issue to resolve this for mainnet. https://github.com/Joystream/joystream/issues/4401


dobertRowneySr (approved):
LGTM

PR #4381, opened 2022-10-17
Colossus: verify bag <-> bucket assignment

zeeshanakram3:
fixes #4292

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203197954194485) by [Unito](https://www.unito.io)

Issue #4295, opened 2022-09-14
Colossus: validate upload request params before upload is finished


kdembler:
## Background
For various reasons, a Colossus upload request may fail. It may be client failure (wrong params), operator misconfiguration or probably tons of other reasons we haven't seen yet. That will happen from time to time. However, a lot of errors that can occur on the operator side, don't depend on the file being uploaded itself, but on request params (`storageBucketId`, `bagId`, etc.), or external factors (sync status of operator QN, etc.). For this reason, the upload UX could be much improved, if the validation that happens after upload is finished, could happen immediately on request sent. This way, if the upload is doomed to fail anyway, user doesn't have to wait for lengthy upload and can try another operator right away.

## Proposal
Our goal should be to allow upload request validation before the file upload is finished. There are 2 ways I see we could go about this:

1. Change the `/api/v1/files` endpoint - at the moment validation logic that happens here will start only after the file upload. Ideally, we modify the request handling logic to run before the full file is streamed, do validation and short-circuit if it fails and then optionally wait for the upload to finish before marking it as accepted.
1. Add additional endpoint that does the same validation as `/api/v1/files` but doesn't accept any file. This way, request params validation could be done and instruct whether it makes sense to start an actual upload. This could even accept additional params like (`dataObjectSize` and `dataObjectIpfsHash`) to validate whether that data (as seen by client) matches what the operator would expect.

I think option 1. works better overall as we don't require additional request roundtrip, but I'm not sure how easy that is on the technical side, to postpone awaiting file streaming to finish. I'm sure somebody solved this already though. My suggestion is to go for 1. if it doesn't include significant effort increase, otherwise just go with 2.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203011491012080) by [Unito](https://www.unito.io)

bedeho:
Adding to mainnet scope!

bedeho:
Done.

PR #4402, opened 2022-10-25
Colossus prevalidate upload params

zeeshanakram3:
Addresses #4295

Based on #4381 & #4383

### Solution 
This PR moves the parameters that needs to be pre validated (before upload is completed) from the request body to query params i.e., `dataObjectId`, `storageBucketId` and `bagId`. This way the query params can be validated right away in the middleware when the request hits the server. And returning an error if the parameters are not valid, both providing the better upload UX for end users, and saving network resources of the storage node

### DIFF
--- a/cli/src/base/UploadCommandBase.ts
+++ b/cli/src/base/UploadCommandBase.ts
@@ -274,9 +274,6 @@ export default abstract class UploadCommandBase extends ContentDirectoryCommandB
       // cli.action.start('Waiting for the file to be processed...')
     })
     const formData = new FormData()
-    formData.append('dataObjectId', objectId.toString())
-    formData.append('storageBucketId', storageNodeInfo.bucketId)
-    formData.append('bagId', bagId)
     formData.append('file', fileStream, {
       filename: path.basename(filePath),
       filepath: filePath,
--- a/cli/src/base/UploadCommandBase.ts
+++ b/cli/src/base/UploadCommandBase.ts
@@ -285,6 +285,11 @@ export default abstract class UploadCommandBase extends ContentDirectoryCommandB
     this.log(`Uploading object ${objectId.toString()} (${filePath})`)
     try {
       await axios.post(`${storageNodeInfo.apiEndpoint}/files`, formData, {
+        params: {
+          dataObjectId: objectId.toString(),
+          storageBucketId: storageNodeInfo.bucketId,
+          bagId,
+        },
         maxBodyLength: Infinity,
         maxContentLength: Infinity,
         headers: {
--- a/distributor-node/src/commands/dev/batchUpload.ts
+++ b/distributor-node/src/commands/dev/batchUpload.ts
@@ -92,7 +92,7 @@ export default class DevBatchUpload extends AccountsCommandBase {
               url: urljoin(endpoint, 'api/v1/files'),
               data: formData,
               params: {
-                dataObjectId: dataObject.toString(),
+                dataObjectId: dataObjectId.toString(),
                 storageBucketId: bucketId.toString(),
                 bagId,
               },
--- a/distributor-node/src/commands/dev/batchUpload.ts
+++ b/distributor-node/src/commands/dev/batchUpload.ts
@@ -84,9 +84,6 @@ export default class DevBatchUpload extends AccountsCommandBase {
         batch.map(async ([, dataObject], k) => {
           const dataObjectId = nextObjectId + k
           const formData = new FormData()
-          formData.append('dataObjectId', dataObjectId.toString())
-          formData.append('storageBucketId', bucketId.toString())
-          formData.append('bagId', bagId)
           formData.append('file', dataObject, { filename: 'test.jpg', knownLength: dataObject.byteLength })
           this.log(`Uploading object ${dataObjectId}`)
           try {
@@ -94,6 +91,11 @@ export default class DevBatchUpload extends AccountsCommandBase {
               method: 'POST',
               url: urljoin(endpoint, 'api/v1/files'),
               data: formData,
+              params: {
+                dataObjectId: dataObject.toString(),
+                storageBucketId: bucketId.toString(),
+                bagId,
+              },
               headers: {
                 'content-type': 'multipart/form-data',
                 ...formData.getHeaders(),
--- a/storage-node/src/services/webApi/app.ts
+++ b/storage-node/src/services/webApi/app.ts
@@ -17,6 +17,7 @@ import {
   errorLogger,
 } from '../../services/logger'
 import { parseBagId } from '../helpers/bagTypes'
+import BN from 'bn.js'
 
 /**
  * Creates Express web application. Uses the OAS spec file for the API.
@@ -44,10 +45,12 @@ export async function createApp(config: AppConfig): Promise<Express> {
     // Pre validate file upload params
     (req: express.Request, res: express.Response<unknown, AppConfig>, next: NextFunction) => {
       if (req.path === '/api/v1/files') {
-        validateUploadFileParams(req, res).catch((error) => sendResponseWithError(res, next, error, 'upload'))
+        validateUploadFileParams(req, res)
+          .then(next)
+          .catch((error) => sendResponseWithError(res, next, error, 'upload'))
+      } else {
+        next()
       }
-
-      next()
     },
 
     // Setup OpenAPiValidator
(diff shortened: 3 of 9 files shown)


storage-node/src/services/webApi/app.ts
```diff
@@ -38,6 +40,20 @@ export async function createApp(config: AppConfig): Promise<Express> {
 
       next()
     },
+
+    // Pre validate file upload params
+    async (req: express.Request, res: express.Response<unknown, AppConfig>, next: NextFunction) => {
```
Lezek123: The linter is now failing: https://github.com/Joystream/joystream/actions/runs/3319733677/jobs/5485256182#step:4:483.
I think this can be fixed by changing the `async/await` syntax to `Promise.then().catch()`

storage-node/src/services/webApi/app.ts
```diff
@@ -38,6 +40,20 @@ export async function createApp(config: AppConfig): Promise<Express> {
...
+      try {
+        if (req.path === '/api/v1/files') {
+          await validateUploadFileParams(req, res)
+        }
+      } catch (error) {
+        sendResponseWithError(res, next, error, 'upload')
+      }
+
+      next()
```
Lezek123: `next` doesn't need to be called after `sendResponseWithError`

storage-node/src/services/webApi/app.ts
```diff
@@ -186,3 +202,27 @@ function verifyUploadTokenData(accountAddress: string, token: UploadToken, reque
     throw new Error('Nonce not found')
   }
 }
+
+async function validateUploadFileParams(req: express.Request, res: express.Response<unknown, AppConfig>) {
+  const { api, queryNodeEndpoint, workerId } = res.locals
+
+  const storageBucketId = Number(req.query.storageBucketId)
+  const dataObjectId = Number(req.query.dataObjectId)
+  const bagId = req.query.bagId!.toString()
```
Lezek123: It's dangerous to use non-null assertion here
```suggestion
  const bagId = req.query.bagId?.toString() || ''
```

storage-node/src/services/webApi/app.ts
```diff
@@ -186,3 +202,27 @@ function verifyUploadTokenData(accountAddress: string, token: UploadToken, reque
...
+
+  const parsedBagId = parseBagId(bagId)
+
+  const [dataObject] = await Promise.all([
+    api.query.storage.dataObjectsById(parsedBagId, Number(dataObjectId)),
```
Lezek123: `dataObjectId` is already a `Number`:
```suggestion
    api.query.storage.dataObjectsById(parsedBagId, dataObjectId),
```

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -89,24 +89,22 @@ export async function uploadFile(
   res: express.Response<unknown, AppConfig>,
   next: express.NextFunction
 ): Promise<void> {
-  const uploadRequest: RequestData = req.body
+  const uploadRequest: RequestData = req.query as any
```
Lezek123: The use of `any` type is usually a bad idea.
Also the issue here is that `req.query` parameter values are parsed into strings, while `RequestData` defines some of them as numbers.

Lezek123 (changes requested):

Lezek123:
Besides the issues mentioned it seems to work as expected, very nice improvement :+1:

cli/src/base/UploadCommandBase.ts
```diff
@@ -285,6 +285,11 @@ export default abstract class UploadCommandBase extends ContentDirectoryCommandB
     this.log(`Uploading object ${objectId.toString()} (${filePath})`)
     try {
       await axios.post(`${storageNodeInfo.apiEndpoint}/files`, formData, {
+        params: {
+          dataObjectId: objectId.toString(),
+          storageBucketId: storageNodeInfo.bucketId,
+          bagId,
```
Lezek123, line 288: Forgot to mention in the previous review:
1. You can remove those fields from `formData` now
2. You can update `distributor-node/src/commands/dev/batchUpload.ts` in a similar way

zeeshanakram3: Addressed in [b32d4d0](https://github.com/Joystream/joystream/pull/4402/commits/b32d4d0bb53dfb5d4ea80804a85b3b3b58a65776)

storage-node/src/services/webApi/app.ts
zeeshanakram3: addressed in [aa7d7c6](https://github.com/Joystream/joystream/pull/4402/commits/aa7d7c66903226558bf95492578d3ac64eff62e6)

storage-node/src/services/webApi/app.ts
zeeshanakram3: addressed in[aa7d7c6](https://github.com/Joystream/joystream/pull/4402/commits/aa7d7c66903226558bf95492578d3ac64eff62e6)

storage-node/src/services/webApi/app.ts
zeeshanakram3: addressed in[aa7d7c6](https://github.com/Joystream/joystream/pull/4402/commits/aa7d7c66903226558bf95492578d3ac64eff62e6)

storage-node/src/services/webApi/controllers/filesApi.ts
zeeshanakram3: Addressed in [e2953f6](https://github.com/Joystream/joystream/pull/4402/commits/e2953f604e94f85a7dfba2c5729bf6a85d82c7d0)

I have changed the types of all fields in `RequestData` to `string`, to make it consistent

storage-node/src/services/webApi/app.ts
```diff
@@ -38,6 +40,16 @@ export async function createApp(config: AppConfig): Promise<Express> {
...
       next()
     },
+
+    // Pre validate file upload params
+    (req: express.Request, res: express.Response<unknown, AppConfig>, next: NextFunction) => {
+      if (req.path === '/api/v1/files') {
+        validateUploadFileParams(req, res).catch((error) => sendResponseWithError(res, next, error, 'upload'))
+      }
+
+      next()
```
Lezek123: As mentioned before, you need to make sure `next()` is called only once and only after all async jobs have finished
```suggestion
      if (req.path === '/api/v1/files') {
        validateUploadFileParams(req, res)
          .then(next)
          .catch((error) => sendResponseWithError(res, next, error, 'upload'))
      } else {
        next()
      }
```

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -120,20 +119,14 @@ export async function uploadFile(
...
-        `Received already accepted data object. DataObjectId = ${uploadRequest.dataObjectId} WorkerId = ${workerId}`
-      )
-    }
+    await acceptPendingDataObjects(
+      api,
+      bagId,
+      res.locals.storageProviderAccount,
+      workerId,
+      new BN(uploadRequest.storageBucketId),
+      [new BN(uploadRequest.dataObjectId)]
```
Lezek123, line 128: It's not a huge issue, but when doing validation you're using:
```typescript
  const storageBucketId = Number(req.query.storageBucketId)
  const dataObjectId = Number(req.query.dataObjectId)
```
But here you're using `new BN(...)`. There will be some differences in how strings are converted to `Number` vs to `BN`, so ideally you'd use the same method for both the validation & here inside `uploadFile` function.

Lezek123 (changes requested):
Also please bump the `storage-node` version and add a `CHANGELOG.md` with an entry describing those changes as agreed.

distributor-node/src/commands/dev/batchUpload.ts
```diff
@@ -84,16 +84,18 @@ export default class DevBatchUpload extends AccountsCommandBase {
...
-          formData.append('bagId', bagId)
           formData.append('file', dataObject, { filename: 'test.jpg', knownLength: dataObject.byteLength })
           this.log(`Uploading object ${dataObjectId}`)
           try {
             await axios({
               method: 'POST',
               url: urljoin(endpoint, 'api/v1/files'),
               data: formData,
+              params: {
+                dataObjectId: dataObject.toString(),
```
Lezek123:
```suggestion
                dataObjectId: dataObjectId.toString(),
```

Lezek123 (changes requested):

storage-node/src/services/webApi/controllers/filesApi.ts
zeeshanakram3: I have changed types in validation code to `BN` too, since the Number conversion can overflow as JS numbers precision is 53 bits while `storageBucketId` & `dataObjectId` are defined as `u64` in runtime

Lezek123 (approved):
LGTM

PR #4413, opened 2022-10-28
Hydra subscription fix

Lezek123:
Companion for: https://github.com/Joystream/hydra/pull/511


zeeshanakram3 (commented):
I have tested the PR, works as expected. Just one comment the [processor_events_log](https://github.com/lezek123/hydra/blob/9afe516ea754b9cb0163f676dc997207f793ec77/packages/hydra-processor/src/entities/ProcessedEventsLogEntity.ts#L3-L6) entity still exists and is being continuously populated. Shouldn't we just get rid of that?

Lezek123:
> Just one comment the [processor_events_log](https://github.com/lezek123/hydra/blob/9afe516ea754b9cb0163f676dc997207f793ec77/packages/hydra-processor/src/entities/ProcessedEventsLogEntity.ts#L3-L6) entity still exists and is being continuously populated. Shouldn't we just get rid of that?

It's still used by the processor to persist information about its state across runs. For example functions like `loadState` and `countProcessedEvents` rely on its existence.

zeeshanakram3:
> It's still used by the processor to persist information about its state across runs. For example functions like `loadState` and `countProcessedEvents` rely on its existence.

Looks good then. I have merged the corresponding hydra PR and published the packages. Please update this PR now.

Lezek123:
The failing test should be ok after https://github.com/Joystream/joystream/pull/4421

zeeshanakram3 (approved):
LGTM

mnaamani (approved):
The implementation looks correct, based on the new hydra feature. To make use of the new endpoint through our deployment scripts its a simple update to the caddy config files, will need a slightly modified setup on our production nodes with linode load balancer.

.env
```diff
@@ -55,6 +57,9 @@ DISTRIBUTOR_QUERY_NODE_URL=http://graphql-server:8081/graphql
 # Indexer gateway used by processor. If you don't use the local indexer set this to a remote gateway
 PROCESSOR_INDEXER_GATEWAY=http://hydra-indexer-gateway:4000/graphql
 
+# State update endpoint used by prcessor (to send state updates)
+STATE_UPDATE_ENDPOINT=http://graphql-server:8082/update-processor-state
```
mnaamani, line 61: Doesn't this need to be passed in env for the processor? (As in the e2e test setup in hydra repo)
Or is it being passed through .env file through use of the `env_file` directive ?

```
env_file:
      # relative to working directory where docker-compose was run from
      - .env
```

mnaamani: the default using 'localhost' probably doesn't work on mac with docker desktop..

```
    STATE_UPDATE_ENDPOINT: str({
      devDefault: 'http://localhost:8082/update-processor-state',
    }),
```

mnaamani:
@Lezek123 How can I test the endpoint? When I do a simple `GET` request I get an error logged:

```
[ 'Unauthorized access on /update-processor-state: 172.18.0.1' ]
```

.env
Lezek123: > Or is it being passed through .env file through use of the env_file directive ?

Yes, it should be passed because of this directive, it doesn't need to be explicitly named in `environment`

> the default using 'localhost' probably doesn't work on mac with docker desktop..

I'm aware it doesn't work in some setups, but it's just the default. I followed the approach already used for indexer endpoint, ie.: `INDEXER_ENDPOINT_URL: str({ devDefault: 'http://localhost:4001' }),`

Lezek123:
> @Lezek123 How can I test the endpoint? When I do a simple GET request I get an error logged:

This is an additional security measure, so that only the processor is allowed to send update requests to this endpoint.
Also only `POST` requests are supported by this endpoint, as it is just for pushing the updates, not for receiving any data.

If you want to test pushing from localhost (via Postman app for example) you can set `PROCESSOR_HOST=localhost` in `.env` (in this case the processor will not be able to push any updates, but you will be able to push from `localhost`).

mnaamani:
> > @Lezek123 How can I test the endpoint? When I do a simple GET request I get an error logged:
> 
> This is an additional security measure, so that only the processor is allowed to send update requests to this endpoint. Also only `POST` requests are supported by this endpoint, as it is just for pushing the updates, not for receiving any data.
> 
> If you want to test pushing from localhost (via Postman app for example) you can set `PROCESSOR_HOST=localhost` in `.env` (in this case the processor will not be able to push any updates, but you will be able to push from `localhost`).

Thanks for clarification, I ran some tests and works as expected:

configured `STATE_UPDATE_ENDPOINT=http://host.docker.internal:8090/` 

```txt
$ nc -l 8090
POST / HTTP/1.1
Accept: application/json, text/plain, */*
Content-Type: application/json
User-Agent: axios/1.1.3
Content-Length: 110
Accept-Encoding: gzip, deflate, br
Host: host.docker.internal:8090
Connection: close

{"state":{"indexerHead":8,"chainHead":14,"lastScannedBlock":1,"lastProcessedEvent":"0000000001-000002-c112c"}}
```

PR #4417, opened 2022-10-31
Carthage Deployment Scripts

mnaamani:
- Include git commit hash and runtime code shasum to the node version string when doing docker build. We do this manually by setting env variable as we no longer copy the entire .git repo into the docker build context and the rust custom build fails to get git commit.
- Fixed bug in chainspec builder "missing" `//` prefix on seeds when generating new random seeds. (generate sub command)
- Added support for the chainspec builder `new` subcommand to pass either seeds or full validator accounts in `--authorities` argument. More secure than providing seeds on deployment host.
- Added a utility `session-keys` to create validator session keys in a keystore. This allows keys to be generated on individual validators, and deployment scripts do not need to fetch from single host and push to validators. More secure setup.

- Updated ansible deployment scripts to support adding separate boot nodes, instead of using validators as boot nodes, and using session-keys to generate keys on validators. Other minor refactoring how node keys and keystore are stored on target hosts.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1203291205609117) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1202132419573087/1203291205609117


dobertRowneySr (approved):
Rust part LGTM

PR #4421, opened 2022-11-02
Fix occasionally failing "video comments and reactions" test

Lezek123:
SSIA

### DIFF
--- a/tests/network-tests/src/fixtures/content/commentAndReaction/CreateCommentsFixture.ts
+++ b/tests/network-tests/src/fixtures/content/commentAndReaction/CreateCommentsFixture.ts
@@ -63,7 +63,8 @@ export class CreateCommentsFixture extends StandardizedFixture {
     qComments: CommentFieldsFragment[],
     qEvents: CommentCreatedEventFieldsFragment[]
   ): void {
-    qEvents.map((qEvent, i) => {
+    this.events.map((e, i) => {
+      const qEvent = this.findMatchingQueryNodeEvent(e, qEvents)
       const qComment = qComments.find((comment) => comment.id === qEvent.comment.id.toString())
       const commentParams = this.commentsParams[i]
       Utils.assert(qComment, 'Query node: Comment not found')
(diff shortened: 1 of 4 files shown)


tests/network-tests/src/fixtures/content/commentAndReaction/CreateCommentsFixture.ts
```diff
@@ -63,7 +63,8 @@ export class CreateCommentsFixture extends StandardizedFixture {
     qComments: CommentFieldsFragment[],
     qEvents: CommentCreatedEventFieldsFragment[]
   ): void {
-    qEvents.map((qEvent, i) => {
+    this.events.map((e, i) => {
+      const qEvent = this.findMatchingQueryNodeEvent(e, qEvents)
```
zeeshanakram3, line 67: 1. Does this change entails that previously some values of `qEvent` were wrong? Logically this change seems to be doing the same thing; in fact, redundant as [qEvents](https://github.com/lezek123/substrate-runtime-joystream/blob/94d59bc1f757a47ad9c40c548eae71b7b3ed0913/tests/network-tests/src/fixtures/content/commentAndReaction/CreateCommentsFixture.ts#L87-L88) is basically a filtered list of all CommentCreatedEvents.
2. If this fixes the integration test, then do we need to apply similar changes to other fixtures, e.g., [EditCommentFixture](https://github.com/lezek123/substrate-runtime-joystream/blob/94d59bc1f757a47ad9c40c548eae71b7b3ed0913/tests/network-tests/src/fixtures/content/commentAndReaction/EditCommentsFixture.ts#L65-L66), [DeleteCommentFixture](https://github.com/lezek123/substrate-runtime-joystream/blob/94d59bc1f757a47ad9c40c548eae71b7b3ed0913/tests/network-tests/src/fixtures/content/commentAndReaction/DeleteCommentsFixture.ts#L64-L65), etc?

Lezek123: 1. The difference is that `qEvents` are not sorted in a deterministic way, while `this.events` are, so if you iterate over `this.events`, you can be sure that you will get correct params when accessing `this.commentsParams[i]`
2.  Good point, I didn't notice there were other fixtures with this mistake, will fix

zeeshanakram3 (approved):
LGTM

mnaamani (approved):

PR #4407, opened 2022-10-26
Argus/Colossus and CLI create/update video tests

Lezek123:
- Addresses https://github.com/Joystream/joystream/issues/4404
- Fixes a few issues in the QN mappings and CLI that were spotted thanks to the new tests

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203244604665451) by [Unito](https://www.unito.io)

PR #4425, opened 2022-11-07
typegen: generate all schemas

mnaamani:
- updated chain metadata, ran yarn bulid in types packages
- ran yarn generate:all in cli,tests,storage and distributor node packages.


Lezek123 (approved):
LGTM, there is a small diff in `/distributor-node/src/services/networking/storage-node/generated/api.ts`, but it doesn't have any effect, since the distributor node is not using this api method and it will be fixed anyway by https://github.com/Joystream/joystream/pull/4407

mnaamani:
> LGTM, there is a small diff in `/distributor-node/src/services/networking/storage-node/generated/api.ts`, but it doesn't have any effect, since the distributor node is not using this api method and it will be fixed anyway by #4407

Nice that you caught that, although the integration tests don't cover colossus and argus api, if anything had broken it would have been caught by the linter I presume.

Issue #4406, opened 2022-10-26
CRT Runtime Tracking Issue


bedeho:
This is a long running tracking issue for ongoing work on `project_token` pallet and related work. It's not yet clear when this gets merged, so we just keep pulling in `carthage`& `mainnet` work until further notice, and we need to start some new branch, say `project_token_runtime`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203242393499031) by [Unito](https://www.unito.io)

bedeho:
Here is a summary of current design: https://gist.github.com/bedeho/e32f117d09c5342ce62d586fd6d5b67b

bedeho:
# 28th October Meeting Summary

## Conclusions
- We agreed that arbitrage that arises from patronage, burning and other suppply dynamics, are acceptable, in the sense that we will not try to directly acomodate them in the AMM in some specific way. The AMM, as in the current working design.
- We had a discussion that perhaps the price floor rigidity that may exist as a result of `f(0) = b` being greater than the price that comes from the fundamental value of a $CRT is not a real problem in terms of potentially preventing liquidity from existing. This could for example happen if `$JOY` goes up very high. The reason this is not a real problem is because AMM can be relaunched with new more efficient parameters. Hence there is no immediate need for some sort of governance based influence on `a` or `b`.
- We discussed how channel feature pausing does allow for whole sale either blocking or allowing CRT issuing by a new default channel, but currently the default is to allow. It is unclear whether this should be the way the platform can regulate access to AMM/public sale launching, or if there should be a new feature for this, e.g. by tying it to runtime-aware notion of being "verified". The benefit of this is that you could have a global policy of restricting or not restricting ability to use such features to this caetgory over all, without having to go through channels one by one to pause and unpause. 

## TODO
- Ignazio will review the prior user research thread  and see if we have not missed something important: https://github.com/Joystream/atlas/issues/2728
- Ignazio will focus on AMM next, and will first specify an precise formulation of how creator fees and slippage controls will be enforced, this will help share understanding in the team, and review. Post to separate issue.

dobertRowneySr:
# Slippage 
Slippage is defined as the difference between the price at which a order is submitted and the price at which the order is filled.
[...]

Issue #4151, opened 2022-08-10
Content directory imposed CRT constraints


bedeho:
# Background
One of the main issues with CRTs on DeSo was lack of long term alignment between issuer and buyers. Unfortunately for them, they did not have vesting, which we do, but we can do even better: governance.

# Proposal
Allow lead, or council, to set global constraints on CRTs that can be issued, e.g.

* max daily number
* only channels that are verified (is this in state?) and/or some how at a given tier (permission levele?)
* min vesting schedule for issuer
* whether permission less demonstrability is ok
* max patronage
* etc.

This is not a spec, it's an idea, implementer should come up with concrete design before starting.

Probably also constraints on [sale terms](https://github.com/Joystream/joystream/issues/2362#issuecomment-1059012676) and other parameters is also sensible. Implementer should provide exhaustive list of constraints.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1202777295294353) by [Unito](https://www.unito.io)

dobertRowneySr:
| Parameter Name                                             | Governance Constraint required | Comments                                                   |
| ---------------------------------------------------------- | ------------------------------ | ---------------------------------------------------------- |
| Revenue split - timeline duration                          | ❌                             |                                                            |
| Revenue split - max split value                            | ❌                             |                                                            |
| Vesting schedule - cliff amount percentage max value       | ❌                             |                                                            |
| Vesting schedule - max duration                            | ❌                             |                                                            |
| Vesting schedule - max blocks before cliff                 | ❌                             |                                                            |
| Token sale - max CRT amount that can be issued in a sale   | ❌                             |                                                            |
| Token sale - max sale duration                             | ❌                             |                                                            |
| Token sale - cap per member                                | ❌                             |                                                            |
| Bonding curve - max value for $a, b$ params                | ✅                             |  100% value for both $a,b$ is maximum                                                          |
| Bonding curve - max reward account                         | ❌                             | Setting 100% value allows the CC to block issuance via amm |
| Content directory - max amount of daily CRT issued         | ✅                             | Consistent with NFT, avoid proliferation of low quality tokens                                      |
| Content directory - only verified channels can issue token | ✅                             | I have listed several reasons below                        |
| Bonding curve - overall constraints to prevent Rugs        | ✅                             | See below                                                  |

## Why we should allow CRT token only by  verified users
- Minimize the chances of imposters & scams
- Platform credibility & quality
[...]

Issue #3754, opened 2022-05-16
AMM Design Discussion


bedeho:
# Background

Great work with this overview @ignazio-bovo 

https://hackmd.io/@msgSgRNqSKmkYgN4EoybcA/BysDdI8Iq

# First thoughts

1. It seems to me the best tradeoff here would be something that never is really bad, but works Ok in a wide range of circumstances, which would be `Linear`? It's also way easier for people to understand.
2. When it says that `Initial Liquidity bootstrapping capabilities` are _very_ good, does this not mean that its simply really cheaply selling off the tokens to buyers, which is at the expense of the issuer? 
3. I think we should just pick a single type of curve for our first implementation, adding additional curves in the future will be very easy, no migration needed, and it will be with the benefit of more concrete information.
4. Ok, so a tax on both AMM transaction types.
5. Price based slippage constraint seems most direct and simple, but is it the same as is used on say Uniswap? may be wise to just stick with their exact same version of slippage tolerance, as people already know this.

dobertRowneySr:
> It seems to me the best tradeoff here would be something that never is really bad, but works Ok in a wide range of circumstances, which would be Linear? It's also way easier for people to understand.
> I think we should just pick a single type of curve for our first implementation, adding additional curves in the future will be very easy, no migration needed, and it will be with the benefit of more concrete information.

Linear would stand somewhat in the middle of all described tradeoffs. I think it is a good choice for initial experimentation since it is not largely exposed to one risk in particular

> When it says that Initial Liquidity bootstrapping capabilities are very good, does this not mean that its simply really cheaply selling off the tokens to buyers, which is at the expense of the issuer?

No I used that term to specify that price growth is the fastest in the beginning of a project. So early investors can benefits of larger price upswings with respect to early investors. This is effectively an incentive to buy early rather than late. This effectively doesn't keep CRT price low for long, if selling initial CRT for cheap is desired polynomial curves are more suitable.

> Price based slippage constraint seems most direct and simple, but is it the same as is used on say Uniswap? may be wise to just stick with their exact same version of slippage tolerance, as people already know this.

Yes Uniswap, Pancakeswap and similar (constant product market makers) use the same slippage protection mechanism, namely
tolerance % + execution deadline. (Uniswap allows you to set these parameters in expert mode)

bedeho:
# Background

The following research

https://github.com/joystream/atlas/issues/2728

# Conclusions

[...]

Issue #3630, opened 2022-04-19
Compelte `project_token` pallet AMM 


bedeho:
Remember: Similar to how NFT fees are collected. Fee itself comes from a global storage parameter.

This is the most latest spec as far as I recall: https://github.com/Joystream/joystream/issues/3832

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201738706333715/1202146640573021) by [Unito](https://www.unito.io)

PR #4434, opened 2022-11-11
amm feature for the CRT pallet

dobertRowneySr:
Adds the AMM (also referred as Bonding Curve) feature to the project token pallet https://github.com/Joystream/joystream/issues/3630

# Features
- [X] `activate, deactivate, bond, unbond` extrinsics
- [X] slippage protection mechanisms: tolerance + deadline
- [X] user can bond and have its account immediately created with bloat bond deposited
- [X] linear pricing formula `ax + b` with, parameters arbirarely chosen by the user
- [ ] chose correct type for parameters `Permill` vs `Perquintill` (all references seems to point out that Permill accuracy should be sufficient)
- [x] added extrinsics to the content pallet for amm activation/deactivation w/ tests 
- [ ] Rug pulling protection mechanisms (~~to be discussed with bedeho?~~ part of the #4151 )
- [X] Doc lines above new extrinsics

# References
- For the choices of parameters `a = .003,b = 0` I used DeSo https://docs.deso.org/about-deso-chain/readme#the-creator-coin-supply-curve
- https://github.com/Joystream/joystream/issues/4406
- https://github.com/Joystream/joystream/issues/3754
- https://www.trality.com/blog/what-is-slippage-in-crypto (suggests that average slippage tolerance is typically in the 2-10% range)

# Computation summary (for the `eval` function)
Relation between JOY and CRT to:
## Bond
$$ \int_X^{X + x} (at + b) dt = \frac{a}{2} [(X + x)^2 - X^2] + b x $$
## Unbond
$$ \int_X^{X - x} (at + b) dt = - \bigg( \frac{a}{2} [X^2 - (X - x)^2] + b x  \bigg) = -\text{JOYs necessary for unbonding}$$
- $X$ = `token_info.total_supply` value before calling the extrinsic
- $x$ = `amount` parameter in the extrinsic

I have expanded the computation into 3 terms to avoid chances of overflowing for big numbers like $X^2$


runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,34 @@ decl_error! {
...
+        /// not in AMM state
+        NotInAmmState,
+
+        /// Invalid bonding curve construction parameters
+        InvalidCurveParameters,
+
+        /// Deadline timestamp lower than now
+        InvalidTimestampSpecified,
+
+        /// Deadline constraint not respected
```
Lezek123:
```suggestion
        /// Deadline constraint not satisfied
```

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,34 @@ decl_error! {
...
+        DeadlineExpired,
+
+        /// SLippage tolerance constraint tolerance not respected
```
Lezek123:
```suggestion
        /// Slippage tolerance constraint tolerance not satisfied
```

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,34 @@ decl_error! {
...
+        SlippageToleranceExceeded,
+
+        /// SLippage tolerance constraint tolerance not respected
+        UnsufficientTokenAmount,
```
Lezek123:
```suggestion
        /// Creator token balance is insufficient 
        InsufficientTokenBalance,
```

runtime-modules/project-token/src/events.rs
```diff
@@ -155,5 +155,34 @@ decl_event! {
         /// - member id
         /// - number of tokens burned
         TokensBurned(TokenId, MemberId, Balance),
+
+        /// BondingCurve activated
+        /// Params:
+        /// - token id
+        /// - member id
+        /// - params for the bonding curve
+        BondingCurveActivated(TokenId, MemberId, BondingCurve),
```
Lezek123: My general feeling about this naming strategy (both for events and extrinsics) is that it's very difficult to parse for multiple reasons:
- bonding/unbonding may refer to many things and is often used in context of staking (for example, when bonding to nominate/validate).
- It's not clear what you actually `bond`. Is it JOY? But then again, the event is called `TokenBonded`, even though when you call `bond` you actually deposit JOY (and mint the token).

What do you think about using more descriptive names like:
- `AMMActivated`
- `TokensBoughtOnAMM` (you deposit JOY and recieve tokens)
- `TokensSoldOnAMM` (you deposit (burn) tokens and recieve JOY)
- `AMMDeactivated`

(consistently using this naming for extrinsics/tests/constants etc. too)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// - (origin, member_id) must be a valid authentication pair
+        /// - token_id must exist
+        /// - offering state for `token_id` must be `Idle`
+        ///
+        /// Postconditions
+        /// - token `bonding_curve` activated with specified parameters
+        /// - amm account created with existential deposit (if necessary)
+        /// - event deposited
+        #[weight = 100_000_000] // TODO: adjust weight
+        fn activate_amm(origin, token_id: T::TokenId, member_id: T::MemberId, curve: BondingCurve) -> DispatchResult {
```
Lezek123: This shouldn't be an extrinsic, instead it should be a method executed by the content pallet, just like `init_token_sale` for example.

The content pallet should perform the authentication, because it must check that the caller is channel owner.

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            T::MemberOriginValidator::ensure_member_controller_account_origin(
+                origin,
+                member_id
+            )?;
+            let token_data = Self::ensure_token_exists(token_id)?;
+
+            ensure!(OfferingStateOf::<T>::ensure_idle_of::<T>(&token_data).is_ok(), Error::<T>::TokenIssuanceNotInIdleState);
```
Lezek123: This is redundant, `ensure_idle_of` already returns `Error::<T>::TokenIssuanceNotInIdleState` on failure:
```suggestion
            OfferingStateOf::<T>::ensure_idle_of::<T>(&token_data)?;
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            if Joy::<T>::usable_balance(&amm_treasury_account).is_zero() {
+                let _ = Joy::<T>::deposit_creating(&amm_treasury_account, T::JoyExistentialDeposit::get());
+            }
+
+            Self::deposit_event(RawEvent::BondingCurveActivated(token_id, member_id, curve));
+
+            Ok(())
+        }
+
+        /// Activate Amm functionality for the token
```
Lezek123: Invalid comment for `fn bond`

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// Preconditions
+        /// - origin, member_id pair must be a valid authentication pair
+        /// - token_id must exist
+        /// - user usable JOY balance must be enough for bonding (+ existential deposit eventually)
```
Lezek123:
```suggestion
        /// - user usable JOY balance must be enough for bonding (+ existential deposit)
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// - deadline constraint respected if provided
+        /// - slippage tolerance constraints respected if provided
+        /// - token total supply and amount value must be s.t. `eval` function doesn't overflow
+        ///
+        /// Postconditions
+        /// - token `bonding_curve` activated with specified parameters
```
Lezek123: Invalid postcondition for `fn bond`

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// - `amount` CRT minted into account (which is created if necessary and existential deposit added to it)
```
Lezek123:
```suggestion
        /// - `amount` CRT minted into account (which is created if necessary with existential deposit transferred to it)
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+                price.saturating_add(bloat_bond)
+            } else {
+                price
+            };
+
+            Self::ensure_can_transfer_joy(&sender, joys_required)?;
+
+            // slippage tolerance check
+            if let Some((slippage_tolerance, desired_price)) = slippage_tolerance {
+                ensure!(price.saturating_sub(desired_price) < slippage_tolerance.mul_floor(desired_price), Error::<T>::SlippageToleranceExceeded);
```
Lezek123: I think this would be more correct
```suggestion
                ensure!(price.saturating_sub(desired_price) <= slippage_tolerance.mul_floor(desired_price), Error::<T>::SlippageToleranceExceeded);
```
(For example, with `price = 100`, `slippage_tolerance = 0.02`, `102` should still be an acceptable price)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            }
+
+            // timestamp deadline check
+            if let Some(deadline) = deadline {
+                ensure!(<timestamp::Pallet<T>>::get() < deadline, Error::<T>::DeadlineExpired);
```
Lezek123:
```suggestion
                ensure!(<timestamp::Pallet<T>>::get() <= deadline, Error::<T>::DeadlineExpired);
```
(same as with `slippage_tolerance`, for consistency)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            }
+
+            // == MUTATION SAFE ==
+
+            let creator_member_id = token_data.creator_member_id;
```
Lezek123: Keeping `creator_member_id` in `token_data` is not a good idea, because the creator token should be effectively owned by the current channel owner (which is subject to change).

If we take _Sale_ as an example, each time a sale is initialized, the content pallet passes `earnings_dst` to `T::ProjectToken::init_token_sale`, and the `earnings_dst` always depends on the current channel owner.

Additionally, **one cannot transfer a channel while creator token sale is ongoing**. A similar mechanism would be needed for AMM (_see `Content::initialize_channel_transfer`_).

This way we can ensure than only the current channel owner profits from a creator token Sale/AMM.

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            let creator_amount = curve.creator_reward.mul_floor(amount);
+            if AccountInfoByTokenAndMember::<T>::contains_key(token_id, creator_member_id) {
+                AccountInfoByTokenAndMember::<T>::mutate(token_id, creator_member_id, |account_data| {
+                    account_data.amount = account_data.amount.saturating_add(creator_amount)
```
Lezek123:
```suggestion
                    account_data.increase_amount_by(creator_amount);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+                });
+            }
+
+            let user_amount = amount - creator_amount;
```
Lezek123:
```suggestion
            let user_amount = amount.saturating_sub(creator_amount);
```
That's a shame that `deny(clippy::integer_arithmetic)` doesn't catch that :/
We will probably need to use https://github.com/rust-lang/rust-clippy/pull/9130 in the furute.

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+                            user_amount,
+                            // No restrictions on repayable bloat bond,
+                            // since only usable balance is allowed
+                            RepayableBloatBond::new(bloat_bond, None)
+                    );
+                Self::do_insert_new_account_for_token(token_id, &member_id, new_account_info);
+                Self::transfer_joy(&sender, &Self::module_treasury_account(), bloat_bond)?;
+            } else {
+                AccountInfoByTokenAndMember::<T>::mutate(token_id, member_id, |account_data| {
+                    account_data.amount = account_data.amount.saturating_add(user_amount);
```
Lezek123:
```suggestion
                    account_data.increase_amount_by(user_amount);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+                token_data.tokens_issued = token_data.tokens_issued.saturating_add(amount);
+            });
+            Self::transfer_joy(&sender, &amm_reserve_account, price)?;
+
+            Self::deposit_event(RawEvent::TokenBonded(token_id, member_id, amount, price));
+
+            Ok(())
+        }
+
+        /// Mint token and transfer joys to the AMM
```
Lezek123: Invalid comment for `fn unbond`

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// - token_id must exist
+        /// - token_id, member_id must be valid account coordinates
+        /// - user usable CRT balance must be at least `amount`
+        /// - deadline constraint respected if provided
+        /// - slippage tolerance constraints respected if provided
+        /// - token total supply and amount value must be s.t. `eval` function doesn't overflow
+        /// - amm treasury acconut must have sufficient JOYs for the operation
+        ///
+        /// Postconditions
+        /// - `amount` burn from user account
```
Lezek123:
```suggestion
        /// - `amount` burned from user account
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+
+            let amm_reserve_account = Self::module_bonding_curve_reserve_account(token_id);
+
+            let price = curve.eval::<T>(amount, token_data.total_supply, BondOperation::Unbond)?;
+
+            Self::ensure_can_transfer_joy(&amm_reserve_account, price)?;
+
+            // slippage tolerance check
+            if let Some((slippage_tolerance, desired_price)) = slippage_tolerance {
+                ensure!(desired_price.saturating_sub(price) < slippage_tolerance.mul_floor(desired_price), Error::<T>::SlippageToleranceExceeded);
```
Lezek123:
```suggestion
                ensure!(desired_price.saturating_sub(price) <= slippage_tolerance.mul_floor(desired_price), Error::<T>::SlippageToleranceExceeded);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            }
+
+            // timestamp deadline check
+            if let Some(deadline) = deadline {
+                ensure!(<timestamp::Pallet<T>>::get() < deadline, Error::<T>::DeadlineExpired);
```
Lezek123:
```suggestion
                ensure!(<timestamp::Pallet<T>>::get() <= deadline, Error::<T>::DeadlineExpired);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            }
+
+            // == MUTATION SAFE ==
+
+            AccountInfoByTokenAndMember::<T>::mutate(token_id, member_id, |account_data| {
+                account_data.amount = account_data.amount.saturating_sub(amount)
```
Lezek123:
```suggestion
                account_data.decrease_amount_by(amount);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            });
+
+            TokenInfoById::<T>::mutate(token_id, |token_data| {
+                token_data.total_supply = token_data.total_supply.saturating_sub(amount);
```
Lezek123:
```suggestion
                token_data.decrease_supply_by(amount);
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            T::MemberOriginValidator::ensure_member_controller_account_origin(
+                origin,
+                member_id
+            )?;
+
+            let token_data = Self::ensure_token_exists(token_id)?;
+            let curve = token_data.bonding_curve.ok_or(Error::<T>::NotInAmmState)?;
+            let user_acc_data = Self::ensure_account_data_exists(token_id, &member_id)?;
+
+            ensure!(user_acc_data.amount >= amount, Error::<T>::UnsufficientTokenAmount);
```
Lezek123: To prevent vesting funds leakage you should check `transferrable` balance here:
```suggestion
            ensure!(
                user_acc_data.transferrable::<T>(Self::current_block()) >= amount,
                Error::<T>::InsufficientTokenAmount
            );
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+                Self::transfer_joy(&sender, &Self::module_treasury_account(), bloat_bond)?;
+            } else {
+                AccountInfoByTokenAndMember::<T>::mutate(token_id, member_id, |account_data| {
+                    account_data.amount = account_data.amount.saturating_add(user_amount);
+                });
+            }
+
+            TokenInfoById::<T>::mutate(token_id, |token_data| {
+                token_data.total_supply = token_data.total_supply.saturating_add(amount);
+                token_data.tokens_issued = token_data.tokens_issued.saturating_add(amount);
```
Lezek123:
```suggestion
                token_data.increase_supply_by(amount);
```
(already does both)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+        /// - (origin, member_id) must be a valid authentication pair
+        /// - token_id must be a valid
+        /// - token must be in `BondigCurve` state
+        ///
+        /// Postconditions
+        /// - Bonding Curve set to None
+        /// - state set to idle
+        /// - event deposited
+        #[weight = 10_000_000] // TODO: Adjust weight
+        fn deactivate_amm(origin, token_id: T::TokenId, member_id: T::MemberId) -> DispatchResult {
```
Lezek123: Just as `activate_amm` - this shouldn't be an extrinsic (for the same reason)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+            T::MemberOriginValidator::ensure_member_controller_account_origin(
+                origin,
+                member_id
+            )?;
+
+            let token_data = Self::ensure_token_exists(token_id)?;
+
+            ensure!(token_data.creator_member_id == member_id, Error::<T>::UserNotAuthorized);
```
Lezek123: This will go away if we shift owner authentication to `content` pallet (which is where it should be done)

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+
+            ensure!(OfferingStateOf::<T>::ensure_bonding_curve_of::<T>(&token_data).is_ok(), Error::<T>::NotInAmmState);
```
Lezek123: `ensure_bonding_curve_of` already returns `Error::<T>::NotInAmmState` on failure:
```suggestion
            OfferingStateOf::<T>::ensure_bonding_curve_of::<T>(&token_data)?;
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +801,239 @@ decl_module! {
...
+
+            let amm_treasury_account = Self::module_bonding_curve_reserve_account(token_id);
+            ensure!(Joy::<T>::usable_balance(amm_treasury_account) == T::JoyExistentialDeposit::get(), Error::<T>::AmmTreasuryBalanceNotEmpty);
```
Lezek123: In order to satisfy this condition, as many tokens as were ever bought on this AMM would need to be sold, which may be impossible. I think there should be some alternative way of closing the AMM (perhaps with some minimum deadline, after which all remaining JOYs are burned?)

runtime-modules/project-token/src/lib.rs
```diff
@@ -1619,6 +1853,11 @@ impl<T: Config> Module<T> {
         <T as Config>::ModuleId::get().into_sub_account_truncating(Vec::<u8>::new())
     }
 
+    /// Returns teh account for the bonding curve treasury
+    pub fn module_bonding_curve_reserve_account(token_id: T::TokenId) -> T::AccountId {
+        <T as Config>::ModuleId::get().into_sub_account_truncating(&("BONDING_RESERVE", token_id))
```
Lezek123: It's risky to use long strings here (longer than 10-12 bytes), because `into_sub_account_truncating` will:
- Add a prefix which is 12 bytes
- Concat it with your input (`BONDING_RESERVE` string itself is already 15 bytes)
- Turncate the result to match `AccountId` length (which is 32 bytes)

If the input is too long, an important part of it (like `token_id` in this case) may get turncated and this will cause huge issues.

How about:
```suggestion
        <T as Config>::ModuleId::get().into_sub_account_truncating(&("AMM", token_id))
```

runtime-modules/project-token/src/types.rs
```diff
@@ -569,6 +575,57 @@ where
     }
 }
 
+/// Represents token's bonding curve with linear pricing function y = ax + b
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Default, Encode, Decode, TypeInfo, Clone, Debug, Eq, PartialEq, MaxEncodedLen)]
+pub struct BondingCurve {
+    /// Slope parameter : a
+    pub slope: Permill,
```
Lezek123, line 591: This prevents a slope from being > 1, which is very limiting

runtime-modules/project-token/src/types.rs
```diff
@@ -569,6 +575,57 @@ where
...
+
+    /// Intercept : b
+    pub intercept: Permill,
```
Lezek123, line 594: Same problem as with the slope

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                .with_member_id(user_member_id)
+                .execute_call()
+                .unwrap();
+
+            let account_number_post = Token::token_info_by_id(token_id).accounts_number;
+            let account_data =
+                Token::ensure_account_data_exists(token_id, &user_member_id).unwrap();
+            let user_amount = BONDING_CURVE_CREATOR_REWARD
+                .left_from_one()
+                .mul_floor(DEFAULT_BONDING_AMOUNT);
```
Lezek123: That's not always going to match `amount - creator_amount`.
For example, imagine `amount = 101` and `BONDING_CURVE_CREATOR_REWARD = 1%`.
Then:
```
creator_amount = 1
amount - creator_amount = 100
(0.99).mul_floor(amount) = 99
```

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                .with_member_id(user_member_id)
+                .execute_call()
+                .unwrap();
+
+            let user_amount_post = Token::ensure_account_data_exists(token_id, &user_member_id)
+                .unwrap()
+                .amount;
+            let user_amount_increase = BONDING_CURVE_CREATOR_REWARD
+                .left_from_one()
+                .mul_floor(DEFAULT_BONDING_AMOUNT);
```
Lezek123: Same comment as before: this doesn't always match `amount - creator_amount`

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+            BondFixture::default()
+                .with_amount(DEFAULT_BONDING_AMOUNT)
+                .with_sender(user_account_id)
+                .with_member_id(user_member_id)
+                .execute_call()
+                .unwrap();
+
+            let amm_reserve_post = Balances::usable_balance(amm_reserve_account);
+            let correctly_computed_joy_amount = 1000500;
```
Lezek123: Some explaination (comment) of why and under what conditions (ie. assuming `DEFAULT_BONDING_AMOUNT=x`) this is the correct amount would be advised

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+        assert_err!(result, Error::<Test>::InsufficientJoyBalance);
+    })
+}
+
+#[test]
+fn user_joy_balance_correctly_decreased_during_bonding() {
+    let (_, user_account) = member!(2);
+    let (user_account_id, user_balance) = (member!(2).1, joy!(5_000_000));
+    let correctly_computed_joy_amount = 1000500;
```
Lezek123: Same comment as before

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+            let user_reserve_post = Balances::usable_balance(user_account);
+            assert_eq!(
+                user_reserve_pre - user_reserve_post,
+                correctly_computed_joy_amount
+            );
+        })
+}
+
+#[test]
+fn crt_correctly_minted_to_user_during_bonding() {
```
Lezek123: Isn't this test effectively repeating the assertions that are already part of `bonding_succeeds_with_existing_user` / `bonding_succeeds_with_new_user`?

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                DEFAULT_BONDING_AMOUNT - creator_amount
+            );
+        })
+}
+
+#[test]
+fn bond_ok_with_event_deposit() {
+    let token_id = token!(1);
+    let ((user_member_id, user_account_id), user_balance) = (member!(2), joy!(5_000_000));
+    let price = joy!(1000500);
```
Lezek123: Same as before - a comment explaining where `1000500` comes from would be advised

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                .execute_call();
+
+            let state_post = sp_io::storage::root(sp_storage::StateVersion::V1);
+            assert_ok!(result);
+            assert_eq!(state_pre, state_post);
+        })
+}
+
+#[test]
+fn unbond_fails_with_user_not_having_enought_crt() {
```
Lezek123: After fixing the vesting leak issue, it'll be a good idea to add test for that too (ie. vesting funds cannot be used to "unbond")

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+            UnbondFixture::default()
+                .with_amount(DEFAULT_UNBONDING_AMOUNT)
+                .with_sender(user_account_id)
+                .with_member_id(user_member_id)
+                .execute_call()
+                .unwrap();
+
+            let amm_reserve_post = Balances::usable_balance(amm_reserve_account);
+            let correctly_computed_joy_amount = 100095;
```
Lezek123: Please add a comment describing where this value comes from

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+            assert_err!(result, Error::<Test>::InsufficientJoyBalance);
+        })
+}
+
+#[test]
+fn unbonding_ok_with_user_joy_balance_correctly_increased() {
+    let (_, user_account) = member!(2);
+    let (user_account_id, user_balance) = (member!(2).1, joy!(5_000_000));
+    build_default_test_externalities_with_balances(vec![(user_account_id, user_balance)])
```
Lezek123:
```suggestion
    let (user_account, user_balance) = (member!(2).1, joy!(5_000_000));
    build_default_test_externalities_with_balances(vec![(user_account, user_balance)])
```

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                .with_creator_reward(Permill::zero())
+                .execute_call()
+                .unwrap();
+            BondFixture::default().execute_call().unwrap();
+            let user_reserve_pre = Balances::usable_balance(user_account);
+
+            UnbondFixture::default().execute_call().unwrap();
+
+            let user_reserve_post = Balances::usable_balance(user_account);
+            let correctly_computed_joy_amount = 95;
```
Lezek123: Please add an explainer (comment) as to where this value comes from

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                Token::account_info_by_token_and_member(token_id, user_member_id).amount;
+            assert_eq!(user_crt_pre - user_crt_post, DEFAULT_UNBONDING_AMOUNT);
+        })
+}
+
+#[test]
+fn unbonding_ok_with_event_deposited() {
+    let token_id = token!(1);
+    let ((user_id, user_account), user_balance) = (member!(2), joy!(5_000_000));
+    let price = joy!(95); // see last test
```
Lezek123: Please add explainer comment

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+            .with_member_id(creator_id)
+            .execute_call()
+            .unwrap();
+
+        last_event_eq!(RawEvent::BondingCurveDeactivated(token_id, creator_id,));
+    })
+}
+
+#[test]
+fn token_state_idle_after_amm_deactivation() {
```
Lezek123: Already checked in `deactivate_ok_with_status_set_to_idle`

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+                .with_creator_reward(Permill::zero())
+                .execute_call()
+                .unwrap();
+            let TokenData {
+                total_supply,
+                bonding_curve,
+                ..
+            } = Token::token_info_by_id(token_id);
+
+            // .001 * ((1ml + x)^2 - 1ml^2)/2
```
Lezek123: 1. I think it's better to use simplified formula `1/2 * ax^2 + asx + bx`, which is also used by the `BondingCurve` type (`s` is supply, `a`, `b` are linear function params, `x` is amount)
2. Ideally the comment would state both the formula and the actual values used, ie.:
```suggestion
            // 1/2 * a * x^2 + a * s * x + b * x = 
            // 1/2 * .001 * x^2 + .001 * 10^6 * x =
            // 0.0005 * x^2 + 1000 * x
```

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+        let token = Token::token_info_by_id(token_id);
+        assert_eq!(IssuanceState::of::<Test>(&token), IssuanceState::Idle);
+    })
+}
+
+// --------------------------------- EXTRA ----------------------------------------
+
+#[test]
+fn review_eval_function() {
```
Lezek123: Also, I think you should use some non-zero `b` for this test, otherwise it seems kind of incomplete

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,1079 @@
...
+            let bonding_values = vec![10, 100, 1000, 10000]
+                .into_iter()
+                .map(|x: Balance| {
+                    curve
+                        .eval::<Test>(x, total_supply, BondOperation::Bond)
+                        .unwrap()
+                })
+                .collect::<Vec<_>>();
+
+            // -[.001 * ((1ml - x)^2 - 1ml^2)/2]
```
Lezek123: I suggest expanding this comment in a similar fashion to how I did it for the previous one

Lezek123 (changes requested):

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I am thinking whether to simply add transaction fees instead. Since we already have patronage that has a similar mechanism to this. 
I have used this mechanics because Deso works in a similar fashion, but at this point I think that having a platform-wide buy/sell tx % fees would be overall simpler

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I have added an helper function with the full explicit formula i.e `a/2 * ((supply + amount)^2 - supply^2) + b * amount`

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I have added an helper function to make things clearer

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I have made them method + added the extrinsic version (with test) on the content pallet

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I have added an extra variable on the `BondingCurve` data structure that tracks how much supply has been created via the AMM. The condition for deactivating the amm is now:

`(amount_minted_through_amm/total_supply) <= 5%`

The threshold 5% can be fixed by the governance

runtime-modules/project-token/src/types.rs
dobertRowneySr: I don't really think that would be a problem. If you look at Deso you see that `a=.003`. The square term already makes the result quite large.
But in any case I think this problem can be solved on the next PR which takes care on all the constraints for the AMM

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I removed the creator reward in favor of tx fees, as it seemed more apt to me

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I removed the creator reward in favor of tx fees, as it seemed more apt to me

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I added an helper function to make the computation clearer

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I added an helper function to make the computation clearer

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I added an helper function to make the computation clearer

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I added an helper function to make the computation clearer

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: I added an helper function to make the computation clearer

runtime-modules/project-token/src/types.rs
dobertRowneySr: see reply for the `slope` parameter

dobertRowneySr:
### Extra addition
- I removed the creator token reward in favor of tx fees %
- I added `activate_amm` and `deactivate_amm` extrinsics to the content token pallet, they cannot be invoked when a channel transfer is taking place

bedeho:
> I removed the creator token reward in favor of tx fees %

What does this mean?

runtime-modules/project-token/src/lib.rs
```diff
@@ -1348,6 +1519,78 @@ impl<T: Config>
 
         Ok(sale.funds_collected)
     }
+
+    /// Activate Amm functionality for the token
+    /// Preconditions
+    /// - (origin, member_id) must be a valid authentication pair
```
Lezek123: No longer applies, since auth is handled by content pallet
```suggestion
```

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,31 @@ decl_error! {
...
+        SlippageToleranceExceeded,
+
+        /// Creator token balance is insufficient
+        InsufficientTokenBalance,
+
+        /// Amm treasury balance is not zero
+        OutstandingBondedAmountTooLarge,
+
+        /// Generic user not authorized error
+        UserNotAuthorized,
```
Lezek123: Unused now
```suggestion
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -163,6 +164,15 @@ decl_storage! { generate_storage_info
 
         /// Platform fee (percentage) charged on top of each sale purchase (in JOY) and burned
         pub SalePlatformFee get(fn sale_platform_fee) config(): Permill;
+
+        /// Percentage threshold for deactivating the amm functionality
+        pub AmmDeactivationThreshold get(fn amm_deactivation_threshold) config(): Permill;
```
Lezek123, line 172: I'm not sure what exact problem is this solving.

Suppose the channel owner starts the AMM and then the channel popularity increases significantly, which causes the organic growth of the CRT value. In this case it will still be almost impossible to ever close the AMM, unless someone is willing to sell the CRT way below the market value.

Even more extreme example is someone using AMM as initial strategy of distributing the tokens. In that case the percentage of supply minted through an AMM will always be 100%, so all tokens would need to be sold in order for the AMM to be closed.

I think the main reason for limiting the owner's ability to close the AMM in that people may have expectations about market's availability when taking certain positions, ie. some people will just buy CRT only to sell them later. The most obvious way I see to solve this is to force a proper heads up before the market is closed (like 2 weeks for example), by introducing a "gracing period" for AMM closure.

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,31 @@ decl_error! {
...
+        DeadlineExpired,
+
+        /// Slippage tolerance constraint tolerance not satisfied
+        SlippageToleranceExceeded,
+
+        /// Creator token balance is insufficient
+        InsufficientTokenBalance,
+
+        /// Amm treasury balance is not zero
+        OutstandingBondedAmountTooLarge,
```
Lezek123: The comment does not match the implementation anymore, but as I mentioned before, I think the implementation should actually be changed

runtime-modules/project-token/src/types.rs
```diff
@@ -640,6 +733,9 @@ pub struct TokenAllocation<Balance, VestingScheduleParams> {
 /// Input parameters for token issuance
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Default, Debug, TypeInfo)]
 pub struct TokenIssuanceParameters<Hash, TokenAllocation, TransferPolicyParams, MemberId: Ord> {
+    /// Creator Member id
+    pub creator_member_id: MemberId,
+
```
Lezek123: Should be removed, there is no longer a use-case for this.
Also the benchmark checks fail: https://github.com/Joystream/joystream/actions/runs/3513430675/jobs/5886244315

runtime-modules/project-token/src/types.rs
```diff
@@ -569,9 +572,86 @@ where
...
+                .ok_or(Error::<T>::ArithmeticError)?
+                .checked_sub(&first_term)
+                .ok_or(Error::<T>::ArithmeticError)?,
+        };
+        Ok(res.into())
+    }
+
+    pub(crate) fn increase_bonded_amount_by(&mut self, amount: Balance) {
+        self.amount_minted = self.amount_minted.saturating_add(amount);
+    }
```
Lezek123: As I mentioned in the previous review, this naming is very confusing, I didn't just mean the events:
- What is actually being "bonded" is JOY, but here you're referring to `bonded_amount` as if it's the CRT
- You then add the "bonded_amount" to `amount_minted`, as if the terms `bonded` and `minted` are interchangeable, but they actually mean two completely different things

As I mentioned before, I think buy/sell terminology is much simpler, but if you want to stick with bond/unbond, it should be specified what is actually being bonded and unbonded, because in the current implementation it's JOY, not the CRT, ie.:
- You bond JOY, you mint CRT
- You burn CRT, you unbond JOY

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +810,166 @@ decl_module! {
...
+                    account_data.increase_amount_by(amount);
+                });
+            }
+
+            TokenInfoById::<T>::mutate(token_id, |token_data| {
+                token_data.increase_supply_by(amount);
+                token_data.increase_bonded_amount_by(amount);
+            });
+
+            // TODO: redirect tx fees revenue to council
```
Lezek123, line 892: Was it already decided that the tx fees should go to the council?
Some of the alternatives include:
- Burning (just like in case of CRT sale fees)
- Fees going to channel owner

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +810,166 @@ decl_module! {
...
+            if let Some((slippage_tolerance, desired_price)) = slippage_tolerance {
+                ensure!(desired_price.saturating_sub(price) <= slippage_tolerance.mul_floor(desired_price), Error::<T>::SlippageToleranceExceeded);
+            }
+
+            // timestamp deadline check
+            if let Some(deadline) = deadline {
+                ensure!(<timestamp::Pallet<T>>::now() <= deadline, Error::<T>::DeadlineExpired);
+            }
+
+            let unbond_price = Self::unbond_tx_fees().left_from_one().mul_floor(price);
```
Lezek123: This is actually more like an `unbonded_amount` (which comes back to the naming issue)

runtime-modules/content/src/lib.rs
```diff
@@ -3748,6 +3749,72 @@ decl_module! {
...
+        /// Deactivate Amm functionality for token
+        #[weight = 100_000_000] // TODO: Adjust weight
+        pub fn deactivate_amm(
+            origin,
+            actor: ContentActor<T::CuratorGroupId, T::CuratorId, T::MemberId>,
+            channel_id: T::ChannelId,
+        ) {
+            let channel = Self::ensure_channel_exists(&channel_id)?;
+
+            channel.ensure_has_no_active_transfer::<T>()?;
```
Lezek123: I don't think this check is needed, instead you should check whether there is an active AMM during `initialize_transfer` in the first place and disallow the transfer in this case.

See: https://github.com/Joystream/joystream/blob/mainnet/runtime-modules/content/src/lib.rs#L3061, we're already ensuring there is no active revenue split and/or active/scheduled sale, adding the same condition for AMM would make sense.

runtime-modules/project-token/src/lib.rs
```diff
@@ -1348,6 +1519,78 @@ impl<T: Config>
...
+
+        TokenInfoById::<T>::mutate(token_id, |token_data| {
+            token_data.bonding_curve = None;
+        });
+
+        // burn amount exceeding existential deposit
+        let amm_treasury_account = Self::module_bonding_curve_reserve_account(token_id);
+        let amount_to_slash = Joy::<T>::usable_balance(&amm_treasury_account)
+            .saturating_sub(T::JoyExistentialDeposit::get());
+        let _ = Joy::<T>::slash(&amm_treasury_account, amount_to_slash);
```
Lezek123: It's not really and issue here, but I'd still advise using a fallible `burn_from_usable` instead, to make sure the entire provided amount is always burned

runtime-modules/project-token/src/lib.rs
```diff
@@ -1348,6 +1519,78 @@ impl<T: Config>
...
+
+        Self::deposit_event(RawEvent::AmmDeactivated(token_id, member_id));
```
Lezek123: I think burned amount would be a useful information to include in this event

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,924 @@
...
+        BondFixture::default()
+            .with_sender(bob_account)
+            .with_member_id(bob_id)
+            .with_amount(DEFAULT_BONDING_AMOUNT)
+            .execute_call()
+            .unwrap();
+
+        // ----------------- act -------------------
+        let result = UnbondFixture::default()
+            .with_amount(DEFAULT_BONDING_AMOUNT)
```
Lezek123: Alice only purchased `DEFAULT_UNBONDING_AMOUNT` (100) tokens, but you're trying to burn `DEFAULT_BONDING_AMOUNT` which is 1000, so the test doesn't really check what it's supposed to check.

Notice that `DEFAULT_UNBONDING_AMOUNT` is also kind of a confusing term, because what is getting unbonded is JOY, and the CRT would be in this case a "burned amount".

Another suggestion: Please use `with_sender(alice_account).with_member_id(alice_id)`, I know those are the default values, but it was by no means clear for me until I checked the `UnbondFixture` implementation.

runtime-modules/project-token/src/tests/bonding_curve.rs
```diff
@@ -0,0 +1,924 @@
...
+                .execute_call()
+                .unwrap();
+
+            // ----------------- act -------------------
+            let result = UnbondFixture::default()
+                .with_amount(2 * DEFAULT_BONDING_AMOUNT)
+                .execute_call();
+
+            // ---------------- assert -----------------
+            // Alice is now being vested but she has 0 transferrable amount
```
Lezek123: This comment doesn't apply here

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +810,166 @@ decl_module! {
...
+        /// - token_id must exist
+        /// - user usable JOY balance must be enough for bonding (+ existential deposit)
+        /// - deadline constraint respected if provided
+        /// - slippage tolerance constraints respected if provided
+        /// - token total supply and amount value must be s.t. `eval` function doesn't overflow
+        ///
+        /// Postconditions
+        /// - `amount` CRT minted into account (which is created if necessary with existential deposit transferred to it)
+        /// - respective JOY amount transferred from user balance to amm treasury account
+        /// - percentage of the minted CRT transferred to the token content creator (creator reward)
```
Lezek123: No longer applies

runtime-modules/project-token/src/lib.rs
```diff
@@ -800,6 +810,166 @@ decl_module! {
...
+        /// Burn desired `token_id` amount from user account and get JOY from treasury account
+        /// Preconditions
+        /// - origin, member_id pair must be a valid authentication pair
+        /// - token_id must exist
+        /// - token_id, member_id must be valid account coordinates
+        /// - user usable CRT balance must be at least `amount`
+        /// - deadline constraint respected if provided
+        /// - slippage tolerance constraints respected if provided
+        /// - token total supply and amount value must be s.t. `eval` function doesn't overflow
+        /// - amm treasury acconut must have sufficient JOYs for the operation
```
Lezek123:
```suggestion
        /// - amm treasury account must have sufficient JOYs for the operation
```

runtime-modules/project-token/src/types.rs
Lezek123: In the end it all depends on the token supply and decimals choosen by the creator.
If I want my token supply to be 1k and have only 2 decimals, then there will be just 100000 units of my CRT. In this case `a=1` is not a very high value, as buying/selling all 100000 units will only move the price (per unit) by 100000 HAPI, which is almost nothing

runtime-modules/project-token/src/types.rs
Lezek123: For `b` the problem is even more pronounced, because this is the main parameter which allows controlling the initial price. If it cannot be more than 1, then creator has almost no control over this.

I don't have an issue with this being fixed in the next PR, but I don't agree it's not a problem (this applies to both `a` and `b` actually)

Lezek123 (changes requested):

runtime-modules/project-token/src/lib.rs
dobertRowneySr: This is the intended behavior.
The CC cannot close the AMM as he pleases. If the public buys 100 CRT through the AMM then the CC cannot close the AMM unless the public has sold back to the AMM at least 95 CRT, this is the intended behavior

Now, the question is: do we want to allow the CC to close the AMM at his discretion, maybe say allowing for a gracing period as suggested (to prevent CC to immediately lock out CRT holders from selling back their token)? @bedeho

runtime-modules/project-token/src/types.rs
dobertRowneySr: see e257c40d78619458146c9281225b6d566f6450d9 for details

runtime-modules/project-token/src/lib.rs
dobertRowneySr: The destination from the proceedings from TX fees is not established yet.

runtime-modules/project-token/src/tests/bonding_curve.rs
dobertRowneySr: that was a typo on my end, good catch

runtime-modules/project-token/src/lib.rs
dobertRowneySr: I will leave the TODO as it is as I am creating an issue with all open questions for AMM to be addressed

runtime-modules/project-token/src/lib.rs
dobertRowneySr: see e257c40d78619458146c9281225b6d566f6450d9 for details

runtime-modules/content/src/lib.rs
```diff
@@ -3760,8 +3765,6 @@ decl_module! {
         ) {
             let channel = Self::ensure_channel_exists(&channel_id)?;
 
-            channel.ensure_has_no_active_transfer::<T>()?;
-
```
Lezek123: In case of `activate_amm` we still need this to prevent someone from starting transfer and then activating AMM before it's finalized (for `deactivate_amm` it's not needed, because you already ensured amm cannot be active when initializing a transfer)

runtime-modules/project-token/src/lib.rs
```diff
@@ -1351,6 +1530,81 @@ impl<T: Config>
...
+        Self::deposit_event(RawEvent::AmmActivated(token_id, member_id, curve));
+
+        Ok(())
+    }
+
+    /// Deactivate the amm functionality
+    /// Preconditions
+    /// - (origin, member_id) must be a valid authentication pair
+    /// - token_id must be a valid
+    /// - token must be in `BondigCurve` state
```
Lezek123:
```suggestion
    /// - token must be in `Amm` state
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -1351,6 +1530,81 @@ impl<T: Config>
...
+    ///
+    /// Postconditions
+    /// - Bonding Curve set to None
```
Lezek123:
```suggestion
    /// - Amm Curve set to None
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -1622,6 +1876,11 @@ impl<T: Config> Module<T> {
         <T as Config>::ModuleId::get().into_sub_account_truncating(Vec::<u8>::new())
     }
 
+    /// Returns teh account for the bonding curve treasury
```
Lezek123:
```suggestion
    /// Returns the account for the AMM treasury
```

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -0,0 +1,919 @@
...
+            let user_reserve_post = Balances::usable_balance(user_account);
+            assert_eq!(
+                user_reserve_pre - user_reserve_post,
+                correctly_computed_joy_amount
+            );
+        })
+}
+
+#[test]
+fn bond_ok_with_event_deposit() {
```
Lezek123:
```suggestion
fn amm_buy_ok_with_event_deposit() {
```

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -0,0 +1,919 @@
...
+                .execute_call();
+
+            let state_post = sp_io::storage::root(sp_storage::StateVersion::V1);
+            assert_ok!(result);
+            assert_eq!(state_pre, state_post);
+        })
+}
+
+#[test]
+fn unbond_fails_with_user_not_having_leaking_funds_from_vesting_schedule() {
```
Lezek123:
```suggestion
fn amm_sell_fails_with_user_not_having_leaking_funds_from_vesting_schedule() {
```

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -0,0 +1,919 @@
...
+            .with_amount(DEFAULT_AMM_BUY_AMOUNT)
+            .execute_call();
+
+        // ---------------- assert -----------------
+        // Alice is now being vested but she has 0 transferrable amount
+        assert_err!(result, Error::<Test>::InsufficientTokenBalance);
+    })
+}
+#[test]
+fn unbond_fails_with_user_not_having_enough_token_balance() {
```
Lezek123:
```suggestion
fn amm_sell_fails_with_user_not_having_enough_token_balance() {
```

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,28 @@ decl_error! {
...
+        DeadlineExpired,
+
+        /// Slippage tolerance constraint tolerance not satisfied
+        SlippageToleranceExceeded,
+
+        /// Creator token balance is insufficient
+        InsufficientTokenBalance,
+
+        /// Outstanding amount of CRT bought from AMM too large
+        OutstandingBondedAmountTooLarge,
```
Lezek123:
```suggestion
        /// Oustanding AMM-provided supply constitutes too large percentage of the token's total supply
        OutstandingAmmProvidedSupplyTooLarge,
```

runtime-modules/project-token/src/types.rs
```diff
@@ -569,9 +572,86 @@ where
...
+#[derive(Default, Encode, Decode, TypeInfo, Clone, Debug, Eq, PartialEq, MaxEncodedLen)]
+pub struct AmmCurve<Balance> {
+    /// Slope parameter : a
+    pub slope: Permill,
+
+    /// Intercept : b
+    pub intercept: Permill,
+
+    // amount public bought on amm so far
+    pub amount_bought_on_amm: Balance,
```
Lezek123:
```suggestion
    // amount of token added to circulation by the AMM so far
    pub provided_supply: Balance,
```

runtime-modules/project-token/src/events.rs
```diff
@@ -155,5 +156,35 @@ decl_event! {
...
         TokensBurned(TokenId, MemberId, Balance),
+
+        /// AMM activated
+        /// Params:
+        /// - token id
+        /// - member id
+        /// - params for the bonding curve
+        AmmActivated(TokenId, MemberId, AmmCurve),
+
+        /// Token Bonded
```
Lezek123:
```suggestion
        /// Tokens Bought on AMM
```

runtime-modules/project-token/src/events.rs
```diff
@@ -155,5 +156,35 @@ decl_event! {
...
+        /// Params:
+        /// - token id
+        /// - member id
+        /// - amount of CRT minted
+        /// - amount of JOY deposited into curve treasury
+        TokensBoughtOnAmm(TokenId, MemberId, Balance, JoyBalance),
+
+        /// Token Unbonded
```
Lezek123:
```suggestion
        /// Tokens Sold on AMM
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -166,6 +167,15 @@ decl_storage! { generate_storage_info
 
         /// Platform fee (percentage) charged on top of each sale purchase (in JOY) and burned
         pub SalePlatformFee get(fn sale_platform_fee) config(): Permill;
+
+        /// Percentage threshold for deactivating the amm functionality
+        pub AmmDeactivationThreshold get(fn amm_deactivation_threshold) config(): Permill;
+
+        /// Bond Transaction fee percentage
```
Lezek123:
```suggestion
        /// AMM buy transaction fee percentage
```

runtime-modules/project-token/src/lib.rs
```diff
@@ -166,6 +167,15 @@ decl_storage! { generate_storage_info
...
+        pub AmmBuyTxFees get(fn amm_buy_tx_fees) config(): Permill;
+
+        /// Bond Transaction fee percentage
```
Lezek123:
```suggestion
        /// AMM sell transaction fee percentage
```

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -0,0 +1,919 @@
...
+        let result = DeactivateAmmFixture::default()
+            .with_token_id(token_id)
+            .execute_call();
+
+        assert_err!(result, Error::<Test>::TokenDoesNotExist);
+    })
+}
+
+#[test]
+fn deactivate_fails_with_too_much_bonded_token_outstanding() {
```
Lezek123:
```suggestion
fn deactivate_fails_with_too_much_amm_provided_supply_outstanding() {
```

Lezek123 (changes requested):
1. One issue with transfers logic
2. A few more renaming suggestions
3. Content benchmarks are failing: https://github.com/Joystream/joystream/actions/runs/3532148688/jobs/5926148030

runtime-modules/project-token/src/errors.rs
```diff
@@ -159,5 +162,28 @@ decl_error! {
...
+        /// ------ AMM ---------------------------------------------------------
+
+        /// not in AMM state
+        NotInAmmState,
+
+        /// Invalid bonding curve construction parameters
+        InvalidCurveParameters,
+
+        /// Deadline timestamp lower than now
+        InvalidTimestampSpecified,
```
Lezek123: Unused (also I don't think there's any use-case for this error)

runtime-modules/project-token/src/lib.rs
```diff
@@ -803,6 +813,165 @@ decl_module! {
...
+                origin,
+                member_id
+            )?;
+
+            let token_data = Self::ensure_token_exists(token_id)?;
+            let curve = token_data.amm_curve.ok_or(Error::<T>::NotInAmmState)?;
+
+            let user_account_data_exists = AccountInfoByTokenAndMember::<T>::contains_key(token_id, &member_id);
+            let amm_treasury_account = Self::amm_treasury_account(token_id);
+            let price = curve.eval::<T>(amount, token_data.total_supply, AmmOperation::Buy)?;
```
Lezek123: The implementation works as expected, I think the only remaining question is what do we want the `supply` parameter in the pricing function to be. There are 2 approaches:
1. We can use actual token's total supply (which can come from sale, patronage etc. and is also affected by non-amm burning)
2. We can use supply provided by the AMM only

I think `2.` would be better, because in this case we don't have to worry about non-amm burns and the patronage affecting the price on the AMM.

### What happens if we use `1.`

If we use token's total supply then the price on the AMM raises when the creator claims the patronage, but the JOY reserve remains unaffected, causing the minimum sell price to raise aswell.

If someone burns tokens outside of the AMM, it's the opposite situation - the price drops, but no JOY is being paid to whoever burned the tokens, which means someone else would eventually claim this JOY from the AMM. In this case the minimum price also drops, making it either harder or impossible to ever close the AMM. Another caveat of non-amm burns is: if we were to allow negative `b`, then it would be possible for the price to drop below 0, meaning someone would get paid for buying CRT (otherwise the AMM would have some JOY locked forever). On the other hand if we don't allow negative `b`, we're limiting the slope quite substantially assuming there were already some tokens sold before the AMM started (for example, via a token sale).

### What happens if we use `2.`

If we were to go with `2.`, everything becomes much more clear. The `b` parameter of the curve is both the initial and the minimum price and it never changes. `a` can be anything `>0`, which means the curve can be anything from almost horizontal line to almost vertical line, giving issuer a lot of freedom when deciding which slope works best for them. We can enforce `b` to be `>= last_sale_price`, which means no AMM can be started that would allow the issuer to dilute the CRT by buying it at a lower price than other members had to pay before and that's it!

runtime-modules/project-token/src/tests/fixtures.rs
```diff
@@ -111,6 +111,47 @@ impl IssueTokenFixture {
...
+    pub fn with_allocation(self, beneficiary_id: MemberId, amount: Balance) -> Self {
+        Self {
+            params: self.params.with_allocation(&beneficiary_id, amount, None),
+            ..self
+        }
+    }
+
+    pub fn with_empty_allocation(self) -> Self {
+        self.with_allocation(member!(1).0, 0u64.into())
+    }
```
Lezek123: I tried adding some tests locally and I was confused why the allocation is `> 0` even when `with_empty_allocation` is used. Then I realized `with_allocation` only adds to existing allocation. How about making this method actually clear any existing allocation:
```suggestion
    pub fn with_empty_allocation(self) -> Self {
        Self {
            params: IssuanceParams {
                initial_allocation: BTreeMap::default(),
                ..self.params
            },
            ..self
        }
    }
```

runtime-modules/project-token/src/lib.rs
dobertRowneySr: This is an old debate.
When it came up I suggested 2 but Bedeho expresses his desire for 1. That's why it is implemented like this
I am in favour of 2 because in that case `b` has the meaning of "price for the first token bought on the amm".
This can (and should be enforced) allow to set `b=last_sale_price`

dobertRowneySr: I passed the `provided_supply` as an argument.
I will refactor the function as part of the next AMM PR, with all the constraints

Lezek123 (approved):
LGTM, I'm not sure what the conclusion will be in https://github.com/Joystream/joystream/issues/4151, but we can always adjust in the next PR.

dobertRowneySr:
Merging into new branch `crt_release`

Issue #4433, opened 2022-11-11
Make `@joystream/types` package hybrid (CJS+ESM)


zeeshanakram3:
Currently `@joystream/types` package is a commonJS module (CJS). However, it's better for the `Atlas` build system if it only uses ES modules (ESM) as its dependencies. 

As Klaudiusz described the context: 

> Our build system can handle CJS dependencies, but there are some issues around it. If we use @joystream/types without patching, it will work, but its dependencies will also be used in CJS versions (so all of @polkadot packages). Then, if we also use those @polkadot packages in our app, we will use ESM versions so the end result is that we have 2 copies of each library, one CJS and one ESM
Doing patching allows us to get rid of this issue and use only ESM versions

The hybrid module setup for Channels Payouts lib was done in https://github.com/Joystream/joystream/pull/4334 so lets do that for the types package too.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203351090466495) by [Unito](https://www.unito.io)

bedeho:
Done in #4435

PR #4435, opened 2022-11-12
make @joystream/types hybrid package

zeeshanakram3:
addresses #4433


kdembler (dismissed):
Seems to work nicely on our end, thanks a bunch! Such a quick change but will make our lives much easier.

We probably should bump the package version so that we can publish it? Also I see you are targetting `mainnet`, couldn't we merge it to `master`?

zeeshanakram3:
> We probably should bump the package version so that we can publish it? Also, I see you are targetting `mainnet`, couldn't we merge it to `master`?

I have bumped the version in [f4f5531](https://github.com/Joystream/joystream/pull/4435/commits/f4f553194ac3d6716dc00265585de60330ba4f75). Regarding the target branch I was not actually sure so I targeted `mainnet`, anyways I will rebase to `master` now

types/package.json
```diff
@@ -1,13 +1,29 @@
...
+      "primitives": [
+        "lib/primitives.d.ts"
+      ]
+    }
+  },
   "scripts": {
     "prepublishOnly": "yarn clean && yarn build",
     "prepack": "yarn clean && yarn build",
-    "compile": "tsc --build tsconfig.json",
+    "compile": "tsc --project tsconfig.esm.json & tsc --project tsconfig.cjs.json",
```
mnaamani: Better to fail if esm build fails, with single `&` it was running in background. I think what we want is:

```suggestion
    "compile": "tsc --project tsconfig.esm.json && tsc --project tsconfig.cjs.json",
```

types/package.json
```diff
@@ -1,13 +1,29 @@
 {
   "name": "@joystream/types",
-  "version": "0.20.3",
+  "version": "0.20.4",
   "description": "Types for Joystream Substrate Runtime - Carthage release",
-  "main": "index.js",
-  "types": "index.d.ts",
+  "main": "lib/index.js",
```
mnaamani: I `npm pack`ed the package, then tried a test to 'require' the package in node:

```
> require("./")
Uncaught:
Error: Cannot find module '/Users/mokhtar/tmp/package/lib/index.js'. Please verify that the package.json has a valid "main" entry
    at tryPackage (internal/modules/cjs/loader.js:303:19)
    at Function.Module._findPath (internal/modules/cjs/loader.js:516:18)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:867:27)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18) {
  code: 'MODULE_NOT_FOUND',
  path: '/Users/mokhtar/tmp/package/package.json',
  requestPath: './'
}
```

There is no index.js produced in `lib/` should it be:

```suggestion
  "main": "lib/cjs/index.js",
```

That worked:

```
> require('.')
{
  registry: TypeRegistry {},
  createType: [Function: createType],
  keysOf: [Function: keysOf],
  entriesByIds: [Function: entriesByIds],
  JOYSTREAM_ADDRESS_PREFIX: 126
}
```

This will be needed for vanilla javascript projects to use the package. Otherwise tested the typescript projects work fine.

mnaamani (changes requested):
Two small change requests, otherwise works as expected. Nice work.

types/package.json
mnaamani: Or am I testing it incorrectly?

mnaamani: seems to be fine if I create a simple npm package in the workspace and consume the types package.

zeeshanakram3: Yes, you are right the error only occurs in vanilla javascript projects. 

`@joystream/types` package.json has an `exports` field.

```json
 "exports": {
    ".": {
      "import": "./lib/mjs/index.js",
      "require": "./lib/cjs/index.js"
    }
```
Node documentation states that you can use both `"main"` & `"exports"` as module [entrypoints](https://nodejs.org/api/packages.html#package-entry-points). Also, from the docs,

> For packages supporting Node.js 10 and below, the ["main"](https://nodejs.org/api/packages.html#main) field is required. If both ["exports"](https://nodejs.org/api/packages.html#exports) and ["main"](https://nodejs.org/api/packages.html#main) are defined, the ["exports"](https://nodejs.org/api/packages.html#exports) field takes precedence over ["main"](https://nodejs.org/api/packages.html#main) in supported versions of Node.js.

In fact, "main" is redundant when "exports" is used. so I tried removing `"main"` field, so that vanilla node project would resolve the module using the "exports" entry, but it wasn't the case, and I got the error. No sure what's causing this issue.

```
Uncaught Error: Cannot find module './'
Require stack:
- <repl>
    at Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    at Module._load (node:internal/modules/cjs/loader:841:27)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at require (node:internal/modules/cjs/helpers:103:18) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '<repl>' ]
}
```

Anyways I have updated the main entry as requested in [8e05b04](https://github.com/Joystream/joystream/pull/4435/commits/8e05b04b84239b34db9f442ccfe9b907b2aa0d20)

mnaamani (approved):

PR #4436, opened 2022-11-12
Fix computing code shasum inside docker build

mnaamani:
Computing shasum of code inside of docker build will not produce the same result as computing on the repo because ~the files are copied and their dates change~ we weren't copying the joystream-node.Dockerfile and setting the RUNTIME_PROFILE env var.

So decided to just pass in the computed value as a build arg.


mnaamani:
I also updated docker image build to include the compressed .wasm file which we now use for the runtime upgrade.

Lezek123 (approved):
LGTM

PR #4441, opened 2022-11-16
Ephesus channel payouts proposal cli Commands

zeeshanakram3:
Adds following CLI commands:

- `content:updateChannelPayoutsProposal` For creating/updating payouts proposal parameters 
- `council:fundBudget` For funding council by any member by given amount


kdembler (approved):
LGTM

cli/src/commands/content/updateChannelPayoutsProposal.ts
```diff
@@ -0,0 +1,122 @@
...
+        description,
+        exactExecutionBlock,
+        stakingAccountId,
+      },
+      {
+        UpdateChannelPayouts: createType('PalletContentUpdateChannelPayoutsParametersRecord', {
+          commitment: payloadFilePath ? await generateCommitmentFromPayloadFile('PATH', payloadFilePath) : null,
+          payload: payloadFilePath
+            ? {
+                uploaderAccount: keypair.address,
```
mnaamani: The runtime has been updated, the uploaderAccount is no longer passed in the payload. It will be the account of the proposal. See https://github.com/Joystream/joystream/pull/4570

mnaamani (changes requested):
Command needs to be updated to match change in the runtime

cli/src/commands/content/updateChannelPayoutsProposal.ts
zeeshanakram3: Addressed in [7dfea16](https://github.com/Joystream/joystream/pull/4441/commits/7dfea165b390d0bf30dcfe5da5110a6043b931c4)

cli/src/commands/council/fundBudget.ts
```diff
@@ -0,0 +1,48 @@
...
+    const member = await this.getRequiredMemberContext()
+    const keypair = await this.getDecodedPair(member.membership.controllerAccount)
+
+    this.jsonPrettyPrint(JSON.stringify({ amount, rationale }))
+
+    await this.requireConfirmation('Do you confirm the provided input?', true)
+
+    const result = await this.sendAndFollowNamedTx(keypair, 'council', 'fundCouncilBudget', [
+      member.id,
+      amount,
```
mnaamani, line 32: This is the amount in base units (HAPIs), to make it more convenient it might be better to have input to be a decimal input in JOY and convert it to a BN representing the Balance type.

If this is the convention for other commands that take a balance amount argument, then better keep it consistent.

mnaamani (approved):
Both commands work as expected.

PR #3803, opened 2022-05-27
qn - *not* filtering support

ondratra:
This PR brings support for negative filtering in QN queries + fixed variant `or` filtering  https://github.com/Joystream/hydra/issues/485 & https://github.com/Joystream/hydra/issues/476.

This type of queries can be used now
```
# selects all nfts except the one with id 3
query {
  ownedNfts(
     where: {
          NOT: [
            {id_eq: 3},
          ]
        }
    ) {
      id,
      transactionalStatus {
        __typename
      }
    }
  }
}
```

Includes work from PRs #3815 and #3819.

PR #4468, opened 2022-11-26
cli v0.9.0

mnaamani:
We forgot to bump cli version for carthage release.

There seems to be an issue however running some commands that use the "chalk" module.

```
bin/run account:info
    Error: Must use import to load ES Module: /Users/mokhtar/tmp/package/node_modules/chalk/source/index.js
    require() of ES modules is not supported.
    require() of /Users/mokhtar/tmp/package/node_modules/chalk/source/index.js from 
    /Users/mokhtar/tmp/package/lib/base/DefaultCommandBase.js is an ES module file as it is a .js file whose nearest parent 
    package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
    Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from 
    /Users/mokhtar/tmp/package/node_modules/chalk/package.json.

    Code: ERR_REQUIRE_ESM
```

To reproduce, `npm pack`  the cli, extract then run `npm install` in extracted location, finally run with `./bin/run account:info` or `bin/run api:getUri`

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -26,7 +26,7 @@ $ npm install -g @joystream/cli
 $ joystream-cli COMMAND
 running command...
 $ joystream-cli (-v|--version|version)
-@joystream/cli/0.8.0 linux-x64 node-v14.18.0
+@joystream/cli/0.9.0 darwin-x64 node-v14.16.1
 $ joystream-cli --help [COMMAND]
 USAGE
   $ joystream-cli COMMAND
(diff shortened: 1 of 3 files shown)


Lezek123:
> There seems to be an issue however running some commands that use the "chalk" module.

`chalk` was updated in https://github.com/Joystream/joystream/pull/3803/files#diff-088b5e0ac52dd36816ebc6b121d5423c763c3c18c098b91757c014937c7d632d to a version that is not compatible with `commonjs` build target.

The changelog can be seen here: https://github.com/chalk/chalk/releases

Reverted to `4.1.2` in https://github.com/mnaamani/joystream/pull/17

mnaamani:
> > There seems to be an issue however running some commands that use the "chalk" module.
> 
> `chalk` was updated in https://github.com/Joystream/joystream/pull/3803/files#diff-088b5e0ac52dd36816ebc6b121d5423c763c3c18c098b91757c014937c7d632d to a version that is not compatible with `commonjs` build target.
> 
> The changelog can be seen here: https://github.com/chalk/chalk/releases
> 
> Reverted to `4.1.2` in [mnaamani#17](https://github.com/mnaamani/joystream/pull/17)

Thanks for the quick fix. Merged and published https://www.npmjs.com/package/@joystream/cli/v/0.9.0

Lezek123 (approved):

PR #4469, opened 2022-11-28
Unset the `WG.leader` after the lead is terminated

thesan:
Issue discovered during the QA of https://github.com/Joystream/pioneer/issues/2326


thesan:
@zeeshanakram3

- check active video counters:
- nft auction and offers
- video comments and reactions
- init storage and distribution buckets via CLI

Keep on failing. @mnaamani told me it was an unrelated issue fixed somewhere (I forgot where exactly though :facepalm:) so I rebased on the latest `mainnet`  but the fix might not be on `mainnet` yet.

tests/network-tests/src/scenarios/workingGroups.ts
```diff
@@ -8,7 +8,7 @@ import groupBudget from '../flows/working-groups/groupBudget'
 
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 scenario('Working groups', async ({ job }) => {
-  const sudoHireLead = job('sudo lead opening', leadOpening())
+  const sudoHireLead = job('sudo lead opening', leadOpening(true))
```
zeeshanakram3, line 12: This change does not help in re-hiring the leads, while full scenario is being run, i.e. the way in CI check. This code will only run if integration-tests are bring run using `./tests/network-tests/run-full-scenario.sh workingGroups`. So better to revert this change.

query-node/mappings/src/workingGroups.ts
```diff
@@ -335,27 +335,40 @@ async function handleWorkingGroupMetadataAction(
...
+  }
 
-  const terminatedEvent = new EventConstructor({
+  const terminatedWorkerEvent = new TerminatedWorkerEvent({
     ...genericEventFields(event),
     group,
     worker,
     penalty: optPenalty.unwrapOr(undefined),
     rationale: optRationale.isSome ? bytesToString(optRationale.unwrap()) : undefined,
   })
```
zeeshanakram3, line 361: Shouldn't the creation of `TerminatedLeaderEvent` & `TerminatedWorkerEvent` event entities be exclusive, i.e, only one of these should be created, at least if we follow runtime [implementation](https://github.com/thesan/joystream/blob/fac29b7cd1e9f53f36d1ad8179039cce6db24af8/runtime-modules/working-group/src/lib.rs#L760-L764)? Here, `TerminatedWorkerEvent` entity is being created every time, even if the Lead is being fired

zeeshanakram3 (changes requested):
The integration-tests are failing because those tests are dependant on working group leads being already set.

This PR terminates all leads in the `workerActions` flow, so at [line 84](https://github.com/thesan/joystream/blob/fac29b7cd1e9f53f36d1ad8179039cce6db24af8/tests/network-tests/src/scenarios/full.ts#L84-L85) in the `tests/network-tests/src/scenarios/full.ts`, all leads will be terminated, and all the subsequent tests dependent on the lead actions would fail.

Tow options for you to fix this, 

1. Move the terminate Leads action from `workersActions` into separate flow and run this at the end of all jobs
2. Rehire all the workers, right after their termination in the full scenario, they way it's been done [here](https://github.com/thesan/joystream/blob/fac29b7cd1e9f53f36d1ad8179039cce6db24af8/tests/network-tests/src/scenarios/full.ts#L78-L80)

tests/network-tests/src/scenarios/workingGroups.ts
thesan: @zeeshanakram3 thank you for the review. This change is not about the current failing tests but when running the `workingGroups` scenario (`./query-node/run-tests.sh workingGroups`) this job fails because `storageWorkingGroup` and `distributionWorkingGroup` leads are already hired at this point (probably from the "Init storage and distribution" job).

query-node/mappings/src/workingGroups.ts
```diff
@@ -335,27 +335,40 @@ async function handleWorkingGroupMetadataAction(
...
 
-  await store.save(terminatedEvent)
+  await store.save<TerminatedWorkerEvent>(terminatedWorkerEvent)
 
   const status = new WorkerStatusTerminated()
-  status.terminatedWorkerEventId = terminatedEvent.id
```
thesan, line 354: @zeeshanakram3 about [the comment on `TerminatedLeaderEvent` & `TerminatedWorkerEvent` entites](https://github.com/Joystream/joystream/pull/4469/files#r1040627135):

Here either `TerminatedLeaderEvent` or `TerminatedWorkerEvent` are passed to the `WorkerStatusTerminated` variant even though it expects `TerminatedWorkerEvent` exclusively:
```gql
type WorkerStatusTerminated @variant {
  "Related event emitted on worker termination"
  terminatedWorkerEvent: TerminatedWorkerEvent!
}
```
(plus isn't the ids of these 2 events non mutually unique ?)
I'm confuse about this but if this is an error I see is 3 solutions:
1. add a variant into the `WorkerStatus` union but this breaks the code which check for `WorkerStatusTerminated` for terminated leads (also this would be the only lead specific status).
2. make `WorkerStatusTerminated.terminatedWorkerEvent` into a union of variant this one is better IMO but it's  also a breaking change (although I'm not sure how much change is needed maybe it's not too much).
3. emitting `TerminatedLeaderEvent` along with `TerminatedWorkerEvent` but as you mentioned it differs from the runtime.

I had not realized that `3` was inconsistent with the runtime working, however I'm not sure if it's bad enough to go for `1` or `2` or maybe I'm missing something.

WDYT @zeeshanakram3 ?

traumschule:
This is about hydra, for the runtime side see #4491.

thesan:
@zeeshanakram3 after some fixing, the full workflow works for me locally so I relaunched the tests

tests/network-tests/src/scenarios/workingGroups.ts
zeeshanakram3: Sure

query-node/mappings/src/workingGroups.ts
zeeshanakram3: > I had not realized that 3 was inconsistent with the runtime working, however I'm not sure if it's bad enough to go for 1 or 2 or maybe I'm missing something.

Yeah, you are right, btw even `3` doesn't introduce any state inconsistency; max, it's just introducing the problem of data redundancy, so let's stick with that.

zeeshanakram3 (commented):
There were still some issues with the integration tests (as you see, they are failing on the current PR); the issue is that the [workerActions](https://github.com/thesan/joystream/blob/c5e4ad1c8f4377055903f5b98ee45deabbfcc1bf/tests/network-tests/src/scenarios/full.ts#L85) job is running concurrently with other jobs, so many times leads will not be terminated at the point where you are trying to [re-hire](https://github.com/thesan/joystream/blob/c5e4ad1c8f4377055903f5b98ee45deabbfcc1bf/tests/network-tests/src/scenarios/full.ts#L99) them. And will fully terminate at some later stage in tests execution. That's why, In the CI integration test, you still get the `Lead not Hired` error in some of the subsequent jobs. 

I have created a fix [PR](https://github.com/thesan/joystream/pull/6) on you fork, please merge that, hopefully it will fix the issue

thesan:
@zeeshanakram3 after this [commit](https://github.com/Joystream/joystream/pull/4469/commits/4da80146d51eff978ee5ffc78e1be7134a57d9cb) which was only about the working-group scenario the CI failed twice: [here](https://github.com/Joystream/joystream/actions/runs/3694315192/jobs/6262707234#step:8:12883) and [here](https://github.com/Joystream/joystream/actions/runs/3694315192/jobs/6262707234#step:8:12883). So I decided to go with the "Rehire all the workers" option you suggested earlier, since any of the jobs depending on `sudoHireLead` could potentially fail too. Plus another of this kind of test will probably be needed soon because it looks like the QN doesn't reflect leads leaving their position well either (see: Joystream/pioneer#3850).

zeeshanakram3 (approved):
LGTM

mnaamani:
I can't seem to get integration tests to run successfully on my localhost, will try on linux (I always seem to have issues on mac)

mnaamani (approved):

PR #4470, opened 2022-11-28
Fix Hydra services potentially starting in a wrong order

Lezek123:
Since https://github.com/Joystream/joystream/pull/4413 and until some better fix is implemented in Hydra we need to make sure that `graphql-server` service always starts **after** `processor`.

This is because the `graphql-server` will currently try to retrieve the `processor` ip address from the dns on startup and then limit the access to `/update-processor-state` only to this ip.


mnaamani (approved):
Is it guaranteed this will always work. It is certainly still better than not having this dependency.

PR #4475, opened 2022-11-30
Devops

mnaamani:
Final tweaks for network deployment


devops/ansible/roles/rpc/templates/joystream-node.service.j2
```diff
@@ -10,17 +10,20 @@ ExecStart={{ template_working_dir }}/bin/joystream-node \
...
         --pruning archive \
-        --ws-max-connections 1000 \
+        --ws-max-connections 512 \
         --telemetry-url "wss://telemetry.polkadot.io/submit/ 0" \
         --no-mdns \
-        --in-peers 128 \
-        --out-peers 128
+        --in-peers 64 \
+        --out-peers 64 \
+        --max-runtime-instances 64
```
Lezek123, line 21: Since we have 1 GB heap allocation set for WASM, isn't setting this to such a high value putting the node at risk of memory overflow?

Lezek123 (approved):
Didn't encounter any issues, works as expected

devops/ansible/roles/rpc/templates/joystream-node.service.j2
mnaamani: Yes perhaps this value is a bit large. I suppose it should be manually updated on each rpc node depending on its physical RAM. The default is 8 instances. Maybe 16 or 20 would be better values for a machine with 32GB memory for example?

Issue #4401, opened 2022-10-24
Integration Tests without Sudo calls


mnaamani:
After main net is deployed, the sudo pallet will be removed from the runtime.
Many Integration/Network tests depend on sudo pallet for putting the runtime in certain states or bypassing the council. These tests will all break see: https://github.com/Joystream/joystream/pull/4400

The tests need to be updated to work around this limitation.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203661753438153) by [Unito](https://www.unito.io)

bedeho:
We have to decide whether we actually want to keep such Sudo calls as part of testing, @mnaamani suggest this is not a good long term solution.

bedeho:
We concluded
- sudo pallet will be removed
- tests are fixed by making tests sufficiently complex so as to bootstrap council and other steps in order to get to the right milestone for real tests to begin.

PR #4477, opened 2022-11-30
Mainnet: liberated runtime (disable sudo call)

mnaamani:
Disable sudo pallet `Call`, in production and staging runtimes.
Sudo still available in testing and playground runtimes to facilitate testing and running integration tests. https://github.com/Joystream/joystream/issues/4401

The way the sudo dispatch calls are disabled is simply by not passing the `Call` type to the sudo pallet in the construct_runtime! macro like so:

```rust
Sudo: pallet_sudo::{Pallet, Storage, Event<T>, Config<T>},
```

And for test runtimes it is passed:
```rust
Sudo: pallet_sudo::{Call, Pallet, Storage, Event<T>, Config<T>},
```

If we don't wish to use this new approach we can just go for dropping the pallet altogether https://github.com/Joystream/joystream/pull/4478


mnaamani:
It was decided that we will stick to with https://github.com/Joystream/joystream/pull/4478 approach of dropping the sudo pallet, as we have already tested it and with mainnet only a few days away, we will not risk using not fully tested approach in this PR.

Issue #4128, opened 2022-08-05
Fix non ASCII encoding on some fields


thesan:
## Context
Right now some fields containing non ASCII characters will be decoded as a HEX instead of UTF8 after being submitted in an extrinsic parameters.
For example the title of [this proposal](https://dao-git-carthage-joystream.vercel.app/#/proposals/preview/3?network-config=https://54.172.154.45.nip.io/config.json)
See:
- https://github.com/Joystream/pioneer/issues/3418
- https://github.com/Joystream/pioneer/issues/3918

## Scope
There is a small chance this could be fixed in `@joystream/types` but it looks more like a `polkadot.js` issue.
In this case it could be patch in the query nodes mappings. Maybe in the string sanitizing function.

PR #4493, opened 2022-12-13
Fix proposals non ASCII text encoding

thesan:
This PR addresses:
- #4128

~~AFAICS from the code this problem should have only affected proposals.~~

In short the problem was that `createType('Bytes', string).toHuman()` returns the original string when string is pure ASCII but weirdly it returns the hex of the string when it contains non ASCII chars.

`bytesToString` encodes the bytes into a UTF8 string and escapes it.

### DIFF
--- a/query-node/mappings/src/content/channel.ts
+++ b/query-node/mappings/src/content/channel.ts
@@ -28,6 +28,7 @@ import {
   logger,
   saveMetaprotocolTransactionSuccessful,
   saveMetaprotocolTransactionErrored,
+  bytesToString,
 } from '../common'
 import {
   processBanOrUnbanMemberFromChannelMessage,
@@ -153,7 +154,7 @@ export async function content_ChannelAssetsDeletedByModerator({
     actor: await convertContentActor(store, actor),
     channelId: channelId.toNumber(),
     assetIds: Array.from(dataObjectIds).map((item) => Number(item)),
-    rationale: rationale.toHuman() as string,
+    rationale: bytesToString(rationale),
   })
 
   await store.save<ChannelAssetsDeletedByModeratorEvent>(channelAssetsDeletedByModeratorEvent)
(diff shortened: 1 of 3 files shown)


zeeshanakram3 (commented):
LGTM.

Thanks for the fix. There are a couple of other places in the mappings, mainly in the event entities where this issue exists, would you please fix in all those places too. e.g.

- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/channel.ts#L156
- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/channel.ts#L191
- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/channel.ts#L232
- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/video.ts#L286
- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/video.ts#L306
- https://github.com/thesan/joystream/blob/8bbe7ea0cbf0f6a9b04cc4675ba0b3815f054841/query-node/mappings/src/content/video.ts#L350

zeeshanakram3 (approved):
LGTM

Lezek123 (approved):
LGTM.

I noticed there is another occurence of `Bytes.toString()` in https://github.com/Joystream/joystream/blob/78b4b83e970cdcaeb600a3f6ddef8b1fe8ff5c51/query-node/mappings/src/membership.ts#L203, but this method appears to be unused anyway. Could you please remove it?

Issue #4500, opened 2022-12-21
cli installation fails


traumschule:
Command takes a long time and finally fails.

`npm install -g @joystream/cl`
```
18703 verbose stack Error: iso-constants@0.1.2 install: `node build.js > index.browser.js`
18703 verbose stack Exit status 2
18703 verbose stack     at EventEmitter.<anonymous> (/root/.volta/tools/image/node/14.21.1/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:332:16)
18703 verbose stack     at EventEmitter.emit (events.js:400:28)
18703 verbose stack     at ChildProcess.<anonymous> (/root/.volta/tools/image/node/14.21.1/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
18703 verbose stack     at ChildProcess.emit (events.js:400:28)
18703 verbose stack     at maybeClose (internal/child_process.js:1088:16)
18703 verbose stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:296:5)
18704 verbose pkgid iso-constants@0.1.2
18705 verbose cwd /root
18706 verbose Linux 5.10.0-18-amd64
18707 verbose argv "/root/.volta/tools/image/node/14.21.1/bin/node" "/root/.volta/tools/image/node/14.21.1/bin/npm" "install" "-g" "@joystream/cli"
18708 verbose node v14.21.1
18709 verbose npm  v6.14.17
18710 error code ELIFECYCLE
18711 error errno 2
18712 error iso-constants@0.1.2 install: `node build.js > index.browser.js`
18712 error Exit status 2
18713 error Failed at the iso-constants@0.1.2 install script.
18713 error This is probably not a problem with npm. There is likely additional logging output above.
18714 verbose exit [ 2, true ]
```

env debian 11.5
```
root@joystreamstats:~# npm --version
6.14.17
root@joystreamstats:~# node --version
v14.21.1
```

## Solution
Can this be simplified?

Not sure if there could be faster recommended way to install.

For example a binary release like for `joystream-node`. Or is there?

traumschule:
Thanks @Lezek123!
> Ok, I was able to reproduce the exact same issue
> I did some digging and it looks like the answer can be found here: https://github.com/ipfs/js-ipfs/issues/3241#issuecomment-678953760
[...]

Issue #4463, opened 2022-11-22
setup.sh


traumschule:
On debian 10.13 `./setup,sh` fails with
```
The following packages have unmet dependencies:
 containerd.io : Conflicts: runc
E: Unable to correct problems, you have held broken packages.
```
Removing runc does not seem advisable.
```
The following packages will be REMOVED:
  docker.io runc
0 upgraded, 0 newly installed, 2 to remove and 0 not upgraded.
After this operation, 249 MB disk space will be freed.
Do you want to continue? [Y/n]
```

Would be great to make the script more robust for non-root users on non-sudo installations. There can be two scripts for normal users and one to be executed with elevated privileges to install dependencies.

PR #4501, opened 2022-12-31
CLI Readme: recommend not to install with root privileges

traumschule:
Concludes #4500

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -22,7 +22,7 @@ Command Line Interface for Joystream community and governance activities
 <!-- usage -->
 
 ```sh-session
-$ npm install -g @joystream/cli
+$ npm install @joystream/cli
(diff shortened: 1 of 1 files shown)


mnaamani:
I don't think this solves the problem described. This will install the cli into the local working directory/workspace.
What I would recommend is to still use the `-g` option but with either: 
1.`volta` installed
2. or by setting a path in the user's home directory to be the location for global installs by editing the [npmrc](https://docs.npmjs.com/cli/v9/configuring-npm/npmrc) (so user will not require sudo/root privileges).

traumschule:
I was looking for an **easier** solution also `WORKSFORME` :)
From app developer perspective it's not very polite to claim root permission where not actually needed. #4463 makes the same assumption to be "the boss" of the system but maybe that's intended?

Issue #4307, opened 2022-09-23
Apps as first-class citizens


bedeho:
# Background

When an application brings a creator to the platform, or a creator publishes a video, often this is positive spillover for the system as a whole. It makes the system more attractive for other applications, consumers and creators, and it also allows the DAO to capture value from downstream activities associated with this publishing, like NFT fees, gateway burns, etc. This means that it is efficient for the DAO to directly subsidies such onboarding in a way which is informed by how much value a creator brings to the platform down stream, for example as measured by various burns or shares of creator payouts. Such subsidisation however requires attribution. It must be known what application actually was the vehicle for capturing and onboarding the creator and content. Such attribution has another strong benefit. When content is published with authentic application attribution, other applications receive a signal about the properties of the content being published, by virtue of things like their creator/content pool post processing steps (screening, transcoding, ...). Lastly, attribution involves creating an application index which will serve as an excellent public index of what exactly is being built on and served on the platform at any given time.

# Proposal

Introduce applications at the metaprotocol level, and wrap existing metaprotocol messaging inside authenticated application action messages. The reason we put applications at the metaprotocol is that we don't need automated value transfer, e.g. x% of creator payout going to onboarding application. This is not only overkill right now, but we are unlikely to pick the exact right approach out of the box, so sticking to a more light weight and flexible approach is better.

## Member & Content lead remarks

We make applications owned by either members or the DAO itself, as presented by the content lead. We add the following new messages with obvious semantics to be used over content lead or member remark extrinsics:

- `CreateApp`: creates new app with incremental ID.
  - name: immutable
  - website_url: where to read more about the project, company for this app.
  - use_uri: where to actually get or use this app.
  - small_icon: url+hash.
  - medium_icon: url+hash.
  - big_icon: url+hash.
  - one_liner: tagline.
  - description: longer description.
  - terms_of_service: terms of service that will be clickwrapped, some URL+hash is enough, raw text will be too long.
  - auth_key: public key used for authenticating messages, may as well be same signature scheme as blockchain accounts.
  - platform: mobile, tablet, roku, web, native (win), native (osx), smart tv,...
  - ~versions: native win, native osx, iphone, android,...~
  - category: messaging, streaming, adult, ...
- `UpdateApp`: updates app properties
  - app ID.
  - optional new values for everything except `name`
- `DeleteApp`: drops an app.
  - app ID.

## Content Directory Creation Extrinsics

_As preface note that we still allow people to create channels and publish just like before._

[...]

Issue #4158, opened 2022-08-10
Apps as channels


bedeho:
# Background

We are in the process of incorporating applications as metaprotocol level concept, here https://github.com/Joystream/joystream/issues/4307, primarily to unlock attribution that finally enables the council to pay apps for the content and creators they onboard based on accurate information. That proposal made applications owned by _memberships_, which was a natural starting point, it does however have limitations:

1. **Scalable Payments:** a practical and efficient way to reward apps
2. **Financing:** Running an app requires investment of various kinds of front, with uncertain payoffs down the road.
3. **Transferable ownership:** Being able to sell your app to someone else later, so they are the full beneficiary and administrator, generates better incentives for long term stewardship, even if time horizon for personal involvement is uncertain or finite. This is of particular importance, as apps can harm end users.

# Proposal

These could all be implemented by making the metaprotocol more elaborate, adding new proposals and adding new integrations between membership pallet and `project_token` pallet, but _even then_, there would be no functional software that would be able to allow people to use all these features.

Instead, I propose we simply overload the `channel` entity in the `content` pallet to also be to function as an app. By doing that, you get 1,2,3 solved for free, without any work. One could also very easily add a new section to Atlas, or just make a standalone product, which allowed for using all the $CRT features, but only for apps, not creator channels, and very limited extra work would be needed. 

Technically, one could more specifically do something like this:

- The actions `CreateApp`, `UpdateApp` are sent over `content::channel_owner_remark` instead of `membership::member_remark`. I think `DeleteApp` causes more trouble than its worth. 
- Apps have their own ID identifier space, as distinct from the channel IDs
- There can be at most one app per channel.
- In QN/Orionv2
  - Add `Apps` as a new schema entity, it holds all the metadata already defined for an app.
  - Add nullable 1-1 relationship `Channel::app` to `Apps`

This is not the only way to do this, but it is a way which causes minimal change requirements other places. **Feel free to improve!**

Atlas can now choose to either systematically ignore all channels, or videos from channels, which have an app, or they can just be included as normal channels. Notice that its still possible to do all the normal channel stuff with apps. If we wanted, Atlas could also choose to add a flare or other sort of signal to channels which are apps. Most regular users would just be non-the wiser.

_Question: with current QN or coming Orinv2, will it be easy to avoid channels which are apps, and also videos that correspond to channels that are apps, in all places needed? (search, within categories, etc.)_

bedeho:
[...]

PR #4504, opened 2023-01-02
Add support for registering applications via a metaprotocol calls

### LINKED ISSUE #3517: Add support for registering applications via a metaprotocol calls
Part of https://github.com/Joystream/joystream/issues/4307

Handle `Member & Content lead remarks` section from above issue. That means:

1. Add `CreateApp`, `UpdateApp` and `DeleteApp` messages to protobuf definitions
2. Handle member remarks with App messages in the Query Node. QN should maintain a list of all the apps
	a. Make sure that updates and deletion can be done only by the member that created the app in the first place

drillprop:
Fix https://github.com/Joystream/atlas/issues/3517 https://github.com/Joystream/joystream/issues/4158

Context: #4307


query-node/schemas/app.graphql
```diff
@@ -0,0 +1,23 @@
+type App @entity {
+  id: ID!
```
zeeshanakram3: As the feature issue describes, the app itself will be owned by either member or curator groups, so having ownership field is important to describe the relationship of `App` entity with the owner entity. 

For this add two new nullable fields e.g. `ownerMember` & `ownerCuratorGroup` and then set those in the mappings based on the creator.

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+    category: appMetadata?.category || undefined,
+    authKey: appMetadata?.authKey || undefined,
+  })
+  await store.save<CreateApp>(newApp)
+  logger.info('App has been created', { name })
+}
+
+export async function processUpdateAppMessage(
+  store: DatabaseManager,
+  event: SubstrateEvent,
```
zeeshanakram3: unused argument

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+  app.category = appMetadata?.category || app.category
+  app.authKey = appMetadata?.authKey || app.authKey
+
+  await store.save<App>(app)
+  logger.info('App has been updated', { appId })
+}
+
+export async function processDeleteAppMessage(
+  store: DatabaseManager,
+  event: SubstrateEvent,
```
zeeshanakram3: unused argument

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+  let tries = 0
+
+  // make sure app id is unique
+  while (await store.get<App>(App, { where: { id: appId } })) {
+    tries++
+    appId = `${event.blockNumber}-${event.indexInBlock}-${tries}`
+  }
+
+  return appId
+}
```
zeeshanakram3: This way of generating app IDs is only needed if entity ID input is required from user, e.g. in the creation of video category. Here, `let appId = ${event.blockNumber}-${event.indexInBlock}` would be sufficient to generate unique IDs every time since

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+    mediumIcon: appMetadata?.mediumIcon || undefined,
+    bigIcon: appMetadata?.bigIcon || undefined,
+    oneLiner: appMetadata?.oneLiner || undefined,
+    description: appMetadata?.description || undefined,
+    termsOfService: appMetadata?.termsOfService || undefined,
+    platforms: appMetadata?.platforms || undefined,
+    category: appMetadata?.category || undefined,
+    authKey: appMetadata?.authKey || undefined,
+  })
+  await store.save<CreateApp>(newApp)
```
zeeshanakram3: Incorrect type here
```suggestion
  await store.save<App>(newApp)
```

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+): Promise<App | undefined> {
+  const app = await store.get(App, {
+    where: {
+      id: appId,
+      createdById: memberId.toString(),
+    },
+  })
+
+  return app
+}
```
zeeshanakram3: 1. As this `createdById` property is scalar so while querying it won't give corresponding relational entity
2. As I said earlier app can be owner by both member & curator group so its better to have to distinct relations

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+  const isAppExists = await store.get(App, {
+    where: {
+      name: name,
+    },
+  })
+
+  if (isAppExists) {
+    logger.error('App already exists', { name })
+    return
+  }
```
zeeshanakram3: For the parent issue App name should be: `name: immutable`. So it does not says anything about being unique so I think we should allow having apps with duplicate names, anyways please ask Beheho/Dmitry for clarity on this.

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,123 @@
...
+  const { appId } = message
+
+  const app = await getAppByIdAndMemberId(store, appId, memberId)
+
+  if (!app) {
+    logger.error("App doesn't exists or doesn't belong to the member", { appId, memberId: memberId.toString() })
+    return
+  }
+
+  await store.remove<App>(new App({ id: appId }))
```
zeeshanakram3: You can just pass the `app` object returned in line 86, no need to create new one.
```suggestion
  await store.remove<App>(app)
```

zeeshanakram3 (changes requested):
First off great effort.

1. Since the `App` can be owner by curatorGroup along with members so there should be a way for also creating curator group owner Apps (created by content WG lead), For that purpose use `lead_remark` instead of simple `member_remark` extrinsic. Also note that you will have to extend protobuf message schema with additional, say `LeadRemarked` message and wrap `CreateApp`, `UpdateApp` & `DeleteApp` actions inside it. And then add mapping to process the message

```proto
message LeadRemarked {
    oneof lead_remarked {
        CreateApp create_app = 1;
        UpdateApp update_app = 2;
        DeleteApp delete_app = 3;
    }
}
```

2. For integration tests as you are creating new scenario so you should add new flows into scenario/full.ts file so they can be run during GitHub CI checks

query-node/mappings/src/content/app.ts
drillprop: Talked to Bedeho and we decided that the app name should be unique.

query-node/schemas/content.graphql
```diff
@@ -8,6 +8,29 @@ type Language @entity {
   createdInBlock: Int!
 }
 
+type App @entity {
+  id: ID!
+  name: String!
+  ownerMember: Membership
+  ownerCuratorGroup: CuratorGroup
+  # Url where user can read more about the project or company for this app
+  websiteUrl: String
```
zeeshanakram3: Please use the same formatting scheme as used to define other entity type, to have coherent formatting. Also, use double-quotes(`""`) to describe a field instead of using (`#`) comments. `""` comments are used to auto-generate the documentation associated with a field/entity which will be visible to you when using playground etc.

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,108 @@
...
+    ownerMember: channel.ownerMember,
+    channel,
+  })
+  await store.save<App>(newApp)
+  logger.info('App has been created', { name })
+
+  channel.app = newApp
+
+  await store.save<Channel>(channel)
+  logger.info('Channel has been updated', { channel })
```
zeeshanakram3: Redundant. Doesn't really need this, other part of the relation will be automatically created.

```suggestion
```

query-node/mappings/src/content/app.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+async function createAppId(event: SubstrateEvent): Promise<string> {
+  return `${event.blockNumber}-${event.indexInBlock}`
+}
+
+export async function proccessUpdateApp(
```
zeeshanakram3: Incorrect spelling.

```suggestion
export async function processUpdateApp(
```

zeeshanakram3 (changes requested):
Looks good overall. Please address few remaining minor changes

zeeshanakram3 (approved):
LGTM

Lezek123 (approved):

Issue #4505, opened 2023-01-03
Add claimed amount field to claim event


bedeho:
# Background

https://github.com/Joystream/atlas/issues/3505

# Proposal

Add `amount` from `ensure_can_claim_channel_reward` to event, also update query node.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203640998332462) by [Unito](https://www.unito.io)

PR #4506, opened 2023-01-03
Ephesus: add claimed reward amount field

zeeshanakram3:
addresses #4505

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203711791393525) by [Unito](https://www.unito.io)


dobertRowneySr (approved):
LGTM

mnaamani:
The docker build fix is available in https://github.com/Joystream/joystream/pull/4528 perhaps we can cherry-pick it into this work

runtime-modules/content/src/lib.rs
```diff
@@ -4907,7 +4907,7 @@ decl_event!(
 
         // Rewards
         ChannelPayoutsUpdated(UpdateChannelPayoutsParameters, Option<DataObjectId>),
-        ChannelRewardUpdated(Balance, ChannelId),
+        ChannelRewardUpdated(Balance, Balance, ChannelId),
```
mnaamani, line 4910: Can we guarantee that the query node will not have processed a `ChannelRewardUpdated` event before ephesus release? Otherwise it is not safe to change the type this way. It would be safer to add the cashout_amount at the end of the tuple.

mnaamani: Answering myself, reason I ask because the extrinsic `claim_channel_reward` is not filtered, but of course `set_channel_paused_features_as_moderator` is filtered, so it is not possible to "unpause" `CreatorCashout` feature prior to ephesus.

mnaamani (approved):

Issue #4507, opened 2023-01-03
Add payments to `member_remarks` 


bedeho:
# Background

There are metaprotocol use cases where having payments also makes sense, most imminent example is channel tips from patrons or gateways.

# Proposal

Add optional payments to `member_remark` which is funded out of caller controller account and respects all locks and vesting, hence invitation funds cannot be used. I _think_ same logic as for paying for NFTs as a member, but double check!

Also add QN support!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203642373930500) by [Unito](https://www.unito.io)

PR #4517, opened 2023-01-10
Ephesus: add optional payment to member remark

zeeshanakram3:
Addresses #4507. This PR addresses all the relevant runtime, QN and integration tests work.

depends on https://github.com/Joystream/hydra/pull/513

---
- Adds Optional `payment: Option<(T::AccountId, T::Balance)>` parameter to `member_remark` extrinsic.
- Adds new protobuf `MakeChannelPayment` message, with an option of adding any of the three payment contexts as described in https://github.com/Joystream/atlas/issues/3591

```proto
message MakeChannelPayment {
    // Reason why payment is being made
    optional string rationale = 1;

    // Context of payment, if it is being made for a specific video, channel, or from an app/gateway
    oneof payment_context {
        // payment as a tip associated with a specific video
        uint64 video_id = 2;
        // payment as a channel wide tip
        uint64 channel_id = 3;
        // payment from member representing a gateway
        uint32 app_id = 4;
    }
}
```
- Adds the following QN entities:
```gql
type PaymentContextVideo @variant {
  "Video for which the payment was made"
  video: Video!
}

type PaymentContextChannel @variant {
  "Channel for which the payment was made"
  channel: Channel!
}

type PaymentContextApp @variant {
  "Gateway/App that made the payment to specific channel"
  app: App!
}

"Various Channel Payment Contexts"
union PaymentContext = PaymentContextVideo | PaymentContextChannel | PaymentContextApp

"Direct channel payment by any member by-passing the council payouts."
type ChannelPaymentMadeEvent @entity {
  ### GENERIC DATA ###
  # ...
  
  ### SPECIFIC DATA ###

  "Actor that made the payment"
  payer: Membership!

  "Amount of the payment"
  amount: BigInt!

  "Payment and payee context"
  paymentContext: PaymentContext

  "Channel that received the payment (if any)"
  payeeChannel: Channel

  "Reason of the payment"
  rationale: String
}
```
- Although payment can be sent with any arbitrary message type, however, right now, mappings will only process the payment if the metadata message type is  `MakeChannelPayment`. Maybe, we can also support sending payment with other existing metaprotocol messages e.g., `ReactComment`, with the intention of tipping the comment creator. This is one example, etc.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203686525354113) by [Unito](https://www.unito.io)


runtime-modules/membership/src/tests/mod.rs
```diff
@@ -1378,6 +1383,47 @@ fn successful_member_remark() {
...
+        );
+
+        assert_eq!(validation_result, Ok(()),);
+    });
+}
+
+#[test]
+fn successful_member_remark_with_payment() {
+    build_test_externalities_with_lead_set().execute_with(|| {
+        const ALICE_FREE_BALANCE: u64 = 100;
```
ignazio-bovo, line 1416: - `ALICE.free_balance = 100` => `ALICE.usable_balance = 100` since no locks are enforced here
- the payment check in the *-remark extrinsic requires that `ALICE.usable_balance >= amount_to_pay  + ExistentialDeposit::get()`
- in the invocation you are setting `amount_to_pay = ALICE.usable_balance` and `ExistentialDeposit = 10` (in the mock configuration)

Is this test passing?

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentFixture.ts
```diff
@@ -0,0 +1,83 @@
...
+import { ChannelPaymentMadeEventFieldsFragment } from '../../../graphql/generated/queries'
+import { EventDetails, EventType } from '../../../types'
+import { Utils } from '../../../utils'
+
+type MemberRemarkedEventDetails = EventDetails<EventType<'members', 'MemberRemarked'>>
+
+export type ChannelPaymentParams = {
+  asMember: MemberId
+  msg: IMakeChannelPayment
+  payment: [string, BN]
```
dobertRowneySr:
```suggestion
  payment: [string, BN] // [address, amount]
```

dobertRowneySr (changes requested):
Great job, I would add the following recommendations:
- The extrinsic  `member_remark` should be benchmarked again 
- I think you should add an edge case, where the extrinsic succeeds but the channel reward account is not found by the QN. Since you have added just the happy path case

bedeho:
Since this is going into `ephesus`, where #4307 is not in scope, let us drop `uint32 app_id = 4;`, make sure to update #4518 also.

dobertRowneySr (changes requested):
The requests are addressed :+1: however there are some `yarn lint` errors, probably you need to re-build the metadata-protobuf package `yarn workspace @joystream/metadata-protobuf build` should update them (this is needed because `appId` was removed)

query-node/mappings/src/content/channel.ts
```diff
@@ -544,3 +556,66 @@ export async function content_ChannelFundsWithdrawn({ store, event }: EventConte
...
+  message: IMakeChannelPayment,
+  [payeeAccount, amount]: [AccountId32, Balance]
+): Promise<ChannelPaymentMadeEvent> {
+  const member = await getMemberById(store, memberId)
+
+  // Only channel reward accounts are being checked right now as payment destination.
+  // Transfers to any other destination will be ignored by the query node.
+  const channel = await store.get(Channel, { where: { rewardAccount: payeeAccount.toString() } })
+  if (!channel) {
+    inconsistentState('Payment made to unknown channel reward account')
```
Lezek123: This is not really inconsistent state, I think it qualifies for `invalidMetadata`. `inconsistentState` should only occur when some critical assumptions about the query node state are broken and it should cause the QN to shutdown (which is not the case here, since the error is being caught later)

query-node/mappings/src/content/channel.ts
```diff
@@ -544,3 +556,66 @@ export async function content_ChannelFundsWithdrawn({ store, event }: EventConte
...
+  }
+
+  // Get payment context from the metadata
+  const getPaymentContext = async (msg: IMakeChannelPayment) => {
+    if (msg.channelId) {
+      const paymentContext = new PaymentContextChannel()
+      if (msg.channelId.toString() !== channel.id) {
+        invalidMetadata(
```
Lezek123: `invalidMetadata` doesn't throw an error, so the metaprotocol transaction is successful regardless of the channel id provided.

Since `msg.channelId` doesn't really matter, as we always use the channel queried by reward account anyway, perhaps the user shouldn't even need to provide `channelId`?

query-node/mappings/src/content/channel.ts
```diff
@@ -544,3 +556,66 @@ export async function content_ChannelFundsWithdrawn({ store, event }: EventConte
...
+        return
+      }
+
+      if (video.channel.id !== channel.id) {
+        invalidMetadata(
+          `payment context (video) added in the metadata of 'member_remark' call is different from` +
+            `queried video based on payee address, EXPECTED: ${channel.id}, ACTUAL: ${video.channel.id}`
+        )
+      }
+      paymentContext.videoId = channel.id
```
Lezek123: You're using `channel.id` as `videoId` here

query-node/mappings/src/membership.ts
```diff
@@ -666,6 +661,19 @@ async function processMemberRemark(
     return { commentDeletedId: comment.id }
   }
 
+  // Though the payments can be sent along with any arbitrary metadata message type,
+  // however they will only be processed if the message type is 'makePayment'
+  if (messageType === 'makeChannelPayment' && payment) {
```
Lezek123: In case the `payment` is empty, the function will fallback to `Unsupported message type` error, which in this case is incorrect. I think there should be a separate error case for empty `payment`.

query-node/schemas/contentEvents.graphql
```diff
@@ -670,3 +670,53 @@ type ChannelPayoutsUpdatedEvent @entity {
...
+  payer: Membership!
+
+  "Amount of the payment"
+  amount: BigInt!
+
+  "Payment and payee context"
+  paymentContext: PaymentContext
+
+  "Channel that received the payment (if any)"
+  payeeChannel: Channel
```
Lezek123, line 718: There is a redundancy here, since `paymentContext.channel` and `payeeChannel` are essentially the same relation.

However due to issues with deep filtering in Hydra perhaps it's ok to leave it as is for now, for Orion v2 mappings it'll probably be a good idea to remove `payeeChannel` completely though.

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentHappyCaseFixture.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export class DirectChannelPaymentsHappyCaseFixture extends StandardizedFixture {
+  protected paymentParams: ChannelPaymentParams[]
+
+  public constructor(api: Api, query: QueryNodeApi, paymentParams: ChannelPaymentParams[]) {
+    super(api, query)
+    this.paymentParams = paymentParams
+  }
+
+  public async getCreatedCommentsIds(): Promise<string[]> {
```
Lezek123: This code is unrelated to this fixture

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentHappyCaseFixture.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise'>[]> {
+    return this.paymentParams.map((params) => {
+      const msg: IMemberRemarked = {
+        makeChannelPayment: {
+          rationale: params.msg.rationale,
+          videoId: params.msg.videoId,
+          channelId: params.msg.channelId,
+        },
```
Lezek123:
```suggestion
        makeChannelPayment: params.msg,
```

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentHappyCaseFixture.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+      return this.api.tx.members.memberRemark(
+        params.asMember,
+        Utils.metadataToBytes(MemberRemarked, msg),
+        params.payment
+      )
+    })
+  }
+
+  protected assertQueryNodeEventIsValid(qEvent: ChannelPaymentMadeEventFieldsFragment, i: number): void {
```
Lezek123, line 55: Missing checks for `payeeChannel` and `paymentContext`

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentWithInvalidRewardAccountFixture.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class DirectChannelPaymentsWithInvalidRewardAccountFixture extends StandardizedFixture {
+  protected paymentParams: ChannelPaymentParams[]
+
+  public constructor(api: Api, query: QueryNodeApi, paymentParams: ChannelPaymentParams[]) {
+    super(api, query)
+    this.paymentParams = paymentParams
+  }
+
+  public async getCreatedCommentsIds(): Promise<string[]> {
```
Lezek123: Unrelated code

tests/network-tests/src/fixtures/content/channelPayouts/DirectChannelPaymentWithInvalidRewardAccountFixture.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise'>[]> {
+    return this.paymentParams.map((params) => {
+      const msg: IMemberRemarked = {
+        makeChannelPayment: {
+          rationale: params.msg.rationale,
+          videoId: params.msg.videoId,
+          channelId: params.msg.channelId,
+        },
```
Lezek123:
```suggestion
        makeChannelPayment: params.msg,
```

Lezek123 (changes requested):

query-node/mappings/src/content/channel.ts
zeeshanakram3: I agree that using `inconsistentState` in any of the Meta-protocol actions seems an incoherent approach since semantically it signifies inconsistent state in QN (possibly against runtime) which basically is not possible in metaprotocol transactions. However I used it here to end control-flow(since only payments to channels are being processed), as you said error is being caught so based on action success/failure we are saving meta-transaction state in the QN event entity, which by using `invalidMetadata` would not be possible since it just logs the info. 

I found another helper, `unexpectedData` which exactly behaves as `inconsistentState` but does not signifies anything about broken QN state. Should I use that instead of `inconsistentState`?

query-node/mappings/src/content/channel.ts
zeeshanakram3: > Since msg.channelId doesn't really matter, as we always use the channel queried by reward account anyway, perhaps the user shouldn't even need to provide channelId?

Yeah, totally makes sense, will address it.

query-node/mappings/src/content/channel.ts
```diff
@@ -600,9 +590,13 @@ export async function processChannelPaymentFromMember(
             `queried video based on payee address, EXPECTED: ${channel.id}, ACTUAL: ${video.channel.id}`
```
Lezek123: The message here should say that the video's `channelId` does not match with the channel that was queried based on reward (or payee) account.

Alternatively you can provide `channelId` as part of the `where` condition when querying for the video, so that you only search within the correct channel to begin with.

**Edit:** I now also noticed that you don't `return` here, the way you do if the video is not found. I think it's better to treat both those cases the same way (ie. if the video does not belong to the channel, the context is invalid and should be set to `null`)

query-node/mappings/src/content/channel.ts
Lezek123: I think ideally you would throw a new, speicifc kind of error like `MetaprotocolError` and then when catching the error you could distinguish between the non-critical `MetaprotocolError` and other errors that should cause the QN to crash, ie.:
```typescript
catch (error) {
  if (error instanceof MetaprotocolError) {
    /* Handle the non-critical metaprotocol error */
  }
  else {
    throw error;
  }
}
```
But there are other places where it would need to be fixed too (like comments and reactions etc.)
Perhaps you could make an issue to potentially fix this in the future (in case it's still needed after launching Orion v2)?

Lezek123 (changes requested):
See previous review (I accidentally submitted it by adding a comment)

Lezek123 (approved):
LGTM

PR #4528, opened 2023-01-16
Fix joystream-node docker build

mnaamani:
use rust nightly to install cargo-chef to fix docker build issue seen here: https://github.com/Joystream/joystream/actions/runs/3890545247/jobs/6662531689#step:12:360


dobertRowneySr (approved):
LGTM

PR #4564, opened 2023-01-23
Feature/app action

WRadoslaw:


WRadoslaw:
@zeeshanakram3 Logic is ready for review, not sure why query-node is failing since I've rebuilt the `josytream/js` package and bumped the version on both `QN` and `test`

WRadoslaw:
It might be better to review after #4504 is merged

Issue #4555, opened 2023-01-19
Re-enable and fix runtime upgrade CI check


mnaamani:
The github workflow job https://github.com/Joystream/joystream/blob/ephesus/.github/workflows/run-network-tests.yml#L188 needs to be updated to test against the latest mainnet runtime. The associated tooling may also not work correctly in `tests/network-tests/run-migration-tests.sh` and should be updated.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203779180308558) by [Unito](https://www.unito.io)

dobertRowneySr:
# Intro & Scope
Substrate forkless updates can be classified in pure *runtime-migration* and *data-migration*, according whether mutation of the *existing* on-chain state is needed, and thus some on-chain transition logic `old_state -> new_state` is needed.
## Fork-off Script
1. Downloads the complete storage state from a live substrate chain from a ws endpoint
2. Generates empty chainspec
3. Fork-off populates the chainspec with the downloaded storage
4. starts node with current live storage
5. Test transition runtime logic by applying a Substrate fork-less upgrade
## Forkless upgrade
via the [frame system `set_code` extrinsic](https://docs.substrate.io/tutorials/get-started/upgrade-a-running-network/), in our case via a governance proposal
# TODOs
Current runtime upgrade mainnet -> ephesus is a pure runtime migration. However if we want to fix this script now: 

- [x] Add a flag that select the fork-off script whether *data-migration* is needed to be tested or not
- [x] remove the root key from the sudo pallet condition if `Alice` account is specified (since we don't use the sudo pallet anymore)
- [ ] update Joystream rpc endpoints from where storage is obtained for the fork-off script
- [ ] Replace `runtime.wasm` swap with runtime upgrade integration test (with the appropriate proposal scenario)

PR #4569, opened 2023-01-24
Ephesus runtime upgrade ci

dobertRowneySr:
Refers to #4555

This needs #4599 in order to have `joystream-node.compact.compressed.wasm` as runtime code for the proposal in order to avoid triggering a block weight exceeded error.

This also needs #4608 as it is based on top of that

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203822321386864) by [Unito](https://www.unito.io)


.github/workflows/run-network-tests.yml
```diff
@@ -185,39 +189,114 @@ jobs:
           export NO_STORAGE=${{ matrix.no_storage }}
           tests/network-tests/run-tests.sh ${{ matrix.scenario }}
 
-  runtime_upgrade:
-    # Re-Enable when we want to test carthage runtime updates
-    if: ${{ false }}
-    name: Runtime Upgrade From Carthage Spec 0
+  build_master_testing:
```
mnaamani, line 192: There is some yaml formatting error here that I can't spot that is preventing workflows from running.
Maybe try a yaml format checker ..

query-node/run-tests.sh
```diff
@@ -37,6 +37,6 @@ export IGNORE_HIRED_LEADS=true # this directive is needed to run `full` scenario
 # pass the scenario name without .ts extension
 SCENARIO=$1
 # fallback if scenario if not specified
-SCENARIO=${SCENARIO:="content-directory"}
+SCENARIO=${SCENARIO:="full"}
```
mnaamani, line 40: why this change? I think the scenario chosen here was perhaps to test only query-node without depending on storage infra

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
 cd $SCRIPT_PATH
 
-# The joystream/node docker image tag which contains WASM runtime to upgrade chain with
-TARGET_RUNTIME_TAG=${TARGET_RUNTIME_TAG:=$(../../scripts/runtime-code-shasum.sh)}
-# The joystream/node docker image tag to start the chain with
-RUNTIME_TAG=${RUNTIME_TAG:=6740a4ae2bf40fe7c670fb49943cbbe290277601}
-# Post migration assertions by means of typescript scenarios required
-POST_MIGRATION_ASYNC_ASSERTIONS=${POST_MIGRATION_ASYNC_ASSERTIONS:=$true}
-# source common function used for node setup
-source ./node-utils.sh
```
mnaamani, line 16: node-utils.sh is not "sourced" anymore, so we can drop the whole script?

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
+                echo $INITIAL_BALANCES >${DATA_PATH}/initial-balances.json
+            else
+                echo >&2 "Failed to parse INITIAL_BALANCES as JSON, or got false/null"
+            fi
+        fi
+    fi
+}
+
+# Create a chain spec file
+function create_hex_chain_spec() {
```
mnaamani: "hex" is not very meaningful here. "raw" would make more sense.

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
+    docker run --rm -v ${DATA_PATH}:/spec --entrypoint ./chain-spec-builder joystream/node:${RUNTIME} \
+        new \
+        --fund-accounts \
+        --authorities //Alice \
+        --sudo-account ${SUDO_ACCOUNT} \
```
mnaamani: Can we drop sudo related arguments so this workflow would not need to be modified again post ephesus. Perhaps using 'generate' command instead of 'new'. In that case you would need to parse the stdout to get the authority keys and specify the keystore path when running the node.

tests/network-tests/run-migration-tests.sh
mnaamani: I think this script should be renamed to run-runtime-upgrade-tests.sh its not just about migration really.

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
+    if [ "${NO_QN}" != true ]; then
+        ../../query-node/start.sh
+    fi
 
-    echo "***** STARTING NODE WITH FORKED STATE *****"
-    CONTAINER_ID=$(start_node)
+    CONTAINER_ID=""
+    export JOYSTREAM_NODE_TAG=${RUNTIME}
+    if [ $TARGET_RUNTIME != $RUNTIME ]; then
+        if ! [[ -f ${DATA_PATH}/chain-spec-raw.json ]]; then
```
mnaamani: I had a bit of a hard time testing when some steps failed. Why not just always clear the data folder and start from scratch.

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
+    # Start a query-node
+    if [ "${NO_QN}" != true ]; then
+        ../../query-node/start.sh
+    fi
 
-    echo "***** STARTING NODE WITH FORKED STATE *****"
-    CONTAINER_ID=$(start_node)
+    CONTAINER_ID=""
+    export JOYSTREAM_NODE_TAG=${RUNTIME}
+    if [ $TARGET_RUNTIME != $RUNTIME ]; then
```
mnaamani: I think the script should exit with an error if `$TARGET_RUNTIME == $RUNTIME` it doesn't make sense to continue really, because its a user input mistake.

utils/api-scripts/src/fork-off.ts
```diff
@@ -2,17 +2,19 @@ import { xxhashAsHex } from '@polkadot/util-crypto'
...
+const cmdArgs = process.argv.slice(2)
+const dataPath = cmdArgs[0].toString() || ''
+const specPath = path.join(dataPath, 'chain-spec-raw.json')
+const forkedSpecPath = path.join(dataPath, 'chain-spec-forked.json')
+const storagePath = path.join(dataPath, 'storage.json')
+
 
 // this might not be of much use
-const provider = new WsProvider(process.env.WS_RPC_ENDPOINT || 'ws://localhost:9944')
+const rpcRemoteEndpoint = cmdArgs[1].toString() || 'ws://localhost:9944'
```
mnaamani: All these globals can be local in main()
And not clear what the comment `"// this might not be of much use"` is for

.env
```diff
@@ -94,3 +94,6 @@ BALANCE_LOCKED=300000000000
 # We do not provide a default value - scripts that startup a joystream-node service
 # Should be explicit about what version to use.
 # JOYSTREAM_NODE_TAG=latest
+
+# Whether data migration is needed for the runtime upgrade
+DATA_MIGRATION_NEEDED=$false
```
mnaamani: I didn't find any reference to this variable anywhere.

bin/utils/chain-spec-builder/src/main.rs
```diff
@@ -61,7 +61,7 @@ impl Into<ChainType> for ChainDeployment {
     }
 }
 
-/// A utility to easily create a testnet chain spec definition with a given set
```
mnaamani: No need to to make any changes to runtime file here, just on principle. This means also the image for ephesus is already built and available from docker hub.

query-node/run-tests.sh
mnaamani, line 40: I don't understand why this was changed.

tests/network-tests/node-utils.sh
```diff
@@ -61,8 +61,8 @@ function create_chainspec_file {
 	   --authorities //Alice \
 	   --deployment dev \
 	   --chain-spec-path /data/chain-spec.json \
-	   --initial-balances-path /data/initial-balances.json
-}
+	   # --initial-balances-path /data/initial-balances.json # no balances
```
mnaamani: This whole file can be dropped? It is not imported anywhere.

tests/network-tests/run-migration-tests.sh
```diff
@@ -1,87 +1,194 @@
...
+# Log only to stderr
+# Only output from this script should be the container id of the node at the very end
+
+# Location that will be mounted to /spec in containers
+# This is where the initial balances files and generated chainspec files will be located.
+DATA_PATH=$PWD/data
+mkdir -p ${DATA_PATH}
+
+# The latest docker image tag to use for joystream/node (testing profile)
+RUNTIME=${RUNTIME:=mainnetDev}
```
mnaamani: if we can keep the names of the runtimes general would be better. mainnet doesn't make sense, but maybe "current" and "target".. something that works when this code is on a future branch post ephesus .. and with the comment "The latest docker image tag" it is even more confusing.

mnaamani (changes requested):
I struggled to test locally, but overall the tests are working successfully.
My comments are mainly improvements and some clarification to make code a bit more maintainable, easier to use to test locally. It should become part of regular tooling used by council/community to test runtime upgrade proposals.

mnaamani:
Last comment, it might not be necessary to rebuild the docker image from master in the run-network-tests, it should already be available in docker hub. And it might make sense to skip the test when it is running on the master branch, because the test would probably fail, and wouldn't make sense anyways. (starting runtime and target would be the same)

tests/network-tests/src/scenarios/full.ts
```diff
@@ -43,23 +43,16 @@ scenario('Full', async ({ job, env }) => {
   // Runtime upgrade should always be first job
   // (except councilJob, which is required for voting and should probably depend on the "source" runtime)
   const councilJob = job('electing council', electCouncil)
-  const runtimeUpgradeProposalJob = env.RUNTIME_UPGRADE_TARGET_WASM_PATH
```
mnaamani: Looking at this again, I think this job should be kept as it is designed to test all possible outcomes.
For the runtime upgrade github action specifically, we only need to perform a normal scenario where one proposal is made and passes. So maybe have a flow just for that purpose?

.env
dobertRowneySr: deleted

tests/network-tests/node-utils.sh
dobertRowneySr: removed

tests/network-tests/run-runtime-upgrade-tests.sh
```diff
@@ -0,0 +1,195 @@
...
+# Location that will be mounted to /spec in containers
+# This is where the initial balances files and generated chainspec files will be located.
+DATA_PATH=$PWD/data
+mkdir -p ${DATA_PATH}
+
+# The latest docker image tag to use for joystream/node (testing profile)
+RUNTIME=${RUNTIME:=current}
+TARGET_RUNTIME=${TARGET_RUNTIME:=target}
+
+# Initial account balance for sudo account
```
mnaamani: Do we sill need //Alice to have a balance?

.github/workflows/run-network-tests.yml
```diff
@@ -189,39 +189,115 @@ jobs:
...
+          path: joystream-node-docker-image-master.tar.gz
+
   runtime_upgrade:
-    # Re-Enable when we want to test carthage runtime updates
-    if: ${{ false }}
-    name: Runtime Upgrade From Carthage Spec 0
-    needs: build_images
+    name: Runtime Upgrade From Mainnet To Ephesus
+    needs: [build_master_testing, build_images]
+    if: github.ref != 'refs/heads/master'
```
mnaamani, line 284: previous stage `build_master_testing` can also be similarly optional

.github/workflows/run-network-tests.yml
```diff
@@ -189,39 +189,115 @@ jobs:
...
+        with:
+          name: ${{ steps.compute_shasum.outputs.shasum }}-joystream-node-docker-image.tar.gz
+          path: joystream-node-docker-image-master.tar.gz
+
   runtime_upgrade:
-    # Re-Enable when we want to test carthage runtime updates
-    if: ${{ false }}
-    name: Runtime Upgrade From Carthage Spec 0
-    needs: build_images
+    name: Runtime Upgrade From Mainnet To Ephesus
```
mnaamani:
```suggestion
    name: Runtime Upgrade From production runtime
```

.github/workflows/run-network-tests.yml
```diff
@@ -189,39 +189,115 @@ jobs:
...
+          ref: master
+
+      - id: compute_shasum
+        env:
+          RUNTIME_PROFILE: 'TESTING'
+        run: |
+          export RUNTIME_CODE_SHASUM=`scripts/runtime-code-shasum.sh`
+          echo "::set-output name=shasum::${RUNTIME_CODE_SHASUM}"
+
+      - name: Setup cache directory
```
mnaamani, line 212: I think we should be able to simply just fetch the image from docker hub because the `joystream-node-docker-dev.yml` workflow pushes all the profile builds when code is committed to a branch.

mnaamani (changes requested):
The tests are still failing.
Left just a few final change requests to speedup the runtime upgrade flow.

tests/network-tests/src/scenarios/full.ts
```diff
@@ -47,9 +47,7 @@ scenario('Full', async ({ job, env }) => {
     ? job('runtime upgrade proposal', runtimeUpgradeProposal).requires(councilJob)
     : undefined
 
-  const coreJob = runtimeUpgradeProposalJob || councilJob
-
-  // All other jobs should be executed after coreJob
+  const coreJob = councilJob || runtimeUpgradeProposalJob
```
mnaamani: This change is not correct.
If a runtime upgrade job is running then the core job is the runtime upgrade job (which itself depends on council job). Or you will have tests potentially running before and after upgrade.

dobertRowneySr:
there's a problem with the full scenario itself see: https://github.com/Joystream/joystream/pull/4679
I think we should merge this and then move on in order to fix CI checks as a subsequent PR

mnaamani (approved):
LGTM
Failing test in full scenario will be addressed separately.

Issue #4524, opened 2023-01-13
Channel payouts proposal parameters should not include `uploader_account`


Lezek123:
**The problem**

When creating `UpdateChannelPayouts` proposal the user provides [UpdateChannelPayoutsParameters](https://github.com/Joystream/joystream/blob/master/runtime-modules/content/src/types.rs#L780). Those parameters include `payload.uploader_account` which is then used to pay the bloat bond and fees for uploading the payload to council storage bag.

The problem is that there is no signature verification for this account, so it can be set by malicious proposal creator to any account at all (including runtime module accounts), which with the help of malicious or irresponsible council can lead to serious security issues.

**Possible fix**

One of the possible fixes would be to disallow providing this account as part of the parameters and let the runtime set it to proposal creator's member controller account instead (during proposal creation).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203721599603662) by [Unito](https://www.unito.io)

PR #4570, opened 2023-01-24
Prevent providing arbitrary `uploader_account` as part of `UpdateChannelPayoutsParameters`

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4524

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203822787368861) by [Unito](https://www.unito.io)


dobertRowneySr (changes requested):
Runtime logic looks correct. Did you update also the `weight.rs` file?

Lezek123:
> Runtime logic looks correct. Did you update also the weight.rs file?

Updated in https://github.com/Joystream/joystream/pull/4570/commits/e50573800705aa5fd17185570510ccfed9eff0b2, but ofc final weights need to be generated on the reference machine (cc: @mnaamani)

dobertRowneySr (approved):
LGTM

mnaamani (approved):
Using the proposer as the up loader account seems like the right approach here.
Tested locally. I think we can do a final weight generation for content  module before release.

PR #4584, opened 2023-01-30
JoystreamJS: Generate commitment from arbitrary bytes source

thesan:
This is needed by Joystream/pioneer#4095 because`fs` is not available in the browser. Instead the payload file is provided to the browser as a [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File).


thesan:
@zeeshanakram3 BTW this is needed in Pioneer [here](https://github.com/Joystream/pioneer/blob/dev/packages/ui/src/common/utils/crypto/index.ts#L17)

zeeshanakram3 (approved):
LGTM

PR #4586, opened 2023-01-30
Joystream js app action commitment helper

WRadoslaw:
This function will be used by: QN, network-tests and Orion


Lezek123 (approved):
LGTM

Issue #4579, opened 2023-01-27
QN: Event handlers incorrectly update `channel.cumulativeRewardClaimed`


Lezek123:
The 3rd parameter in `ChannelRewardClaimedAndWithdrawn` event is the cashout amount, but the query node sets `channel.cumulativeRewardClaimed` to this amount when handling the event: https://github.com/Joystream/joystream/blob/ephesus/query-node/mappings/src/content/channel.ts#L540

The problem occurs when there are 2 payouts to the same channel:
1. First we set channel payouts where channel `1` receives `x` JOY
2. We claim the payout
3. We update channel payouts, now channel `1` recieves `x + y`
4. We now claim using `claimAndWithdrawChannelReward` extrinsic
5. The query node state shows `channel.cumulativeRewardClaimed` for channel `1` to be `y`, while in the chain state it is `x + y`

I executed those steps with `x = 2,666,666,666,660` and `y = 3,000,000,000,000` and confirmed the bug.

Chain state:
![image](https://user-images.githubusercontent.com/19978260/215140142-49315324-b922-42df-8b40-5ce00b49a274.png)

QN state:
![image](https://user-images.githubusercontent.com/19978260/215140328-dbd3218d-ceca-4613-abe5-73030ef9de2f.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203846261411672) by [Unito](https://www.unito.io)

Lezek123:
Edit: Found another issue.
In https://github.com/Joystream/joystream/blob/ephesus/query-node/mappings/src/content/channel.ts#L494 :
```
channel.cumulativeRewardClaimed = channel.cumulativeRewardClaimed?.add(claimedAmount.toBn())
```
This means that if `channel.cumulativeRewardClaimed` is currently `undefined`/`null`, the `add` will never be triggered and the `channel.cumulativeRewardClaimed` will remain unset. So until `ChannelRewardClaimedAndWithdrawn` is processed, the `cumulativeRewardClaimed` remains `null`

PR #4594, opened 2023-01-31
Fix channel payouts mappings and add missing tests

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4579 and adds test coverage

### DIFF
--- a/query-node/mappings/src/content/channel.ts
+++ b/query-node/mappings/src/content/channel.ts
@@ -470,7 +470,7 @@ export async function content_ChannelPayoutsUpdated({ store, event }: EventConte
 
 export async function content_ChannelRewardUpdated({ store, event }: EventContext & StoreContext): Promise<void> {
   // load event data
-  const [, claimedAmount, channelId] = new Content.ChannelRewardUpdatedEvent(event).params
+  const [cumulativeRewardEarned, claimedAmount, channelId] = new Content.ChannelRewardUpdatedEvent(event).params
 
   // load channel
   const channel = await store.get(Channel, { where: { id: channelId.toString() } })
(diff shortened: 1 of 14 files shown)


query-node/mappings/src/content/channel.ts
```diff
@@ -501,7 +501,7 @@ export async function content_ChannelRewardClaimedAndWithdrawn({
   event,
 }: EventContext & StoreContext): Promise<void> {
   // load event data
-  const [owner, channelId, amount, accountId] = new Content.ChannelRewardClaimedAndWithdrawnEvent(event).params
+  const [owner, channelId, amount, destination] = new Content.ChannelRewardClaimedAndWithdrawnEvent(event).params
```
dobertRowneySr: maybe replace `amount` with `withdrawnAmount` since it's easier to understand

dobertRowneySr (changes requested):
it's ok, I just left a cosmetic suggestion.

 I just want to mention 2 things:
1. we are not having any tests for curator channel in which `claim_and_withdraw` would leave the council budget unaffected theoretically. 
2. we are not making any checks to ensure proper `api.query.council.budget` variations

I am not sure if they should be implemented here or they can be checked in other way (QA)

query-node/mappings/src/content/channel.ts
Lezek123: Done in https://github.com/Joystream/joystream/pull/4594/commits/d7013f7523758e1d1211056e968d91fc2544002e

Lezek123:
> we are not having any tests for curator channel in which claim_and_withdraw would leave the council budget unaffected theoretically.

We have unit tests in the runtime, but we don't have integration tests. We don't really have integration tests for curator channels in general, which is quite unfortunate.

> we are not making any checks to ensure proper api.query.council.budget variations

Added some basic checks in https://github.com/Joystream/joystream/pull/4594/commits/1320d219cc9dacba79ab5087f39867511b94768f

dobertRowneySr (approved):
LGTM, merging

Issue #4573, opened 2023-01-25
QN: Improve Metaprotocol transactions error handling


zeeshanakram3:
As pointed out by [Lezsek](https://github.com/Joystream/joystream/pull/4517#discussion_r1086299357), we need improved & efficient implementation of error handling logic for metaprotocol txs which include features like `comments/reactions`, `videoCategory`, `channelpayment`, `apps` etc.

Further ...
> Going forward it would be good to make metaprotocol transaction errors just enum variants instead of storing the metaprotocol error message in the database, as this would eventually become a lot of wasted storage, and it's unproportionally cheap to spam QN database this way. The frontend could decide what kind of error message to display based on the error code/variant instead. Also it would be language-agnostic approach

PR #4597, opened 2023-02-02
Implement apps as member owned

### LINKED ISSUE #4307: Apps as first-class citizens

drillprop:
Fix #4307 and https://github.com/Joystream/atlas/issues/3517


query-node/mappings/src/content/app.ts
```diff
@@ -1,43 +1,46 @@
...
+
+  let ownerMember: Membership | undefined
+  if (memberId) {
+    ownerMember = await store.get(Membership, {
+      where: {
+        id: memberId,
+      },
+    })
+    if (!ownerMember) {
+      return inconsistentState(`Member with ${memberId} not found`)
```
Lezek123: This should never actually occur, as the runtime should never emit `MemberRemarked` with non-existing member id.

If it does occur, it's a critical error and should crash the QN. This will however not happen in this case, because any error you throw from `processCreateAppMessage` is later caught in a `catch` block of `members_MemberRemarked` (there is already an issue to address this: https://github.com/Joystream/joystream/issues/4573).

Therefore:
- You can remove this check as it is redundant
-  You can set `ownerMember` to simply `memberId && new Membership({ id: memberId })`
- If you want to add `inconsistentState` safeguard in case there is a runtime/QN state mismatch, it should be added in `members_MemberRemarked` function instead (and outside of the `try-catch` block)

query-node/mappings/src/content/app.ts
```diff
@@ -1,43 +1,46 @@
...
+    }
+  }
 
   const isAppExists = await store.get(App, {
     where: {
-      name: name,
+      name: message?.name,
```
Lezek123: message cannot be empty, so it should be either:
```suggestion
      name: message.name,
```
or `message` should have `ICreateApp | undefined` type (if it's possible for it to be empty).

query-node/mappings/src/content/app.ts
```diff
@@ -1,43 +1,46 @@
...
-  // load channel
-  const channel = await store.get(Channel, {
-    where: { id: channelId.toString() },
-  })
-
-  // ensure channel exists
-  if (!channel) {
-    return inconsistentState('Non-existing channel update requested', channelId)
+    logger.error('App already exists', { name })
+    return
```
Lezek123: Here the function returns nothing and throws no error, so it behaves the same way it does when the processing was successful, which means it should later trigger `saveMetaprotocolTransactionSuccessful` in `members_MemberRemarked`.

query-node/mappings/src/content/app.ts
```diff
@@ -1,43 +1,46 @@
...
   }
 
   const newApp = new App({
     name,
     id: appId,
+    ownerMember: ownerMember || undefined,
+    isLeadOwned: !memberId,
```
Lezek123, line 37: Isn't this redundant, since it should be always equal to `!!app.ownerMember`?

query-node/mappings/src/content/app.ts
```diff
@@ -49,32 +52,19 @@ export async function processCreateAppMessage(
...
-  const app = await getAppById(store, appId)
+  const app = await getAppByIdAndMemberId(store, appId)
 
   if (!app) {
-    inconsistentState("App doesn't exists; appId:", appId)
-  }
-  if (app?.channel.id !== channelId.toString()) {
-    inconsistentState(`Cannot update app; app does not belong to the channelId: `, channelId)
+    logger.error("App doesn't exists or doesn't belong to the member", { appId })
+    return
```
Lezek123: Same issue as previously pointed out: returning nothing is equivalent to success condition

query-node/mappings/src/content/app.ts
```diff
@@ -97,12 +87,26 @@ export async function processUpdateApp(
...
 
-async function getAppById(store: DatabaseManager, appId: string): Promise<App | undefined> {
+export async function processDeleteAppMessage(store: DatabaseManager, message: IDeleteApp): Promise<void> {
+  const { appId } = message
+
+  const app = await getAppByIdAndMemberId(store, appId)
+
+  if (!app) {
+    logger.error("App doesn't exists", { appId })
+    return
```
Lezek123: Same issue as in the previous comment

query-node/mappings/src/content/app.ts
```diff
@@ -49,32 +52,19 @@ export async function processCreateAppMessage(
...
-export async function processUpdateApp(
-  store: DatabaseManager,
-  channelId: ChannelId,
-  message: IUpdateApp
-): Promise<void> {
+export async function processUpdateAppMessage(store: DatabaseManager, message: IUpdateApp): Promise<void> {
   const { appId, appMetadata } = message
 
-  const app = await getAppById(store, appId)
+  const app = await getAppByIdAndMemberId(store, appId)
```
Lezek123: The function is called `getAppByIdAndMemberId` but only the `appId` is used in the query.
There is actually a security issue here too, because you're not checking anywhere that `app.ownerMember` is actually the member that submitted the remark transaction (or lead in case `app.ownerMember` is not set)

query-node/mappings/src/content/app.ts
```diff
@@ -97,12 +87,26 @@ export async function processUpdateApp(
   logger.info('App has been updated', { appId })
 }
 
-async function getAppById(store: DatabaseManager, appId: string): Promise<App | undefined> {
+export async function processDeleteAppMessage(store: DatabaseManager, message: IDeleteApp): Promise<void> {
+  const { appId } = message
+
+  const app = await getAppByIdAndMemberId(store, appId)
```
Lezek123: Same issue as in `processUpdateAppMessage`. Owner not being actually checked.

query-node/mappings/src/membership.ts
```diff
@@ -655,6 +656,18 @@ async function processMemberRemark(
...
+  }
+
+  if (messageType === 'updateApp') {
+    await processUpdateAppMessage(store, decodedMessage.updateApp!)
+  }
+
+  if (messageType === 'deleteApp') {
+    await processDeleteAppMessage(store, decodedMessage.deleteApp!)
+  }
+
```
Lezek123: In all those if-blocks there is no `return` statement, so the function continues until it hits:
```typescript
  // unknown message type
  return inconsistentState('Unsupported message type in member_remark action', messageType)
```

query-node/mappings/src/workingGroups.ts
```diff
@@ -688,6 +691,17 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
   await store.save<StatusTextChangedEvent>(statusTextChangedEvent)
 }
 
+export async function contentWorkingGroups_LeadRemarked({ store, event }: EventContext & StoreContext): Promise<void> {
+  const [metadataByte] = new WorkingGroups.LeadRemarkedEvent(event).params
+
+  const leadRemarkMetadata = deserializeMetadata(LeadRemarked, metadataByte)
```
Lezek123:
```suggestion
  const [metadataBytes] = new WorkingGroups.LeadRemarkedEvent(event).params

  const leadRemarkMetadata = deserializeMetadata(LeadRemarked, metadataBytes)

File: query-node/mappings/src/workingGroups.ts [hunk 11]
```diff
@@ -688,6 +691,17 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
+  if (leadRemarkMetadata) {
+    await processLeadRemarked(store, event, leadRemarkMetadata)
+  } else {
+    return invalidMetadata('Unrecognized remarked action')
+  }
```
Lezek123: You're using the same functions to process app-related actions in both `LeadRemarked` and `MemberRemarked` handlers, but note that in `MemberRemarked` handler there is a try-catch block around `await processMemberRemark` and here there is none. This means that throwing an error from `processCreateAppMessage` / `processUpdateAppMessage` / `processDeleteAppMessage` has 2 different meanings depending on which extrinsic you're processing:
1. When one of those function throws while `MemberRemarked` is being processed - it signals metaprotocol transaction error (so an error is just stored in `MetaprotocolTransactionStatusEvent`)
2. When one of those function throws while `LeadRemarked` is being processed - it crashes the QN

Also note that here you're not saving any `MetaprotocolTransactionStatusEvent`, so unless there is some additional logic introduced in Atlas, it won't recognize the result of this transaction.

You should keep the implementation here consitent with `members_MemberRemarked`

tests/network-tests/src/Api.ts
```diff
@@ -807,48 +808,72 @@ export class Api {
...
-        name,
-        appMetadata,
-      },
+  async updateApp(memberId: u64, appId: string, appMetadata: IAppMetadata): Promise<ISubmittableResult> {
+    const memberAccount = await this.getMemberControllerAccount(memberId.toNumber())
+    if (!memberAccount) {
+      throw new Error('invalid member id')
     }
 
+    const meta = new LeadRemarked({
```
Lezek123: Shouldn't that depend on the app owner? (ie. `LeadRemark` for DAO-owned apps, `MemberRemark` for member-owned apps)

tests/network-tests/src/Api.ts
```diff
@@ -807,48 +808,72 @@ export class Api {
...
-  ): Promise<ISubmittableResult> {
+  async deleteApp(memberId: u64, appId: string): Promise<ISubmittableResult> {
     const memberAccount = await this.getMemberControllerAccount(memberId.toNumber())
     if (!memberAccount) {
       throw new Error('invalid member id')
     }
 
-    const meta = new ChannelOwnerRemarked({
-      updateApp: new UpdateApp({
+    const meta = new LeadRemarked({
```
Lezek123: Same comment as above

query-node/mappings/src/workingGroups.ts
Lezek123: Although `deserializeMetadata` is a recommended way of processing the metadata, there is now an inconsistency, since `MemberRemarked` still uses `MemberRemarked.decode`, so `processUpdateAppMessage` will behave differently depending on whether it's part of `MemberRemark` or `LeadRemark`.

tests/network-tests/src/flows/membership/createApp.ts
```diff
@@ -0,0 +1,68 @@
...
+      assert.equal(appsByName?.[0]?.category, appOwnedByMemberMetadata.category)
+      assert.equal(appsByName?.[0]?.oneLiner, appOwnedByMemberMetadata.oneLiner)
+      assert.equal(appsByName?.[0]?.description, appOwnedByMemberMetadata.description)
+      assert.equal(appsByName?.[0]?.termsOfService, null)
+      assert.equal(appsByName?.[0]?.websiteUrl, null)
+      assert.deepEqual(appsByName?.[0]?.platforms, appOwnedByMemberMetadata.platforms)
+    }
+  )
+
+  const appOwnedByLead = 'app_owned_by_lead'
```
Lezek123: This suggests that perhaps `membership` is not the right category for those flows. I think it should probably be in `content`, since we already have other metaprotocol stuff that relies on `MemberRemark` there, like comments & reactions.

tests/network-tests/src/scenarios/full.ts
```diff
@@ -113,6 +115,7 @@ scenario('Full', async ({ job, env }) => {
   )
   job('create app', createApp).after(sudoHireLead)
   job('update app', updateApp).after(sudoHireLead)
+  job('update app', deleteApp).after(sudoHireLead)
```
Lezek123:
```suggestion
  job('delete app', deleteApp).after(sudoHireLead)
```

Lezek123 (changes requested):

query-node/mappings/src/content/app.ts
```diff
@@ -49,32 +46,32 @@ export async function processCreateAppMessage(
...
   const app = await getAppById(store, appId)
 
   if (!app) {
-    inconsistentState("App doesn't exists; appId:", appId)
+    inconsistentState("App doesn't exists", { appId })
   }
-  if (app?.channel.id !== channelId.toString()) {
-    inconsistentState(`Cannot update app; app does not belong to the channelId: `, channelId)
+
+  if (memberId && app?.ownerMember?.id !== memberId) {
```
Lezek123, line 73: This means the lead can update any app, because `memberId` is `undefined` in this case, so the condition is never met. Was this intended?

query-node/mappings/src/content/app.ts
```diff
@@ -97,12 +94,40 @@ export async function processUpdateApp(
...
+    inconsistentState("memberId wasn't provided")
+  }
+
+  const app = await getAppById(store, appId)
+
+  if (!app) {
+    inconsistentState("App doesn't exists", { appId })
+  }
+
+  if (memberId && app?.ownerMember?.id !== memberId) {
```
Lezek123, line 120: Same comment as above

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -356,10 +357,10 @@ export async function processCreateCommentMessage(
   store: DatabaseManager,
   event: SubstrateEvent,
   memberId: MemberId,
-  message: ICreateComment
+  metadata: DecodedMetadataObject<ICreateComment>
 ): Promise<Comment> {
   // in case of null `parentCommentId` protobuf would assign it a default value i.e. ''
```
Lezek123: This comment no longer applies
```suggestion
```

query-node/mappings/src/content/commentAndReaction.ts
```diff
@@ -192,9 +193,9 @@ export async function processReactVideoMessage(
   store: DatabaseManager,
   event: SubstrateEvent,
   memberId: MemberId,
-  message: IReactVideo
+  metadata: DecodedMetadataObject<IReactVideo>
```
Lezek123, line 196: For consistancy, similar change is needed in:
- `processReactCommentMessage`
- `processEditCommentMessage`
- `processDeleteCommentMessage`
- `createVideoCategory` (also please remove type assetion `decodedMetadata?.createVideoCategory as ICreateVideoCategory` in `processMemberRemark`)

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
   store: DatabaseManager,
   event: SubstrateEvent,
   memberId: MemberId,
-  decodedMessage: MemberRemarked
+  decodedMetadata: DecodedMetadataObject<IMemberRemarked> | null
 ): Promise<Partial<MetaprotocolTransactionSuccessful>> {
-  const messageType = decodedMessage.memberRemarked
+  if (decodedMetadata?.createApp) {
+    await processCreateAppMessage(store, event, decodedMetadata.createApp!, memberId.toString())
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    await processCreateAppMessage(store, event, decodedMetadata.createApp, memberId.toString())
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
 
-  if (messageType === 'reactVideo') {
-    await processReactVideoMessage(store, event, memberId, decodedMessage.reactVideo!)
+    return {}
+  }
+
+  if (decodedMetadata?.updateApp) {
+    await processUpdateAppMessage(store, event, decodedMetadata.updateApp!, memberId.toString())
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    await processUpdateAppMessage(store, event, decodedMetadata.updateApp, memberId.toString())
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
+
+    return {}
+  }
+
+  if (decodedMetadata?.deleteApp) {
+    await processDeleteAppMessage(store, event, decodedMetadata.deleteApp!, memberId.toString())
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    await processDeleteAppMessage(store, event, decodedMetadata.deleteApp, memberId.toString())
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
+
+    return {}
+  }
+
+  if (decodedMetadata?.reactVideo) {
+    await processReactVideoMessage(store, event, memberId, decodedMetadata.reactVideo!)
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    await processReactVideoMessage(store, event, memberId, decodedMetadata.reactVideo)
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
 
     return {}
   }
 
-  if (messageType === 'reactComment') {
-    await processReactCommentMessage(store, event, memberId, decodedMessage.reactComment!)
+  if (decodedMetadata?.reactComment) {
+    await processReactCommentMessage(store, event, memberId, decodedMetadata.reactComment!)
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    await processReactCommentMessage(store, event, memberId, decodedMetadata.reactComment)
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
 
     return {}
   }
 
-  if (messageType === 'createComment') {
-    const comment = await processCreateCommentMessage(store, event, memberId, decodedMessage.createComment!)
+  if (decodedMetadata?.createComment) {
+    const comment = await processCreateCommentMessage(store, event, memberId, decodedMetadata.createComment!)
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    const comment = await processCreateCommentMessage(store, event, memberId, decodedMetadata.createComment)
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
 
     return { commentCreatedId: comment.id }
   }
 
-  if (messageType === 'editComment') {
-    const comment = await processEditCommentMessage(store, event, memberId, decodedMessage.editComment!)
+  if (decodedMetadata?.editComment) {
+    const comment = await processEditCommentMessage(store, event, memberId, decodedMetadata.editComment!)
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    const comment = await processEditCommentMessage(store, event, memberId, decodedMetadata.editComment)
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
     return { commentEditedId: comment.id }
   }
 
-  if (messageType === 'deleteComment') {
-    const comment = await processDeleteCommentMessage(store, event, memberId, decodedMessage.deleteComment!)
+  if (decodedMetadata?.deleteComment) {
+    const comment = await processDeleteCommentMessage(store, event, memberId, decodedMetadata.deleteComment!)
```
Lezek123: Inside this if-block the non-null assertion is no longer needed:
```suggestion
    const comment = await processDeleteCommentMessage(store, event, memberId, decodedMetadata.deleteComment)
```

query-node/mappings/src/membership.ts
```diff
@@ -651,46 +653,62 @@ async function processMemberRemark(
...
     return { commentDeletedId: comment.id }
   }
 
-  if (messageType === 'createVideoCategory') {
-    const createParams = decodedMessage.createVideoCategory as ICreateVideoCategory
+  if (decodedMetadata?.createVideoCategory) {
+    const createParams = decodedMetadata?.createVideoCategory as ICreateVideoCategory
 
     const videoCategory = await createVideoCategory(store, event, createParams)
```
Lezek123: Please remove the redundant type assertion
```suggestion
    const videoCategory = await createVideoCategory(store, event, decodedMetadata.createVideoCategory)
```

Lezek123 (changes requested):

metadata-protobuf/proto/Metaprotocol.proto
```diff
@@ -156,11 +134,46 @@ message AppMetadata {
...
 message UpdateApp {
     required string app_id = 1;
-    optional AppMetadata app_metadata = 3;
+    optional AppMetadata app_metadata = 2;
+}
+message DeleteApp {
+    required string app_id = 1;
+}
+
+message LeadRemarked {
```
Lezek123: One more cosmetic change crossed my mind: I think this message should be named `ContentLeadRemarked`, since it is not a generic lead remark, but rather a lead remark from a specific working group in this case.

drillprop:
@Lezek123 I don't know why tests are failing. Tried locally, and everything seems fine to me - tests are passing.

tests/network-tests/src/flows/content/createApp.ts
```diff
@@ -47,7 +49,7 @@ export async function createApp({ api, query }: FlowProps): Promise<void> {
     platforms: ['web', 'mobile'],
   }
 
-  await api.createApp(member.memberId, appOwnedByLead, appOwnedByLeadMetadata, true)
+  await api.createApp(leadId, appOwnedByLead, appOwnedByLeadMetadata, true)
```
Lezek123: I think it's better to just make `memberId` arg optional (which will also imply whether the app is member-owned or lead-owned).
`leadId` is actually not even a member id, it's a worker id

tests/network-tests/src/flows/content/updateApp.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  await query.tryQueryWithTimeout(
+    () => query.getAppsByName(newAppNameOwnedByLead),
+    (appsByName) => {
+      assert.equal(appsByName?.[0]?.id, newAppOwnedByLeadId)
+      assert.equal(appsByName?.[0]?.name, newAppNameOwnedByLead)
+      assert.equal(appsByName?.[0]?.ownerMember?.id, null)
+      assert.equal(appsByName?.[0]?.useUri, updatedMetadata.useUri)
+      assert.equal(appsByName?.[0]?.category, updatedMetadata.category)
```
Lezek123: > I don't know why tests are failing. Tried locally, and everything seems fine to me - tests are passing.

Error from the workflow (https://github.com/Joystream/joystream/actions/runs/4133895124/jobs/7144408740): 
> 2023-02-09T12:19:00.8946237Z 2023-02-09T12:19:00.893Z integration-tests:job:update app flow 0 failed:
2023-02-09T12:19:00.9697378Z AssertionError: expected 'blockchain' to equal undefined
2023-02-09T12:19:00.9698158Z     at /home/runner/work/joystream/joystream/tests/network-tests/src/flows/content/updateApp.ts:105:14

The `updatedMetadata` only specifies `description`, `oneLiner` and `platforms`, so other properties should remain intact (just like in case of the member app):

```suggestion
      assert.equal(appsByName?.[0]?.useUri, newAppOwnedByLeadMetadata.useUri)
      assert.equal(appsByName?.[0]?.category, newAppOwnedByLeadMetadata.category)
```

Lezek123 (changes requested):

Lezek123 (approved):
LGTM

Issue #4571, opened 2023-01-25
Add integration tests for governance blacklisting


bedeho:
For this https://github.com/Joystream/joystream/pull/4409

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203827270827814) by [Unito](https://www.unito.io)

PR #4598, opened 2023-02-02
Ephesus governance opt out int tests

dobertRowneySr:
addressed #4571 
I added tests for the opt-out voting functionality + some refactoring in order to simplify the execution in the `ElectCouncilFixture`
*Side Note* I left out implementing some of the QN checks after refactoring, as they are out of the scope of this pr

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203891683761476) by [Unito](https://www.unito.io)


tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -35,41 +37,87 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
...
-    await api.sendExtrinsicsAndGetResults(applyForCouncilTxs, candidatesMemberAccounts)
+    await new FixtureRunner(announceCandidacyFixture).run()
 
     // Voting stage
     await this.api.untilCouncilStage('Voting')
 
+    const votersStakingAccounts = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
+    const optOutVoters = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
+    await api.treasuryTransferBalanceToAccounts(votersStakingAccounts, voteStake)
+
```
mnaamani: should we also transfer funds to optOutVoters, to be sure they have the required funds to attempt to vote, or is the check for blacklisted account happen before funds check?

tests/network-tests/src/fixtures/referendum/voteFixture.ts
```diff
@@ -0,0 +1,68 @@
...
+    const accountOrAccounts = await this.getSignerAccountOrAccounts()
+    const extrinsics = await this.getExtrinsics()
+    this.extrinsics = super.flattenExtrinsics(extrinsics)
+    await this.api.prepareAccountsForFeeExpenses(accountOrAccounts, this.extrinsics, 10)
+    this.results = await this.api.sendExtrinsicsAndGetResults(extrinsics, accountOrAccounts, 10)
+
+    if (!this.failureExpected) {
+      this.events = await Promise.all(this.results.map((r) => this.getEventFromResult(r)))
+    } else {
+      if (Array.isArray(accountOrAccounts)) {
```
mnaamani: I think we should be looking for event `AccountAlreadyOptedOutOfVoting` to be sure it is failing for the right reason. It could fail for other reason, incase the account setup in the fixture was incorrect for example.

mnaamani (changes requested):
The refactoring is nice, makes the code easier to parse. I left suggestion for improving the failed vote check case.

tests/network-tests/src/fixtures/council/announceCandidacyFixture.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  protected getExtrinsics(): Promise<
+    SubmittableExtrinsic<'promise', ISubmittableResult>[] | SubmittableExtrinsic<'promise', ISubmittableResult>[][]
+  > {
+    const results = Array.from(this.announcements).map(async ([, params]) =>
+      this.api.tx.council.announceCandidacy(params.memberId, params.stakingAccount, params.rewardAccount, params.stake)
+    )
+    return Promise.all(results)
+  }
```
Lezek123:
```suggestion
  protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise', ISubmittableResult>[]> {
    return Array.from(this.announcements).map(([, params]) =>
      this.api.tx.council.announceCandidacy(params.memberId, params.stakingAccount, params.rewardAccount, params.stake)
    )
  }
```

tests/network-tests/src/fixtures/referendum/blackListVoteFixture.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  protected getSignerAccountOrAccounts(): Promise<string | string[]> {
+    return Promise.resolve(this.accountsToBlacklist)
+  }
+
+  protected getExtrinsics(): Promise<
+    SubmittableExtrinsic<'promise', ISubmittableResult>[] | SubmittableExtrinsic<'promise', ISubmittableResult>[][]
+  > {
+    return Promise.all(this.accountsToBlacklist.map(async () => await this.api.tx.referendum.optOutOfVoting()))
+  }
```
Lezek123: async-await is not needed here:

```suggestion
  protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise', ISubmittableResult>[]> {
    return this.accountsToBlacklist.map(() => this.api.tx.referendum.optOutOfVoting())
  }
```

tests/network-tests/src/fixtures/referendum/voteFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  protected getExtrinsics(): Promise<
+    SubmittableExtrinsic<'promise', ISubmittableResult>[] | SubmittableExtrinsic<'promise', ISubmittableResult>[][]
+  > {
+    const results = Array.from(this.votes).map(async ([, params]) =>
+      this.api.tx.referendum.vote(params.commitment, params.stake)
+    )
+    return Promise.all(results)
+  }
```
Lezek123:
```suggestion
  protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise', ISubmittableResult>[]> {
    return Array.from(this.votes).map(([, params]) => this.api.tx.referendum.vote(params.commitment, params.stake))
  }
```

tests/network-tests/src/fixtures/referendum/voteFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+    const extrinsics = await this.getExtrinsics()
+    this.extrinsics = super.flattenExtrinsics(extrinsics)
+    await this.api.prepareAccountsForFeeExpenses(accountOrAccounts, this.extrinsics, 10)
+    this.results = await this.api.sendExtrinsicsAndGetResults(extrinsics, accountOrAccounts, 10)
+    if (!this.failureExpected) {
+      this.events = await Promise.all(this.results.map((r) => this.getEventFromResult(r)))
+    }
+  }
+
+  public assertError(expectedErrName: string) {
```
Lezek123:
```suggestion
  public assertError(expectedErrName: string): void {
```

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -35,41 +37,88 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
...
     )
-    await api.prepareAccountsForFeeExpenses(candidatesMemberAccounts, applyForCouncilTxs)
-    await api.sendExtrinsicsAndGetResults(applyForCouncilTxs, candidatesMemberAccounts)
+    await new FixtureRunner(announceCandidacyFixture).run()
 
     // Voting stage
     await this.api.untilCouncilStage('Voting')
 
+    const votersStakingAccounts = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
+    const optOutVoters = (await this.api.createKeyPairs(numberOfVoters)).map(({ key }) => key.address)
```
Lezek123: I think there should be a separate flow testing the opt-out mechanics, instead of this being part of the `ElectCouncilFixture`. Or at least the opt-out testing part should be skipped by default, as the fixture is meant to be a minimal, reusable unit.

tests/network-tests/src/scenarios/council.ts
```diff
@@ -5,7 +5,8 @@ import { scenario } from '../Scenario'
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 scenario('Council', async ({ job }) => {
   const councilJob = job('electing council', electCouncil)
-  const secondCouncilJob = job('electing second council', electCouncil).requires(councilJob)
+  // const secondCouncilJob = job('electing second council', electCouncil).requires(councilJob)
```
Lezek123: Why was this job commented-out?

tests/network-tests/src/Fixture.ts
```diff
@@ -138,7 +138,7 @@ export abstract class StandardizedFixture extends BaseQueryNodeFixture {
     })
   }
 
-  private flattenExtrinsics(
+  protected flattenExtrinsics(
```
Lezek123: I think this method is actually pretty redundant, as one can just do `extrinsics.flat()` (which I probably didn't know back then)

tests/network-tests/src/fixtures/referendum/voteFixture.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  protected getEventFromResult(result: ISubmittableResult): Promise<EventDetails<unknown>> {
+    return this.api.getEventDetails(result, 'referendum', 'VoteCast')
+  }
+
+  protected assertQueryNodeEventIsValid(qEvent: AnyQueryNodeEvent, i: number): void {
+    // TODO: implement QN checks
+  }
+
+  // overwriting for using decremental tips
```
Lezek123: There was a comment explaining why it was needed in the council fixture, ie.:
```
    // Due to the fact that we need the transactions to be processed in the expected order
    // (which is not guaranteed by the nonce, because we're using different voter accounts),
    // we'll be including a small, decremental tip (10 JOY * (votersStakingAccounts.length - 1 - accIndex))
```
Could you add it here?

Also:
- It would probably be better to make `StandardizedFixture` have a simple config (like a flag for example) to enable/disable decremental tips
- `10` is a very small amount, I think it was used back when we had no fees & token decimals at all. `10 JOY` (`100_000_000_000`) would be more reasonable I think

Lezek123 (changes requested):

tests/network-tests/src/scenarios/council.ts
dobertRowneySr: I commented out while testing, but I forgot to revert this change

tests/network-tests/src/fixtures/referendum/voteFixture.ts
dobertRowneySr: Is it ok if I update all the balances amount (`10 -> 10_000_000_000`) in a future PR, I can open a issue with a a tech-debt, maybe

Lezek123: It's an easy fix I think, the reason why https://github.com/Joystream/joystream/actions/runs/4160646171/jobs/7197872519 was failing was probably because the `tip` was passed as `number`, so `BN.addn()` failed, as it doesn't support large numbers. Using `new BN(10_000_000_000)` should fix the issue.

tests/network-tests/src/fixtures/referendum/voteFixture.ts
```diff
@@ -0,0 +1,68 @@
...
+    const accountOrAccounts = await this.getSignerAccountOrAccounts()
+    const extrinsics = await this.getExtrinsics()
+    this.extrinsics = extrinsics.flat()
+    await this.api.prepareAccountsForFeeExpenses(accountOrAccounts, this.extrinsics, this.decrementalTip ? 10 : 0)
+    this.results = await this.api.sendExtrinsicsAndGetResults(
+      extrinsics,
+      accountOrAccounts,
+      this.decrementalTip ? 10 : 0
+    )
+    if (!this.failureExpected) {
```
Lezek123: I think it could be a good idea to also make expected failure a configurable part of `StandardizedFixture` aswell.

Actually now that I think about it, you could even do the following to make thing much simpler:
1. Add a method like `setExpectedError(errorName)` to `StandardizedFixture`, which would set `this.expectedError` to `errorName` and accordingly change the logic of `execute` if `this.expectedError` is set (something like `if (this.expectedError) { this.results.forEach((r) => this.expectDispatchError(r, this.expectedError) }`
2. Make `votersStakingAccounts` optionally configurable in `ElectCouncilFixture`
3. Add `setExpectedVotingError(errorName)` to `ElectCouncilFixture` which will set `this.expectedVotingError` and call `votingFixture.setExpectedError(this.expectedVotingError)` if `this.expectedVotingError` is set.
4. You now don't need a separate fixture for voting opt-out test. You can just:
    - Create voter keys
    - Run `BlacklistVoteFixture`
    - Create `ElectCouncilFixture` via `const electCouncilFixture = new ElectCouncilFixture(voterKeys).setExpectedVotingError('AccountAlreadyOptedOutOfVoting')`
    - Run the fixture

I think this will be a very nice solution, as `VotersOptingOut` fixture by itself is not very reusable, it just represents a single test case and duplicates a lot of `ElectCouncilFixture` code.

Additionally I think introducing some standarized way of testing for expected failures can possibly pay off a lot in the future.

Lezek123 (changes requested):

Lezek123 (approved):
Great job, LGTM!

tests/network-tests/src/scenarios/full.ts
```diff
@@ -60,6 +61,7 @@ scenario('Full', async ({ job, env }) => {
   // Council (should not interrupt proposalsJob!)
   const secondCouncilJob = job('electing second council', electCouncil).requires(coreJob)
   const councilFailuresJob = job('council election failures', failToElect).requires(secondCouncilJob)
+  job('council election failure with blacklist', failToElectWithBlacklist).requires(secondCouncilJob)
```
mnaamani: Will there be two council jobs running unless..
```suggestion
  job('council election failure with blacklist', failToElectWithBlacklist).requires(councilFailuresJob)
```

mnaamani (changes requested):
Just a minor issues with council jobs running in parallel in the full scenario, otherwise looks good.

mnaamani (approved):

PR #4575, opened 2023-01-25
Ephesus: output blake3 hash of channel payload file when using CLI

zeeshanakram3:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203892409508638) by [Unito](https://www.unito.io)

Issue #4556, opened 2023-01-19
Merge `mainnet-liberated` into `Epheus`


dobertRowneySr:
SSIA
The mainnet-liberated branch should have https://github.com/Joystream/joystream/pull/4542 
merged

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203780550680374) by [Unito](https://www.unito.io)

dobertRowneySr:
# Details
- [x] proposal `UpdateChannelPayouts` should be activated
- [x] final stage conditions (mainnet liberated) extrinsics should be activated
- [x] integration tests should not depend on sudo pallet (already solved by #4401) according to #3792
- [x] Ephesus Gateway WG should be discarded in favor of the APP WG
- [x] `update_curator_group_permission` call enabled in ephesus but it should be disabled in mainnet as the latest

dobertRowneySr:
Conflict list
```
Auto-merging .github/workflows/run-network-tests.yml
Auto-merging Cargo.lock
CONFLICT (content): Merge conflict in Cargo.lock
Auto-merging bin/node/Cargo.toml
CONFLICT (content): Merge conflict in bin/node/Cargo.toml
Auto-merging bin/utils/chain-spec-builder/Cargo.toml
CONFLICT (content): Merge conflict in bin/utils/chain-spec-builder/Cargo.toml
Auto-merging chain-metadata.json
CONFLICT (content): Merge conflict in chain-metadata.json
Auto-merging cli/README.md
CONFLICT (content): Merge conflict in cli/README.md
Auto-merging cli/package.json
Auto-merging cli/src/Api.ts
Auto-merging cli/src/commands/fee-profile/addVideoComment.ts
CONFLICT (content): Merge conflict in cli/src/commands/fee-profile/addVideoComment.ts
Auto-merging cli/src/graphql/generated/schema.ts
Auto-merging joystream-node.Dockerfile
Auto-merging package.json
Auto-merging query-node/codegen/package.json
CONFLICT (content): Merge conflict in query-node/codegen/package.json
Auto-merging query-node/codegen/yarn.lock
CONFLICT (content): Merge conflict in query-node/codegen/yarn.lock
Auto-merging query-node/manifest.yml
Auto-merging query-node/mappings/package.json
CONFLICT (content): Merge conflict in query-node/mappings/package.json
Auto-merging query-node/mappings/src/common.ts
CONFLICT (content): Merge conflict in query-node/mappings/src/common.ts
[...]

PR #4599, opened 2023-02-03
Ephesus merge mainnet liberated

dobertRowneySr:
merge `mainnet-liberated` into `ephesus` #4556 

# Main points
- the runtime & node version should be correct, but the QN / cli versioning should be reviewed
- ~~I left a comment for the `update_curator_group_permissions` call which I was not sure where it should be enabled or not~~ update: I left the proposal disabled as we don't have integration tests for this yet
- `pallet_membership::create_member` relies on `root` as it was used for founding members. Hence it is not used anymore and I have removed the respective tests from the `full` scenario
- There are still outstading PR targeting `ephesus`: 
    - #4594
    - #4575 
    - #4441
    - #4584
    - #4586
    - #4569 (I will retarget this after this PR has been merged)
    - #4598 (I will retarget this after this PR has been merged)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203892485426670) by [Unito](https://www.unito.io)


bedeho:
Who else should review this?

mnaamani:
The two flows a failing consistently in the network integration tests:

```
channel payouts proposal: Failed
group budget: Failed
```

dobertRowneySr:
> The two flows a failing consistently in the network integration tests:
> 
> ```
> channel payouts proposal: Failed
> group budget: Failed
> ```

In the channelclaim-* fixtures there where Council budget checks. They where concurrent to the budget update logic run at each new block, that's why they where failing. Budget checks would have been nice, but implementing them is time consuming.

The other error was due to inviting members being run concurrently to the upgrade Membership Group budget fixture (so you had 2 concurrent budget updates)

runtime/src/lib.rs
```diff
@@ -234,10 +234,9 @@ impl Contains<<Runtime as frame_system::Config>::Call> for CallFilter {
         match call {
             Call::Content(content::Call::<Runtime>::destroy_nft { .. }) => false,
             Call::Content(content::Call::<Runtime>::toggle_nft_limits { .. }) => false,
-            // REVIEW THIS CHANGE:
-            // Call::Content(content::Call::<Runtime>::update_curator_group_permissions {
-            //     ..
-            // }) => false,
+            Call::Content(content::Call::<Runtime>::update_curator_group_permissions {
+                ..
+            }) => false,
```
mnaamani, line 239: Why was this originally filtered?

utils/api-scripts/src/initialize-lead.ts
```diff
@@ -8,7 +8,7 @@ const workingGroupModules = [
   'contentWorkingGroup',
```
mnaamani: I think we can drop this entierly. It uses sudo. We forgot about it in mainnet-liberated branch

mnaamani:
> In the channelclaim-* fixtures there where Council budget checks. They where concurrent to the budget update logic run at each new block, that's why they where failing. Budget checks would have been nice, but implementing them is time consuming.

I would have assumed these were working before. Why would they stop working in this merge?

ignazio-bovo:
> > In the channelclaim-* fixtures there where Council budget checks. They where concurrent to the budget update logic run at each new block, that's why they where failing. Budget checks would have been nice, but implementing them is time consuming.
> 
> I would have assumed these were working before. Why would they stop working in this merge?

The failing assertion were not directly messed up by the periodical budget bookkeeping, but by the `inviteMember` job being run concurrently to the `channelProposal` one.

dobertRowneySr:
> > In the channelclaim-* fixtures there where Council budget checks. They where concurrent to the budget update logic run at each new block, that's why they where failing. Budget checks would have been nice, but implementing them is time consuming.
> 
> I would have assumed these were working before. Why would they stop working in this merge?

``` typescript
    this.expectedClaims = await getExpectedClaims(this.api, this.claimChannelRewardParams)
    const expectedClaimsSum = this.expectedClaims.reduce((sum, claim) => sum.add(new BN(claim)), new BN(0))
    const councilBudgetPre = await this.api.query.council.budget()
    await super.execute()
    const councilBudgetPost = await this.api.query.council.budget()
    assert.equal(councilBudgetPre.toString(), councilBudgetPost.add(expectedClaimsSum).toString())
  }
```
The execute function reads the council budget before and post execution. However there are multiple budget modifications due to the council budget bookkeeping logic (screenshot below). So the assertion checking only for the channel reward variation fails.

Screenshot 2023-02-07 at 12 02 39

I am investigating why this error has not appeared before, 
Edit: the execution block for the `claim_channel_reward` changes at each test run. So it is possible that previous tests were fortuitous
Edit2: I can confirm that these assertions fails on ephesus @ 2676a17390ce1ccba2dbd9c6514d97e0886a8da8 at the moment

mnaamani:
Ok thanks for investigating that, it would be good to have such assertions as part of these fixtures but it is out of scope of this PR.

mnaamani (approved):

Issue #4582, opened 2023-01-29
Re-generate weights for runtime


mnaamani:
Weights were updated in this PR https://github.com/Joystream/joystream/pull/4570
But as there are several changes to content module, we should do a final weight generation round before release.

Some changes to referendum and membership modules were also made, so benchmarking should cover those as well.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203851001134606) by [Unito](https://www.unito.io)

PR #4409, opened 2022-10-27
Referendum: Add voting opt-out functionality

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/2927

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203688139710670) by [Unito](https://www.unito.io)

PR #4602, opened 2023-02-06
Ephesus adjusted weights

mnaamani:
Implements https://github.com/Joystream/joystream/issues/4582

- Generated weights on reference machine
- Reverted weight changes for extrinsics that were not modified in ephesus

To confirm the revert works, diff against master branch the weights.rs of the pallets

```git
git diff master -- runtime-modules/**/weights.rs
```

You should see only the new generated weights specific to ephesus


runtime-modules/referendum/src/weights.rs
```diff
@@ -140,7 +140,8 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	}
 	// Storage: Instance1Referendum AccountsOptedOut (r:0 w:1)
 	fn opt_out_of_voting() -> Weight {
-		(16_920_000 as Weight)
```
dobertRowneySr, line 143: We do check for duplicates in `opt-out` extr., but that read seems not to be accounted for.

dobertRowneySr (approved):
LGTM

runtime-modules/referendum/src/weights.rs
mnaamani: Its a shame the comments are not very accurate, but notice the added db read weight on the following lines. Originally the placeholder benchmarks was added in first iteration of [implementation](https://github.com/Joystream/joystream/pull/4409) (didn't have a read), after the suggested fix, the weight generation wasn't run. So it showed up now when  I did the final generation.

Lezek123 (commented):
When comparing to master I see quite a few extrinsics that were modified in Ephesus, but don't have new weights generated:

Content module:
- `update_channel_payouts` - **missing**
- `claim_channel_reward` - **missing**

Membership:
- `member_remark` (with/without payment) - **OK**

Project token (perhaps it's out of Ephesus scope to update those weights though):
- `transfer` - **missing**
- `issuer_transfer` - **missing**

Proposals codex:
- all weights should be updated due to a change in `create_proposal` function - **missing** (except `create_proposal_update_channel_payouts`) 

Referendum:
- `vote` - **missing** (only the comment was updated it seems)
- `opt_out_of_voting` - **OK**

Issue #4601, opened 2023-02-06
Deleting channel with a collaborator crashed the QN's processor


zeeshanakram3:
## Context

As reported, deletion  of channel with a collaborator crashed the QN's processor, see logs: 

```
INFO: Last block: 1264132     : 38680 blocks behind
INFO: Starting the event queue
INFO: Processing block: 0001264154-865d6, events count: 4 
Asset has been disconnected from Channel { channelId: '17', dataObjectId: '143' }
Asset has been disconnected from Channel { channelId: '17', dataObjectId: '144' }
ERROR: Stopping the proccessor due to errors: {
  "query": "DELETE FROM \"channel\" WHERE \"id\" = $1",
  "parameters": [
    "17"
  ],
  "driverError": {
    "length": 309,
    "name": "error",
    "severity": "ERROR",
    "code": "23503",
    "detail": "Key (id)=(17) is still referenced from table \"collaborator\".",
    "schema": "public",
    "table": "collaborator",
    "constraint": "FK_e69d65ddb224176321e63ec3585",
    "file": "ri_triggers.c",
    "line": "2476",
    "routine": "ri_ReportViolation"
  },
  "length": 309,
  "severity": "ERROR",
  "code": "23503",
  "detail": "Key (id)=(17) is still referenced from table \"collaborator\".",
  "schema": "public",
  "table": "collaborator",
  "constraint": "FK_e69d65ddb224176321e63ec3585",
  "file": "ri_triggers.c",
  "line": "2476",
  "routine": "ri_ReportViolation"
} name: QueryFailedError, message: update or delete on table "channel" violates foreign key constraint "FK_e69d65ddb224176321e63ec3585" on table "collaborator", stack: QueryFailedError: update or delete on table "channel" violates foreign key constraint "FK_e69d65ddb224176321e63ec3585" on table "collaborator"
    at QueryFailedError.TypeORMError [as constructor] (/joystream/node_modules/typeorm/error/TypeORMError.js:9:28)
I 
```

## Fix

During processing the channel deletion mappings, we need to make sure that all the entities that have non-nullable reference (e.g. `Collaborator`) to channel are removed before we actually remove channel from the DB.

zeeshanakram3:
addressed in #4603

PR #4603, opened 2023-02-06
fix: channel deletion mapping bug

zeeshanakram3:
addresses #4601


query-node/mappings/src/content/channel.ts
```diff
@@ -412,3 +420,31 @@ async function processModeratorRemark(
...
+    return await store.getMany(entityType, {
+      where: { channel: { id: channelId } } as FindOptionsWhere<T>,
+    })
+  }
+
+  const removeRelations = async <T>(store: DatabaseManager, entities: T[]) => {
+    await Promise.all(entities.map(async (r) => await store.remove<T>(r)))
+  }
+
+  const referencingEntities: typeof BaseModel[] = [Collaborator, ChannelNftCollectors, MemberBannedFromChannelEvent]
```
Lezek123: If you declare it as:
```typescript
  const referencingEntities: { new (): BaseModel & { channel: Partial<Channel> } }[] = [
    Collaborator,
    ChannelNftCollectors,
    MemberBannedFromChannelEvent,
  ]
```
then you can remove `entity as any` assertion below and preserve the typesafety

Lezek123 (changes requested):
Additional notes:
1. I don't think we have any tests for banning members from a channel. If it's not time-consuming, could you add a call to `BanOrUnbanMemberFixture` in `flows/commentsAndReactions`, to make sure removing `MemberBannedFromChannelEvent` entities works as expected?

query-node/mappings/src/content/channel.ts
zeeshanakram3: Adressed in [550314a](https://github.com/Joystream/joystream/pull/4603/commits/550314a18c8fe98af2eb5b65e96815cda80bb394)

Lezek123 (approved):
LGTM, but will wait until the checks pass before merging.

A note about updating atlas-dev: Because the atlas-dev network is running from `carthage` branch and the mappings between `carthage` and `master` are not fully compatible (there was `Gateway` => `Apps` working group renaming), you probably just want to cherry-pick [463c141](https://github.com/Joystream/joystream/pull/4603/commits/463c141bdc9b69988b3916f943801e003cab3e6b) and [550314a](https://github.com/Joystream/joystream/pull/4603/commits/550314a18c8fe98af2eb5b65e96815cda80bb394) into `carthage` before re-building and restarting the processor.

PR #4560, opened 2023-01-22
expose QN status in Argus and Colossus

zeeshanakram3:
This PR addresses
- #4256 
- #4445

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203798699334216) by [Unito](https://www.unito.io)

PR #4607, opened 2023-02-07
bump package versions for ephesus

mnaamani:
Bump package versions to be higher than master branch.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203922112778457) by [Unito](https://www.unito.io)


dobertRowneySr (approved):
LGTM

zeeshanakram3 (commented):
1. Shouldn't we also bump `query-node` version? Here is the [list](https://github.com/Joystream/joystream/issues/4593) of all versions we selected for ephesus release
image


2. After https://github.com/Joystream/hydra/pull/516 has been merged, and new hydra version released, we should also update the hydra dependencies in the QN

mnaamani:
> 1. Shouldn't we also bump `query-node` version? Here is the [list](https://github.com/Joystream/joystream/issues/4593) of all versions we selected for ephesus release

Thank you I was looking for this to cross check. Will revise to match.

> After https://github.com/Joystream/hydra/pull/516 has been merged, and new hydra version released, we should also update the hydra dependencies in the QN

Noted

mnaamani:
There are a couple of PRs updating joystream-js package, so I assume that will need to be bumped also and published to npm eventually? @zeeshanakram3

zeeshanakram3:
> There are a couple of PRs updating joystream-js package, so I assume that will need to be bumped also and published to npm eventually? @zeeshanakram3

I talked to Theo about the PR that he created, and he said that merging it is not important as they have found workaround for it. The other one should be merged I believe as this will be required for Apps. So yes, newer version of joystream-js package should be published to npm

cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@joystream/cli",
   "description": "Command Line Interface for Joystream community and governance activities",
-  "version": "0.10.0",
+  "version": "0.11.0",
```
Lezek123: I'm wondering if we should bump versions of all packages that still start with `0.` to `1.` now that we're on mainnet,

tests/network-tests/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@apollo/client": "^3.2.5",
     "@joystream/distributor-node-client": "^0.2.0",
-    "@joystream/types": "^0.20.6",
+    "@joystream/types": "^0.20.7",
```
Lezek123: Same comment as w.r.t. `@joystream/cli`, it looks strange to not have the `patch` version at all and only update the `minor` one during a new release.

cli/package.json
```diff
@@ -11,8 +11,8 @@
     "@apidevtools/json-schema-ref-parser": "^9.0.6",
     "@apollo/client": "^3.3.13",
     "@ffprobe-installer/ffprobe": "^1.4.1",
-    "@joystream/metadata-protobuf": "^2.5.0",
-    "@joystream/types": "^0.20.6",
+    "@joystream/metadata-protobuf": "^2.7.0",
```
Lezek123: There will be apps release before ephesus and `2.6.0` is already taken, so I think `2.7.0` should be the apps and `2.8.0` - Ephesus?

query-node/mappings/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "query-node-mappings",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
Lezek123: Another question is whether we should bump the major version already for the apps release or for Ephesus? Otherwise we'd have to use something like `0.1.2` for the apps release.

mnaamani: Sounds good to me. With reference to bumping version on master https://github.com/Joystream/joystream/pull/4616#pullrequestreview-1295274065 lets make it 
`1.0.0` for master branch, 
`1.1.0` for apps release,
`1.2.0` for ephesus?

cli/package.json
mnaamani: That's a sensible suggestion.

tests/network-tests/package.json
mnaamani: So we go for `0.21.0` for ephesus, unless we want to also start at `1.0.0` for mainnet, so making ephesus version `1.1.0` (as there are no runtime changes for apps release)

cli/package.json
mnaamani: I have no objection.

mnaamani: 1.0.0 - mainnet
1.1.0 - apps
1.20 - ephesus

@Lezek123

tests/network-tests/package.json
Lezek123: I think we should start at `1.0.0` for mainnet for consistency

query-node/mappings/package.json
Lezek123: Sounds good! :+1:

cli/package.json
Lezek123: If you mean `1.2.0` for Ephesus then sounds good.

mnaamani:
Bumped versions to also take into consideration incoming PR to master https://github.com/Joystream/joystream/pull/4560

cli/package.json
mnaamani: yep, typo there :)

Lezek123 (changes requested):
1. CLI/network-tests should depend on `@joystream/js` of version at least `1.3.0` (currently `^1.0.0`)
2. W.r.t. `@joystream/types`, I think the combination of using `^1.1.0` and bumping only the minor version during an upgrade like Ephesus is risky. We should probably either bump the major version when there are non-backward-compatible changes (like the removal of sudo pallet, changes in existing event parameters etc.) or use `~1.1.0` to avoid some of our apps using non-compatible version of `@joystream/types` in the future.

mnaamani:
> 1. CLI/network-tests should depend on `@joystream/js` of version at least `1.3.0` (currently `^1.0.0`)
> 
>     2. W.r.t. `@joystream/types`, I think the combination of using `^1.1.0` and bumping only the minor version during an upgrade like Ephesus is risky. We should probably either bump the major version when there are non-backward-compatible changes (like the removal of sudo pallet, changes in existing event parameters etc.) or use `~1.1.0` to avoid some of our apps using non-compatible version of `@joystream/types` in the future.

1. Well spotted
2. That is a good point. Going with 2.0.0 sounds best.

mnaamani:
Updated as suggested. You will note I had to bump the dependency in migration-scripts as well, although that set of migration scripts should all be dropped, in addition to some in api-scripts as they depend on sudo. I will prepare a separate PR for that.

Lezek123 (approved):
LGTM

PR #4608, opened 2023-02-07
fix: re-enable storage in CI & fix setupNewChain test

### LINKED ISSUE #4609: Failing Playground Deploy CD action
The playground deployment CD action is failing here
https://github.com/Joystream/joystream/actions/runs/4116799098

I need to edit the `setupNewChain` & `setupNewChainMultiStorage` scenarios in order to have the council elected first, then try to start the playground locally

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203926687296484) by [Unito](https://www.unito.io)

dobertRowneySr:
Fixes #4609

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203935971261566) by [Unito](https://www.unito.io)


dobertRowneySr:
The playground deployment is failing locally due to council election timing out, I have updated the timeout limits and I am testing

dobertRowneySr:
> The playground deployment is failing locally due to council election timing out, I have updated the timeout limits and I am testing

This is was due to the `tests/network-tests/src/Api::untilCouncilStage` final check for which the `IdlePeriod (=1) >= reserve (=4)` (in the playground config). So the playground election used to get stuck.
I have set reserve = 1 and then ran the local playground successfully

dobertRowneySr:
I have also set shorter duration parameters for duration election and lead proposal

mnaamani:
Was able to do a deployment, and running a test now also through github action https://github.com/Joystream/joystream/actions/runs/4134320191

One useful addition I can think of is to perhaps give the council member member controller account a custom key, like we do for the working group workers to setup the storage infrastructure with pre-determined accounts. This will make it easy for testers of the playground to add these keys rather than needed to setup a new council with their own keys. As they will need to test specific proposals..

mnaamani:
> Was able to do a deployment, and running a test now also through github action https://github.com/Joystream/joystream/actions/runs/4134320191

The ansible job timed out, so prepared a PR to increase the timeout https://github.com/ignazio-bovo/joystream/pull/9

Successful deployment https://github.com/Joystream/joystream/actions/runs/4138130693/jobs/7154156247

mnaamani (changes requested):
Great job fixing this. Please update the timeout for ansible job  by merging https://github.com/ignazio-bovo/joystream/pull/9

If you can add the suggested deterministic account for council memebers it would be great finishing touch.

runtime/src/proposals_configuration/playground.rs
```diff
@@ -61,7 +61,7 @@ pub(crate) fn funding_request_proposal() -> ProposalParameters<BlockNumber, Bala
 pub(crate) fn create_working_group_lead_opening_proposal(
 ) -> ProposalParameters<BlockNumber, Balance> {
     ProposalParameters {
-        voting_period: 200,
+        voting_period: 20,
```
Lezek123: I don't think voting period length of the proposals adds to the delay, because the proposal should be approved instantly once the quorum+threshold is reached, so perhaps it's better to keep it intact.

Probably the main delay was the `council announcing period + referendum voting + referendum revealing period`, which are together `100 + 100 + 100 = 300 blocks (~1800 sec)` now, and were `300 + 100 + 100 = 500 blocks (~3000 sec)` before

```suggestion
        voting_period: 200,
```

runtime/src/proposals_configuration/playground.rs
```diff
@@ -76,7 +76,7 @@ pub(crate) fn create_working_group_lead_opening_proposal(
 pub(crate) fn fill_working_group_lead_opening_proposal() -> ProposalParameters<BlockNumber, Balance>
 {
     ProposalParameters {
-        voting_period: 200,
+        voting_period: 20,
```
Lezek123:
```suggestion
        voting_period: 200,
```

tests/network-tests/src/Api.ts
```diff
@@ -661,9 +661,10 @@ export class Api {
   public async untilCouncilStage(
     targetStage: 'Announcing' | 'Voting' | 'Revealing' | 'Idle',
     announcementPeriodNr: number | null = null,
-    blocksReserve = 4,
+    blocksReserve = 1, // TODO dynamically adjust this as it stuck the election process
```
Lezek123: What do you mean by "stuck the election process"?
I think there is a risk in having the reserve default to as low as `1`, as this may not be enough time for the integration tests to issue all transactions intended for a given stage (like announcing all candidacies during announcing stage) and we may experience some issues w/ the integration tests occasionally failing.

Lezek123 (changes requested):

tests/network-tests/src/Api.ts
dobertRowneySr: > This is was due to the tests/network-tests/src/Api::untilCouncilStage final check for which the IdlePeriod (=1) >= reserve (=4) (in the playground config). So the playground election used to get stuck.
I have set reserve = 1 and then ran the local playground successfully

I will to set Idle period to 5 for the playground

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -5,16 +5,32 @@ import { blake2AsHex } from '@polkadot/util-crypto'
...
+    const councilSize = this.api.consts.council.councilSize
+    const numberOfCandidates = this.candidatesAddresses.length
     const numberOfVoters = numberOfCandidates
 
-    // Prepare memberships
-    const candidatesMemberAccounts = (await this.api.createKeyPairs(numberOfCandidates)).map(({ key }) => key.address)
+    // get/create candidates member accounts
+    const candidatesMemberAccounts =
+      this.candidatesAddresses.length === 0
+        ? (await this.api.createKeyPairs(numberOfCandidates)).map(({ key }) => key.address)
```
Lezek123: If `this.candidatesAddresses.length` then `numberOfCandidates` is also `0`, I think you need to get `numberOfCandidates` from `api.consts` again.

tests/network-tests/src/flows/council/elect.ts
```diff
@@ -2,13 +2,25 @@ import { FlowProps } from '../../Flow'
...
   const debug = extendDebug('flow:elect-council')
+  const { api, query } = props
   debug('Started')
   api.enableDebugTxLogs()
 
-  const electCouncilFixture = new ElectCouncilFixture(api, query)
+  const councilSize = api.consts.council.councilSize.toNumber()
+
+  // if single councilor (playground setting) use custom key
+  const candidates = councilSize === 1 ? createSingleCouncilorAddress(props) : undefined
```
mnaamani: Here we are making assumption that the extra candidates required will always be 0.

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -5,17 +5,34 @@ import { blake2AsHex } from '@polkadot/util-crypto'
...
+    return addresses
+  }
+
   public async execute(): Promise<void> {
     const { api, query } = this
-    const { councilSize, minNumberOfExtraCandidates } = api.consts.council
-    const numberOfCandidates = councilSize.add(minNumberOfExtraCandidates).toNumber()
+
+    // get/create candidates member accounts
+    const candidatesMemberAccounts = this.candidatesAddresses ?? (await this.createCandidates())
```
mnaamani: Great to have the custom candidate addresses. Instead of having the special case for size of council = 1 only for which `this.candidatesAddresses` was introduced, we might as well just have all addresses generated for council candidates have custom addresses, created in `function createCandidates()` above, with pattern like `//CouncilCandidate//N` where you need to store the counter N and increment for each new candidate. Or to avoid saving this 'global' counter, you could have `//CouncilCandidate//R//N` - use the election round as `R` and `N` can just repeat 0..n for each election attempt.

mnaamani (changes requested):

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
dobertRowneySr: I have followed the second approach

Lezek123 (changes requested):
The test is failing because `createKeyPairs` also initializes the accounts with existential deposit by default, but `createCustomKeyPair` doesn't, so the `BuyMembership` fixture fails for candidate accounts now:
```
2023-02-14T07:56:52.088Z integration-tests:Sender:2:elect council:electCouncil-0 Dispatch Error: NotEnoughBalanceToBuyMembership
```

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -40,6 +64,8 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
 
     // Announcing stage
     await this.api.untilCouncilStage('Announcing')
+    const x = await this.api.query.council.announcementPeriodNr()
+    this.debug(`announcement period ${x.toNumber()}`)
```
mnaamani: Perhaps your intention here is to debug/check that we are still in the announcing period expected. Maybe assert the condition instead of logging and continuing?

tests/network-tests/src/flows/council/elect.ts
```diff
@@ -2,9 +2,11 @@ import { FlowProps } from '../../Flow'
 import { extendDebug } from '../../Debugger'
 import { FixtureRunner } from '../../Fixture'
 import { ElectCouncilFixture } from '../../fixtures/council/ElectCouncilFixture'
+import { assert } from 'chai'
 
-export default async function electCouncil({ api, query }: FlowProps): Promise<void> {
+export default async function electCouncil(props: FlowProps): Promise<void> {
   const debug = extendDebug('flow:elect-council')
+  const { api, query } = props
```
mnaamani: all changes in this file seem no longer necessary

tests/network-tests/.env
```diff
@@ -10,6 +10,8 @@ TREASURY_ACCOUNT_URI=//Bob
 START_KEY_ID=0
 # Mini-secret or mnemonic used in SURI for deterministic key derivation
 SURI_MINI_SECRET=""
+# Single councilor account SURI used for testing
+COUNCILLOR_SURI=//Councillor
```
mnaamani: this can be removed now.

devops/ansible/deploy-playground-playbook.yml
```diff
@@ -69,7 +69,7 @@
         TREASURY_ACCOUNT_URI: '{{ treasury_suri }}'
         INITIAL_BALANCES: '{{ initial_balances }}'
         SKIP_NODE: 'true'
-      async: 1800
+      async: 3600
```
mnaamani: Lets just add a little bit more time as we are just on the edge based on test runs I did.

```suggestion
      async: 4000
```

devops/ansible/deploy-playground-playbook.yml
```diff
@@ -79,7 +79,7 @@
       register: job_result
       until: job_result.finished
       # Max number of times to check for status
-      retries: 18
+      retries: 36
```
mnaamani:
```suggestion
      retries: 40
```

mnaamani (changes requested):
Looks good, just left a couple of code cleanup points, and suggestion to increase the timeout of the ansible job slightly, otherwise its good to go.

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
dobertRowneySr: I used it to display the announcement period, because the first election starts in Announcing stage while subsequent test elections starts in Idle, that way the announcing period between the first and the second election was the same and so the addresses for the candidates

dobertRowneySr: I have removed it now

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -40,6 +64,7 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
 
     // Announcing stage
     await this.api.untilCouncilStage('Announcing')
+    const x = await this.api.query.council.announcementPeriodNr()
```
Lezek123: Unused variable
```suggestion
```

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -7,14 +7,38 @@ import { assert } from 'chai'
...
+  // meanwhile subsequent elections tests start in the Idle period
+  protected async getUpcomingAnnouncingPeriod(): Promise<number> {
+    const stageUpdate = await this.api.query.council.stage()
+    const stage = stageUpdate.stage
+    const announcementPeriodId = await this.api.query.council.announcementPeriodNr()
+    if (stage.isAnnouncing) {
+      return announcementPeriodId.toNumber()
+    } else {
+      return announcementPeriodId.toNumber() + 1
+    }
```
Lezek123: This will not always work, because:
- It doesn't take into account the "block reserve" (ie. whether the announcing period has less than `blockReserve` blocks left)
- There are a few more actions ran after this check, for example, the `BuyMembershipHappyCaseFixture`. After those actions are finished, the stage may already be different (if the announcing period was close to an end).

One way to solve this would be to update the controller accounts of council members after they are already elected. I don't think you even need to include the election cycle number in this case.

Lezek123 (changes requested):

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
mnaamani: That might be best., that's how we do it for the working group workers -> https://github.com/Joystream/joystream/blob/997c5ae929b7422486593c3f0054154eef5f5cd2/tests/network-tests/src/Api.ts#L846

dobertRowneySr: There's no way to change a controller account for a member or councilor once set

Lezek123: You can do it via `update_accounts` extrinsic in the membership module

dobertRowneySr: 😃  good catch I have added the suggested fix + assertion

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -69,12 +80,13 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
 
   public async execute(): Promise<void> {
     const { api, query } = this
-    const { councilSize, minNumberOfExtraCandidates } = api.consts.council
+
+    // get/create candidates member accounts
```
Lezek123:
```suggestion
```

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -160,10 +172,20 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
...
     const councilMembers = await api.query.council.councilMembers()
+    this.councilMembersIds = councilMembers.map((m) => m.membershipId)
     assert.sameMembers(
-      councilMembers.map((m) => m.membershipId.toString()),
+      this.councilMembersIds.map((id) => id.toString()),
       candidatesToWinIds
     )
+
+    // change accounts to known accounts
+    const oldCouncilMemberAccounts = await this.getCouncilMembersControllerAccounts()
```
Lezek123: Instead of having to fetch those accounts here, you can query for the member's `rootAccount` in `updateMemberControllerAccount`, I think that will be simpler. (`update_accounts` extrinsic should be called using member's `rootAccount`, not `controllerAccount`, although in the integration tests they are usually the same)

tests/network-tests/src/fixtures/council/ElectCouncilFixture.ts
```diff
@@ -160,10 +172,20 @@ export class ElectCouncilFixture extends BaseQueryNodeFixture {
...
+    const newCouncilMemberAccounts = await this.api.updateCouncillorsAccounts(
+      oldCouncilMemberAccounts,
+      this.councilMembersIds
+    )
```
Lezek123: `updateCouncillorsAccounts` can query `api.query.council.councilMembers`, so that they don't need to be passed as an argument.

Lezek123 (changes requested):

tests/network-tests/src/Api.ts
```diff
@@ -859,6 +862,42 @@ export class Api {
...
+      memberIds.map(async (id) => {
+        const membership = await this.query.members.membershipById(id)
+        return membership.unwrap().rootAccount.toString()
+      })
+    )
+    // path to append to base SURI
+    const debug = extendDebug('api-factory')
+    debug(`assigning Well Known Councillors Account`)
+    const newAccounts = memberIds.map((id) => {
+      const uri = `//Councillor//` + id.toString()
```
Lezek123: I noticed there is a small issue here. Because the `uri` starts with `//` and `isFinalPath = false`, the final uri becomes:
```typescript
`${miniSecret}//testing////Councillor//${memberId}`
```
_(notice 4 slashes before `Councillor`)_

Lezek123 (changes requested):

Lezek123 (approved):
LGTM

mnaamani (approved):

PR #4621, opened 2023-02-09
Fix storage node syncing query

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4615 by splitting `getAllAssignedDataObjects` query into smaller queries, so that no individual query has more than 1000 storage bag ids provided in the input (to avoid hitting the 100kb request size limit)


mnaamani (approved):
Fix looks correct, serially making multiple queries is fine. Would doing multiple queries in parallel give a faster response?

zeeshanakram3 (commented):
Thanks for the fast fix!

1. I think we should add the `onError` error handling Apollo link while initializing the Query-node apollo client in storage node, similar to how it's implemented in the distributor-node's [QueryNodeApi](https://github.com/Lezek123/substrate-runtime-joystream/blob/c51db9ea3334a453c63b17ec3ba5e87f869b8f05/distributor-node/src/services/networking/query-node/api.ts#L48). This will imporve error handling/logging for graphql syntax & network errors
2. Should not be in the scope of this PR I think, but we should analyze all the queries having structure similar to [storageDataObjectsConnection](https://github.com/Lezek123/substrate-runtime-joystream/blob/c51db9ea3334a453c63b17ec3ba5e87f869b8f05/storage-node/src/services/queryNode/generated/queries.ts#L203) that are being used in all of the services/nodes, i.e. queries having one or more array params. And then if there is a chance that those queries could too break the default request size limit, we should split those queries  too into smaller ones

Lezek123:
> think we should add the onError error handling Apollo link while initializing the Query-node apollo client in storage node, similar to how it's implemented in the distributor-node's [QueryNodeApi](https://github.com/Lezek123/substrate-runtime-joystream/blob/c51db9ea3334a453c63b17ec3ba5e87f869b8f05/distributor-node/src/services/networking/query-node/api.ts#L48). This will imporve error handling/logging for graphql syntax & network errors

Added in https://github.com/Joystream/joystream/pull/4621/commits/fa7cfb33c4a9b4c7920980350f6835e44c2a3775

> Should not be in the scope of this PR I think, but we should analyze all the queries having structure similar to [storageDataObjectsConnection](https://github.com/Lezek123/substrate-runtime-joystream/blob/c51db9ea3334a453c63b17ec3ba5e87f869b8f05/storage-node/src/services/queryNode/generated/queries.ts#L203) that are being used in all of the services/nodes, i.e. queries having one or more array params. And then if there is a chance that those queries could too break the default request size limit, we should split those queries too into smaller ones

Created a new issue: https://github.com/Joystream/joystream/issues/4624

zeeshanakram3 (approved):
LGTM

Issue #4625, opened 2023-02-10
Missing mappings for `WorkingGroupBudgetFunded` event


Lezek123:
Query node mappings for `WorkingGroupBudgetFunded` aren't implemented, which may result in mismatch between runtime and QN state w.r.t. to current WG budget amount.

mochet:
Not sure if it helps but I've been able to find WG budget increases via the following query:
`query MyQuery {
  budgetUpdatedEvents(limit: 1000) {
    groupId
    budgetChangeAmount
    inBlock
    id
  }
}`

Lezek123:
Done in https://github.com/Joystream/joystream/pull/4632

PR #4626, opened 2023-02-10
Ephesus: post-merge fixes

Lezek123:
- In `run-network-tests.yml`: `actions/checkout@v3` and `actions/setup-node@v3` were (probably accidently) downgraded to `v1` during the merge, updated to `v3`.
- `project-token/src/lib.rs`: Fixed duplicated `TransferDestinationMemberDoesNotExist` check
- Removed no longer useful `SUDO_ACCOUNT_URI` env variable from integration tests
- `network-tests/src/consts.ts`: Removed `getWorkingGroupNameByModuleName` as it was pretty much a duplicate of `workingGroupNameByModuleName` const
- Network tests: Refactored flows that need to top up WG budget to use `fundWorkingGroupBudget` tx instead of proposal for simplicity
- Added missing `WorkingGroupBudgetFunded` event mapping (https://github.com/Joystream/joystream/issues/4625)
- `network-tests/src/cli/joystream.ts` - removed redundant `exitCode` checks for commands executed without `expectSuccess=false` flag. They were redundant, because in this case any `exitCode` other than `0` will just throw an error.
- Fixed resource locking in `network-tests/src/flows/proposals/index.ts`
- Removed a few unused fixtures that relied on sudo: `CreateLeadOpeningFixture`, `SetBudgetFixture`, `SetLeaderInvitationQuotaFixture`, `CreateMemberHappyCaseFixture`
- Removed unused flows that relied on sudo: `creatingFoundingMembers`, `creatingMembers`
- Removed a few other references to `sudo` and some unused code


dobertRowneySr (approved):
LGTM

mnaamani:
> * Added missing `WorkingGroupBudgetFunded` event mapping (

Do we need this on master as well?

Lezek123:
> Do we need this on master as well?

Oh, sure, I was planning to prepare a PR for master as well, I'll create one in a sec.

mnaamani (approved):
Nice cleanup, there are other scripts in `utils/api-scripts` with dependency on sudo which we can probably drop. Even the migration scripts for that matter. But not critical now.

PR #4631, opened 2023-02-14
Feature/app action

WRadoslaw:


query-node/mappings/src/content/channel.ts
```diff
@@ -77,8 +86,26 @@ export async function content_ChannelCreated(ctx: EventContext & StoreContext):
       inconsistentState(`storageBag for channel ${channelId} does not exist`)
     }
 
-    const metadata = deserializeMetadata(ChannelMetadata, channelCreationParameters.meta.unwrap()) || {}
-    await processChannelMetadata(ctx, channel, metadata, dataObjects)
+    const appAction = deserializeMetadata(AppAction, channelCreationParameters.meta.unwrap())
+
+    if (appAction && Object.keys(appAction).length) {
+      const appCommitment = generateAppActionCommitment(
+        channelOwner.ownerMember?.id ?? channelOwner.ownerCuratorGroup?.id ?? '',
```
Lezek123: `memberId` and `curatorGroupId` have the same format, so they can overlap.
I think we need a prefix to distinguish between the two (like `m:${membeId}`, `c:${curatorGroupId}`)

query-node/mappings/src/content/channel.ts
```diff
@@ -117,8 +144,14 @@ export async function content_ChannelUpdated(ctx: EventContext & StoreContext):
...
-    const newMetadata = deserializeMetadata(ChannelMetadata, newMetadataBytes) || {}
-    await processChannelMetadata(ctx, channel, newMetadata, newDataObjects)
+    const newMetadata = deserializeMetadata(AppAction, newMetadataBytes)
+
+    if (newMetadata && Object.keys(newMetadata)) {
+      await processChannelMetadata(ctx, channel, newMetadata.channelMetadata ?? {}, newDataObjects)
+    } else {
+      const realNewMetadata = deserializeMetadata(ChannelMetadata, newMetadataBytes)
+      await processChannelMetadata(ctx, channel, realNewMetadata ?? {}, newDataObjects)
+    }
```
Lezek123: No commitment / nonce checks done in this case and I don't think it makes sense to support `AppAction` for channel/video updates currently (?)

Was this mentioned in some spec?

query-node/mappings/src/content/utils.ts
```diff
@@ -166,6 +173,89 @@ async function processVideoSubtitleAssets(
...
+  actionOwnerId: string | undefined,
+  appAction: DecodedMetadataObject<IAppAction>
+): Promise<boolean> {
+  if (!appAction.metadata?.nonce || !actionOwnerId) {
+    return false
+  }
+
+  if (appAction.contentMetadata?.videoMetadata) {
+    return (
+      (await ctx.store.getMany(Video, { where: { channel: { ownerMember: { id: actionOwnerId } } } })).length ===
```
Lezek123: One can still easily reuse a nonce by removing some arbitrary videos, so that `(await ctx.store.getMany(Video, { where: { channel: { ownerMember: { id: actionOwnerId } } } })).length` matches the nonce.

I think what is needed is a new field in `Membership` entity which represents number of videos/channels ever created by the member, which is only incremented on each new video/channel and never decremented.

query-node/mappings/src/content/utils.ts
```diff
@@ -166,6 +173,89 @@ async function processVideoSubtitleAssets(
...
+      parseInt(appAction.metadata.nonce)
+    )
+  }
+
+  if (appAction.channelMetadata) {
+    return (
+      (await ctx.store.getMany(Channel, { where: { ownerMember: { id: actionOwnerId } } })).length ===
```
Lezek123: Same as above

query-node/mappings/src/content/video.ts
```diff
@@ -184,9 +217,18 @@ export async function content_ContentUpdated(ctx: EventContext & StoreContext):
...
-
-    await processUpdateVideoMessage(ctx, video, contentMetadata?.videoMetadata || undefined, contentUpdatedEventData)
+    const appAction = newMeta.isSome ? deserializeMetadata(AppAction, newMeta.unwrap()) : undefined
+    if (appAction && 'signature' in appAction) {
+      await processUpdateVideoMessage(
+        ctx,
+        video,
+        appAction?.contentMetadata?.videoMetadata || undefined,
+        contentUpdatedEventData
+      )
```
Lezek123: Same comment as w.r.t. channel updates

tests/network-tests/package.json
```diff
@@ -19,6 +19,7 @@
     "@joystream/types": "^0.20.5",
     "@polkadot/api": "8.9.1",
     "@polkadot/keyring": "9.5.1",
+    "@polkadot/util-crypto": "^10.2.6",
```
Lezek123:
```suggestion
```

`@polkadot/util-crypto` is already a dependency of `@polkadot/api`. We also don't want to introduce version conflicts (although this is currently fixed via `resolutions` in the monorepo)

tests/network-tests/src/Api.ts
```diff
@@ -761,22 +763,30 @@ export class Api {
...
     memberControllerAccount = memberControllerAccount || (await this.getMemberControllerAccount(memberId))
 
     if (!memberControllerAccount) {
       throw new Error('invalid member id')
     }
 
+    const txParameters = channelMeta
+      ? {
+          expectedDataObjectStateBloatBond: await this.api.query.storage.dataObjectPerMegabyteFee(),
+          expectedChannelStateBloatBond: await this.api.query.content.channelStateBloatBondValue(),
```
Lezek123: Those are not needed when `assets` are empty

tests/network-tests/src/Api.ts
```diff
@@ -787,17 +797,30 @@ export class Api {
...
 
     if (!memberControllerAccount) {
       throw new Error('invalid member id')
     }
 
+    const txParameters = videoMeta
+      ? {
+          storageBucketsNumWitness: await this.storageBucketsNumWitness(channelId),
+          expectedVideoStateBloatBond: await this.getVideoStateBloatBond(),
+          expectedDataObjectStateBloatBond: await this.getDataObjectStateBloatBond(),
```
Lezek123: Those are not needed when `assets` are empty

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -0,0 +1,147 @@
...
+  const signature = u8aToHex(ed25519Sign(appChannelCommitment, keypair, true))
+  const appChannelInput = {
+    appId: appFragment?.[0].id,
+    channelMetadata: channelInput,
+    signature,
+    metadata: appActionMeta,
+  }
+  const storageBuckets = await createChannelFixture.selectStorageBucketsForNewChannel()
+  const distBuckets = await createChannelFixture.selectDistributionBucketsForNewChannel()
+  const channelId = await api.createMockChannel(
```
Lezek123: You're instantializing `createChannelFixture` which you never actually run, you only use it to select the buckets. This is a little confusing. I think you should extend `CreateChannelsAsMemberFixture` to take optional `appActionMeta` argument instead, possibly even add the query node check there as well (implement `runQueryNodeChecks` from `BaseQueryNodeFixture`)

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -0,0 +1,147 @@
...
+  await query.tryQueryWithTimeout(
+    () => query.channelById(channelId.toString()),
+    (channel) => {
+      Utils.assert(channel, 'Channel not found')
+      assert.equal(channel.title, appChannelInput.channelMetadata?.title)
+      assert.equal(channel.description, appChannelInput.channelMetadata?.description)
+      assert.equal(channel.isPublic, appChannelInput.channelMetadata?.isPublic)
+      assert.equal(channel.language?.iso, appChannelInput.channelMetadata?.language)
+      assert.equal(channel.entryApp?.id, appFragment?.[0].id)
+      assert.equal(channel.entryApp?.name, appFragment?.[0].name)
```
Lezek123: This is a minor issue, but I suggest assigning `appFragment?.[0].id` to a non-nullable `appId` variable instead to avoid having to do optional chaining every time (for app name you can already use `appOwnedByMember`). It would also make the code a little more clear.

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -0,0 +1,147 @@
...
+  )
+
+  await query.tryQueryWithTimeout(
+    () => query.channelById(channelId.toString()),
+    (channel) => {
+      Utils.assert(channel, 'Channel not found')
+      assert.equal(channel.title, appChannelInput.channelMetadata?.title)
+      assert.equal(channel.description, appChannelInput.channelMetadata?.description)
+      assert.equal(channel.isPublic, appChannelInput.channelMetadata?.isPublic)
+      assert.equal(channel.language?.iso, appChannelInput.channelMetadata?.language)
```
Lezek123:
```suggestion
      assert.equal(channel.title, channelInput.title)
      assert.equal(channel.description, channelInput.description)
      assert.equal(channel.isPublic, channelInput.isPublic)
      assert.equal(channel.language?.iso, channelInput.language)
```

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+  const videoId = await api.createMockVideo(member.memberId.toNumber(), channelId.toNumber(), undefined, appVideoInput)
+
+  await query.tryQueryWithTimeout(
+    () => query.videoById(videoId.toString()),
+    (video) => {
+      Utils.assert(video, 'Video not found')
+      assert.equal(video.title, appVideoInput.contentMetadata?.videoMetadata?.title)
+      assert.equal(video.description, appVideoInput.contentMetadata?.videoMetadata?.description)
+      assert.equal(video.duration, appVideoInput.contentMetadata?.videoMetadata?.duration)
```
Lezek123:
```suggestion
      assert.equal(video.title, contentMetadata.videoMetadata.title)
      assert.equal(video.description, contentMetadata.videoMetadata.description)
      assert.equal(video.duration, contentMetadata.videoMetadata.duration)
```

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -0,0 +1,147 @@
...
+      assert.equal(video.entryApp?.id, appFragment?.[0].id)
+      assert.equal(video.entryApp?.name, appFragment?.[0].name)
+      assert.equal(video.ytVideoId, appVideoInput.metadata?.videoId)
```
Lezek123:
```suggestion
      assert.equal(video.ytVideoId, videoAppActionMeta.videoId)
```

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto2";
+
+import "proto/Video.proto";
+import "proto/Channel.proto";
+
+message AppActionMetadata {
+  optional string nonce = 1;
+  // YouTube video ID
+  optional string video_id  = 2;
+}
```
Lezek123: Based on how I understand the description in https://github.com/Joystream/joystream/issues/4307, more specificly:
```
metadata: raw metadata, can be anything.
```
And what I later raised in https://github.com/Joystream/joystream/issues/4307#issuecomment-1430043806, I think the second field of the `AppActionMetadata` should be just a raw string, that can be interpreted differently depending on the app that issued the action.

So the way I understand this, it would be something like:
```suggestion
message AppActionMetadata {
  optional string nonce = 1;
  optional string raw_custom_metadata  = 2;
}
```

The query node I think doens't even need to try to process `raw_custom_metadata`, it can just include it in the `Video` /  `Channel` entity "as-is" and let the apps process it on their side (for example, YPP can understand that the `video.appActionMetadata` is just a youtube video id, provided that `video.entryApp` is YPP).

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
...
+  optional string app_id = 1;
+
+  // Metadata
+  optional AppActionMetadata metadata = 2;
+
+  // Raw metadata of wrapped action
+  oneof raw_action {
+    ContentMetadata content_metadata  = 3;
+    ChannelMetadata channel_metadata = 4;
+  }
```
Lezek123: Similarly to the previous comment, https://github.com/Joystream/joystream/issues/4307 states:
> The reason raw_action is not locked to a specific set of messages (...)

In you implementation it is actually being locked to a specific set of messages: `ContentMetadata` and `ChannelMetadata`.

As I understand this should be just a raw string, the QN can try to decode it as `ContentMetadata` or `ChannelMetadata` depending on which event it processes, but a failure to decode `raw_action` shouldn't cause a failure to decode the entire `AppAction`. In case `raw_action` is not in a recognized format, it should be just treated as empty metadata, but the channel/video should still be connected to a given app.

Lezek123 (changes requested):
I'm not sure if my understanding of the spec at https://github.com/Joystream/joystream/issues/4307 is correct, but I added a comment where we can further discuss it: https://github.com/Joystream/joystream/issues/4307#issuecomment-1430043806 and reviewed accordingly to my current assumptions.

query-node/mappings/src/content/channel.ts
WRadoslaw: I remember that updates weren't in the initial spec

query-node/mappings/src/content/utils.ts
WRadoslaw: I wonder how will it imply for curator group ownership. If we add the fields to the member, then we shouldn't validate curator group none at all?

Maybe instead of fetching video/channel as it can be deleted we would fetch `..._CreatedEvent` and just validate against its length.

Lezek123: 1. The reason I didn't mention curator groups is that I noticed they are not fully supported currently, for example, in
https://github.com/Joystream/joystream/pull/4631/files#diff-526bb3b7420aeed56e6c5bfa94e07ce3a8302d6ed7ad874b394b2185d542a003R102 you are only passing `ownerMember?.id` and in https://github.com/Joystream/joystream/pull/4631/files#diff-1b8138e018fcb5011c92af46bc3ed5a7a04579644c72675900d69dbdfe6b5fa3R182 you make the nonce check fail in case there is no owner id provided, which makes it impossible to pass this check in case of curator-group-owned channels anyway. If you wish to add support for curator-group-owned channels you need a similar counter in `CuratorGroup` entity perhaps.

2. Counting `ChannelCreated` / `VideoCreated` events would be another way to go about it, but we currently don't store those in the processor database, so I think that's a more complex change.

tests/network-tests/src/Api.ts
WRadoslaw: Tried to remove these changes, but tests are failing due to channel TX failure (bloat bond error)

tests/network-tests/src/Api.ts
WRadoslaw: Same as above

tests/network-tests/src/Api.ts
Lezek123: My bad, I didn't realize this function was not actually used in a while.
It seems like only `expectedDataObjectStateBloatBond` is optional when there are no assets, but `expectedChannelStateBloatBond` is **always** required (regardless of whether there is `channelMeta` set or not)

tests/network-tests/src/Api.ts
Lezek123: Here `expectedDataObjectStateBloatBond` is optional, but `storageBucketsNumWitness` and `expectedVideoStateBloatBond` is required regardless of whether `videoMeta` is set or not

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -33,62 +34,59 @@ export async function createAppActions({ api, query }: FlowProps): Promise<void>
...
   const appFragment = await query.tryQueryWithTimeout(
-    () => query.getAppsByName(appOwnedByMember),
+    () => query.getAppsByName(appOwnedByMemberName),
     (appsByName) => {
-      assert.equal(appsByName?.[0]?.name, appOwnedByMember)
+      assert.equal(appsByName?.[0]?.name, appOwnedByMemberName)
       assert.equal(appsByName?.[0].authKey, appPublicKeyHex)
     }
   )
+  const appId = appFragment?.[0]?.id as string
```
Lezek123, line 47:
```suggestion
  const appId = appFragment?.[0]?.id
  Utils.assert(appId)
```

tests/network-tests/src/scenarios/app.ts
```diff
@@ -11,8 +11,8 @@ scenario('App', async ({ job }) => {
   job('Initialize Faucet', initFaucet)
 
   const leads = job('Set WorkingGroup Leads', leaderSetup(true))
-  job('Create app', createApp).after(leads)
-  job('Update app', updateApp).after(leads)
+  // job('Create app', createApp).after(leads)
+  // job('Update app', updateApp).after(leads)
   job('Create app actions', createAppActions).after(leads)
-  job('Delete app', deleteApp).after(leads)
+  // job('Delete app', deleteApp).after(leads)
```
Lezek123:
```suggestion
  job('Create app', createApp).after(leads)
  job('Update app', updateApp).after(leads)
  job('Create app actions', createAppActions).after(leads)
  job('Delete app', deleteApp).after(leads)
```

tests/network-tests/src/Api.ts
```diff
@@ -761,22 +763,30 @@ export class Api {
...
+    channelMeta?: Bytes
   ): Promise<ChannelId> {
     memberControllerAccount = memberControllerAccount || (await this.getMemberControllerAccount(memberId))
 
     if (!memberControllerAccount) {
       throw new Error('invalid member id')
     }
 
+    const txParameters = channelMeta
+      ? {
```
Lezek123: Please make those parameters independent of whether `channelMeta` is set or not.

tests/network-tests/src/Api.ts
Lezek123: Please make those parameters independent of whether `videoMeta` is set or not.

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
...
+  optional string app_id = 1;
+
+  // Metadata in hex
+  optional string metadata = 2;
+
+  // Raw metadata of wrapped action in hex
+  optional string raw_action = 3;
+
+  // Signature over app commitment
+  optional string signature = 4;
```
Lezek123: I just realized that having those parameters provided as hex is quite suboptimal. You need 2 hex characters to represent one byte, so there are at least 2 bytes used to represent a single byte of data at the moment.

There is a `bytes` type in protobuf, perhaps you can give it a try? `protobufjs` represents it as `UInt8Array` in JavaScript

query-node/mappings/src/content/utils.ts
```diff
@@ -166,6 +173,65 @@ async function processVideoSubtitleAssets(
   }
 }
 
+async function validateAndGetApp<T>(
+  ctx: EventContext & StoreContext,
+  entity: T,
```
Lezek123: Unused parameter

query-node/mappings/src/content/channel.ts
```diff
@@ -77,13 +85,37 @@ export async function content_ChannelCreated(ctx: EventContext & StoreContext):
       inconsistentState(`storageBag for channel ${channelId} does not exist`)
     }
 
-    const metadata = deserializeMetadata(ChannelMetadata, channelCreationParameters.meta.unwrap()) || {}
-    await processChannelMetadata(ctx, channel, metadata, dataObjects)
+    const appAction = deserializeMetadata(AppAction, channelCreationParameters.meta.unwrap())
+
+    if (appAction && Object.keys(appAction).length) {
```
Lezek123: I noticed there is a backward-compatibility problem both here and in the `CreateVideo` event handler.
The problem is that in protobuf you cannot distinguish between different message types (like `AppAction` vs `ChannelMetadata`) just by looking at the serialized message.

For example, a message like:
```
ChannelMetadata {
  title: "Test",
  description: "Test"
}
```
Will be exactly the same as:
```
AppAction {
  appId: "Test",
  metadata: "Test"
}
```
once serialized. Both of them will be serialized to `0x0a0454657374120454657374`.

You can only distringuish a serialized `AppAction` from `ChannelMetadata` if a property which has an index that exists only in one of the messages (for example, property `6` only exists in `ChannelMetadata`) is specified, or if a property which has different type depending on the message (for example, `3` is `bool` in `ChannelMetadata`, but `string` in `AppAction`) is encoded in such a way that you can derive the type from the encoding.

So what we need is a way to be able to definitely tell whether a serialized message was supposed to be `AppAction` or `ChannelMetadata` / `ContentMetadata`. You can require a prefix for `AppAction` messages for example, or make `app_id` a `required` property with some unique index, for example:
```
    message AppAction {
      required string app_id = 9999;
      ...
    }
```

Lezek123 (changes requested):

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
...
+  required string app_id = 999;
+
+  // Metadata
+  optional bytes metadata = 2;
+
+  // Raw metadata of wrapped action
+  optional bytes raw_action = 3;
+
+  // Signature over app commitment
+  optional string signature = 4;
```
Lezek123: Could you make this `bytes` as well?

query-node/mappings/src/content/video.ts
```diff
@@ -93,19 +98,25 @@ export async function content_ContentCreated(ctx: EventContext & StoreContext):
   }
 
   // deserialize & process metadata
+  const appAction = meta.isSome ? deserializeMetadata(AppAction, meta.unwrap()) : undefined
   const contentMetadata = meta.isSome ? deserializeMetadata(ContentMetadata, meta.unwrap()) : undefined
```
Lezek123: This will generate a lot of warnings, because only one of those calls to `deserializeMetadata` will be successful.
Could you add a flag like `skipWarning` to `deserializeMetadata` and skip calling `invalidMetadata` in case it's provided?

Then please also update the flow to be as follows:
1. Try to deserialize as `AppAction` with `skipWarning=true`
2. If it's `AppAction` (deserialization doesn't fail), don't try to deserialize as `ContentMetadata` anymore
3. If `AppAction` deserialization fails - try to deserialize as `ContentMetadata`

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto2";
+
+import "proto/Video.proto";
+import "proto/Channel.proto";
```
Lezek123: No longer needed
```suggestion
```

metadata-protobuf/proto/App.proto
```diff
@@ -0,0 +1,26 @@
...
+  optional bytes metadata = 2;
+
+  // Raw metadata of wrapped action
+  optional bytes raw_action = 3;
+
+  // Signature over app commitment
+  optional string signature = 4;
+
+  // Nonce to prevent signature reusal
+  optional string nonce = 5;
```
Lezek123: Not a huge issue, but I think this should be `u32` or `u64`

query-node/mappings/src/content/channel.ts
```diff
@@ -77,13 +85,37 @@ export async function content_ChannelCreated(ctx: EventContext & StoreContext):
       inconsistentState(`storageBag for channel ${channelId} does not exist`)
     }
 
-    const metadata = deserializeMetadata(ChannelMetadata, channelCreationParameters.meta.unwrap()) || {}
-    await processChannelMetadata(ctx, channel, metadata, dataObjects)
+    const appAction = deserializeMetadata(AppAction, channelCreationParameters.meta.unwrap())
```
Lezek123: Please add `skipWarning` here as well

query-node/mappings/src/content/channel.ts
Lezek123: This is now equivalent to:
```suggestion
    if (appAction) {
```

query-node/mappings/src/content/channel.ts
```diff
@@ -117,8 +149,16 @@ export async function content_ChannelUpdated(ctx: EventContext & StoreContext):
       inconsistentState(`storageBag for channel ${channelId} does not exist`)
     }
 
-    const newMetadata = deserializeMetadata(ChannelMetadata, newMetadataBytes) || {}
-    await processChannelMetadata(ctx, channel, newMetadata, newDataObjects)
+    const newMetadata = deserializeMetadata(AppAction, newMetadataBytes)
```
Lezek123: Please add `skipWarning` here as well

query-node/mappings/src/content/channel.ts
```diff
@@ -117,8 +149,16 @@ export async function content_ChannelUpdated(ctx: EventContext & StoreContext):
...
+
+    if (newMetadata && 'rawAction' in newMetadata) {
```
Lezek123:
```suggestion
    if (newMetadata) {
```

query-node/mappings/src/content/utils.ts
```diff
@@ -41,12 +44,16 @@ import {
   PalletContentPermissionsContentActor as ContentActor,
   PalletContentIterableEnumsChannelActionPermission,
 } from '@polkadot/types/lookup'
-import { DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
+import { AnyMetadataClass, DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
```
Lezek123: Unused import
```suggestion
import { DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
```

query-node/mappings/src/content/utils.ts
```diff
@@ -166,6 +173,64 @@ async function processVideoSubtitleAssets(
   }
 }
 
+async function validateAndGetApp<T>(
```
Lezek123: Unused generic:
```suggestion
async function validateAndGetApp(
```

query-node/mappings/src/content/utils.ts
```diff
@@ -166,6 +173,64 @@ async function processVideoSubtitleAssets(
...
+    typeof validationContext.ownerNonce === 'undefined' ||
+    validationContext.ownerNonce !== parseInt(appAction.nonce)
+  ) {
+    invalidMetadata('Invalid app action nonce')
+
+    return undefined
+  }
+
+  try {
+    return ed25519Verify(validationContext.appCommitment, appAction.signature, app.authKey) ? app : undefined
```
Lezek123: There are logs like: `invalidMetadata('Invalid app action nonce')`, but there is no log for `Invalid signature` case, which would probably be quite useful for debugging

query-node/mappings/src/content/video.ts
```diff
@@ -123,12 +134,42 @@ export async function processCreateVideoMessage(
     reactionsCount: 0,
   })
 
-  if (metadata) {
-    await processVideoMetadata(ctx, video, metadata, newDataObjectIds)
+  if (metadata && 'rawAction' in metadata) {
```
Lezek123: The only bulletproof way to differentiate between the 2 types is now the `appId` (`rawAction` is optional in `AppAction`):
```suggestion
  if (metadata && 'appId' in metadata) {
```

query-node/mappings/src/content/video.ts
```diff
@@ -184,9 +225,15 @@ export async function content_ContentUpdated(ctx: EventContext & StoreContext):
   })
 
   if (video) {
-    const contentMetadata = newMeta.isSome ? deserializeMetadata(ContentMetadata, newMeta.unwrap()) : undefined
-
-    await processUpdateVideoMessage(ctx, video, contentMetadata?.videoMetadata || undefined, contentUpdatedEventData)
+    const appAction = newMeta.isSome ? deserializeMetadata(AppAction, newMeta.unwrap()) : undefined
+    if (appAction && 'signature' in appAction) {
```
Lezek123:
```suggestion
    if (appAction) {
```
If the signature is not present is should be handled later, with a clear warning message. The fact that we could deserialize the message as `AppAction` already means it was supposed to be deserialized this way (due to a new required `appId = 999` prop)

query-node/mappings/src/content/video.ts
```diff
@@ -184,9 +225,15 @@ export async function content_ContentUpdated(ctx: EventContext & StoreContext):
   })
 
   if (video) {
-    const contentMetadata = newMeta.isSome ? deserializeMetadata(ContentMetadata, newMeta.unwrap()) : undefined
-
-    await processUpdateVideoMessage(ctx, video, contentMetadata?.videoMetadata || undefined, contentUpdatedEventData)
+    const appAction = newMeta.isSome ? deserializeMetadata(AppAction, newMeta.unwrap()) : undefined
```
Lezek123: Please add `skipWarning` here as well

query-node/mappings/src/content/video.ts
```diff
@@ -184,9 +225,15 @@ export async function content_ContentUpdated(ctx: EventContext & StoreContext):
...
+    if (appAction && 'signature' in appAction) {
+      const contentMetadataBytes = u8aToBytes(appAction?.rawAction)
```
Lezek123:
```suggestion
      const contentMetadataBytes = u8aToBytes(appAction.rawAction)
```

Lezek123 (changes requested):

Lezek123 (approved):
LGTM

PR #4633, opened 2023-02-14
Fix network-deploy action 

mnaamani:
Only install python3-netaddr as ubuntu-latest doesn't have python2 anymore and would cause deployment failure.
ref: https://github.com/Joystream/joystream/actions/runs/4172483517/jobs/7223639256#step:5:13


bwhm (approved):
:+1:

PR #4642, opened 2023-02-16
Increase storageSync QN timeout even more (5m)

Lezek123:
Storage sync failing again: https://github.com/Joystream/joystream/actions/runs/4196136901/jobs/7278686492

- Increased total timeout to 5 minutes
- Fixes arguments to `tryQueryWithTimeout`: number of seconds was passed instead of number of miliseconds 
- Removed the arbitrary `Giving the query node X minutes to sync...` period in favor of using more retries in `tryQueryWithTimeout`

### DIFF
--- a/tests/network-tests/src/flows/storage/storageSync.ts
+++ b/tests/network-tests/src/flows/storage/storageSync.ts
@@ -62,15 +62,13 @@ export async function storageSync({ api, query }: FlowProps): Promise<void> {
     await api.sendExtrinsicsAndGetResults(createChannelTxBatch, memberAddr)
   }
 
-  debug('Giving the query node 2 minutes to sync...')
-  await Utils.wait(120_000)
-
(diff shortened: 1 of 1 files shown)


bedeho:
Is this just a testing artefact, or is there a fundamental issue here?

Lezek123:
> Is this just a testing artefact, or is there a fundamental issue here?

I just realized that the main reason the check was failing was that I was passing number of seconds instead of number of miliseconds to `tryQueryWithTimeout` function as `retryTime`.

The main issue I was trying to solve was that it takes over 2 minutes to process `10_000` new channels in the query node (during CI checks), which is quite slow, but to be fair this is only experienced in CI, because the machines used to run the checks are probably not very high spec.

Locally it takes less than a minute on my machine, so I wouldn't say it's a big isssue unless we experience a huge spike in demand.

mnaamani (approved):

Issue #4643, opened 2023-02-17
Add initial invitation balance to `MemberInvited` event


Lezek123:
Add initial invitation balance to `MemberInvited` event to avoid having to hardcode it in the query node: https://github.com/Joystream/joystream/blob/master/query-node/mappings/src/membership.ts#L74

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203993382260318) by [Unito](https://www.unito.io)

bedeho:
Done in #4646

PR #4646, opened 2023-02-17
Add initial invitation balance to `MemberInvited` event, remove workaround, bring back tests

Lezek123:
Addresses https://github.com/Joystream/joystream/issues/4643
- **Runtime:** add initial invitation balance to `MemberInvited` event
- **Query node:** remove hardcoded initial invitation balance and add the new property to `MemberInvitedEvent`
- **Integration tests:** Bring back the tests originally introduced in https://github.com/Joystream/joystream/pull/4533 (with adjustments due to sudo module removal)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1204004850546542) by [Unito](https://www.unito.io)


zeeshanakram3 (approved):
LGTM.

mnaamani (changes requested):
Looks good, I like that the tests change the invitation balance is updated twice between invites.
Only change request would be to bump types package version to `v2.1.0` since we already published `2.0.0` to npm.

Lezek123:
> Only change request would be to bump types package version to v2.1.0 since we already published 2.0.0 to npm.

https://github.com/Joystream/joystream/pull/4646/commits/f899922645013cd5aad6d8036aeae2180f123cf7

I bumped types to `2.1.0`, but since only QN depends on this change I only bumped the dependency there.
Other packages depend on `^2.0.0`, so will use the monorepo version as well.
The only exception was `@joystream/js`, so I changed the types dependency from `~2.0.0` to `^2.0.0` there.

mnaamani (approved):

mnaamani:
> The only exception was `@joystream/js`, so I changed the types dependency from `~2.0.0` to `^2.0.0` there.

So just to confirm, there is no need to publish a new version of `@joystream/js` as it doesn't care about the MemberInvited event changes?

Lezek123:
> So just to confirm, there is no need to publish a new version of @joystream/js as it doesn't care about the MemberInvited event changes?

Yes, it doesn't depend on those changes. The only inconvinience would be if someone tries to use `@joystream/js` together with `@joystream/types@2.1.0`, as then they would end up with either 2 versions of types (`2.0.0` and `2.1.0` since `@joystream/js` depended on `~2.0.0` the last time it was released), or they would have to use `resolutions`/`overrides`.

mnaamani:
FYI @thesan https://www.npmjs.com/package/@joystream/types/v/2.1.0

PR #4648, opened 2023-02-20
Remove lead-owned apps

Lezek123:
https://discord.com/channels/811216481340751934/1035123033196732427/1075886121080336515

### DIFF
--- a/query-node/mappings/src/content/app.ts
+++ b/query-node/mappings/src/content/app.ts
@@ -9,7 +9,7 @@ export async function processCreateAppMessage(
   store: DatabaseManager,
   event: SubstrateEvent,
   metadata: DecodedMetadataObject<ICreateApp>,
-  memberId?: string
+  memberId: string
 ): Promise<void> {
   const { name, appMetadata } = metadata
 
@@ -33,8 +33,7 @@ export async function processCreateAppMessage(
   const newApp = new App({
     name,
     id: appId,
-    ownerMember: memberId ? new Membership({ id: memberId }) : undefined,
-    isLeadOwned: !memberId,
+    ownerMember: new Membership({ id: memberId }),
     websiteUrl: appMetadata?.websiteUrl || undefined,
     useUri: appMetadata?.useUri || undefined,
     smallIcon: appMetadata?.smallIcon || undefined,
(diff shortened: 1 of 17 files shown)


query-node/mappings/src/content/app.ts
```diff
@@ -33,8 +33,7 @@ export async function processCreateAppMessage(
   const newApp = new App({
     name,
     id: appId,
-    ownerMember: memberId ? new Membership({ id: memberId }) : undefined,
```
drillprop, line 36: I can't comment on line 16 on GitHub, but there is a redundant condition and a comment. 
```tsx

  // memberId is required for MemberRemark, but not for LeadRemark
  if (event.method === 'MemberRemarked' && !memberId) {
    inconsistentState("memberId wasn't provided")
  }
```
We should update it as well

query-node/mappings/src/content/app.ts
```diff
@@ -55,29 +54,20 @@ export async function processUpdateAppMessage(
   store: DatabaseManager,
   event: SubstrateEvent,
```
drillprop: unused parameter

query-node/mappings/src/content/app.ts
```diff
@@ -102,29 +92,20 @@ export async function processDeleteAppMessage(
...
```
drillprop: unused parameter

tests/network-tests/src/flows/content/createApp.ts
```diff
@@ -30,28 +30,7 @@ export async function createApp({ api, query }: FlowProps): Promise<void> {
...
-  await query.tryQueryWithTimeout(
-    () => query.getAppsByName(appOwnedByLead),
-    (appsByName) => {
-      assert.equal(appsByName?.[0]?.name, appOwnedByLead)
-      assert.equal(appsByName?.[0].ownerMember?.id, null)
-      assert.equal(appsByName?.[0].isLeadOwned, true)
+      assert.equal(appsByName?.[0].ownerMember.id, member.memberId.toString())
       assert.equal(appsByName?.[0]?.category, appMetadata.category)
       assert.equal(appsByName?.[0]?.oneLiner, appMetadata.oneLiner)
       assert.equal(appsByName?.[0]?.description, appMetadata.description)
```
drillprop, line 35: Maybe I'm nitpicking here, but we could rename all the variables like appOwnedByMember to just `app` or `newApp`. Just for the sake of clarity? WDYT? We don't need to differentiate anymore between the app owned by the lead and the app owned by members.

tests/network-tests/src/flows/content/deleteApp.ts
```diff
@@ -43,35 +43,5 @@ export async function deleteApp({ api, query }: FlowProps): Promise<void> {
     )
   }
```
drillprop: same question here

tests/network-tests/src/flows/content/updateApp.ts
```diff
@@ -61,51 +61,5 @@ export async function updateApp({ api, query }: FlowProps): Promise<void> {
     }
```
drillprop: same question here

drillprop (changes requested):

drillprop (approved):
Thanks for taking care of it! LGTM!

PR #4656, opened 2023-02-22
Feature/app cli

WRadoslaw:


Lezek123:
Could you re-target the PR to `apps-metaprotocol` branch?

cli/src/base/AppCommandBase.ts
```diff
@@ -0,0 +1,54 @@
...
+    authKey: 'App public auth key?',
+    websiteUrl: 'Site URL where user can learn more about the app?',
+    useUri: 'App URL?',
+    platforms: 'Platforms supported by app? Format eg.: "web,mobile,native"',
+    category: 'Categories covered by app? Format eg.: "messaging,adult"',
+    oneLiner: 'Tagline for the app?',
+    termsOfService: 'Terms of service for the app?',
+    bigIcon: 'Big icon blob?',
+    mediumIcon: 'Medium icon blob?',
+    smallIcon: 'Small icon blob?',
```
Lezek123: It would be very unefficient and expensive to send blobs in a transaction, I though those are supposed to be URLs?

cli/src/commands/apps/createApp.ts
```diff
@@ -0,0 +1,64 @@
...
+import { IMemberRemarked, MemberRemarked } from '@joystream/metadata-protobuf'
+import ExitCodes from '../../ExitCodes'
+import { metadataToBytes } from '../../helpers/serialization'
+import { getInputJson } from '../../helpers/InputOutput'
+import AppCommandBase from '../../base/AppCommandBase'
+import { AppInputDetails } from 'src/Types'
+import MembershipsCommandBase from '../../base/MembershipsCommandBase'
+
+export default class GenerateAppCreationMessage extends AppCommandBase {
+  static description = 'App creation message factory'
```
Lezek123: This command is not only generating the message, but also sending a remark transaction to actually create the app, so I think the class name and description is a little misleading

cli/src/commands/apps/createApp.ts
```diff
@@ -0,0 +1,64 @@
...
+      const name = await this.promptAppName()
+      const appMetadata = await this.promptAppMetadata()
+
+      createAppRemarked = {
+        createApp: {
+          name,
+          appMetadata,
+        },
+      }
+    }
```
Lezek123: I think this can be simplified to something like:
```suggestion
    const defaults: Partial<AppInputDetails> = input ? await getInputJson<Partial<AppInputDetails>>(input) : {}
    const name = defaults?.name || (await this.promptAppName())
    const appMetadata = skip ? defaults : await this.promptAppMetadata(defaults)
    const createAppRemarked = {
      createApp: {
        name,
        appMetadata,
      },
    }
```

cli/src/base/AppCommandBase.ts
```diff
@@ -0,0 +1,54 @@
...
+      return value.split(',').map((str: string) => str.trim())
+    }
+    return value
+  }
+
+  async sendRemark(message: Bytes): Promise<u64> {
+    const {
+      id,
+      membership: { controllerAccount },
+    } = await this.getRequiredMemberContext(true)
```
Lezek123, line 56: I think it's a good practice to get context at the beginning of command execution, since it's a fallible call.
Otherwise the command may fail right after the user already provided all the input:
image

cli/src/base/AppCommandBase.ts
```diff
@@ -0,0 +1,54 @@
...
+import { u64 } from '@polkadot/types'
+
+export default abstract class AppCommandBase extends MembershipsCommandBase {
+  appMetadataForm = {
+    description: 'App description?',
+    authKey: 'App public auth key?',
+    websiteUrl: 'Site URL where user can learn more about the app?',
+    useUri: 'App URL?',
+    platforms: 'Platforms supported by app? Format eg.: "web,mobile,native"',
+    category: 'Categories covered by app? Format eg.: "messaging,adult"',
```
Lezek123: Only one category per app is currently supported, so I think the description should be updated
```suggestion
    category: 'Category of the app? Format eg.: "messaging"',
```

cli/src/commands/apps/createApp.ts
```diff
@@ -0,0 +1,64 @@
...
+      description: "If true command won't prompt missing fields",
+    }),
+    ...MembershipsCommandBase.flags,
+  }
+
+  async run(): Promise<void> {
+    let createAppRemarked: IMemberRemarked | null
+    const { input, skip } = this.parse(GenerateAppCreationMessage).flags
+    if (input) {
+      const inputBody = await getInputJson<AppInputDetails>(input)
```
Lezek123: It would be good to have an example input files like `cli/examples/content/CreateApp.json` and `cli/examples/content/UpdateApp.json`

cli/src/commands/apps/createApp.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      createAppRemarked = {
+        createApp: {
+          name,
+          appMetadata,
+        },
+      }
+    }
+    const message = metadataToBytes(MemberRemarked, createAppRemarked)
+    const memberId = await this.sendRemark(message)
```
Lezek123: Could you log all final parameters of the app and add a final confirmation here? (Something like: https://github.com/Joystream/joystream/blob/master/cli/src/commands/content/createChannel.ts#L73)

cli/src/commands/apps/createApp.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.log(chalk.green(`App created with owner member ID: ${memberId}`))
```
Lezek123: This will not always be true, for example, if an app already exists with a provided name.
You can either adjust the message or actually verify if the app was created against the QN (the 2nd approach is ofc more complicated)

cli/src/commands/apps/updateApp.ts
```diff
@@ -0,0 +1,57 @@
+import { flags } from '@oclif/command'
+import chalk from 'chalk'
+import { IAppMetadata, IMemberRemarked, MemberRemarked } from '@joystream/metadata-protobuf'
+import { metadataToBytes } from '../../helpers/serialization'
+import { getInputJson } from '../../helpers/InputOutput'
+import AppCommandBase from '../../base/AppCommandBase'
+import MembershipsCommandBase from '../../base/MembershipsCommandBase'
+
+export default class GenerateAppCreationMessage extends AppCommandBase {
+  static description = 'App creation message factory'
```
Lezek123: Invalid class name and description

cli/src/commands/apps/deleteApp.ts
```diff
@@ -0,0 +1,31 @@
+import { flags } from '@oclif/command'
+import chalk from 'chalk'
+import { IMemberRemarked, MemberRemarked } from '@joystream/metadata-protobuf'
+import { metadataToBytes } from '../../helpers/serialization'
+import AppCommandBase from '../../base/AppCommandBase'
+import MembershipsCommandBase from '../../base/MembershipsCommandBase'
+
+export default class GenerateAppCreationMessage extends AppCommandBase {
+  static description = 'App creation message factory'
```
Lezek123: Invalid class name and description

cli/src/commands/apps/updateApp.ts
```diff
@@ -0,0 +1,57 @@
...
+    }),
+    ...MembershipsCommandBase.flags,
+  }
+
+  async run(): Promise<void> {
+    let updateAppRemarked: IMemberRemarked | null
+    const { input, appId, skip } = this.parse(GenerateAppCreationMessage).flags
+    if (input) {
+      const inputBody = await getInputJson<IAppMetadata>(input)
+      const appMetadata = skip ? inputBody : await this.promptAppMetadata(inputBody)
```
Lezek123: Unless using a combination of `--input` and `--skip` it's impossible to partially update the app.
I suggest not setting any value of `prev[key]` (or setting it to `null` / `undefined`) if `promptValue` is an empty string in `promptAppMetadata`.
This of course introduces another issue, which is inability to explicity _unset_ a filed, but I think that's ok (unless you want to solve that one too, for exmaple, by asking whether user wants to update each given field or not)

cli/src/commands/apps/updateApp.ts
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      const appMetadata = await this.promptAppMetadata()
+
+      updateAppRemarked = {
+        updateApp: {
+          appId,
+          appMetadata,
+        },
+      }
+    }
```
Lezek123: Can be simplified to:
```suggestion
    const inputBody = input ? await getInputJson<IAppMetadata>(input) : {}
    const appMetadata = skip ? inputBody : await this.promptAppMetadata(inputBody)
    const updateAppRemarked = {
      updateApp: {
        appId,
        appMetadata,
      },
    }
```

cli/src/commands/apps/updateApp.ts
```diff
@@ -0,0 +1,57 @@
...
+      description: `Path to JSON file containing app details`,
+    }),
+    appId: flags.string({
+      required: true,
+      description: `ID of the app to update`,
+    }),
+    skip: flags.boolean({
+      char: 's',
+      description: "If true command won't prompt missing fields",
+    }),
```
Lezek123: Actually here I don't think `skip` makes any sense without `input`, so you can add a dependency:
```suggestion
    skip: flags.boolean({
      char: 's',
      description: "If true command won't prompt missing fields",
      dependsOn: ['input'],
    }),
```

tests/network-tests/src/flows/content/createApp.ts
```diff
@@ -14,6 +15,10 @@ export async function createApp({ api, query }: FlowProps): Promise<void> {
   await new FixtureRunner(createMembersFixture).run()
   const [member] = createMembersFixture.getCreatedItems().members
 
+  const joystreamCli = await createJoystreamCli()
+  await joystreamCli.init()
```
Lezek123:
```suggestion
```
This is already called by `createJoystreamCli`

tests/network-tests/src/flows/content/createAppActions.ts
```diff
@@ -35,7 +36,12 @@ export async function createAppActions({ api, query }: FlowProps): Promise<void>
     authKey: appPublicKeyHex,
   }
   const appName = 'app_for_actions'
-  await api.createApp(appName, appMetadata, member.memberId)
+
+  const joystreamCli = await createJoystreamCli()
+  await joystreamCli.init()
```
Lezek123:
```suggestion
```
This is already called by `createJoystreamCli`

tests/network-tests/src/flows/content/deleteApp.ts
```diff
@@ -22,7 +23,11 @@ export async function deleteApp({ api, query }: FlowProps): Promise<void> {
     platforms: ['web', 'mobile'],
   }
 
-  await api.createApp(appToDeleteName, appToDeleteMetadata, member.memberId)
+  const joystreamCli = await createJoystreamCli()
+  await joystreamCli.init()
```
Lezek123:
```suggestion
```
This is already called by `createJoystreamCli`

tests/network-tests/src/flows/content/updateApp.ts
```diff
@@ -14,6 +15,10 @@ export async function updateApp({ api, query }: FlowProps): Promise<void> {
   await new FixtureRunner(createMembersFixture).run()
   const [member] = createMembersFixture.getCreatedItems().members
 
+  const joystreamCli = await createJoystreamCli()
+  await joystreamCli.init()
```
Lezek123:
```suggestion
```
This is already called by `createJoystreamCli`

Lezek123 (changes requested):
Also, in `cli/package.json` please add `apps` under `oclif.topcis`

cli/src/commands/apps/updateApp.ts
```diff
@@ -0,0 +1,57 @@
...
+        updateApp: {
+          appId,
+          appMetadata,
+        },
+      }
+    }
+
+    const updateAppMessage = metadataToBytes(MemberRemarked, updateAppRemarked)
+    await this.sendRemark(updateAppMessage)
+    this.log(chalk.green(`Updated App with ID: ${appId}`))
```
Lezek123: Same issue as w/ `createApp` - it's not known whether the metatransaction was actually successful (for example, the `appId` could be wrong)

cli/src/commands/apps/deleteApp.ts
```diff
@@ -0,0 +1,31 @@
...
+    const { appId } = this.parse(GenerateAppCreationMessage).flags
+    const deleteAppRemarked: IMemberRemarked = {
+      deleteApp: {
+        appId,
+      },
+    }
+
+    const deleteAppMessage = metadataToBytes(MemberRemarked, deleteAppRemarked)
+    await this.sendRemark(deleteAppMessage)
+    this.log(chalk.green(`Deleted app with ID: ${appId}`))
```
Lezek123: Same issue as in the previous comment

Lezek123 (changes requested):

cli/src/commands/apps/createApp.ts
WRadoslaw: In fixes, I just changed the message, but WDYT about adding a warning after the message:
"However this doesn't mean that changes were applied. To verify try to run `apps:memberApps`"
The extra command will just fetch apps where the owner member eq selected member.

cli/src/commands/apps/updateApp.ts
WRadoslaw: Maybe just pressing ENTER to skip or pass a single dash to unset a field will be straightforward enough. Can't think of any misusage.

Looks like metaprotocol drops fields with a `null` value, and the same goes for `undefined`, to resolve it I went with an empty string to unset a field.

cli/src/commands/apps/updateApp.ts
```diff
@@ -22,36 +23,32 @@ export default class GenerateAppCreationMessage extends AppCommandBase {
...
-          appId,
-          appMetadata,
+    const defaults: Partial<AppInputDetails> = input ? await getInputJson<Partial<AppInputDetails>>(input) : {}
+    const appMetadata = skip ? defaults : await this.promptAppMetadata(defaults)
+    const updateAppRemarked: IMemberRemarked = {
+      updateApp: {
+        appId,
+        appMetadata: {
+          ...appMetadata,
+          description: undefined,
```
Lezek123:
```suggestion
```

cli/src/base/AppCommandBase.ts
```diff
@@ -27,12 +31,18 @@ export default abstract class AppCommandBase extends MembershipsCommandBase {
...
+      }
       prev[key] = this.processValue(key, promptValue)
       return prev
     }, Promise.resolve({} as Record<string, string | string[]>))
   }
 
   processValue(key: string, value: string): string | string[] {
+    if (value === '-') {
+      return ''
+    }
```
Lezek123: Nice solution!
The only field it doesn't work for is `platforms`, but arrays are tricky to deal with in protobuf in general, because there is no difference between empty array and no value provided (in terms of encoding) so best we can do in this case is use `['']` (array with empty string)

cli/src/commands/apps/createApp.ts
Lezek123: Sounds good, but I think this can be introduced in another PR.

Lezek123 (changes requested):
Looks good, just cosmetic changes requested (I think the `description` line was commited by accident)

Lezek123 (approved):
LGTM

PR #4659, opened 2023-02-23
Network tests: Fix memory leak, increase storageSync wait time

Lezek123:
Example failure caused by insufficient wait time: https://github.com/Joystream/joystream/actions/runs/4245801993/jobs/7381940413

Example failure caused by memory leak: https://github.com/Joystream/joystream/actions/runs/4242498829/jobs/7375275681


mnaamani (approved):
Oh dear, I think this memory leak has been around for ages :) great that you found it.
Just curios what tooling did you use to identify this is where the leak was coming from.

PR #4662, opened 2023-02-23
CLI: Metaprotocol utils

Lezek123:
Utility commands to encode and decode metatada-protobuf messages


zeeshanakram3 (changes requested):
1. Just an enhancement, but I think that the `type` flag should be made optional, and it makes sense to prompt for a message type in all of the commands if the user doesn't specify the message type. e.g.

```ts
  private async promptForMessageType(
    message = 'Choose what message type you will use to encode the message:'
  ): Promise<string> {
    return this.simplePrompt({
      message,
      type: 'list',
      choices: Object.keys(allMessages)
        .sort()
        .map((c) => ({ name: c, value: c })),
    })
  }
  
  
  async run(): Promise<void> {
    // ...
    let { type } = this.parse(EncodeMessageCommand).flags
    if (!type) {
      type = await this.promptForMessageType()
    }
    // ...
  }
```

WDYT?

2. I tried `util:messageStructure` for few of the messages, and it failed for `AppAction` message with the following error:
```
Error: Cannot understand the message structure
```

zeeshanakram3 (approved):
LGTM

PR #4663, opened 2023-02-24
Orion v1: Add missing APP_PRIVATE_KEY env

Lezek123:


mnaamani (approved):
LGTM
Thanks for updating the k8 code, although we don't actually deploy k8 its good to at least have them correct. Most of the arguments that are secrets should actually be using pulumi secrets module, but honestly we need a full revamp of all the pulumi/k8 code.

Issue #4661, opened 2023-02-23
QN: Deletion of an app will fail if there are already videos/channels associated with the app


Lezek123:
There's no easy way to set things like `ON DELETE CASCADE` or `ON DELETE SET NULL` currently in QN, so deletion of any entity will fail (crash the QN) if there are already other entities that reference it.

**The problem:** We introduced deletion for the apps, but later we introduced `Video.entryApp` and `Channel.entryApp` relations. Now when we're deleting the app we're not checking if any videos/channels reference it (https://github.com/Joystream/joystream/blob/apps-metaprotocol/query-node/mappings/src/content/app.ts#L102), which introduces a possibly to break the QN.

**Solutions:**
- Disallow app deletion
- Don't drop the actual entity when an app is deleted (for example: just set `isDeleted` to `true)
- Query videos and channels that reference the app and unset the relations manually (could become expensive once there are a lot of videos/channels referencing the app)

drillprop:
I would go with the first or second solution. I don't think we should bother with the third one. 

If we decide to pick a second, we will need to make sure, that it's impossible to use an app action, on the app that has `isDeleted` set to true. Not sure how we should tackle this. If someone for example wants to create a channel with the deleted app should we don't allow him and throw an error? Or just silently allow him to create a channel without `entryApp`? Not sure if the latter introduces more complexity @WRadoslaw

bedeho:
> will fail (crash the QN) if there are already other entities that reference it.

- This should not happen in any real example I should think of? mapping validation should prevent this.
- Option 2 is worst imho, option 1 is OK - main virtue is that its simple, option 3 seems ideal in terms of retaining flexibility for apps - but payoff in having deletion seems small. For other domains it may not be, for apps it does.

so option 1 seems best here.

Lezek123:
> This should not happen in any real example I should think of? mapping validation should prevent this.

I'm not sure what you mean. It can currently happen in the following scenario:
- A member creates an app
- The member then includes a valid, signed `AppAction` message as part of `create_video` / `create_channel` extrinsic. This results in video/channel being connected to the app in the QN
- The member (app owner) then issues a message to delete the app
- The QN should crash due to database constraint violation

bedeho:
[...]

PR #4664, opened 2023-02-24
Fix/apps deletion

WRadoslaw:


Lezek123 (commented):
Based on discussion in https://github.com/Joystream/joystream/issues/4661 I think we should remove the app deletion completely.

WRadoslaw:
Yup, I will take care of it

PR #4666, opened 2023-02-27
Apps: Bump package versions, update QN CHANGELOG and CLI REAMDE

Lezek123:
Depends on https://github.com/Joystream/joystream/pull/4667 (final CLI README update)


mnaamani (approved):
Versions looks correct, crossed checked against future ephesus package versions.

PR #4671, opened 2023-02-27
Fix/import commitment factory from package

WRadoslaw:


Lezek123 (commented):
I feel a little worried about having 2 different versions of `@joystream/types` and `@joystream/metadata-protobuf` in the monorepo.

But perhaps we could solve it by temporarily adding `@joystream/types` and `@joystream/metadata-protobuf` to yarn `resolutions` after https://github.com/Joystream/joystream/pull/4666

PR #4673, opened 2023-02-28
Fix/commitment issue

WRadoslaw:


Lezek123 (approved):
LGTM

Lezek123:
@WRadoslaw I think we also need a PR to `ephesus` to update the `generateAppActionCommitment` implementation in `@joystream/js`

Issue #4543, opened 2023-01-18
Publish final changelog and version bump for runtime & node in `Ephesus`


bedeho:
SSIA

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1203774483976270) by [Unito](https://www.unito.io)

PR #4675, opened 2023-03-01
Ephesus Runtime changelog

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4543


bwhm (approved):
Looks good!

Might have been nice to emphasize more clearly that this supersedes `Liberated`, but not really the right place...

PR #4678, opened 2023-03-01
`generateAppActionCommitment` fixes

Lezek123:
- Include new `ownerType` filed in app action commitment to simplify differentiation between different types of owners (member, curator group, channel)
- Remove `nonce` and `actionType` fields from `AppAction` message. They don't actually need to be there, since the correct values are known by the QN, so they can be only included in the commitment/signature
- `ActionType` and `CreatorType` enums are still included in `App.proto` for the purpose of standarizing the encoding of those values across different apps.
- I simplified `generateAppActionCommitment` so that it doesn't rely on values like `rawAction` and `rawAppActionMetadata` being provided as `Bytes`. The `Bytes` type is `@polkadot/types`-specific and is harder to construct than `Uint8Array` (which is universal), as it requires a type registry, importing from `@polkadot/types` or `@joystream/types` library etc. Also now the encoidng of those values is more consistent with `assets` 
- Fixed a few other small issues

### DIFF
--- a/metadata-protobuf/proto/App.proto
+++ b/metadata-protobuf/proto/App.proto
@@ -18,14 +18,14 @@ message AppAction {
   // Signature over app commitment
   optional bytes signature = 4;
 
-  // Nonce to prevent signature reusal
-  optional uint32 nonce = 5;
-
   enum ActionType {
     CREATE_VIDEO = 0;
     CREATE_CHANNEL = 1;
   }
 
-  // Type of the runtime-level action to be performed
-  optional ActionType action_type = 6;
+  enum CreatorType {
+    CHANNEL = 0;
+    MEMBER = 1;
+    CURAOTR_GROUP = 2;
+  }
 }
--- a/metadata-protobuf/proto/App.proto
+++ b/metadata-protobuf/proto/App.proto
@@ -26,6 +26,6 @@ message AppAction {
   enum CreatorType {
     CHANNEL = 0;
     MEMBER = 1;
-    CURAOTR_GROUP = 2;
+    CURATOR_GROUP = 2;
   }
 }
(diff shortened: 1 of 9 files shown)


WRadoslaw (approved):
LGTM

metadata-protobuf/proto/App.proto
```diff
@@ -18,14 +18,14 @@ message AppAction {
...
     CREATE_CHANNEL = 1;
   }
 
-  // Type of the runtime-level action to be performed
-  optional ActionType action_type = 6;
-}
+  enum CreatorType {
+    CHANNEL = 0;
+    MEMBER = 1;
+    CURAOTR_GROUP = 2;
```
drillprop:
```suggestion
    CURATOR_GROUP = 2;
```

query-node/mappings/src/content/channel.ts
```diff
@@ -93,20 +94,20 @@ export async function content_ChannelCreated(ctx: EventContext & StoreContext):
     if (appAction) {
       const channelMetadataBytes = u8aToBytes(appAction.rawAction)
       const channelMetadata = deserializeMetadata(ChannelMetadata, channelMetadataBytes)
-      const appCommitment = generateAppActionCommitment(
+      const creatorType = channel.ownerMember ? AppAction.CreatorType.MEMBER : AppAction.CreatorType.CURAOTR_GROUP
```
drillprop:
```suggestion
      const creatorType = channel.ownerMember ? AppAction.CreatorType.MEMBER : AppAction.CreatorType.CURATOR_GROUP
```

drillprop (changes requested):

zeeshanakram3 (approved):

drillprop (approved):

PR #4694, opened 2023-03-09
Fix failing tests on ephesus inviting members job

dobertRowneySr:
I have added locks for Membership WG budget resource, as the failures appear to be non-deterministic (eg some scenario runs succeed while others don't, also the failure is not reproducible on all machines).
This made me think that NodeJS threads are competing for spending from the membership WG.
#### Current extrinsics that decrease Membership WG
- `working_group::spend_from_budget`
- `membership::invite_member`

I have added a membership wg budget mutex for the fixtures referencing the above extrinsics

### DIFF
--- a/.github/workflows/run-network-tests.yml
+++ b/.github/workflows/run-network-tests.yml
@@ -143,7 +143,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        scenario: ['full', 'setupNewChain', 'setupNewChainMultiStorage', 'bonding', 'storageSync', 'memberships']
+        scenario: ['full', 'setupNewChain', 'setupNewChainMultiStorage', 'bonding', 'storageSync']
         include:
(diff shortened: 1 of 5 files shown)


dobertRowneySr:
the `invite_members` job is succeeding, running a second time to make sure it is not by chance

dobertRowneySr:
I have rearranged the execution flow. I will run the tests for another 2/3 times if they are all successful I guess we can declare success

mnaamani (approved):
It seems it was primarily having the invitation tests run before the group budget tests that did the trick. Although the lock look like it would have also solved it without re-arranging the order. Did you not find that to be the case?

dobertRowneySr:
Setting the lock did not solve the problem sometimes, but the setting the lock AND rearranging the jobs order did
There's only some other runtime hooks for rewarding the worker, but that should not happen during tests since it is every 24 hours + 40 blocks

Issue #4519, opened 2023-01-10
New proposal for setting CRT constraints


bedeho:
I don't know what the final constraints will be that the council directly control, here is a preliminary breakdown https://github.com/Joystream/joystream/issues/4151#issuecomment-1319867653, but when finalized, a new proposal is needed for setting these, and it may as well cover all of them at the same time for simplicity.

PR #4696, opened 2023-03-10
Crt governance parameters update proposal

dobertRowneySr:
addresses: #4519 
depends on : #4750 

### Changelist
- add extrinsic for updating the following CRT parameters:
   - max yearly rate
   - min amm slope parameter
   - min sale duration
   - min revenue split time to start
   - amm buy tx fees
   - amm sell tx fees
   - bloat bond
 - added integration with proposal system
 - added benchmark (but no weights)


dobertRowneySr:
in favour of #5103

PR #4719, opened 2023-03-27
add mass discretionary spending to cli (#4715)

chrlschwb:
solves #4715
Sometimes there is a need to send small amounts of tokens to multiple, like hundreds, accounts.
This new command will help with that.


cli/src/commands/working-groups/discretionarySpending.ts
```diff
@@ -0,0 +1,49 @@
+import WorkingGroupsCommandBase from '../../base/WorkingGroupsCommandBase'
+import { apiModuleByGroup } from '../../Api'
+import chalk from 'chalk'
+import ExitCodes from '../../ExitCodes'
+import { isValidBalance } from '../../helpers/validation'
+
+export default class WorkingGroupsDiscetionarySpending extends WorkingGroupsCommandBase {
```
zeeshanakram3, line 10: Incorrect spelling 
```suggestion
export default class WorkingGroupsDiscretionarySpending extends WorkingGroupsCommandBase {
```

cli/src/base/ApiCommandBase.ts
```diff
@@ -592,6 +592,34 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
     }
   }
 
+  async sendAndFollowTxMulti(account: KeyringPair, tx: SubmittableExtrinsic<'promise'>): Promise<SubmittableResult> {
```
zeeshanakram3, line 595: What is the purpose of this function? Why send `sendAndFollowTx` can't be used in CLI commands for sending extrinsics? It seems a redundant addition, so please remove it

cli/src/base/ApiCommandBase.ts
```diff
@@ -623,6 +651,21 @@ export default abstract class ApiCommandBase extends StateAwareCommandBase {
...
+    account: KeyringPair,
+    module: Module,
+    method: Method,
+    params: Submittable extends (...args: any[]) => any ? Parameters<Submittable> : []
+  ): Promise<SubmittableResult> {
+    // TODO: Replace all usages with "sendAndFollowTx"
+    const tx = await this.getUnaugmentedApi().tx[module][method](...params)
+    return this.sendAndFollowTxMulti(account, tx)
+  }
+
```
zeeshanakram3, line 668: Same as above.
What is the purpose of this function? Why send `sendAndFollowNamedTx` can't be used in CLI commands for sending extrinsics? It seems a redundant addition, so please remove it

cli/src/commands/working-groups/discretionarySpending.ts
```diff
@@ -0,0 +1,49 @@
...
+  static args = [
```
zeeshanakram3, line 11: Add the CLI command's description here.

zeeshanakram3 (changes requested):
1. [test.json](https://github.com/Joystream/joystream/blob/7dc70b2f98f76a6d0218303c04a028a848d0758b/cli/bin/test.json) file should be placed in the` examples/working-group` directory, and not in the `bin/` dir. Also, please rename it accordingly.

2. `yarn.lock` file is changed in this PR, which I think should not be the case since this PR does not add or remove any dependency from `package.json` file. So please revert the changes in `yarn.lock` file

3. What's the purpose of two different CLI commands? You can do both single or multiple discretionary spending transactions from the single file as well. So no need to create two different Commands. You can add `input` flag to `WorkingGroupsDiscretionarySpending` command.

cli/src/commands/working-groups/discretionarySpending.ts
chrlschwb: Do you mean as a comment?
The command help is working by typing, "working-groups:discretionarySpending --help"

cli/src/base/ApiCommandBase.ts
chrlschwb: The `sendAndFollowNamedTx` function doesn't require the user to confirm transaction each time.
This is very useful if the user is using mass discretionary spending. For example, if they send 100 JOY to 100 people, they do not need to confirm 100 transactions in CLI.

cli/src/commands/working-groups/discretionarySpending.ts
zeeshanakram3: I mean a `description` property of the class, like it's part of other [commands](https://github.com/Joystream/joystream/blob/7dc70b2f98f76a6d0218303c04a028a848d0758b/cli/src/commands/working-groups/decreaseWorkerStake.ts#L10) too.

cli/src/base/ApiCommandBase.ts
zeeshanakram3: If you see [requireConfirmation](https://github.com/chrlschwb/joystream/blob/d8a021cb0cf2adfc031b28ecfc4b73e6b8447004/cli/src/base/DefaultCommandBase.ts#L58) function there is a `AUTO_CONFIRM` env var which is used to decided whether to prompt for confirmation or not. Please set this variable before initiating a transaction and it will solve your problem

PR #4749, opened 2023-05-05
Nara crt release

dobertRowneySr:
updates crt release with nara


mnaamani:
Update looks good, but to make it cleaner I suggest (based on what we discussed in chat):

- Bump the spec_version to 2002
- fix the query-node metadata (take them from nara branch)
  - `query-node/chain-metadata/1001.json`
  - `query-node/chain-metadata/2001.json`
- re-generate chain metadata (which will also generate a copy as `query-node/chain-metadata/2002.json`

mnaamani (approved):

PR #4481, opened 2022-12-06
Amm benchmark

dobertRowneySr:
benchmarks the following extrinsics:
- `buy_on_amm`
- `sell_on_amm`
- `activate_amm`
- `deactivate_amm`

Issue #4378, opened 2022-10-17
Patronage interest rate calculation


dobertRowneySr:
### Current patronage computation
`patronage_amount = issuance * patronage_rate_per_block * number_of_blocks * blocks_per_year`
### Desired patronage calculation
`patronage_amount = issuance * ( 1 + patronage_rate_per_year)^{number_of_blocks/blocks_per_year} - issuance`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201958687417145/1203361526020352) by [Unito](https://www.unito.io)

bedeho:
1. isnt the patronage rate annual? not per block?
2. why is `issuance` being subtracted?

dobertRowneySr:
> 1. isnt the patronage rate annual? not per block?

`blocks` meant number of blocks in a year (I have renamed it now). The user gets to specified the annual rate, but we store it as per block rate to have easier computations (which is similar to what the pallet vesting is doing)

> 2. why is `issuance` being subtracted?

To obtain the net amount that will be granted to the creator account:
If I put 10 $ into a risk free bond with 3% interest rate in 10 years the amount will be `10 * (1 + 3%)^10` the net gain `10 * (1 + 3%)^10 - 10`

bedeho:
ok, makes sense.

ignazio-bovo:
- $r$ yearly interest rate
- $t$ = number of years
- $n$ = number of blocks in a year 

If we subdivide the years in $n$ compounding periods (corresponding to each block)

$$ I = \bigg(1 + \frac{r}{n} \bigg)^{n t} - 1 \approx e^{rt} - 1 $$

with $n$ large enough (in our case 5 259 487.66) -> ok

$$ I = e^{rt} - 1 \approx rt + \frac{(r t)^2}{2} + \frac{(rt)^3}{6} + \frac{(rt)^4}{24} $$

# Plan
- drop the convertion from yearly rate to block rate
- compute $rt + \frac{(r t)^2}{2} + \frac{(rt)^3}{6} + \frac{(rt)^4}{24} = y$ and then split $y = y_{int} + y_{frac}$ (integer + fractional)
- patronage amount is found by multiplying the `total_supply` value by `y_int` with ordinary `checked_mul` and then after add the value `Perquintill::from_float(y_frac}.mul_floor(total_supply)`

ignazio-bovo:
Addressed in #4516

ignazio-bovo:
[...]

PR #4750, opened 2023-05-05
CRT constraints and AMM benchmarks

dobertRowneySr:
fixes:
- max yearly patronage rate 
- patronage rate computation & approximation updated
- patronage tests refactored & updated
- add max limit for transfers
- adds benchmarks for amm (as #4481 has lost all changes)

## Resources and References:
- [summary table with CRT governance constraints](https://github.com/Joystream/joystream/issues/4151#issuecomment-1455821245) 
- [AMM manual page reference](https://handbook.joystream.org/system/content-directory/projecttoken#amm)
- [AMM easy to read summary (Bonding Curve means our AMM functionality)](https://medium.com/linum-labs/intro-to-bonding-curves-and-shapes-bf326bc4e11a)
- [Update patronage rate computation](https://github.com/Joystream/joystream/issues/4378)


runtime-modules/project-token/src/lib.rs
```diff
@@ -819,16 +825,15 @@ decl_module! {
...
         /// - slippage tolerance constraints respected if provided
         /// - token total supply and amount value must be s.t. `eval` function doesn't overflow
         ///
         /// Postconditions
         /// - `amount` CRT minted into account (which is created if necessary with existential deposit transferred to it)
         /// - respective JOY amount transferred from user balance to amm treasury account
         /// - event deposited
-        #[weight = 100_000_000] // TODO: adjust weight
-        fn buy_on_amm(origin, token_id: T::TokenId, member_id: T::MemberId, amount: <T as Config>::Balance, deadline: Option<<T as timestamp::Config>::Moment>, slippage_tolerance: Option<(Permill, JoyBalanceOf<T>)>) -> DispatchResult {
+        #[weight = WeightInfoToken::<T>::buy_on_amm_with_existing_account()]
```
dobertRowneySr, line 826: Replace this with a `max(buy_on_amm_with_existing_account, buy_on_amm_with_new_account)`
as the weight might differ on the ref. machine

runtime-modules/project-token/src/lib.rs
```diff
@@ -972,6 +966,19 @@ decl_module! {
             Ok(())
         }
 
+        #[weight = 100_000_000] // TODO: adjust weight
```
dobertRowneySr: might as well benchmark this

runtime-modules/project-token/src/lib.rs
```diff
@@ -913,10 +912,10 @@ decl_module! {
         /// - total supply decreased by amount
         /// - respective JOY amount transferred from amm treasury account to user account
         /// - event deposited
-        #[weight = 100_000_000] // TODO: adjust weight
-        fn sell_on_amm(origin, token_id: T::TokenId, member_id: T::MemberId, amount: <T as Config>::Balance, deadline: Option<<T as timestamp::Config>::Moment>, slippage_tolerance: Option<(Permill, JoyBalanceOf<T>)>) -> DispatchResult {
```
dobertRowneySr, line 917: This is good, timestamp info can be manipulated by validator node (https://neptunemutual.com/blog/understanding-block-timestamp-manipulation/)

runtime-modules/project-token/src/lib.rs
```diff
@@ -247,20 +245,28 @@ decl_module! {
...
         pub fn transfer(
             origin,
             src_member_id: T::MemberId,
             token_id: T::TokenId,
-            outputs: TransfersOf<T>,
+            outputs: TransferOutputsOf<T>,
             metadata: Vec<u8>
         ) -> DispatchResult {
+
+            // // security check
```
dobertRowneySr: Remove this if dead code

runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -264,271 +173,245 @@ fn decreasing_patronage_rate_fails_with_target_rate_exceeding_current_rate() {
...
+        increase_block_number_by(DEFAULT_BLOCK_INTERVAL);
 
-        let result = Token::claim_patronage_credit(token_id, owner_id);
+        let result = ReducePatronageRateToFixture::default().execute_call();
 
         assert_ok!(result);
     })
 }
 
+// TODO(mrbovo): allow for more than 100% claim over supply
```
dobertRowneySr: no longer relevant: remove this

runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -264,271 +173,245 @@ fn decreasing_patronage_rate_fails_with_target_rate_exceeding_current_rate() {
...
-            owner_id,
+            DEFAULT_TOKEN_ID,
+            compute_correct_patronage_amount(DEFAULT_INITIAL_ISSUANCE),
+            DEFAULT_ISSUER_MEMBER_ID,
         ));
     })
 }
 
 #[test]
 fn claim_patronage_ok_with_credit_accounting() {
```
dobertRowneySr:
```suggestion
fn claim_patronage_ok_with_correct_patronage_credit_accounting() {
```

runtime-modules/project-token/src/tests/patronage.rs
dobertRowneySr:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -886,6 +886,7 @@ parameter_types! {
     pub const ProjectTokenModuleId: PalletId = PalletId(*b"mo:token"); // module: token
     pub const MaxVestingSchedulesPerAccountPerToken: u32 = 5;
     pub const BlocksPerYear: u32 = 5259600; // 365,25 * 24 * 60 * 60 / 6
+    pub const MaxOutputs: u32 = 256; // TODO(Martin, Ignazio) : find a suitable value
```
dobertRowneySr, line 904: Need a separate issue for:
- max patronage rate
- min amm slope

runtime-modules/project-token/src/types.rs
```diff
@@ -871,6 +892,11 @@ impl<Balance, VestingScheduleParams>
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, TypeInfo)]
 pub struct Transfers<MemberId, Payment>(pub BTreeMap<MemberId, Payment>);
 
+// #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, TypeInfo)]
```
dobertRowneySr: Test code, remove this

dobertRowneySr (commented):
Some fixes are needed

bin/node/src/chain_spec/project_token_config.rs
```diff
@@ -11,6 +11,8 @@ pub fn production_config() -> ProjectTokenConfig {
         min_revenue_split_duration: days!(21),
         min_revenue_split_time_to_start: 0,
         sale_platform_fee: Permill::from_percent(2),
+        max_yearly_patronage_rate: Permill::from_percent(5).into(),
```
dobertRowneySr, line 12: no migration is needed for these parameters as they are new with V2002

runtime-modules/content/src/benchmarks/benchmarking.rs
```diff
@@ -1776,6 +1776,70 @@ benchmarks! {
             );
         }
 
+    activate_amm {
```
dobertRowneySr, line 1779: Consider adding a comment about why member channel all max doesn't make op. more expensive

runtime-modules/content/src/lib.rs
dobertRowneySr, line 1: Ok it appears no (data) migrations are needed here on content pallet

runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -264,271 +173,244 @@ fn decreasing_patronage_rate_fails_with_target_rate_exceeding_current_rate() {
...
-    let blocks = block!(2); // patronage rate for period = 110% > 100%
-
-    let params = TokenDataBuilder::new_empty().with_patronage_rate(rate);
-    // rate for period = 110% but effective rate is capped at 100% -> 100% supply + supply
-    let expected = balance!(100) + supply;
-
-    let config = GenesisConfigBuilder::new_empty()
-        .with_token_and_owner(token_id, params.build(), owner_id, supply)
-        .build();
+fn claim_patronage_ok_with_correct_credit_accounting_and_more_than_100_percent_supply() {
```
dobertRowneySr, line 249: This also accounts for testing numeric precision, however there are several tests in the `common/numerical.rs` library for this

query-node/mappings/src/proposals.ts
```diff
@@ -333,9 +337,15 @@ async function handleRuntimeUpgradeProposalExecution(event: SubstrateEvent, stor
...
-export async function proposalsCodex_ProposalCreated({ store, event }: EventContext & StoreContext): Promise<void> {
+export async function proposalsCodex_ProposalCreated({
+  store,
+  event,
+  block,
+}: EventContext & StoreContext): Promise<void> {
+  const specVersion = block.runtimeVersion.specVersion
   const [proposalId, generalProposalParameters, runtimeProposalDetails, proposalThreadId] =
-    new ProposalCreatedEvent_V1001(event).params
+    specVersion >= 2002 ? new ProposalCreatedEvent_V2002(event).params : new ProposalCreatedEvent_V1001(event).params
```
dobertRowneySr, line 347: Should we update the mappings in order to account for the new update max yearly patronage rate proposal?
If so I need to open an issue about this

bin/node/src/chain_spec/project_token_config.rs
ignazio-bovo: Plus these needs to be changed into `Balances`

runtime-modules/content/src/benchmarks/benchmarking.rs
dobertRowneySr: curator channel operations seems to be more expensive due to permission checking

runtime-modules/project-token/src/lib.rs
dobertRowneySr: Code is no longer neeed due to Bounded Vec type use as parameter

bin/node/src/chain_spec/project_token_config.rs
```diff
@@ -11,6 +11,9 @@ pub fn production_config() -> ProjectTokenConfig {
         min_revenue_split_duration: days!(21),
         min_revenue_split_time_to_start: 0,
         sale_platform_fee: Permill::from_percent(2),
+        max_yearly_patronage_rate: Permill::from_percent(20).into(), // TODO: update, raising might make benchmarks fail
+        min_amm_slope_parameter: 10, // TODO: update, raising might make benchmarks fail
```
mnaamani: The default storage value set in project-token decl_storage macro is 15% for max_yearly_patronage_rate and 0 for min_amm_slope_parameter. In a runtime upgrade those defaults will be used because they are now being introduced in nara upgrade. If we are not happy with the defaults, we need to add a migration on_runtime_upgrade.

query-node/mappings/src/proposals.ts
mnaamani: Will it depends on when we decide to enable creating tokens, or start adding integration tests then yes we would need to update the mappings starting with handling new proposal type in `parseProposalDetails()`

runtime-modules/common/src/numerical.rs
```diff
@@ -0,0 +1,156 @@
+use core::ops::{Div, Mul};
+
+use sp_runtime::{
+    traits::{One, Saturating},
+    FixedPointNumber, FixedU128, PerThing, Permill, Perquintill,
+};
+
+pub const ORDER: usize = 20;
```
mnaamani: could you comment what this  constant stands for?

runtime/src/proposals_configuration/testing.rs
```diff
@@ -331,3 +331,18 @@ pub(crate) fn update_channel_payouts_proposal() -> ProposalParameters<BlockNumbe
         constitutionality: 1,
     }
 }
+
+// Proposal parameters for the 'Update Max Yearly Patronage Rate' proposal
+pub(crate) fn update_max_yearly_patronage_rate_proposal() -> ProposalParameters<BlockNumber, Balance>
+{
+    ProposalParameters {
+        voting_period: days!(7),
```
mnaamani: voting and grace periods for testing,playground need to be much lower. See values for other proposals.

runtime/src/weights/block_weights.rs
```diff
@@ -16,23 +16,22 @@
 // limitations under the License.
 
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-10-20 (Y/M/D)
+//! DATE: 2023-05-11 (Y/M/D)
```
mnaamani: block, extrinsics and rocksdb _weights.rs should not be modified in this PR.

runtime-modules/content/src/weights.rs
```diff
@@ -807,20 +801,42 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 			.saturating_add(T::DbWeight::get().reads(6 as Weight))
 			.saturating_add(T::DbWeight::get().writes(3 as Weight))
 	}
 	// Storage: Content ChannelById (r:1 w:0)
 	// Storage: Instance3WorkingGroup WorkerById (r:1 w:0)
 	// Storage: Content CuratorGroupById (r:1 w:0)
+	// Storage: Token TokenInfoById (r:1 w:1)
+	// Storage: Token MinAmmSlopeParameter (r:1 w:0)
+	// Storage: System Account (r:1 w:1)
+	fn activate_amm() -> Weight {
```
mnaamani, line 1061: perhaps we can keep only the new weight functions (and maybe issue_new_token which was modified) in this file and checkout all other changes.
We will do a final weight generation step before nara release.

runtime-modules/project-token/src/lib.rs
mnaamani: It doesn't look like the deadline was used for any purpose anyway, so it makes sense to drop it.

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -79,15 +79,15 @@ const MAX_KILOBYTES_METADATA: u32 = 100;
 
 // Creator tokens
 const MAX_CRT_INITIAL_ALLOCATION_MEMBERS: u32 = 1024;
-const MAX_CRT_ISSUER_TRANSFER_OUTPUTS: u32 = 1024;
+const MAX_CRT_ISSUER_TRANSFER_OUTPUTS: u32 = 256; // same as in Token MaxOuputs
```
mnaamani: If it is a runtime constant, perhaps it can be read from the pallet instead of re-defining a separate constant and having to remember to keep them in sync?

mnaamani (commented):
I left just an initial high level review, did not go into implementation details yet.

runtime-modules/common/src/numerical.rs
```diff
@@ -0,0 +1,156 @@
...
+
+// does not work with 100% interest
+pub fn natural_log_1_plus_x(interest: Permill) -> Perquintill {
```
mnaamani: I see this is not used expect internally from `one_plus_interest_pow_frac` so maybe keep it private?
Similarly for `one_plus_interest_pow_frac`.

If this function "does not work with 100% interest" would would the output be, and should we guard against invalid input?

runtime-modules/common/src/numerical.rs
```diff
@@ -0,0 +1,156 @@
...
+        result = result.saturating_add(terms[2 * i + 1]);
+    }
+    for i in 0..ORDER.div(2) {
+        result = result.saturating_sub(terms[2 * i]);
+    }
+
+    result
+}
+
+pub fn one_plus_interest_pow_frac(interest: Permill, exp: Perquintill) -> FixedU128 {
```
mnaamani: Can you add rust-doc just to show what is being calculated

runtime-modules/project-token/src/benchmarking.rs
```diff
@@ -28,15 +30,18 @@ const DEFAULT_TOKENS_ON_SALE: u32 = 100_000;
...
-// Patronage
-const DEFAULT_PATRONAGE: YearlyRate = YearlyRate(Permill::from_percent(1));
+const DEFAULT_SPLIT_ALLOCATION: u32 = 4_000_000; // DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE
+const DEFAULT_SPLIT_PAYOUT: u32 = 2_000_000;
+const DEFAULT_SPLIT_PARTICIPATION: u32 =
+    DEFAULT_SPLIT_PAYOUT * ((DEFAULT_TOKEN_ISSUANCE / DEFAULT_SPLIT_ALLOCATION) as u32);
+
+// Amm
+const DEFAULT_AMM_AMOUNT: u32 = 1000;
+const DEFAULT_AMM_jOY_AMOUNT: u32 = 5_100_000; // (a = 10) * amount^2 /2 + (b = 100) * amount, amount = 1000
```
mnaamani:
```suggestion
const DEFAULT_AMM_JOY_AMOUNT: u32 = 5_100_000; // (a = 10) * amount^2 /2 + (b = 100) * amount, amount = 1000
```

runtime-modules/project-token/src/benchmarking.rs
```diff
@@ -28,15 +30,18 @@ const DEFAULT_TOKENS_ON_SALE: u32 = 100_000;
...
+const DEFAULT_SPLIT_REVENUE: u32 = 8_000_000;
 const DEFAULT_REVENUE_SPLIT_RATE: Permill = Permill::from_percent(50);
-const DEFAULT_SPLIT_ALLOCATION: u64 = 4_000_000; // DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE
-const DEFAULT_SPLIT_PAYOUT: u64 = 2_000_000;
-const DEFAULT_SPLIT_PARTICIPATION: u64 =
-    DEFAULT_SPLIT_PAYOUT * DEFAULT_TOKEN_ISSUANCE / DEFAULT_SPLIT_ALLOCATION;
-
-// Patronage
-const DEFAULT_PATRONAGE: YearlyRate = YearlyRate(Permill::from_percent(1));
+const DEFAULT_SPLIT_ALLOCATION: u32 = 4_000_000; // DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE
```
mnaamani, line 33: Why not evaluate it as per comment`DEFAULT_REVENUE_SPLIT_RATE * DEFAULT_SPLIT_REVENUE`

runtime-modules/project-token/src/benchmarking.rs
```diff
@@ -237,26 +265,18 @@ benchmarks! {
...
-                (
-                    member_id,
-                    Payment::<<T as Config>::Balance> {
-                        amount: DEFAULT_TX_AMOUNT.into()
-                    }
-                )
-            })
-            .collect()
-        );
+        let mut outputs = TransferOutputsOf::<T>::with_bounded_capacity((o + 1) as usize);
```
mnaamani: by adding + 1 are not going over the `MAX_TX_OUTPUTS`. range of `o` starts at 1, so I don't see the need to add +1 if we are concerned calling `with_bounded_capacity` with args `0`..

runtime-modules/content/src/benchmarks/mod.rs
```diff
@@ -227,19 +228,20 @@ fn get_byte(num: u64, byte_number: u8) -> u8 {
...
 fn handle_from_id<T: membership::Config>(id: u64) -> Vec<u8> {
-    let min_handle_length = 1;
+    id.to_be_bytes().to_vec()
+    // let min_handle_length = 1;
 
-    let mut handle = vec![];
+    // let mut handle = vec![];
 
-    for i in 0..16 {
-        handle.push(get_byte(id, i));
```
mnaamani, line 235: this is the only use of `get_byte()`, if we are sure it is not needed anymore, then we can remove commented code and drop implementation of get_byte()

runtime-modules/project-token/src/lib.rs
```diff
@@ -247,20 +245,22 @@ decl_module! {
...
             metadata: Vec<u8>
         ) -> DispatchResult {
+
             let sender = T::MemberOriginValidator::ensure_member_controller_account_origin(
                 origin,
                 src_member_id
             )?;
 
             // Currency transfer preconditions
+            // let btree_outputs = Transfers(outputs.into_iter().map(|(member_id, amount)| (member_id, PaymentWithVesting::from(amount))).collect::<BTreeMap<_,_>>());
```
mnaamani: drop unused code?

runtime-modules/project-token/src/lib.rs
```diff
@@ -176,6 +168,12 @@ decl_storage! { generate_storage_info
 
         /// AMM sell transaction fee percentage
         pub AmmSellTxFees get(fn amm_sell_tx_fees) config(): Permill;
+
+        /// Max patronage rate allowed
+        pub MaxYearlyPatronageRate get(fn max_yearly_patronage_rate) config(): YearlyRate = YearlyRate(Permill::from_percent(15));
+
+        /// Minimum slope parameters allowed for AMM curve
+        pub MinAmmSlopeParameter get(fn min_amm_slope_parameter) config(): TokenBalanceOf<T>;
```
mnaamani, line 174: there is no need for a disptachable call to set/update this value ?

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -63,7 +63,7 @@ fn amm_buy_fails_with_member_and_origin_auth() {
 #[test]
 fn amm_buy_succeeds_with_new_user() {
     let token_id = token!(1);
-    let ((user_member_id, user_account_id), user_balance) = (member!(2), joy!(5_000_000));
+    let ((user_member_id, user_account_id), user_balance) = (member!(2), joy!(5_000_000_000));
```
mnaamani: All tests are updated to give member a larger initial balance. Assuming tests were failing. Can we use a constant that might highlight what and why there is a minimum for tests to pass?

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -810,10 +786,14 @@ fn deactivate_fails_with_invalid_token_id() {
 fn deactivate_fails_with_too_much_amm_provided_supply_outstanding() {
     let amount = Permill::from_percent(10).mul_floor(DEFAULT_INITIAL_ISSUANCE);
     // give enough balance to bond amount
-    let ((user_member_id, user_account_id), user_balance) = (member!(2), joy!(10_000_000_000));
+    let ((user_member_id, user_account_id), user_balance) =
+        (member!(2), joy!(10_000_000_000_000_000_000));
```
mnaamani: How is are these values being selected?

runtime-modules/project-token/src/tests/amm.rs
```diff
@@ -810,10 +786,14 @@ fn deactivate_fails_with_invalid_token_id() {
...
     build_default_test_externalities_with_balances(vec![(user_account_id, user_balance)])
         .execute_with(|| {
-            IssueTokenFixture::default().execute_call().unwrap();
+            IssueTokenFixture::default()
+                .with_supply(1_000u128)
```
mnaamani: how is it determined that 1_000 will result in the expected error?

runtime-modules/project-token/src/tests/canonical.rs
```diff
@@ -1133,37 +1132,6 @@ fn issue_token_ok_with_token_info_added() {
     })
 }
 
-#[test]
```
mnaamani, line 1136: why did we drop this test?

runtime-modules/project-token/src/tests/mock.rs
```diff
@@ -617,11 +628,19 @@ macro_rules! block {
 }
 
 // ------ General constants ---------------
+pub const DEFAULT_TOKEN_ID: u64 = 1;
+pub const DEFAULT_ISSUER_ACCOUNT_ID: u64 = 1001;
+pub const DEFAULT_ISSUER_MEMBER_ID: u64 = 1;
 pub const DEFAULT_BLOAT_BOND: u128 = 0;
-pub const DEFAULT_INITIAL_ISSUANCE: u128 = 1_000_000;
+pub const DEFAULT_INITIAL_ISSUANCE: u128 = 100_000_000;
```
mnaamani: Do tests fail with smaller values?

runtime-modules/project-token/src/tests/mock.rs
```diff
@@ -732,3 +752,8 @@ pub(crate) fn amm_function_values(
         AmmOperation::Sell => DEFAULT_AMM_SELL_FEES.left_from_one().mul_floor(res),
     }
 }
+
+// computes  supply * [(1 + 10%/100%)^(10/BlocksPerYear::get()) - 1]
+pub(crate) fn compute_correct_patronage_amount(supply: Balance) -> Balance {
+    Perquintill::from_parts(181215751000).mul_floor(supply)
```
mnaamani: from comment I assume this is dependant on AMM_CURVE_SLOPE and AMM_CURVE_INTERCEPT?
If so we shouldn't have a magic number `181215751000`, but base it on a const function computing it?

runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -36,226 +29,142 @@ fn issue_token_ok_with_patronage_tally_count_zero() {
...
-        .build();
-    let config = GenesisConfigBuilder::new_empty()
-        .with_token_and_owner(token_id, token_info, owner_id, init_supply)
-        .build();
-
-    build_test_externalities(config).execute_with(|| {
-        let result = Token::reduce_patronage_rate_to(token_id, decrement);
+        let result = ReducePatronageRateToFixture::default()
+            .with_target_rate(DEFAULT_YEARLY_PATRONAGE_RATE.into())
+            .execute_call();
```
mnaamani, line 86: we should also be testing the correct value is set, not just Ok result

runtime-modules/project-token/src/tests/patronage.rs
```diff
@@ -264,271 +173,244 @@ fn decreasing_patronage_rate_fails_with_target_rate_exceeding_current_rate() {
...
+        increase_block_number_by(DEFAULT_BLOCK_INTERVAL);
+
+        ClaimPatronageCreditFixture::default()
+            .execute_call()
+            .unwrap();
+
+        let approx =
+            Token::account_info_by_token_and_member(DEFAULT_TOKEN_ID, DEFAULT_ISSUER_MEMBER_ID)
+                .transferrable::<Test>(System::block_number());
+        let target = 1000000181215750585898368181876;
```
mnaamani: I'm having a hard time accepting these pre-computed values.

runtime-modules/project-token/src/types.rs
```diff
@@ -1,32 +1,48 @@
...
+    ops::Add,
     vec::Vec,
 };
 use storage::{BagId, DataObjectCreationParameters};
 
 // crate imports
 use crate::{errors::Error, Config, RepayableBloatBondOf};
 
+// trait "aliases"
+// `BlockNumber` will be implemented as `u64` in the runtime configuration
```
mnaamani: The runtime implements BlockNumber as u32, not u64. Is this assumption here problematic?

runtime-modules/project-token/src/types.rs
```diff
@@ -602,43 +616,14 @@ pub(crate) enum AmmOperation {
...
-    pub(crate) fn from_params(params: AmmParams) -> Self {
+impl<Balance: TokenBalanceTrait> AmmCurve<Balance> {
+    pub(crate) fn from_params<T: Config>(params: AmmParams<Balance>) -> Self {
         Self {
             slope: params.slope,
             intercept: params.intercept,
             provided_supply: Balance::zero(),
         }
     }
-    pub(crate) fn eval<T: Config>(
```
mnaamani, line 613: This seemed like a better place for the 'eval' implementation., but it was made into a standalone function in the content pallet. What is the reason behind this?

runtime-modules/project-token/src/types.rs
```diff
@@ -789,9 +774,52 @@ pub enum MerkleSide {
...
+
+        let result = one_plus_interest_pow_fixed(self.0, rate);
+        result.saturating_sub(FixedU128::one())
+    }
+}
+
+impl Add for YearlyRate {
+    type Output = Self;
+    fn add(self, rhs: Self) -> Self::Output {
+        Self(self.0.add(rhs.0))
```
mnaamani: should we be concerned with overflows?

runtime/src/utils.rs
```diff
@@ -148,7 +148,7 @@ macro_rules! hours {
 #[macro_export]
 macro_rules! days {
     ($a:expr) => {{
-        hours!(24) * $a
+        ((24 * 60 * 60 * 1000) / ExpectedBlockTime::get()) as u32 * $a
```
mnaamani, line 153: mathematically it looks the same to me, is this implementation more accurate (due to integer maths?)

mnaamani:
I left some general comments in my last review. I think bedeho will still be in better position to review some of the test cases. I will run through full project-token implementation and perhaps do one more pass.

mnaamani:
I've ran weight generation and re-compiled benchmarks and ran benchmark tests successfully.

runtime-modules/project-token/src/lib.rs
dobertRowneySr: This will be taken care via proposal (next PR on the roadmap)

runtime-modules/project-token/src/tests/mock.rs
dobertRowneySr: No this is not, it's just patronage computation

runtime/src/utils.rs
dobertRowneySr: this was in order to silence clippy, I will revisit

runtime-modules/project-token/src/tests/patronage.rs
dobertRowneySr: fixed

runtime-modules/project-token/src/types.rs
dobertRowneySr: No, it should not. According to the trait

runtime-modules/project-token/src/benchmarking.rs
dobertRowneySr: because `DEFAULT_PATRONAGE: Permill` and the multiplaction is a non constant operation, so rustc will throw error

mnaamani:
@ignazio-bovo should we close this PR, it was merged into nara if I'm not mistaken?

PR #4752, opened 2023-05-09
[Argus] add auth options to elastic logs config

kdembler:
Allow adding authentication options to Elasticsearch logging so that secured clusters can be used as well.

### DIFF
--- a/distributor-node/src/app/index.ts
+++ b/distributor-node/src/app/index.ts
@@ -211,7 +211,7 @@ function hideConfigPath(config: object, path: string) {
         if (typeof pathValueEl === 'object') {
           return _.mapValues(pathValueEl, () => replaceValue)
         }
-        return '###SECRET###' as const
+        return replaceValue
       })
     } else {
       mappedValue = _.mapValues(pathValue, () => replaceValue)
--- a/distributor-node/src/app/index.ts
+++ b/distributor-node/src/app/index.ts
@@ -90,14 +90,11 @@ export class App {
     }
   }
 
-  private hideSecrets(config: Config) {
-    const displaySafeConfig: DisplaySafeConfig = {
-      ...config,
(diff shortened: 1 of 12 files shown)


distributor-node/src/app/index.ts
```diff
@@ -90,14 +90,11 @@ export class App {
...
 
-  private hideSecrets(config: Config) {
-    const displaySafeConfig: DisplaySafeConfig = {
-      ...config,
-      keys: config.keys?.map((k) => _.mapValues(k, () => '###SECRET###' as const)),
-      operatorApi: _.mapValues(config.operatorApi, () => '###SECRET###' as const),
-    }
-
-    return displaySafeConfig
+  private hideSecrets(config: Config): DisplaySafeConfig {
```
kdembler: I had to rework this functionality slightly since hiding value on a nested path was quite hard to do (and very unreadable). I'm open to alternatives if there are issues with the new approach

Lezek123: Since this new approach already sacrifaces the typesafety, as it's very difficult to maintain here, I decided to drop the `DisplaySafeConfig` type as well and replace it with a simpler `Record<string, unknown>` type.

I also reworked your `hideConfigPath` function to work recursively, as I think that makes it both clearer and more futureproof.

Here is my PR: https://github.com/kdembler/joystream/pull/1

Lezek123 (changes requested):
Thank you very much for this PR.
I tested the authentication and it works as expected.
I suggested some changes to the `hideConfigPath` functionality and mentioned in the reply above (https://github.com/kdembler/joystream/pull/1)

Since it's going to be a new release, there are also few additional changes needed:
- Please bump the version in `distributor-node/package.json` to `1.2.0`
- Please add new entry in `distributor-node/CHANGELOG.md`

Lezek123 (approved):
LGTM

PR #4758, opened 2023-05-15
[Argus] add workerId to status response

kdembler:
Sometimes operators will have `workerId` misconfigured which leads to issues, it'd be useful to just expose that config publicly for troubleshooting


Lezek123 (approved):
LGTM

PR #4759, opened 2023-05-15
[Argus] allow specifying elastic logs index

kdembler:
It may be useful in the future to create a separate index for each worker to better partition logs


Lezek123 (approved):
LGTM

Issue #4760, opened 2023-05-15
Argus trying to set headers after response is set


kdembler:
Seeing those in my logs from time to time:
```
2023-05-15 14:24:22:2422 PublicApi error: middlewareError
{
    "err": {
        "code": "ERR_HTTP_HEADERS_SENT",
        "message": "Cannot set headers after they are sent to the client",
        "stack": "Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client\n    at new NodeError (internal/errors.js:322:7)\n    at ServerResponse.setHeader (_http_outgoing.js:561:11)\n    at PublicApiController.assetHead (/joystream/distributor-node/src/services/httpApi/controllers/public.ts:224:9)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async /joystream/distributor-node/src/services/httpApi/HttpApiBase.ts:30:9"
    },
    "req": {
        "url": "/api/v1/assets/13702",
        "headers": {
            "host": "joystream.koalva.io",
            "user-agent": "undici",
            "accept": "*/*",
            "accept-encoding": "br, gzip, deflate",
            "accept-language": "*",
            "sec-fetch-mode": "cors",
            "x-forwarded-host": "joystream.koalva.io",
            "x-forwarded-proto": "https"
        },
        "method": "HEAD",
        "httpVersion": "1.1",
        "originalUrl": "/api/v1/assets/13702",
        "query": {}
    }
}
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at new NodeError (internal/errors.js:322:7)
    at ServerResponse.setHeader (_http_outgoing.js:561:11)
    at PublicApiController.assetHead (/joystream/distributor-node/src/services/httpApi/controllers/public.ts:224:9)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async /joystream/distributor-node/src/services/httpApi/HttpApiBase.ts:30:9
```

zeeshanakram3:
A couple of questions to further investigate this issue:
- Does this error happens only for `HEAD assets/` request
- What is the error code of the response

kdembler:
> Does this error happens only for HEAD assets/ request

No, I've seen this happen also with GET requests

> What is the error code of the response

I'm not sure, I think because of the error, the status code doesn't get logged

[...]

PR #4772, opened 2023-05-24
fix: properly handle prematurely closed requests

zeeshanakram3:
addresses #4760

# Problem
Sometimes a request can be prematurely closed by client or due to bad network conditions. Whenever that happens, an event handler on response object (`res`), manually [closes](https://github.com/Joystream/joystream/blob/master/distributor-node/src/services/httpApi/HttpApiBase.ts#LL25) the response by calling `res.end()`.  However, the problem is that even if the response closes the the control flow is passed the request handlers which then tries to set headers on a closed response, and hence the error.

# How to reproduce

In [routeWrapper](https://github.com/Joystream/joystream/blob/master/distributor-node/src/services/httpApi/HttpApiBase.ts#L22) function, manually close the response by calling `res.socket?.destory()`, 

```diff
  protected routeWrapper(handler: express.RequestHandler) {
    return async (req: express.Request, res: express.Response, next: express.NextFunction): Promise<void> => {
      // Fix for express-winston in order to also log prematurely closed requests
      
+     res.socket.?destroy()
      res.on('close', () => {
        res.locals.prematurelyClosed = !res.writableFinished
        res.end()
      })
      
+     await sleep(1000) // sleep for 1 second to make sure that request does not finish before socket destroy event is emitted
      try {
        await handler(req, res, next)
      } catch (err) {
        next(err)
      }
    }
  }

```

### DIFF
--- a/distributor-node/src/services/httpApi/HttpApiBase.ts
+++ b/distributor-node/src/services/httpApi/HttpApiBase.ts
@@ -23,17 +23,9 @@ export abstract class HttpApiBase {
       // Fix for express-winston in order to also log prematurely closed requests
       res.on('close', () => {
         res.locals.prematurelyClosed = !res.writableFinished
-        res.end()
       })
       try {
-        /**
-         * Call the request handler only if the response has not been sent yet (e.g. because req or res was
-         * closed prematurely). Otherwise, the request handler would try to set header or send the response
-         * again, which would result in an error e.g., "Cannot set headers after they are sent to the client"
-         * */
-        if (!res.headersSent) {
-          await handler(req, res, next)
-        }
+        await handler(req, res, next)
       } catch (err) {
         next(err)
       }
--- a/distributor-node/src/services/httpApi/HttpApiBase.ts
+++ b/distributor-node/src/services/httpApi/HttpApiBase.ts
@@ -1,12 +1,11 @@
+import cors from 'cors'
 import express from 'express'
 import * as OpenApiValidator from 'express-openapi-validator'
 import { HttpError, OpenApiValidatorOpts } from 'express-openapi-validator/dist/framework/types'
-import { ReadonlyConfig } from '../../types/config'
 import expressWinston from 'express-winston'
-import { Logger } from 'winston'
 import { Server } from 'http'
-import cors from 'cors'
-
+import { Logger } from 'winston'
+import { ReadonlyConfig } from '../../types/config'
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


distributor-node/src/services/httpApi/HttpApiBase.ts
```diff
@@ -27,7 +26,14 @@ export abstract class HttpApiBase {
         res.end()
       })
       try {
-        await handler(req, res, next)
+        /**
+         * Call the request handler only if the response has not been sent yet (e.g. because req or res was
+         * closed prematurely). Otherwise, the request handler would try to set header or send the response
+         * again, which would result in an error e.g., "Cannot set headers after they are sent to the client"
+         * */
+        if (!res.headersSent) {
```
mnaamani: This should not be necessary. My hunch is that the root cause of this behavior is the invocation of `res.end()` in the `close` event handler on line 26. It just looks wrong. I'll test and confirm

mnaamani: confirmed that by removing the `res.end()` and the newly added conditional on res.headerSent, invoking socket.destroy(), execution does not reach invokation of `await handler(req, res, next)`.
Also confirmed the log correctly includes the meta information for `prematurelyClosed`:

```
2023-06-12 18:57:17 2023-06-12 14:57:17:5717 PublicApi http: HTTP GET /api/v1/status
2023-06-12 18:57:17 {
2023-06-12 18:57:17     "meta": {
2023-06-12 18:57:17         "req": {
2023-06-12 18:57:17             "url": "/api/v1/status",
2023-06-12 18:57:17             "headers": {
2023-06-12 18:57:17                 "host": "localhost:3334",
2023-06-12 18:57:17                 "user-agent": "curl/8.1.2",
2023-06-12 18:57:17                 "accept": "*/*"
2023-06-12 18:57:17             },
2023-06-12 18:57:17             "method": "GET",
2023-06-12 18:57:17             "httpVersion": "1.1",
2023-06-12 18:57:17             "originalUrl": "/api/v1/status",
2023-06-12 18:57:17             "query": {}
2023-06-12 18:57:17         },
2023-06-12 18:57:17         "res": {
2023-06-12 18:57:17             "statusCode": 200
2023-06-12 18:57:17         },
2023-06-12 18:57:17         "responseTime": 60,
2023-06-12 18:57:17         "prematurelyClosed": true
2023-06-12 18:57:17     }
2023-06-12 18:57:17 }
```

mnaamani (changes requested):
A simpler solution is to just not call `res.end()` in the `closed` event handler.

mnaamani (approved):

PR #4776, opened 2023-05-31
[Colossus] add support for elasticsearch logs auth

kdembler:
Allows specifying Elasticsearch auth options for sending logs in Colossus, same was done for Argus: #4752

### DIFF
--- a/storage-node/src/commands/server.ts
+++ b/storage-node/src/commands/server.ts
@@ -75,20 +75,24 @@ export default class Server extends ApiCommandBase {
     elasticSearchEndpoint: flags.string({
       char: 'e',
       required: false,
+      env: 'ELASTIC_ENDPOINT',
       description: `Elasticsearch endpoint (e.g.: http://some.com:8081).
 Log level could be set using the ELASTIC_LOG_LEVEL enviroment variable.
 Supported values: warn, error, debug, info. Default:debug`,
     }),
     elasticSearchIndex: flags.string({
       required: false,
+      env: 'ELASTIC_INDEX',
       description: 'Elasticsearch index name.',
     }),
     elasticSearchUser: flags.string({
-      required: false,
+      dependsOn: ['elasticSearchEndpoint', 'elasticSearchPassword'],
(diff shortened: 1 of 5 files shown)


storage-node/src/commands/server.ts
```diff
@@ -79,6 +79,18 @@ export default class Server extends ApiCommandBase {
...
     }),
+    elasticSearchIndex: flags.string({
+      required: false,
+      description: 'Elasticsearch index name.',
+    }),
+    elasticSearchUser: flags.string({
+      required: false,
+      description: 'Elasticsearch user for basic authentication.',
+    }),
+    elasticSearchPassword: flags.string({
```
mnaamani, line 93: generally I don't like passing passwords as command line arguments, it is less secure (password would be available to see in shell history, process information, deployment scripts (although there are ways to secure secrets in pulumi deployment scripts for example).
But given that it is just for elastic search authentication I think its fine.

mnaamani (approved):

mnaamani:
Just an afterthought, is there a way to auto-generate docs for the new flags?

storage-node/src/commands/server.ts
mnaamani: We can easily add the `env` option to set a default value from the environment.
It might also make sense to use the `dependsOn` property instead of `required` on both `elasticSearchPassword` and `elasticSearchUser` to make them depend on each other.

see https://oclif.io/docs/flags

mnaamani (changes requested):
Just left a recommendation on flag options for the username and password.

storage-node/src/commands/server.ts
kdembler: Done in 139c5b700a99fc54e26f80a177fb9711db5d49f7

kdembler:
> Just an afterthought, is there a way to auto-generate docs for the new flags?

Regenerated docs and bumped version to `v3.3.0`

mnaamani (approved):

mnaamani (approved):

PR #4777, opened 2023-05-31
[Colossus] handle aborted uploads

### LINKED ISSUE #4773: [Colossus] Handle temp file cleanup on failed upload
It is observed that when storage node handling a file upload behind a reverse proxy with a max request_body size configured lower than the asset size, the syncing node doesn't appear to handle the network disconnection correctly and is not cleaning up temporary file.

This resulted in situation where several upload attempts exploded the size of the temp download folder or multiple nodes, until the remote nodes were reconfigured to increase the limit.

The fix for storage node should be able to handle the network disconnection correctly.

We should also investigate if a similar case might occur during sync operating if the reverse proxy applies any limit on the response size.

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4773

### DIFF
--- a/storage-node/CHANGELOG.md
+++ b/storage-node/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 3.3.0
 
 - Added customization options for Elasticsearch logging. Users can now specify index name and auth options.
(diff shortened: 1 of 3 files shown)


zeeshanakram3 (approved):
Tested this fix locally. LGTM

PR #4778, opened 2023-06-01
Update from node v14 to v18.6.0

mnaamani:
Updated all volta sections in repo package.json files to use node 18, as well is in github workflows and docker images.
I did not change the min version in engine sections to prevent breaking any deployments where volta is not used or forcing users to upgrade. But as long as volta is used, it will trigger it to update and use new version.


package.json
```diff
@@ -66,11 +66,11 @@
     }
   },
   "engines": {
-    "node": ">=14.0.0",
+    "node": ">=14.6.1",
```
zeeshanakram3: If Volta is not installed, then this engine requirement could break things. I got the following error while installing dependencies using node `14.7.0`. 

```
[1/6] Validating package.json...
[2/6] Resolving packages...
[3/6] Fetching packages...
error file-type@17.1.1: The engine "node" is incompatible with this module. Expected version "^12.20.0 || ^14.13.1 || >=16.0.0". Got "14.7.0"
error Found incompatible module.
```

So the engine min version needs to be increased. Things seem to work at the minimum of `14.18.0` (version previously defined in "volta" section)

README.md
```diff
@@ -12,7 +12,7 @@ functionality to support the [various roles](https://joystream.gitbook.io/testne
 The following tools are required for building, testing and contributing to this repo:
 
 - [Rust](https://www.rust-lang.org/tools/install) toolchain - _required_
-- [nodejs](https://nodejs.org/) v14.x - _required_
+- [nodejs](https://nodejs.org/) >= v18.6 - _required_
```
zeeshanakram3: shouldn't be `v14.x.x` i.e. the minimum node version defined in the engine section?

zeeshanakram3 (changes requested):

README.md
mnaamani: Will fix, I think I changed that originally when I was forcing all engine filed to be >=18

mnaamani:
@zeeshanakram3 Applied your feedback, I also update Readme to document the fact that old version of linux will not support node v18, so I include a workaround.

zeeshanakram3:
Looks good, I think we should use the same nodes engine requirement (`>=14.18.0`) in all package.json files in monorepo.

mnaamani:
> @zeeshanakram3 Applied your feedback, I also update Readme to document the fact that old version of linux will not support node v18, so I include a workaround.

done in [4a985be](https://github.com/Joystream/joystream/pull/4778/commits/4a985bebe3cfa1ea6c160385f0c51e75404c3c0c)

CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
+# Joystream Monorepo ChangeLog
+
+This is a top-level view of changes in the repo, linking to detailed changelogs of individual packages changed.
+If no packages or cargo crates are changed then the changes are usually "devops" realted.
```
zeeshanakram3:
```suggestion
If no packages or cargo crates are changed then the changes are usually "devops" related.
```

CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
...
+#### Moving to node v18.6.0 from v14
+This is a potentially breaking change for your setup, since it requires a minimum GLIBC_2.28 on gnu/linux systems. If you use Ubuntu then you should be at least on the `20.04` release.
+
+To check your glibc version use this command: `ldd --version`
+
+If your version is older than v2.28 you should really consider updating your OS it is probably quite dated and vulnerable to security bugs.
+
+The minimum required node engine for all monorepo packages is updated to v14.18.0
+
+No npm packages have not been published with new engine version yet, so you are not impacted if your application depedns on them, until the next package version release. To be ready do one of:
```
zeeshanakram3:
```suggestion
No npm packages have not been published with new engine version yet, so you are not impacted if your application depends on them, until the next package version release. To be ready do one of:
```

zeeshanakram3 (approved):
LGTM

Issue #4763, opened 2023-05-18
[Argus] better request tracing logs


kdembler:
Currently Argus logs all HTTP requests with some basic data associated with them, like request headers, result status code and also `responseTime` that is populated by logging middleware. However, it's quite unclear what `responseTime` really represents and it doesn't give us much info. There's many reasons why an Argus node can be slow to respond, for example:
- slow disk lookup
- slow QN query
- slow SP communication
If we're to be able to deal with those issues, we need more visibility into what is happening during request processing. I think some additional explicit (as in done by logic and not HTTP middleware) logging should be added and include various timing (mostly for points above) so distribution lead can collect that data and see what the issues are

kdembler:
Also would be good to log response headers, not only request, so things like Content-Size can be aggregated to get idea on throughput

bedeho:
@zeeshanakram3 would this be feasible ?

kdembler:
I'm currently using logs from `http` level that are implicitly logged by a middleware. This is the data that I'm getting:

<details>
<summary>Details</summary>
"fields": {
  "meta": {
    "res": {
      "statusCode": 200
    },
    "responseTime": 23,
    "prematurelyClosed": false,
    "req": {
      "headers": {
        "x-b3-parentspanid": "3ebf6731ec618482",
        "x-b3-spanid": "94c637a094bc18af",
        "x-b3-traceid": "ccb9c98a684f798a",
        "x-b3-sampled": "false",
        "host": "127.0.0.1:3334",
        "connection": "close",
        "x-forwarded-for": "199.16.157.181",
        "accept-encoding": "gzip, deflate",
        "accept": "*/*",
        "user-agent": "Twitterbot/1.0",
        "x-b3-flags": "17155602618"
      },
      "httpVersion": "1.0",
      "method": "GET",
      "query": {},
      "originalUrl": "/api/v1/assets/2418",
      "url": "/api/v1/assets/2418"
    }
  }
}

</details>

I would love to have the following data:
- asset ID
- status (missing/proxied/cached/etc)
- storage provider selected, if asset wasn't cached
- total bytes of content transferred
- total request processing time
- QN lookup time
- some details about storage provider performance if possible when proxying, ideally even bandwidth based on time and content size
- user-agent
- IP address
- status code

zeeshanakram3:
> @zeeshanakram3 would this be feasible ?

[...]

PR #4779, opened 2023-06-01
Origo: [Argus] Opentelemetry instrumentation for better metrics & tracing

zeeshanakram3:
Addresses #4763


zeeshanakram3:
# How to Test

- Setup Elastic search stack (Elasticsearch + Kibana + APM server). A utility script (`./start-elasticsearch-stack.sh`) has been provided. It will automatically bootstrap all the relevant services & set up all the necessary configurations.
- After bootstraping the ES Stack using `./start-elasticsearch-stack.sh`. go to Kibana dashboard at `http://localhost:5601/` and login with following credentials:
  - username=**_elastic_**
  - password=**_password_**
- Go to Kibana APM integration [endpoint](http://localhost:5601/app/integrations/detail/apm/overview), and click on `Add Elastic APM`. Then click on `Save and Continue`
- Next run any application instrumented with OpenTelemetry protocol. For reference, `distributor-1` service from the monorepo root `docker-compse.yml` file has been configured to export trace logs to the Elasticsearch server(see the configuration variables needed to be exported).
- After the application is up & running, go to [Traces](http://localhost:5601/app/apm/traces) endpoint in the Kibana dashboard, and you should be able to see the trace logs for different requests. e.g., below is a trace for `GET api/v1/assets/:id` request to Argus

image

zeeshanakram3:
@kdembler please test this PR based on the above [instructions](https://github.com/Joystream/joystream/pull/4779#issuecomment-1573127535), and let me know whether it meets your requirements for monitoring purposes. Also, feel free to ping me if there is any issue setting up the infra

zeeshanakram3:
@kdembler please test this PR based on the above [instructions](https://github.com/Joystream/joystream/pull/4779#issuecomment-1573127535), and let me know whether it meets your requirements for monitoring purposes. Also, feel free to ping me if there is any issue setting up the infra.

kdembler:
@zeeshanakram3 This looks _lovely_! I will be happy to test it out, but I will be unavailable next week, will be back on 12th and will check it out as soon as possible

mnaamani:
This looks like a really nice addition to provide visibility into how the application is performing.
I'm having issues however in getting the the distributor-1 to run successfully with instrumentation.

First observation, when starting the ES stack, the apm-server service name is coming up as `apm-server-1`

Secondly the distributor is crashing, with failure to validate the `JOYSTREAM_DISTRIBUTOR__OTLP__ENDPOINT` env variable.

I tried to change it from `http://apm-server:8200` to `"http://apm-server-1:8200"` (with quotes) but it still fails with:

```
2023-06-08 10:49:47 $ node --require ./lib/app/instrumentation.js ./bin/run start
2023-06-08 10:49:51 Starting tracing...
2023-06-08 10:49:51 /joystream/distributor-node/lib/services/parsers/ConfigParserService.js:96
2023-06-08 10:49:51                     throw e;
2023-06-08 10:49:51                     ^
2023-06-08 10:49:51 
2023-06-08 10:49:51 ValidationError: Invalid env value of JOYSTREAM_DISTRIBUTOR__OTLP__ENDPOINT
2023-06-08 10:49:51 
2023-06-08 10:49:51 
2023-06-08 10:49:51     at ConfigParserService.setConfigEnvValue (/joystream/distributor-node/lib/services/parsers/ConfigParserService.js:89:27)
2023-06-08 10:49:51     at /joystream/distributor-node/lib/services/parsers/ConfigParserService.js:109:18
2023-06-08 10:49:51     at Array.forEach (<anonymous>)
2023-06-08 10:49:51     at ConfigParserService.mergeEnvConfigWith (/joystream/distributor-node/lib/services/parsers/ConfigParserService.js:104:14)
2023-06-08 10:49:51     at ConfigParserService.parse (/joystream/distributor-node/lib/services/parsers/ConfigParserService.js:133:14)
2023-06-08 10:49:51     at Object.<anonymous> (/joystream/distributor-node/lib/app/instrumentation.js:27:31)
2023-06-08 10:49:51     at Module._compile (internal/modules/cjs/loader.js:1114:14)
2023-06-08 10:49:51     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
2023-06-08 10:49:51     at Module.load (internal/modules/cjs/loader.js:979:32)
2023-06-08 10:49:51     at Function.Module._load (internal/modules/cjs/loader.js:819:12) {
2023-06-08 10:49:51   errors: [],
2023-06-08 10:49:51   errorMessages: []
2023-06-08 10:49:51 }
2023-06-08 10:49:51 error Command failed with exit code 1.
```

This is happening on Mac Desktop Docker..

I'll try on Linux and report back. But I suspect the same is happening on Linux and that is why the Full scenario is failing integration tests at job `manage channels and videos through CLI: Failed`

mnaamani:
Even without starting with instrumentation, the same error is occuring:

```
2023-06-08 11:58:39 $ ./bin/run start
2023-06-08 11:57:34 
2023-06-08 11:57:34 
2023-06-08 11:57:34 error Command failed with exit code 1.
2023-06-08 11:57:45     Error: Invalid env value of JOYSTREAM_DISTRIBUTOR__OTLP__ENDPOINT
2023-06-08 11:57:45 
```

Is there something perhaps wrong in the changes in `configSchema.ts` ?

mnaamani:
Was able to resolve the problem by uncommenting the `otlp` section in the `distributor-node/config.yml`
I think this is because the properties are "required" so they must exist in the default config file, even if they are overriden by the env value.

mnaamani:
Reverted to use `http://apm-server:8200` endpoint, and was able to setup the APM integration and seeing 
Screenshot 2023-06-08 at 1 00 05 PM

mnaamani:
Interestingly enough, the tracing showed a potential problem with reading the package.json file in which the only code in distributor-node that does this is:

`ConfigParserService.ts`

```
public getNodeVersion(): string {
    const packageJSON = JSON.parse(fs.readFileSync(path.join(__dirname, '../../../package.json')).toString())
    return String(packageJSON.version)
  }
```

Screenshot 2023-06-08 at 1 04 05 PM

distributor-node/config.yml
```diff
@@ -19,6 +19,9 @@ logs:
   #   auth:
   #     username: username
   #     password: password
+# otlp:
```
mnaamani: This needs to be uncommented as the properties are "required".
Otherwise change them to be optional and leave them commented out.
If node is started with instrumentation and endpoint/attributes not provided, exist with error?

mnaamani (changes requested):

distributor-node/config.yml
zeeshanakram3: Actually, the `otlp` properties are not required in the config schema. You can see [here](https://github.com/zeeshanakram3/joystream/blob/8c0494ede006d155f91dad8e3424efd33f0918a9/distributor-node/src/schemas/configSchema.ts#L20). Also, the `distributor-1` service works fine if I remove the `JOYSTREAM_DISTRIBUTOR__OTLP__ENDPOINT`, `JOYSTREAM_DISTRIBUTOR__OTLP__ATTRIBUTES` env vars from the service in docker-compose.yml file, which also enforces the point that `otlp` properties are not required. I think there is a problem with the `ValidationService` class. I tried to debug the problem by adding `console.log(this.avj.errors)` at [line](https://github.com/zeeshanakram3/joystream/blob/8c0494ede006d155f91dad8e3424efd33f0918a9/distributor-node/src/services/validation/ValidationService.ts#L34), and got this:
```
[
  {
    keyword: 'required',
    dataPath: '/otlp',
    schemaPath: '#/properties/otlp/required',
    params: { missingProperty: 'attributes' },
    message: "should have required property 'attributes'"
  }
]
/joystream/distributor-node/lib/services/parsers/ConfigParserService.js:97
                    throw e;
                    ^

ValidationError: Invalid env value of JOYSTREAM_DISTRIBUTOR__OTLP__ENDPOINT
```

zeeshanakram3: Just created an issue for this. Will be investigating it https://github.com/Joystream/joystream/issues/4787

mnaamani: okay, so how do we interpret the line `required: ['endpoint', 'attributes'],` in [configSchema.ts](https://github.com/zeeshanakram3/joystream/blob/8c0494ede006d155f91dad8e3424efd33f0918a9/distributor-node/src/schemas/configSchema.ts#L131)

zeeshanakram3: > okay, so how do we interpret the line `required: ['endpoint', 'attributes'],` in [configSchema.ts](https://github.com/zeeshanakram3/joystream/blob/8c0494ede006d155f91dad8e3424efd33f0918a9/distributor-node/src/schemas/configSchema.ts#L131)

This means `otlp` is an optional object in configSchema. However, when it is present (i.e., not `undefined`),  it `must` have `endpoint` & `attributes` properties. 

FYI I created a PR to address this issue, please have a look #4788

mnaamani: Merged the fix, perhaps you want to update this from master

zeeshanakram3: Done.

distributor-node/src/app/instrumentation.ts
```diff
@@ -0,0 +1,83 @@
...
+import pkgDir from 'pkg-dir'
+import { ConfigParserService } from '../services/parsers/ConfigParserService'
+import { ReadonlyConfig } from '../types'
+
+// For troubleshooting, set the log level to DiagLogLevel.DEBUG
+diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.INFO)
+
+diag.info('Starting tracing...')
+
+// Default config JSON/YAML file path (relative to current working directory)
```
mnaamani: I'm glad we discovered issue with parsing the config because of adding the new parameters in the config file, however I feel perhaps we can just rely on `OTEL_`... env variables, since they are not really configuration parameters for the distributor-node itself. And by dropping this coupling, where we are importing from the 

```
import { ConfigParserService } from '../services/parsers/ConfigParserService'
import { ReadonlyConfig } from '../types'
```

It becomes possible to break this code out of the distributor package into its own package that can be re-used? 

What are your thoughts. I'm not familiar enough with Open Telemetry architecture yet to have a very strong opinion on this.

mnaamani (commented):
Looks and works as expected.
Just left some comments about possible generalization and decoupling the instrumentation config from the main application.

kdembler:
Okay it's been a while, really sorry for the delay. I'm starting testing it out right now

distributor-node/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### 1.2.0
 
+- Integrates OpenTelemetry API/SDK with Argus for exporting improved tracing logs & metrics to Elasticsearch. Adds `./start-elasticsearch-stack.sh` script to bootstrap elasticsearch services (Elasticsearch + Kibana + APM Server) with all the required configurations.
```
kdembler: Since 1.2.0 has been already released and is in use, let's bump the version please

kdembler:
Also seeing `ENOENT: no such file or directory, open '/joystream/distributor-node/src/commands/package.json'`

zeeshanakram3:
> It becomes possible to break this code out of the distributor package into its own package that can be re-used?

- I have created `@joystream/opentelemetry` package in the monorepo that can be used to add opentelemetry tracing capabilities to any arbitrary nodes & services
- Now all the OpenTelemetry code is Orthogonal to the Application packages and can be imported as a dependency
- I tested the storage-node, distributed-node & query-node docker containers with open telemetry integration using the `colossos-1`, `distributor-1` & `graphql-server` services respectively in the `docker-compose.yml`, and it seemed to work as expected
- Building & running docker images of Applications (with OTEL integration) requires publishing of `@joystream/opentelemetry` on npm (so could not test that)

cc. @mnaamani

zeeshanakram3:
@mnaamani I have updated the dockerfiles to install `@josyteam/opentelemtry` in images during build step in [d3a473f](https://github.com/Joystream/joystream/pull/4779/commits/d3a473f0ae07d40e6d6e8b5159b20d3ac57056ff)

colossus.Dockerfile
```diff
@@ -37,5 +40,5 @@ VOLUME ["/data", "/keystore", "/logs"]
 # Colossus node port
 EXPOSE 3333
 
-ENTRYPOINT ["yarn", "storage-node"]
-CMD ["server"]
\ No newline at end of file
+ENTRYPOINT ["yarn"]
+CMD ["start"]
```
mnaamani: I would prefer to keep the previous entrypoint and cmd, keeping in mind that we also want to run other commands. Better to override them in the docker-compose files.

colossus.Dockerfile
```diff
@@ -19,10 +20,12 @@ COPY ./metadata-protobuf metadata-protobuf
 COPY ./devops/eslint-config ./devops/eslint-config
 COPY ./devops/prettier-config ./devops/prettier-config
 COPY ./storage-node storage-node
+COPY ./opentelemetry opentelemetry
 COPY ./tsconfig.json ./tsconfig.json
 
 RUN yarn workspace @joystream/types build
 RUN yarn workspace @joystream/metadata-protobuf build
+RUN yarn workspace @joystream/opentelemetry build
```
mnaamani, line 28: We also need to add `yarn workspace @joystream/opentelemetry build` to `build-packages.sh` please.

distributor-node.Dockerfile
```diff
@@ -40,5 +43,5 @@ RUN yarn --frozen-lockfile --production
 
 EXPOSE 3334
 
-ENTRYPOINT ["yarn", "joystream-distributor"]
+ENTRYPOINT ["yarn"]
```
mnaamani: Please keep the same entry point for same reason as other Dockerfile, and make necessary change in the compose file instead.

distributor-node/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
...
```
mnaamani, line 3: I think we need to bump version to `1.2.1` so we can publish a new docker image with a new tag.

query-node/build.sh
```diff
@@ -34,3 +34,14 @@ yarn workspace query-node-mappings build
 
 cp ./generated/graphql-server/generated/schema.graphql ../storage-node/src/services/queryNode/schema.graphql
 cp ./generated/graphql-server/generated/schema.graphql ../distributor-node/src/services/networking/query-node/schema.graphql
+
+################################################
+# temporary patche TODO: create proper solution
+
```
mnaamani, line 40: Some env variables i think might be getting lost, there is a specific one i noticed which is set at top .env `PROCESSOR_HOST=processor` which is the host/ip the graphql server expects from the processor when it "pings" it to update.  So the graphql-server is constantly logging:


```
2023-07-12 19:30:30 [
2023-07-12 19:30:30   'Unauthorized access on /update-processor-state: 172.18.0.10 (expected: undefined)'
2023-07-12 19:30:30 ]
```

start-elasticsearch-stack.sh
```diff
@@ -0,0 +1,42 @@
...
+# Run docker-compose to start elasticsearch container
+docker-compose -f ./docker-compose.elasticsearch.yml up -d elasticsearch
+
+echo 'Waiting for Elasticsearch...'
+
+sleep 30
+
+# Generate the service token
+# Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/service-accounts.html#service-accounts-tokens
+response=$(curl -f -s -X POST -u "${ELASTIC_USERNAME}":"${ELASTIC_PASSWORD}" "http://localhost:9200/_security/service/elastic/kibana/credential/token/my_kibana_token" -H 'Content-Type: application/json')
```
mnaamani: Might be nice to actually spit out error when this fails to fetch token, I had issue running elasticsearch on mac docker desktop (ran out of space so cluster was failing) so the HTTP response was coming back with status code `503`, and took me a while to figure out why kibana was not starting.

opentelemetry/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      } else {
+        console.log('APM server may not be running correctly.')
+      }
+    } catch (error: any) {
+      console.error(`Failed to reach APM server, attempt ${attempts}:`, error.message)
+      if (attempts < retryAttempts) {
+        await wait(retryInterval)
+      } else {
+        console.log(`Failed to reach APM server after ${attempts} attempts.`)
+        process.exit(1)
```
mnaamani: I think we should drop waiting for APM to be reachable entirely, to avoid potentially restarting loop if it ever restart or goes down. Worst case we will see logs indicating otl export failure and queue will get filled.

opentelemetry/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      diag.info(`Loaded Application Instrumentation: "Storage Node"`)
+    } else if (applicationName === 'query-node') {
+      instrumentation = DefaultInstrumentation
+      diag.info(`Loaded Application Instrumentation: "Query Node"`)
+    } else {
+      diag.error(
+        `Opentelemetry instrumentation for package/application "${applicationName}" does not exist` +
+          ` Add instrumentation for package, or try to run application with default instrumentation`
+      )
+      process.exit(1)
```
mnaamani: Should we warn and just use default instrumentation instead of exiting. Maybe operator misconfiguration the application name in their env settings.

mnaamani (changes requested):
Nice work. Left a few suggested changes, otherwise it works great and seeing nice traces in APM down to the SQL queries.

colossus.Dockerfile
zeeshanakram3: Addressed

opentelemetry/index.ts
zeeshanakram3: Addressed by adding warning.

opentelemetry/index.ts
zeeshanakram3: Addressed

start-elasticsearch-stack.sh
zeeshanakram3: Addressed in [b9aa169](https://github.com/Joystream/joystream/pull/4779/commits/b9aa16950e48965047323317835dc32263404577)

query-node/build.sh
zeeshanakram3: Ok, so it took a while to figure this out.

The `PROCESSOR_HOST` env var is available inside the `graphql-server`, how the problem is happening while dns [lookup](https://github.com/Joystream/hydra/blob/db6b2d4b611bf76de38160c1d800648f85e8f3de/packages/hydra-cli/src/templates/graphql-server/src/processorStateApp.ts.mst#L65), if the `graphql-server` is running with open telemetry instrumentation enabled, then the `lookup` function returns an address string (response type of callback-based function) instead of expected [LookupAddress](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/dns.d.ts#L72) object (response type of promise-based function). So, in short, open telemetry isn't applying a proper patch of the lookup function.

There are two possible solutions I can think of
- Don't  instrument the `dns` package in the `graphql-server` (a much simpler solution)
- Make a change in the `graphql-server` codebase to use `callback` based or `dns/promise` based `lookup` method instead of [promisify](https://github.com/Joystream/hydra/blob/master/packages/hydra-cli/src/templates/graphql-server/src/processorStateApp.ts.mst#L59) lookup function

mnaamani: > * Don't  instrument the `dns` package in the `graphql-server` (a much simpler solution)

I think that is the best option to go with.  :+1:

zeeshanakram3: Addressed in [7c658a8](https://github.com/Joystream/joystream/pull/4779/commits/7c658a8000410da69fdc614458a33a951385d6e9)

opentelemetry/instrumentations/default.ts
```diff
@@ -15,5 +15,10 @@ export const DefaultInstrumentation = new NodeSDK({
   metricReader: new PeriodicExportingMetricReader({
     exporter: new OTLPMetricExporter(),
   }),
-  instrumentations: [getNodeAutoInstrumentations()],
+  instrumentations: [
+    // Disable DNS instrumentation, because the instrumentation does not correctly patches `dns.lookup` function
+    // if the function is converted to a promise-based method using `utils.promisify(dns.lookup)`
+    // See: https://github.com/Joystream/joystream/pull/4779#discussion_r1262515887
+    getNodeAutoInstrumentations({ '@opentelemetry/instrumentation-dns': { enabled: false } }),
```
mnaamani, line 22: This also fixed a problem I noticed with storage and distributor nodes, not finishing reponse on their status endpoint `/api/v1/status`

mnaamani (changes requested):
A few minor fixes, I prepared them in a PR https://github.com/zeeshanakram3/joystream/pull/6

mnaamani (approved):

PR #4786, opened 2023-06-06
fix joystream-node docker build workflow

mnaamani:
Follow up to https://github.com/Joystream/joystream/pull/4770 which fixes workflow when pushing to master: https://github.com/Joystream/joystream/actions/runs/5146012965/jobs/9264471091

With the fix for installing rustup, the workflow is always executing the build because the `docker manifest inspect ..` command to check if a build exists on dockerhub was failing not because image was not found but because of auth failure. eg. https://github.com/Joystream/joystream/actions/runs/5179948547/jobs/9333504405#step:6:9


dobertRowneySr (approved):
LGTM

Issue #4787, opened 2023-06-08
Argus: ConfigParserService expects value to be provided even for optional property


zeeshanakram3:
see https://github.com/Joystream/joystream/pull/4779#discussion_r1223382263

zeeshanakram3:
Addressed in https://github.com/Joystream/joystream/pull/4788

PR #4788, opened 2023-06-08
[Argus] Fix config parsing bug

zeeshanakram3:
addresses #4787 

- Addresses `#4787` in 695c66a0b290bbd911e0c8ce988e7dd953919d6b. The fix allows omitting optional config keys from `config.yml` file when specifying their values as env vars.
- [add uri validation for endpoint properties in schema](https://github.com/Joystream/joystream/commit/c0be3e97a3707a363029af2edea780b80e65fac5)
- [bump avj version number](https://github.com/Joystream/joystream/commit/d27733eea42aa642db29c7162f319f23413e211d)


mnaamani:
Would merging https://github.com/Joystream/joystream/pull/4759 without this fix be problematic?

distributor-node/src/services/validation/ValidationService.ts
```diff
@@ -7,7 +8,7 @@ export class ValidationError extends Error {
 
   public constructor(message: string, errors: Ajv['errors']) {
     const errorMessages: string[] = []
-    errors?.forEach((e) => errorMessages.push(`${e.dataPath}: ${e.message} (${JSON.stringify(e.params)})`))
+    errors?.forEach((e) => errorMessages.push(`${e.instancePath}: ${e.message} (${JSON.stringify(e.params)})`))
```
mnaamani, line 11: I see this change from dataPath to instancePath is due to newer version of ajv package. https://ajv.js.org/api.html#validation-errors

But storage-node also uses ajv for schema validation so we may need to do a similar fix there

zeeshanakram3:
> Would merging #4759 without this fix be problematic?

The error fixed in this PR occurred if we didn't provide the required property of an optional object in config.yml. In #4759, a new property `logs.elastic.index` is being added, which is already optional. So I think we won't have an issue if we merge #4759 without this fix.

distributor-node/src/services/validation/ValidationService.ts
zeeshanakram3: Both storage-node & distributor-node have in-package import of `ajv` dependency from `storage-node/node_modules` & `distributor-node/node_modules/` as it's added separately in ` storage-node/package.json` & `distributor-node/package.json`.

 - However, I have updated `ajv` dependency to `^8..0.0` in `/storage-node` & `/cli`
 - Removed `ajv` dependency from `/types`, as it's unused

mnaamani (approved):
LGTM

Issue #4765, opened 2023-05-18
Add support to mark storage-node/distrubtion-node as under maintenance by the Node operators


zeeshanakram3:
## Context 

When Storage & Distribution node operators put their nodes down for maintenance or upgrade purposes. They should be able to publish that action on the chain so that Leads/Apps are aware of the node status.

bedeho:
Here are some loose requirements I have from a conversation with a lead

- Both lead and worker, in each case, should be able to signal a given host as being unavailable for use. If a given worker has multiple distinct hosts for the same role, then presumably the flag should be apply to a given host, not just the role, but the leads perhaps should be consulted on this. 
- New state is introduced for each host: a status `operational_status` with value range (x,y, use epoch time, not blocks)
  - `normal`: all APIS work, as now.
  - `no_service(bool forced)`: no APIs should be expected to work. A more granular representation could be picked here, but seems premature at this point.
  - `no_service_from(bool forced,x)`: normal mode before time `x`, `no_service` after.
  - `no_service_during(bool forced, x,y)`: normal more before and after `x` and `y` time, `no_service` after.
- The `forced` indicator means the state was entered into by lead, and it is meant to prevent worker from unilaterally reversing.
- Leads can set any host to any new status value at all times using extrinsic with a message over working group lead remark for given corresponding group, and message should include field for free text message explaining background for state change.
- Workers can set their own host status in the corresponding way under the following more restrictive rules
  - any unforced non-service state when currently either normal or in an unforced non-service state.
  - normal state if they are in an unforced non-service state.

I think a reasonable next steps are to

1. Prepare a detailed spec.
2. Have both leads review: but avoid feature bloat!
3. Prepare implementation plan with details about what pieces of software must be updated, what tests should be updated, and how to deploy safely.

bedeho:
# Update

Seems I missed that there are indeed distinct remarks for storage operators and distributor operators, so perhaps those should be used instead, rather than using the generic working group extriniscs, as means the mappings should be a bit cleaner and simpler, however I am not sure the leads could be tackled this way, some more careful examination is required.

zeeshanakram3:
Here is a feature specification highlighting all the required QN, metaprotocol & protobuf changes 

## QN Changes

Following schema changes would be added to query-node's [storage](https://github.com/zeeshanakram3/joystream/blob/2a94e3bb640fac13f16f005531f28f7d27fa127e/query-node/schemas/storage.graphql) related entities

```.gql
### New Variants

[...]

PR #4793, opened 2023-06-21
[Origo] Adds support to mark storage-node/distrubtion-node under maintenance

zeeshanakram3:
Addresses #4765 

- [x]  Adds Storage/Distribution QN mappings
- [x] Adds `operator:set-metadata` & `leader:set-node-operational-status` CLI commands
- [x] ~WIP: Updating Storage/Distribution related integration tests~ https://github.com/Joystream/joystream/issues/4815


mnaamani:
You need to add implicitly new dependency on inquirer-datepicker? (and maybe others) to storage node: this is why this check is failing building docker image for storage node https://github.com/Joystream/joystream/actions/runs/5561050023/jobs/10158465896?pr=4793#step:8:373

zeeshanakram3:
Integration tests will be added in #4815

mnaamani:
@zeeshan perhaps we can try to bump version of `protobufjs` in this PR?

see https://github.com/advisories/GHSA-h755-8qp9-cq85

zeeshanakram3:
> perhaps we can try to bump version of `protobufjs` in this PR?

Bumped package versions and updated CHANGELOG.md in [d133ffb](https://github.com/Joystream/joystream/pull/4793/commits/d133ffbf1f3265f560d5aaece632820c08c21e0f)

mnaamani:
> > perhaps we can try to bump version of `protobufjs` in this PR?
> 
> Bumped package versions and updated CHANGELOG.md in [d133ffb](https://github.com/Joystream/joystream/pull/4793/commits/d133ffbf1f3265f560d5aaece632820c08c21e0f)

Great.
I think QN package also needs a version bump.

zeeshanakram3:
> I think QN package also needs a version bump.

Done in [ba24cfe](https://github.com/Joystream/joystream/pull/4793/commits/ba24cfe683ee6c29f3f5ba43c9152d4736a4f50a)

metadata-protobuf/proto/Storage.proto
```diff
@@ -46,3 +48,31 @@ message DistributionBucketFamilyMetadata {
...
+  enum OperationalStatus {
+    // Node is operating normally
+    NORMAL = 0;
+
+    // Node is not operational 
+    NO_SERVICE = 1;
+  }
+
+  // Node's Operational status to set
+  optional OperationalStatus status = 1;
```
kdembler: Shouldn't the `status` be required here? Full `NodeOperationalStatusMetadata` is already optional in `DistributionBucketOperatorMetadata`

query-node/mappings/src/storage/index.ts
```diff
@@ -585,3 +599,89 @@ async function removeDistributionBucketOperator(store: DatabaseManager, operator
...
+  if (workingGroup.name !== 'distributionWorkingGroup' && workingGroup.name !== 'storageWorkingGroup') {
+    return invalidMetadata(`The ${workingGroup.name} is incompatible with the remarked setNodeOperationalStatus`)
+  }
+
+  const workerId = Number(meta.workerId)
+  const bucketId = String(meta.bucketId)
+
+  if ((await getWorker(store, workingGroup.name, workerId)) === undefined) {
+    return invalidMetadata(`The worker ${workerId} does not exist in the ${workingGroup.name} working group`)
+  }
```
kdembler: I'm wondering, do we need to bother with worker for storage buckets? Those only have a single operator so I guess the worker ID doesn't make any difference?

query-node/mappings/src/storage/metadata.ts
```diff
@@ -66,6 +73,52 @@ async function processNodeLocationMetadata(
...
+  if (isCurrentForced && actorContext === 'worker') {
+    return current
+  }
+
+  // set node state to NoService
+  if (meta.status === NodeOperationalStatusMetadata.OperationalStatus.NO_SERVICE) {
+    if (meta.noServiceFrom && meta.noServiceTo) {
+      const status = new NodeOperationalStatusNoServiceDuring()
+      status.from = new Date(meta.noServiceFrom)
+      status.to = new Date(meta.noServiceTo)
```
kdembler: Are those raw dates validated anywhere? If not, we probably should do that

query-node/schemas/storageEvents.graphql
```diff
@@ -0,0 +1,53 @@
...
+  "Network the block was produced in"
+  network: Network!
+
+  "Index of event in block from which it was emitted."
+  indexInBlock: Int!
+
+  ### SPECIFIC DATA ###
+
+  "Storage Bucket"
+  storageBucket: StorageBucket
```
kdembler: Shouldn't that be a required field?

storage-node/src/commands/leader/set-node-operational-status.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  static flags = {
+    bucketId: flags.integer({
+      char: 'i',
+      required: true,
+      description: 'Storage bucket ID',
+    }),
+    workerId: flags.integer({
+      char: 'w',
+      description: 'ID of the operator (distribution group worker)',
```
kdembler:
```suggestion
      description: 'ID of the operator (storage group worker)',
```

storage-node/src/commands/operator/set-metadata.ts
```diff
@@ -45,16 +57,55 @@ export default class OperatorSetMetadata extends ApiCommandBase {
...
+        },
+      }
+    } else {
+      metadata = { endpoint, operationalStatus }
+    }
 
     const encodedMetadata = '0x' + Buffer.from(StorageBucketOperatorMetadata.encode(metadata).finish()).toString('hex')
 
-    logger.info('Setting the storage operator metadata...')
+    logger.info('Se tting the storage node operational status...')
```
kdembler:
```suggestion
    logger.info('Setting the storage node operational status...')
```

kdembler: Also, I think this still should be the old string since the command is also used for regular metadata?

storage-node/src/commands/operator/set-metadata.ts
```diff
@@ -45,16 +57,55 @@ export default class OperatorSetMetadata extends ApiCommandBase {
...
+      metadata = {
+        ...input,
+        operationalStatus: {
+          ...input.operationalStatus,
+          status:
+            input.operationalStatus?.status === 'Normal'
+              ? NodeOperationalStatusMetadata.OperationalStatus.NORMAL
+              : NodeOperationalStatusMetadata.OperationalStatus.NO_SERVICE,
+        },
+      }
```
kdembler: I think if somebody will try to set metadata without operational status then this will set it to `NO_SERVICE`?

distributor-node/src/commands/operator/set-metadata.ts
```diff
@@ -31,18 +43,58 @@ export default class OperatorSetMetadata extends AccountsCommandBase {
...
+      const params = validation.validate('OperatorMetadata', JSON.parse(fs.readFileSync(input).toString()))
+      metadata = {
+        ...params,
+        operationalStatus: {
+          ...params.operationalStatus,
+          status:
+            params.operationalStatus?.status === 'Normal'
+              ? NodeOperationalStatusMetadata.OperationalStatus.NORMAL
+              : NodeOperationalStatusMetadata.OperationalStatus.NO_SERVICE,
+        },
```
kdembler: I think if somebody will try to set metadata without operational status then this will set it to `NO_SERVICE`?

storage-node/src/services/queryNode/schema.graphql
```diff
@@ -957,6 +962,9 @@ type DistributionBucketOperatorMetadata implements BaseGraphQLObject {
   nodeLocation: NodeLocationMetadata
   nodeLocationId: String
 
+  """Optional node operational status"""
+  nodeOperationalStatus: NodeOperationalStatus
```
kdembler: I think this can be hard to query because of QN nested filtering limitations. I think we need to design a schema in a way that I can easily fetch all the active operators. If I'm not mistaken with the current schema the query would have to be (paraphrasing, I don't remember the query syntax for checking specific union type):
```gql
query {
  distributionBucketOperators(
    where: { metadata: { nodeOperationalStatus: { status: "OKAY" } } }
  ) {
    id
  }
}
```
And that filter would be too deep for QN to execute. Same applies to storage buckets

kdembler: Also, would it be possible to make operational status required part of metadata? Because empty operational status currently would also mean "Okay" but it would be hard to query it I think

kdembler (changes requested):
Great work @zeeshanakram3 and sorry for doing this review so late 🙇 I think one additional feature we could introduce is for distribution nodes to only query active storage operators when doing pings/asset checks. But we can do that as a followup later, this in itself is quite valuable.
Edit: I see that issue for that is already created and scoped: https://github.com/Joystream/joystream/issues/4816

query-node/mappings/src/storage/metadata.ts
zeeshanakram3: Addressed this in storage squid PR https://github.com/Joystream/storage-squid/pull/19

distributor-node/src/commands/operator/set-metadata.ts
zeeshanakram3: Addressed in [6ff4531](https://github.com/Joystream/joystream/pull/4793/commits/6ff45319432ff711ce35e281f01b0fba889089a3)

storage-node/src/commands/operator/set-metadata.ts
zeeshanakram3: Addressed in [6ff4531](https://github.com/Joystream/joystream/pull/4793/commits/6ff45319432ff711ce35e281f01b0fba889089a3)

storage-node/src/services/queryNode/schema.graphql
zeeshanakram3: I think you would not have that problem with storage squid's graphql server

zeeshanakram3:
@kdembler I have addressed the requested changes. Also, removed the QN mappings needed for this feature as I think having the code at two places makes it difficult to review and maintain

Also, did changes as described in #4816 for both Argus/Colossus

distributor-node/src/commands/leader/set-node-operational-status.ts
```diff
@@ -0,0 +1,81 @@
...
+    }),
+    ...DefaultCommandBase.flags,
+  }
+
+  async run(): Promise<void> {
+    const { bucketId, workerId, operationalStatus: statusType } = this.parse(LeadSetNodeOperationalStatus).flags
+    const leadKey = await this.getDistributorLeadKey()
+
+    let operationalStatus: INodeOperationalStatusMetadata = {}
+    switch (statusType) {
```
kdembler, line 45: `statusType` input is not required, should we default to `Normal`?

distributor-node/src/commands/operator/set-metadata.ts
```diff
@@ -31,18 +43,60 @@ export default class OperatorSetMetadata extends AccountsCommandBase {
   }
 
   async run(): Promise<void> {
-    const { bucketId, workerId, input, endpoint } = this.parse(OperatorSetMetadata).flags
+    const { bucketId, workerId, input, endpoint, operationalStatus: statusType } = this.parse(OperatorSetMetadata).flags
     const workerKey = await this.getDistributorWorkerRoleKey(workerId)
 
+    let operationalStatus: INodeOperationalStatusMetadata = {}
+    switch (statusType) {
```
kdembler, line 49: Same question about default status here

distributor-node/src/commands/operator/set-metadata.ts
```diff
@@ -31,18 +43,60 @@ export default class OperatorSetMetadata extends AccountsCommandBase {
...
+        ...params,
+        operationalStatus: params.operationalStatus
+          ? {
+              ...params.operationalStatus,
+              status:
+                params.operationalStatus?.status === 'Normal'
+                  ? NodeOperationalStatusMetadata.OperationalStatus.NORMAL
+                  : NodeOperationalStatusMetadata.OperationalStatus.NO_SERVICE,
+            }
+          : {},
```
kdembler: Would this override the current status? Or skip the update?

distributor-node/src/services/networking/query-node/queries/queries.graphql
```diff
@@ -107,7 +107,17 @@ fragment StorageBucketOperatorFields on StorageBucket {
...
 
 query getActiveStorageBucketOperatorsData {
-  storageBuckets(where: { operatorStatus: { isTypeOf_eq: "StorageBucketOperatorStatusActive" } }) {
+  storageBuckets(
+    where: {
+      operatorStatus: { isTypeOf_eq: "StorageBucketOperatorStatusActive" }
+      operatorMetadata: {
+        OR: [
+          { nodeOperationalStatus_isNull: true }
+          { nodeOperationalStatus: { isTypeOf_eq: "NodeOperationalStatusNormal" } }
```
kdembler: This filter would skip any nodes with future maintenance as well. We probably should get all the operators and later only filter those that are under maintenance **currently**

metadata-protobuf/proto/Storage.proto
kdembler: @zeeshanakram3 still relevant

storage-node/src/services/queryNode/queries/queries.graphql
```diff
@@ -3,19 +3,62 @@ fragment StorageBucketIds on StorageBucket {
...
   storageBuckets(
-    where: { operatorStatus: { isTypeOf_eq: "StorageBucketOperatorStatusActive", workerId_eq: $workerId } }
+    where: {
+      operatorStatus: { isTypeOf_eq: "StorageBucketOperatorStatusActive", workerId_eq: $workerId }
+      operatorMetadata: {
+        OR: [
+          { nodeOperationalStatus_isNull: true }
+          { nodeOperationalStatus: { isTypeOf_eq: "NodeOperationalStatusNormal" } }
+        ]
+      }
```
kdembler: Same comment here as with the query in distribution node

kdembler (changes requested):

distributor-node/src/commands/leader/set-node-operational-status.ts
mnaamani: It makes more sense to make it required, the command name doesn't imply what the default value should be.

metadata-protobuf/proto/Storage.proto
```diff
@@ -46,3 +48,31 @@ message DistributionBucketFamilyMetadata {
...
+  // Date (ISO format) until which the node won't be operational (Should be set if status is NoServiceDuring)
+  optional string no_service_to = 3; 
+
+  // Rationale for setting the current status
+  optional string rationale = 4; 
+}
+
+message SetNodeOperationalStatus {
+  optional NodeOperationalStatusMetadata operational_status = 1; // Node's operational status to set
+  optional string worker_id = 2; // Storage/Distribution Worker ID
```
mnaamani: I would argue only the `bucket_id` is strictly necessary in the message.

distributor-node/src/commands/leader/set-node-operational-status.ts
```diff
@@ -0,0 +1,81 @@
...
+import { NODE_OPERATIONAL_STATUS_OPTIONS, NodeOperationalStatus } from '../../types/metadata'
+
+export default class LeadSetNodeOperationalStatus extends AccountsCommandBase {
+  static description = `Set/update distribution node operational status. Requires distribution working group leader permissions.`
+
+  static flags = {
+    bucketId: flags.bucketId({
+      required: true,
+    }),
+    workerId: flags.integer({
```
mnaamani, line 17: I suggested that worker id is not needed in metadata. If we agree on that, then this workerId can be dropped.

metadata-protobuf/proto/Storage.proto
mnaamani: I would go further and suggest that we should probably have multiple messages, corresponding with the options in the commands for setting the metadata.

```
message NodeOperationalStatusNormal {
  optional string rationale = 1;
  required string bucketId = 2;
}
```

```
message NodeOperationalStatusNoService {
   optional string rationale = 1;
   required string bucketId = 2;
}
```

```
NodeOperationalStatusNoServiceFrom {
  optional string rationale = 1;
  required string from = 2;  // date
  required string bucketId = 3;
}
```


```
NodeOperationalStatusNoServiceUntil {
  optional string rationale = 1;
  optional string from = 2  // date
  required string until = 3;  // date
  required string bucketId = 4;
}
```

Doing this I think will help address a couple issues raised in other places in the review as well as simplifying implementation in the commands and squid mappings I presume.

distributor-node/src/command-base/default.ts
```diff
@@ -98,6 +100,21 @@ export default abstract class DefaultCommandBase extends Command {
...
   }
 
+  async datePrompt(question: DistinctQuestion): Promise<Date> {
+    const { result } = await inquirer.prompt([
+      {
+        ...question,
+        type: 'datepicker',
+        name: 'result',
+        clearable: true,
+        default: new Date('2017-09-28 17:36:05').toISOString(),
```
mnaamani: Tried the date picker, works really nicely in the terminal!

Perhaps a more reasonable default would be the current date `new Date().toISOString()`, so less fiddling with inital year and month for the user.

mnaamani (changes requested):
I'm suggesting to make multiple messages rather than a single one. Feels like a better design, but this will require a bit of refactoring here and in the storage-squid mappings.

metadata-protobuf/proto/Storage.proto
zeeshanakram3: Correct for storage nodes, but for distributor nodes we also need to know about the `worker_id`, since a bucket can be assigned to multiple operators?

distributor-node/src/commands/leader/set-node-operational-status.ts
zeeshanakram3: Please see https://github.com/Joystream/joystream/pull/4793#discussion_r1538138524

distributor-node/src/commands/leader/set-node-operational-status.ts
zeeshanakram3: > It makes more sense to make it required, the command name doesn't imply what the default value should be.

Done

distributor-node/src/commands/operator/set-metadata.ts
zeeshanakram3: It should skip the update but previously it was overriding, I have fixed it

distributor-node/src/command-base/default.ts
zeeshanakram3: Done

distributor-node/src/services/networking/query-node/queries/queries.graphql
zeeshanakram3: Done in [0537ee7](https://github.com/Joystream/joystream/pull/4793/commits/0537ee74c8f5caa8a14217668fc7cab59a5ca68c)

storage-node/src/services/queryNode/queries/queries.graphql
zeeshanakram3: Addressed

zeeshanakram3:
@mnaamani @kdembler I have addressed the requested changes. Also, re-designed the protobuf message schemas as requested. 

If the new protobuf messages are acceptable, then please publish the npm package so I would update the storage-squid mappings as well

mnaamani:
> @mnaamani @kdembler I have addressed the requested changes. Also, re-designed the protobuf message schemas as requested.
> 
> If the new protobuf messages are acceptable, then please publish the npm package so I would update the storage-squid mappings as well

published protobuf and joystream/js packages (for joystream/js I bumped the dependency on protobuf package)

PR #4796, opened 2023-06-26
Colossus 3.6.0

### LINKED ISSUE #4794: [Colossus] Storage Node - artificial operating limitations
### Problem
Currently colossus (storage-node0 imposes artificial constraint on a workers:

- Run a single instance for all buckets
- All bucket `transactorAccountId` must be the same account

This is reflected in the configuration/arguments.
- Buckets to be handled cannot be specified as an argument.
- Only a single account key can be passed

In the implementation, on startup the node checks the bucket obligations and for each bucket checks that the transactor account corresponds to the key provided:

https://github.com/Joystream/joystream/blob/cc43a96c15cb16f01cca6c47752444a8863e6fc3/storage-node/src/commands/server.ts#L281

This is severely limiting the flexibility for a worker to run multiple instances of a storage-node. This problem has been raised by the storage lead. The ability to run multiple nodes is needed to better scale operations of storage workers.

#### Dealing with multiple accounts
Operator must be able to provide multiple keys, corresponding to the buckets that will be handled by the running instance. I propose this be done by replacing the `keyFile` argument with `keyStorePath` who's value is a path to a folder containing multiple files. All of which would be loaded into the keyring of the node.

#### Bucket selection
Following from previous point, if an instance needs to be "read-only", ie operator does not intend it to handle new uploads, but only serve distributors.. technically a key is not required as it will not need to send any transactions. So the buckets to be operated should not be determined simply by which keys are found in the keystore.

Therefore bucket selection should be explicit. I think it makes sense to have a single argument `buckets` string where a matching pattern is passed specifying buckets and other options for each. (see point below about uploads restriction)

Currently there is no built-in transaction coordinator if multiple nodes intend to accept new uploads for the same bucket (presuming they have a shared storage volume) or for different buckets configured with the same transactorAccountId. This can lead to potential situation where two nodes will submit a transaction with the same nonce. To avoid this I propose that there should be an explicit configuration that operator chooses if the node will not-accept new uploads for a bucket on that instance (disabling HTTP POST requests for that bucket).

Keeping in mind that the metadata for a bucket currently only allows specifying a single url for both uploads and downloads, if an operator chooses to place multiple nodes behind a load balancer/nginx/caddy reverse proxy, they must have rules that route POST requests only to the instance that allows uploads.

Perhaps it would be useful also to update the message for setting the operator metadata so different URLs for each purpose can be provided. This is a bit disruptive though on the client applications side, but if done in a backward compatible way it could be rolled out smoothly.

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4794

- Added support for multiple transactor accounts when running the server, so we can take advantage of ability for operators to assign different transactor accounts per bucket.

- Added support specifying which buckets to serve when running a server.

- Bump version to v3.6.0

In combination these two changes now allows workers who operate multiple buckets to be able to run multiple nodes, for example with each node serving a single bucket.

- Also finally added behavior to allow server to startup even if transactor account is not provided to one or more buckets. The server will still serve downloads and sync those buckets. This is very useful for testing. I can run a node and specify a valid worker id and test the nodes behavior, open telemetry traces etc.

```sh
yarn storage-node server -d ~/tmp/ -o 3333 --apiUrl wss://rpc.joystream.org \
    -s -r 1 -w 8 --buckets 0,1,2,3,4,5,6,7

2023-07-16 03:07:35:735 info: Initializing runtime connection to: wss://rpc.joystream.org
2023-07-16 03:07:38:738 info: Waiting for chain to be synced before proceeding.
2023-07-16 03:07:38:738 info: Query node endpoint set: http://localhost:8081/graphql
2023-07-16 03:07:38:738 debug: Query - storageBucketsConnection
 ›   Warning: Missing transactor key for bucket 1
2023-07-16 03:07:39:739 warn: Only subset of buckets will process uploads!
2023-07-16 03:07:39:739 info: Buckets synced and served: 1
2023-07-16 03:07:39:739 info: Buckets accepting uploads: 
2023-07-16 03:07:39:739 info: Removing temp directory ...
2023-07-16 03:07:39:739 info: Creating temp directory ...
2023-07-16 03:07:39:739 debug: Local ID cache loaded.
2023-07-16 03:07:39:739 info: Synchronization enabled.
2023-07-16 03:07:39:739 debug: Max file size runtime parameter: 64424509440
2023-07-16 03:07:39:739 info: Listening on http://localhost:3333
2023-07-16 03:07:39:739 info: Sync paused for 1 minute(s).
```

~Even an invalid worker id can be specified, in which-case the server will startup, but with very limited functionality. This is useful for preparing a new node, testing endpoint accessibility, and general testing.~ We should add a separate command for this purpose.


storage-node/src/services/sync/storageObligations.ts
```diff
@@ -87,15 +87,13 @@ type DataObject = {
  */
 export async function getStorageObligationsFromRuntime(
   queryNodeUrl: string,
-  workerId: number
+  workerId: number,
```
zeeshanakram3: `workerId` param is not used anymore

storage-node/src/commands/server.ts
```diff
@@ -149,21 +151,71 @@ Supported values: warn, error, debug, info. Default:debug`,
...
 
-    if (flags.dev) {
-      await this.ensureDevelopmentChain()
+    const api = await this.getApi()
+
+    const workerId = flags.worker
+
+    if (!(await verifyWorkerId(api, workerId))) {
+      logger.warn(`workerId ${workerId} does not exist in the storage working group`)
+      // this.exit(ExitCodes.InvalidWorkerId)
```
zeeshanakram3: Kindly remove this commented line, since we have decided to bootstrap storage-node even if an invalid `workerId` is specified

storage-node/src/commands/server.ts
```diff
@@ -302,19 +354,9 @@ async function verifyWorkerId(
       )
     )
   )
+    .filter(([bucketId]) => bucketsToServe.length === 0 || bucketsToServe.includes(parseInt(bucketId)))
```
zeeshanakram3, line 357: What should be the behavior if the `workerId` input is invalid & list of `buckets` that needed to be served is valid, should the storage-node serve those buckets? The current implementation does not allow for that. 

I think `workerId` & `buckets` flags can be mutually exclusive as we can derive the worker info if the buckets flag is specified i.e. if `buckets` is provided then we can be assured that worker ID is the worker that is assigned to serve these buckets in the runtime. WDYT?

storage-node/src/command-base/ApiCommandBase.ts
```diff
@@ -123,61 +124,132 @@ export default abstract class ApiCommandBase extends Command {
...
-      accountUri = process.env.ACCOUNT_URI ?? ''
+  async loadKeys(flags: {
+    dev: boolean
+    keyFile?: string
+    password?: string | string[]
+    accountUri?: string | string[]
+    keyStore?: string
+  }): Promise<void> {
+    const keyring = this.getKeyring()
+    const { dev, password, keyFile, accountUri, keyStore } = flags
```
zeeshanakram3, line 166: Just to comfirm, what is the desired behavior here? Should these flags (`keyFile`, `accountUri`, `keyStore`) be allowed to be simultaneously used or they should be exclusive? Currently, they can be simultaneously used, if desired otherwise, these should be marked as exclusive

storage-node/src/services/webApi/app.ts
```diff
@@ -223,6 +226,10 @@ async function validateUploadFileParams(req: express.Request, res: express.Respo
   const dataObjectId = new BN(req.query.dataObjectId?.toString() || '')
   const bagId = req.query.bagId?.toString() || ''
 
+  if (!res.locals.uploadBuckets.includes(req.query.storageBucketId?.toString() || '')) {
```
zeeshanakram3: I guess `storageBucketId` defined above can be used here instead of duplicating the expression `req.query.storageBucketId?.toString() || ''`?

storage-node/src/commands/dev/sync.ts
```diff
@@ -56,13 +61,14 @@ export default class DevSync extends Command {
 
   async run(): Promise<void> {
     const { flags } = this.parse(DevSync)
-
+    const bucketId = '' + flags.bucketId
```
zeeshanakram3:
```suggestion
    const bucketId = flags.bucketId.toString()
```

I think better to use `.toString()` method?

zeeshanakram3 (commented):
Looks Good. I think just some cosmetic changes are required.

I tested this seems to be working as expected. The overall code refactorization and customization were long overdue, so great job.

storage-node/src/command-base/ApiCommandBase.ts
mnaamani: No they are not exclusive, all can be used at the same time.

storage-node/src/commands/server.ts
mnaamani: That's an excellent point I didn't consider that because allowing invalid worker id was a last minute change i didn't without thinking it through, also I just realized that a storage bucket can only have one operator (unlike distribution buckets).

> if buckets is provided then we can be assured that worker ID is the worker that is assigned to serve these buckets in the runtime. WDYT?

But in this case you can pass buckets for different workerId.. should we allow that? (It would only make sense if both 'workers' are the same person or are cooperating (sharing transactor-key and maybe role key))

I'm tempted to completely drop the `workerId` flag and only have the `buckets` flag so you always have to be explicit as to what bucket to serve.

I don't want to make this overly complicated and prone to making mistakes, just to allow a more testable node. I think its better that I revert this change so you must provide a valid worker id, and we can separately make have another command just for testing purpouses which would allow specifying any bucket.

storage-node/src/commands/server.ts
mnaamani: On second thoughts, decided to revert allowing invalid workerId: https://github.com/Joystream/joystream/pull/4796/files#r1270006202

zeeshanakram3: > On second thoughts, decided to revert allowing invalid workerId: https://github.com/Joystream/joystream/pull/4796/files#r1270006202

Sounds good

zeeshanakram3 (approved):
LGTM.

Issue #4802, opened 2023-07-17
Current Implementation of `sendExtrinsic` function may stuck indefinitely if the Tx has already finalised before processing the callback of `tx.signAndSend`


zeeshanakram3:
## Issue
Currently, the [sendExtrinsic](https://github.com/zeeshanakram3/joystream/blob/master/cli/src/base/ApiCommandBase.ts#L524) function (which is a wrapper around PolkadotJS' `tx.signAndSend` function) in Argus, Colossus & CLI may stuck indefinitely if the transaction has already finalized before processing the callback of `tx.signAndSend`.

This is because the [if-condition](https://github.com/zeeshanakram3/joystream/blob/master/cli/src/base/ApiCommandBase.ts#L532) in the `sendExtrinsic` function, would only resolve or reject if the transaction status is `isInBlock` or errored respectively. However, there might be the case that the transaction has already been finalized when the callback was run, so in that case, the promise would not resolve/reject. The function would be stuck and the calling function would be indefinitely waiting for the response.

## Background
I discovered this issue in YT-synch (since it uses the exact implementation from monorepo), where occasionally the video creation service would get stuck, this mostly happened when the YT-synch would be doing intensive IO-bound work. So my hunch is that in such cases callback function would be scheduled to run after the Tx has been finalized.

zeeshanakram3:
Addressed in #4803

PR #4803, opened 2023-07-17
Fix send extrinsic function for finalized txs

zeeshanakram3:
addresses #4802


mnaamani (changes requested):
LGTM.

We experienced this also in running integration tests where a large number of transactions were being generated, and I can see the same solution was applied.

Can you please just do minor version bump for all three packages.

zeeshanakram3:
> Can you please just do minor version bump for all three packages.

Addressed in [048f23c](https://github.com/Joystream/joystream/pull/4803/commits/048f23c33028bc1cd291668cb2b78217e5c461b8)

mnaamani (approved):

PR #4804, opened 2023-07-19
Docker compose tweaks

mnaamani:
- QueryNode `start.sh` simplify query-node start script
- `setup.sh`: 
  - Install latest version of docker-compose
  - do not replace existing installed version of docker
- docker-compose: configurable telemetry by adding new env variable `TELEMETRY_ENABLED` and `TELEMETRY_ENDPOINT`. Telemetry still disabled by default.
- Argus/Colossus/QueryNode can now be built without requiring to force `linux/x86-64` as the platform (works on mac m1, arm64) - there used to be some npm dependencies that didn't build natively probably because we were using node:14 docker image)

### DIFF
--- a/.env
+++ b/.env
@@ -123,6 +123,12 @@ APP_PRIVATE_KEY=this-is-not-so-secret-change-it
 TRUSTED_REVERSE_PROXIES=1
 # =====================================================================================
 
+## Telemetry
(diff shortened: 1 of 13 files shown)


zeeshanakram3 (approved):
LGTM

PR #4806, opened 2023-07-20
Readme update links to binaries

mnaamani:
Update Readme to not reference testnet and make binaries and chainspec easier to find on landing readme page.

Also updated shasum calculation for runtime code to exclude `.md` so in future node rebuild isn't required if we simply change some docs, as it doesn't affect the compile process.

Some github actions are failing because the github runner is running out of space, a fix is available in this PR https://github.com/Joystream/joystream/pull/4804 so maybe we can merge that first


scripts/runtime-code-shasum.sh
```diff
@@ -14,6 +14,7 @@ fi
 
 # sort/owner/group/mtime arguments only work with gnu version of tar!
 ${TAR} -c --sort=name --owner=root:0 --group=root:0 --mode 644 --mtime='UTC 2020-01-01' \
+    --exclude='*.md' \
```
bwhm, line 17: Is this hash actually referred to anywhere?

utils/api-scripts/README.md
```diff
@@ -49,7 +49,7 @@ yarn ts-node src/test-command.ts
 
 Some examples of how to write "scripts" is available under [./scripts/](./scripts) folder.
 These are "code-snippets" designed primarly for doing queries or very simple transactions.
-If you follow the example format you can also copy/paste them into https://testnet.joystream.org/#/js for execution.
+If you follow the example format you can also copy/paste them into https://polkadot.js.org/apps/#/js for execution.
```
bwhm, line 52: I don't think these works...

bwhm (approved):
Misclicked twice on "Add single comment", but these are super nitpicky :)

- All links seem to work
- All changes are to the better

The only `testnet` reference I found to be still around was in `/runtime/README.md`, and probably makes sense to keep for a while.

PR #4811, opened 2023-07-22
Standalone joystrea/node docker image

mnaamani:
Package the mainnet chainspec file with the docker image for joystream/node to make it simpler to deploy without having to provide it separately.

Some github actions are failing because the github runner is running out of space, a fix is available in this PR https://github.com/Joystream/joystream/pull/4804 so maybe we can merge that first


zeeshanakram3 (approved):
LGTM

Issue #4817, opened 2023-07-27
[Query Node] Add bag size info in the `StorageBag` entity


zeeshanakram3:
Currently, QN does not have any field for querying the size of `StorageBag`, so getting this information requires querying the RPC endpoint directly.

Indexing the size information in the QN is specifically useful for Storage Lead, e.g. for filtering bags and moving them based on the disk availability among different nodes.

PR #4818, opened 2023-07-27
[Query Node] Adds `StorageBag.objectsSize` field

zeeshanakram3:
addresses #4817


storage-node/src/services/queryNode/schema.graphql
```diff
@@ -1047,6 +1047,9 @@ type StorageBag implements BaseGraphQLObject {
   deletedById: ID
   version: Int!
   objects: [StorageDataObject!]!
+
+  """Total size of data objects in Bag"""
+  objectsSize: BigInt!
```
mnaamani, line 1052: just as a general question, if an older version of storage node does a query for `StorageBag` and they are not aware of the new schema, how will it behave? Will it fail to parse the response or just not be aware of this new field?

zeeshanakram3: Adding a new field to the graphql schema is generally future-proof (unless we make breaking changes to the schema), So if the older version of storage-node is unaware of the new schema and they don't query the new field, the query would work fine.

For example, the storage-node makes use of [storageBagsConnection](https://github.com/zeeshanakram3/joystream/blob/f9c565fe6acfb1b657a3c77a10dd10f5349eef1c/storage-node/src/services/queryNode/queries/queries.graphql#L81) query to get all assigned bags. As you can see the only field being queried is bag [id](https://github.com/zeeshanakram3/joystream/blob/f9c565fe6acfb1b657a3c77a10dd10f5349eef1c/storage-node/src/services/queryNode/queries/queries.graphql#L68). So being unaware of the new schema does not create the  issues.

mnaamani (approved):

PR #4820, opened 2023-07-27
Integration Tests tidy up scripts

mnaamani:
Too many confusing variables, and scripts to run integration tests.
Just did a bit of house keeping.
Dropped the REUSE_KEYS vaiable, it will always use by default if the `keys.json` is found (renamed from output.json)
The keys.json is explicitly removed in scripts when starting up a new "test network" to execute scenarios on.


mnaamani:
I'm going to rework this a bit and open new PR into nara branch.

Issue #4814, opened 2023-07-26
[Argus] Improve HEAD Request for asset


mnaamani:

zeeshanakram3:
Currently, for each `HEAD api/v1/{id}` request, a request to QN is made if the assets does not exists already. This isn't an optimal if distributor-node has to scale to serve hunderds of thousands of req/sec.

## Proposal

Look into `ApolloClient`'s built-in qraphql caching mechanism for avoiding frequent IO bound queryies for each asset. Or develop some custom solution for that.

kdembler:
@zeeshanakram3 I recall that we already introduced some caching of those requests. Do you think we should close this issue or there's still more work to be done here?

zeeshanakram3:
Completed in #4834

PR #4834, opened 2023-08-21
[Argus] Improves 'HEAD /assets/{id}' requests latency by implementing caching QN requests

zeeshanakram3:
addresses #4814

This PR implements `max-age` based cache strategy for `StorageDataObject` QN entity, whose corresponding data object asset is missing on the distributor-node & for which a QN request is need to be made.

So whenever a  `HEAD /assets/{id}` request is made, it could be satisfied in the following ways:
- If the object, for which a QN request is supposed to be made, does not exist in the Apollo's in-memory cache, the network call would be made to fetch the object and update the cache.
- If object exists in the in-memory cache and is older than the `CACHED_OBJECT_MAX_AGE` value, a network call would still be made to fetch the object and update the cache.
-  If object exists in the in-memory cache and is _NOT_ old then `CACHED_OBJECT_MAX_AGE`, the object would be served from the cache

### DIFF
--- a/distributor-node/src/services/networking/query-node/api.ts
+++ b/distributor-node/src/services/networking/query-node/api.ts
@@ -3,7 +3,6 @@ import {
   DocumentNode,
   FetchPolicy,
   HttpLink,
-  InMemoryCache,
   NormalizedCacheObject,
   from,
   split,
@@ -11,13 +10,14 @@ import {
 import { onError } from '@apollo/client/link/error'
 import { WebSocketLink } from '@apollo/client/link/ws'
 import { getMainDefinition } from '@apollo/client/utilities'
+import { InvalidationPolicyCache } from '@nerdwallet/apollo-cache-policies'
 import fetch from 'cross-fetch'
 import { FragmentDefinitionNode } from 'graphql'
 import { Logger } from 'winston'
 import ws from 'ws'
+import { ReadonlyConfig } from '../../../types'
 import { LoggingService } from '../../logging'
 import {
-  DataObjectDetails,
   DataObjectDetailsFragment,
   DistirubtionBucketWithObjectsFragment,
   GetActiveStorageBucketOperatorsData,
@@ -43,7 +43,7 @@ import { Maybe } from './generated/schema'
 
 const MAX_RESULTS_PER_QUERY = 1000
 
-export type QueryFetchPolicy = Extract<FetchPolicy, 'cache-first' | 'network-only' | 'no-cache'>
+export type QueryFetchPolicy = Extract<FetchPolicy, 'cache-first' | 'no-cache'>
 
 type PaginationQueryVariables = {
   limit: number
@@ -61,12 +61,13 @@ type PaginationQueryResult<T = unknown> = {
 type CustomVariables<T> = Omit<T, keyof PaginationQueryVariables>
 
 export class QueryNodeApi {
+  private config: ReadonlyConfig
   private apolloClient: ApolloClient<NormalizedCacheObject>
   private logger: Logger
-  private CachedObjectsAge: Map<string, Date> = new Map()
-  private CACHED_OBJECT_MAX_AGE = 1000 * 60 // 1 min
 
-  public constructor(endpoint: string, logging: LoggingService, exitOnError = false) {
+  public constructor(config: ReadonlyConfig, logging: LoggingService, exitOnError = false) {
+    this.config = config
+
     this.logger = logging.createLogger('QueryNodeApi')
     const errorLink = onError(({ graphQLErrors, networkError }) => {
       const message = networkError?.message || 'Graphql syntax errors found'
@@ -74,9 +75,9 @@ export class QueryNodeApi {
       exitOnError && process.exit(-1)
     })
 
-    const queryLink = from([errorLink, new HttpLink({ uri: endpoint, fetch })])
+    const queryLink = from([errorLink, new HttpLink({ uri: this.config.endpoints.queryNode, fetch })])
     const wsLink = new WebSocketLink({
-      uri: endpoint,
+      uri: this.config.endpoints.queryNode,
       options: {
         reconnect: true,
       },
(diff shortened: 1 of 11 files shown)


distributor-node/src/services/networking/query-node/api.ts
```diff
@@ -87,13 +93,13 @@ export class QueryNodeApi {
...
       cache: new InMemoryCache({
-        dataIdFromObject: (object) => {
-          // setup cache object id for ProcessorState entity type
-          if (object.__typename === 'ProcessorState') {
-            return object.__typename
-          }
-          return defaultDataIdFromObject(object)
+        typePolicies: {
+          ProcessorState: {
+            keyFields: [],
```
mnaamani: Does this tell apollo client explicitly NOT to cache processor state ?

distributor-node/src/services/networking/query-node/api.ts
```diff
@@ -59,6 +63,8 @@ type CustomVariables<T> = Omit<T, keyof PaginationQueryVariables>
 export class QueryNodeApi {
   private apolloClient: ApolloClient<NormalizedCacheObject>
   private logger: Logger
+  private CachedObjectsAge: Map<string, Date> = new Map()
+  private CACHED_OBJECT_MAX_AGE = 1000 * 60 // 1 min
```
mnaamani: 1min sounds good, but we probably want to make this a configurable in the config.yml file

distributor-node/src/services/networking/query-node/api.ts
```diff
@@ -172,12 +181,53 @@ export class QueryNodeApi {
...
 
-  public getDataObjectDetails(objectId: string): Promise<DataObjectDetailsFragment | null> {
-    return this.uniqueEntityQuery<GetDataObjectDetailsQuery, GetDataObjectDetailsQueryVariables>(
+  protected readFragment<FragmentT, VariablesT extends Record<string, unknown>>(
+    id: string,
+    fragment: DocumentNode
+  ): FragmentT | null {
+    // Get the fragment name, usually first element of the definitions array is the name of the top level fragment
+    const fragmentName = (fragment.definitions[0] as FragmentDefinitionNode).name.value
+    if (!fragmentName) {
```
mnaamani, line 198: Do we want to also check that the `fragmentName` === `"id"`

mnaamani (changes requested):
Implementation looks good. I left just a clarification question.

I have to say I was surprised that there is no builtin cache expiry config, and I presume that is why you had to add way to track the age of cached object.

Lets make the max age configurable in config.yml (or do you think it makes more sense to hardcode it?) and bump the distributor node version.

mnaamani:
> I have to say I was surprised that there is no builtin cache expiry config, and I presume that is why you had to add way to track the age of cached object.

@zeeshanakram3 Found a third party cache implementation for apollo client, is it something we can use to make our code a bit cleaner? https://github.com/NerdWalletOSS/apollo-cache-policies

zeeshanakram3:
> > I have to say I was surprised that there is no builtin cache expiry config, and I presume that is why you had to add way to track the age of cached object.
> 
> @zeeshanakram3 Found a third party cache implementation for apollo client, is it something we can use to make our code a bit cleaner? https://github.com/NerdWalletOSS/apollo-cache-policies

Thanks for sharing, will test it.

distributor-node/src/services/networking/query-node/api.ts
zeeshanakram3: Thank you for pointing this out. This is actually incorrect, basically, this is just the alternative & newer way (introduced in Apollo 3.0) to assign the cache IDs to objects. We used `dataIdFromObject` previously. I have fixed it

distributor-node/src/services/networking/query-node/api.ts
zeeshanakram3: So I checked that `Id` isn't necessarily same as the fragment name. `id` refers to the cached object ID (e.g. for the `StorageDataObject` entity type it can be `StorageDataObject:1`) however the fragment name refers to the specific fragment of the entity type that we want to query from the cache, e.g. [DataObjectDetails](https://github.com/zeeshanakram3/joystream/blob/argus_improve_head_request_latency/distributor-node/src/services/networking/query-node/queries/queries.graphql#L23)

zeeshanakram3:
1. I reimplemented & tested [apollo-cache-policies](https://github.com/NerdWalletOSS/apollo-cache-policies) npm package (using OpenTelemetry traces) and the TTL seems to be working as expected.
2. Also made the TTL configurable using `interval.queryNodeCacheTTL` 
3. Bumped package version + added CHNAGELOG.md

mnaamani:
Works as expected.
To test I commented out the code that checks for file in local fs before trying query:

```ts
  // distributor-node/src/services/content/ContentService.ts
  public async objectStatus(
       objectId: string,
       qnFetchPolicy: QueryFetchPolicy = 'no-cache'
  ): Promise<ObjectStatus> {
    /*
    const pendingDownload = this.stateCache.getPendingDownload(objectId)

    if (!pendingDownload && this.exists(objectId)) {
      return { type: ObjectStatusType.Available, path: this.path(objectId) }
    }

    if (pendingDownload) {
      return { type: ObjectStatusType.PendingDownload, pendingDownload }
    }
    */
   const objectInfo = await this.networking.dataObjectInfo(objectId, qnFetchPolicy)
...
```

then I tested (crudely) 

```
# first request ~ 400ms
time http HEAD http://localhost:3334/api/v1/assets/1
docker stop graphql-server
# confirm we get a cached response, ~200ms response
time http HEAD http://localhost:3334/api/v1/assets/1
.. try a few more times, until we get an error response because distributor tried to do a query to QN
# restart query-node
docker start query-node
time http HEAD http://localhost:3334/api/v1/assets/1
# successful request as expected.
```

Tried with different TTL values also.

mnaamani (approved):
Nice work.
This will also benefit GET requests.
Not clear how much impact this will have but if there is any kind of repeated requests to assets the node doesn't have this fix will cover it.

Issue #4824, opened 2023-08-02
QN: Validator profile verification


thesan:
Once the Profile is bound it can be manually verified. The profile should be "verified" if the links, the about section, and any other information mentioned in the membership data is genuine and actually describes the person or company operating the validator account. This verification is done by the membership working group. Either a worker with `membershipWorkingGroup.workerRemark` or the lead with `membershipWorkingGroup.leadRemark`. These extrinsics will be called via a CLI command to be created.

thesan:
Superseded by #4905

PR #4839, opened 2023-08-24
Fix/add validator mapping

### LINKED ISSUE #4823: QN: Validator profile

```gql
memberMetadata(where: {
  validatorAccount: {
    isVerified_eq: true
    address_in: [
    ]
  }
}) {
  member { handle }
  validatorAccount { address  }
}
```

chrlschwb:
fixes #4823 
- changed `MembershipMetadata` in metadata-protobuf/proto/Membership.proto
- added schema and mapping for the validator entity


query-node/schemas/validator.graphql
```diff
@@ -0,0 +1,6 @@
+type Validator @entity {
+  id: ID!
+  isValidator: Boolean!
+  validatorAccount: String
+  member: String
```
thesan: The member field should be of type `Membership`.

The way the QN are structured, `Validator` should be available on `MembershipMetadata` because this entity linking a validator to a membership only exist in the QN and not really on chain. For this reason maybe `Validator` should be renamed `ValidatorProfile`. That said due to some hydra limitation we have to also link this entity to the `Membership` entity in order to be able to filter members by validator addresses and state. So this should look like this:
```suggestion
  memberMetadata: MemberMetadata!
  member: Membership!
```

query-node/schemas/validator.graphql
```diff
@@ -0,0 +1,6 @@
+type Validator @entity {
+  id: ID!
+  isValidator: Boolean!
```
thesan: `isVerified` is a lot clearer IMO

query-node/mappings/src/validator.ts
```diff
@@ -0,0 +1,35 @@
+import { DatabaseManager } from '@joystream/hydra-common'
+import {
+  MemberMetadata,
+  Membership,
+  MembershipExternalResource,
+  MembershipExternalResourceType,
+  Validator,
+} from 'query-node/dist/model'
+
+export async function saveValidator(
```
thesan, line 8: This function has to be called on membership creation and updates

metadata-protobuf/proto/Membership.proto
```diff
@@ -21,6 +21,8 @@ message MembershipMetadata {
       IRC = 8;
       WECHAT = 9;
       WHATSAPP = 10;
+      VALIDATOR = 11;
```
thesan: I don't think the validator account should be passed as an external resource. It's not really an external resource and also with the current mapping it would still appear in the QN and Pioneer as an external resource which would be confusing.

Instead I think it could just be added like this:
```proto
message MembershipMetadata {
  optional string validatorAccount = 6; // Address of the member's validator account
}
```

thesan (changes requested):
IMO as far as the QN is concern Validators are a Membership concept. So I think instead of adding `query-node/schemas/validator.graphql` and `query-node/schemas/validator.graphql` these changes should be added to `query-node/schemas/membership.graphql` and `query-node/mappings/src/membership.ts`.

Also it might be because I haven't touch the QN in a while but I don't understand the changes to `query-node/chain-metadata/xxxx.json` files is this necessary ?

Finally it's for @zeeshanakram3 to decide but if the changes to `distributor-node/src/services/networking/query-node/schema.graphql` and `storage-node/src/services/queryNode/schema.graphql` are not needed to use this in Pioneer I think these changes should be avoided to reduce the risks of conflicts later on.

zeeshanakram3 (changes requested):
- As Theo said `query-node/chain-metadata/xxxx.json` are auto-generated files, so no need to edit or format these.
- We are now been committing the the generated `graphql.schema` files in both `storage-node` & `distributor-node`, it's basically done for the ease of generating the docker images of storage/distributor nodes, so it's fine  (cc. @thesan)
- Q: If there is a `isVerified` field in the validator entity, then who will do the verification? And there there should be protobuf message for validator verification action too.  (cc. @thesan)

**Also, address all the changes asked by Theo**

thesan:
@zeeshanakram3 thank you for having a look.

> We are now been committing the the generated graphql.schema files in both storage-node & distributor-node, it's basically done for the ease of generating the docker images of storage/distributor nodes, so it's fine (cc. @thesan)

I'll keep this in mind!

> Q: If there is a isVerified field in the validator entity, then who will do the verification? And there there should be protobuf message for validator verification action too. (cc. @thesan)

The HR workers/lead will take care of the verification with their: `workerRemark`/`leadRemark` extrinsics. But that's in the scope of another issue: #4824.

The `isVerified` will always be `false` following a member creation or update event so IMO there is no need to add it to the `Membership.proto`.

@chrlschwb please don't close your PRs. If you didn't mean to merge `master` into it you can still reset your branch to `6d40bd7afd21f2c947eb0d04d6584ad0c4f3a0cb` and force push or if you want to restart from scratch you can still do that and force push a new `fix/addValidatorMapping`.

PR #4843, opened 2023-08-28
fix error update

### LINKED ISSUE #4823: QN: Validator profile

chrlschwb:
fix #4823 update


thesan (changes requested):
@chrlschwb
As I mentioned in here https://github.com/Joystream/joystream/pull/4839#issuecomment-1695776675 let's keep all the work in the PR #4839. This might be a long review process so closing a PR and creating a new one after each review is not acceptable.

PR #4844, opened 2023-08-28
Fix/is verified validator

chrlschwb:
fix input verified state #4824


query-node/mappings/src/workingGroups.ts
```diff
@@ -714,18 +714,18 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
     if (group.name !== 'forumWorkingGroup') {
       return invalidMetadata(`The ${group.name} is incompatible with the remarked moderatePost`)
     }
-    const { postId, rationale } = metadata.moderatePost
+    const { postId, rationale, validator } = metadata.moderatePost
     const actor = await getWorkingGroupLead(store, group.name)
 
-    await moderatePost(store, event, 'leadRemark', postId, actor, rationale)
+    await moderatePost(store, event, 'leadRemark', postId, actor, rationale, validator)
   } else {
```
thesan, line 729: That's not it. The idea is to add one more case to this function. ATM there is just one case which is `if (metadata?.moderatePost) { ... }` otherwise the mapping returns an `invalidMetadata` errror. So in addition to the `metadata?.moderatePost` there should be a e.g `metadata?.verifyValidator` case.

thesan (changes requested):
I detailed the mapping logic a bit more. But more importantly IMO you should focus on #4823 first because that's where the schema will be created. Then you can rebase this PR on the branch solving #4823.

Issue #4813, opened 2023-07-26
Pruning of assets from storage nodes that they are no longer obliged to store


zeeshanakram3:
## Problem

Currently, storage nodes don't delete assets no longer assigned to them in the runtime. And hence dead assets keep on accumulating on the nodes. This means that disk usage of the nodes would continue to grow unless operators manually clean up the space which is very risky.
Other complications of not pruning assets are that DataObjects/Bags can't be effectively moved b/w different buckets/operators, as this action does not physically free up the space, so Storage Lead can't use the Globally available disk space effectively.

zeeshanakram3:
# Proposal 

TBD

yasiryagi:
#there should be a mean for the storage server operator to run a command:

-  that will show the diff between what is assigned to the bucket (i.e QN) and what is in the server.
- the command should have modes:
  - View
  - Action (delete ) , should have a data loss mitigation::
    - Provide a warning before execution. 
    - Check the availability of the object where it should if any.
    - Force a replication if the object does not exist where it should be,
  - Set the storage server to auto do the pruning, the auto pruning ( This mode should also be available in storage server command as an option. )should have a data loss mitigation:
    -  Back off period between detection (logs and metrics should be generated at this stage) and actioning the pruning.
    - Check the availability of the object where it should if any.
    - Force a replication if the object does not exist where it should be,

bedeho:
shouldn't this happen automatically?

yasiryagi:
That is the ideal with couple of two more consideration:

1. Data loss mitigation consideration.
2. Ability of the operator to force it through command.

kdembler:
@zeeshanakram3 Should we close this already or there's more work to be done?

PR #4845, opened 2023-08-28
[Colossus] Add pruning service

zeeshanakram3:
addresses #4813 

This PR adds:
- adds cleanup service to prune outdated assets, and configuration option to enable/disable it on storage-node startup
- adds CLI command 'dev:cleanup' for performing pruning actions (ideally should not be used in production)
- extends '/status' endpoint with additional info, e.g. sync, cleanup/pruning & serving buckets configurations

## How the pruning of assets work

The `cleanupService` runs the data objects cleanup/pruning workflow. It removes all the locally stored data objects that the operator is no longer obliged to keep, because either the data object has been deleted from the runtime or it has been moved to some other bucket/s

**_PRECONDITIONS:_**
- Since the cleanup uses the QueryNode to query the data obligations, the QueryNode processor should not lag more than `MAXIMUM_QN_LAGGING_THRESHOLD` blocks, otherwise the cleanup workflow would not be performed in order to avoid pruning assets based on outdated state.
- If the asset being pruned from this storage node still exists in the runtime (i.e. its storage obligation has been moved), then at least _"X"_ other storage nodes should hold the asset, otherwise, the cleanup workflow would not be performed, where _"X"_ is defined by `MINIMUM_REPLICATION_THRESHOLD`
 - If the asset being pruned from this storage node is currently being downloaded by some external actors, then the cleanup action for this asset would be postponed


storage-node/src/commands/dev/cleanup.ts
```diff
@@ -0,0 +1,63 @@
...
+    help: flags.help({ char: 'h' }),
+    workerId: flags.integer({
+      char: 'w',
+      required: true,
+      description: 'Storage node operator worker ID.',
+    }),
+    bucketId: flags.integer({
+      char: 'b',
+      required: true,
+      description: 'The buckerId to sync',
```
mnaamani:
```suggestion
      description: 'The buckerId to prune/cleanup',
```

storage-node/src/commands/server.ts
```diff
@@ -62,6 +66,16 @@ export default class Server extends ApiCommandBase {
       description: 'Interval between synchronizations (in minutes)',
       default: 1,
     }),
+    cleanup: flags.boolean({
+      char: 's',
```
mnaamani: char 's' already assigned for sync. Better remove it as I don't know what the behavior is if short flag is used and multiple flags assigned same char.

storage-node/src/commands/server.ts
```diff
@@ -62,6 +66,16 @@ export default class Server extends ApiCommandBase {
...
+      description: 'Enable cleanup/pruning of no-longer assigned assets.',
+      default: false,
+    }),
+    cleanupInterval: flags.integer({
+      char: 'i',
+      description: 'Interval between periodic cleanup actions (in minutes)',
+      default: 1,
```
mnaamani: I would say this should be much higher like 1h, maybe even longer?
Depends on how much work/time it takes per cleanup run.

storage-node/src/commands/server.ts
```diff
@@ -308,6 +352,39 @@ async function runSyncWithInterval(
   }
 }
 
+/**
+ * Run the data synchronization process.
```
mnaamani:
```suggestion
 * Run the data cleanup process.
```

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -39,7 +42,7 @@ export async function loadDataObjectIdCache(uploadDir: string, tempDirName: stri
   const tempDirectoryName = path.parse(tempDirName).name
   const ids = localIds.filter((dataObjectId) => dataObjectId !== tempDirectoryName)
 
-  idCache = new Set(ids)
+  ids.map((id) => idCache.set(id, 0))
```
mnaamani:
```suggestion
  ids.forEach((id) => idCache.set(id, 0))
```

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -74,6 +105,24 @@ export async function deleteDataObjectIdFromCache(dataObjectId: string): Promise
...
+ * Get data object ID from the local cache.
+ *
+ * @param dataObjectId - Storage data object ID
+ */
+export async function getDataObjectIdFromCache(
+  dataObjectId: string
+): Promise<{ dataObjectId: DataObjectId; pinnedCount: DataObjectPinCount } | undefined> {
+  if (idCache.has(dataObjectId)) {
+    await lock.acquireAsync()
+    const id = { dataObjectId, pinnedCount: idCache.get(dataObjectId) as DataObjectPinCount }
```
mnaamani: return type of function doesn't seem to match `id`, `dataObjectId` is a string where as in the return type it is `DataObjectId` (u64)

storage-node/src/services/sync/cleanupService.ts
```diff
@@ -0,0 +1,154 @@
...
+  for (const entry of dataObligations.storageBuckets) {
+    // Skip all buckets of the current WorkerId (this storage provider)
+    if (entry.workerId !== currentWorkerId) {
+      bucketOperatorUrlById.set(entry.id, entry.operatorUrl)
+    }
+  }
+
+  const timeoutMs = 60 * 1000 // 1 minute since it's only a HEAD request
+  const deletionTasks: DeleteLocalFileTask[] = []
+  await Promise.all(
```
mnaamani:
```suggestion
  await Promise.allSettled(
```
To avoid the promise being rejected when HEAD request to one data object fails, [Promise.allSettled()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) is more appropriate.

Also I can image if the number of dataobjects could potentially be very large, doing a massive number of HTTP HEAD requests in parallel might be a bad idea. Perhaps we should do it in batches.

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -57,9 +63,14 @@ export async function getFile(
...
     })
 
+    stream.on('end', async () => {
+      await unpinDataObjectIdFromCache(req.params.id)
+    })
+
     stream.pipe(res)
   } catch (err) {
     sendResponseWithError(res, next, err, 'files')
+    await unpinDataObjectIdFromCache(req.params.id)
```
mnaamani, line 70: instead of unpinning twice, it could be placed in `finally { .. }`

mnaamani (changes requested):
Overall this looks good and will be quite valuable. Left some feedback. Also needs an update from master to fix some merge conflicts.

storage-node/src/services/sync/cleanupService.ts
zeeshanakram3: Done, but I have to change `target` option in `tsconfig.json` to `es2020` as `Promise.allSettled` does not seem to be available in `es2017`

storage-node/src/services/caching/localDataObjects.ts
zeeshanakram3: The DataObjectId is also a string, as you can see
https://github.com/Joystream/joystream/blob/d8fcf541ffc9d10dfb0973be22a78824d453d3c9/storage-node/src/services/caching/localDataObjects.ts#L7

storage-node/src/commands/server.ts
zeeshanakram3: Yeah, ideally it should be higher that the `sync` interval, I just set it to 1min, basically test the pruning behavior locally. Now I have set it to `6hrs`

storage-node/src/commands/server.ts
zeeshanakram3: Done, changed the short flag to `c`. WDYT?

storage-node/src/services/caching/localDataObjects.ts
mnaamani: Ah yes, i didn't spot that type definition.

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -74,6 +104,24 @@ export async function deleteDataObjectIdFromCache(dataObjectId: string): Promise
...
+/**
+ * Get data object ID from the local cache.
+ *
+ * @param dataObjectId - Storage data object ID
+ */
+export async function getDataObjectIdFromCache(
+  dataObjectId: string
+): Promise<{ dataObjectId: DataObjectId; pinnedCount: DataObjectPinCount } | undefined> {
+  if (idCache.has(dataObjectId)) {
+    await lock.acquireAsync()
```
mnaamani: shouldn't the lock be acquired before checking `idCache.has()` ?

mnaamani (changes requested):
One final small change and this should be ready to test in production :)
And updating from master to resolve merge conflict.

storage-node/src/services/caching/localDataObjects.ts
zeeshanakram3: Addressed. However, placing a lock before the if condition can lead to unnecessary blocking if the `dataObjectId` is not present in the cache. So I reimplemented by first checking the cache without acquiring the lock. This is a quick operation and if the `dataObjectId` is not present, just return without the overhead of acquiring the lock.

mnaamani (approved):
I will merge this work into a new branch `colossus-beta` where we can do final tweaks for a "beta" release.

PR #4849, opened 2023-08-29
Fix/4823 validator profile

### LINKED ISSUE #4823: QN: Validator profile

chrlschwb:
fix #4823


query-node/schemas/membership.graphql
```diff
@@ -47,6 +47,9 @@ type MemberMetadata @entity {
 
   "Social media handles, email address..."
   externalResources: [MembershipExternalResource] @derivedFrom(field: "memberMetadata")
+
+  isVerified: Boolean!
```
thesan: Since this is not in a separate entity anymore. Please rename it to something more explicit like `isVerifiedValidator`

query-node/mappings/src/membership.ts
```diff
@@ -326,6 +328,11 @@ export async function members_MemberProfileUpdated({ store, event }: EventContex
     }
   }
 
+  if (typeof metadata?.validatorAccount === 'string') {
+    member.metadata.validatorAccount = (metadata.validatorAccount || null) as string
```
thesan: This could be:
```suggestion
  if (typeof metadata?.validatorAccount === 'string' && metadata.validatorAccount !== member.metadata.validatorAccount) {
    member.metadata.validatorAccount = metadata.validatorAccount || undefined
```
To prevent resetting the verification by mistake when the same account is passed and to allow removing the validator account by passing an empty string.

thesan (changes requested):
The mapping looks good. Please reset the changes to the `.yarnrc.yml` and `query-node/schemas/content.graphql`

Also please test the mapping in `tests/network-tests/src/flows/membership/updatingProfile.ts`

thesan (changes requested):
The mappings and schema are great now but as I mentioned in the last review please add a test to the  membership updating profile flow.

To run the tests locally you can follow the [query node README](https://github.com/Joystream/joystream/blob/master/query-node/README.md#development):
1. In `.env` uncomment `JOYSTREAM_NODE_TAG` and set it to the latest testing image, currently it's: `44832c0fe83c518779d52b5b3a2dcce94ae2d805` (do not commit this change.
2. Run the `./query-node/build.sh` script to build the schema and mappings locally.
3. Try running `DEBUG=true ./query-node/run-tests.sh memberships`. Hopefully it works otherwise you will have to run the full scenario all the time with `DEBUG=true ./query-node/run-tests.sh` (which is more time consuming).
4. After running the tests once in (1) you might to modify them and re-run the a lot faster with `REUSE_KEYS=true yarn workspace network-tests run-test-scenario memberships`

If you change the mappings or the schema, you will have to redo (2) and (3).

You can check this commit for more details and ask me if you get stuck.

metadata-protobuf/proto/Membership.proto
```diff
@@ -25,5 +25,11 @@ message MembershipMetadata {
     optional ResourceType type = 1;
     optional string value = 2;
   }
+  
   repeated ExternalResource externalResources = 5;
+  
+  optional bool isVerifiedValidator = 6;
```
thesan: I just realized there shouldn't be any `isVerifiedValidator` here. The `MembershipMetadata` can only define a `validatorAccount` but not whether or not it's verified.

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -16,6 +16,8 @@ export type MemberProfileData = {
   about?: string | null
   avatarUri?: string | null
   externalResources?: MembershipMetadata.IExternalResource[] | null
+  isVerifiedValidator?: boolean
```
thesan: Same here
```suggestion
```

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -66,6 +68,12 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
...
         ? this.newValues.externalResources || null
         : this.oldValues.externalResources,
+      isVerifiedValidator: isSet(this.newValues.isVerifiedValidator)
+        ? this.newValues.isVerifiedValidator
+        : this.oldValues.isVerifiedValidator,
+      validatorAccount: isSet(this.newValues.validatorAccount)
+        ? this.newValues.validatorAccount
+        : this.oldValues.validatorAccount,
     }
   }
```
thesan, line 77: The `assertQueryNodeEventIsValid` method should:
```ts
assert.equal(newMetadata.isVerifiedValidator, false)
assert.equal(newMetadata.validatorAccount, expected.validatorAccount)
```

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -66,6 +68,12 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
       externalResources: isSet(this.newValues.externalResources)
         ? this.newValues.externalResources || null
         : this.oldValues.externalResources,
+      isVerifiedValidator: isSet(this.newValues.isVerifiedValidator)
+        ? this.newValues.isVerifiedValidator
+        : this.oldValues.isVerifiedValidator,
```
thesan: So I don't think this is needed
```suggestion
```

tests/network-tests/src/flows/membership/updatingProfile.ts
```diff
@@ -18,15 +18,31 @@ export default async function updatingProfile({ api, query }: FlowProps): Promis
...
+    {
+      handle: 'New handle 1',
+      name: 'New name',
+      isVerifiedValidator: false,
+      validatorAccount: 'validator address',
+    },
+    {
+      handle: 'New handle 2',
+      isVerifiedValidator: false,
+      validatorAccount: 'validator address',
```
thesan: So it actually test something:
```suggestion
      validatorAccount: 'validator address 2',
```

tests/network-tests/src/flows/membership/updatingProfile.ts
```diff
@@ -18,15 +18,31 @@ export default async function updatingProfile({ api, query }: FlowProps): Promis
...
       avatarUri: '',
-      externalResources: [{ type: MembershipMetadata.ExternalResource.ResourceType.EMAIL, value: 'A@example.com' }],
+      externalResources: [
+        {
+          type: MembershipMetadata.ExternalResource.ResourceType.EMAIL,
+          value: 'A@example.com',
+        },
+      ],
+      isVerifiedValidator: false,
+      validatorAccount: 'validator address',
```
thesan: Then you could try not to pass it to check it remains:
```suggestion
```

tests/network-tests/src/flows/membership/updatingProfile.ts
```diff
@@ -35,9 +51,17 @@ export default async function updatingProfile({ api, query }: FlowProps): Promis
...
+          type: MembershipMetadata.ExternalResource.ResourceType.EMAIL,
+          value: 'B@example.com',
+        },
+        {
+          type: MembershipMetadata.ExternalResource.ResourceType.HYPERLINK,
+          value: 'example.com',
+        },
       ],
+      isVerifiedValidator: false,
+      validatorAccount: 'validator address',
```
thesan: Finally based on the mappings it should be removable:
```suggestion
      validatorAccount: '',
```

tests/network-tests/src/flows/membership/updatingProfile.ts
```diff
@@ -18,15 +18,31 @@ export default async function updatingProfile({ api, query }: FlowProps): Promis
   const updates: MemberProfileData[] = [
     // Partial updates
     // FIXME: Currently handle always need to be provided, see: https://github.com/Joystream/joystream/issues/2503
-    { handle: 'New handle 1', name: 'New name' },
-    { handle: 'New handle 2' },
+    {
+      handle: 'New handle 1',
+      name: 'New name',
+      isVerifiedValidator: false,
```
thesan:
```suggestion
```

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -66,6 +68,12 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
...
+      validatorAccount: isSet(this.newValues.validatorAccount)
+        ? this.newValues.validatorAccount
```
thesan: We expect the `validatorAccount` to be `null` (i.e removed) when an empty string was passed:
```suggestion
        ? this.newValues.validatorAccount || null
```

thesan (changes requested):
You can reset the `yarn.lock` by running `git checkout master -- yarn.lock` and committing the changes.

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -93,6 +97,7 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
       expected.externalResources?.map(asMembershipExternalResource) ?? []
     )
     assert.isFalse(Utils.hasDuplicates(newMetadata.externalResources?.map(({ type }) => type)))
+    assert.equal(newMetadata.isVerifiedValidator, false)
```
thesan: If you remove `assert.equal(newMetadata.validatorAccount, expected.validatorAccount)` you're not actually checking if the `validatorAccount` was added to the member metadata in the QN. `validatorAccount` shouldn't be blank. It should be 'validator address' after the first update, then 'validator address 2', then it should remain 'validator address 2' and finaly it should be `null` after the 4th update. If it's not the case something is wrong either with the test or the mapping.

thesan (changes requested):

query-node/mappings/src/membership.ts
chrlschwb: removing `null` creates error during membership update test where `validatorAccount` should be set to null

thesan: Nice catch ! I'm not sure of this but maybe `null` actually unset the value on hydra while `undefined` is ignored (I think I read an issue about something like that a while ago).

tests/network-tests/src/Api.ts
```diff
@@ -124,7 +124,10 @@ export class ApiFactory {
 
   constructor(api: ApiPromise, treasuryAccountUri: string, miniSecret: string) {
     this.api = api
-    this.keyring = new Keyring({ type: 'sr25519', ss58Format: JOYSTREAM_ADDRESS_PREFIX })
+    this.keyring = new Keyring({
```
thesan: Please use git to remove the changes on this file to reduce potential conflicts later

tests/network-tests/src/fixtures/membership/utils.ts
```diff
@@ -19,6 +19,8 @@ type MemberCreationParams = {
   externalResources?: MembershipMetadata.IExternalResource[] | null
   metadata: Bytes
   is_founding_member: boolean
+  isVerifiedValidator?: boolean
```
thesan:
```suggestion
```

tests/network-tests/src/flows/membership/updatingProfile.ts
```diff
@@ -18,15 +18,28 @@ export default async function updatingProfile({ api, query }: FlowProps): Promis
...
-    { handle: 'New handle 1', name: 'New name' },
-    { handle: 'New handle 2' },
+    {
+      handle: 'New handle 1',
+      name: 'New name',
+      validatorAccount: 'validator address',
+    },
+    {
+      handle: 'New handle 2',
+      validatorAccount: 'validator address2',
```
thesan: I know I suggested 'validator address2' before but just to make sure it works with `undefined`:
```suggestion
```
It would be a more useful test IMO.

thesan (commented):
Good job just a few last minor changes

tests/network-tests/src/flows/membership/updatingProfile.ts
chrlschwb: the fourth update is `undefined`

thesan: Yes but before the fourth update `validatorAccount` has been set to `undefined` already (during the 3rd one) so I think having `undefined` on the 2de update is more valuable

thesan (approved):
:100:

zeeshanakram3 (changes requested):
Looks good overall. Please revert the changes in `query-node/chain-metadata/*.json` files. These are auto-generated files that should be edited.

zeeshanakram3 (approved):
LGTM

Issue #4842, opened 2023-08-25
Review QN for faults in mappings


bedeho:
Recently, it was found that mappings that process metadata messages in a way which was not robust, namely causing the processor to halt if the message was malformed in some way. It is therefore suggested that a systematic review is conducted across all mappings to make sure there is no other instance of such fault behaviour.

PR #4856, opened 2023-09-06
QN mappings review & fixes

### LINKED ISSUE #4853: Implement mapping for `UpdateGlobalNftLimit` proposal type
# Context
Discovered while working on #4842 

In the `ProposalCreated` mappings, the condition for checking `proposalDetails.type === "UpdateGlobalNftLimit"` does not exist. And since the `ProposalCreated` mappings throw an error for unimplemented/unsupported proposal types, this mapping would throw an exception when the `UpdateGlobalNftLimit` proposal type is unfiltered (in future runtime upgrade) and a proposal is created of this type. 

https://github.com/Joystream/joystream/blob/d1448236349154afa2c1a725f7144990c37d5042/query-node/mappings/src/proposals.ts#L306

# Fix
Handle the `UpdateGlobalNftLimit` proposal type in the mapping

### LINKED ISSUE #4855: Gracefully handle actions by `Lead` as `ContentActor`
# Context
Discovered while working on https://github.com/Joystream/joystream/issues/4842

If the content Lead creates a video with an `autoIssueNft` parameter set, then the mappings would crash as mappings currently don't recognize `Lead` as a valid content actor. This issue is present in many other mappings such as `OpenAuctionStarted`, `EnglishAuctionStarted`, `NFTIssued` etc. because a common function (`convertContentActorToChannelOrNftOwner`) is used everywhere.

https://github.com/Joystream/joystream/blob/d1448236349154afa2c1a725f7144990c37d5042/query-node/mappings/src/content/utils.ts#L446

image

# Fix 
Gracefully handle actions by `Lead` as `ContentActor`

zeeshanakram3:
addresses https://github.com/Joystream/joystream/issues/4842

This PR:
- fixes #4855
- fixes #4853
- Creates a new custom Eslint rule for disallowing the use of the `throw` keyword
- Removes all instances of `throw` from the mappings and adds helper function `getById` & `getByIdOrFail` instead


query-node/mappings/package.json
```diff
@@ -16,14 +16,15 @@
...
     "@joystream/hydra-common": "5.0.0-alpha.4",
     "@joystream/hydra-db-utils": "5.0.0-alpha.4",
-    "@joystream/warthog": "^2.41.9",
     "@joystream/js": "^1.4.0",
-    "@apollo/client": "^3.2.5"
+    "@joystream/warthog": "^2.41.9",
+    "@polkadot/types": "8.9.1"
   },
   "devDependencies": {
+    "eslint-plugin-local-rules": "2.0.0",
```
mnaamani, line 27: Adding this package seems to require native build on mac/arm64 so I had to add `node-gyp` as a dev dependency to successfully build docker image of query-node.

query-node/mappings/src/common.ts
```diff
@@ -88,18 +95,20 @@ export function inconsistentState(extraInfo: string, data?: unknown): never {
...
+ * Reports an unimplemented mapping/variant error for which a runtime implementation logic exists and is not filtered.
+ */
+export function unimplementedError(extraInfo: string, data?: unknown): never {
+  const errorMessage = 'unimplemented error: ' + extraInfo
 
   // log error
   logger.error(errorMessage, data)
 
+  // eslint-disable-next-line local-rules/no-throw
   throw errorMessage
```
mnaamani: Would it make sense to call `process.exit` here instead of throwing?
This would be to prevent any code capturing this throw, and also caller doesn't need to `return` from the calling function.

mnaamani (changes requested):
This is a great improvement, and code looks so much more readable.
Small change and merge conflict fix needed otherwise good to go.

I ran processor against mainnet works great. I think it would be nice to have a tool to compare state between two query-node instances.

query-node/mappings/package.json
zeeshanakram3: Addressed in [b3f01a3](https://github.com/Joystream/joystream/pull/4856/commits/b3f01a374390d97a8ae149135c6e2520902ccc0b)

query-node/mappings/src/common.ts
zeeshanakram3: Addressed in [810240d](https://github.com/Joystream/joystream/pull/4856/commits/810240d2ae555d114778650db8b76823769a603e)

mnaamani:
@zeeshanakram3 I fixed conflicts and updated version here https://github.com/zeeshanakram3/joystream/pull/7
I would like to merge this to master soon, so I can updated nara branch from master soon.

mnaamani (approved):

PR #4857, opened 2023-09-06
fix orion environment not working

dobertRowneySr:
Fixes the current orion environment not working when starting the playground
I have also removed the squid-archive services in favour of our new archive endpoint


.env
```diff
@@ -92,35 +92,65 @@ BALANCE_CREDIT=300000000000
...
+#   rections weights,
+#   [joystream creation weight, YT creation weight]
+# ]
+RELEVANCE_WEIGHTS="[1, 0.03, 0.3, 0.5, [7,3]]"
 MAX_CACHED_ENTITIES=1000
 APP_PRIVATE_KEY=this-is-not-so-secret-change-it
-TRUSTED_REVERSE_PROXIES=1
+SESSION_EXPIRY_AFTER_INACTIVITY_MINUTES=60
+SESSION_MAX_DURATION_HOURS=720
+EMAIL_CONFIRMATION_ROUTE=http://localhost:4074/api/v1/confirm-email?token={token}
```
mnaamani, line 142: obviously this will only work when opening email from the machine. But I gather we would most likely disable email confirmation and notifications in a playground setup so thats fine.

docker-compose.yml
```diff
@@ -318,92 +301,84 @@ services:
     ports:
       - "3002:3002"
 
-  squid-archive:
-    container_name: squid-archive
```
mnaamani, line 322: Wouldn't make more sense for a playground's default deployment to have its own squid-archive and gateway and configure default value of `ARCHIVE_GATEWAY_URL` to point to it?

mnaamani (changes requested):
Works well, but the default archive gateway on startup is the production chain which doesn't make sense. I think we should keep an archive + gateway in the setup. If we choose to run orion from this repo for the production chain then we set the `CUSTOM_ARCHIVE_GATEWAY_URL` env variable to `https://archive.joystream.org/graphql`

docker-compose.yml
dobertRowneySr: ok

mnaamani:
@ignazio-bovo The CI check that is failing:
https://github.com/Joystream/joystream/actions/runs/6099420853/job/16551528071?pr=4857#step:12:16435 

because it tries to use the adjacent docker-compose file `docker-compose-no-bind-volumes.yml` by first [copying it over to](https://github.com/Joystream/joystream/blob/d1448236349154afa2c1a725f7144990c37d5042/.github/workflows/joystream-apps-docker.yml#L84) `docker-compose.yml`

Please apply the same fixes to that as well.

.env
```diff
@@ -92,35 +92,71 @@ BALANCE_CREDIT=300000000000
...
+EMAIL_CONFIRMATION_TOKEN_EXPIRY_TIME_HOURS=24
+EMAIL_CONFIRMATION_TOKEN_RATE_LIMIT=5
+ACCOUNT_OWNERSHIP_PROOF_EXPIRY_TIME_SECONDS=300 # 5 minutes
+COOKIE_SECRET=this-is-not-so-secret-change-it
+
+TRUST_PROXY=uniquelocal
+
+# Sendgrid API
+SENDGRID_API_KEY=
```
mnaamani: There is an identical `SENDGRID_API_KEY` variable below for the faucet.
perhaps we can prefix the one for orion with `ORION_SENDGRID_API_KEY` then pass that env variable in docker-compose environment section (so you don't have to modify orion code to expect a different variable name)

postgres.conf
```diff
@@ -0,0 +1,4 @@
+listen_addresses = '*'
```
mnaamani: can we rename this file to `orion-postgres.conf`, there is another PR where I was planning to add a postgres.conf file for the query-node db. It makes sense to have two different conf files for each db instance.

docker-compose-no-bind-volumes.yml
```diff
@@ -307,92 +290,159 @@ services:
...
+  squid-archive-explorer:
+    container_name: orion_archive_explorer
+    hostname: orion_archive_explorer
+    restart: unless-stopped
+    image: subsquid/substrate-explorer:firesquid
+    env_file:
+      - .env
+    environment:
+      DB_TYPE: postgres # set to `cockroach` for Cockroach DB
+      DB_HOST: archive_db
```
mnaamani: I think the hostname is incorrect, (should it be orion_archive_db), also the correct port for the archive db `ARCHIVE_DB_PORT` is not used here.

I'm seeing connection error log for the explorer container:
```
2023-09-11 10:54:19 {"level":3,"time":1694415259620,"ns":"sqd:substrate-explorer","msg":"failed to connect to postgres://postgres@archive_db:5432/squid-archive","err":{"errno":-3008,"code":"ENOTFOUND","syscall":"getaddrinfo","hostname":"archive_db","stack":"Error: getaddrinfo ENOTFOUND archive_db\n    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:26)"}}
2023-09-11 10:54:19 {"level":2,"time":1694415259730,"ns":"sqd:substrate-explorer","msg":"listening on port 3000"}
```

mnaamani (changes requested):
A couple minor fixes needed of archive explorer config. Otherwise working nicely.

With so env variables now in the `.env` file and with some similarly named ones, i think in future we need to do a bit of refactoring to make it more manageable.

orion-postgres.conf
mnaamani, line 1: You also need to change the name in the docker-compose files where this file is mapped.

docker-compose-no-bind-volumes.yml
```diff
@@ -437,7 +437,7 @@ services:
       - .env
     environment:
       DB_TYPE: postgres # set to `cockroach` for Cockroach DB
-      DB_HOST: archive_db
+      DB_HOST: orion_archive_db
```
mnaamani, line 442: you also need
```
DB_PORT: ${ARCHIVE_DB_PORT}
```

mnaamani (changes requested):
You need the `DB_PORT` also to be set for the explorer, and the renamed `postgres.conf` file needs to be adjusted in the docker-compose file for the orion-db service to match the new name:

```
    volumes:
      - orion-db-data:/var/lib/postgresql/data
      - ./orion-postgres.conf:/etc/postgresql/postgresql.conf
 ```

File: .env [hunk 7]
```diff
@@ -92,35 +92,71 @@ BALANCE_CREDIT=300000000000
...
+EMAIL_CONFIRMATION_TOKEN_EXPIRY_TIME_HOURS=24
+EMAIL_CONFIRMATION_TOKEN_RATE_LIMIT=5
+ACCOUNT_OWNERSHIP_PROOF_EXPIRY_TIME_SECONDS=300 # 5 minutes
+COOKIE_SECRET=this-is-not-so-secret-change-it
+
+TRUST_PROXY=uniquelocal
+
+# Sendgrid API
+ORION_SENDGRID_API_KEY=
+ORION_SENDGRID_FROM_EMAIL=gateway@example.com
```
mnaamani, line 152: These need to be used in docker-compose to pass the expected env var name to orion like so:

```
environment:
      - SENDGRID_API_KEY=${ORION_SENDGRID_API_KEY}
      - SENDGRID_FROM_EMAIL=${ORION_SENDGRID_FROM_EMAIL}
```

mnaamani (approved):

PR #4863, opened 2023-09-11
QueryNode: Run graphql-server with pm2 cluster mode

mnaamani:
Enable horizontal scaling graphql-server on machine using pm2 process manager in cluster mode.

The updated script to start graphql-server will only start a single node cluster. Once the graphql-server is running we can manage the cluster with pm2:

Get status:
```sh
docker exec -it graphql-server yarn pm2 status
```

Add one more instance to the cluster
```sh
docker exec -it graphql-server yarn pm2 scale graphql-server +1
```

Monitor the cluster:
```sh
docker exec -it graphql-server yarn pm2 status
```

Its important to note that the processor send status updates to the graphql-server. As more instances are added, these udpates are balanced across the cluster, so they will appear less frequent from the point of view of applications that are subscribing through the graphql-server to these updates.


zeeshanakram3 (approved):
LGTM

PR #4868, opened 2023-09-13
Fix/4824 verify validator

chrlschwb:
#4824


query-node/mappings/src/workingGroups.ts
```diff
@@ -744,6 +752,14 @@ export async function workingGroups_WorkerRemarked({ store, event }: EventContex
     const actor = await getWorker(store, group.name, workerId)
 
     await moderatePost(store, event, 'workerRemark', postId, actor, rationale)
+  } else if (metadata?.verifyValidator) {
+    if (group.name !== 'operationsWorkingGroupBeta') {
+      return invalidMetadata(`The ${group.name} is incompatible with the remarked moderatePost`)
+    }
+    const member = await getMemberById(store, workerId, ['metadata', 'metadata.externalResources'])
```
thesan: No need for `metadata.externalResources`

metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -63,8 +63,12 @@ message ModeratePost {
   required string rationale = 2;
 }
 
+message VerifyValidator{
+  required uint64 memberId = 1;
```
thesan, line 67: This is enough in theory but please also add the validator account too. So in the mappings you can make sure the validator account on the member metadata is actually the one the HR worker intends to verify (and it wasn't switched at some point).

thesan (changes requested):
So far so good!

query-node/mappings/src/workingGroups.ts
```diff
@@ -744,6 +752,20 @@ export async function workingGroups_WorkerRemarked({ store, event }: EventContex
...
+    if (group.name !== 'operationsWorkingGroupBeta') {
+      return invalidMetadata(`The ${group.name} cannot verify validator accounts`)
+    }
+
+    const actor = await getWorker(store, group.name, workerId)
+    if (actor === undefined) {
+      return invalidMetadata(`${actor} is not an HR account`)
+    }
+
+    const member = await getMemberById(store, workerId, ['metadata'])
```
chrlschwb: workerId is the memberId of the HR worker not that of the validator.

`const {memberId, validatorAccount} =  const { postId, rationale } = metadata.verifyValidator`

The memberId is string but I need u64 for the getMemberById function. What should I do?

thesan: Oh I completely missed that.

You could use:
```ts
import { createType } from '@joystream/types'
createType('MemberId', Number(metadata.verifyValidator.memberId))
```
But `getMemberById` just converts the id into a string so maybe just change the `getMemberById` `id` argument to: `MemberId | string`.

tests/network-tests/src/flows/membership/updatingVerifyAccount.ts
```diff
@@ -0,0 +1,44 @@
...
+  MemberProfileData,
+  UpdateProfileHappyCaseFixture,
+} from '../../fixtures/membership'
+
+import { extendDebug } from '../../Debugger'
+import { FixtureRunner } from '../../Fixture'
+import { generateParamsFromAccountId } from '../../fixtures/membership/utils'
+import { MembershipMetadata } from '@joystream/metadata-protobuf'
+
+export default async function updatingVerifyAccount({ api, query }: FlowProps): Promise<void> {
```
thesan: This is not the actually testing the validator profile verification. To test this I think you would need to create a fixture like `RemarkModeratePostsFixture`. But that would be a lot of effort. @zeeshanakram3 in your opinion is this test worth the effort or can we skip it here ?

tests/network-tests/src/fixtures/forum/RemarkModeratePostsFixture.ts
```diff
@@ -33,7 +33,10 @@ export class RemarkModeratePostsFixture extends WithForumWorkersFixture {
   protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise'>[]> {
     return this.moderations.map((u) => {
       const metadata = Utils.metadataToBytes(RemarkMetadataAction, {
-        moderatePost: { postId: Long.fromString(String(u.postId)), rationale: u.rationale },
+        moderatePost: {
+          postId: Long.fromString(String(u.postId)),
+          rationale: u.rationale,
+        },
```
thesan: Please revert this change

thesan (changes requested):
The rest looks good :+1:

zeeshanakram3:
> This is not the actually testing the validator profile verification. To test this I think you would need to create a fixture like `RemarkModeratePostsFixture`. But that would be a lot of effort. @zeeshanakram3 in your opinion is this test worth the effort or can we skip it here ?

Yes, so I think the integration-test should be added for this action. Although, there isn't any specific place to add fixture for this specific action, but still I would say that add a new fixture `VerifyVealidatorAccountFixture.ts` in `tests/network-tests/src/fixtures/membership`. 

After creating the new fixture, add a new flow (say `validatorProfile.ts`) in `tests/network-tests/src/flows/membership`, in which you would have to both first create a validator profile (most of the code from the `updatingProfile` flow will be reused) , and then after that do the validator verification using the `VerifyVealidatorAccountFixture.ts`  fixture.

Last step, add the flow (`validatorProfile.ts`) to the `tests/network-tests/src/scenarios/memberships.ts` scenarios file. Make sure that this flow has dependency on the `hireLeads` job/flow.

cc. @chrlschwb

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -0,0 +1,78 @@
...
+import { assert } from 'chai'
+import { QueryNodeApi } from '../../QueryNodeApi'
+import { SubmittableExtrinsic } from '@polkadot/api/types'
+import { BaseQueryNodeFixture } from '../../Fixture'
+import { MemberContext, EventDetails } from '../../types'
+import { MembershipFieldsFragment, MemberProfileUpdatedEventFieldsFragment } from '../../graphql/generated/queries'
+import { MembershipMetadata } from '@joystream/metadata-protobuf'
+import { Utils } from '../../utils'
+
+export class VerifyValidatorAccountFixture extends BaseQueryNodeFixture {
```
thesan: The fixture needs a `getExtrinsics` method, to allow it to actually send the transaction that it's meant to test, which is either `operationsWorkingGroupBeta.leadRemark(msg)` or `operationsWorkingGroupBeta.workerRemark(workerId, msg)`. You can check [this fixture](https://github.com/Joystream/joystream/blob/master/tests/network-tests/src/fixtures/forum/RemarkModeratePostsFixture.ts#L33) for example.

tests/network-tests/src/scenarios/memberships.ts
```diff
@@ -24,4 +25,5 @@ scenario('Memberships', async ({ job }) => {
   job('transferring invites', transferringInvites)
   job('managing staking accounts', managingStakingAccounts)
   job('updating member verification status', updatingVerificationStatus).after(hireLeads)
+  job('updating member verify account status', updateValidatorAccount)
```
thesan:
```suggestion
  job('updating member validator account', updateValidatorAccount).after(hireLeads)
```
Also I think this job needs to be added the `full.ts` scenario too in order to be ran by the CI.

thesan (changes requested):

metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -63,8 +63,13 @@ message ModeratePost {
...
+  required string validatorAccount = 2;
```
thesan: Hey @chrlschwb the scope has changed:
The verification will come from the membership group instead of HR and the worker will actually be verifying the member profile of the validator accounts not the validator accounts. So there shouldn't be a `validatorAccount` field here.
Also the workers should be able to "unverify" a membership (in case of error). So there should be a:
```suggestion
  required bool is_veryfied = 2;
}
```
instead. As a result the mappings and tests code will have to be changed too.

thesan (changes requested):

tests/network-tests/src/flows/membership/validatorProfile.ts
```diff
@@ -0,0 +1,30 @@
...
+  api.enableDebugTxLogs()
+
+  const VerifyValidator = [
+    {
+      memberId: '1',
+      validatorAccount: 'validator address',
+    },
+    {
+      memberId: '2',
+      validatorAccount: '',
```
thesan: Please update the parameters here:
```suggestion
      memberId: '1',
      isVerified: true,
    },
    {
      memberId: '2',
      isVerified: true,
      asWorker: {{a membership group worker id}}, // This would be good to make sure non lead worker can call the extrinsic too
```

Also there should be a second call where `isVerified` is set to `false` to make sure workers can reverse the verification if something comes up.

thesan (changes requested):

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -24,17 +24,6 @@ export class VerifyValidatorAccountFixture extends BaseQueryNodeFixture {
...
-    return this.verifyValidator.map((u) => {
-      const metadata = Utils.metadataToBytes(RemarkMetadataAction, {
-        verifyValidator: { memberId: Long.fromString(String(u.memberId)), validatorAccount: u.validatorAccount },
-      })
-      return u.asWorker
-        ? this.api.tx.operationsWorkingGroupBeta.workerRemark(u.asWorker, metadata)
-        : this.api.tx.operationsWorkingGroupBeta.leadRemark(metadata)
-    })
-  }
-
```
thesan: As I said this morning: these are still needed. They are the extrinsics used by the membership WG to verify the memberships

tests/network-tests/src/scenarios/memberships.ts
```diff
@@ -24,4 +25,5 @@ scenario('Memberships', async ({ job }) => {
   job('transferring invites', transferringInvites)
   job('managing staking accounts', managingStakingAccounts)
   job('updating member verification status', updatingVerificationStatus).after(hireLeads)
+  job('updating member verify account', updateValidatorAccount).after(hireLeads)
```
thesan: This job should also be added to `tests/network-tests/src/scenarios/full.ts` so the CI actually runs it

tests/network-tests/src/fixtures/membership/InviteMembersHappyCaseFixture.ts
```diff
@@ -61,7 +61,7 @@ export class InviteMembersHappyCaseFixture extends StandardizedFixture {
         handle,
         rootAccount,
         controllerAccount,
-        metadata: { name, about, avatar, externalResources, validatorAccount },
+        metadata: { name, about, avatar, externalResources },
```
thesan: Please reverse the changes on this file from this PR. This test has nothing to do with this PR (it's about #4915)

tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -57,7 +57,6 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
     )
     assert.isFalse(Utils.hasDuplicates(metadata.externalResources?.map(({ type }) => type)))
     assert.equal(metadata.isVerifiedValidator, false)
-    assert.equal(metadata.validatorAccount, expected.validatorAccount)
```
thesan: Same thing for this file

thesan (changes requested):
In addition please address https://github.com/Joystream/joystream/pull/4868#discussion_r1342824534

tests/network-tests/src/scenarios/full.ts
```diff
@@ -126,4 +127,5 @@ scenario('Full', async ({ job, env }) => {
   job('init storage and distribution buckets via CLI', [initDistributionBucket, initStorageBucket]).after(
     contentDirectoryJob
   )
+  job('updating member verify account', updateValidatorAccount).after(hireLeads)
```
thesan: This is correct but for organization sake please move this just under the `updating member verification status` job

tests/network-tests/src/flows/membership/validatorProfile.ts
```diff
@@ -20,6 +20,11 @@ export default async function validatorAccount({ api, query }: FlowProps): Promi
       isVerified: true,
       asWorker: 'j4VEC6FcJtBrwYQKhBAoB6Rj83jDeVua6azuHBrri1zoksBkz',
     },
+    {
+      memberId: '3',
+      isVerified: false,
+      asWorker: 'j4VEC6FcJtBrwYQKhBAoB6Rj83jDeVua6azuHBrri1zoksBkz',
+    },
```
thesan: The problem here is that the member 3 is already not verified so this doesn't really test unverifying a member. Instead the idea would be to do something like:
```ts
const verifyValidatorFixture = new VerifyValidatorAccountFixture(api, query, VerifyValidator)
const verifyValidatorRunner = new FixtureRunner(verifyValidatorFixture)
await verifyValidatorRunner.run()
await verifyValidatorRunner.runQueryNodeChecks()

const unVerifyValidatorFixture = new VerifyValidatorAccountFixture(api, query, UnVerifyValidator)
const unVerifyValidatorRunner = new FixtureRunner(unVerifyValidatorFixture)
await unVerifyValidatorRunner.run()
await unVerifyValidatorRunner.runQueryNodeChecks()
```
Here `UnVerifyValidator` would unverify a member that was verified in `VerifyValidator`.

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -12,7 +12,7 @@ import { BaseQueryNodeFixture } from '../../Fixture'
 
 export type ValidaotrAccountInput = {
```
thesan, line 17: Please rename this:
```suggestion
export type VerifyValidatorInput = {
```

tests/network-tests/src/scenarios/memberships.ts
thesan: Please change the description to `updating member verify validator profile` also `updateValidatorAccount` should probably be renamed to `updateValidatorProfile`

tests/network-tests/src/scenarios/full.ts
thesan: same here

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Utils } from '../../utils'
+import { BaseQueryNodeFixture } from '../../Fixture'
+
+export type ValidaotrAccountInput = {
+  memberId: string
+  isVerified: boolean
+  asWorker?: string
+}
+
+export class VerifyValidatorAccountFixture extends BaseQueryNodeFixture {
```
thesan: I think this fixture should renamed too:
```suggestion
export class VerifyValidatorProfileFixture extends BaseQueryNodeFixture {
```

thesan: But more importantly the fixture should define `runQueryNodeChecks` and this method should actually check whether all the memberships `isVerifiedValidator` have been updated correctly

thesan (changes requested):

tests/network-tests/src/flows/membership/validatorProfile.ts
chrlschwb: Currently, there is no `unverifying` mechanism.
`Membership.leadRemark` and `Membership.workerRemark` can only `verify` validator profile.

Once, a validator profile is verified, it will automatically unverify when the validator's membership details are edited.

thesan: Yes but this is why you added `is_verified` to `VerifyValidator` in `metadata-protobuf/proto/WorkingGroups.proto` when it's `false` the profile should get unverified.

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -2,48 +2,91 @@ import { assert } from 'chai'
...
+    //   this.getEventFromResult(this.tx)
+    // }
+  }
+
+  protected assertQueryNodeEventIsValid(qEvent: MemberVerificationStatusUpdatedEventFieldsFragment, i: number): void {
+    console.log(qEvent, '---------------------------------------------------')
+
+    // assert.equal(qEvent.member.id, this.inputs[i % this.inputs.length].asMember.toString())
+    // assert.equal(qEvent.account, this.inputs[i % this.inputs.length].account.toString())
+  }
```
thesan: AFAIK this is not necessary:
```suggestion
```

tests/network-tests/src/fixtures/membership/VerifyValidatorAccountFixture.ts
```diff
@@ -2,48 +2,91 @@ import { assert } from 'chai'
...
-      return u.asWorker
-        ? this.api.tx.operationsWorkingGroupBeta.workerRemark(u.asWorker, metadata)
-        : this.api.tx.operationsWorkingGroupBeta.leadRemark(metadata)
+      return u.memberId
+        ? this.api.tx.membershipWorkingGroup.workerRemark(u.memberId, metadata)
+        : this.api.tx.membershipWorkingGroup.leadRemark(metadata)
     })
   }
 
+  private assetVerifyValidatorTest(qMember: MembershipFieldsFragment[] | null): void {
```
thesan: Please rename this method to:
```suggestion
  private assertQueriedMembershipAreValid(qMember: MembershipFieldsFragment[] | null): void {
```

thesan: also please remove the commented code and console.log

tests/network-tests/src/flows/membership/validatorProfile.ts
```diff
@@ -3,31 +3,29 @@ import { BuyMembershipHappyCaseFixture } from '../../fixtures/membership'
...
+      memberId: '27',
+      isVerifiedValidator: false,
     },
   ]
 
-  const verifyAccountFixture = new VerifyValidatorAccountFixture(api, query, VerifyValidator)
+  const verifyAccountFixture = new VerifyValidatorProfileFixture(api, query, VerifyValidator)
   const remarkModerateRunner = new FixtureRunner(verifyAccountFixture)
   await remarkModerateRunner.run()
   await remarkModerateRunner.runQueryNodeChecks()
```
thesan, line 37: Please address this https://github.com/Joystream/joystream/pull/4868#discussion_r1364249766

tests/network-tests/src/flows/membership/validatorProfile.ts
```diff
@@ -3,31 +3,29 @@ import { BuyMembershipHappyCaseFixture } from '../../fixtures/membership'
...
-      asWorker: 'j4VEC6FcJtBrwYQKhBAoB6Rj83jDeVua6azuHBrri1zoksBkz',
+      isVerifiedValidator: false,
     },
     {
-      memberId: '3',
-      isVerified: false,
-      asWorker: 'j4VEC6FcJtBrwYQKhBAoB6Rj83jDeVua6azuHBrri1zoksBkz',
+      memberId: '27',
+      isVerifiedValidator: false,
     },
```
thesan: These 3 inputs are not actually changing the member metadata `isVerifiedValidator`. `isVerifiedValidator` should be set to `true`. If the tests don't pass it probably means that the mappings have a bug.

Also the 3 inputs are basically identical, so there's no point having them all. If you manage to add a membership WG worker and pass their id `asWorker` then 2 inputs should be enough: one without the `asWorker` field (so ran as a lead) and another with it. If not just one with `{ memberId: '1', isVerifiedValidator: false }`.

tests/network-tests/src/scenarios/full.ts
```diff
@@ -127,5 +127,5 @@ scenario('Full', async ({ job, env }) => {
   job('init storage and distribution buckets via CLI', [initDistributionBucket, initStorageBucket]).after(
     contentDirectoryJob
   )
-  job('updating member verify account', updateValidatorAccount).after(hireLeads)
+  job('updating member verify profile', updateValidatorAccount).after(hireLeads)
```
thesan: Please address https://github.com/Joystream/joystream/pull/4868#discussion_r1364238450

tests/network-tests/src/fixtures/membership/UpdateVerificationStatusFixture.ts
```diff
@@ -31,6 +31,7 @@ export class UpdateVerificationStatusFixture extends StandardizedFixture {
   }
 
   protected async getExtrinsics(): Promise<SubmittableExtrinsic<'promise'>[]> {
+    console.log(this.updates, '================+++++++++++++++++++=================')
```
thesan: Please remove this change

thesan (changes requested):
Again the changes to these files are unnecessary:
- `metadata-protobuf/proto/Membership.proto`
- `query-node/mappings/src/membership.ts`
- `query-node/schemas/membership.graphql`
- `tests/network-tests/src/fixtures/membership/GiftMembershipHappyCaseFixture.ts`
- `tests/network-tests/src/fixtures/membership/InviteMembersHappyCaseFixture.ts`
- `tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts`
- `tests/network-tests/src/fixtures/membership/utils.ts`
- `tests/network-tests/src/flows/membership/updatingProfile.ts`
- `tests/network-tests/src/graphql/queries/membership.graphql`

They just make this PR more complex. Please remove them for now. You can add them back on #4915 later on. If after removing all these schema changes one these fixtures fails please just push the failing code and I'll have a look into it.

Issue #4789, opened 2023-06-13
Symbol string for `issue_token` 


dobertRowneySr:
Needed to be set in the orion as a human-readable character, eg "JOY","AVA", "HAPI" etc.
I don't think it is a good idea to pass a whole string as an argument which might make the extrinsic more expensive among other things.

What I would suggest is:
- remove the symbol bookkeeping from runtime
- add a metadata parameters to the `TokenIssuanceParameters`
- have the metadata with the symbol parsed and then set in orion

_Originally posted by @Lezek123 in https://github.com/Joystream/orion/pull/99#discussion_r1227206158_

bedeho:
I don't understand the problem, can you restate a clear problem section here? what is this symobl bookkeping in the runtime and what is wrong about it?

bedeho:
His comment seems to be saying something different:

> Currently, the runtime emits hash of the symbol, not the symbol itself, so I think a change in the runtime is required to support this.

dobertRowneySr:
1. What is the purpose of storing the symbol on chain if we just set it at initialization and then we don't even use it. We just store it on a map `SymbolUsed`? Can we remove it, plain and simple?
2. In order to solve the above problem can we just have a metadata item for the token symbol to allow orion to store it in the `Token` entity as a field?
3. I don't think it makes sense to store them both, as one user can specify the hash for `JOY` in the parameters but then set another string like `ETH` in the metadata, and the runtime cannot check this case (as far as I know)

bedeho:
> What is the purpose of storing the symbol on chain if we just set it at initialization

The premise is incorrect, it is used again when a new token is created to check for possible conflict, so it is indeed read back. I think if ticker only lives in metadata, it raises the question of how you deal with conflicts. So you create many tokens in a row with same ticker, what do they get as ticker? If first one gets to reserve it, then what will ticker of the remaining ones be, nothing? what if first token is deleted, then the old ones still have nothing? Its possible to do something like that, but then we are allowing "null" as ticker symbols, I don't love that, but its an open question.

So its up to you, if you want to simplify runtime, with no ticker sympbol accounting, you need a proper logic for managing conflicts and deletions. I guess the simplest one is 

- if ticker is available, you get it
- if ticker is not available, you don't get it, you have no ticker
- you can never change the ticker later, even if you have no ticker and you want to turn it into something which is free, because this causes phishing risk

up to you.

But I think the most important thing here is
- but ticker in transaction somehow still
[...]

Issue #4713, opened 2023-03-27
Creator Token Metadata


dobertRowneySr:
- [x] add a `creator_token_remark` extrinsic for metadata support
- [x] compile list of metadata message format (see down below)
- [x] ~publish~ patch protobuf package with new metadata for CRT?
- [x] process metadata in orion v2?
- [x] Review and open PR in orion repo https://github.com/Joystream/orion/pull/99

dobertRowneySr:
# CRT Metadata message format
``` go
syntax = "proto2";

message TokenMetadata {
  optional string name = 1; // Title
  optional string description = 2; // token description
  oneof avatar {
    uint32 avatar_object = 3; // avatar for token - index into external [assets array](#.Assets)
    string avatar_uri = 4; // Url to member's avatar
  }
  repeated Benefit benefits = 5; // benefits for tokne
  optional string whitelist_application_note = 6; // note for applicant
  optional string whitelist_application_apply_link = 7; // link to application process
}

message Benefit {
  optional string title = 1;
  optional string description = 2;
  optional string emoji = 3;
  uint32 display_order = 4;
}

message SaleMetadata {
  optional string terms_and_conditions = 1
}

```

dobertRowneySr:
symbol should be added.
Addressed in #4871

Issue #4488, opened 2022-12-07
Consideration on AMM parameters


dobertRowneySr:
AMM pricing formula:
```
price = a * x + b
```
### Considerations
- `x` = AMM provided liquidity: i.e. only the token minted through the AMM will be taken into account for price computation). This will make the semantics for `b` to be "initial unit price". I think that this is way more manageable than using the `token_total_supply` from a runtime POV.
- There are no trivial consraints for `a`. Can `a` be zero?  I don't think so, otherwise you should issue a sale **usability** 
- Can `b` be zero? I think it should if we allow a sale to have `sale_price = 0`.
- Dangers of having `a,b` too high: Numeric overflow in price computation. But the eval function will protect against this.
- Dangers of having `a,b` too low: Supply dilution (which can be view as **vulnerability** ) thus they should be set by the governance
- `a,b` should be thought as on a $10^{-10}$ scale since we are quoting the price in HAPI

bedeho:
> Can a be zero? I don't think so, otherwise you should issue a sale usability

I don't see why not really, some may even want it, but more imporantly, there are lots of edge cases of "problems" with any values, see response here: https://github.com/Joystream/joystream/issues/4151#issuecomment-1375627080

> Can b be zero? I think it should if we allow a sale to have sale_price = 0.

I think this value also should pretty much be unconstrained.

> Dangers of having a,b too high: Numeric overflow in price computation. But the eval function will protect against this.

What is eval function?

dobertRowneySr:
- `a` should have a min cap to avoid dilution by issuer
- `a` should not necessarely have a max cap, as overflow checks are in place

dobertRowneySr:
Runtime expects the extrinsic caller to specify bot parameter a and b used for the amm price point computation: 
https://github.com/Joystream/joystream/blob/564e4e0fa028785d640a4dfb6ef69c4962f8defe/runtime-modules/project-token/src/types.rs#L575-L576

https://github.com/Joystream/joystream/blob/564e4e0fa028785d640a4dfb6ef69c4962f8defe/runtime-modules/content/src/lib.rs#L3721-L3722

During our design meetings however, we established  that however there should have been a default value.

This can be enforced either:
A. by Atlas (front end)
B. by runtime by allowing the `AmmParams` to take optional `a,b` parameters and setting the default ones (above) in case they are null

[...]

PR #4871, opened 2023-09-14
Crt release v1

dobertRowneySr:
This PR contains all the fixes for v1 in order for allow Orion (and Atlas) to progress
#### Changes (and issues addressed)
- [x]  remove symbol map from runtime #4789 
- [x] Avoid that Amm and revenue split being active #4783
- [x] update all CRT metadata #4713 
- [x] remove the deadline parameter from amm extr.s #4483 
- [x] both module account and amm treasury account are guaranteed to be initialised with `ExistentialDeposit` #4198 
- [x] update with nara #4876
- [x] benchmark Amm extrinsics #4446 
- [x] merge changes for #4378
- [x] Added minor security fixes such as make the `TransferOutput` parameter a bounded vector and add a minimum allowed value for the Amm `a`
- [x] Also after carefully reviewing AMM parameters default recommendations https://github.com/Joystream/joystream/issues/4488#issuecomment-1768635722 Settled on types for amm curve parameters

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205679220762292) by [Unito](https://www.unito.io)


types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@joystream/types",
-  "version": "2.1.0",
+  "version": "2.3.0",
```
mnaamani: I know I already suggested 2.3.0, but seeing that nara will also need a version bump. Lets update major version instead (I think this is what we came to as best approach with Leszek)

```suggestion
  "version": "4.0.0",
```

And Nara will be version `3.0.0`

distributor-node/package.json
```diff
@@ -13,7 +13,7 @@
     "@joystream/metadata-protobuf": "^2.8.0",
     "@joystream/opentelemetry": "1.0.0",
     "@joystream/storage-node-client": "^3.0.0",
-    "@joystream/types": "^3.0.0",
+    "@joystream/types": "^2.0.0",
```
mnaamani: Not clear why this is being reverted .. but I can imagine it was quite difficult to keep things in sync with the multitude of times you rebased

runtime/src/lib.rs
```diff
@@ -240,16 +240,9 @@ impl Contains<<Runtime as frame_system::Config>::RuntimeCall> for CallFilter {
             RuntimeCall::Content(content::Call::<Runtime>::update_channel_privilege_level {
                 ..
             }) => false,
-            RuntimeCall::Content(content::Call::<Runtime>::update_channel_nft_limit { .. }) => {
```
mnaamani: Is it essential to enable `update_channel_nft_limit` for creator tokens to work?

runtime/src/lib.rs
```diff
@@ -240,16 +240,9 @@ impl Contains<<Runtime as frame_system::Config>::RuntimeCall> for CallFilter {
...
-                false
-            }
-            RuntimeCall::Content(
```
mnaamani: Is `set_channel_paused_features_as_moderator` essential for creator tokens to be released?

tests/network-tests/src/scenarios/content-directory.ts
```diff
@@ -33,5 +33,6 @@ scenario('Content directory', async ({ job }) => {
     'curators and collaborators permissions',
     collaboratorAndCuratorPermissions
   ).after(videoCommentsAndReactionsJob)
-  job('direct channel payment by members', directChannelPayment).after(collaboratorAndCuratorPermissionsJob)
+  // TODO fix: https://github.com/Joystream/joystream/issues/4756
```
mnaamani: Seems this comment can be removed now?

types/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@joystream/types",
-  "version": "3.0.0",
-  "description": "Types for Joystream Substrate Runtime - Nara release",
+  "version": "2.2.0",
```
mnaamani: Please revert back to 3.0.0. Hmm didn't we publish 4.0.0 which was originally for CRT.. In that case this will need to be 4.1.0 we will just have to skip 3.x

query-node/chain-metadata/1001.json
mnaamani, line 1: What happened here? This file should not change.

query-node/chain-metadata/2001.json
mnaamani, line 1: What happened here? This file should not change.

query-node/mappings/src/content/channel.ts
```diff
@@ -86,14 +87,17 @@ import {
...
 
 export async function content_ChannelCreated(ctx: EventContext & StoreContext): Promise<void> {
-  const { store, event } = ctx
+  const { store, event, block } = ctx
+  const { specVersion } = block.runtimeVersion
+
   // read event data
   const [channelId, { owner, dataObjects, channelStateBloatBond }, channelCreationParameters, rewardAccount] =
-    new ChannelCreatedEvent_V1001(event).params
+    specVersion < 2002 ? new ChannelCreatedEvent_V1001(event).params : new ChannelCreatedEvent_V2002(event).params
```
mnaamani: It doesn't look like the `ChannelCreated` event was changed in this PR.. or `nara`, So why is this needed? Is it related possible to `1001.json` being modified? Perhaps since work on crt was started such a long time ago, the spec version was 1001 at that time..?

query-node/mappings/src/content/channel.ts
```diff
@@ -160,9 +164,12 @@ export async function content_ChannelCreated(ctx: EventContext & StoreContext):
...
 
 export async function content_ChannelUpdated(ctx: EventContext & StoreContext): Promise<void> {
-  const { store, event } = ctx
+  const { store, event, block } = ctx
+  const { specVersion } = block.runtimeVersion
+
   // read event data
-  const [, channelId, channelUpdateParameters, newDataObjects] = new ChannelUpdatedEvent_V1001(event).params
+  const [, channelId, channelUpdateParameters, newDataObjects] =
+    specVersion < 2002 ? new ChannelUpdatedEvent_V1001(event).params : new ChannelUpdatedEvent_V2002(event).params
```
mnaamani: Same issue here I think, the 1001,2001 .json files are incorrect.

mnaamani (changes requested):
I had to stop in middle of review to bring up mainly issue with changes being done in query-node mapping. They are not wrong, but seems most of the changes related to picking the event type based on runtime spec version is because you have wrong metadata in files 1001.json and 2001.json

Can we confirm this is not a general issue for event types in nara related to new version of substrate?

distributor-node/package.json
dobertRowneySr: It was due to the `yarn build:packages` failing

runtime/src/lib.rs
dobertRowneySr: No, it is probably due to older modifications. 
I have it re-filtered again

runtime/src/lib.rs
dobertRowneySr: No, it is probably due to older modifications. 
I have it re-filtered again

types/package.json
dobertRowneySr: I have noticed that setting "^4.0.0" doesn't cause the CI to break
https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

distributor-node/package.json
dobertRowneySr: Setting it to ^4.0.0 fixes the CI

query-node/chain-metadata/1001.json
dobertRowneySr: fixed by copying 1001.json and 2001.json from nara https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

query-node/chain-metadata/2001.json
dobertRowneySr: fixed by copying 1001.json and 2001.json from nara https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

query-node/mappings/src/content/channel.ts
dobertRowneySr: fixed by copying 1001.json and 2001.json from nara https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

query-node/mappings/src/content/channel.ts
dobertRowneySr: fixed by copying 1001.json and 2001.json from nara https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

tests/network-tests/src/scenarios/content-directory.ts
dobertRowneySr: fixed in https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

runtime/src/lib.rs
dobertRowneySr: here https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

runtime/src/lib.rs
dobertRowneySr: here https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

distributor-node/package.json
dobertRowneySr: here https://github.com/Joystream/joystream/pull/4871/commits/ead0209600d711681888c6aff1ed222e07422b4e

runtime-modules/content/src/lib.rs
```diff
@@ -3675,6 +3633,98 @@ decl_module! {
...
+            // Call to ProjectToken
+            T::ProjectToken::deactivate_amm(
+                token_id,
+                member_id,
+            )?;
+
+        }
+
+        /// Allow crt issuer to update metadata for an existing token
+        #[weight = 100_000_000] // TODO: Adjust weight
```
mnaamani: Do we need a benchmarked weight function for this?

query-node/mappings/src/content/channel.ts
mnaamani: Ok, but technically the code now is suggesting that there is a difference in types. Its not a big deal but it might confuse the reader.

query-node/mappings/src/content/channel.ts
mnaamani: Same here, v1001 of the event will work in nara.

query-node/mappings/src/proposals.ts
```diff
@@ -333,9 +337,15 @@ async function handleRuntimeUpgradeProposalExecution(event: SubstrateEvent, stor
...
-export async function proposalsCodex_ProposalCreated({ store, event }: EventContext & StoreContext): Promise<void> {
+export async function proposalsCodex_ProposalCreated({
+  store,
+  event,
+  block,
+}: EventContext & StoreContext): Promise<void> {
+  const specVersion = block.runtimeVersion.specVersion
   const [proposalId, generalProposalParameters, runtimeProposalDetails, proposalThreadId] =
-    new ProposalCreatedEvent_V1001(event).params
+    specVersion >= 2002 ? new ProposalCreatedEvent_V2002(event).params : new ProposalCreatedEvent_V1001(event).params
```
mnaamani, line 347: Here again, the `ProposalCreated` event hasn't changed since v1001, but since we are planning a new proposal type to be added for freezing, `ProposalDetails` will change (which is one of the data elements of `ProposalCreated`) we can keep this change.

runtime-modules/project-token/src/lib.rs
```diff
@@ -1622,6 +1848,11 @@ impl<T: Config> Module<T> {
         <T as Config>::ModuleId::get().into_sub_account_truncating(Vec::<u8>::new())
     }
 
+    /// Returns the account for the AMM treasury
+    pub fn amm_treasury_account(token_id: T::TokenId) -> T::AccountId {
```
mnaamani, line 1914: This is a new "module" account that was not initialized at genesis. I think we need a runtime migration to give it an existential deposit.

runtime-modules/project-token/src/lib.rs
```diff
@@ -161,11 +146,27 @@ decl_storage! { generate_storage_info
...
+        pub AmmBuyTxFees get(fn amm_buy_tx_fees) config(): Permill;
+
+        /// AMM sell transaction fee percentage
+        pub AmmSellTxFees get(fn amm_sell_tx_fees) config(): Permill;
+
+        /// Max patronage rate allowed
+        pub MaxYearlyPatronageRate get(fn max_yearly_patronage_rate) config(): YearlyRate = YearlyRate(Permill::from_percent(15));
+
+        /// Minimum slope parameters allowed for AMM curve
+        pub MinAmmSlopeParameter get(fn min_amm_slope_parameter) config(): TokenBalanceOf<T>;
```
mnaamani: Just to be clear, all these new storage values will have value of zero (with exception of `max_yearly_patronage_rate`) after the runtime upgrade as they are not initialized at genesis.
If we need them to have some other initial value we will want a runtime migration.

mnaamani (changes requested):
Left a few more comments.
The most important issue however is the need to initialize the `amm_treasury_account` with existential deposit which we do for all module accounts, with an on_runtime_update migration.

types/package.json
mnaamani: So that would be using the published version to npm (the yarn.lock is modified), not the version in the monorepo.
Setting the dependency in cli to ^4.0.0 is okay, but the types package in the monorepo should be updated to 4.1.0 then probably need to revert change in yarn.lock

runtime-modules/project-token/src/migrations.rs
```diff
@@ -0,0 +1,78 @@
...
+        fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+            let onchain = Pallet::<T>::on_chain_storage_version();
+
+            ensure!(onchain < 1, "this migration can be deleted");
+
+            let module_account_id = Module::<T>::module_treasury_account();
+            let deposit = T::JoyExistentialDeposit::get();
+            let treasury_free_balance = Joy::<T>::free_balance(&module_account_id);
+
+            ensure!(treasury_free_balance >= Joy::<T>::existential_deposit());
```
mnaamani: Does it make sense to assert this? The whole point is that this is a new account which would have no free balance.. So this will fail when run against mainnet chain..If anything you would be asserting the free balance is in-fact zero.

runtime-modules/project-token/src/migrations.rs
```diff
@@ -0,0 +1,78 @@
+use super::*;
+use frame_support::traits::OnRuntimeUpgrade;
+
+// syntactic sugar for logging.
+#[macro_export]
+macro_rules! log {
+	($level:tt, $patter:expr $(, $values:expr)* $(,)?) => {
+		log::$level!(
```
mnaamani: I think you are missing the log crate dependency in Cargo.toml and LOG_TARGET definition

runtime-modules/project-token/src/migrations.rs
mnaamani: Besides that, the ensure!() macro needs a second argument which is the error if assertion fails.

yarn.lock
```diff
@@ -2347,6 +2347,19 @@
     yaml "^1.10.0"
     yaml-validator "^3.0.0"
 
+"@joystream/types@^4.0.0":
+  version "4.0.0"
+  resolved "https://registry.yarnpkg.com/@joystream/types/-/types-4.0.0.tgz#b7019466e5fb82fe4a97d46fcbcf840e19ab3ff6"
+  integrity sha512-3NE+sQC9cXBudYIPrCJCDdM4N9p6a4/XLUCjaZLaHwiD1Ds7Dbsfwjbfw67wU1m2V9hUeQCacV/ixumjcOIBsw==
```
mnaamani: run `yarn` in the root to remove this entry in yarn.lock

runtime/src/lib.rs
```diff
@@ -901,6 +900,7 @@ parameter_types! {
     pub const ProjectTokenModuleId: PalletId = PalletId(*b"mo:token"); // module: token
     pub const MaxVestingSchedulesPerAccountPerToken: u32 = 5;
     pub const BlocksPerYear: u32 = 5259600; // 365,25 * 24 * 60 * 60 / 6
+    pub const MaxOutputs: u32 = 256; // TODO(Martin, Ignazio) : find a suitable value
```
mnaamani: Can you make an issue for this if this is not the final value?

tsconfig.json
```diff
@@ -1,3 +1,22 @@
 {
-
+    "compilerOptions": {
```
mnaamani: What problem is this solving?

runtime-modules/project-token/Cargo.toml
```diff
@@ -5,28 +5,35 @@ authors = ['Joystream contributors']
...
+codec = { package = 'parity-scale-codec', version = '3.1.2', default-features = false, features = [
+	'derive',
+] }
+serde = { version = '1.0.101', features = ['derive'], optional = true }
+scale-info = { version = "2.1.1", default-features = false, features = [
+	"derive",
+] }
+
+# util
+log = { version = "0.4.17", default-features = false }
```
mnaamani: This is no longer used, it was used at one point in the migrations.rs module which has now been removed.

mnaamani (changes requested):
LGTM
just some minor points.

mnaamani (approved):

Issue #4589, opened 2023-01-30
Restrict content curator deletion powers


bedeho:
# Background

Currently, moderators can do the following deletions

1. `delete_video_as_moderator`: requires that video has no NFT
2. `delete_channel_as_moderator`: requires that channel has no videos, and also in the future no creator token.
3. `delete_channel_assets_as_moderator`: can always be done
4. `delete_video_assets_as_moderator`: can always be done

There is a big distinction between 1&2 vs. 3&4. The former group really means that a publisher who wants to can effectively _block_ curators from removing these objects by making sure requirements are not satisfied (have NFTs, videos and creator tokens). This group also the most contentious forms of moderation, where you basically accept that the DAO can _remove_, in principle, anyone for any reason. The latter group not only cannot be blocked by the publisher, but it is also the case where lack of moderation could not even be an option in theory. Even leaving the side of the economic ROI of certain content being possibly so low as not being worth while, other content would be effectively toxic for the entire system, and it would be catastrophic for the infra operators to have to host or distribute.

This combination of characteristics leads me to think that its best to deprecate 1&2. Their presence undercuts the immutable ownership property the ledger can offer - which many will find attractive, while simultaneously not actually protecting the ledger against an active adversary. The worst thing someone could do when 1&2 is removed - which they still could do if it was there, is to have metadata which somehow is considered out of policy, for example by linking to resources outside the Joystream Network itself, e.g. a hyperlink somewhere. This seems to me to be both unavoidable, but also not really a problem at all.

# Proposal

Propose plan for how to most gracefully deprecate 1&2, with full consideration of possible requirements to migrate permissions, featuring blocking and other representations that currently assume their presence.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1204404644374030) by [Unito](https://www.unito.io)

PR #4874, opened 2023-09-15
[Nara] Restric curators deletion powers

mnaamani:
Implements https://github.com/Joystream/joystream/issues/4589

- [x] Dropped `delete_video_as_moderator`
- [x] Dropped `delete_channel_as_moderator`
- [x] Regenerated chain metadata and types
- [x] Bumped types version to `v3.0.0` (convention to bump major version on each new runtime version)
- [x] Removed use of extrinsics in cli,integration tests
- [x] Update benchmarks and weights (remove unused weight functions related to dropped extrinsics
- [x] Remove `ContentModerationAction::DeleteVideo`, `ContentModerationAction::DeleteChannel` variants, and add migration to update permissions by level of each curator group in `content::CuratorGroupById` storage map.

Includes work from https://github.com/Joystream/joystream/pull/4875

The runtime migration can be tested against the live network by running:

```sh
RUN_MIGRATION_TESTS=true RUST_LOG=info cargo +nightly-2022-11-15 \
    test remote_tests::run_migrations --release --features try-runtime
```

The runtime upgrade can also be tested against any other chain with:

```sh
WS=ws://127.0.0.1:9944/ RUN_MIGRATION_TESTS=true RUST_LOG=info cargo +nightly-2022-11-15 \
    test remote_tests::run_migrations --release --features try-runtime
```

or

```sh
cargo +nightly-2022-11-15 build --release --features try-runtime
RUST_LOG=info,runtime=debug ./target/release/joystream-node try-runtime \
   --runtime ./target/release/wbuild/joystream-node-runtime/joystream_node_runtime.compact.compressed.wasm \
   on-runtime-upgrade live --uri ws://127.0.0.1:9944/
```

We have a problem with the query-node. We have already solved how to allow processor to handle different version of an event type after runtime upgrade, however it cannot handle an event being removed entierly from a runtime.
So for now the `ChannelDeletedByModerator` and `VideoDeletedByModerator` events are still part of the runtime despite no longer being emitted, only to be able to handle old runtime events. - I guess that is one more reason to move to subsquid.. (not like we needed any more reasons) - created an issue https://github.com/Joystream/joystream/issues/4877 -> resolved in https://github.com/Joystream/joystream/pull/4875 and https://github.com/Joystream/hydra/pull/531

With respect to mappings related to curator permissions there are none specific to `ContentModerationAction`. Query node is only keeping track of the "channel agent" permissions. So no changes to mappings is required.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205538268914413) by [Unito](https://www.unito.io)


runtime-modules/content/src/lib.rs
```diff
@@ -90,6 +91,25 @@ use sp_std::{borrow::ToOwned, collections::btree_set::BTreeSet, vec::Vec};
 
 type WeightInfoContent<T> = <T as Config>::WeightInfo;
 
+/// The log target of this pallet.
+pub const LOG_TARGET: &str = "runtime::content";
+
+// syntactic sugar for logging.
+#[macro_export]
+macro_rules! log {
```
dobertRowneySr: maybe I would put this in the `migration.rs` file where it is used

dobertRowneySr (commented):
I have no major remarks.
I am just wandering: shouldn't we replace the runtime migration tests in the CI with the `try_runtime` ones?

dobertRowneySr (changes requested):
> I have no major remarks.
I am just wandering: shouldn't we replace the runtime migration tests in the CI with the `try_runtime` ones?

runtime-modules/content/src/migrations.rs
```diff
@@ -0,0 +1,203 @@
...
+
+            <CuratorGroupById<T>>::translate_values(
+                |old_value: CuratorGroupV0<T>| -> Option<CuratorGroup<T>> {
+                    log!(info, "Migrating group {:?}", old_value);
+                    let mut permissions_by_level = ModerationPermissionsByLevel::<T>::new();
+                    old_value
+                        .permissions_by_level
+                        .iter()
+                        .for_each(|(level, permissions)| {
+                            let mut permissions_v1 = BTreeSet::new();
```
dobertRowneySr, line 131: Also maybe add a `log!` statement here

runtime-modules/content/src/lib.rs
mnaamani: I will try, although I think the macro export statement might only work in the crate root.
Although we don't use the log statement anywhere else, it could be useful if we want to add some logs in future?

mnaamani:
> I am just wandering: shouldn't we replace the runtime migration tests in the CI with the `try_runtime` ones?

Yes we do here: 
https://github.com/Joystream/joystream/actions/runs/6245191588/job/16953377405?pr=4874#step:7:2
and
https://github.com/Joystream/joystream/actions/runs/6245191588/job/16953377405?pr=4874#step:8:2

runtime-modules/content/src/migrations.rs
mnaamani: Done in https://github.com/Joystream/joystream/pull/4874/commits/e759b320d6432bb5f4fa0e8c3ece52249a0c27f0

dobertRowneySr (approved):
LGTM

runtime-modules/content/src/migrations.rs
```diff
@@ -0,0 +1,219 @@
...
+                            log!(warn, "Failed to add curator to group");
+                        }
+                    });
+                    Some(group)
+                },
+            );
+
+            current.put::<Pallet<T>>();
+
+            <T as frame_system::Config>::BlockWeights::get().max_block
```
bedeho, line 193: Why is this set to max? and is this the right notion of max? is the idea here to exclude all other processing in this block, so as to avoid having to come up with an accurate weight function to begin with?

runtime-modules/content/src/migrations.rs
```diff
@@ -0,0 +1,219 @@
...
+                    log!(info, "Migrating group {:?}", old_value);
+                    let mut permissions_by_level = ModerationPermissionsByLevel::<T>::new();
+                    old_value
+                        .permissions_by_level
+                        .iter()
+                        .for_each(|(level, permissions)| {
+                            log!(info, "level: {:?}, permissions: {:?}", level, permissions);
+                            let mut permissions_v1 = BTreeSet::new();
+                            permissions.iter().for_each(|action| match action {
+                                ContentModerationActionV0::HideVideo => {
```
bedeho, line 133: I think it would have been better to separate out conversion from writing/translation, but don't think its worth fiddling with now.

runtime-modules/content/src/lib.rs
```diff
@@ -90,6 +91,14 @@ use sp_std::{borrow::ToOwned, collections::btree_set::BTreeSet, vec::Vec};
 
 type WeightInfoContent<T> = <T as Config>::WeightInfo;
 
+/// The log target of this pallet.
+pub const LOG_TARGET: &str = "runtime::content";
+
+// Nara release. enum variants removed:
+// - ContentModerationAction::DeleteVideo
+// - ContentModerationAction::DeleteChannel
+const CURRENT_STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
bedeho, line 100: Rather than using these opaque version identifiers, what if we used some more explicit representation which would be easier to self-document, like they seemingly do in Substrate frame pallets: 

https://paritytech.github.io/polkadot-sdk/master/src/pallet_staking/migrations.rs.html#36

I guess we could could use release names? like nara, etc.?

runtime-modules/content/src/migrations.rs
```diff
@@ -0,0 +1,219 @@
...
+            ensure!(onchain < 2, "this migration needs to be removed");
+            ensure!(onchain == 1, "this migration needs to be run");
+            let group_count = CuratorGroupById::<T>::iter().count();
+            log!(
+                info,
+                "Number of curator groups post_upgrade {}",
+                group_count
+            );
+            let pre_upgrade_group_count = usize::from_be_bytes(state.try_into().unwrap());
+            ensure!(
```
bedeho, line 208: I'm not sure I understand the consequences of this sort of check at runtime, is that really appropriate? I see the case for asserting this in debug mode or something, but if there is some sort of bug, is there any added profit from failing this hook? what happens then?

runtime-modules/content/src/migrations.rs
```diff
@@ -0,0 +1,219 @@
...
+            let onchain = Pallet::<T>::on_chain_storage_version();
+            let current = Pallet::<T>::current_storage_version();
+
+            if onchain > 0 {
+                return T::DbWeight::get().reads(1);
+            }
+
+            use iterable_enums::ContentModerationActionV0;
+
+            <CuratorGroupById<T>>::translate_values(
```
bedeho, line 122: Are we not concerned about this, for some reason being gigantic by some weird accident or attack, and then trying to iterate it all in one block? Seems like this iteration must be either unrolled across blocks if it is truly to be dynamic, or it has to have a hardcap which we know allows compute to finish in time.

bedeho (approved):
So this looks like excellent work, but I have to admit that getting back to this really showed me how stale my Substrate skills are, and with limited time, my due dilligence is of limited value unfortunately. I'm ok with merging this, but also asked for a superficial change and a question to make sure I'm understanding properly.

I'm glad there is a somewhat standardised approach to migrations in Substrate now, it looks like a nice approach which can work for complex changes.

mnaamani:
Will address your questions bedeho, but will merge for Ignazio's sake.

runtime-modules/content/src/migrations.rs
mnaamani: When we build the production runtime we will not build it with the `try-runtime` feature, so the pre_/post_ methods will not exist and will not execute.

The feature is meant specifically for testing a new runtime upgrade during development against some existing chain state.

runtime-modules/content/src/lib.rs
mnaamani: If you look carefully at the `enum ObsoleteReleases` you will note they are actually phasing out that pattern for the more common convention I saw used in multiple other pallets. eg. https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/assets/src/migration.rs

Although I agree a `u32` it is not very self documenting which is why I used the `nara` name for the module.

runtime-modules/content/src/migrations.rs
mnaamani: You are right, I didn't invest too much time on code style here.

runtime-modules/content/src/migrations.rs
mnaamani: I was going through a lot of the migrations which we are including the nara update and saw more than one returning this max block weight. And since one of the migrations was doing it, it become almost irrelevant what was returned here. 

As far as I know the amount of block space allocated for on_runtime_upgrade is not constrained, but accurate weight computation would be important to know if we were planning a migration of a large map for example. This way we could plan ahead and decide if the total weight was larger than what validators could produce in required time.

Found this relevant discussion https://github.com/paritytech/substrate/issues/10064

runtime-modules/content/src/migrations.rs
mnaamani: That is a great point, I guess it also goes back to the discussion here https://github.com/Joystream/joystream/pull/4874#discussion_r1344448892

I believe there is already a limit on the number of curators per group but not on number of groups.
So yes it could be a problem if the content lead decides to create a very large number of groups before the runtime upgrade.

Should we tackle this possibility ?

runtime-modules/content/src/migrations.rs
mnaamani: But to answer your question more clearly, yes by returning [max_block](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/system/src/limits.rs#L203) in this (or any other migration) will ensure that the block being executed (the first block with the new runtime active) will prevent any extrinsics from being included. So only migrations, on_initialize and on_finalize logic will execute.

The value is constructed in https://github.com/Joystream/joystream/blob/14b0fa39825209dc66264123f017fc51cd33c07f/runtime/src/lib.rs#L186

Issue #4822, opened 2023-08-02
runtime: increase storage + distribution `maxWorkerNumberLimit`


mochet:
# Problem
Currently the storage WG is experiencing a rapid requirement to grow, and one of the limitations is the maximum number of workers. There are some other avenues of improving this already (such as allowing storage workers to increase the number of servers that they can run as one worker). Regardless, it seems like considering an increase to the `maxWorkerNumberLimit` is worth considering for the next runtime upgrade.
And although there are no current issues with the Distributor WG, it would seem prudent to also consider this WG along the same lines.

# Solution
* The current `maxWorkerNumberLimit` for both Storage + Distributors is `30`
* Increase it to 50? maybe even as high as 100?
* I'm not sure what the scaling limitations are here, but it could be that some of the operations WGs have their `maxWorkerNumberLimit` decreased to balance this since those groups do not have distinct on-chain permissions like Storage + Distribution--they also have the option of paying non-workers using discretionary spending from their WG.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205195329012145) by [Unito](https://www.unito.io)

freakstatic:
# TODO
- [ ] Define new max worker limit
- [x] Runtime change with the value `50`
- [x] Add integration test to check new value

bedeho:
We decided to introduce a new constant for each working group, only distributors and storage providers should be bumped to 50,all others remain at 30.

PR #4882, opened 2023-09-21
[Nara] Increase workers runtime limit

freakstatic:
Implements #4822


mnaamani (approved):
We should probably do some tests with payouts when the max number of workers is hired. However I think we should be fine since payouts to different groups is spaced out by 10 blocks.

mochet:
> We should probably do some tests with payouts when the max number of workers is hired. However I think we should be fine since payouts to different groups is spaced out by 10 blocks.

Over a long enough time period wouldn't there be times when two or more working group payment periods line up and occur on the same block?

Issue #4866, opened 2023-09-12
Runtime Upgrade to increase council duration


chrlschwb:
We are planning on a runtime upgrade to change the following constants.**Council Term Duration** (https://github.com/Joystream/joystream/blob/d1448236349154afa2c1a725f7144990c37d5042/runtime/src/lib.rs#L935)【 code  】

pub const VoteStageDuration: BlockNumber = days!(6);

pub const RevealStageDuration: BlockNumber = days!(4);

pub const AnnouncingPeriodDuration: BlockNumber = days!(4);

pub const IdlePeriodDuration: BlockNumber = days!(14);

【 end code 】

**Forum Subcategory Limit** 

(https://github.com/Joystream/joystream/blob/d1448236349154afa2c1a725f7144990c37d5042/runtime/src/lib.rs#L1227)

```
MaxDirectSubcategoriesInCategory: u64 = 10;
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205488807612673) by [Unito](https://www.unito.io)

chrlschwb:
Endpoint for the chain testing:

Pioneer:
https://pioneer.joystreamstats.live/?network-config=https://pl.joystreamstats.live/joy/config.json

RPC: 
wss://longer-idle-period.joystreamstats.live/rpc

Graphql: 
https://longer-idle-period.joystreamstats.live/graphql

bedeho:
Actual numbers are not final.

freakstatic:
# TODO
- [x] Define new council/election stages durations
- [x] Runtime change
-  [ ] ~~-Add integration test to check new value~~

freakstatic:
Parameters values voted here:
https://pioneerapp.xyz/#/proposals/preview/547

**Idle Period**:    14 days
**Announcing Period**:    6 days
**Voting Period**:    4 days
**Revealing Period**:    4 days

PR #4884, opened 2023-09-24
[Nara] Change referendum + council parameters

freakstatic:
Implements #4866

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -946,15 +946,15 @@ pub type CouncilModule = council::Module<Runtime>;
 parameter_types! {
     // referendum parameters
     pub const MaxSaltLength: u64 = 32;
-    pub const VoteStageDuration: BlockNumber = days!(3);
-    pub const RevealStageDuration: BlockNumber = days!(3);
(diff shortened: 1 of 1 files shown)


runtime/src/runtime_api.rs
```diff
@@ -89,6 +90,55 @@ impl Get<&'static str> for StakingMigrationV11OldPallet {
     }
 }
 
+pub struct ReferendumParametersCheck;
+impl OnRuntimeUpgrade for ReferendumParametersCheck {
+    fn on_runtime_upgrade() -> Weight {
+        let voting_stage_duration =
+            <Runtime as referendum::Config<ReferendumInstance>>::VoteStageDuration::get();
+        assert!(
```
mnaamani: We definitely do not want to `assert` here, this will halt the chain.
I would recommend leaving out any of these "tests" from the runtime.

mnaamani (changes requested):
New values look correct as decided by DAO. But please leave out the migrations, they are not the correct way to test these new values.

mnaamani (approved):

PR #4886, opened 2023-09-25
Docker images: selectively publish package

mnaamani:
Update workflow to make it possible to select which package to publish to docker hub instead of always publishing all packages.


mnaamani:
When buildx is cross-compiling for arm64 it seems to have network issues and the build can't fetch packages from yarn...
```
#29 431.6 info There appears to be trouble with your network connection. Retrying...
```

zeeshanakram3 (approved):
LGTM.

PR #4888, opened 2023-09-26
[Nara] add new workers limits

freakstatic:
Implements #4822


freakstatic:
@mnaamani please review

mnaamani (approved):
Nice work.

Just a note, the benchmarks for on_initialize use the membership working group max worker limit to do the analysis, but should still be accurate enough for the larger groups.

Issue #4894, opened 2023-09-28
WIP: Fast Runtime Configuration


bedeho:
WIP: Mokhtar will compelet spec., freakstatic executes

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205602859471050) by [Unito](https://www.unito.io)

mnaamani:
The idea here was to introduce a runtime configuration that could be used to run a test/staging network for QA testing where runtime and storage based configuration would match the "production" runtime, with one key difference. The block production interval would be 1s instead of 6s, but any period lengths (defined in blocks) (provided they are not hard coded but actually use helper macros (days!, minutes!, hours!) for council/election periods, proposal voting/gracing periods etc would still be computed based on 6s blocks. This would mean time would be accelerated or "warped" . 

I've implemented this in https://github.com/Joystream/joystream/pull/4913

Note there is already a way to run a local development chain with production runtime settings if all that is needed is to verify/confirm constants are as expected by running node with the following arguments:

```
joystream-node --chain prod-test --tmp --alice --rpc-cors=all --pruning=archive
```

PR #4913, opened 2023-10-06
runtime: add warp-time feature for "fast" runtime

mnaamani:
Added a "warp-time" cargo feature for use with production runtime configuration to address https://github.com/Joystream/joystream/issues/4894

To test build and run local chain (with production configration)

```sh
WASM_BUILD_TOOLCHAIN=nightly-2022-11-15 cargo build --release --features warp-time
```

```
./target/release/joystream-node --chain prod-test --tmp --alice --rpc-cors=all --pruning=archive
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205678974517544) by [Unito](https://www.unito.io)

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -226,12 +226,26 @@ runtime-benchmarks = [
     "project-token/runtime-benchmarks",
 ]
 
-# Staging and testing configurations
-# configuration suitable for staging networks and playground
+# Configuration suitable for staging networks and playground
 staging-runtime = []
 playground-runtime = []
-# configuration suitable for integration testing
-testing-runtime = []
+# Configuration suitable for automated integration testing only
+testing-runtime = [
+    "fast-block-production"
+]
+
+# Default block production interval is 6s. Enabling this feature will configure chain for 1s interval instead.
+fast-block-production = []
+
+# Configures 1s block interval for the chain, but changes behaviour of `minutes!`, `hours!`, `days!` util macros
+# to compute periods (in blocks) based on 6s block interval.
+# This configuration is meant for QA (manual testing) of production runtime configuration.
+# that has long time periods for things like council elections, proposals voting periods etc.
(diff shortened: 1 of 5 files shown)


freakstatic (commented):
@mnaamani are you sure you want to have "two names" for the same "feature"?
warp-time vs fast-block-production

runtime/Cargo.toml
```diff
@@ -226,12 +226,26 @@ runtime-benchmarks = [
...
-# Staging and testing configurations
-# configuration suitable for staging networks and playground
+# Configuration suitable for staging networks and playground
 staging-runtime = []
 playground-runtime = []
-# configuration suitable for integration testing
-testing-runtime = []
+# Configuration suitable for automated integration testing only
+testing-runtime = [
+    "fast-block-production"
```
freakstatic, line 234: will not this make testing-runtime with fast-block-production by default?

freakstatic:
oh maybe I misunderstood, so fast-block-production it's the regular testing environment and warp-time is the one with block production of 1s right?

freakstatic (approved):

runtime/Cargo.toml
mnaamani: That is already the case. The testing runtime uses 1s blocks (but doesn't warp the timing)

mnaamani:
> oh maybe I misunderstood, so fast-block-production it's the regular testing environment and warp-time is the one with block production of 1s right?

~Correct.~
Not exactly.

`testing-runtime` feature and `warp-time` both enable 1s block times
testing-runtime uses different runtime constants for council/election/proposal periods than production.
warp-time if enabled with production runtime will use the production runtime constants and for periods which are defined using hours!,minutes!,days! macros will not be real minutes,hours,days .. they will be warped times.. so for example `hours!(1)` will be number of blocks (based on 6s) so 60 * 60 / 6 =  600 blocks. But since a blocks are actually at 1s intervals, that is actually 600 / 60 = 10 (real) minutes.

Issue #4905, opened 2023-10-02
QN: Validator profile scope changes


thesan:
Follow-up to #4823

### Context
After discussing the Pioneer validator page it appeared that the initial scope was wrong:
1. The Validator accounts are bound to a membership profile using the `members.addStakingAccountCandidate` / `members.confirmStakingAccount` extrinsics. As a result the accounts are not stored in the QN (they should be just retrieved directly from the chain). So both QN and protobufs `validatorAccount` fields introduced in #4849 should be removed.
2. The membership profile is what is being verified not the validator account. So the `isVerifiedValidator` from #4849 should be kept but it should be reset to `false` whenever any membership metadata field is changed.
3. The verification is the membership working group responsibility (not the HR group).

### Scope
- [ ] @chrlschwb  please address (3) in the currently in progress #4868 PR.
- [ ] (1) and (2) can be addressed in a new PR.

eshark9312:
@thesan 
> 2. The membership profile is what is being verified not the validator account. So the isVerifiedValidator from https://github.com/Joystream/joystream/pull/4849 should be kept but it should be reset to false whenever any membership metadata field is changed.
> 3. The verification is the membership working group responsibility (not the HR group).

AFAIK, there is no tx to verify the validator membership with `membership working group` role.
Do we need to update the runtime?
Can we name the transaction `api.tx.membershipWorkingGroup.verifyValidatorMembership` or `verifyValidatorProfile`?

thesan:
> AFAIK, there is no tx to verify the validator membership with `membership working group` role. Do we need to update the runtime? Can we name the transaction `api.tx.membershipWorkingGroup.verifyValidatorMembership` or `verifyValidatorProfile`?

No need to modify the runtime because this verification is pure metadata (it only leaves in the QN). Here's the issue for it: https://github.com/Joystream/joystream/issues/4824 and here's the PR https://github.com/Joystream/joystream/pull/4868.

In short these extrinsics will be used: ~~`api.tx.operationsWorkingGroupBeta.workerRemark(workerId, msg)`~~ `api.tx.membershipWorkingGroup.workerRemark(workerId, msg)` and ~~`api.tx.operationsWorkingGroupBeta.leadRemark(msg)`~~  `api.tx.membershipWorkingGroup.leadRemark(msg)`

PR #4915, opened 2023-10-09
Remove validator account from the QN

chrlschwb:
#4905 

- QN and protobufs `validatorAccount` fields removed
- `isVerifiedValidator`is reset to `false` whenever any membership metadata field is changed.


tests/network-tests/src/fixtures/membership/UpdateProfileHappyCaseFixture.ts
```diff
@@ -56,8 +55,6 @@ export class UpdateProfileHappyCaseFixture extends BaseQueryNodeFixture {
       expected.externalResources?.map(asMembershipExternalResource) ?? []
     )
     assert.isFalse(Utils.hasDuplicates(metadata.externalResources?.map(({ type }) => type)))
-    assert.equal(metadata.isVerifiedValidator, false)
```
thesan, line 59: Please keep this `assert.equal(metadata.isVerifiedValidator, false)`. Even though it doesn't mean what we thought it did, this value should still be set to`false` when any metadata changes.

tests/network-tests/src/fixtures/membership/InviteMembersHappyCaseFixture.ts
```diff
@@ -75,7 +75,6 @@ export class InviteMembersHappyCaseFixture extends StandardizedFixture {
       assert.equal(about, metadata.about)
       assert.equal(inviteCount, 0)
       assert.equal(avatar?.avatarUri, metadata.avatarUri || undefined)
-      assert.equal(validatorAccount, metadata.validatorAccount)
```
thesan, line 78: Please check `assert.equal(metadata.isVerifiedValidator, false)` here too

tests/network-tests/src/fixtures/membership/GiftMembershipHappyCaseFixture.ts
```diff
@@ -81,7 +81,6 @@ export class GiftMembershipHappyCaseFixture extends StandardizedFixture {
       Utils.assert(entry.__typename === 'MembershipEntryGifted', 'Query node: Invalid membership entry method')
       Utils.assert(entry.membershipGiftedEvent)
       assert.equal(entry.membershipGiftedEvent.id, qEvent.id)
-      assert.equal(validatorAccount, metadata.validatorAccount)
```
thesan, line 84: And here too

query-node/mappings/src/workingGroups.ts
```diff
@@ -52,6 +52,7 @@ import {
   getWorkingGroupByName,
   getWorkingGroupLead,
   invalidMetadata,
+  getMemberById,
```
thesan:
```suggestion
```

metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -63,8 +63,14 @@ message ModeratePost {
...
 
+message VerifyValidator{
+  required uint64 memberId = 1;
+  required bool is_verified = 2;
+}
+
 message RemarkMetadataAction {
   oneof action {
     ModeratePost moderate_post = 1;
+    VerifyValidator verfy_validator = 2;
```
thesan: This is correct but it should be on #4868 not here (because you'll need it for the tests)

thesan (changes requested):

tests/network-tests/src/fixtures/membership/InviteMembersHappyCaseFixture.ts
chrlschwb: I can't add command. because isVerifiedValidator is not defined.

tests/network-tests/src/fixtures/membership/GiftMembershipHappyCaseFixture.ts
chrlschwb: here same

PR #4916, opened 2023-10-09
Joystream-node embed mainnet chainspec

mnaamani:
Embed `joy-mainnet.json` chainspec file in joystream-node binary to make it simpler to get started with running a node for mainnet.
So now only need to download binary, and then run a mainnet node like this:

`joystream-node --chain joy-mainnet`

rather than needing to separately download the `joy-mainnet.json` file and running with:

```
joystream-node --chain /path/to/joy-mainnet.json
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2737,7 +2737,7 @@ dependencies = [
 
 [[package]]
 name = "joystream-node"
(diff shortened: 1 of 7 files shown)


dobertRowneySr (approved):
LGTM

PR #4917, opened 2023-10-09
bash scripts: proper error handling when invoking sub shell scripts

mnaamani:
Depends on https://github.com/Joystream/joystream/pull/4913 and https://github.com/Joystream/joystream/pull/4935

General fix to all shell scripts to correctly exit script when calling a sub script fails.

TLDR:
we have `set -e` in scripts that needs to exit when a subscript fails.

Using this:
```sh
output=$( expression )
```
captures output but doesn't cause calling script to exit with error. (You have to manually check exit code with `$?`) 

```sh
output=`expression`
```

Works better because it captures output and a non zero exit code from the expression/script with make the calling script exit correctly with an error code.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205728346610756) by [Unito](https://www.unito.io)

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,4 @@
 COMPOSE_PROJECT_NAME=joystream
-PROJECT_NAME=query_node
 
 # We use a single postgres service for both the query node indexer and processor.
 # The default `DB_*` environment variables point to query node processor's database.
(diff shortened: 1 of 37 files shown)


dobertRowneySr (approved):
LGTM

PR #4921, opened 2023-10-09
[Argus] Fix distributor node syncing query

zeeshanakram3:
## Problem

The distributor node executes [getDistributionBucketsWithObjectsByWorkerId](https://github.com/Joystream/joystream/blob/0d2a068ecd7d4eb07189a48e0be1a1de6c740351/distributor-node/src/services/networking/NetworkingService.ts#L376C33-L376C76) query to get all the data objects that given node is supposed to distribute. The problem is that with the consistent growth of the storage directory, the response size of this query is becoming larger and larger. For some time this query was experiencing `Timeout`, upon investigating it turned out that while executing this query the graphql-server was consistently crashing with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

image

## Fix

This PR divides the given query into multiple smaller queries so that the graphql-server is successfully able to process it, until we find the reason for the memory leak happening in the graphql-server and create a proper fix.

### DIFF
--- a/distributor-node/src/services/networking/query-node/api.ts
+++ b/distributor-node/src/services/networking/query-node/api.ts
@@ -17,24 +17,28 @@ import ws from 'ws'
 import { LoggingService } from '../../logging'
 import {
   DataObjectDetailsFragment,
-  DistirubtionBucketWithObjectsFragment,
   GetActiveStorageBucketOperatorsData,
   GetActiveStorageBucketOperatorsDataQuery,
   GetActiveStorageBucketOperatorsDataQueryVariables,
   GetDataObjectDetails,
   GetDataObjectDetailsQuery,
   GetDataObjectDetailsQueryVariables,
-  GetDistributionBucketsWithObjectsByIds,
-  GetDistributionBucketsWithObjectsByIdsQuery,
-  GetDistributionBucketsWithObjectsByIdsQueryVariables,
-  GetDistributionBucketsWithObjectsByWorkerId,
-  GetDistributionBucketsWithObjectsByWorkerIdQuery,
-  GetDistributionBucketsWithObjectsByWorkerIdQueryVariables,
+  GetDataObjectsWithBagsByIds,
+  GetDataObjectsWithBagsByIdsQuery,
+  GetDataObjectsWithBagsByIdsQueryVariables,
+  GetDistributionBucketsWithBagsByIds,
+  GetDistributionBucketsWithBagsByIdsQuery,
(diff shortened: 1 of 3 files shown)


distributor-node/src/services/networking/query-node/api.ts
```diff
@@ -180,20 +184,48 @@ export class QueryNodeApi {
...
+    const allBagIds = [...bagIds] // Copy to avoid modifying the original array
+    const fullResult: StorageBagWithObjectsFragment[] = []
+    while (allBagIds.length) {
+      const bagIdsBatch = allBagIds.splice(0, 1000)
+      fullResult.push(
+        ...(await this.multipleEntitiesQuery<
+          GetDataObjectsWithBagsByIdsQuery,
+          GetDataObjectsWithBagsByIdsQueryVariables
+        >(GetDataObjectsWithBagsByIds, { bagIds: bagIdsBatch, limit: bagIdsBatch.length }, 'storageBags'))
+      )
```
kdembler, line 197: What do you think of adding a small delay between batches? Like 0.5s? I'd rather have this operation take slightly longer but not put such a heavy load on server

kdembler (approved):
Tested this and works properly! I've checked that the number of objects is same as the previous version.
You can also see less pressure on resources:
CleanShot 2023-10-09 at 21 45 08@2x

Thanks for your work @zeeshanakram3!

mnaamani (approved):
Great work.
I wouldn't be surprised if the storage-node needs a similar fix.
@kdembler suggested adding a small delay between fetching batches. That is probably a good idea. What might work better is to use generators, `yield`ing one batch at a time, allow consumer/caller to process the batch of objects before coming back for more.

I will merge then bump the version of argus and prepare a docker release (using the https://github.com/Joystream/joystream/pull/4886 branch)

bwhm:
> Great work. I wouldn't be surprised if the storage-node needs a similar fix. @kdembler suggested adding a small delay between fetching batches. That is probably a good idea. What might work better is to use generators, `yield`ing one batch at a time, allow consumer/caller to process the batch of objects before coming back for more.
> 
> I will merge then bump the version of argus and prepare a docker release (using the #4886 branch)

The storage node already uses pagination at least, and from my testing, it wasn't as bad. It would help the overall system load to bump the `syncInterval` from the current 1 minute...

mnaamani:
> The storage node already uses pagination at least, and from my testing, it wasn't as bad. It would help the overall system load to bump the `syncInterval` from the current 1 minute...

https://github.com/Joystream/joystream/pull/4924

Issue #4897, opened 2023-09-29
`FreezePallet` Proposal and pallet


bedeho:
# Background

We want to be able to allow the council to block the functioning of entire pallets on very short notice to deal with 0-day problems. The idea is then that this freezing will be in place until an upgrade can pass. This is an imperfect solution, but its still useful.

# Proposal

1. Introduce a new pallet `pallet_freezer` which simply stores an individual indicator of whether each of our pallets are enabled or not. This means that this pallet is coupled with whatever the set of pallets will be in the full runtime, which is fine. The storage representation can be something simple as `BtreeSet<FreezablePallet>` where `FreezablePallet` is an enum representing all the pallets that are freezable. There is no need for any extrinsics, but there does need to be a public getter and setter which can be used by other pallets to check if they are frozen or the proposal to update what pallets are frozen, respectively. Getter can have signature like `is_fronze(FreezablePallet)` and `set_freezer(BTreeMap<FreezablePallet, Bool>)`. 
2. Introduce a new proposal in the codex, it takes `BTreeMap<FreezablePallet, Bool>` as a proposal parameter.
3. Proposal parameters are needed!
4. Make sure to add a distinct getter trait that other pallets can use for their runtime configuration to dial out and check if they are frozen.

We can incorporate all pallets that are written as part of Joystream code base, but we don't need to actually implement sensitivity to the freezing off the bat, only really CRT freezing is needed.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205619440282237) by [Unito](https://www.unito.io)

bedeho:
**NOTE:**
- Do propossal and pallet in two distinct PRs.
- Remember integration tests as well!

bedeho:
We decided to use a different approach which is to locally store whether any given pallet is frozen or not as a configurable boolean that is consulted on all extrinsics and public methods in the pallet. So no new pallet is needed, however a cross plallet proposal is still needed.

freakstatic:
@bedeho should root still be allowed to perform an extrinsics if a pallet was frozen?

bedeho:
> @bedeho should root still be allowed to perform an extrinsics if a pallet was frozen?

We agreed on Nara call that it does not matter for Nara, because we are only covering project_token pallet, which does not have root calls. But for the future, probably root should not be frozen.

bedeho:
[...]

PR #4922, opened 2023-10-09
[Nara] Add freeze pallet extrinsic to CRT pallet

freakstatic:
Implements the freeze part of #4897

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205728875665994) by [Unito](https://www.unito.io)


freakstatic:
@mnaamani can you please review?

runtime-modules/project-token/src/lib.rs
```diff
@@ -803,6 +825,27 @@ decl_module! {
...
+        ///
+        /// <weight>
+        ///
+        /// ## Weight
+        /// `O (1)`
+        /// - DB:
+        ///    - O(1) doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = WeightInfoToken::<T>::freeze_pallet()]
+        pub fn freeze_pallet(origin, freeze: bool) {
```
dobertRowneySr:
```suggestion
        pub fn set_freezing_status(origin, freeze: bool) {
```
I would prefer this name (or similar ones possibly more expressive) , otherwise from the name it looks like you can set `frozen` state only

runtime-modules/project-token/src/lib.rs
```diff
@@ -1351,6 +1414,10 @@ impl<T: Config>
 
         Ok(sale.funds_collected)
     }
+
```
dobertRowneySr: Where is this used?
In any case you can use `Self::pallet_frozen()` directly
or `Self::ensure_unfrozen_state().is_err()`

runtime-modules/project-token/src/events.rs
```diff
@@ -155,5 +155,7 @@ decl_event! {
         /// - member id
         /// - number of tokens burned
         TokensBurned(TokenId, MemberId, Balance),
+
```
dobertRowneySr, line 158:
```suggestion
/// Pallet Frozen state toggled
/// Params:
/// - current frozen status (true | false)
```

runtime-modules/project-token/src/traits.rs
```diff
@@ -81,4 +81,6 @@ pub trait PalletToken<
 
     /// Establish whether the token has an unfinalized sale
     fn is_sale_unscheduled(token_id: TokenId) -> bool;
+
+    fn is_pallet_frozen() -> bool;
```
dobertRowneySr: Is this used somewhere, I couldn't find any extra instances

dobertRowneySr (changes requested):
I have reviewed the companion issue. But the implementation presented here is different from what's prescribed. Am I missing something?
Also it's still possible for a creator `issue_token`, because there is not frozen-pallet check. Is this desired?
What follows below are minor suggestions

runtime-modules/project-token/src/lib.rs
freakstatic: @ignazio-bovo what you think about `set_frozen_status`?

runtime-modules/project-token/src/lib.rs
freakstatic: well notice, I  forgot to delete :)

runtime-modules/project-token/src/traits.rs
freakstatic: I removed since it was not used

freakstatic:
> I have reviewed the companion issue. But the implementation presented here is different from what's prescribed. Am I missing something? Also it's still possible for a creator `issue_token`, because there is not frozen-pallet check. Is this desired? What follows below are minor suggestions

@ignazio-bovo thanks for the review 👍 
Yeah you are right, the objective was changed, please check
https://github.com/Joystream/joystream/issues/4897#issuecomment-1742799791

Regarding `issue_token` well notice, I forgot to protect that function as well, I fixed that and added a test to ensure that it was working as expected.

freakstatic:
@ignazio-bovo FYI the proposal part will be submitted in another PR

dobertRowneySr (approved):
LGTM

runtime-modules/common/src/lib.rs
```diff
@@ -69,6 +72,10 @@ pub trait StorageOwnership {
         + PartialEq;
 }
 
+pub trait FrezablePalletConfig {
```
mnaamani: I see this trait implemented for Test and Runtime, but the value is never used anywhere. So it should be dropped. Maybe you meant to use it as the default storage value for `PalletFrozen` ? As you probably figures out the default for bool is false so its not necessary.

runtime-modules/content/src/tests/creator_tokens/finalize_creator_token_sale.rs
```diff
@@ -213,3 +215,47 @@ fn successful_finalize_curator_channel_creator_token_sale_by_lead() {
...
+            existential_deposit + DEFAULT_CREATOR_TOKEN_SALE_UNIT_PRICE * amount,
+        );
+
+        let result = project_token::Module::<Test>::purchase_tokens_on_sale(
+            RuntimeOrigin::signed(SECOND_MEMBER_ACCOUNT_ID),
+            project_token::Module::<Test>::next_token_id() - 1,
+            SECOND_MEMBER_ID,
+            amount,
+        );
+        assert_noop!(result, project_token::Error::<Test>::PalletFrozen);
```
mnaamani: The use of `assert_noop!` here is equivalent to `assert_err!`
If you really want to assert no storage was modified the expression returning the result should be the argument to the macro like so:

```suggestion
        assert_noop!(
          project_token::Module::<Test>::purchase_tokens_on_sale(
            RuntimeOrigin::signed(SECOND_MEMBER_ACCOUNT_ID),
            project_token::Module::<Test>::next_token_id() - 1,
            SECOND_MEMBER_ID,
            amount),
          project_token::Error::<Test>::PalletFrozen
        );
```

runtime-modules/project-token/src/events.rs
```diff
@@ -155,5 +155,10 @@ decl_event! {
...
+        /// Pallet Frozen status toggled
+        /// Params:
+        /// - current frozen status (true | false)
```
mnaamani:
```suggestion
        /// - new frozen status (true | false)
```

runtime-modules/content/src/tests/creator_tokens/claim_patronage_credit.rs
```diff
@@ -93,3 +96,17 @@ fn claim_creator_token_patronage_credit_fails_during_trasfer() {
...
+    with_default_mock_builder(|| {
+        ContentTest::with_member_channel().setup();
+        IssueCreatorTokenFixture::default().call_and_assert(Ok(()));
+
+        let freeze_result = Token::set_frozen_status(RawOrigin::Root.into(), true);
+        assert_ok!(freeze_result);
+
+        ClaimCreatorTokenPatronageCreditFixture::default()
+            .call_and_assert(Err(project_token::Error::<Test>::PalletFrozen.into()));
+    })
```
mnaamani, line 112: For all the new tests being added, since you already are doing the setup and testing the call fails if pallet is frozen, you should follow it up with a call to unfreeze the pallet and then ensure that the original fixture/call succeeds.

mnaamani (changes requested):
Great contribution, just left some change requests and general comments on use of assert_noop and unit tests.

runtime-modules/common/src/lib.rs
freakstatic: Yeah you are right, I forgot to delete it

runtime-modules/project-token/src/lib.rs
```diff
@@ -69,6 +69,7 @@ pub mod weights;
 pub use weights::WeightInfo;
 
 type WeightInfoToken<T> = <T as Config>::WeightInfo;
+pub use frame_support::weights::Weight;
```
mnaamani: I don't this import is needed anymore.

mnaamani (changes requested):
Runtime changes looks good.

Just a general comment. It is generally not recommended to force push changes after a PR is opened and reviews are going on. It will make it harder to do followup reviews as reviewer cannot review new changes and would have to look at the PR as whole again.

As a final step when we make runtime changes is to update the types package.
To do this build the node, and run it with `joystream-node --dev` then run `yarn update-chain-metadata` This produces the chain-metadata.json file at the root.

Then run build script for types package:
`yarn workspace @joystream/types build`
commit changes:
  `types/src/augment/*`
  `chain-metadata.json`
   `query-node/chain-metadata/2002.json`

freakstatic:
> Runtime changes looks good.
> 
> Just a general comment. It is generally not recommended to force push changes after a PR is opened and reviews are going on. It will make it harder to do followup reviews as reviewer cannot review new changes and would have to look at the PR as whole again.
> 
> As a final step when we make runtime changes is to update the types package. To do this build the node, and run it with `joystream-node --dev` then run `yarn update-chain-metadata` This produces the chain-metadata.json file at the root.
> 
> Then run build script for types package: `yarn workspace @joystream/types build` commit changes: `types/src/augment/*` `chain-metadata.json` `query-node/chain-metadata/2002.json`

I apologize for the force push, I wanted to avoid having too many commits for the same feature going to the "master" branch. I will avoid it in the future.

I committed the types packages now, @mnaamani please review again

mnaamani (approved):

PR #4924, opened 2023-10-10
colossus: sync - longer default interval

mnaamani:
As per recommendation in https://github.com/Joystream/joystream/pull/4921#issuecomment-1754633973 

- Changed default syncInterval to 20min if not specified on commandline
- Do not wait initial interval before starting sync
- Add another interval defaulting to 3min on sync failure

This is not an urgent update since it is really only changing the default, but it would be good to confirm with Storage lead if it does indeed help storage operators overall.


storage-node/src/commands/server.ts
```diff
@@ -284,12 +290,12 @@ async function runSyncWithInterval(
   tempDirectory: string,
   syncWorkersNumber: number,
   syncWorkersTimeout: number,
-  syncIntervalMinutes: number
+  syncIntervalMinutes: number,
+  syncRetryIntervalMinutes: number
 ) {
   const sleepInteval = syncIntervalMinutes * 60 * 1000
```
```suggestion
  const sleepInterval = syncIntervalMinutes * 60 * 1000
```

zeeshanakram3 (commented):
Looks good, just one small comment. And for the delay, I repeatedly missed reviewing this.

mnaamani:
> Looks good, just one small comment. And for the delay, I repeatedly missed reviewing this.

Thanks for review. Fixed the typo in [c25416d](https://github.com/Joystream/joystream/pull/4924/commits/c25416db7ad3c2d37e5868c64a283f9f969d48ea)

zeeshanakram3 (approved):
LGTM

PR #4925, opened 2023-10-10
[argus] few tweaks

kdembler:
- Add 1s delay between fetching pages for QN state sync. This will allow QN to have some breathing space and process other requests that may have been stuck because of heavy processing.
- Add `Access-Control-Expose-Headers` response header to allow web clients checking cache status of downloaded file.
- Include response headers in `http` logs

### DIFF
--- a/distributor-node/CHANGELOG.md
+++ b/distributor-node/CHANGELOG.md
@@ -1,8 +1,12 @@
 ##
 
+### 1.4.0
+
+- Add 1s delay between fetching pages for QN state sync. This will allow QN to have some breathing space and process other requests that may have been stuck because of heavy processing.
+
 ### 1.3.1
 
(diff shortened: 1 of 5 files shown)


zeeshanakram3 (approved):
LGTM

mnaamani:
The browser is getting the header `Access-Control-Expose-Headers` twice for each value:

Screenshot 2023-10-16 at 10 39 26 AM

I would have expected one header with the multiple values. Is this the expected response?

In the past I have observed firefox at least complain about saying a header multiple times.. with CORS headers for example. But it doesn't seem to be an issue with this header.

mnaamani (changes requested):
Just a minor merge conflict fix needed otherwise I think this is good to go.

kdembler:
Good call @mnaamani, changed to a single header with comma-separated values. Also rebased and fixed conflicts

mnaamani (approved):

PR #4927, opened 2023-10-12
fix/linkedin-missing-list-1

glisten-star:
- https://github.com/Joystream/pioneer/issues/4540


thesan (changes requested):
@GoldWolf115 I don't think these should be changed:
- `utils/migration-scripts/src/olympia-carthage/olympia-query-node/olympia.schema.graphql`
- `utils/migration-scripts/src/olympia-carthage/olympia-query-node/generated/schema.ts`

Apart from that it looks good to me!

thesan (approved):
LGTM ! could you double check please @zeeshanakram3 ?

zeeshanakram3 (approved):
LGTM

Issue #4930, opened 2023-10-13
CLI: Add a command for the membership WG to verify validator profile


thesan:
SSIA

PR #4933, opened 2023-10-15
add CLI command `verifyValidator`

chrlschwb:
- #4930


thesan (commented):
TBH I don't know anything about the CLI. @zeeshanakram3 could you check this PR please ?

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,39 @@
+import { flags } from '@oclif/command'
+import WorkingGroupsCommandBase from '../../base/WorkingGroupsCommandBase'
+import { WorkingGroups } from '../../Types'
+
+export default class MembershipValidatorAccountCommand extends WorkingGroupsCommandBase {
```
zeeshanakram3: Maybe rename this class to something like `VerifyValidatorAccountCommand` to better reflect the usecase?

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,39 @@
...
+  static description = 'Membership lead/worker verifies validator membership profile'
+  static flags = {
+    message: flags.string({
+      required: true,
+      description: 'Memberhip ID of the validator to verify',
```
zeeshanakram3:
```suggestion
      description: 'Membership ID of the validator to verify',
```

zeeshanakram3: This argument's name is confusing and not aligned with the description of the argument. rename it to something like `memberId`

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,39 @@
...
+    const { message } = this.parse(MembershipValidatorAccountCommand).flags
+
+    const members = await this.getApi().groupMembers(WorkingGroups.Membership)
+    const lead = await this.getApi().groupLead(WorkingGroups.Membership)
+    // const lead = await this.getRequiredLeadContext()
+    const pairs = this.getPairs()
+
+    const membersRows = members.filter((m) =>
+      pairs.find((p) => p.address === m.roleAccount.toString()) ? m.memberId.toString() : undefined
+    )
```
zeeshanakram3: You can use [getRequiredWorkerContext](https://github.com/chrlschwb/joystream/blob/17b39cf69fc297fab10ab6528b4ee5b67441b581/cli/src/base/WorkingGroupCommandBase.ts#L50) helper function here to get the given worker

zeeshanakram3 (changes requested):

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,38 @@
...
+  async run(): Promise<void> {
+    const { memberId } = this.parse(VerifyValidatorAccountCommand).flags
+
+    const members = await this.getApi().groupMembers(WorkingGroups.Membership)
+    const lead = await this.getApi().groupLead(WorkingGroups.Membership)
+
+    const membersRows = members.filter((m) => {
+      const arrayValue: u64[] = [m.memberId]
+      return this.getRequiredWorkerContext('MemberController', arrayValue)
+    })
```
zeeshanakram3: As I mentioned you don't need explicitly get all the workers & lead and then find the actual worker. Please refer to https://github.com/chrlschwb/joystream/blob/d8518bb7ad015052b080e793fce297a34c763abf/cli/src/commands/working-groups/leaveRole.ts example command to see how to use `getRequiredWorkerContext` method properly

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if (!membersRows || membersRows.length === 0) {
+      this.error('Only membership WG lead/worker can perform this command')
+    } else {
+      if (!lead) {
+        this.getOriginalApi().tx.membershipWorkingGroup.workerRemark(Number(membersRows[0]), memberId!)
+      } else {
+        this.getOriginalApi().tx.membershipWorkingGroup.leadRemark(memberId)
```
zeeshanakram3: You also don't need to differentiate b/w workers and lead, since lead is also a worker so you can simply use the `Lead` through it's worker privileges for this action

zeeshanakram3 (changes requested):

cli/src/commands/working-groups/verifyValidator.ts
chrlschwb: The reason why I'm looping through all workers is to determine if the caller is a membership WG worker or a worker in a different WG.

zeeshanakram3: You can simply change the default working group by executing the following function at the start of the command 

```ts
await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
```

Then running `this.getRequiredWorkerContext()` would check the worker in the context of the Membership working group

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,28 @@
...
+  async run(): Promise<void> {
+    const { memberId } = this.parse(VerifyValidatorAccountCommand).flags
+
+    await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
+    const worker = await this.getRequiredWorkerContext()
+
+    if (!worker) {
+      this.error('Only membership WG lead/worker can perform this command')
+    } else {
+      this.getOriginalApi().tx.membershipWorkingGroup.workerRemark(Number(worker), memberId!)
```
zeeshanakram3:
```suggestion
      this.getOriginalApi().tx.membershipWorkingGroup.workerRemark(Number(worker), memberId || '')
```

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,28 @@
...
+      description: 'Membership ID of the validator to verify',
+    }),
+
+    ...WorkingGroupsCommandBase.flags,
+  }
+
+  async run(): Promise<void> {
+    const { memberId } = this.parse(VerifyValidatorAccountCommand).flags
+
+    await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
```
zeeshanakram3, line 38: This is fine, but this will mutate the default working group preference set by the User through CLI to `Membership` every time you run the command, so after running this command you actually want to revert that change, and re-set the default working group whatever it was set before.

zeeshanakram3 (changes requested):

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const nowGroup = this.group;
+
+    await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
+    const worker = await this.getRequiredWorkerContext()
+
+    if (!worker) {
+      this.error('Only membership WG lead/worker can perform this command')
+    } else {
+      this.getOriginalApi().tx.membershipWorkingGroup.workerRemark(Number(worker), memberId || "")
```
mnaamani: The second argument here is supposed to be the serialized protobuf message, not a memberId.

mnaamani (changes requested):
Shouldn't the `workerRemak` call be passed an encoded protobuf `RemarkMetadataAction` message with the `VerifyValidator` action set?

zeeshanakram3:
> Shouldn't the `workerRemak` call be passed an encoded protobuf `RemarkMetadataAction` message with the `VerifyValidator` action set?

Correct. @chrlschwb please address that.

chrlschwb:
> > Shouldn't the `workerRemak` call be passed an encoded protobuf `RemarkMetadataAction` message with the `VerifyValidator` action set?
> 
> Correct. @chrlschwb please address that.

Please let me know which is correct.
- The user creates an encoded protobuf and uses it as input for the CLI command
- The user inputs member_id and is_verified, then the code encodes it to be used in the workerRemark extrinsic

zeeshanakram3:
> The user inputs member_id and is_verified, then the code encodes it to be used in the workerRemark extrinsic

This. For reference, please look at the implementation of [memberRemark](https://github.com/chrlschwb/joystream/blob/a761f9bf380cf6dc3fc28611751b1e7a3adb3867/cli/src/commands/membership/memberRemark.ts) CLI command

chrlschwb:
> > The user inputs member_id and is_verified, then the code encodes it to be used in the workerRemark extrinsic
> 
> This. For reference, please look at the implementation of [memberRemark](https://github.com/chrlschwb/joystream/blob/a761f9bf380cf6dc3fc28611751b1e7a3adb3867/cli/src/commands/membership/memberRemark.ts) CLI command

I have changed as required

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  async run(): Promise<void> {
+    const { memberId, isVerified } = this.parse(VerifyValidatorAccountCommand).flags
+
+    const meta = new ModeratePost({
+      VerifyValidator: new VerifyValidator({
+        memberId,
+        isVerified,
+      }),
+    })
```
zeeshanakram3: The build is failing because the message construction is invalid. Also, why are you wrapping the `VerifyValidator` messaging in `ModeratePost`, please look at the protobuf message structures to encode the message properly https://github.com/chrlschwb/joystream/blob/2bfefe793774cff2fd3699db1df06da9fc300846/metadata-protobuf/proto/WorkingGroups.proto#L66

chrlschwb: please check again

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,49 @@
...
+    required: true,
+    description: 'Membership ID of the validator to verify',
+  },
+  {
+
+    name: "isVerified",
+    required: true,
+    description: 'Verification state of the validator',
+  }
+  ]
```
zeeshanakram3: Please use flags `instead` of `args`. Using flags you have to explicitly write each flag name before providing the value, so less chance of errors, as compared to `args` which are just order dependant positional arguments

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const nowGroup = this.group
+
+    await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
+    const worker = await this.getRequiredWorkerContext()
+
+    if (!worker) {
+      this.error('Only membership WG lead/worker can perform this command')
+    } else {
+      this.getOriginalApi().tx.membershipWorkingGroup.workerRemark(Number(worker), message)
```
zeeshanakram3: This line just only prepares the encoded`tx`. It actually does not sends/broadcasts it.

zeeshanakram3: `worker` is a type of [GroupMember](https://github.com/chrlschwb/joystream/blob/bcc88ee4ac7f3a1be87e79bd7fb5568e102aa774/cli/src/Types.ts#L90), while you are converting it to Number. Which is obviously incorrect.

zeeshanakram3 (changes requested):

cli/src/commands/working-groups/verifyValidator.ts
zeeshanakram3: This comment is still not resolved, please look at any example CLI implementation (e.g. [memberRemark](https://github.com/chrlschwb/joystream/blob/a761f9bf380cf6dc3fc28611751b1e7a3adb3867/cli/src/commands/membership/memberRemark.ts)) to see how transactions are being constructed & dispatched

zeeshanakram3 (changes requested):
- Please see this comment https://github.com/Joystream/joystream/pull/4933#discussion_r1422488832
- I don't think this PR should alter `yarn.lock` file? Can you explain why changes are being made to `yarn.lock` file?

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,61 @@
...
+        memberId: id,
+        isVerified,
+      }),
+    })
+
+    const message = metadataToString(RemarkMetadataAction, meta)
+
+    const nowGroup = this.group
+
+    await this.setPreservedState({ defaultWorkingGroup: WorkingGroups.Membership })
```
thesan, line 38: Hey @zeeshanakram3 it looks to me like the code from this PR should work. I'm just wondering whether these 2 lines and the line 59 are necessary.

Could you have another look please ? This is blocking the validator dashboard QA

thesan (approved):

cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,61 @@
...
+import chalk from 'chalk'
+export default class VerifyValidatorAccountCommand extends WorkingGroupsCommandBase {
+  static description = 'Membership lead/worker verifies validator membership profile'
+  static flags = {
+    memberId: flags.integer({
+      required: true,
+      description: 'Membership ID of the validator to verify',
+    }),
+    isVerified: flags.boolean({
+      required: true,
```
zeeshanakram3, line 16:
```suggestion
     default: false,
```

We can't specify boolean flags as `isVerified=true` or `isVerified=false`, so the only way to signal that `isVerified` is true/false is by presence/absence of the flag itself, i.e.

```
./bin/run working-groups:verifyValidator --memberId=2 --isVerified # isVerified is 'true'
./bin/run working-groups:verifyValidator --memberId=2                     # isVerified is 'false'
```

Since we would define the default value of `isVerified` false so if we omit the flag then it was be assumed false in the command.

cli/src/commands/working-groups/verifyValidator.ts
zeeshanakram3: Right, we can actually specify the working group using `--group` flag, so this is not necessary at all. Please remove these

thesan:
I'm closing this one in favour of #5057

Issue #4903, opened 2023-10-02
[Cli] Update `advanced-transactions` MultiSig calls


mnaamani:
Based on https://github.com/Joystream/joystream/issues/4891#issuecomment-1742622794
The current implementation of the [advanced-transactions](https://github.com/Joystream/joystream/tree/nara/cli/src/commands/advanced-transactions) (for creating multi-sig transactions) will not work with nara which has some changes to the multi-sig api.

- Update to correct `@substrate/tx-wrapper` version
- Fix arguments to `asMulti` and `approveAsMulti` calls.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205629388840428) by [Unito](https://www.unito.io)

bedeho:
Take note of : https://github.com/Joystream/joystream/issues/4928 , but we must be extremely conservative with any broader polkadotjs changes.

Issue #4928, opened 2023-10-13
polkadot-js api/txwrapper update?


mnaamani:
Relates to #4903 

We are able to get away for most applications with using same version of polkadot-js api in nara as ephesus. `v8.9.1`
The joystream-cli depends on https://github.com/paritytech/txwrapper-core `v3.1.3` which also depends on polkadot-js api.

To work with nara we need to use a slightly newer version of txwrapper `v5` which will conflict with the version polkadot-js api we are forcing through resolutions section of monorepo package.json

So it makes sense to update to correct version as recommended by https://forum.polkadot.network/t/polkadot-release-analysis-v0-9-39/2277#hammer_and_wrench-tooling-section-4

This will have some breaking changes and will require some refactoring to most applications.

Alternatively we can change joystream-cli to not require txwrapper core dependency.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205714454743816) by [Unito](https://www.unito.io)

PR #4939, opened 2023-10-19
MultiSign CLI update

chrlschwb:
#4903

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205781567965720) by [Unito](https://www.unito.io)


cli/src/commands/advanced-transactions/constructUnsignedTxApproveMs.ts
```diff
@@ -68,6 +68,16 @@ export default class ConstructUnsignedTxApproveMsCommand extends AdvancedTransac
...
+    refTime: flags.integer({
+      required: false,
+      default: 0,
+      description: 'reftime',
+    }),
+    proofSize: flags.integer({
+      required: false,
+      default: 0,
+      description: 'proofsize',
+    }),
```
mnaamani: Why is the weight being passed as command line argument instead of getting the correct weight via rpc call to `query_info()` ?
If this approach will be taken, we still need to assist the user to find the appropriate values to provide. This is one extra step that makes using these commands harder.

Also if the user doesn't provide these arguments and default values of 0 are used the call will not execute.

cli/src/base/AdvancedTransactionsCommandBase.ts
```diff
@@ -266,10 +275,21 @@ export default abstract class AdvancedTransactionsCommandBase extends AccountsCo
     return unsigned
   }
 
-  async getWeight(call: Call): Promise<number> {
-    const callData = this.getOriginalApi().tx(call)
-    const paymentWeight = await this.getOriginalApi().rpc.payment.queryInfo(callData.toHex())
-    return paymentWeight.weight.toNumber()
+  async getWeight(
+    // call: Call,
```
mnaamani: In general we don't leave commented code around.

cli/src/base/AdvancedTransactionsCommandBase.ts
```diff
@@ -266,10 +275,21 @@ export default abstract class AdvancedTransactionsCommandBase extends AccountsCo
...
+    refTime: string | number,
+    proofSize: string | number
+  ): Promise<{
+    refTime: string | number
+    proofSize: string | number
+  }> {
+    // const callData = this.getOriginalApi().tx(call)
+    // const paymentWeight = await this.getOriginalApi().rpc.payment.queryInfo(callData.toHex())
```
mnaamani: Getting the weight through the rpc call is the correct approach. Why have we switched to passing it manually this way.
If user doesn't sufficient max_weight the call will not get executed.

yarn.lock
```diff
@@ -240,32 +240,6 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.18.8.tgz#2483f565faca607b8535590e84e7de323f27764d"
   integrity sha512-HSmX4WZPPK3FUxYp7g2T6EyO8j96HlZJlxmKPSh6KAcqwyDrfx7hKjXpAW/0FhFfTJsR0Yt4lAjLI2coMptIHQ==
 
-"@babel/compat-data@^7.20.5":
```
mnaamani, line 243: Usually when we change dependencies, we run `yarn && yarn build:packages` to get the final yarn.lock file
Please do so and re-check in the updated yarn.lock

package.json
```diff
@@ -52,7 +52,7 @@
     "pg": "^8.4.0",
     "chalk": "^4.0.0",
     "@types/bn.js": "^5.1.0",
-    "@substrate/txwrapper-core": "=3.1.3"
+    "@substrate/txwrapper-core": "7.1.0"
```
mnaamani: How did you select this version of tx-wrapper ?

mnaamani (changes requested):

package.json
chrlschwb: It's the latest version of tx-wrapper

cli/src/base/AdvancedTransactionsCommandBase.ts
```diff
@@ -154,14 +160,17 @@ export default abstract class AdvancedTransactionsCommandBase extends AccountsCo
...
+      proofSize: string | number
+    }
   ): Promise<MultisigAsMulti> {
     if (input) {
       const args = await this.getAsMultiInputFromFile(input)
       const otherSignatories = args.otherSignatories as string[]
       const otherSignatoriesSorted = sortAddresses(otherSignatories, JOYSTREAM_ADDRESS_PREFIX)
       args.otherSignatories = otherSignatoriesSorted
-      const maxWeightChanged = maxWeight !== (args.maxWeight as number)
+      const maxWeightChanged = maxWeight
```
mnaamani, line 173: I interpret `maxWeightChanged` is a boolean but you are assigning it the actual weight value here.

cli/src/base/AdvancedTransactionsCommandBase.ts
```diff
@@ -266,10 +275,19 @@ export default abstract class AdvancedTransactionsCommandBase extends AccountsCo
...
+  async getWeight(call: Call): Promise<{
+    refTime: string | number
+    proofSize: string | number
+  }> {
     const callData = this.getOriginalApi().tx(call)
     const paymentWeight = await this.getOriginalApi().rpc.payment.queryInfo(callData.toHex())
-    return paymentWeight.weight.toNumber()
+    const proofSize = 0
+    const value = {
+      refTime: paymentWeight.weight.toNumber(),
```
mnaamani, line 286: I think this is still not correct. My guess it would probably also throw an exception decoding the queryInfo() result? But not certain because in other places we do [tx.paymentInfo()](https://github.com/Joystream/joystream/blob/332e22d0b06a27c76c4e7a48db8662072a82dfdb/cli/src/Api.ts#L146) which also return a `DispatchInfo` type but we use the `partialFee` field not the weight.

The problem is despite updating version of tx-wrapper we are still at the workspace level forcing resolution for polkadot-js api v8.9.1 and so the returned type from `queryInfo()` is expected to be:

```ts
/** @name RuntimeDispatchInfo */
export interface RuntimeDispatchInfo extends Struct {
    readonly weight: Weight;
    readonly class: DispatchClass;
    readonly partialFee: Balance;
}

/** @name Weight */
export interface Weight extends u64 {}
```
But infact the weight type is not a u64.

We can see the warning when yarn is installing dependencies:

```sh
warning Resolution field "@polkadot/api@8.9.1" is incompatible with requested version "@polkadot/api@^10.9.1"
```

When I originally refactored some code in repo that relates to estimating fees I didn't give it the attention it needed. For example in this fee-profile I now see my error: https://github.com/Joystream/joystream/blob/332e22d0b06a27c76c4e7a48db8662072a82dfdb/cli/src/base/FeeProfileCommandBase.ts#L102

package.json
mnaamani: I think we only really needed to update to v5 https://github.com/Joystream/joystream/issues/4928 , [ref](https://forum.polkadot.network/t/polkadot-release-analysis-v0-9-40/2468#hammer_and_wrench-tooling-section-4)


TxWrapper Core - [v5.0.0 2](https://github.com/paritytech/txwrapper-core/releases/tag/v5.0.0)
Polkadot JS API - [v9.14.2 2](https://github.com/polkadot-js/api/releases/tag/v9.14.2)

mnaamani (commented):
I think we have no choice but to update polkadot-js/api dependency if we want to do a proper fix. Lets discuss in Nara release call.

mnaamani:
Superceded by https://github.com/Joystream/joystream/pull/5087

PR #4944, opened 2023-10-23
[Argus] fix: Parse Axios error before logging & exporting to the Elasticsearch

zeeshanakram3:
# Problem

Thanks to @kdembler for investigation
> it seems that full axios client is logged with error and we are trying to serialize it including buffers for sending to elastic
which fails because of circular reference

### DIFF
--- a/distributor-node/src/services/networking/NetworkingService.ts
+++ b/distributor-node/src/services/networking/NetworkingService.ts
@@ -333,7 +333,7 @@ export class NetworkingService {
 
       objectDownloadQueue.on('error', (err) => {
         this.logger.error('Download attempt from storage node failed after availability was confirmed:', {
-          err: parseAxiosError(err),
+          err: axios.isAxiosError(err) ? parseAxiosError(err) : err,
         })
       })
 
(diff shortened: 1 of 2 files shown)


kdembler (approved):
LGTM 🫡

distributor-node/src/services/networking/NetworkingService.ts
```diff
@@ -332,7 +332,9 @@ export class NetworkingService {
       })
 
       objectDownloadQueue.on('error', (err) => {
-        this.logger.error('Download attempt from storage node failed after availability was confirmed:', { err })
+        this.logger.error('Download attempt from storage node failed after availability was confirmed:', {
+          err: parseAxiosError(err),
```
mnaamani: If err is not an instance of `AxiosError` will parsing fail?

Perhaps parse only if `axios.isAxiosError(err) == true`

zeeshanakram3: Addressed here [8ebd130](https://github.com/Joystream/joystream/pull/4944/commits/8ebd130675124bfda4e239d0cecb2a9e38eeb8ae)

mnaamani (approved):

PR #4945, opened 2023-10-24
[Nara] make tests dependent of block production time to not run on "fast production" runtime

freakstatic:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205793205691623) by [Unito](https://www.unito.io)


mnaamani (approved):
Not sure why the CI checks are not running, but fix is correct.

PR #4946, opened 2023-10-25
[Nara] add freeze pallet proposal

freakstatic:
Implements the proposal part of #4897


runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -933,6 +939,7 @@ impl<T: Config> Module<T> {
             ProposalDetails::UpdateChannelPayouts(..) => {
                 T::UpdateChannelPayoutsProposalParameters::get()
             }
+            ProposalDetails::FreezePallet(..) => T::SetInvitationCountProposalParameters::get(),
```
mnaamani:
```suggestion
            ProposalDetails::FreezePallet(..) => T::FreezePalletProposalParameters::get(),
```

runtime-modules/common/src/lib.rs
```diff
@@ -69,6 +69,27 @@ pub trait StorageOwnership {
...
+    #[cfg(feature = "std")]
+    use serde::{Deserialize, Serialize};
+    #[cfg(feature = "std")]
+    use strum_macros::EnumIter;
+
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize, EnumIter))]
+    #[derive(
+        Encode, Decode, Clone, PartialEq, Eq, Copy, Debug, PartialOrd, Ord, TypeInfo, MaxEncodedLen,
+    )]
+    pub enum FrezablePallet {
```
mnaamani: small typo:
```suggestion
    pub enum FreezablePallet {
```

Also I don't know if it needs to be enumerable?

runtime-modules/proposals/codex/src/types.rs
```diff
@@ -118,6 +119,9 @@ pub enum ProposalDetails<
 
     /// `Update Channel Payouts` proposal
     UpdateChannelPayouts(UpdateChannelPayoutsParameters),
+
+    /// `Freeze Palet` proposal
+    FreezePallet(bool, FrezablePallet),
```
mnaamani: Since there is only one proposal to do two actions, a better name would be (like the dispatch call) `SetPalletFozenStatus` ?

runtime/src/proposals_configuration/defaults.rs
```diff
@@ -343,3 +343,17 @@ pub(crate) fn update_channel_payouts_proposal() -> ProposalParameters<BlockNumbe
         constitutionality: 1,
     }
 }
+
+// Proposal parameters for the 'Freeze Pallet' proposal
+pub(crate) fn freeze_pallet_proposal() -> ProposalParameters<BlockNumber, Balance> {
+    ProposalParameters {
+        voting_period: days!(3),
+        grace_period: hours!(2),
```
mnaamani:
```suggestion
        grace_period: 0,
```

I think it was suggested by tomato @mochet that it doesn't need a grace period.

mnaamani (changes requested):
Great work, left a few comments and shared in https://github.com/freakstatic/joystream/pull/1 howto fix the build failure seen in the CI checks related to modified runtime types

runtime/src/proposals_configuration/defaults.rs
mnaamani: Tomato perhaps you could also weigh in on other parameters for this proposal.

mochet: > Tomato perhaps you could also weigh in on other parameters for this proposal.

LGTM! It has 2/3 approval vote requirement which seems enough. I don't think this kind of proposal would need 3/3 votes as it can sometimes take time to get the 3rd vote.

runtime-modules/proposals/codex/src/lib.rs
mnaamani: Also please "export" the proposal parameters as a runtime pallet constant in `decl_module! { ... }` like other proposal, as it is needed by pioneer to display and in the integration [tests](https://github.com/freakstatic/joystream/pull/1#discussion_r1379691315).

query-node/mappings/src/proposals.ts
```diff
@@ -305,6 +307,12 @@ async function parseProposalDetails(
     const asPayload = unwrap(specificDetails.payload)?.objectCreationParams
     details.payloadHash = asPayload && bytesToString(asPayload.ipfsContentId)
 
+    return details
+  } else if ((proposalDetails as RuntimeProposalDetails_V2002).isSetPalletFozenStatus) {
+    const details = new UpdatePalletFrozenStatusProposalDetails()
+    const specificDetails = (proposalDetails as RuntimeProposalDetails_V2002).asSetPalletFozenStatus
+    details.frozen = specificDetails[0].isTrue
+    details.pallet = specificDetails[1].type
```
mnaamani: I realized I made a mistake in original PR, although this kind of works because there is a single enum variant.
The correct way would be (assuming the variants have no encapsulated data) would be:

```suggestion
    details.pallet = specificDetails[1].toString()
```

mnaamani (changes requested):
`enum.type` was wrong choice on my part, should be `enum.toString()`
I believe you need to generate types one last time :)
Finally I was surprised to see query-node/chain-metadata/1001.json and 2001.json modified then realized that was result of running `yarn format` in query-node workspace.
I think we should add a `query-node/.prettierignore` with content:
```
chain-metadata/
```

query-node/mappings/src/proposals.ts
mnaamani: Perhaps even clearer code:
```suggestion
    const [frozen, pallet] = (proposalDetails as RuntimeProposalDetails_V2002).asSetPalletFozenStatus
    details.frozen = frozen.isTrue
    details.pallet = pallet.toString()
```

freakstatic: @mnaamani yeah I also thought it was a bit strange but I assumed that it was ok to commit them formatted, should I revert that?

freakstatic:
> `enum.type` was wrong choice on my part, should be `enum.toString()` I believe you need to generate types one last time :) Finally I was surprised to see query-node/chain-metadata/1001.json and 2001.json modified then realized that was result of running `yarn format` in query-node workspace. I think we should add a `query-node/.prettierignore` with content:
> 
> ```
> chain-metadata/
> ```

@mnaamani yeah I also thought it was a bit strange but I assumed that it was ok to commit them formatted, should I revert that?

mnaamani:
> @mnaamani yeah I also thought it was a bit strange but I assumed that it was ok to commit them formatted, should I revert that?

Yes please. It just makes it easier to detect if they are truly modified when we fetch the metadata with `yarn update-chain-metadata` by simply doing `git diff`

PR #4950, opened 2023-10-30
hotfix: QueryNode handling of membership handle

mnaamani:
Problem:
If member handle has non-valid utf-8 chars and is replaced with the `�` replacement character, two unique (at byte level) handles when converted to string can actually become identical, breaking the uniqueness constraint in the database for the handle column.

Solution:
Update query-node schema to:
- do best effort converting handle to utf-8 string an store it in non-unique column
- store raw bytes as hex string in new column

### DIFF
--- a/query-node/mappings/src/membership.ts
+++ b/query-node/mappings/src/membership.ts
@@ -7,6 +7,7 @@ import { DecodedMetadataObject } from '@joystream/metadata-protobuf/types'
 import { isSet } from '@joystream/metadata-protobuf/utils'
 import { MemberId } from '@joystream/types/primitives'
 import { AccountId32, Balance } from '@polkadot/types/interfaces'
+import { Bytes } from '@polkadot/types'
 import {
   PalletMembershipBuyMembershipParameters as BuyMembershipParameters,
   PalletMembershipCreateMemberParameters as CreateMemberParameters,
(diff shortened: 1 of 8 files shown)


query-node/mappings/src/membership.ts
```diff
@@ -193,7 +193,7 @@ async function createNewMemberFromParams(
     rootAccount: rootAccount.toString(),
     controllerAccount: controllerAccount.toString(),
     handle: bytesToString(memberHandle),
-    handleRaw: memberHandle.toHex(),
+    handleRaw: Buffer.from(memberHandle.toU8a(true)).toString('hex'),
```
dobertRowneySr: maybe refactor all the `handle`-setting logic into one function and use it in both places? Since you are setting it in 2 different ways

dobertRowneySr (changes requested):

query-node/mappings/src/membership.ts
dobertRowneySr: also does it automatically sets the '0x' in front of the string?

mnaamani: its doesn't, and I actually pushed a later commit to fallback to simply using `.toHex()`

zeeshanakram3 (approved):
LGTM

dobertRowneySr (approved):
LGTM

dobertRowneySr (approved):
ok

PR #4959, opened 2023-11-07
hydra v5.0.0-alpha.7

mnaamani:
This version of hydra better handles indexing when runtime upgrade happens without the process exiting.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1204404358291458/1205893464537779) by [Unito](https://www.unito.io)

### DIFF
--- a/tests/network-tests/run-runtime-upgrade-tests.sh
+++ b/tests/network-tests/run-runtime-upgrade-tests.sh
@@ -162,14 +162,13 @@ function main {
     set_new_runtime_wasm_path
     echo >&2 "new wasm path set"
 
-    CONTAINER_ID=""
-    export JOYSTREAM_NODE_TAG=${RUNTIME}
     # 4. early chain db init
+    export JOYSTREAM_NODE_TAG=${RUNTIME}
     init_chain_db
     echo >&2 "chain db initialized"
     # 5. start node using new version
     export JOYSTREAM_NODE_TAG=${TARGET_RUNTIME}
-    CONTAINER_ID=$(start_joystream_node)
+    start_joystream_node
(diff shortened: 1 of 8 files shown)


tests/network-tests/run-runtime-upgrade-tests.sh
```diff
@@ -167,7 +167,8 @@ function main {
     # 4. early chain db init
     init_chain_db
     echo >&2 "chain db initialized"
-    # 5. start node
+    # 5. start node using new version
+    export JOYSTREAM_NODE_TAG=${TARGET_RUNTIME}
     CONTAINER_ID=$(start_joystream_node)
```
zeeshanakram3: The `CONTAINER_ID` variable is no longer used except for assignment here. I think it should be removed?

zeeshanakram3 (commented):
Looks good. Just a suggestion I think we should update the hydra dependencies at all places to `v5.0.0-alpha.7` version, for consistency. As `package.json` files in `query-node/*` has older hydra dependencies. WDYT?

mnaamani:
> Looks good. Just a suggestion I think we should update the hydra dependencies at all places to `v5.0.0-alpha.7` version, for consistency. As `package.json` files in `query-node/*` has older hydra dependencies. WDYT?

That might be a good idea, although those packages were not actually modified it doesn't hurt to keep them all at the same version to avoid any confusion about why they are different.

mnaamani:
> 5.0.0-alpha

Done in a8722d26937bdae561097c35d538b5a638f3d3b6

zeeshanakram3 (approved):
LGTM

PR #4963, opened 2023-11-10
FIX gracefully handle logging exceptions in Argus and Colossus

zeeshanakram3:
# Context

As reported by @kdembler the Distributor node crashed with restarting the ES node, or if providing the wrong credentials. This happened because the code was not gracefully the exceptions

# Fix

This PR gracefully handles logging exceptions in Argus and Colossus

### DIFF
--- a/distributor-node/package.json
+++ b/distributor-node/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@joystream/distributor-cli",
   "description": "Joystream distributor node CLI",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "author": "Joystream contributors",
   "bin": {
     "joystream-distributor": "./bin/run"
@@ -50,7 +50,7 @@
     "url-join": "^4.0.1",
     "winston": "^3.3.3",
     "winston-daily-rotate-file": "^4.5.5",
-    "winston-elasticsearch": "^0.15.8",
+    "winston-elasticsearch": "^0.17.4",
     "yaml": "^1.10.2"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


distributor-node/src/services/logging/LoggingService.ts
```diff
@@ -70,12 +70,21 @@ export class LoggingService {
...
+    this.esTransport?.on('error', (error) => {
+      console.error('Error in logger caught:', error)
+    })
   }
 
   public static withAppConfig(config: ReadonlyConfig): LoggingService {
     const transports: winston.LoggerOptions['transports'] = []
 
     let esTransport: ElasticsearchTransport | undefined
+    console.log('config.logs?.elastic', config.logs?.elastic)
```
kdembler: 🙈

distributor-node/package.json
```diff
@@ -50,7 +50,7 @@
     "url-join": "^4.0.1",
     "winston": "^3.3.3",
```
kdembler, line 51: Should we maybe bump winston as well while we're at it?

distributor-node/src/services/logging/LoggingService.ts
zeeshanakram3: Removed

distributor-node/package.json
zeeshanakram3: Ok. I tried to update the `winston` to the most recent version and it turned out that it contains breaking changes (so will have to update code at couple of places). So unless we absolutely want to update the dependency, I would say it's fine for now. WDYT?

kdembler: yes, let's ignore in that case

kdembler (approved):
For some weird reason I can't get it to connect to local ES when developing, but seems that was also the case for me before this PR.

That said, tested with public ES and it indeed caught the exception and didn't crash on failed write, so all good, thanks!

mnaamani (approved):

PR #4964, opened 2023-11-12
[Argus] use ES data streams, enable log correlation

kdembler:
- Changed Argus Elasticsearch transport to use data streams instead of regular indices. Renamed `config.logs.elastic.index` to `config.logs.elastic.indexPrefix`. Node ID from config will be automatically appended to the index name.
- Changed Colossus Elasticsearch transport to use data streams instead of regular indices. Removed `--elasticSearchIndex` option and replaced with `--elasticSearchIndexPrefix`. Node ID from config will be automatically appended to the index name.
- Updated example distributor node config to match values currently used in production.

### DIFF
--- a/distributor-node/package.json
+++ b/distributor-node/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@joystream/distributor-cli",
   "description": "Joystream distributor node CLI",
-  "version": "1.4.1",
+  "version": "1.5.0",
   "author": "Joystream contributors",
   "bin": {
     "joystream-distributor": "./bin/run"
@@ -161,7 +161,7 @@
     "checks": "tsc --noEmit --pretty && prettier ./ --check && yarn lint",
     "cli": "./bin/run",
     "start": "./bin/run start",
-    "start:with-instrumentation": "export OTEL_APPLICATION=distributor-node; node --require @joystream/opentelemetry ./bin/run start"
+    "start:with-instrumentation": "export OTEL_APPLICATION=distributor-node; export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8200; export OTEL_RESOURCE_ATTRIBUTES=service.name=argus,deployment.environment=production; node --require @joystream/opentelemetry ./bin/run start"
   },
   "types": "lib/index.d.ts"
 }
--- a/storage-node/src/services/logger.ts
+++ b/storage-node/src/services/logger.ts
@@ -148,7 +148,7 @@ function createCustomLogger(customOptions: LogConfig): winston.Logger {
       createElasticTransport(
         customOptions.elasticSearchlogSource,
         customOptions.elasticSearchEndpoint,
-        customOptions.elasticSearchIndex,
+        customOptions.elasticSearchIndexPrefix,
         customOptions.elasticSearchUser,
         customOptions.elasticSearchPassword
       )
@@ -210,6 +210,9 @@ function createElasticTransport(
     elasticLogLevel = 'debug' // default
   }
 
+  const indexPrefix = elasticSearchIndex || 'logs-colossus'
+  const index = `${indexPrefix}-${logSource}`.toLowerCase()
+
   const esTransport = new ElasticsearchTransport({
     level: elasticLogLevel,
     clientOpts: {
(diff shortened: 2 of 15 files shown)


kdembler:
@zeeshanakram3 FYI, did the same for Colossus as well

storage-node/src/services/logger.ts
zeeshanakram3, line 1: please change this function argument to [elasticSearchIndexProfix](https://github.com/Joystream/joystream/pull/4964/files#diff-7189f20f39097d9ede9fda7ae09c2d9cc76ea109fab783ce717fd3e159c67722R200)

storage-node/src/services/logger.ts
```diff
@@ -224,10 +227,36 @@ function createElasticTransport(
...
+      }
+      if (logData.meta.span_id) {
+        transformed.span = { id: logData.meta.span_id }
+      }
+      if (logData.meta.transaction_id) {
+        transformed.transaction = { id: logData.meta.transaction_id }
+      }
+
+      return transformed
+    },
```
zeeshanakram3, line 259: What type of traces are being assumed here? Opentelemetry traces?

Also, the Github resource you linked has `ids` in the format `trace.id`, while you are using `trace_id`. Is this correct.

kdembler: So the fields (`trace_id`, etc) are added by [our auto Node instrumentation](https://github.com/Joystream/joystream/blob/master/opentelemetry/instrumentations/default.ts#L22), winston has its own integration. My best guess is that at some recent version APM changed the name of those values and `winston-elasticsearch` wasn't updated. Or they are renamed for some reason at some point. Anyhow, here's an example log with tracing:
```
...
        "res": {
            "headers": {
                "x-powered-by": "Express",
                "access-control-allow-origin": "*",
                "timing-allow-origin": "*",
                "access-control-expose-headers": "x-cache, x-data-source",
                "x-cache": "hit",
                "x-data-source": "local",
                "content-disposition": "inline",
                "content-type": "image/png",
                "accept-ranges": "bytes",
                "cache-control": "public, max-age=31536000",
                "etag": "W/\"bded-18bd5d45f9b\"",
                "content-length": 48621
            },
            "statusCode": 200
        },
        "responseTime": 1,
        "prematurelyClosed": false
    },
    "trace_id": "5baade5b12b703e2f75ffb7c2132809a",
    "span_id": "20e622119c2e385e",
    "trace_flags": "01"
}
```
Thanks to the transform fields get renamed to `trace.id` and elastic can properly correlate.

storage-node/src/services/logger.ts
kdembler: Thanks, done

zeeshanakram3 (approved):

zeeshanakram3:
As the integration-test logs suggest that the [colossus-1](https://github.com/Joystream/joystream/actions/runs/6891652323/job/18747416981#step:8:353) is not serving any buckets (i.e. storage node during initialization could not find [worker/buckets](https://github.com/kdembler/joystream/blob/86a96be0362bf32fd4d847878775aae41311ee69/storage-node/src/commands/server.ts#L345) that it's supposed to serve).  I think this is mainly because QN is lagging and info about worker/buckets created in the previous [step](https://github.com/Joystream/joystream/actions/runs/6891652323/job/18747416981#step:8:315) hasn't been processed yet. 

## Fix

Add `sleep` after completion of [initStorageAndDistribution](https://github.com/kdembler/joystream/blob/86a96be0362bf32fd4d847878775aae41311ee69/tests/network-tests/start-storage.sh#L13) scenario to give time to QN for blocks processing

distributor-node/package.json
```diff
@@ -161,7 +161,7 @@
     "checks": "tsc --noEmit --pretty && prettier ./ --check && yarn lint",
     "cli": "./bin/run",
     "start": "./bin/run start",
-    "start:with-instrumentation": "export OTEL_APPLICATION=distributor-node; node --require @joystream/opentelemetry ./bin/run start"
+    "start:with-instrumentation": "export OTEL_APPLICATION=distributor-node; export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8200; export OTEL_RESOURCE_ATTRIBUTES=service.name=argus,deployment.environment=production; node --require @joystream/opentelemetry ./bin/run start"
```
mnaamani, line 164: I don't think it is the best place to hardcode the otlp endpoint in this npm script.
But since in production we are using docker compose file which doesn't directly use this script I think its okay, but I just wanted to make a note of it.

mnaamani (approved):

PR #4965, opened 2023-11-16
Nara master update

mnaamani:


mnaamani:
The storageSync test is failing. I don't think it is getting enough time to sync files, even though there is a period in the test pausing before checking for files. This is because the sync process only start after the set interval on startup. There is a fix for that in https://github.com/Joystream/joystream/pull/4924 I think it would be good to include this fix before trying running the tests again @zeeshanakram3

PR #4966, opened 2023-11-17
Opentelemetry fixes

zeeshanakram3:
This PR

- Fixes loading env vars from `.env` file in `@joystream/opentelemetry` module. Due to this bug the `OTEL_APPLICATION` was not exported and hence [DefaultInstrumentation](https://github.com/zeeshanakram3/joystream/blob/a91634aae25735e9d5baf73d56d1496306493ce0/opentelemetry/index.ts#L23) was used for Argus, and hence argus related attributes were not exported to ES.
- make OTEL `maxQueueSize` & `maxExportBatchSize` span processor options configurable
- Add 'nodeId' attribute to all Argus spans

### DIFF
--- a/opentelemetry/instrumentations/distributor-node.ts
+++ b/opentelemetry/instrumentations/distributor-node.ts
@@ -12,8 +12,8 @@ import { ClientRequest, ServerResponse } from 'http'
 
 export const DistributorNodeInstrumentation = new NodeSDK({
   spanProcessor: new BatchSpanProcessor(new OTLPTraceExporter(), {
-    maxQueueSize: 8192 /* 4 times of default queue size */,
-    maxExportBatchSize: 1024 /* 2 times of default batch size */,
+    maxQueueSize: parseInt(process.env.OTEL_MAX_QUEUE_SIZE || '8192'),
+    maxExportBatchSize: parseInt(process.env.OTEL_MAX_EXPORT_BATCH_SIZE || '1024'),
   }),
   metricReader: new PeriodicExportingMetricReader({
     exporter: new OTLPMetricExporter(),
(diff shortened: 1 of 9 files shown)


opentelemetry/instrumentations/distributor-node.ts
```diff
@@ -5,15 +5,21 @@ import { FsInstrumentation } from '@opentelemetry/instrumentation-fs'
...
 import { NodeSDK } from '@opentelemetry/sdk-node'
-import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-node'
+import { BatchSpanProcessor, Span } from '@opentelemetry/sdk-trace-node'
 import { ClientRequest, ServerResponse } from 'http'
 
 /** Opentelemetry Instrumentation for Joystream Distributor Node */
 
+class CustomSpanProcessor extends BatchSpanProcessor {
+  onStart(span: Span) {
+    span.setAttribute('nodeId', process.env.JOYSTREAM_DISTRIBUTOR__ID)
```
mnaamani, line 15: Question: I see the code in the config parser setting this env variable, but it happens in the distributor code. Do we know that `onStart()` is called after the config parser has done it's job?

mnaamani (commented):
Just left a question, otherwise LGTM.

opentelemetry/instrumentations/distributor-node.ts
zeeshanakram3: Good question, the instrument code is loaded & initialized first and then the config parser will do its job, i.e. exporting the env vars. Because of this reason the first few spans when the distributor node starts will have nodeId as undefined (if the config was set through `config.yml` file and not as env vars).

mnaamani (approved):

PR #4969, opened 2023-11-20
HotFix for Critical sync error: RangeError: Maximum call stack size exceeded

mnaamani:
We had a case where a new storage node was doing an initial sync. The number of objects to sync was around 380,000.
Sync tasks being added was using pattern `Array.push(...tasks)` this results in a call to `Array.push(....)` with 380,000 arguments which are all added to the javascript call stack which throws an exception, this resulted in the node never being able to complete a sync run.

The fix is to just iterate over the tasks array elements and push them individually into the target array.

### DIFF
--- a/storage-node/CHANGELOG.md
+++ b/storage-node/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 3.8.1
+
+- Hotfix: Fix call stack size exceeded when handling large number of initial object to sync.
+
 ### 3.8.0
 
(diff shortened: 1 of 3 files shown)


zeeshanakram3 (approved):
LGTM. Thanks for investigating & fixing this issue. Just a related point, I think we use the spread operator notation (i/e.`...tasks`) in a number of places, so maybe we should find all the occurrences and add fixes (But that for some other task I think).

PR #4970, opened 2023-11-21
Test script runners refactoring

mnaamani:
Problem: Too many confusing variables, and scripts to run integration tests.

Solution:
 - A bit of house keeping.
 - Dropped the `REUSE_KEYS` environment variable. When a scenario is run and `output.json` is found, the generated keys will be loaded from it.
- Updated scripts to explicitly remove `output.json` where appropriate. eg. when scenario is run to initialize a new network.
- Adding a new scenario called flow which can be used to run an individual flow. This allows us to more flexibly run flows without need to group them in a scenario.
- Changed default behavior of `yarn test` and `yarn start` to start a test chain (TESTING runtime profile)


zeeshanakram3 (commented):
Looks good, just a few comments:

- In the Github workflows we still use `yarn build:packages`, maybe better to replace with `yarn build` for consistency? 
- We should also remove `run-test-node-docker.sh` script too, it only exporting `RUNTIME_PROFILE=TESTING` , maybe do that in the calling script?

zeeshanakram3 (approved):
LGTM

PR #4971, opened 2023-11-21
[Colossus] Accept pending data objects enhancements

### LINKED ISSUE #4968: [Bug] Object upload request to the storage-node fail even though the call to `accept_pending_data_objects` in the same request succeeds
## Context 

```
        "trace": [
          {
            "file": "/joystream/node_modules/promise-timeout/index.js",
            "method": "TimeoutError",
            "native": false,
            "line": 44,
            "function": "new module.exports.TimeoutError",
            "column": 16
          },
          {
            "file": "/joystream/node_modules/promise-timeout/index.js",
            "method": "timeout",
            "native": false,
            "line": 20,
            "function": "module.exports.timeout",
            "column": 15
          },
          {
            "file": "/joystream/storage-node/lib/services/runtime/extrinsics.js",
            "method": null,
            "native": false,
            "line": 207,
            "function": "extrinsicWrapper",
            "column": 45
          },
          {
            "file": "/joystream/storage-node/lib/services/runtime/extrinsics.js",
            "method": null,
            "native": false,
            "line": 115,
            "function": "acceptPendingDataObjects",
            "column": 18
          },
          {
            "file": "/joystream/storage-node/lib/services/webApi/controllers/filesApi.js",
            "method": null,
            "native": false,
            "line": 99,
            "function": "uploadFile",
            "column": 57
          }
        ],

```

As recorded, some data object upload requests (endpoint: `POST /api/v1/files`) to the storage node fail (`Error 500: Timeout`) even though the call to `accept_pending_data_objects` in the same request succeeds. This is happening because the extrinsicWrapper - which calls the `accept_pending_data_objects` after the object upload has been completed on the storage node and the storage node verifies that the object is valid (e.g. it's size, and hash matches the runtime values) - is hardcoded with [25s](https://github.com/zeeshanakram3/joystream/blob/a91634aae25735e9d5baf73d56d1496306493ce0/storage-node/src/services/runtime/extrinsics.ts#L266) default timeout.

**_Now, the problem is that after sending the `accept_pending_data_objects` tx, [extrinsicWrapper](https://github.com/zeeshanakram3/joystream/blob/a91634aae25735e9d5baf73d56d1496306493ce0/storage-node/src/services/runtime/extrinsics.ts#L68) will timeout after 25s, even though tx may get included in the block after this timeout and eventually gets succeeded, so the side-effect in runtime has happened and the storage-node is not aware of it, or incorrectly it assumes that every timed out tx would eventually fail._**  

# Implications

After the error is thrown from the [acceptPendingDataObjects](https://github.com/zeeshanakram3/joystream/blob/a91634aae25735e9d5baf73d56d1496306493ce0/storage-node/src/services/webApi/controllers/filesApi.ts#L133) the catch block of the calling function performs the file cleanup, so major implication of this bug is that if the object is successfully accepted in the runtime, then the cleanup step effectively leads to the permanent loss of the data object, since storage-node by design also does not allow re-uploading of the [accepted](https://github.com/zeeshanakram3/joystream/blob/a91634aae25735e9d5baf73d56d1496306493ce0/storage-node/src/services/webApi/app.ts#L246) object as a measure against denial of services attack.

zeeshanakram3:
fixes #4968

This PR:

- Increase the default timeout value in the extrinsic wrapper to match the transaction validity in the transaction pool
- Adds new `AcceptPendingObjectsService` that is responsible for periodically sending batch `accept_pending_data_objects` for all the pending data objects
- The `POST /files` endpoint now no longer calls the `accept_pending_data_objects` extrinsic for individual uploads, instead, it registers all the pending objects with `AcceptPendingObjectsService`
- Updates '/state/data' endpoint response headers to return data objects status too i.e. (`pending` or `accepted`)


storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  private async loadPendingDataObjects(): Promise<void> {
+    const pendingIds = await this.getPendingObjectsFromLocalDir(this.pendingDataObjectsDir)
+
+    await this.loadPendingDataObjectsFromIDs(pendingIds)
+
+    logger.debug(`Pending data objects ID cache loaded.`)
+  }
+
+  public push(dataObjectId: string, storageBucketId: string, bagId: string): void {
```
mnaamani:
```suggestion
  public add(dataObjectId: string, storageBucketId: string, bagId: string): void {
```

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+    this.pendingDataObjects.set(dataObjectId, [storageBucketId, bagId])
+  }
+
+  private async popN(n: number): Promise<AcceptPendingDataObjectsParams[]> {
```
mnaamani: `pop` is a poor choice of name because you are implying the underyling data structure is a stack, but since it is a map and I don't believe there is any guaranteed ordering. Closer to an "unordered queue".

Maybe `take(n)`? And separate this function into two, one `take` returns `n` removed elements and second function to process taken elements and produces the params array.

storage-node/src/services/runtime/extrinsics.ts
```diff
@@ -129,6 +131,64 @@ export async function updateStorageBucketsForBags(
...
+  // a list of failed data objects
+  const failedDataObjects: string[] = []
+
+  const txsByTransactorAccount = acceptPendingDataObjectsParams.map(({ account, storageBucket }) => {
+    const txs = storageBucket.bags.map((bag) =>
+      api.tx.storage.acceptPendingDataObjects(
+        workerId,
+        storageBucket.id,
+        parseBagId(bag.id),
+        api.createType('BTreeSet<u64>', bag.dataObjects)
```
mnaamani, line 159: I seem to recollect that there is a need to make sure the elements in the set must be ordered, otherwise we get an error in the runtime code when the arguments are being decoded.

mnaamani: Nvm: It was fixed in https://github.com/polkadot-js/api/pull/3789

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+  private constructor(
+    private qnApi: QueryNodeApi,
+    private pendingDataObjectsDir: string,
+    private bucketKeyPairs: Map<string, KeyringPair>,
+    private uploadBuckets: string[]
+  ) {
+    this.pendingDataObjects = new Map()
+  }
+
+  static async init(
```
mnaamani: Makes sense to only allow creating of one instance "singelton" to prevent multiple instances of this service being created. Maybe this should be named "create", and on subsequent calls returns existing instance (stored as a private this.instance)

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+  private async loadPendingDataObjectsFromIDs(pendingIds: string[]): Promise<void> {
+    const pendingDataObjects = await this.qnApi.getDataObjectsByIds(pendingIds)
+
+    pendingDataObjects.forEach((dataObject) => {
+      const storageBucket = dataObject.storageBag.storageBuckets.find(({ id }) => this.uploadBuckets.includes(id))
+      if (storageBucket) {
+        this.push(dataObject.id, storageBucket.id, dataObject.storageBag.id)
+      } else {
+        logger.warn(
+          `Data object ${dataObject.id} in pending directory is not assigned to any of the upload buckets: ${this.uploadBuckets}.`
```
mnaamani: We would probably want to remove these objects at some point, probably in the "pruning" worker once we merge.

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+    return this.pendingDataObjects.get(objectId)
+  }
+
+  private async loadPendingDataObjectsFromIDs(pendingIds: string[]): Promise<void> {
+    const pendingDataObjects = await this.qnApi.getDataObjectsByIds(pendingIds)
+
+    pendingDataObjects.forEach((dataObject) => {
+      const storageBucket = dataObject.storageBag.storageBuckets.find(({ id }) => this.uploadBuckets.includes(id))
+      if (storageBucket) {
+        this.push(dataObject.id, storageBucket.id, dataObject.storageBag.id)
```
mnaamani:
```suggestion
        this.insert(dataObject.id, storageBucket.id, dataObject.storageBag.id)
```
renaming just for correct symantics given the underlying data structure.

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+   * @param directory - local directory to get file names from
+   */
+  private async getPendingObjectsFromLocalDir(directory: string): Promise<string[]> {
+    // Check if directory exists and if not, create it
+    if (!fs.existsSync(directory)) {
+      await fsPromises.mkdir(directory)
+    }
+
+    // Read the directory contents
+    return await fsPromises.readdir(directory)
```
mnaamani, line 207: I have an issue with both the temp dir and now the pending dir being sub directories of the uploads dir.

But that aside (I'll make a separate issue for it), we should make sure the sync worker which also loads objects with `readdir` ignores the pending dir here https://github.com/Joystream/joystream/blob/9c5383d213667d750d9150d4483982cab84ff0e4/storage-node/src/services/caching/localDataObjects.ts#L34

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    let failedObjectsIds: string[]
+    try {
+      failedObjectsIds = await acceptPendingDataObjectsBatch(api, workerId, params)
+    } catch (err) {
+      // Re-push all params as failed since the entire batch failed
+      params.forEach((param) => {
+        param.storageBucket.bags.forEach((bag) => {
+          bag.dataObjects.forEach((dataObjectId) => {
+            this.push(dataObjectId, param.storageBucket.id, bag.id)
```
mnaamani: Is it possible the bag doesn't belong in the bucket anymore, and we should let the service drop it since the transaction will never work?

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    return acceptPendingObjectsService
+  }
+
+  getPendingDataObject(objectId: string): [string, string] | undefined {
+    return this.pendingDataObjects.get(objectId)
+  }
+
+  private async loadPendingDataObjectsFromIDs(pendingIds: string[]): Promise<void> {
+    const pendingDataObjects = await this.qnApi.getDataObjectsByIds(pendingIds)
```
mnaamani, line 77: I think we want to filter out any objects that have not been accepted yet? To avoid wasting resources trying to submit a call that will fail. Imagine a scenario where tx was submitted to accept upload, but node stopped before the file was moved out of the pending dir. 

Perhaps the directory being parsed had some stale objects, operator switched to a directory with existing files.

How will this behave if there are some files in the directory with unexpected file names..

mnaamani (changes requested):
Left a few suggestions.

I think we need to start thinking about add a better way to manage the state of the node between runs. Depending on presence of files in directories is brittle and many things can go wrong as the file system is easy to corrupt my moving files around accidentally.

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: Done.

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: Do you mean runtime transaction would fail if any of the calls are invalid?  If yes, that will not be the case, since we are using [forceBatch](https://github.com/zeeshanakram3/joystream/blob/51901e12660b6e7cd5589cd29443da7a6db17bd5/storage-node/src/services/runtime/extrinsics.ts#L167)

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: Yeah, wasn't sure where that should be done, here OR in the pruning service, that's why I left it as it is.

This scenario would mostly happen if the data object is moved (i.e. object's bag is moved). And, I think removing objects due to this action is best suited for the Pruning service (which extensively ensures that moved data objects at least exist in `N` different buckets before deleting it). WDYT?

Maybe I should create an issue for that?

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: Yes, in the sense that it will remain in pending folder and will be retried again later. But I think this is similar to discussion here https://github.com/Joystream/joystream/pull/4971#discussion_r1409373556

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: Yes lets leave it to the pruning service, and create an issue for it.
Just fyi the version of colossus with the pruning service is merged into [`colossus-beta`](https://github.com/Joystream/joystream/tree/colossus-beta) branch.

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -0,0 +1,251 @@
...
+    return this.instance
+  }
+
+  private runWithInterval(api: ApiPromise, workerId: number, maxTxBatchSize: number, intervalMs: number) {
+    const run = () => {
+      this.acceptPendingDataObjects(api, workerId, maxTxBatchSize).catch((err) =>
+        logger.error(`Failed to register pending data objects as accepted in runtime: ${err}`)
+      )
+
+      setTimeout(run, intervalMs)
```
mnaamani:
```suggestion
      ).finally(() => setTimeout(run, intervalMs))
```

mnaamani (changes requested):
Minor fix in `runWithInterval()`

mnaamani (approved):

PR #4984, opened 2023-11-29
twitter-name fixed

glisten-star:
Twitter name was changed X


thesan:
This will be handled in Pioneer

Issue #4830, opened 2023-08-14
QN: Label Metaprotocol in Forum Thread and Proposals


chrlschwb:
### Background
https://github.com/Joystream/pioneer/issues/4497

### Requirement
A new metaprotocol `label` will be created and added to `forumThreads` and `proposals` entities.
It should have the following parameters.
- name
- description
- visibility

PR #4987, opened 2023-11-29
metaprotocol labeling: mappings and protobufs

mkbeefcake:
Created Protobuf type and mapping handler in query-node for Label
For the issue #4830

### DIFF
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -127,19 +127,13 @@ message UnassignTagFromProposal {
 }
 
 message AllowTagToWorker {
-  // ID of tag
-  required string tag_id = 1;
-
   // ID of worker
-  required uint64 worker_id = 2;  
+  required uint64 worker_id = 1;
 }
 
 message DisallowTagToWorker {
-  // ID of tag
-  required string tag_id = 1;
-
   // ID of worker
-  required uint64 worker_id = 2;  
+  required uint64 worker_id = 1;  
 }
 
 message RemarkMetadataAction {
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -94,33 +94,33 @@ message UpdateTag {
   optional bool visibility = 5;
 }
 
-message AssignTagToThread {
+message AssignTagsToThread {
   // ID of tag
-  required string tag_id = 1;
+  repeated string tag_ids = 1;
 
   // ID of thread
   required string thread_id = 2;  
 }
 
-message UnassignTagFromThread {
+message UnassignTagsFromThread {
   // ID of tag
-  required string tag_id = 1;
+  repeated string tag_ids = 1;
 
   // ID of thread
   required string thread_id = 2;  
 }
 
-message AssignTagToProposal {
+message AssignTagsToProposal {
   // ID of tag
-  required string tag_id = 1;
+  repeated string tag_ids = 1;
 
   // ID of proposal
   required string proposal_id = 2;  
 }
 
-message UnassignTagFromProposal {
+message UnassignTagsFromProposal {
   // ID of tag
-  required string tag_id = 1;
+  repeated string tag_ids = 1;
 
   // ID of proposal
   required string proposal_id = 2;  
@@ -142,10 +142,10 @@ message RemarkMetadataAction {
     VerifyValidator verify_validator = 2;
     CreateTag create_tag = 3;
     UpdateTag update_tag = 4;
-    AssignTagToThread assign_tag_to_thread = 5;
-    UnassignTagFromThread unassign_tag_from_thread = 6;
-    AssignTagToProposal assign_tag_to_proposal = 7;
-    UnassignTagFromProposal unassign_tag_from_proposal = 8;
+    AssignTagsToThread assign_tags_to_thread = 5;
+    UnassignTagsFromThread unassign_tags_from_thread = 6;
+    AssignTagsToProposal assign_tags_to_proposal = 7;
+    UnassignTagsFromProposal unassign_tags_from_proposal = 8;
     AllowTagToWorker allow_tag_to_worker = 9;
     DisallowTagToWorker disallow_tag_to_worker = 10;
   }
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -82,7 +82,7 @@ message UpdateTag {
   required string tag_id = 1;
 
   // name of tag
-  required string name = 2;
+  optional string name = 2;
 
   // Updated Description of Tag
   optional string description = 3;
--- a/metadata-protobuf/proto/WorkingGroups.proto
+++ b/metadata-protobuf/proto/WorkingGroups.proto
@@ -63,8 +63,95 @@ message ModeratePost {
   required string rationale = 2;
 }
 
+message CreateTag {
+  // Name of new Tag, e.g : roadmap
+  required string name = 1;
+
+  // Description of new Tag, e.g : Roadmap covering the Joystream ecosystem
+  required string description = 2;
+
+  // Type of new Tag : e.g : roadmap
+  required string type = 3;
+
+  // Visibility of new Tag : e.g : true
+  required bool visibility = 4;
+}
+
+message UpdateTag {
+  // ID of tag 
+  required string tag_id = 1;
+
+  // name of tag
+  required string name = 2;
+
+  // Updated Description of Tag
+  optional string description = 3;
+
+  // Updated Type of Tag
+  optional string type = 4;
+
+  // Updated Visibility of Tag
+  optional bool visibility = 5;
+}
+
+message AssignTagToThread {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of thread
+  required string thread_id = 2;  
+}
+
+message UnassignTagFromThread {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of thread
+  required string thread_id = 2;  
+}
+
+message AssignTagToProposal {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of proposal
+  required string proposal_id = 2;  
+}
+
+message UnassignTagFromProposal {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of proposal
+  required string proposal_id = 2;  
+}
+
+message AllowTagToWorker {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of worker
+  required uint64 worker_id = 2;  
+}
+
+message DisallowTagToWorker {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of worker
+  required uint64 worker_id = 2;  
+}
+
 message RemarkMetadataAction {
   oneof action {
     ModeratePost moderate_post = 1;
+    CreateTag create_tag = 2;    
+    UpdateTag update_tag = 3;
+    AssignTagToThread assign_tag_to_thread = 4;
+    UnassignTagFromThread unassign_tag_from_thread = 5;
+    AssignTagToProposal assign_tag_to_proposal = 6;
+    UnassignTagFromProposal unassign_tag_from_proposal = 7;
+    AllowTagToWorker allow_tag_to_worker = 8;
+    DisallowTagToWorker disallow_tag_to_worker = 9;
   }
 }
--- a/query-node/mappings/src/label/tag.ts
+++ b/query-node/mappings/src/label/tag.ts
@@ -3,10 +3,10 @@ import { DatabaseManager } from '@joystream/hydra-common'
 import {
   ICreateTag,
   IUpdateTag,
-  IAssignTagToThread,
-  IAssignTagToProposal,
-  IUnassignTagFromThread,
-  IUnassignTagFromProposal,
+  IAssignTagsToThread,
+  IAssignTagsToProposal,
+  IUnassignTagsFromThread,
+  IUnassignTagsFromProposal,
   IAllowTagToWorker,
   IDisallowTagToWorker,
 } from '@joystream/metadata-protobuf'
@@ -86,13 +86,13 @@ export async function processUpdateTag(
   return tag
 }
 
-export async function processAssignTagToThread(
+export async function processAssignTagsToThread(
   store: DatabaseManager,
-  metadata: DecodedMetadataObject<IAssignTagToThread>,
+  metadata: DecodedMetadataObject<IAssignTagsToThread>,
   isLead: boolean,
   workerId: number
 ): Promise<any> {
-  const { tagId, threadId } = metadata
+  const { tagIds, threadId } = metadata
 
   if (!isLead) {
     const tagPermittedWorker: TagPermittedWorker | undefined = await getOneBy(store, TagPermittedWorker, {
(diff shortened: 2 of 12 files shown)


metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -63,10 +63,97 @@ message ModeratePost {
...
+  // Visibility of new Tag : e.g : true
+  required bool visibility = 4;
+}
+
+message UpdateTag {
+  // ID of tag 
+  required string tag_id = 1;
+
+  // name of tag
+  required string name = 2;
```
kdembler: Do we need to require name when updating?

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,270 @@
...
+} from 'query-node/dist/model'
+import { MetaprotocolTxError, getOneBy, getById, logger, getOneByOrFail, getByIdOrFail } from 'src/common'
+
+export async function processCreateTag(
+  store: DatabaseManager,
+  event: SubstrateEvent,
+  metadata: DecodedMetadataObject<ICreateTag>,
+  workgroup: WorkingGroup,
+  isLead: boolean,
+  workerId: number
```
kdembler: There's a couple of warnings about unused parameters/variables, please fix

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,270 @@
...
+    return MetaprotocolTxError.TagPermNotAllowed
+  }
+
+  const tagId = `${event.blockNumber}-${event.indexInBlock}`
+
+  const isTagExists = await getOneBy(store, Tag, { name: name })
+
+  if (isTagExists) {
+    return MetaprotocolTxError.TagAlreadyExists
+  }
```
kdembler: Here you require new tags to have unique names, but later when processing tag update, you don't check it any more. I think it's okay if we allow this

metadata-protobuf/proto/WorkingGroups.proto
```diff
@@ -63,10 +63,97 @@ message ModeratePost {
...
+  required uint64 worker_id = 2;  
+}
+
+message DisallowTagToWorker {
+  // ID of tag
+  required string tag_id = 1;
+
+  // ID of worker
+  required uint64 worker_id = 2;  
+}
```
kdembler: Was there a requirement for this fine-grained control? I feel per-tag permissions will be cumbersome to deal with. For example you have a worker that's supposed to assign tags, a lead creates a new one and then needs to give permissions again. If there aren't other reasons, I suggest to allow/disallow workers to assign all tags.

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,270 @@
...
+  event: SubstrateEvent,
+  metadata: DecodedMetadataObject<IAssignTagToThread>,
+  workingGroup: WorkingGroup,
+  isLead: boolean,
+  workerId: number
+): Promise<any> {
+  const { tagId, threadId } = metadata
+
+  if (!isLead) {
+    await getOneByOrFail(store, TagToWorker, { tagId: tagId, workerId: workerId.toString() })
```
kdembler: As I understand it, you shouldn't use `getOneByOrFail` anywhere here. That will throw an error and possibly crash QN. Let's only use `MetaprotocolTxError` early returns

query-node/mappings/src/workingGroups.ts
```diff
@@ -709,74 +721,110 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
+async function processWorkingGroupsRemark(
+  store: DatabaseManager,
+  event: SubstrateEvent,
+  group: WorkingGroup,
+  decodedMetadata: DecodedMetadataObject<IRemarkMetadataAction> | null,
+  isLead = false,
+  workerId: number
+): Promise<any> {
+  // Moderate Post
+  if (decodedMetadata?.moderatePost) {
```
kdembler: I'm not sure about refactor of WG remarks processing. I think it's more error-prone to mix lead and workers logic. Previously only leads could moderate posts, now you allowed all FWG workers to do it. You also use phantom `0` worker ID for leads which can lead to errors later on as well.

I suggest to keep current logic for forum moderation and verifying validators intact in `workingGroups_LeadRemarked`. Instead you can extract all the `if`s for tag messages into a function and then call it both in `LeadRemarked` and `WorkerRemarked`

query-node/schemas/tag.graphql
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type tagToProposal @entity {
+  "Runtime entity identifier (EntityId)"
+  id: ID!
+  "Id of tag"
+  tagId: ID!
+  "Id of proposal"
+  proposalId: ID!
+}
```
kdembler: Do we need those intermediate entities? I think having `Thread` or `Proposal` entities have `tags` relation directly makes more sense, WDYT?

query-node/mappings/src/workingGroups.ts
```diff
@@ -709,74 +721,110 @@ export async function workingGroups_StatusTextChanged({ store, event }: EventCon
...
     }
     member.metadata.isVerifiedValidator = isVerified
     await store.save<MemberMetadata>(member.metadata)
-    await store.save<Membership>(member)
-  } else {
-    return invalidMetadata('Unrecognized remarked action')
+    return await store.save<Membership>(member)
+  }
+
+  if (decodedMetadata?.createTag) {
```
kdembler: I suggest to limit tag-management to Forum Working Group only. I don't see a need for other WGs to manage this tbh. Also the permissions you have currently only check worker ID which is duplicated across WGs.

kdembler (changes requested):
Hello @mkbeefcake, sorry for a late look at this. Left some comments, hope we can get back to this and move quickly! 🚀

metadata-protobuf/proto/WorkingGroups.proto
mkbeefcake: https://github.com/Joystream/pioneer/issues/4497
In Moderation Permissions table, There is the roles and permissions for tagging.

query-node/mappings/src/label/tag.ts
mkbeefcake: In updateTag(), I have implemented to get the tag by tagId for content update.

query-node/mappings/src/label/tag.ts
mkbeefcake: Yes. I will refactor all getOneByOrFail, getByIdOrFail to getOneBy and getById

query-node/mappings/src/workingGroups.ts
mkbeefcake: Yes. I will recover the original workingGroups.ts file and add the elseif branches for tagging system

query-node/schemas/tag.graphql
mkbeefcake: well... In the requirement, the tags verus thread/proposal are Many : one relationship. I checked this kind of implementation in the original code. For example, in proposals.graphql and proposalsEvents.graphql, the proposal contains the proposalStatusUpdates[] relation (one : many) by using @derivedFrom(field: proposal). The proposalStatusUpdatedEvent contains the proposal field to indicate which proposal is associated. 
If I follow this mechanism, each tag data needs to contain proposalid or threadid, and there will be happened same tag content with different proposalid or threadid. so I suggested to use intermediate mapping structure for it. also this will not apply any changes to proposal and thread associated code and can be integrated.  I hope to hear from you about this.

query-node/mappings/src/workingGroups.ts
mkbeefcake: Yes. I will update the code

metadata-protobuf/proto/WorkingGroups.proto
kdembler: Yes, but it doesn't mention need for managing those permissions for each tag separately. We can simplify this and just let the lead select workers that can assign/unassign _all_ labels

query-node/mappings/src/label/tag.ts
kdembler: On second thought, let's keep the unique name requirement, but also implement it when updating a tag.

query-node/schemas/tag.graphql
kdembler: Please refer to implementation of legacy forum tags:
1. https://github.com/Joystream/joystream/blob/8e88b27c27204d06f9930c9f1cac129ab4671147/query-node/schemas/forum.graphql#L118-L127
2. https://github.com/Joystream/joystream/blob/8e88b27c27204d06f9930c9f1cac129ab4671147/query-node/schemas/forum.graphql#L115
Let's follow the same approach

query-node/mappings/src/workingGroups.ts
```diff
@@ -739,6 +749,22 @@ export async function workingGroups_LeadRemarked({ store, event }: EventContext
...
+  } else if (metadata?.assignTagsToProposal && group.name === 'forumWorkingGroup') {
+    return await processAssignTagsToProposal(store, metadata?.assignTagsToProposal, true, 0)
+  } else if (metadata?.unassignTagsFromThread && group.name === 'forumWorkingGroup') {
+    return await processUnassignTagsFromThread(store, metadata?.unassignTagsFromThread, true, 0)
+  } else if (metadata?.unassignTagsFromProposal && group.name === 'forumWorkingGroup') {
+    return await processUnassignTagsFromProposal(store, metadata?.unassignTagsFromProposal, true, 0)
+  } else if (metadata?.allowTagToWorker && group.name === 'forumWorkingGroup') {
+    return await processAllowTagToWorker(store, metadata?.allowTagToWorker)
+  } else if (metadata?.disallowTagToWorker && group.name === 'forumWorkingGroup') {
+    return await processDisallowTagToWorker(store, metadata?.disallowTagToWorker)
```
kdembler: Let's extract those to a function as suggested earlier so the logic is not repeated. Additionally, with the current conditions if somebody from outside forum working group would try to work with tags, the message would just be ignored. Instead, it should return `invalidMetadata` error as is done above:
```
    if (group.name !== 'membershipWorkingGroup') {
      return invalidMetadata(`The ${group.name} can't verify the validator's membership`)
    }
```
I think what makes most sense here:
1. Replace all the conditions (same for regular worker message) with single:
```ts
else if (metadata?.createTag || metadata?.updateTag || ...) {
  return await processTagMessage(metadata, group.name, isLead, workerId)
}
```
2. In `processTagMessage` do:
  2.1. Check working group and return error if not forum
  2.2. Do permissions check so we don't have to duplicate across specific message handlers
  2.3. Have this long if-else for specific messages

WDYT?

kdembler: Also, please don't use `0` as worker ID for lead, that's misleading. If it's not needed for lead, just make it an optional parameter and skip

query-node/schemas/forum.graphql
```diff
@@ -113,8 +113,12 @@ type ForumThread @entity {
 
   "Assigned thread tags"
   tags: [ForumThreadTag!]
+
+  "New tag/label properties"
+  newTags: [Tag!]
```
kdembler, line 118: We will need to merge those functionalities actually, but I'm fine with this as a temporary measure to not bloat this PR too much. Once we're done with this PR we should most likely remove all the old tags related stuff

kdembler (changes requested):

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+    tag.type = type
+  }
+
+  if (visibility) {
+    tag.visibility = visibility
+  }
+
+  await store.save<Tag>(tag)
+
+  logger.info('Tag has been created', { name })
```
kdembler:
```suggestion
  logger.info('Tag has been updated', { tagId, name, description })
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  const forumThread: ForumThread | undefined = await getById(store, ForumThread, threadId)
+  if (!forumThread) {
+    return MetaprotocolTxError.TagInvalidThreadId
+  }
+
+  const currentTagIds = (forumThread.newTags || []).map((t) => t.id)
+  const tagIdsToSet = _.union(currentTagIds, tagIds)
+  if (tagIdsToSet) {
```
kdembler: Do we need this `if`? I don't think `tagIdsToSet` will ever be nullish

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+    const tags = await Promise.all(tagIdsToSet.map(async (tagId: string) => await getById(store, Tag, tagId)))
+    forumThread.newTags = tags.filter((t): t is Tag => !!t)
+    await store.save<ForumThread>(forumThread)
+  }
+
+  logger.info('new Tag is assigned to ForumThread', { tagIds, threadId })
```
kdembler:
```suggestion
  logger.info('Assigned forum thread tags', { threadId, tagIdsToSet })
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  const proposal: Proposal | undefined = await getById(store, Proposal, proposalId)
+  if (!proposal) {
+    return MetaprotocolTxError.TagInvalidProposalId
+  }
+
+  const currentTagIds = (proposal.tags || []).map((t) => t.id)
+  const tagIdsToSet = _.union(currentTagIds, tagIds)
+  if (tagIdsToSet) {
```
kdembler: Same as above

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+    const tags = await Promise.all(tagIdsToSet.map(async (tagId: string) => await getById(store, Tag, tagId)))
+    proposal.tags = tags.filter((t): t is Tag => !!t)
+    await store.save<Proposal>(proposal)
+  }
+
+  logger.info('new Tag is assigned to proposal', { tagIds, proposalId })
```
kdembler:
```suggestion
  logger.info('Assigned proposal tags', { proposalId, tagIdsToSet })
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+      return MetaprotocolTxError.TagPermNotAllowed
+    }
+  }
+
+  const forumThread: ForumThread | undefined = await getById(store, ForumThread, threadId)
+  if (!forumThread) {
+    return MetaprotocolTxError.TagInvalidThreadId
+  }
+
+  const remainedTags = (forumThread.newTags || []).filter((t) => !(tagIds || []).includes(t.id))
```
kdembler:
```suggestion
  const remainingTags = (forumThread.newTags || []).filter((t) => !(tagIds || []).includes(t.id))
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+  forumThread.newTags = remainedTags
+  await store.save<ForumThread>(forumThread)
+
+  logger.info('tag is unassigned from forumThread', { tagIds, threadId })
```
kdembler:
```suggestion
  logger.info('Unassigned forum thread tags', { threadId, remainingTags })
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+      return MetaprotocolTxError.TagPermNotAllowed
+    }
+  }
+
+  const proposal: Proposal | undefined = await getById(store, Proposal, proposalId)
+  if (!proposal) {
+    return MetaprotocolTxError.TagInvalidProposalId
+  }
+
+  const remainedTags = (proposal.tags || []).filter((t) => !(tagIds || []).includes(t.id))
```
kdembler:
```suggestion
  const remainingTags = (proposal.tags || []).filter((t) => !(tagIds || []).includes(t.id))
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+  proposal.tags = remainedTags
+  await store.save<Proposal>(proposal)
+
+  logger.info('tag is unassigned from proposal', { tagIds, proposalId })
```
kdembler:
```suggestion
  logger.info('Unassigned proposal tags', { proposalId, remainingTags })
```

query-node/mappings/src/label/tag.ts
```diff
@@ -0,0 +1,289 @@
...
+  })
+
+  if (!tagPermittedWorker) {
+    const newTagToWorker = new TagPermittedWorker({
+      workerId: assigneeId,
+    })
+    await store.save<TagPermittedWorker>(newTagToWorker)
+  }
+
+  logger.info('TagPermittedWorker has been allowed', { assigneeId })
```
kdembler: Let's move this log into the `if` above and add an `else` branch that will return an error indicating that the user already has permissions. Let's do a similar thing for disallow function

PR #4989, opened 2023-12-04
colossus: do not serve content from pending folder

mnaamani:
It is not clearly obvious how the node will behave, if while service a file from temp pending, the file is moved/renamed into the uploads folder. Will the move fail or will the http request handler to getFile fail. So it seems safer and more predictable behavior to only serve objects in the main uploads folder.

This PR reverts change added in https://github.com/Joystream/joystream/pull/4971 which allows serving objects in pending folder.

### DIFF
--- a/storage-node/src/services/webApi/controllers/filesApi.ts
+++ b/storage-node/src/services/webApi/controllers/filesApi.ts
@@ -39,10 +39,8 @@ export async function getFile(
 
     const dataObjectId = new BN(req.params.id)
     const uploadsDir = res.locals.uploadsDir
-    const pendingObjectsDir = res.locals.pendingDataObjectsDir
-    const pending = res.locals.acceptPendingObjectsService.getPendingDataObject(dataObjectId.toString())
 
-    const fullPath = path.resolve(pending ? pendingObjectsDir : uploadsDir, dataObjectId.toString())
+    const fullPath = path.resolve(uploadsDir, dataObjectId.toString())
 
     const fileInfo = await getFileInfo(fullPath)
     const fileStats = await fsPromises.stat(fullPath)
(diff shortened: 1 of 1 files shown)


zeeshanakram3 (commented):
Looks good to me, the only implication of this change is that even the uploader won't able to access the file till it's has been accepted. 

The other possible solution (and might not be trivial to implement) is that whenever the conflict happens we let the HTTP request fail and the move should always succeed. WDYT?

mnaamani:
> Looks good to me, the only implication of this change is that even the uploader won't able to access the file till it's has been accepted.

Yes that is unfortunate, best case scenario is they would have to wait up to 6s for file to become accessible.

> 
> The other possible solution (and might not be trivial to implement) is that whenever the conflict happens we let the HTTP request fail and the move should always succeed. WDYT?

Yes I can think of a few ways, but it does seem complex. The root issue is of course having separate location for pending folder. Would it be less complex if we eliminated the pending folder, and track the objects that need to be "confirmed" with onchain tx in a local db, or simply a JSON file?

zeeshanakram3:
>  The root issue is of course having separate location for pending folder. Would it be less complex if we eliminated the pending folder, and track the objects that need to be "confirmed" with onchain tx in a local db, or simply a JSON file?

Sounds good. Maybe this can be implemented as a feature in `storage-squid` (i.e. as a mutation that will store relevant info in storage-squid db)

zeeshanakram3 (approved):

mnaamani:
> Sounds good. Maybe this can be implemented as a feature in `storage-squid` (i.e. as a mutation that will store relevant info in storage-squid db)

Worth considering.

PR #4993, opened 2023-12-07
skip url endpoints of buckets operated by the node

### LINKED ISSUE #4990: Colossus: Sync/Cleanup - Do not skip fetching from all buckets operated by worker.
## Problem

When preparing download tasks, and cleanup tasks, the node is preparing list of urls/endpoints to fetch and check availability of objects at those endpoints.

https://github.com/Joystream/joystream/blob/5d911ce1bd546da16c983c852203a2a6eac8d8ea/storage-node/src/services/sync/synchronizer.ts#L117

https://github.com/Joystream/joystream/blob/5d911ce1bd546da16c983c852203a2a6eac8d8ea/storage-node/src/services/sync/cleanupService.ts#L126

The implementation attempts to avoid using endpoints of buckets associated with the worker running the current node. Although the intent is correct: "do not try to fetch from self", it ignores the case where the worker is operating buckets on different nodes (which is recommended) and needs to fetch an object from one of their remote nodes.

## Solution(s)
1. Instead of passing the current worker id, pass in the current buckets being operated by the node and skip those buckets endpoints.
or
2. Add a new argument to the server command with the public endpoint of the node, and only that will be used to identify endpoints not to connect to.

## Prevent Self-Connecting

This also brings up a subtle issue that can arise from operators that mis-configure endpoints, causing a node to connect to itself to fetch an object (eg. operator operates multiple buckets, but has only configured the current node with one bucket, so will attempt to fetch/check presence of an object from/on a remote bucket which is actually pointing to itself). This isn't really a problem for sync, but more for cleanup since it will be a false-positive counting to replication count.

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4990

- When selecting "candidate" urls to fetch from, instead of ignoring all endpoints of buckets operated by worker id, we select all urls of buckets except the ones operated by the node. This ensures we don't miss any remote urls of buckets (even if they are operated by the same worker) on a different host.
- Just in-case there is a misconfiguration by the operator with respect to url endpoints, and to avoid the node connecting to itself, we send an http header `X-COLOSSUS-HOST-ID` with a new random value when the node starts up, which is checked by express server when handling requests. If the value belongs to the node, the request is dropped.

### DIFF
--- a/storage-node/src/commands/dev/cleanup.ts
+++ b/storage-node/src/commands/dev/cleanup.ts
@@ -54,7 +54,7 @@ export default class DevCleanup extends Command {
     logger.info('Cleanup...')
 
     try {
-      await performCleanup([bucketId], flags.cleanupWorkersNumber, qnApi, flags.uploads)
+      await performCleanup([bucketId], flags.cleanupWorkersNumber, qnApi, flags.uploads, '')
     } catch (err) {
       logger.error(err)
       logger.error(stringify(err))
--- a/storage-node/src/commands/dev/cleanup.ts
+++ b/storage-node/src/commands/dev/cleanup.ts
@@ -54,7 +54,7 @@ export default class DevCleanup extends Command {
     logger.info('Cleanup...')
 
     try {
-      await performCleanup(flags.workerId, [bucketId], flags.cleanupWorkersNumber, qnApi, flags.uploads)
+      await performCleanup([bucketId], flags.cleanupWorkersNumber, qnApi, flags.uploads)
     } catch (err) {
       logger.error(err)
       logger.error(stringify(err))
--- a/storage-node/src/commands/dev/sync.ts
+++ b/storage-node/src/commands/dev/sync.ts
@@ -69,7 +69,6 @@ export default class DevSync extends Command {
     try {
       await performSync(
         undefined,
-        flags.workerId,
         [bucketId],
         flags.syncWorkersNumber,
         flags.syncWorkersTimeout,
--- a/storage-node/src/commands/dev/sync.ts
+++ b/storage-node/src/commands/dev/sync.ts
@@ -74,6 +74,7 @@ export default class DevSync extends Command {
         flags.syncWorkersTimeout,
         qnApi,
         flags.uploads,
+        '',
         flags.dataSourceOperatorUrl
       )
     } catch (err) {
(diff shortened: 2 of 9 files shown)


zeeshanakram3 (commented):
> causing a node to connect to itself to fetch an object (eg. the operator operates multiple buckets, but has only configured the current node with one bucket, so will attempt to fetch/check presence of an object from/on a remote bucket which is actually pointing to itself).

> to avoid the node connecting to itself, we send an http header X-COLOSSUS-HOST-ID with a new random value when the node starts up, which is checked by express server when handling requests. If the value belongs to the node, the request is dropped.

Instead of doing this, shouldn't we consult QN i.e. if node operator didn't specify any bucket/s with the current node, and [storageBucket.operatorMetadata.nodeEndpoint](https://github.com/Joystream/joystream/blob/master/query-node/schemas/storage.graphql#L63) of that bucket/s is equal to the endpoint [^1] of the current running node, then we just ignore it.
 
[^1]: basically equal to the endpoints of buckets served by the current node.
Just extending that maybe we should add a condition upon storage-node bootstrapping that all the buckets specified should have the same endpoint, so the operators could not mistakenly serve multiple buckets with different node endpoints on the same server

mnaamani:
> Instead of doing this, shouldn't we consult QN i.e. if node operator didn't specify any bucket/s with the current node, and [storageBucket.operatorMetadata.nodeEndpoint](https://github.com/Joystream/joystream/blob/master/query-node/schemas/storage.graphql#L63) of that bucket/s is equal to the endpoint [1](#user-content-fn-1-fd4f98553cb7bf5f64643c3d7b4d0ee9) of the current running node, then we just ignore it.

That a good suggestion. I added this check in [a3cb8e5](https://github.com/Joystream/joystream/pull/4993/commits/a3cb8e5e5480623b7ba63a59c7921db36c054037)

I would argue, there is still value in using the header as well to catch the case where two url endpoints might be different (by string comparison), but still actually point to the same host (ie DNS records point to same host, irrespective if intentional or accidental)

mnaamani:
Just a note on your comment "if node operator didn't specify any bucket/s with the current node" - if no buckets are specified as command line arguments, then all buckets the worker is associated with are selected. 

Regarding "add a condition upon storage-node bootstrapping that all the buckets specified should have the same endpoint". Are you suggesting the node should exit with failure if that is not the case or just log a warning? It could be useful to do an info log so the operator is aware what the metadata endpoint url is for each bucket they are serving.

zeeshanakram3:
> Regarding "add a condition upon storage-node bootstrapping that all the buckets specified should have the same endpoint". Are you suggesting the node should exit with failure if that is not the case or just log a warning? It could be useful to do an info log so the operator is aware what the metadata endpoint url is for each bucket they are serving.

I was suggesting exiting the node with failure, but since as you pointed out two different endpoints can resolve to the same host, I think showing a warning log would be ideal.

zeeshanakram3 (approved):
LGTM

Issue #4995, opened 2023-12-08
colossus: crash when fetch failed


kdembler:
```
2023-12-08 09:46:18:4618 ESC[31merrorESC[39m: Sync - fetching data error for https://storage.freakstatic.com/storage/api/v1/files/1348170: Error: Inva
lid file hash. Expected: gW93EBBCozPxBRche8Xx79qJPjjmsHyjFa66yUWAeZWpuf - real: gWA8ejdyVhcaiwBSc9AqtN8nQidkDJinZEFL498TQRxrR2
2023-12-08 09:46:18:4618 ESC[33mwarnESC[39m: Cleaning up file /data/temp/c726755c-2904-4431-a18a-bf0200b5474b
2023-12-08 09:47:07:477 ESC[31merrorESC[39m: Sync - fetching data error from https://sieemmastorage.com/storage/api/v1/state/data-objects: Error: Time
out of 120000ms exceeded
node:events:495
      throw er; // Unhandled 'error' event
      ^

Error: aborted
    at connResetException (node:internal/errors:720:14)
    at TLSSocket.socketCloseListener (node:_http_client:461:19)
    at TLSSocket.emit (node:events:529:35)
    at node:net:350:12
    at TCP.done (node:_tls_wrap:614:7)
Emitted 'error' event on Response instance at:
    at IncomingMessage.emit (node:events:517:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  code: 'ECONNRESET'
}

Node.js v18.18.2
2023-12-08 09:53:21:5321 ESC[33mwarnESC[39m: Sync - cannot get operator URLs for 1362932
```

kdembler:
@zeeshanakram3 I'm on 3.9.1 and I still see the issue:
```
2023-12-19 09:26:11:2611 ESC[33mwarnESC[39m: Sync - cannot get operator URLs for 1348436
2023-12-19 09:26:11:2611 ESC[33mwarnESC[39m: Sync - cannot get operator URLs for 1348170
node:events:495
      throw er; // Unhandled 'error' event
      ^

Error: aborted
    at connResetException (node:internal/errors:720:14)
    at TLSSocket.socketCloseListener (node:_http_client:461:19)
    at TLSSocket.emit (node:events:529:35)
    at node:net:350:12
[...]

PR #4998, opened 2023-12-12
[colossus] fix: unhandled error event on `superagent.get` instance

zeeshanakram3:
addresses #4995

### DIFF
--- a/storage-node/CHANGELOG.md
+++ b/storage-node/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 3.9.1
+
+- **FIX**: Added event handler to handle `error` event on `superagent.get` instance. Previously the unhandled error was causing the application to crash.
+
 ### 3.9.0
 
 - Increase default interval between sync runs. Start sync run immediately do not wait initial interval on startup before starting sync. Adds additional optional argument to specify retry interval on failure. [#4924](https://github.com/Joystream/joystream/pull/4924)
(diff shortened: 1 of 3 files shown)


mnaamani (approved):

mnaamani:
The fix addresses the node crashing on request errors which is definitely needed.

Some ideas for further enhancements:

1. Timeout should be a function of object size
The original issue showed:
```
Sync - fetching data error from https://sieemmastorage.com/storage/api/v1/state/data-objects: Error: Time
out of 120000ms exceeded
```
So the node was configured with a 2minute timeout to fetch. It seems a bit arbitrary and wouldn't work for very large files.
In general I think a timeout should be much larger, but a more effective mechanism to drop the connection would be monitoring the download progress. There is a `superagent-progress` plugin that could be useful.

2. Keep some state about prior sync run
Moreover there isn't any kind of "feedback loop" for the synchronizer worker to make more intelligent selection on next run.

kdembler:
@mnaamani Was that a timeout for the full request or just a timeout for initial response (without the full content?)

mnaamani:
> @mnaamani Was that a timeout for the full request or just a timeout for initial response (without the full content?)

That two minute timeout looks like it was just for the the fetching of the data-objects list. And is for the full response or "deadline" because code uses this call `.timeout(120000)`

Now that you ask, there are in-fact two possible timeouts that can be configured the response and deadline with this call for example:

```js
.timeout({
        response: 5000,  // Wait 5 seconds for the server to start sending,
        deadline: 60000, // but allow 1 minute for the file to finish loading.
      })
```

fyi: @zeeshanakram3 @kdembler

PR #5000, opened 2023-12-12
temp folder refactor

### LINKED ISSUE #4977: colossus: fully qualified path for temp folder
Background

The temp folder path where active uploads to the node and objects being fetched by sync worker from other nodes is not configurable, but is always created as a sub folder under the uploads directory (long term storage of data objects). 

Some ideas for scaling storage capacity of the node will require separation of temp files from the long term storage path. If we decide to use object storage (s3 compatible) then its not possible to store incomplete download files there.

Solution

Add new parameter to `server` command to specify a fully qualified path for the temp folder to use.

Challenges ?

Typically the path to the uploads folder will be on a volume with considerable capacity, and this volume might be mounted from a network storage cluster. It might be more suitable for the temp folder to be a path on the local host file system with not as much disk capacity. Should the node actively try to manage the number of downloads and uploads at a rate that takes the temp folder size into consideration?

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/4977

- Temp folder can now be configured with argument with a fully qualified path, with a check that it is not the same as the uploads folder.
- Added check to ensure logs folder is not the same directory as the uploads folder.
- Improved initialization of the uploads directory, reading data objects from the uploads folder filters out any files/subfolders without a name matching the data-object-id (ie. a number)
- Improved initialization of temp directory. It will not be recursively deleted on startup (risky).
- Renamed the 'dev' commands to 'util' commands although I have some reservation about using the cleanup command outside of the context of the running server.. Perhaps we should drop them.

### DIFF
--- a/storage-node/src/commands/server.ts
+++ b/storage-node/src/commands/server.ts
@@ -53,7 +53,7 @@ export default class Server extends ApiCommandBase {
     }),
     tempFolder: flags.string({
       description:
-        'Directory to store tempory files during sync and upload (absolute path).\n,Temporary directory (absolute path). If not specified a subfolder under the uploads directory will be used.',
+        'Directory to store tempory files during sync and upload (absolute path).\nIf not specified a subfolder under the uploads directory will be used.',
     }),
     port: flags.integer({
       char: 'o',
@@ -230,9 +230,10 @@ Supported values: warn, error, debug, info. Default:debug`,
 
     if (!flags.tempFolder) {
       logger.warn(
-        'It is recommended to specify a unique file path for temporary files.' +
-          'For now a temp folder under the uploads folder will be used.' +
-          'In future this will be warning will become and error!'
+        'You did not specify a path to the temporary directory. ' +
+          'A temp folder under the uploads folder willl be used. ' +
+          'In a future release passing an absolute path to a temporary directory with the ' +
+          '"tempFolder" argument will be required.'
       )
     }
 
--- a/storage-node/src/commands/server.ts
+++ b/storage-node/src/commands/server.ts
@@ -1,14 +1,11 @@
 import { flags } from '@oclif/command'
 import { ApiPromise } from '@polkadot/api'
-import { KeyringPair } from '@polkadot/keyring/types'
-import { PalletStorageStorageBucketRecord } from '@polkadot/types/lookup'
-import fs from 'fs'
+import sleep from 'sleep-promise'
 import _ from 'lodash'
 import path from 'path'
-import rimraf from 'rimraf'
-import sleep from 'sleep-promise'
-import { promisify } from 'util'
-import ApiCommandBase from '../command-base/ApiCommandBase'
+import fs from 'fs'
+import { PalletStorageStorageBucketRecord } from '@polkadot/types/lookup'
+import { KeyringPair } from '@polkadot/keyring/types'
 import { customFlags } from '../command-base/CustomFlags'
 import { loadDataObjectIdCache } from '../services/caching/localDataObjects'
 import logger, { DatePatternByFrequency, Frequency, initNewLogger } from '../services/logger'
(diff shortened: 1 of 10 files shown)


storage-node/src/commands/server.ts
```diff
@@ -219,16 +228,25 @@ Supported values: warn, error, debug, info. Default:debug`,
...
 
-    await recreateTempDirectory(flags.uploads, TempDirName)
-
-    if (fs.existsSync(flags.uploads)) {
-      await loadDataObjectIdCache(flags.uploads, TempDirName, PendingDirName)
+    if (!flags.tempFolder) {
+      logger.warn(
+        'It is recommended to specify a unique file path for temporary files.' +
+          'For now a temp folder under the uploads folder will be used.' +
+          'In future this will be warning will become and error!'
```
```
In future this will be warning will become and error!
```

Can you please rephrase that?

storage-node/src/commands/server.ts
```diff
@@ -157,11 +160,19 @@ Supported values: warn, error, debug, info. Default:debug`,
...
-    const logSource = `StorageProvider_${flags.worker}`
+    const api = await this.getApi()
+
+    if (flags.dev) {
+      await this.ensureDevelopmentChain()
+    }
+
+    if (flags.logFilePath && path.relative(flags.logFilePath, flags.uploads) === '') {
+      this.error('Paths for logs and uploads must be unique.')
+    }
```
zeeshanakram3, line 171:
```suggestion
    if (flags.logFilePath && path.relative(flags.logFilePath, flags.uploads) === '..') {
      this.error('Paths for logs and uploads must be unique.')
    }
```

Shouldn't it be `path.relative(flags.logFilePath, flags.uploads) === '..'`, I tested the `path.relative` function as `path.relative('/uploads/temp', '/uploads')` and I got `..` as return value

storage-node/src/commands/server.ts
```diff
@@ -219,16 +228,25 @@ Supported values: warn, error, debug, info. Default:debug`,
...
+          'In future this will be warning will become and error!'
+      )
     }
 
-    if (flags.dev) {
-      await this.ensureDevelopmentChain()
+    const tempFolder = flags.tempFolder || path.join(flags.uploads, TempDirName)
+
+    if (path.relative(tempFolder, flags.uploads) === '') {
+      this.error('Paths for temporary and uploads folders must be unique.')
```
zeeshanakram3, line 243:
```suggestion
    if (path.relative(tempFolder, flags.uploads) === '..') {
      this.error('Paths for temporary and uploads folders must be unique.')
```

Please see the other comment

zeeshanakram3 (commented):
Looks Good. Can you please add `--tempFolder` flag in docker-compose file, as an example on how to use it?

storage-node/src/commands/server.ts
mnaamani: Here is was only trying to protect against the two paths being the same location. I was not concerned with temp or logs being subfolders of the main uploads directory. That is why I was comparing with `''` null string.

Originally I was doing `if (lags.logFilePath === flags.uploads) { }` but I decided to use `path.relative()` to take into account trailing `/` path separator..
 
Given that I added proper filtering of filenames when they are loaded from uploads directory, and if we assume the file names of logs and temp files will not clash with the object id name, perhaps this constraint is not really necessary at all?

storage-node/src/commands/server.ts
mnaamani: Will rephrase the whole warning to:
```
You did not specify a path to the temporary directory. A temp folder under the uploads folder will
be used. In a future release passing an absolute path to a temporary directory with the tempFolder argument
will become mandatory.
```

WDYT?

storage-node/src/commands/server.ts
mnaamani: commented here https://github.com/Joystream/joystream/pull/5000#discussion_r1435673895

storage-node/src/commands/server.ts
mnaamani: done in [a287fdc](https://github.com/Joystream/joystream/pull/5000/commits/a287fdc229c42642bb89b838ae0fe795f8c21ffe)

mnaamani:
> Looks Good. Can you please add `--tempFolder` flag in docker-compose file, as an example on how to use it?

Added in https://github.com/Joystream/joystream/pull/5000/commits/3dd8e0aa733430733d5a23d9b23cee3ade388fb2

zeeshanakram3 (approved):
LGTM. I believe the version bump and CHANGELOG will be done in separate release PR?

mnaamani:
> LGTM. I believe the version bump and CHANGELOG will be done in separate release PR?

Yes will do that next.

Issue #4732, opened 2023-04-09
Expose information about QN used in core nodes


bedeho:
# Background

Multiple different services now are powered by a QN, specifically

- Argus
- Colossus
- YT-synch

It may be useful for leads and others to understand what QN version a given operator is using, as this may be important in preparation for upgrades and debugging. This capability was explicitly requested by a lead around the Ephesus upgrade.

# Proposal

Add some sort of public diagnostics endpoint which describes this, among other possibly useful things. I don't think it is a security or denial of service risk if done properly.

0x2bc:
let's adjust this requirement in the following way
> It may be useful for leads and others to understand what QN version a given operator is using, **and what's QN URL a given operator is using (localhost or third party one)**

traumschule:
> "queryNodeStatus":{"url":"https://query.joystream.org/graphql","chainHead":1740687,"blocksProcessed":1739709}}

Just need to add the version to the status output.

0x2bc:
@traumschule this works for DP. 
For SP node we are lack both QN URL and QN version

PR #5001, opened 2023-12-13
Storage Squid integration with Argus/Colossus

### LINKED ISSUE #4957: Storage squid support for Argus and Colossus
There is a consensus among SWG, DWG and JSG that we should migrate our Colossus and Argus nodes away from QN and create a storage-focused squid to serve those nodes' data needs.

Some quick notes on the current status:
1. We want to create a squid and place it in a new repo.
2. Storage squid should borrow as much code as possible from existing Orion mappings.
3. We should use the latest ArrowSquid release with focus on long-term maintenance.
4. One thing I've recently discovered is missing from QN/Orion mappings is some kind of reference to which storage operator accepted the upload of a given object. It would be nice to include that.
5. @zeeshanakram3 already started work on some basic version

### LINKED ISSUE #4983: colossus: command docs
colossus commands documentation is outdated and doesn't seem to get generated on build. Several changes to command line args have been made since the docs were last generated.

solution: update npm scripts to generate the docs (like distributor node)

zeeshanakram3:
closes #4957

This PR:

- [integrates Storage-Squid with Distributor Node](https://github.com/Joystream/joystream/commit/de5166d6473b5ad17d166ee51d4890fa15b1e87e)
- [integrates Storage-Squid with Storage Node](https://github.com/Joystream/joystream/commit/8f6e650de78e03d8599b9c896d3ec43db8c56997)
- adds NPM script to regenerate storage-node CLI docs + regenerated document (fixes #4983)
- adds `docker-compose.storage-squid.yml` file to run storage-squid services
- updates integration tests bash scripts to start `storage-squid` services when running Argus/Colossus
- [Enable telemetry for storage-squid's `graphql-server`](https://github.com/Joystream/joystream/pull/5001/commits/986d4d08e02648ff58c8426b6a604e14d7a49f03)


distributor-node/config.yml
```diff
@@ -1,6 +1,6 @@
 id: test-node
 endpoints:
-  queryNode: http://localhost:8081/graphql
+  queryNode: http://localhost:4352/graphql
```
kdembler: I think we should make it clear that this is not _the_ query node now. I suggest to rename to `storageSquid` or just `squid`

kdembler (commented):
Amazing to see this realize, great work @zeeshanakram3 

I'll try to get to proper code review later, for now tried Argus out and seeing some problems:
1. Getting `{"type":"exception","message":"Unexpected end of JSON input"}` from `/status` when squid is not available
2. On start (serving bucket `0:0`) I'm seeing
```
2023-12-28 22:59:51:5951 ContentService info: ContentService initializing...
{
    "supportedObjects": 0,
    "filesCountOnStartup": 0,
    "cacheItemsCountOnStartup": 0
}
2023-12-28 22:59:51:5951 ContentService info: ContentService initialized
{
    "filesDropped": 0,
    "cacheItemsDropped": 0,
    "contentSizeSum": 0
}
```
I expected the `supportedObjects` to give proper number. I can  get objects fine

.env
```diff
@@ -169,3 +169,22 @@ TELEMETRY_ENDPOINT=http://collector:4318
...
+## Storage-Squid configuration
+
+# Db config
+SQUID_DB_HOST=squid_db
+SQUID_DB_NAME=squid
+SQUID_DB_PASS=squid
+SQUID_DB_PORT=23332
+
+# Processor service prometheus port
+SQUID_PROCESSOR_PROMETHEUS_PORT=3337
```
mnaamani: `PROCESSOR_PROMETHEUS_PORT` earlier in the file is also assigned 3337
But this is only a problem if orion processor is also running on same host.

distributor-node/config.yml
mnaamani: I don't mind the name so much. Just wondering where the port number selection came from, is it just 4250 (orion graphql port) + 2 ?

distributor-node/src/services/networking/NetworkingService.ts
```diff
@@ -1,27 +1,28 @@
...
+import { StateCacheService } from '../cache/StateCacheService'
+import { LoggingService } from '../logging'
 import { parseAxiosError } from '../parsers/errors'
 import { PendingDownload, PendingDownloadStatusType } from './PendingDownload'
-import Mime from 'mime/lite'
+import { QueryFetchPolicy, QueryNodeApi } from './query-node/api'
+import { DataObjectDetailsFragment } from './query-node/generated/queries'
+import { DistributionBucketOperatorStatus } from './query-node/generated/schema'
+import { RuntimeApi } from './runtime/api'
+import { StorageNodeApi } from './storage-node/api'
```
mnaamani: I think we discussed this once, but it is hard to see what imports were added or removed when they are also being re-ordered.

distributor-node/package.json
```diff
@@ -138,6 +138,7 @@
     "prepack": "rm -rf lib && tsc -b && oclif-dev manifest && generate:all",
     "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\"",
     "version": "generate:docs:cli && git add docs/cli/*",
+    "generate:schema:graphql": "docker run --rm joystream/storage-squid:latest npm run get-graphql-schema > src/services/networking/query-node/schema.graphql",
```
mnaamani:
```suggestion
    "generate:schema:graphql": "docker inspect joystream/storage-squid:latest && docker run --rm joystream/storage-squid:latest npm run get-graphql-schema > src/services/networking/query-node/schema.graphql",
```

mnaamani: `get-graphql-schema` is existing with value `0` when graphql endpoint is unreachable:

```
FetchError: request to http://localhost:4352/graphql failed, reason: connect ECONNREFUSED 127.0.0.1:4352
```

In cases docker run fails or scripts doesn't connect, the resulting `src/services/networking/query-node/schema.graphql` is overwritten with no content.

This step is done manually and not part of automated build step, but it would be better for it to fail in a better way.

I'm guessing on my machine [5 seconds](https://github.com/Joystream/storage-squid/blob/7402805e1f969c899a29bfd52f677e026aba05c6/scripts/get-graphql-schema.sh#L11C8-L11C8) was not long enough for graphql-server to startup.

Interactively it works fine..

`docker run -it joystream/storage-squid:latest sh`

docker-compose.storage-squid.yml
```diff
@@ -0,0 +1,79 @@
...
+    restart: unless-stopped
+    env_file:
+      - .env
+    environment:
+      - DB_NAME=${SQUID_DB_NAME}
+      - DB_PASS=${SQUID_DB_PASS}
+      - DB_PORT=${SQUID_DB_PORT}
+      - DB_HOST=${SQUID_DB_HOST}
+      - GQL_PORT=${SQUID_GQL_PORT}
+      - ARCHIVE_GATEWAY_URL=${SQUID_ARCHIVE_GATEWAY_URL}
```
mnaamani, line 35: SQUID_ARCHIVE_GATEWAY_URL is empty in .env, what is the default if `ARCHIVE_GATEWAY_URL` is empty? Does it sync directly from chain using rcp endpoint?

tests/network-tests/start-storage.sh
```diff
@@ -1,21 +1,20 @@
 set -e
 
 TMP=$0
-THIS_DIR=`dirname $TMP`
+THIS_DIR=$(dirname $TMP)
```
mnaamani: Why did you need to change to use `$(..)`?

Please see https://github.com/Joystream/joystream/pull/4917
After than PR I was meaning to apply the same for all scripts on master branch but never got around to it.

tests/network-tests/start-storage.sh
```diff
@@ -1,21 +1,20 @@
...
 
 echo "Staring storage infrastructure"
 
-HOST_IP=`$THIS_DIR/get-host-ip.sh`
```
mnaamani, line 8: same question about switching to `$(...)`

tests/network-tests/start-storage.sh
```diff
@@ -1,21 +1,20 @@
...
+# Start Storage-Squid
+docker-compose -f $THIS_DIR/../../docker-compose.storage-squid.yml up -d
```
mnaamani, line 9: I think we want to also start the storage-squid in project root 
`start.sh` and `start-multi-storage.sh`

distributor-node/src/services/networking/NetworkingService.ts
```diff
@@ -65,6 +67,12 @@ export class NetworkingService {
     this.downloadQueue.on('error', (err) => {
       this.logger.error('Data object download failed', { err })
     })
+
+    this.initRuntimeApi().catch((err) => this.logger.error('Runtime API initialization failed:', { err }))
```
mnaamani, line 71: I suppose it is not critical for runtimeApi to be functioning, as only place i see it used is in the public api endpoint to get the status. But if the api is not initialized and the endpoint is called and `getQueryNodeStatus()` thorws, we should ensure the request handler catches it and return a HTTP 503 status rather than cause the express server to cause the process to exit with an unhandled expection.

distributor-node/src/services/networking/query-node/api.ts
```diff
@@ -79,33 +65,11 @@ export class QueryNodeApi {
...
-        const definition = getMainDefinition(query)
-        return definition.kind === 'OperationDefinition' && definition.operation === 'subscription'
-      },
-      wsLink,
-      queryLink
-    )
-
     this.apolloClient = new ApolloClient({
-      link: splitLink,
+      link: from([errorLink, new HttpLink({ uri: this.config.endpoints.queryNode, fetch })]),
```
mnaamani: I don't see where this `from()` function is imported. Is it a typescript/es6 syntax?

storage-node/src/services/webApi/controllers/stateApi.ts
```diff
@@ -158,7 +159,7 @@ export async function getStatus(req: express.Request, res: express.Response<Stat
     downloadBuckets,
     sync,
     cleanup,
-    queryNodeStatus: await getQueryNodeStatus(qnApi),
+    queryNodeStatus: await getQueryNodeStatus(api, qnApi),
```
mnaamani, line 162: how does express handle the response if we get an uncaught exception here? Is it handled by https://github.com/Joystream/joystream/blob/691e35f6b846fb07637fefe9372e4a49ff0543b5/storage-node/src/services/webApi/app.ts#L109 ?

mnaamani (changes requested):
Great work. I left a few comments.

I see the annoying test `curator moderation actions: Failed` failing often. I think this started happening more frequently when we introduced the accept pending service because of delay between object being uploaded and the accepted status being updated?

docker-compose.storage-squid.yml
mnaamani: Answering myself: Yes :)

storage-node/src/services/webApi/controllers/stateApi.ts
mnaamani: I guess this is a similar point made in https://github.com/Joystream/joystream/pull/5001#pullrequestreview-1798426461

tests/network-tests/start-storage.sh
zeeshanakram3: Reverted the change. I actually didn't notice that the change happened, as I have installed shell formatter extension in VSCode which formatted the line this way on save.

tests/network-tests/start-storage.sh
zeeshanakram3: Please see https://github.com/Joystream/joystream/pull/5001#discussion_r1442676343

tests/network-tests/start-storage.sh
zeeshanakram3: Addressed in [8c03814](https://github.com/Joystream/joystream/pull/5001/commits/8c03814027f6a270e19e5930cbc55fa477c1d823)

.env
zeeshanakram3: Addressed in [36212f0](https://github.com/Joystream/joystream/pull/5001/commits/36212f0242f1ac362f3d251b23beef20a5cdc3cc)

distributor-node/src/services/networking/query-node/api.ts
zeeshanakram3: It's being imported from Apollo 
https://github.com/zeeshanakram3/joystream/blob/36212f0242f1ac362f3d251b23beef20a5cdc3cc/distributor-node/src/services/networking/query-node/api.ts#L1

distributor-node/src/services/networking/NetworkingService.ts
zeeshanakram3: Reverted the re-ordering change

distributor-node/src/services/networking/NetworkingService.ts
zeeshanakram3: > we should ensure the request handler catches it and return a HTTP 503 status rather than cause the express server to cause the process to exit with an unhandled expection.

I don't think this is currently happening for `distributor-node` (yes, `storage-node` is crashing), if `getQueryNodeStatus()` throws, then it is being caught by the express [error handler](https://github.com/zeeshanakram3/joystream/blob/master/distributor-node/src/services/httpApi/HttpApiBase.ts#L92) middleware.

storage-node/src/services/webApi/controllers/stateApi.ts
zeeshanakram3: Actually, error handler middleware is not really used/called in the `storage-node` (even in all the previous versions), so whenever the unhandled exception occurred in the route handlers, it caused the app to crash. 

If async route handlers throw then its the [responsibility](https://expressjs.com/en/guide/error-handling.html#:~:text=For%20errors%20returned%20from%20asynchronous%20functions%20invoked%20by%20route%20handlers%20and%20middleware%2C%20you%20must%20pass%20them%20to%20the%20next()%20function%2C%20where%20Express%20will%20catch%20and%20process%20them.%20For%20example%3A) of the app to pass this error to the `next()` function which then will be handled by the default error handler

I have [fixed](https://github.com/Joystream/joystream/pull/5001/commits/7d5d37fd25a955ea06adcda5779acd22361e2c43) this for storage-node, for `distributor-node` this is not a problem as error thrown from the route handlers are already passed to the [next()](https://github.com/zeeshanakram3/joystream/blob/master/distributor-node/src/services/httpApi/HttpApiBase.ts#L30) function

distributor-node/config.yml
zeeshanakram3: > I don't mind the name so much. Just wondering where the port number selection came from, is it just 4250 (orion graphql port) + 2 ?

@mnaamani yes.

distributor-node/package.json
zeeshanakram3: I have created a new shell script that will only write to output file if the returned schema response is not empty. Let me know if it's fine.

Added in 3a6b0e4

zeeshanakram3:
> I expected the `supportedObjects` to give proper number. I can get objects fine

@kdembler Addressed in [2bbf0af](https://github.com/Joystream/joystream/pull/5001/commits/2bbf0af1b1bc924908794f38f42263c6806c8cc2). You should be able to see the supported objects count now

distributor-node/config.yml
kdembler: > I don't mind the name so much.

It's not purely cosmetic, I'm just thinking about operators upgrading and possibly forgetting to update this config value. If we leave the current name and somebody keeps the old value, then they will try to query old QN, which will most likely lead to some weird GraphQL errors that may not be as readable and clear.

If we update the config key, any operator that forgot to update their config will get a clear error about missing config key.

distributor-node/src/services/networking/NetworkingService.ts
kdembler: > if getQueryNodeStatus() throws, then it is being caught by the express error handler middleware.

Was this added recently? Referencing my previous comment:

> Getting {"type":"exception","message":"Unexpected end of JSON input"} from /status when squid is not available

kdembler: Hmm actually it is handled and Argus doesn't crash, so I guess we just need better error message in that case

kdembler:
btw @zeeshanakram3 did you manage to resolve https://github.com/Joystream/storage-squid/issues/1? My squid crashed 1 day after starting

distributor-node/src/services/networking/NetworkingService.ts
```diff
@@ -437,16 +447,16 @@ export class NetworkingService {
   }
 
   async getQueryNodeStatus(): Promise<StatusResponse['queryNodeStatus']> {
-    const qnState = await this.queryNodeApi.getQueryNodeState()
+    const squidStatus = await this.queryNodeApi.getQueryNodeState()
```
kdembler, line 448: I think we should try-catch this? Relating to https://github.com/Joystream/joystream/pull/5001/files#r1443286342

distributor-node/src/services/networking/NetworkingService.ts
```diff
@@ -437,16 +447,16 @@ export class NetworkingService {
...
-      this.logger.error("Couldn't fetch the state from connected query-node")
+    if (squidStatus === null) {
+      this.logger.error("Couldn't fetch the state from connected storage-squid")
     }
 
     return {
       url: this.config.endpoints.queryNode,
-      chainHead: qnState?.chainHead || 0,
-      blocksProcessed: qnState?.lastCompleteBlock || 0,
+      chainHead: (await this.runtimeApi.derive.chain.bestNumber()).toNumber() || 0,
```
kdembler, line 456: Possibly worth try-catching as well so if `runtimeApi` is unavailable we get a clear error message?

storage-node/src/commands/server.ts
```diff
@@ -91,8 +91,8 @@ export default class Server extends ApiCommandBase {
     queryNodeEndpoint: flags.string({
```
kdembler: Following https://github.com/Joystream/joystream/pull/5001/files#r1443274494 I would recommend to update the name here as well. This would also apply to other command that use the same flag

storage-node/src/commands/server.ts
```diff
@@ -91,8 +91,8 @@ export default class Server extends ApiCommandBase {
...
       char: 'q',
       required: true,
-      default: 'http://localhost:8081/graphql',
-      description: 'Query node endpoint (e.g.: http://some.com:8081/graphql)',
+      default: 'http://localhost:4352/graphql',
```
kdembler, line 94: Actually if the flag is required does default value make any difference?

storage-node/src/services/queryNode/api.ts
```diff
@@ -285,11 +217,11 @@ export class QueryNodeApi {
    * @param bucketIds - query filter: bucket IDs
    */
   public async getStorageBagsDetails(bucketIds: string[]): Promise<Array<StorageBagDetailsFragment>> {
-    const result = await this.multipleEntitiesWithPagination<
-      StorageBagDetailsFragment,
-      GetBagConnectionQuery,
-      GetBagConnectionQueryVariables
-    >(GetBagConnection, { limit: MAX_RESULTS_PER_QUERY, bucketIds }, 'storageBagsConnection')
+    const result = await this.multipleEntitiesQuery<GetStorageBagDetailsQuery, GetStorageBagDetailsQueryVariables>(
```
kdembler, line 223: Why are we moving away from paginated query here?

storage-node/src/services/queryNode/api.ts
```diff
@@ -303,22 +235,18 @@ export class QueryNodeApi {
...
         ...fullResult,
-        ...(await this.multipleEntitiesWithPagination<
-          DataObjectByBagIdsDetailsFragment,
-          GetDataObjectsByBagIdsConnectionQuery,
-          GetDataObjectsByBagIdsConnectionQueryVariables
-        >(
-          GetDataObjectsByBagIdsConnection,
-          { limit: MAX_RESULTS_PER_QUERY, bagIds: input },
-          'storageDataObjectsConnection'
+        ...(await this.multipleEntitiesQuery<GetDataObjectsByBagIdsQuery, GetDataObjectsByBagIdsQueryVariables>(
```
kdembler, line 249: Same question here

kdembler (commented):
Other than those small things, LGTM. Fantastic work @zeeshanakram3. Once the issue with squid crashing is resolved, I will run some manual tests with this version

distributor-node/config.yml
mnaamani: > If we update the config key, any operator that forgot to update their config will get a clear error about missing config key.

I think that is a good argument actually.

distributor-node/src/services/networking/query-node/api.ts
mnaamani: :+1: thanks I guess it was hard to spot because of so many occurrences of the `from` keyword.

storage-node/src/services/queryNode/api.ts
zeeshanakram3: I think the pagination was mainly used due to the limitation of the QN, and storage squid does not have this limitation and returns the result successfully. In fact, having pagination adds overhead, both network (multiple round-trips) as well as DB, since db has to execute the same query each time with a different offset. 

Also, we need to rethink these expensive queries anyway, either using generators (as Mokhtar suggested), or some tweaks which does not involve fetching all the assets/bags each time.

storage-node/src/services/queryNode/api.ts
zeeshanakram3: Please see https://github.com/Joystream/joystream/pull/5001#discussion_r1444623747

zeeshanakram3:
> btw @zeeshanakram3 did you manage to resolve [Joystream/storage-squid#1](https://github.com/Joystream/storage-squid/issues/1)? My squid crashed 1 day after starting

I wasn't able to reproduce the error. I will try to test again locally against the mainnet

storage-node/src/commands/server.ts
zeeshanakram3: > Actually if the flag is required does default value make any difference?

If you don't provide the value, then default value will be used, otherwise (if default value was not set) the command will throw an error that required flag is missing

storage-node/src/commands/server.ts
zeeshanakram3: Addressed in [e2ce216](https://github.com/Joystream/joystream/pull/5001/commits/e2ce21622463a09f62c639ab6eee332d0678ecfd)

storage-node/src/services/webApi/controllers/stateApi.ts
zeeshanakram3: Addressed in [e2ce216](https://github.com/Joystream/joystream/pull/5001/commits/e2ce21622463a09f62c639ab6eee332d0678ecfd)

kdembler:
> I wasn't able to reproduce the error. I will try to test again locally against the mainnet

I'd suggest to do it on a server so that it indexes all the time. Locally (only syncing when my laptop was on) it didn't happen to me as well

kdembler:
> I'd suggest to do it on a server so that it indexes all the time. Locally (only syncing when my laptop was on) it didn't happen to me as well

After your comment I have dropped my squid, started fresh sync and it crashed again

kdembler:
@zeeshanakram3 one more question: do you think it would be a lot of effort to include storage squid version in argus&colossus status endpoints? https://github.com/Joystream/joystream/issues/4732

zeeshanakram3:
> @zeeshanakram3 one more question: do you think it would be a lot of effort to include storage squid version in argus&colossus status endpoints? #4732

I will implement it.

storage-node/src/services/queryNode/api.ts
mnaamani: Just a comment on generators. I was experimenting here https://github.com/Joystream/joystream/pull/5026/files#diff-0da8a15d4ae4a081c01cccf049103d2c94906628d9d0d57b715c991d3f2596bd with using generator to yield a page at a time to avoid having to concat results into very large arrays.

Until we persist some state there is always going to be this need when colossus starts up to refetch "everything".

And as Zeeshan is suggesting, we both recognize that it make more sense after this initial fetch everything, subsequent sync runs only need to look at the "diff", what new objects have been created since last sync run (in addition to what needs to be re-tried from last run)

docker-compose.storage-squid.yml
```diff
@@ -0,0 +1,79 @@
...
+    hostname: squid-db
+    image: postgres:14
+    restart: unless-stopped
+    environment:
+      POSTGRES_DB: ${SQUID_DB_NAME}
+      POSTGRES_PASSWORD: ${SQUID_DB_PASS}
+    networks:
+      - joystream
+    ports:
+      - '${SQUID_DB_PORT}:${SQUID_DB_PORT}'
```
mnaamani:
```suggestion
      - '127.0.0.1:${SQUID_DB_PORT}:${SQUID_DB_PORT}'
```

To prevent the db being exposed on public interface?

docker-compose.storage-squid.yml
```diff
@@ -0,0 +1,79 @@
...
+    command: ['postgres', '-c', 'config_file=/etc/postgresql/postgresql.conf', '-p', '${SQUID_DB_PORT}']
+    shm_size: 1g
```
mnaamani: Should this be a bit more than what postgres tries to allocate as configured in postgresql.conf?

docker-compose.storage-squid.yml
```diff
@@ -0,0 +1,79 @@
+version: '3'
+
+services:
+  squid_db:
+    container_name: squid_db
+    hostname: squid-db
+    image: postgres:14
```
mnaamani: Does squid work with `postgres:16` ? Is there any downside of using latest if its compatible?

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -88,29 +88,25 @@ export class AcceptPendingObjectsService {
...
+    if (maybeDeletedObjectIds.length) {
+      logger.debug(`Found ${maybeDeletedObjectIds.length} stale pending objects in pending folder`)
+
+      // Check if the stale objects have been ACTUALLY deleted using the deleted events from Storage Squid
+      const deletedObjectIds = await this.qnApi.getDataObjectDeletedEvents(maybeDeletedObjectIds)
+
+      // Delete the stale objects from the pending folder
+      await Promise.allSettled(
+        deletedObjectIds.map(({ data: { dataObjectId } }) =>
+          fsPromises.unlink(path.join(this.pendingDataObjectsDir, dataObjectId))
```
mnaamani, line 107: should also remove the id from `pendingDataObjects` array so we don't attempt to process it in next step.

storage-node/src/services/webApi/app.ts
```diff
@@ -82,7 +79,8 @@ export async function createApp(config: AppConfig): Promise<Express> {
       validateRequests: true,
       operationHandlers: {
         basePath: path.join(__dirname, './controllers'),
-        resolver: OpenApiValidator.resolvers.modulePathResolver,
+        resolver: (basePath: string, route: RouteMetadata, apiDoc: OpenAPIV3.Document) =>
+          asyncHandler(OpenApiValidator.resolvers.modulePathResolver(basePath, route, apiDoc)),
```
mnaamani, line 83: What is this change fixing?

storage-node/src/services/webApi/app.ts
```diff
@@ -125,109 +121,11 @@ export async function createApp(config: AppConfig): Promise<Express> {
         message: err.message,
       })
     }
-
-    next()
```
mnaamani, line 129: yep that looked out of place.

storage-node/src/services/webApi/app.ts
```diff
@@ -125,109 +121,11 @@ export async function createApp(config: AppConfig): Promise<Express> {
...
   })
 
   return app
 }
 
-/**
```
mnaamani, line 135: Since we are getting rid of these.. we can also drop

```ts
    // when enabling upload auth ensure the keyring has the operator role key and set it here.
    const enableUploadingAuth = false
    const operatorRoleKey = undefined
```

in `server.ts` and the corresponding fields in the `AppConfig` type

Does it make sense then to also remove the relevant parts in the api spec `openapi.yaml`?

mnaamani (changes requested):
Some more code to drop (upload authentication related)
A couple suggestions on docker-compose config for squid.

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: If an object has been deleted from the runtime then, `pendingDataObjects` wont contain it in the first place since QN wont return it. Wouldn't that be the case?

storage-node/src/services/webApi/app.ts
zeeshanakram3: It catches the errors thrown by the async route handlers and passes it to the default [error handler](https://github.com/zeeshanakram3/joystream/blob/0b8c3f8bdd6539a408bfcf6d9d4539a53cbb7649/storage-node/src/services/webApi/app.ts#L105). Without this change if any of the route handlers (e.g. handler for `/status`) threw an error then it would crash the application since it would not be caught anywhere.

It's the equivalent of following piece of code in distributor node
https://github.com/zeeshanakram3/joystream/blob/0b8c3f8bdd6539a408bfcf6d9d4539a53cbb7649/distributor-node/src/services/httpApi/HttpApiBase.ts#L27-L31

docker-compose.storage-squid.yml
zeeshanakram3: Yeah, should be safe upgrading to `postgres:16`, so changed it.

docker-compose.storage-squid.yml
zeeshanakram3: Addressed.

docker-compose.storage-squid.yml
zeeshanakram3: Addressed.

storage-node/src/services/webApi/app.ts
zeeshanakram3: Done in [fd44d0d](https://github.com/Joystream/joystream/pull/5001/commits/fd44d0d3c38c0b28b4785694aed766920aca9797). Also removed relevant parts from the API spec file.

docker-compose.yml
```diff
@@ -237,7 +237,7 @@ services:
       "
 
   indexer:
-    image: joystream/hydra-indexer:v5.0.0-alpha.1
+    image: hydra-indexer
```
mnaamani: I think like me you were using local build of hydra-indexer ;)

zeeshanakram3: Yeah ;), fixed that in https://github.com/Joystream/joystream/pull/5001/commits/190daa0cb75ed5ec88e8c7b7a470daa125cab15b

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: Correct, but that original array i'm refering to includes files picked up from the pending folder. So here we are deleting when we find that it has been deleted by runtime. Now it is odd for that file to get placed there but non-the less if it was placed, or somehow uploaded again, then we will delete it but later still try to actually moving to uploads folder because no matching bucket will be found for it.

mnaamani: we might actually try to compute the ipfs hash of it first.. but still the general point applies. correct?

mnaamani: Sorry my bad, I confused it with the `pendingIds` array.

docker-compose.storage-squid.yml
```diff
@@ -0,0 +1,79 @@
...
+      - DB_NAME=${SQUID_DB_NAME}
+      - DB_PASS=${SQUID_DB_PASS}
+      - DB_PORT=${SQUID_DB_PORT}
+      - DB_HOST=${SQUID_DB_HOST}
+      - GQL_PORT=${SQUID_GQL_PORT}
+      - ARCHIVE_GATEWAY_URL=${SQUID_ARCHIVE_GATEWAY_URL}
+      - RPC_ENDPOINT=${JOYSTREAM_NODE_WS}
+
+    ports:
+      - '127.0.0.1:${PROCESSOR_PROMETHEUS_PORT}:${PROCESSOR_PROMETHEUS_PORT}'
```
mnaamani: change to use `SQUID_PROCESSOR_PROMETHEUS_PORT` instead. Otherwise services are failing to start with:

```
Error response from daemon: driver failed programming external connectivity on endpoint squid_processor (e1a9bb40428326391a37414aa61f52dfd4817f99dc1662c278013f4780d7c979): Bind for 127.0.0.1:3337 failed: port is already allocated
```

When we start network with `start.sh` or `start-multistorage.sh` that starts orion services first which is listening on `PROCESSOR_PROMETHEUS_PORT` port.

mnaamani (changes requested):
Everything looks good, final changes 
- fix the prometheus port conflict in docker compose file.
- Bump storage node version to v3.11.0 and the storage cli package to v3.3.0

docker-compose.storage-squid.yml
zeeshanakram3: Done in [897c7c0](https://github.com/Joystream/joystream/pull/5001/commits/897c7c0e38110a5fcbadd8cc20a3add82642215c)

zeeshanakram3:
> * Bump storage node version to v3.11.0 and the storage cli package to v3.3.0

@mnaamani  regarding the package versions, shouldn't we bump the major versions in both Argus & Colossus, since we have a breaking change in configuration options (changed `--queryNodeEndpoint` to `--storageSquidEndpoint`)?

mnaamani:
> > * Bump storage node version to v3.11.0 and the storage cli package to v3.3.0
> 
> @mnaamani regarding the package versions, shouldn't we bump the major versions in both Argus & Colossus, since we have a breaking change in configuration options (changed `--queryNodeEndpoint` to `--storageSquidEndpoint`)?

That might actually make more sense yes.

mnaamani:
Should we also consider forking current master to keep the old version on a separate branch so we can slowly transition all operators just in-case we need to create some important patch while operators are still on older version?

zeeshanakram3:
> Should we also consider forking current master to keep the old version on a separate branch so we can slowly transition all operators just in-case we need to create some important patch while operators are still on older version?

Yeah, makes sense

zeeshanakram3:
Failing integration [tests](https://github.com/Joystream/joystream/actions/runs/7555614511/job/20571202057?pr=5001#step:8:335) should be fixed by https://github.com/Joystream/storage-squid/pull/13

mnaamani (approved):

kdembler:
@zeeshanakram3 could you reply to this? https://github.com/Joystream/joystream/pull/5001#discussion_r1443286342 I think Argus (and Colossus) should still provide valid response for /status request even when the squid is down. Currently we're just forwarding the raw error to the user.

As a counterargument, we need that kind of handling for asset requests as well, not only for status and that currently isn't handled

kdembler (approved):
Created followup for the thing I mentioned: https://github.com/Joystream/joystream/issues/5056

Great work on this @zeeshanakram3, LGTM!

Issue #5008, opened 2023-12-19
[Colossus] Avoid fetching list of all data objects (i.e. calling `api/v1/state/data-objects`) when performing sync


zeeshanakram3:
# Problem 

While syncing the data object/s, the storage node needs to know the existence of required object/s from peer nodes and then pick a URL to download object/s from. However, the problem is that for each asset the node needs to sync, it calls `api/v1/state/data-objects` on all the peer nodes until it picks a URL to download the asset from.

https://github.com/Joystream/joystream/blob/46e75506e9639dae4bf67a8ff7e322166ad522ee/storage-node/src/services/sync/tasks.ts#L220-L225

`/state/data-objects` does not return a constant size response, and hence response size and latency grow linearly, for reference, currently some nodes return the data objects response over 5MB in size. 

# Solution

1. Maybe use `HEAD /files/{Id}` to know the availability of assets on a given node, OR
2. Or create a new lightweight endpoint, that maybe return a boolean response indicating the presence of asset.

yasiryagi:
Can we call /state/data-objects once per sync timer and cache it. That will reduce the overhead and could more efficient than head per object.

zeeshanakram3:
> Can we call /state/data-objects once per sync timer and cache it. That will reduce the overhead and could more efficient than head per object.

Yeah, I guess that should work too, but instead of getting all the data objects per sync (`/state/data-objects`), we should only get the data-objects of bags that we need to sync (`/state/bags/{bagId}/data-objects`) and then cache it.

mnaamani:
> Can we call /state/data-objects once per sync timer and cache it. That will reduce the overhead and could more efficient than head per object.

Currently it does cache the result, but for a short period [3min](https://github.com/Joystream/joystream/blob/10baf39c222ce42a7fc7ab70518835b00f67bdd8/storage-node/src/services/sync/remoteStorageData.ts#L7). So if the sync interval is larger than this caching period, which I think  the operators are setting to at minimum 10min, the data is always fetched again.

mnaamani:
Suggested solution on zoom call:

No need to pre-determine if an operator has object before attempting to fetch it.
Just do best effort to fetch from other operators that should be storing the same bag which the object belongs in.

kdembler:
@mnaamani @zeeshanakram3 can we close this after sync rework?

PR #5014, opened 2023-12-26
[Colossus] Sync rework - 

zeeshanakram3:
colses #5008

This PR:
- removes `api` parameter from the `performSync` function & fetch the object hash from QN instead of fetching from the runtime.
- directly download data objects from remote nodes instead of pre-checking their existence first.
- The logging message in the `superagent` error handler - when the object does not exist in the remote node - won't be logged, since technically missing asset won't be categorized as an error, and logging this as an error may confuse the Operator/Lead


storage-node/src/services/sync/tasks.ts
```diff
@@ -77,39 +72,62 @@ export class DeleteLocalFileTask implements SyncTask {
...
   description(): string {
-    return `Sync - downloading file: ${this.url} to ${this.uploadsDirectory} ....`
+    return `Sync - Trying for download of: ${this.dataObjectId} ....`
   }
 
   async execute(): Promise<void> {
+    // Create an array of operator URL indices to maintain a random URL choice
+    // cannot use the original array because we shouldn't modify the original data.
+    // And cloning it seems like a heavy operation.
+    const operatorUrlIndices: number[] = [...Array(this.operatorUrls.length).keys()]
```
mnaamani: I can suggest a slightly cleaner approach. Start with an array already containing random order of the indices using a combination of _.range() and _.shuffle(). You could then just shift or pop the array at the start of the loop. Or the while loop then can be a simply a `for..each` iteration over the array, without needing to modify it.

`const operatorUrlIndices: number[] = _.shuffle(_.range(this.operatorUrls.length))`

mnaamani (changes requested):
Looks great. Small code style change suggested.

storage-node/src/services/sync/tasks.ts
```diff
@@ -77,39 +72,53 @@ export class DeleteLocalFileTask implements SyncTask {
...
 
   async execute(): Promise<void> {
+    const operatorUrlIndices: number[] = _.shuffle(_.range(this.operatorUrls.length))
+
+    for (const randomUrlIndex of operatorUrlIndices) {
+      const chosenBaseUrl = this.operatorUrls[randomUrlIndex]
+      logger.debug(`Sync - random storage node URL was chosen ${chosenBaseUrl}`)
+
+      // Remove random url from the original list.
+      _.remove(operatorUrlIndices, (index) => index === randomUrlIndex)
```
mnaamani: removing the element is no longer necessary, even potentially problematic.

zeeshanakram3: Ah right. Missed that.

zeeshanakram3: Addressed in [3c02240](https://github.com/Joystream/joystream/pull/5014/commits/3c02240ef9fcf9a64945e94c880f33a16849de16)

mnaamani (approved):

Issue #5003, opened 2023-12-19
[Colossus]  storage node crashed on upload


yasiryagi:
Description: Storage node unhandled crash during an object upload, while it is accepted
```
{
  "data": {
    "storageDataObjects": [
      {
        "id": "1454542",
        "size": "20022",
        "isAccepted": true,
        "videoMedia": null,
        "videoThumbnail": {
          "id": "709318"
        }
      }
    ]
  }
}
```
Product: Colossus 
Version: 3.7.2
Log:

```
{
  "_index": "storage-node-adovrn",
  "_id": "sATJeIwBh1Mrve8iRUku",
  "_version": 1,
  "_score": 0,
  "_ignored": [
    "fields.meta.message.keyword",
    "fields.meta.error.stack.keyword",
    "fields.meta.stack.keyword"
  ],
  "_source": {
    "@timestamp": "2023-12-17T17:18:23.213Z",
    "message": "POST /api/v1/files: Error 500: Timeout",
    "severity": "error",
    "fields": {
      "meta": {
        "error": {
          "stack": "Error\n    at new module.exports.TimeoutError (/joystream/node_modules/promise-timeout/index.js:44:16)\n    at module.exports.timeout (/joystream/node_modules/promise-timeout/index.js:20:15)\n    at extrinsicWrapper (/joystream/storage-node/lib/services/runtime/extrinsics.js:207:45)\n    at acceptPendingDataObjects (/joystream/storage-node/lib/services/runtime/extrinsics.js:115:18)\n    at uploadFile (/joystream/storage-node/lib/services/webApi/controllers/filesApi.js:99:57)",
          "message": "Timeout"
        },
        "level": "error",
        "message": "uncaughtException: Timeout\nError\n    at new module.exports.TimeoutError (/joystream/node_modules/promise-timeout/index.js:44:16)\n    at module.exports.timeout (/joystream/node_modules/promise-timeout/index.js:20:15)\n    at extrinsicWrapper (/joystream/storage-node/lib/services/runtime/extrinsics.js:207:45)\n    at acceptPendingDataObjects (/joystream/storage-node/lib/services/runtime/extrinsics.js:115:18)\n    at uploadFile (/joystream/storage-node/lib/services/webApi/controllers/filesApi.js:99:57)",
[...]

Issue #5002, opened 2023-12-19
[Colossus] storage failing to query data objects


yasiryagi:
Issue: when query data objects intermittently get error response
detail:
```

~ # curl https://storage.freakstatic.com/storage/api/v1/state/data-objects
{"type":"request_exception","message":".response[565943] should be string"}
joystream ~ # curl https://joystream.joystream-arsi44.space/storage/api/v1/state/data-objects
{"type":"request_exception","message":".response[404642] should be string"}
```

Product: colossus
Version: 3.9.0, 3.9.1

freakstatic:
This seems to be happening on >= v3.9.0
I tested on 3.8.1 and it's not happening

freakstatic:
v3.9.1 URL: 
https://joystream.joystream-arsi44.space/storage/api/v1/state/data-objects

mnaamani:
This error is coming from response validation failing, we use Ajv package - the schema for the response is defined in:

https://github.com/Joystream/joystream/blob/10baf39c222ce42a7fc7ab70518835b00f67bdd8/storage-node/src/api-spec/openapi.yaml#L361

Somehow, the map is having a key set that is a number, not an string. Not sure how that is possible at least at compile time since the map is strongly typed. 
https://github.com/Joystream/joystream/blob/10baf39c222ce42a7fc7ab70518835b00f67bdd8/storage-node/src/services/caching/localDataObjects.ts#L11

I inspected all code paths that update the map and couldn't identify where we might be mistakenly setting a number.

Only thing I can think of is if somehow the spread operator used to generate the array: `const ids = [...idCache.keys()]` could be doing something funny on the node that has this issue? 

Not sure if we wan't to investigate this further if we plan on removing this api endpoint see: https://github.com/Joystream/joystream/issues/5008 ?

mnaamani:
Finally managed to find the fault.
When HTTP GET or HEAD request is made to the node which does not have the object id, we hit these lines of code:

https://github.com/Joystream/joystream/blob/c4789f9bf556eee32e03853554c124631d34845c/storage-node/src/services/webApi/controllers/filesApi.ts#L65

https://github.com/Joystream/joystream/blob/c4789f9bf556eee32e03853554c124631d34845c/storage-node/src/services/webApi/controllers/filesApi.ts#L93

[...]

Issue #4991, opened 2023-12-06
Colossus: In memory local data objects cache improvements


mnaamani:
### Background

The node keeps track of the local data objects in storage in a `Map`. This is acceptable for now but is not scalable, as it will grow memory usage linearly with the number of objects stored.

https://github.com/Joystream/joystream/blob/5d911ce1bd546da16c983c852203a2a6eac8d8ea/storage-node/src/services/caching/localDataObjects.ts#L21

Some room for improvement.
1. `addDataObjectIdToCache()` should check for existence of entry before adding
2. `loadDataObjectIdCache()` should be protected to only run once
3. `pin/unpinDataObjectIdToCache()` should check existence of entry before pinning/unpinning
4. `getDataObjectIdFromCache()` - lock before check for id (I previously suggested it was not necessary in PR) I was wrong.
5. `getDataObjectIDs()` returns an array. At some point this is going to be expensive and we will hit the JavaScript limit on size of array elements. (32-bit index so ~4Billion elements)

kdembler:
closed by https://github.com/Joystream/joystream/pull/5016

PR #5016, opened 2023-12-28
Colossus bug fixes for 3.10.0 release

mnaamani:
1. Multiple bug fixes
2. Refactored and improved pending objects service
3. Allow re-upload of already "accepted" objects via files api endpoint, in-case they are lost.
4. Fixed issue in batch extrinsic sender which was not returning correct failed calls.

closes:
- https://github.com/Joystream/joystream/issues/4991 - removed async lock, added some assertions
- https://github.com/Joystream/joystream/issues/5002 - fixed bug in files api that was injecting `undefined` key into idCache map
- https://github.com/Joystream/joystream/issues/5003
  - timeout issue sending extrinsic (which caused the next issue) - with use of latest improvements this should no longer happen.
  - failure to re-upload - support now added.
  - basic 404 file note found. Not a problem really, but maybe it shouldn't be logged as an error?


storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -34,14 +38,19 @@ export async function getFile(
...
-    await pinDataObjectIdToCache(req.params.id)
+  const { id } = req.params
+  const dataObjectId = new BN(id).toString()
 
-    const dataObjectId = new BN(req.params.id)
-    const uploadsDir = res.locals.uploadsDir
+  if (!(await getDataObjectIdFromCache(dataObjectId))) {
+    sendResponseWithError(res, next, new WebApiError('File Not Found', 404), 'files')
+    return
+  }
```
zeeshanakram3: Wouldn't this be redundant, since now `pinDataObjectIdToCache` ensures that ID exists before pinning it? If this is added to return a better error message in the response, then we can provide this error in as an argument of `asset` in `pinDataObjectIdToCache` function itself. e.g.

```ts
  assert(idCache.has(dataObjectId), new WebApiError('File Not Found', 404))

```

zeeshanakram3 (commented):
Thanks for investigating and fixing the issue. Wouldn't it be much simpler to store data object ID (`req.params.id`) at the start of the both `GET`/`HEAD` handlers, before the `try` block starts.  This way the data object ID in the `finally` block would still be valid and we won't have to separately do the unpinning at multiple different places. For example, following change fixes the bug

image

storage-node/src/services/webApi/controllers/filesApi.ts
zeeshanakram3: We can also pass the `Error` as a second optional argument to the `pinDataObjectIdToCache` function

mnaamani:
> Thanks for investigating and fixing the issue. Wouldn't it be much simpler to store data object ID (`req.params.id`) at the start of the both `GET`/`HEAD` handlers, before the `try` block starts. This way the data object ID in the `finally` block would still be valid and we won't have to separately do the unpinning at multiple different places.

Yes that was one of the reasons I moved the data object id instantiation outside of the `try` block, but then realized that I had to get rid of the `finally` clause because it can be reached before the stream ends. So we would be unpinning too early, defeating the whole point.

It is easy to miss such cases when mixing `async/await` code with traditional node event and steam based paradigm in the same function.

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: Remember assertions don't throw and it is intended to fail the process completely. Meaning the caller needs to make sure they are calling at the right time. So if these new assertions fail it means our code is not correct.

Are you suggesting we should be returning/throwing an error/exception instead of these assertions?

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -68,12 +71,17 @@ export async function addDataObjectIdToCache(dataObjectId: string): Promise<void
  *
  * @param dataObjectId - Storage data object ID
  *
- * @returns empty promise.
+ * @returns promise<boolean> - true if object exists and was pinned
```
mnaamani: I forgot to revert this after experimenting with returning boolean :)

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: I picked assertions because I was treating the user of the idCache as "our own" code. So its not like a library.
With reference to https://softwareengineering.stackexchange.com/questions/15515/when-to-use-assertions-and-when-to-use-exceptions#15518 

Let me know what you think.

zeeshanakram3: > Remember assertions don't throw and it is intended to fail the process completely.

Maybe I am misunderstanding, but In javascript, the `assert` module throws the Error of type [AssertionError](https://nodejs.org/api/assert.html#class-assertassertionerror), and we can actually pass a custom error too, that should be thrown in case assertion fails.

mnaamani: You are absolutely correct. And given that is true I think I might be introducing more problems.
Let me think this through and come back to it.

storage-node/src/services/webApi/app.ts
```diff
@@ -250,12 +261,27 @@ async function validateUploadFileParams(req: express.Request, res: express.Respo
     throw new WebApiError(`Data object ${dataObjectId} doesn't exist in storage bag ${parsedBagId}`, 400)
   }
 
+  // Skipping this check to allow re-upload of "lost" objects
   if (dataObject.accepted.valueOf()) {
-    throw new WebApiError(`Data object ${dataObjectId} has already been accepted by storage node`, 400)
+    // throw new WebApiError(`Data object ${dataObjectId} has already been accepted by storage node`, 400)
```
zeeshanakram3: Please remove this commented code.

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -44,84 +37,83 @@ export async function loadDataObjectIdCache(uploadDir: string): Promise<void> {
...
-export async function addDataObjectIdToCache(dataObjectId: string): Promise<void> {
-  await lock.acquireAsync()
-
-  idCache.set(dataObjectId, 0)
-
-  lock.release()
+export function addDataObjectIdToCache(dataObjectId: string): void {
+  assert(typeof dataObjectId === 'string')
+  // assert(!idCache.has(dataObjectId))
+  // existence check to avoid resetting pin count
```
zeeshanakram3: Isn't it better to have this assertion?

storage-node/src/commands/server.ts
```diff
@@ -237,27 +241,40 @@ Supported values: warn, error, debug, info. Default:debug`,
       )
     }
 
+    if (!flags.pendingFolder) {
+      logger.warn(
+        'You did not specify a path to the pending directory. ' +
+          'A pending folder under the uploads folder willl be used. ' +
```
```suggestion
          'A pending folder under the uploads folder will be used. ' +
```

storage-node/src/commands/server.ts
```diff
@@ -55,6 +55,10 @@ export default class Server extends ApiCommandBase {
       description:
         'Directory to store tempory files during sync and upload (absolute path).\nIf not specified a subfolder under the uploads directory will be used.',
     }),
+    pendingFolder: flags.string({
+      description:
+        'Directory to store pending files which are uploaded upload (absolute path).\nIf not specified a subfolder under the uploads directory will be used.',
+    }),
```
zeeshanakram3, line 61: I think making the `temp` folder configurable does not create any risks, but in the case of the `pending` folder what would happen if the operator specifies a different `pendingFolder` in different node runs? Would this lead to considering the objects in the previous pending folder as effectively lost (unless they were re-uploaded)

storage-node/src/services/webApi/app.ts
```diff
@@ -73,6 +74,16 @@ export async function createApp(config: AppConfig): Promise<Express> {
...
+    // Pre validate data object id in GET/HEAD file requests
+    (req: express.Request, res: express.Response<unknown, AppConfig>, next: NextFunction) => {
+      if (req.path === '/api/v1/files' && (req.method === 'GET' || req.method === 'HEAD')) {
+        validateDataObjectId(req, res)
+          .then(next)
+          .catch((error) => sendResponseWithError(res, next, error, 'upload'))
+      } else {
+        next()
+      }
+    },
```
zeeshanakram3: This middleware handler would never be invoked since `req.path` would always be `/api/v1/files/{id}`, i.e. it would include the data object ID as part of the path.

Also, it's it better to do this validation at the openApi schema level (e.g. [example](https://github.com/mnaamani/joystream/blob/0001eef6b11aac812895bbb3f234d890d2117ea9/storage-node/src/api-spec/openapi.yaml#L109))?

storage-node/src/services/caching/localDataObjects.ts
```diff
@@ -44,84 +37,83 @@ export async function loadDataObjectIdCache(uploadDir: string): Promise<void> {
...
  * @param dataObjectId - Storage data object ID
+ *
+ * @returns void
  */
-export async function deleteDataObjectIdFromCache(dataObjectId: string): Promise<void> {
-  await lock.acquireAsync()
-
+export function deleteDataObjectIdFromCache(dataObjectId: string): void {
+  assert(typeof dataObjectId === 'string')
+  // assert(idCache.has(dataObjectId))
```
zeeshanakram3: Isn't it better to have this assertion?

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -34,35 +34,45 @@ export async function getFile(
...
+      // we assume that if stream.pipe throws, then stream.on('end') will never fire
+      unpinDataObjectIdFromCache(dataObjectId)
+    })
+
     stream.pipe(res)
   } catch (err) {
     sendResponseWithError(res, next, err, 'files')
-  } finally {
-    await unpinDataObjectIdFromCache(req.params.id)
+    // we assume that if stream.pipe throws, then stream.on('end') will never fire
```
zeeshanakram3: I think the comment should not be here?

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -54,61 +59,87 @@ export class AcceptPendingObjectsService {
...
+          const objectsToAccept = await this.processPendingObjects(ids)
+          await this.acceptPendingDataObjects(api, workerId, objectsToAccept, maxTxBatchSize)
+          await new Promise((resolve) => setTimeout(resolve, intervalMs))
+        } catch (err) {
+          logger.error(`Error processing pending objects: ${err}`)
+        }
+      }
+    })().catch((err) => {
+      logger.error(`Accept pending objects service died: ${err}`)
+    })
```
zeeshanakram3:
```suggestion
    })()
```
Since the async function is in `try...catch`, I don't think any promise rejections will ever be caught in `.catch` function. WDYT?

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -54,61 +59,87 @@ export class AcceptPendingObjectsService {
...
           } else {
-            this.add(dataObject.id, storageBucket.id, dataObject.storageBag.id)
+            objectsToAccept.push([dataObject.id, [storageBucket.id, dataObject.storageBag.id]])
           }
         } else {
-          logger.warn(
+          logger.debug(
             `Data object ${dataObject.id} in pending directory is no longer assigned to any of the upload buckets: ${this.uploadBuckets}.`
           )
+          await fsPromises.unlink(path.join(this.pendingDataObjectsDir, dataObject.id))
```
zeeshanakram3: I think we discussed that doing this is risky, and should probably by handled by pruning service (https://github.com/Joystream/joystream/pull/4971#discussion_r1409373556)

So better to move such objects to the uploads folder, and pruning service can handle from there?

zeeshanakram3 (changes requested):

storage-node/src/commands/server.ts
mnaamani: Correct there is a risk of that happening. The operator can after start node with a new path for pending folder, simply copy the files to the new location and they will be picked up. But I think it is a necessary step towards making the uploads folder separate in preparation for external object storage back-end.

storage-node/src/services/webApi/app.ts
mnaamani: Good catch. Yeah I was surprised actually that the schema was for a string. Maybe it could be of type `number` rather than regex pattern on a string? Was a string perhaps chosen to support data object ids larger than javascript max number ?

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: You are correct, I think I was forced to add a catch handler because typescript complained. I'll try to use the previous code style to make this whole function cleaner. The reason I ended up with this was this was the point where I was debugging the issue which ended up being the invalid spread operator on the extrinsic result.

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: The reason for the comment at both lines where we are invoking the unpin call is to explain that it is not a mistake (based on the assumption), and we should never actually get the result where unpin is invoked twice.
I suppose I could have a locally scoped function that will only even call `unpinDataObjectIdFromCache()` once and log a warning if it is every invoked multiple times as a signal that this assumption is false without causing a possible corrupt state or unpin throwing the assertion error.

storage-node/src/services/webApi/app.ts
mnaamani: Even that regex expression could allow numbers too large for BN to fail.
With this custom function we could also do checks like if the id outside the range of valid dataobject ids?

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: Ah yes now I remember. That is a good idea actually.

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: About this whole approach of pinning to prevent the cleanup service deleting a file being served, I may have overlooked that on linux file systems at least, it may not be necessary at all. I should have considered that when reviewing first PR that introduced the clean up service.

I do think we can safely get rid of this.

On top of that, I can't find the comment in the PR but I do remember at one point we were serving content from pending folder if objects was not yet moved to uploads, and I suggested that it was a bad idea as we didn't know what would happen if files was being moved while being served.

I think we should re-introduce ability to serve from pending folder. It will become more important when we move to object storage, if there is a delay in transferring the pending file from local file system to object storage (compared with what we do currently which is just renaming on the save disk volume).

What do you think?

storage-node/src/services/sync/acceptPendingObjects.ts
```diff
@@ -54,61 +59,81 @@ export class AcceptPendingObjectsService {
...
+      // we could store the bag-id and object id in the filename to make it easy to lookup in runtime state.
     }
 
-    await Promise.all(
+    const objectsToAccept: PendingObjectDetails = []
+
+    await Promise.allSettled(
       pendingDataObjects.map(async (dataObject) => {
-        const storageBucket = dataObject.storageBag.storageBuckets.find(({ id }) => this.uploadBuckets.includes(id))
+        // verify the ipfshash of the object matches against runtime!
```
mnaamani: @zeeshanakram3 what are your thoughts about doing this additional check here?

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -34,35 +34,45 @@ export async function getFile(
...
+  try {
     const uploadsDir = res.locals.uploadsDir
-
-    const fullPath = path.resolve(uploadsDir, dataObjectId.toString())
-
+    const fullPath = path.resolve(uploadsDir, dataObjectId)
     const fileInfo = await getFileInfo(fullPath)
-    const fileStats = await fsPromises.stat(fullPath)
+
+    // verify ipfs has of file we are about to serve, just in-case it is corrupt
```
mnaamani: @zeeshanakram3 similar question to https://github.com/Joystream/joystream/pull/5016/files#r1438660236

zeeshanakram3: I understand the rationale, but I think it would affect the performance at scale, since hashing is CPU-bound work with linear time complexity (`O(n)`), so it could severely impact the throughput of `GET` requests if a large number of concurrent requests are trying to get some object especially if the object size is very large.

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: This check is already being [done](https://github.com/mnaamani/joystream/blob/3af466c526e37c40642f9045153f8377a5393739/storage-node/src/services/webApi/controllers/filesApi.ts#L139) after the file upload completes and before it is moved to the pending `dir`. So I think doing that would be redundant?

storage-node/src/services/webApi/controllers/filesApi.ts
zeeshanakram3: > On top of that, I can't find the comment in the PR but I do remember at one point we were serving content from pending folder if objects was not yet moved to uploads, and I suggested that it was a bad idea as we didn't know what would happen if files was being being while being served.

This is the PR where you suggested and mage the change https://github.com/Joystream/joystream/pull/4989

storage-node/src/services/webApi/app.ts
```diff
@@ -73,6 +74,16 @@ export async function createApp(config: AppConfig): Promise<Express> {
...
+    // Pre validate data object id in GET/HEAD file requests
+    (req: express.Request, res: express.Response<unknown, AppConfig>, next: NextFunction) => {
+      if (req.path.startsWith('/api/v1/files/') && (req.method === 'GET' || req.method === 'HEAD')) {
+        validateDataObjectId(req, res)
+          .then(next)
+          .catch((error) => sendResponseWithError(res, next, error, 'upload'))
+      } else {
+        next()
+      }
+    },
```
zeeshanakram3: Actually, this still won't work correctly as `req.params` inside `validateDataObjectId` is empty object (`{}`). This is because the express does not know about the structure of the URL yet (i.e. whether  `1` in `/api/v1/files/1` is parameter or part of the URL itself), and it will only know about the `params` after the next middleware (`OpenApiValidator.middleware`) is executed.

So unfortunately I think there is no way then doing this validation at openapi schema level

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: Ok, that is a good point.

storage-node/src/services/sync/acceptPendingObjects.ts
mnaamani: It is of course redundant for objects that just came in through the http api, but not for files that might have been copied by operator directly. I'm thinking of scenario where operator maybe recovering their setup, moving files from other location after chaing pending folder path. Maybe somehow recovering "lost" objects from a backup/archive and placing them into the pending folder for processing..

storage-node/src/services/webApi/app.ts
mnaamani: Ah thanks for checking this. 
So I'll drop it, and update the handlers to not assume the id has been validated fully, and at least update the schema validation.

storage-node/src/services/sync/acceptPendingObjects.ts
zeeshanakram3: Yeah, in those cases it makes sense to have this validation. So maybe just add a comment, briefly explaining why are we doing this? In which scenarios it would be beneficial? As you described above

mnaamani: done in [0cdd273](https://github.com/Joystream/joystream/pull/5016/commits/0cdd2732bb3333040b16b714a081aeb46d8a8bc5)

storage-node/src/services/webApi/controllers/filesApi.ts
mnaamani: Made unpinning less error prone and guaranteed to only happen once after pinning in getFile handler.
I read that end event on stream may/or may not fire if error event is fired, so there are potentially 3 places where we would need to unpin.. 

Done in [d5ec0ec](https://github.com/Joystream/joystream/pull/5016/commits/d5ec0ec0d15dc3271972c4c3b0885e124480b30b)

storage-node/src/services/webApi/app.ts
mnaamani: Done in [0cdd273](https://github.com/Joystream/joystream/pull/5016/commits/0cdd2732bb3333040b16b714a081aeb46d8a8bc5)

storage-node/src/services/webApi/controllers/filesApi.ts
zeeshanakram3: > Made unpinning less error prone and guaranteed to only happen once after pinning in getFile handler.

Yeah, this is a good change actually.

zeeshanakram3 (approved):
LGTM.

PR #5021, opened 2024-01-03
[colossus] fix: QN query `Maximum call stack size exceeded`

zeeshanakram3:
SSIA.

image


mnaamani:
I would have expected this error only if `MAX_RESULTS_PER_QUERY` was set to a very large value, but it is a constant 1000. On further inspection I see now this is only the size limit result per page. So the name of the constant was misleading and why I did not make these changes in similar fix https://github.com/Joystream/joystream/pull/4969

mnaamani (approved):

PR #5024, opened 2024-01-03
storage-node: moveFile

mnaamani:
Introduce `moveFile(src, dest)` in place of  `fs.promises.rename(src, dest)`, to avoid `EXDEV: cross-device link not permitted` error when moving a file from `temp` and `pending` folders into the `uploads` folder when they exist on different volumes.


zeeshanakram3 (approved):
LGTM.

PR #5026, opened 2024-01-05
colossus: annotations - looking for performance issues

mnaamani:
Exploring areas in code that have room for improvement in terms of overall performance.

### DIFF
--- a/storage-node/src/services/queryNode/api.ts
+++ b/storage-node/src/services/queryNode/api.ts
@@ -161,14 +161,25 @@ export class QueryNodeApi {
     NodeT,
     QueryT extends { [k: string]: PaginationQueryResult<NodeT> },
     CustomVariablesT extends Record<string, unknown>
-  >(
-    query: DocumentNode,
-    variables: CustomVariablesT,
-    resultKey: keyof QueryT,
-    itemsPerPage = MAX_RESULTS_PER_QUERY
-  ): Promise<NodeT[]> {
+  >(query: DocumentNode, variables: CustomVariablesT, resultKey: keyof QueryT, itemsPerPage = MAX_RESULTS_PER_QUERY) {
+    const results = []
+    for await (const result of this.multipleEntitiesWithPaginationGenerator<NodeT, QueryT, CustomVariablesT>(
+      query,
+      variables,
+      resultKey,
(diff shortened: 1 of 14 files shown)


storage-node/src/services/sync/synchronizer.ts
```diff
@@ -45,17 +45,26 @@ export async function performSync(
   selectedOperatorUrl?: string
 ): Promise<void> {
   logger.info('Started syncing...')
+  logger.info('Sync - Fetching obligations...')
+  // This large query is also done by cleanup service.. sync the two services to use the same promise to get result
+  // one-time if queries are concurrent.
   const [model, files] = await Promise.all([getStorageObligationsFromRuntime(qnApi, buckets), getDataObjectIDs()])
```
mnaamani, line 51: When thinking only about "new" objects to fetch, it makes sense to fetch all obligations on first run.
On subsequent calls we should find an approach to limit the size of the data set.

storage-node/src/services/sync/workingProcess.ts
```diff
@@ -37,7 +37,7 @@ export class WorkingStack implements TaskSink, TaskSource {
   }
 
   async get(): Promise<SyncTask | null> {
-    const task = this.workingStack.pop()
+    const task = this.workingStack.pop() // expensive if array is very large?
```
mnaamani, line 40: If implemented correctly pop is very fast as it only involves reducing array length and returning last element

storage-node/src/services/queryNode/api.ts
```diff
@@ -161,14 +161,25 @@ export class QueryNodeApi {
...
-  ): Promise<NodeT[]> {
+  >(query: DocumentNode, variables: CustomVariablesT, resultKey: keyof QueryT, itemsPerPage = MAX_RESULTS_PER_QUERY) {
+    const results = []
+    for await (const result of this.multipleEntitiesWithPaginationGenerator<NodeT, QueryT, CustomVariablesT>(
+      query,
+      variables,
+      resultKey,
+      itemsPerPage
+    )) {
+      results.push(...result)
```
mnaamani, line 172: don't bring this syntax back.. just testing things out here.

PR #5033, opened 2024-01-10
Colossus 3.10.2

mnaamani:
After doing some [measurements](https://github.com/Joystream/joystream/blob/cf117617388ae41dc9f82407d2561d433142e695/storage-node/src/commands/util/test-1.ts) on the performance of `lodash.differenceWith` it became evident that this was main cause of high CPU usage at start of sync run when handling very large arrays.

In this PR we use the `Map.has()` and `Set.has()` method in combination with `Array.filter` to get faster performance and avoid the node process hanging for multiple seconds at a time.

In addition to task processor is updated to catch task failures to avoid entire sync run ending prematurely.

I have chosen not to include additional improvements from https://github.com/Joystream/joystream/pull/5026 for now although they will be very beneficial.

edit:
I was curios why `v3.8.x` version had no issues, and it seems there is a huge performance hit we got when we switched from using `_.difference()` to `_.differenceWith()` in `v3.9.x` and up

```
_.difference   23ms
_.differenceWith   14480ms
filter with array.includes   8678ms
filter map.has   2ms
```
Example for running test on my branch https://github.com/mnaamani/joystream/tree/colossus-tweaks
`yarn storage-node util:test-1 --obligationCount 300000 --storedCount 100000`

### DIFF
--- a/storage-node/src/services/sync/cleanupService.ts
+++ b/storage-node/src/services/sync/cleanupService.ts
@@ -1,4 +1,3 @@
-import _ from 'lodash'
 import superagent from 'superagent'
 import urljoin from 'url-join'
 import { getDataObjectIDs } from '../../services/caching/localDataObjects'
@@ -73,8 +72,8 @@ export async function performCleanup(
     getStorageObligationsFromRuntime(qnApi, buckets),
     getDataObjectIDs(),
   ])
-  const assignedObjectsIds = model.dataObjects.map((obj) => obj.id)
-  const removedIds = _.difference(storedObjectsIds, assignedObjectsIds)
+  const assignedObjectsIds = new Set(model.dataObjects.map((obj) => obj.id))
+  const removedIds = storedObjectsIds.filter((id) => !assignedObjectsIds.has(id))
   const removedObjects = await getDataObjectsByIDs(qnApi, removedIds)
 
   logger.debug(`Cleanup - pruning ${removedIds.length} obsolete objects`)
--- a/storage-node/src/services/sync/cleanupService.ts
+++ b/storage-node/src/services/sync/cleanupService.ts
@@ -7,6 +7,7 @@ import { DataObjectDetailsFragment } from '../queryNode/generated/queries'
 import { DataObligations, getDataObjectsByIDs, getStorageObligationsFromRuntime } from './storageObligations'
 import { DeleteLocalFileTask } from './tasks'
 import { TaskProcessorSpawner, WorkingStack } from './workingProcess'
+import _ from 'lodash'
 
 /**
  * The maximum allowed threshold by which the QN processor can lag behind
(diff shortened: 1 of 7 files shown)


storage-node/src/services/sync/cleanupService.ts
```diff
@@ -73,8 +72,8 @@ export async function performCleanup(
     getStorageObligationsFromRuntime(qnApi, buckets),
     getDataObjectIDs(),
   ])
-  const assignedObjectsIds = model.dataObjects.map((obj) => obj.id)
-  const removedIds = _.difference(storedObjectsIds, assignedObjectsIds)
```
mnaamani, line 77: Should I revert this change?, because the problem seems only to be with `differenceWith` implementation, not `difference`

zeeshanakram3: Yeah, sure, if there appears to be no problem with the `_. difference` function, then lets have it

storage-node/src/services/sync/synchronizer.ts
```diff
@@ -45,15 +44,13 @@ export async function performSync(
...
+  const model = await getStorageObligationsFromRuntime(qnApi, buckets)
 
   const required = model.dataObjects
 
-  const added = _.differenceWith(required, files, (required, file) => required.id === file)
-  const removed = _.differenceWith(files, required, (file, required) => file === required.id)
+  const added = required.filter((obj) => !isDataObjectIdInCache(obj.id))
 
   logger.debug(`Sync - new objects: ${added.length}`)
-  logger.debug(`Sync - obsolete objects: ${removed.length}`)
```
zeeshanakram3, line 56: I think it's good to have this log, especially if the cleanup service is disabled so that you can know about the state/status of the data objects in your node. WDYT?

For example, operator misconfigured their node with the incorrect bucket, so the log `Sync - new objects: ` probably wont provide us any useful info but `Sync - obsolete objects:`  will show that all/most  of the local data objects are obsolete?

storage-node/src/services/sync/cleanupService.ts
mnaamani: Yeah it is cleaner with `_.difference`.
Will do.

storage-node/src/services/sync/synchronizer.ts
mnaamani: Good point. Will revert.

storage-node/src/services/sync/cleanupService.ts
mnaamani: Worth noting, _.difference was still 10x slower than doing filter+Map.has

Also there is use of  `Array.some()` later in the cleanup service.. maybe worth doing some timing tests on that?

mnaamani:
The failing CI tests has finally got me frustrated enough to actually fix in https://github.com/Joystream/joystream/pull/5037

zeeshanakram3 (approved):
LGTM.

PR #5036, opened 2024-01-11
[Argus] always parse Axios errors before logging

kdembler:
Added winston format that will always pre-process logs and make sure we don't log raw axios error. This way we don't need to remember to do this in all the different places where the error can be logged

### DIFF
--- a/distributor-node/CHANGELOG.md
+++ b/distributor-node/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.5.1
+
+- Added parsing of Axios errors on logger level so that we never log the whole Axios client instance (which is a circular object and causes the node to crash)
+
 ## 1.5.0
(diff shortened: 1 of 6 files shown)


mnaamani (changes requested):
Looks good, just a dangling import in `src/services/networking/storage-node/api.ts` making the linter unhappy.
```
/home/runner/work/joystream/joystream/distributor-node/src/services/networking/storage-node/api.ts
Error:   6:10  error  'parseAxiosError' is defined but never used  @typescript-eslint/no-unused-vars
```

mnaamani (approved):

mnaamani:
LGTM

Don't worry about the failing tests I've got a fix here https://github.com/Joystream/joystream/pull/5037

PR #5046, opened 2024-01-16
Expose Storage Squid version in Argus & Colossus status endpoints

### LINKED ISSUE #4732: Expose information about QN used in core nodes

zeeshanakram3:
closes #4732

based on #5001


mnaamani:
Perhaps this can be rebased/updated from master ? @zeeshanakram3

zeeshanakram3:
> Perhaps this can be rebased/updated from master ? @zeeshanakram3

Done

kdembler (approved):
LGTM, tested both Argus and Colossus

PR #5054, opened 2024-01-22
[Colossus] Remove sync i/o calls

### LINKED ISSUE #5050: Avoid using synchronous fs api
To minimize event loop blocking, eliminate where possible use of synchronous file i/o.
eg.
  - `fs.existsSync`
  - `readChunk.sync` in `services/helpers/fileInfo.ts:getFileInfo()` -> maybe fileInfo can be cached in memory so we don't need to repeat it on every HEAD/GET request for a file.

mnaamani:
Fixes https://github.com/Joystream/joystream/issues/5050

- Changed sync io to async io.
- Did not add caching results of fileInfo.


storage-node/src/services/webApi/app.ts
```diff
@@ -251,13 +251,13 @@ async function validateUploadFileParams(req: express.Request, res: express.Respo
...
-
   const isInStorage = getDataObjectIdFromCache(dataObjectId.toString())
   if (isInStorage) {
-    throw new WebApiError(`Data object ${dataObjectId} already exists`, 400)
+    throw new WebApiError(`Data object ${dataObjectId} already exists (in storage)`, 400)
+  }
+
+  const isObjectPending = await res.locals.acceptPendingObjectsService.pendingObjectExists(dataObjectId.toString())
+  if (isObjectPending) {
+    throw new WebApiError(`Data object ${dataObjectId} already exists (pending)`, 400)
```
zeeshanakram3: I think it makes sense to check the existence of the data object in `pending` directory first and then check in the `upload` state, _doing otherwise might incorrectly report that the data object is missing (due to race condition) but in fact, it is present_

For example based on your change:
- Say the object exists in the pending dir, so `isInStorage` would be false, but then the object is moved to the upload dir so `isObjectPending` would also be false. Which is not what we want

But if we check the object in pending dir first this case would never happen. WDYT?

zeeshanakram3 (commented):
Looks good, do you think it makes sense to do this change for the `distributor-node` too?

storage-node/src/services/sync/tasks.ts
```diff
@@ -104,15 +109,18 @@ export class DownloadFileTask implements SyncTask {
...
+        } catch (err) {
+          continue
         }
       } catch (err) {
         logger.error(`Sync - fetching data error for ${this.dataObjectId}: ${err}`, { err })
       }
     }
 
-    logger.warn(`Sync - cannot get operator URLs for ${this.dataObjectId}`)
+    logger.warn(`Sync - Failed to download ${this.dataObjectId}`)
```
kdembler, line 123: Not part of this PR but shouldn't the severity be the other way around actually? Warning for a single download and error if it wasn't possible to get an object at all?

storage-node/src/services/webApi/app.ts
mnaamani: Good point, I hadn't considered that.
The reason I switched order was based on idea that it is more likely for an object to already exist in the uploads folder than in the pending folder, and was trying to reduce a potentially wasteful check in pending first.
But catching that case makes more sense to avoid storage node wasting bandwidth allowing another upload for existing object in pending state. Will switch order back.

mnaamani:
> Looks good, do you think it makes sense to do this change for the `distributor-node` too?

Any opportunity where it makes sense to eliminate blocking the event loop for too long  should be investigated. I'm not as familiar with the distributor-node code base but yes in general it makes sense.

storage-node/src/services/sync/tasks.ts
mnaamani: Yeah I think we need to have a PR just for improving the logging.
Also it is a bit more nuanced. We would need better error type checks to know if a download failing is an error, due to a problem from the local operator side, or from the remote node.

storage-node/src/services/webApi/app.ts
mnaamani: Fixed in [d06dd32](https://github.com/Joystream/joystream/pull/5054/commits/d06dd3274d90afff5ad8c862ec345b593a9ed07a)

mnaamani:
> * Did not add caching results of fileInfo.

Should I include this in the PR, it would reduce latency accessing files.

Edit:
Added in [8bb4be3](https://github.com/Joystream/joystream/pull/5054/commits/8bb4be39d42e78ef14888031f7c0fa111aa7dcec)

storage-node/src/services/helpers/fileInfo.ts
```diff
@@ -37,19 +45,20 @@ const MINIMUM_FILE_CHUNK = 4100
...
   const { size } = await fsPromises.stat(fullPath)
 
-  return {
+  const info = {
     mimeType: fileType ? fileType.mime.toString() : DEFAULT_FILE_INFO.mimeType,
     ext: fileType ? fileType.ext.toString() : DEFAULT_FILE_INFO.ext,
     size,
   }
+
+  FileInfoCache.set(fullPath, info)
```
zeeshanakram3: Shouldn't we just use data object Ids as keys? This way the key size would be much smaller.

mnaamani: Yes it would, however since this function is taking a path to a file, it is not "aware" of object ids per say. This also suggests it is the wrong place to actually do the caching. Maybe in the express app, the request handlers for files GET/HEAD should be responsible for the caching instead.

mnaamani: refactored in [d5f021a](https://github.com/Joystream/joystream/pull/5054/commits/d5f021a0661aad1cb4459862f9f120ebd04583b9)

zeeshanakram3 (approved):
LGTM

PR #5057, opened 2024-01-23
[CLI] Add  `working-groups:verifyValidator`

thesan:
Replace #4933

- #4930

@zeeshanakram3 
- This can only be called by membership working group workers so regarding https://github.com/Joystream/joystream/pull/4933#discussion_r1463151390 I also enforced `WorkingGroups.Membership` but in a different way than #4933. If you think it's a bad idea I can remove it. (Also for the same reason I didn't document the `--group` `-g` flags.
- I updated the Readme but I'm not sure whether it's supposed to be automatized.
- I tested the command on [Atlas Dev](https://atlas-dev.joystream.org/network/config.json) it executed without error but the validator status wasn't set correctly `true` (but maybe the this network processor isn't up to date :shrug:). I'll have to check locally.

### DIFF
--- a/cli/src/commands/working-groups/verifyValidator.ts
+++ b/cli/src/commands/working-groups/verifyValidator.ts
@@ -7,20 +7,21 @@ import Long from 'long'
 import chalk from 'chalk'
 
 export default class VerifyValidatorCommand extends WorkingGroupsCommandBase {
-  static description = 'Membership lead/worker verifies validator membership profile'
+  static description =
+    'Verify or un-verify the membership profile bound to a validator account. Available to membership workers only.'
 
   static args = [
     {
       name: 'memberId',
       required: true,
-      description: 'Membership ID of the validator to verify',
+      description: 'ID of the membership bound to the validator account.',
     },
   ]
 
   static flags = {
     unverify: flags.boolean({
       default: false,
-      description: 'Verification state of the validator',
+      description: 'Whether the profile should be un-verified.',
     }),
 
     ...WorkingGroupsCommandBase.flags,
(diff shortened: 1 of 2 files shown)


cli/src/commands/working-groups/verifyValidator.ts
```diff
@@ -0,0 +1,60 @@
...
+      description: 'Whether the profile should be un-verified.',
+    }),
+
+    ...WorkingGroupsCommandBase.flags,
+  }
+
+  async init(): Promise<void> {
+    await super.init()
+    this._group = WorkingGroups.Membership
+  }
```
zeeshanakram3:
```
 ./bin/run working-groups:verifyValidator 2 
Initializing the query node connection (http://localhost:8081/graphql)...... done
Initializing the api connection (ws://localhost:9944)...... done
Current Group: storageProviders
```
Due to this change, whenever you run the command `Current Group: storageProviders` [log](https://github.com/thesan/joystream/blob/6ffd6ec85e03ebf3e19f95f8c683c133f062c103/cli/src/base/WorkingGroupsCommandBase.ts#L88) will be printed, which might be confusing for the user

Also, instead of changing the group in code wouldn't it better to throw the error if user didn't specify the `--group=membership` flag (i.e. explicitly asking user to run command in membership group context)

zeeshanakram3 (commented):
> I updated the Readme but I'm not sure whether it's supposed to be automatized.

Good idea. Maybe add this as part of build step?

zeeshanakram3 (approved):
LGTM

PR #5061, opened 2024-01-25
[QN] Fix validator verification

thesan:
- #4824

based on #5057


tests/network-tests/src/fixtures/membership/VerifyValidatorMembershipFixture.ts
```diff
@@ -49,22 +50,38 @@ export class VerifyValidatorMembershipFixture extends WithMembershipWorkersFixtu
...
 
   async runQueryNodeChecks(): Promise<void> {
     await super.runQueryNodeChecks()
-    const qMembers = await this.query.getMembersByIds(this.verifications.map((m) => m.memberId))
-    this.assertQueriedMembershipsAreValid(qMembers)
+
+    const memberIds = this.verifications.reduce(
+      (ids, { memberId }) => (ids.some((id) => memberId.eq(id)) ? ids : [...ids, memberId]),
+      [] as MemberId[]
+    )
```
zeeshanakram3: Don't you think following would be much simpler 
```suggestion
    const memberIds = [...new Set(this.verifications.map((v) => v.memberId))]
```

zeeshanakram3 (changes requested):
Look good, while testing this PR I found a logical bug in the `verifyValidator` CLI command which I overlooked in the PR review.

https://github.com/Joystream/joystream/blob/master/cli/src/commands/working-groups/verifyValidator.ts#L32

It should be the following 
```ts
 if (!worker || this.group !== WorkingGroups.Membership) {
```

tests/network-tests/src/fixtures/membership/VerifyValidatorMembershipFixture.ts
thesan: :100:

thesan:
> ```ts
>  if (!worker || this.group !== WorkingGroups.Membership) {
> ```

:sweat: 
Nice catch!

zeeshanakram3 (approved):
LGTM

Issue #3494, opened 2022-03-25
Proposal: Reduce from council budget


bedeho:
Council should be able to burn budget capacity as a way to take future minting off the table.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202132419573087/1202202225872277) by [Unito](https://www.unito.io)

PR #5062, opened 2024-01-26
Luxor/reduce council budget

dobertRowneySr:
addresses: #3494

Budget is simply reduced by the amount (there is no token minting until workers/councilors gets paid)
I have also added a proposal, with the same parameters as the `freeze_pallet` proposal
- [x] add extrinsic to council pallet
- [x] add tests to council pallet
- [x] add proposal to codex pallet
- [x] add tests to codex pallet
- [x] add runtime configuration
- [x] add integration tests


kdembler:
@ignazio-bovo please change target branch to `luxor`

runtime-modules/proposals/codex/src/lib.rs
```diff
@@ -503,6 +512,11 @@ decl_module! {
         const SetMaxValidatorCountProposalMaxValidators: u32 =
             T::SetMaxValidatorCountProposalMaxValidators::get();
 
+        /// Decrease Cuncil budget parameters
```
kdembler:
```suggestion
        /// Decrease Council budget parameters
```

runtime-modules/proposals/codex/src/weights.rs
```diff
@@ -1019,6 +1020,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(7_u64))
 			.saturating_add(T::DbWeight::get().writes(9_u64))
 	}
+
+	fn create_proposal_decrease_council_budget(t: u32, d: u32, ) -> Weight {
+		Weight::from_parts(0, 0)
```
kdembler, line 1025: This needs to be adjusted right?

runtime/src/proposals_configuration/defaults.rs
```diff
@@ -357,3 +357,17 @@ pub(crate) fn freeze_pallet_proposal() -> ProposalParameters<BlockNumber, Balanc
...
+pub(crate) fn decrease_council_budget() -> ProposalParameters<BlockNumber, Balance> {
+    ProposalParameters {
+        voting_period: days!(3),
+        grace_period: 0,
+        approval_quorum_percentage: TWO_OUT_OF_THREE,
+        approval_threshold_percentage: TWO_OUT_OF_THREE,
+        slashing_quorum_percentage: ALL,
+        slashing_threshold_percentage: ALL,
+        required_stake: Some(dollars!(50)),
+        constitutionality: 1,
```
kdembler: We need to adjust those, will get back to you

kdembler (commented):
Look good to me, will approve once we finalize the params for proposal, waiting for council feedback

freakstatic (approved):
LGTM!

runtime/src/proposals_configuration/defaults.rs
kdembler: Council has approved a signal to make this 3/3 approval, 1 constitutionality: https://pioneer.joyutils.org/#/proposals/preview/797

runtime-modules/proposals/codex/src/weights.rs
dobertRowneySr: Hi, yes but they will be generated by @mnaamani on the ref. hardware

dobertRowneySr: And not on this pr

runtime/src/proposals_configuration/defaults.rs
dobertRowneySr: Let me know if also the quorum percentage is correct

textureAtlas.png
kdembler, line 1: You probably didn't want to include this :D

mnaamani (changes requested):
Runtime implementation looks good.
I think you ran the query-node build before the runtime spec was updated, that is why the query-node/chain-metadata/2002.json file is modified. Please revert it to one from master.
The integration test fails because the mappings do not handle case of the new proposal, so the processor fails to handle the event. Do you want to add the mapping and schema update for QN in this PR or a separately?

dobertRowneySr:
The QN part is handled separately

runtime-modules/council/src/lib.rs
```diff
@@ -843,6 +851,38 @@ decl_module! {
...
+        /// Decrease the council total budget
+        ///
+        /// # <weight>
+        ///
+        /// ## weight
+        /// `O (1)`
+        /// - db:
+        ///    - `O(1)` doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = 10_000_000] // TODO: adjust
```
mnaamani: Should we add a weight function for this dispatch

runtime-modules/proposals/codex/src/tests/mock.rs
```diff
@@ -105,6 +104,7 @@ frame_support::construct_runtime!(
         OperationsWorkingGroupBeta: working_group::<Instance7>::{Pallet, Call, Storage, Event<T>},
         OperationsWorkingGroupGamma: working_group::<Instance8>::{Pallet, Call, Storage, Event<T>},
         DistributionWorkingGroup: working_group::<Instance9>::{Pallet, Call, Storage, Event<T>},
+        Council: council::{Pallet, Call, Storage, Event<T>},
```
mnaamani, line 107: Why did you find it necessary to move the Council pallet down here.

mnaamani (changes requested):
Just need to add a weight function for the `decrease_council_budget` dispatch call.

runtime-modules/council/src/lib.rs
dobertRowneySr: This was part of a bigger issue that there are no benchmarks.
I have added that for the `reduce_council_budget`, thanks for pointing it out

runtime-modules/proposals/codex/src/tests/mock.rs
dobertRowneySr:
```
error[E0277]: the trait bound `RuntimeEvent: From<pallet_council::RawEvent<u64, u64, u64, u64>>` is not satisfied
   --> runtime-modules/proposals/codex/src/tests/mock.rs:696:25
    |
696 |     type RuntimeEvent = RuntimeEvent;
    |                         ^^^^^^^^^^^^ the trait `From<pallet_council::RawEvent<u64, u64, u64, u64>>` is not implemented for `RuntimeEvent`
```

dobertRowneySr: without that

query-node/mappings/src/proposals.ts
```diff
@@ -109,7 +108,7 @@ async function getOrCreateRuntimeWasmBytecode(store: DatabaseManager, bytecode:
 async function parseProposalDetails(
   event: SubstrateEvent,
   store: DatabaseManager,
-  proposalDetails: RuntimeProposalDetails_V1001 | RuntimeProposalDetails_V2002
+  proposalDetails: RuntimeProposalDetails_V2003
```
mnaamani, line 111: I think this change is okay, although we probably should implement this function in future to account for any removed proposals between runtimes. To date we have only been adding new proposals.

runtime-modules/council/src/lib.rs
```diff
@@ -843,6 +851,38 @@ decl_module! {
...
+        ///
+        /// # <weight>
+        ///
+        /// ## weight
+        /// `O (1)`
+        /// - db:
+        ///    - `O(1)` doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = CouncilWeightInfo::<T>::decrease_council_budget()]
+        pub fn decrease_council_budget(origin, reduction_amount: Balance<T>) -> Result<(), Error<T>> {
```
mnaamani, line 864: when adding a new dispatch that is invoked by a proposal, it is best to try to add it at the end so that the method index of existing dispatches doesn't change, so we can have the option to not remove all pending proposals on runtime upgrades.

runtime-modules/proposals/codex/src/tests/mock.rs
mnaamani: Is is something about the way this particular mock is implemented with respect to types. We do not seem to need to do the same for the main runtime/lib.rs

mnaamani (approved):
I think this is good to merge, just left some general comments.

Issue #4948, opened 2023-10-26
WGs: Discretionary spending with vesting


mochet:
Similar to: https://github.com/Joystream/joystream/issues/4808

Make WG discretionary spending feature with the ability to apply a vesting schedule. There are many payments that the council entrusts WGs to make that could be made to service providers and similar and vesting payments for these have a use--the current system means the WG lead must withdraw the funds to a personal account and then send them with a vesting schedule to the recipient.

dobertRowneySr:
There shouldn't be much be much complexity here, there's a premade extrinsic called `vested_transfer` which is can be leveraged.
https://github.com/paritytech/substrate/blob/ad5399644aebc54e32a107ac37ae08e6cd1f0cfb/frame/vesting/src/lib.rs#L349

kdembler:
https://github.com/Joystream/joystream/pull/5072

PR #5072, opened 2024-02-05
Luxor/vested wg spending

dobertRowneySr:
Addresses: #4948 
based on top of #5062 

Note: I have used the extrinsic `vested_transfer` inside the extrinsic `spend_from_budget` all extrinsic in substrate are now `transactional` by default, this means that if the `vested_transfer` fails then `spend_from_budget` is equivalent to a no-op.
Also I have created one internal account for each WG where tokens are minted before they are transferred with vesting.


dobertRowneySr:
To be rebased onto #5062

mnaamani:
I think there is a simpler way to do this, without needing to have an account on each working group.
Instead of `pay_from_budget` into the working group account then doing `vesting::vested_transfer` you should be able to pay from the budget directly into the target account, then do `vesting::force_vested_transfer` with source and target being the final destination. The balances::transfer inside force_vested_transfer will be a successful no-op, and then the vesting schedule will be applied.

mnaamani:
> Also I have created one internal account for each WG where tokens are minted before they are transferred with vesting.

https://github.com/Joystream/joystream/pull/5072#issuecomment-2002164477

runtime-modules/working-group/src/lib.rs
```diff
@@ -1140,31 +1152,60 @@ decl_module! {
...
+                // we do this to avoid migrations when updating to Luxor
+                let treasury_account_id = T::ModuleId::get().into_sub_account_truncating(I::INDEX);
+                let treasury_amount = Balances::<T>::usable_balance(&treasury_account_id);
+                if treasury_amount < T::ExistentialDeposit::get() {
+                    let treasury_initial_deposit = T::ExistentialDeposit::get().saturating_sub(treasury_amount);
+                    let _ = <balances::Pallet<T>>::deposit_creating(&treasury_account_id, treasury_initial_deposit);
+                }
+
+                // first mint into the treasury account
+                Self::pay_from_budget(&treasury_account_id, amount);
```
mnaamani: ref: https://github.com/Joystream/joystream/pull/5072#issuecomment-2002164477

```rust
Self::pay_from_budget(account_id, amount);
vesting::Pallet::<T>::force_vested_transfer(
  RawOrigin::Root,
  account_id,
  account_id,
  vesting_schedule
)
```

mnaamani (changes requested):
Can you confirm that [force_vested_transfer](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/vesting/src/lib.rs#L371) the way I'm suggested can be done to simplify this implementation?

dobertRowneySr:
I need to add benchmarks

runtime-modules/working-group/src/lib.rs
```diff
@@ -1626,6 +1681,11 @@ impl<T: Config<I>, I: Instance> Module<T, I> {
             .map(|(worker_id, _)| worker_id)
             .collect()
     }
+
+    /// Returns the account for the  module used for vested spending
+    pub fn module_treasury_account() -> T::AccountId {
```
mnaamani: this can be dropped

runtime-modules/working-group/src/lib.rs
```diff
@@ -129,6 +137,9 @@ pub trait Config<I: Instance = DefaultInstance>:
 
     /// Stake needed to create an opening
     type LeaderOpeningStake: Get<Self::Balance>;
+
+    /// Type for the Module Id account in order to locked payment the budget.
+    type ModuleId: Get<PalletId>;
```
mnaamani: As we no longer need to store funds in the working group, this can be dropped

runtime-modules/working-group/src/tests/fixtures.rs
```diff
@@ -1159,6 +1162,78 @@ impl SetStatusTextFixture {
...
+
+        let new_budget = TestWorkingGroup::budget();
+        let new_balance = self
+            .vesting_schedule
+            .locked_at::<BlockNumberToBalance>(<frame_system::Pallet<Test>>::block_number());
+        let amount = self.vesting_schedule.locked();
+
+        if actual_result.is_ok() {
+            assert_eq!(new_budget, old_budget - amount);
+            assert_eq!(new_balance, new_balance);
```
mnaamani: not sure what is being checked here?

mnaamani (changes requested):

mnaamani:
I think this now needs to be rebased on luxor since I merged the reduce council budget PR

runtime-modules/working-group/src/tests/fixtures.rs
dobertRowneySr: legacy code that I did not remove..

tests/network-tests/src/fixtures/workingGroups/SpendBudgetFixture.ts
```diff
@@ -9,15 +9,27 @@ import { ISubmittableResult } from '@polkadot/types/types/'
...
+export type VestingSchedule = {
+  locked: BN
+  perBlock: BN
+  startingBlock?: BN
+}
+
 export class SpendBudgetFixture extends BaseWorkingGroupFixture {
   protected recievers: string[]
-  protected amounts: BN[]
+  protected vestingSchedules: VestingSchedule[]
```
thesan: Should this be changed ? The fixture is still called `SpendBudgetFixture` not `SpendVestedBudgetFixture`

dobertRowneySr: change reverted, integration tests needs to be taken care of in a separate PR. 
Namely add this Fixture with Flow

runtime-modules/content/src/tests/mock.rs
```diff
@@ -476,6 +479,15 @@ impl common::council::CouncilBudgetManager<U256, u64> for CouncilBudgetManager {
     }
 }
 
+parameter_types! {
+    pub const DWGModuleId: PalletId = PalletId(*b"m_distwg");
+    pub const SWGModuleId: PalletId = PalletId(*b"m_storwg");
+    pub const CWGModuleId: PalletId = PalletId(*b"m_contwg");
```
mnaamani: ModuleId not required anymore

runtime-modules/content/src/tests/mock.rs
```diff
@@ -490,6 +502,8 @@ impl working_group::Config<StorageWorkingGroupInstance> for Test {
     type WeightInfo = ();
     type MinimumApplicationStake = MinimumApplicationStake;
     type LeaderOpeningStake = LeaderOpeningStake;
+    type VestingBalanceToBalance = BalanceConverter;
+    type ModuleId = SWGModuleId;
```
mnaamani: ModuleId not a configuration trait anymore so should be removed

runtime/src/lib.rs
```diff
@@ -1355,6 +1355,7 @@ parameter_types! {
     pub const MinimumApplicationStake: Balance = dollars!(20);
     // This should be more costly than `add_opening` fee
     pub const LeaderOpeningStake: Balance = dollars!(100);
+    pub const WorkingGroupModuleId: PalletId = PalletId(*b"mworking"); // module storage
```
mnaamani, line 1358: not required

tests/network-tests/src/fixtures/workingGroups/SpendBudgetFixture.ts
```diff
@@ -9,6 +9,12 @@ import { ISubmittableResult } from '@polkadot/types/types/'
 import { Utils } from '../../utils'
 import { BudgetSpendingEventFieldsFragment, WorkingGroupFieldsFragment } from '../../graphql/generated/queries'
 
+export type VestingSchedule = {
```
mnaamani, line 12: unused type, I think you mentioned integration tests will be updated in a separate PR?

mnaamani (changes requested):
some additional cleanup for build to works after refactor (removing the moduleId in working groups)

tests/network-tests/src/fixtures/workingGroups/SpendBudgetFixture.ts
dobertRowneySr: yes

mnaamani (commented):
Benchmark is failing

```
Error: Input("Benchmark working_group::vested_spend_from_budget failed: AmountLow")
2024-03-29 22:54:25 Starting benchmark: working_group::vested_spend_from_budget    
There was a problem generating the weights for working_group, check the error above
Error: Process completed with exit code 1.
```

runtime-modules/working-group/src/benchmarking.rs
```diff
@@ -787,12 +788,34 @@ benchmarks_instance! {
...
+    }
+
+    vested_spend_from_budget {
+        let (lead_id, _) = insert_a_worker::<T, I>(
+            OpeningType::Leader,
+            0,
+            None
+        );
+
+        let amount_vesting = VestingBalanceOf::<T>::from(10_000u32);
```
mnaamani: Benchmark is failing with error here`AmountLow` because amount is less than vesting pallet configuration of `<T as vesting::Config>::MinVestedTransfer::get()`

mnaamani: So perhaps use that constant to make the `amount_vesting` at least that amount.

runtime-modules/working-group/src/benchmarking.rs
```diff
@@ -787,12 +788,34 @@ benchmarks_instance! {
...
+        let budget = BalanceOf::<T>::from(100_000u32);
+        let block_no = 100u32;
```
mnaamani: `block_no` is not used

mnaamani (changes requested):
from earlier

runtime-modules/working-group/src/benchmarking.rs
ignazio-bovo: There's a constraint in the runtime lib: 
```rust
parameter_types! {
    pub const MinVestedTransfer: Balance = dollars!(1);
    pub UnvestedFundsAllowedWithdrawReasons: WithdrawReasons = WithdrawReasons::empty();
}
```

I cannot reach 1 dollar with a `u32`, as I was setting `VestedBalance::from(200000000u32)` and it complains if you try to use from with a u64. hence I have added a feature flag to enable this only when runtime benchmarks to disable this during benchmarks.
u32 and u64 are copied by value in rust so there should not be difference in benchmarking result

mnaamani:
strange that I'm seeing errors like `../../query-node/start.sh: line 11: docker-compose: command not found` in CI workflow runs..
Has it been deprecated and we should replace all occurrences of that command with `docker compose` command?

I think you forgot to commit changes to: `types/src/augment/augment-api-rpc.ts` and  `types/src/augment/augment-types.ts`

mnaamani (approved):
LGTM, next PR to luxor should commit changes to types/src/*.rs missed in this PR

Running integration tests locally on my machine works, we need to investigate why they are failing on github runners.

PR #5073, opened 2024-02-07
feat: :art: set the amm threshold to zero

dobertRowneySr:
set the amm threshold to zero


mnaamani (approved):
Approving as agreed. Will just wait for cargo unit tests to complete before merging.

kdembler (approved):
LGTM

PR #5075, opened 2024-02-07
bump QN version for Nara

kdembler:


mnaamani:
@kdembler Could you also bump to same version in `query-node/mappings/package.json`

mnaamani (approved):

PR #5076, opened 2024-02-16
Nara fixes for amm

dobertRowneySr:
As agreed :
- selling is not possible if the `amm_provided_supply < amount_of_crt_to_sell`
- transaction fees from the amm are burned immediately
- I have added a helper function on the `numerical.rs` utility library, I plan on adding more test, in the meantime the review can start


freakstatic (approved):
LGTM!

runtime-modules/common/src/numerical.rs
```diff
@@ -124,11 +119,7 @@ pub fn amm_eval_inner<
         })
         .flatten();
 
-    if lower_bound == x {
-        res.map(|res_| SignedResult::Positive(res_))
-    } else {
-        res.map(|res_| SignedResult::Negative(res_))
-    }
+    res
```
freakstatic, line 122: typo?

kdembler:
~@ignazio-bovo sorry if it's a stupid question, but why does this PR result in changes in metadata?~

Edit: nevermind, that's because of new error type

runtime-modules/project-token/src/lib.rs
```diff
@@ -893,10 +894,13 @@ decl_module! {
                 token_data.increase_amm_bought_amount_by(amount);
             });
 
-            // TODO: redirect tx fees revenue to council
+            // transfer tx_fee * price + price
             Self::transfer_joy(&sender, &amm_treasury_account, buy_price)?;
 
-            Self::deposit_event(RawEvent::TokensBoughtOnAmm(token_id, member_id, amount, buy_price));
+            // burn tx_fee * price
+            let _ = burn_from_usable::<T>(&amm_treasury_account, buy_tx_fee);
```
kdembler, line 901: Are there any noteworthy tradeoffs between:
a) transferring to treasury and then burning fee from treasury
b) burning fee from user account and then transferring the rest to the treasury

runtime-modules/project-token/src/lib.rs
```diff
@@ -893,10 +894,13 @@ decl_module! {
...
+
+            Self::deposit_event(RawEvent::TokensBoughtOnAmm(token_id, member_id, amount, price));
```
kdembler: Previously this event emitted full price including the fee, now it will include only non-burnt price. Was that a desired change?

dobertRowneySr: Thank you for pointing this out
I have realised that, keeping in mind that using the orion code for crt below:

```typescript
  overlay.getRepository(AmmTransaction).new({
    ammId: activeAmm.id,
    accountId: buyerAccount.id,
    id: overlay.getRepository(AmmTransaction).getNewEntityId(),
    transactionType: AmmTransactionType.BUY,
    createdIn: block.height,
    quantity: crtMinted,
    pricePaid: joysDeposited, 
    pricePerUnit: joysDeposited / crtMinted, // truncates decimal values
  })
}
```

`pricePaid` is the amount paid by the user
`pricePaid - TxFees` is the amount in the treasury account (not accounted anywhere in Orion)

Hence I need to put it back to `buy_price` (and `sell_price` for the sell extr.) because that is the amount the user cares about

runtime-modules/project-token/src/lib.rs
dobertRowneySr: No that comes to mind

kdembler (approved):
LGTM, would be good to get Mokhtar's approval as well

kdembler (approved):
LGTM, would be good to get Mokhtar's approval as well

Issue #3702, opened 2022-05-03
Control validator security budget


bedeho:
# Background
Right now (`Olympia`), it is not feasible for the council to set the total $JOY reward devoted to validators over a given period of time, this not only negatively affects the testnet incentives, but it also 

# Proposal
Introduce ability for this to be controlled directly. Add support in proposal #3626

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202078489349362/1202217411332176) by [Unito](https://www.unito.io)

traumschule:
The council should be able to set `max_inflation` and `ideal_stake` with low grace period (similar to max validator count).

mochet:
For some context on urgency with this GH issue, this forum post shares a % look into how much we are currently paying validators/nominators, it is a significant amount: https://pioneerapp.xyz/#/forum/thread/809

kdembler:
Needs estimation of the effort

kdembler:
To give a status update: we've considered including this functionality in Nara but decided not to.

The validator rewards are controlled by a curve defined in the runtime code: https://github.com/Joystream/joystream/blob/01444997bd3bb0e0c870c91406e4c9c5d40fdadd/runtime/src/lib.rs#L515-L524

Our initial idea was to add a proposal that can adjust the parameters to construct the curve as seen above. However, Mokhtar has pointed out that the process of compiling the curve - using the parameters to calculate set of curve points - is currently done during runtime build and may be too heave of a calculation to actually do at runtime with a proposal.

So an alternative idea is to have the curve itself be a mutable storage value and have a proposal that changes the curve directly (not its parameters). That means we would compute the curve off-chain and only submit the result of that calculation as a proposal.

This however requires some basic tooling - we need to be able to construct the curve given a set of parameters but also to somehow verify a proposed curve. We need to research how the curve is stored and what the inputs to a proposal would look like

dobertRowneySr:
We want the same curve structure just the ability to change parameters, right? @bedeho @kdembler

kdembler:
@ignazio-bovo I think yes, but what exactly do you mean by the curve structure?

dobertRowneySr:
The math formula being used.
I have figured out a way to do it in a clean way, ETA: ??days (depending on what do we want to change)
**EDIT**:
The way substrate construct the curve is by using a macro, for which the curve is quantised into around 1,000,000 intervals (via while loop with a lot of iterations)
[...]

PR #5086, opened 2024-02-27
Luxor/reward-curve-proposal

dobertRowneySr:
addresses: https://github.com/Joystream/joystream/issues/3702


runtime-modules/council/src/lib.rs
```diff
@@ -367,6 +368,10 @@ decl_storage! { generate_storage_info
 
         /// Councilor reward per block
         pub CouncilorReward get(fn councilor_reward) config(): Balance<T>;
+
+        /// Era payou damping factor: a parameter in [0,1] that can be used to reduce the era
+        /// payout without changing the reward curve directly
+        pub EraPayoutDampingFactor get(fn era_payout_damping_factor) config(): Percent;
```
mnaamani: We probably want the default/initial damping factor to be 100%

runtime/src/lib.rs
```diff
@@ -536,11 +537,21 @@ impl EraPayout<Balance> for NoInflationIfNoEras {
             // PoA mode: no inflation.
             (0, 0)
         } else {
-            <pallet_staking::ConvertCurve<RewardCurve> as EraPayout<Balance>>::era_payout(
-                total_staked,
-                total_issuance,
-                era_duration_millis,
-            )
+            // rescale the era payout according to the damping factor while keeping the invariant
+            // era_payou + rest = max_payout
```
mnaamani, line 541: By keeping this invariant, the damping factor is only reducing the payout to the validator, the rest, which increases if damping factor is less than 100% goes to the nominators. So we are not reducing the inflation only payout amount to validators. Is this the intention?
If we wish to reduce the overall inflation, then the max_payout needs to be scaled as well.

mnaamani: Sorry my mistake, the staking pallet will send the "remainder" (rest) amount to be handled by `RewardRemainder` trait (not for nominators). We have set this trait as empty tuple `()`, so the amount will be discarded.

runtime-modules/council/src/lib.rs
dobertRowneySr: I believe I have set it to 100% to the chainspec file

runtime-modules/council/src/lib.rs
```diff
@@ -843,6 +859,64 @@ decl_module! {
...
+        ///
+        /// # <weight>
+        ///
+        /// ## weight
+        /// `O (1)`
+        /// - db:
+        ///    - `O(1)` doesn't depend on the state or parameters
+        /// # </weight>
+        #[weight = 10_000_000] // TODO: adjust
+        pub fn decrease_council_budget(origin, reduction_amount: Balance<T>) -> Result<(), Error<T>> {
```
freakstatic: @ignazio-bovo are you sure you want to include this change?
seems to be out of scope

dobertRowneySr (commented):
I think I still have to add benchmarks for this but at this point I would wait before 
- Council reduce budget
- Budget WG spending

runtime-modules/council/src/lib.rs
mnaamani: Yes I see it. We will need to add a runtime migration to also set it to 100%.

runtime-modules/council/src/lib.rs
mnaamani: I think its because he is building on prior PRs

types/src/augment/augment-api-consts.ts
```diff
@@ -585,10 +585,6 @@ declare module '@polkadot/api-base/types/consts' {
        * Exports `Set Councilor Reward Proposal Parameters` proposal parameters.
        **/
       setCouncilorRewardProposalParameters: PalletProposalsEngineProposalParameters & AugmentedConst<ApiType>;
-      /**
-       * Era payout damping factor
-       **/
-      setEraPayoutDampingFactorProposalParameters: PalletProposalsEngineProposalParameters & AugmentedConst<ApiType>;
```
thesan: Hey @ignazio-bovo this last commit removes the set era payout damping factor proposal types to add the update token pallet governance ones, is it a mistake ?

FYI It also makes the network tests fail (including the chain setup scenario) so I had to rebase the QN work on the previous commit.

runtime-modules/council/src/lib.rs
mnaamani: Prepared this https://github.com/ignazio-bovo/joystream/pull/14

types/src/augment/augment-api-consts.ts
mnaamani: Yeah the PR needs to be rebased on luxor and types regenerated.

runtime-modules/council/src/lib.rs
dobertRowneySr: Constraints:
- Zero is allowed
- Max percent, currently 100%, potentially to increasing to above 100%

dobertRowneySr: We want this to be above 100% :(

kdembler:
Let's allow the multiplier to be 0.1-2.0. Mokhtar suggests to rename from damping to just multiplier

ignazio-bovo:
add the init default value

runtime-modules/council/src/lib.rs
mnaamani: @ignazio-bovo suggested this might be simpler than doing a migration

```suggestion
        pub EraPayoutDampingFactor get(fn era_payout_damping_factor) config(): Percent = Percent::from_percent(100);
```

mnaamani: @ignazio-bovo suggested this might be simpler than doing a migration

tests/network-tests/src/misc/postRuntimeUpgrade.ts
```diff
@@ -64,5 +64,9 @@ export default async function assertValues({ api }: FlowProps): Promise<void> {
   // assert.deepEqual(curatorGroup.permissionsByLevel, expectedPermissions)
   assert.equal(`${curatorGroup.permissionsByLevel}`, `${expectedPermissions}`)
 
+  // Damping factor after runtime should be 100% = default
```
mnaamani, line 15: Some of the previous assertions in this function need to be removed or fixed, they were for ephesus -> nara upgrade. As well as the setup in the `preRuntimeUpgrade.ts`

mnaamani (changes requested):

runtime/src/proposals_configuration/staging.rs
```diff
@@ -372,6 +372,21 @@ pub(crate) fn update_token_pallet_token_governance_parameters(
     }
 }
 
+// TODO (luxor/reward-curve-proposal): adjust parameters
+// Proposal parameters for the 'Set Era Payout Damping Factor' proposal
+pub(crate) fn set_era_payout_damping_factor() -> ProposalParameters<BlockNumber, Balance> {
+    ProposalParameters {
+        voting_period: days!(3),
```
mnaamani: `days` macro is not imported. But I think for staging profile we probably want minutes here.

mnaamani (approved):
Ok LGTM.
Changes to runtime-upgrade integration test do not look complete and they will be addressed separately as per https://github.com/Joystream/joystream/issues/5109

PR #5091, opened 2024-03-04
Add a `SKIP_ORION` option to the start script

thesan:
This is just a DX thing but I keep on adding this to the start script.

Here's why: only Atlas usues Orion, and a lot of the work on Atlas is done using a playgrounds ATM. OTOH when working on Orion directly IMO it's easier to run Orion from the local repository not from the monorepos.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -35,11 +35,14 @@ fi
 ## Query Node Infrastructure
 ./query-node/start.sh
 
-## Orion
-./start-orion.sh
+if [ "${SKIP_ORION}" != true ]
+then
+  ## Orion
+  ./start-orion.sh
 
-## Storage Squid
-docker-compose -f ./docker-compose.storage-squid.yml up -d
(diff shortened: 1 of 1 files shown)


start.sh
```diff
@@ -35,11 +35,14 @@ fi
...
 
-## Orion
-./start-orion.sh
+if [ "${SKIP_ORION}" != true ]
+then
+  ## Orion
+  ./start-orion.sh
 
-## Storage Squid
-docker-compose -f ./docker-compose.storage-squid.yml up -d
```
mnaamani, line 42: I don't think you mean to also prevent storage-squid from starting up if SKIP_ORION == true
The storage infra needs it to function, as they do not rely on query node anymore.

mnaamani (changes requested):
This makes sense, I find myself doing the same as well.
But we still need storage-squid to run ..

Issue #5081, opened 2024-02-22
[Colossus] uploads failing


kdembler:
This is investigation of upload failures we've been getting reports about lately.

From user reports, I have heard that they had uploads just abort mid-flight. Users have reported seeing CORS issues in the browser logs.

From my testing, I have had a case when the upload failed mid-flight and I got 408 response from the reverse proxy. This would also result in a CORS error as mentioned above.

yasiryagi:
I had test this and looked in the logs.

Affected storage providers: all
![image](https://github.com/Joystream/joystream/assets/4862448/888a0ee1-e0ac-4160-ac0e-6efe5385c416)

Geo location: all
![image](https://github.com/Joystream/joystream/assets/4862448/702394b6-5c2e-4787-aca1-fd9b71595512)

Client type: all


Observed when this issue happens the reverse proxy receives a reset from the peer i.e. Colossus and respond to the client with 408.

```
{"level":"error","ts":1708069037.667749,"logger":"http.handlers.reverse_proxy","msg":"reading from backend","error":"read tcp 172.21.0.7:33060->172.21.0.15:3333: read: connection reset by peer"}
{"level":"error","ts":1708069037.6678162,"logger":"http.handlers.reverse_proxy","msg":"aborting with incomplete response"," #error":"read tcp 172.21.0.7:33060->172.21.0.15:3333: read: connection reset by peer"}
```


Below is samples of the error 

This error is the majority, sometimes a company by other errors (included the StorageObject affected)


```
dataObjectId: 2100687
Internal Server Error: Upload aborted
    at error (/joystream/node_modules/express-openapi-validator/dist/middlewares/openapi.multipart.js:123:20)
    at /joystream/node_modules/express-openapi-validator/dist/middlewares/openapi.multipart.js:16:26
    at Immediate._onImmediate (/joystream/node_modules/multer/lib/make-middleware.js:53:37)
    at process.processImmediate (node:internal/timers:478:21)
```


```
dataObjectId: 1321165
uncaughtException: Upload aborted
Internal Server Error: Upload aborted
    at error (/joystream/node_modules/express-openapi-validator/dist/middlewares/openapi.multipart.js:123:20)
    at /joystream/node_modules/express-openapi-validator/dist/middlewares/openapi.multipart.js:16:26
    at Immediate._onImmediate (/joystream/node_modules/multer/lib/make-middleware.js:53:37)
    at process.processImmediate (node:internal/timers:478:21)

```

[...]

PR #5094, opened 2024-03-09
[Colossus] fix: upload aborted error

zeeshanakram3:
addresses #5081

- Updates `@types/node` dependency version to `18.6.0`
- Fix colossus upload aborted error

# Context

There is a [server.requestTimeout](https://nodejs.org/dist/latest-v18.x/docs/api/http.html#serverrequesttimeout) configuration option in the `http.Server` instance (remember `express` is just a wrapper around nodejs native HTTP Server), from the nodejs official documentation what this option means:

> Sets the timeout value in milliseconds for receiving the entire request from the client.
>
> If the timeout expires, the server responds with status 408 without forwarding the request to the request listener and then closes the connection.

The default value for the option is 300s for node versions `>=18.x`, while for the previous version the default value was 0 (no timeout)

So, the uploads failing problem started to occur when we updated the Nodejs from `v14` to `v18.6.0` in #4778

This timeout value of 300s also validates Ignazio's investigation that Colossus aborts errors after ~5 minutes.

### DIFF
--- a/storage-node/src/commands/server.ts
+++ b/storage-node/src/commands/server.ts
@@ -350,7 +350,11 @@ Supported values: warn, error, debug, info. Default:debug`,
         },
         x_host_id: X_HOST_ID,
       })
-      app.listen(port, () => logger.info(`Listening on http://localhost:${port}`))
+      const server = app.listen(port, () => logger.info(`Listening on http://localhost:${port}`))
+
+      // INFO: https://nodejs.org/dist/latest-v18.x/docs/api/http.html#serverrequesttimeout
+      // Set the server request timeout to 0 to disable it. This was default behaviour pre Node.js 18.x
+      server.requestTimeout = 0
     } catch (err) {
       logger.error(`Server error: ${err}`)
       this.exit(ExitCodes.ServerError)
--- a/cli/package.json
+++ b/cli/package.json
@@ -68,7 +68,7 @@
     "@oclif/test": "^1.2.5",
     "@types/chai": "^4.2.11",
     "@types/mocha": "^5.2.7",
-    "@types/node": "^10.17.18",
+    "@types/node": "^18.6.0",
     "chai": "^4.2.0",
     "eslint": "^7.6.0",
     "eslint-config-oclif": "^3.1.0",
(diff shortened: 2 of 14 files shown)


storage-node/src/commands/server.ts
```diff
@@ -350,7 +350,11 @@ Supported values: warn, error, debug, info. Default:debug`,
         },
         x_host_id: X_HOST_ID,
       })
-      app.listen(port, () => logger.info(`Listening on http://localhost:${port}`))
+      const server = app.listen(port, () => logger.info(`Listening on http://localhost:${port}`))
+
+      // INFO: https://nodejs.org/dist/latest-v18.x/docs/api/http.html#serverrequesttimeout
+      // Set the server request timeout to 0 to disable it. This was default behaviour pre Node.js 18.x
+      server.requestTimeout = 0
```
mnaamani, line 357: I think it makes sense to have the timeout disabled, provided the node is always behind a reverse proxy and not exposed on a public interface.

Alternatively we can have a non zero timeout, but larger value of say 20min, better than default of 5min?

mnaamani (changes requested):
Great that you found this settings. I suppose we should have looked more closely at changes between v14 and v18 that would have impacted our applications. Let keep that in mind when we shift to v20+

If we want a quick deployment of this fix because we are seeing many failed uploads this is good to go. Just need to bump colossus version.

storage-node/src/commands/server.ts
mnaamani: Or disabled timeout but make sure docs and our default deployment configs do not expose public interface.. ?

dobertRowneySr: My initial idea was to set a `env` var as a parameter named `MAX_REQUEST_TIMEOUT = 20 * 60 * 1000`  (20 mins in milliseconds) so that it can be set by the user.

dobertRowneySr:
The fix works in the sense that the above observed error is gone and the request time is in my case 6.4 mins > 5mins

dobertRowneySr (approved):
LGTM, I have also tested the fix locally

mnaamani (approved):

PR #5122, opened 2024-04-01
Add proxy pallet

mnaamani:
- [x] Add the substrate proxy pallet to joystream runtime
- [x] Temporary weight generation for pallet
- [x] Generate chain metadata


runtime/src/lib.rs
```diff
@@ -1812,6 +1812,92 @@ impl pallet_multisig::Config for Runtime {
...
+    WorkingGroups,
+}
+impl Default for ProxyType {
+    fn default() -> Self {
+        Self::Any
+    }
+}
+impl InstanceFilter<RuntimeCall> for ProxyType {
+    fn filter(&self, c: &RuntimeCall) -> bool {
+        match self {
```
kdembler, line 1873: Could we add a `Referendum` proxy type?
It would allow calling only:
- `referendum.vote`
- `referendum.revealVote`
- `referendum.releaseVoteStake`
The `Governance` type is a bit more extensive

kdembler (approved):
LGTM

runtime/src/lib.rs
mnaamani: Added in [2a2304b](https://github.com/Joystream/joystream/pull/5122/commits/2a2304b377044d531262561b4d9402d143ab1e18)

PR #5125, opened 2024-04-04
fix: :bug: amm and revenue shares cannot be concurrent

dobertRowneySr:
### Goals
1. Amm cannot be activate when a revenue share is active
2. Revenue Share cannot be issued when Amm is active

### Changes
- Added unit test for constraint 1
- Added unit test for constraint 2
- Added logic to the code to satisfy constraints above
- Added extra test to highlight that a token sale can still be issued during revenue share
- regenerate metadata and types

### Notes
- No migrations needed to accommodate changes


dobertRowneySr:
Update also the preconditions comment...

runtime-modules/project-token/src/lib.rs
```diff
@@ -1500,6 +1500,11 @@ impl<T: Config>
         let token_info = Self::ensure_token_exists(token_id)?;
         token_info.revenue_split.ensure_inactive::<T>()?;
 
+        ensure!(
```
mnaamani, line 1503: please add this precondition to the code comments

runtime-modules/project-token/src/lib.rs
```diff
@@ -1654,6 +1659,8 @@ impl<T: Config>
 
         let token_data = Self::ensure_token_exists(token_id)?;
 
+        token_data.ensure_can_modify_supply::<T>()?;
```
mnaamani, line 1662: Please add this pre-condition to the code comments as well.

mnaamani (changes requested):
Minor update to comments, otherwise looks good to go.

mnaamani:
> Update also the preconditions comment...

did you forget to push this change maybe?

mnaamani:
This solution works but "it would force creators to open and close their AMMs frequently which is not a good idea"
So replaced by https://github.com/Joystream/joystream/pull/5127

PR #5127, opened 2024-04-05
Amm buy/sell operations not allowed during revenue split

mnaamani:
Alternative to https://github.com/Joystream/joystream/pull/5125
AMM and Revenue split can both be active at the same time, but during revenue split the token supply cannot be modified by amm buy/sell (just as burning tokens is not permitted during revenue split)

cc: @WRadoslaw


dobertRowneySr (approved):
LGTM

Issue #5113, opened 2024-03-25
Add QN support for the validator rewards proposal


thesan:
Runtime issue: #3702
Runtime PR: #5086

PR #5129, opened 2024-04-08
[QN] Set era payout damping factor proposal mappings

thesan:
Address #5113

(Previous PR: https://github.com/ignazio-bovo/joystream/pull/13)


kdembler (approved):
LGTM

Issue #4981, opened 2023-11-29
colossus: add s3 compatible object storage backend


mnaamani:
### Background

With the explosive growth in demand on the storage infrastructure, and as suggested on multiple occasions by Storage lead, scaling the storage capacity of the storage node can be achieved by storing data objects on an object store built on top of clustering technology which allows dynamic growth of the capacity with limited disruption. eg. AWS [S3](https://aws.amazon.com/pm/serv-s3/), [Ceph](https://ceph.com/en/) cluster + Object Storage, S3 compatible object stores from other cloud providers

### Proposal - add support to store objects in object store.

- Choose between api/interface to use object store
   - Option A. Colossus will write into store using s3 HTTP API 
   - Option B. Consider mounting s3 bucket with fuse fs as a local volume with https://github.com/s3fs-fuse/s3fs-fuse. Does that work for non AWS s3 buckets?
- When clients (distributor nodes and other storage nodes) request an object colossus will do an HTTP redirect to an http url to fetch the stored object from the object store directly. (If we use the fuse fs how can we map to such a url?) - Alternatively colossus would have to fetch object from the object storage (perhaps cache it in local fs as well) and then serve it directly. This is less complicated and requires very little to no changes but at cost of more bandwidth consumption by the operator.
- HTTP HEAD requests should be redirected the same way. If this is not possible then colossus needs to keep local info about objects stored in object store to respond to HEAD requests?
- Update argus and colossus to handle HTTP redirects when fetching object from remote storage-nodes.

### Notes
- We cannot store temp files in s3 when syncing or accepting an upload see: https://github.com/Joystream/joystream/issues/4977 (maybe only when mounting with fuse fs?)
- Pick an s3 client package and test against multiple cloud providers.
- Do we want to support more than one object store at a time?
- If node can have multiple storage paths to help during migrating between volumes, we can it move objects to new storage path, new objects are added to new store and still be served from old store during migration.

zeeshanakram3:
> Pick an s3 client package and test against multiple cloud providers
> Do we want to support more than one object store at a time?

@mnaamani  there is a an NPM https://github.com/pkgcloud/pkgcloud#storage package that provides a unified interface to _all/most_ of the obejct storage colud services, Maybe we can look into this and see if it meets the requirements

mnaamani:
Adding links that might be useful for testing/development:

https://ytykhonchuk.medium.com/mock-amazon-s3-bucket-for-local-development-889440f9618e
[...]

PR #5150, opened 2024-05-16
colossus s3 api

dobertRowneySr:
The implementation is the same as suggested here:
#4981 (last comment by Mokhtar)

Instead of utilising a package in order to handle different types of cloud provider I have created a provider agnostic interface which integrates in the colossus logic

I believe this will allow us to easily add support for other providers in the future, by using their official sdk.
This also was useful in order to keep the changes to existing colossus logic to a minimal level (as the logic is already complex by itself).

The files `localCache` has been enhanced in order to keep track of which files are on the bucket and which files are on the local volume.
Only "accepted" data objects are stored on the bucket


.env
```diff
@@ -186,3 +186,18 @@ SQUID_GQL_PORT=4352
...
 # SQUID_ARCHIVE_GATEWAY_URL=${CUSTOM_ARCHIVE_GATEWAY_URL:-https://v2.archive.subsquid.io/network/joystream}
+
+# =====================================================================================
+## Cloud storage provider configuration
+
+# Enable or disable the cloud storage provider
+ENABLE_STORAGE_PROVIDER=true
+
+## Specify the cloud storage provider to use:
+CLOUD_STORAGE_PROVIDER_NAME=aws
```
mnaamani, line 197: `ENABLE_STORAGE_PROVIDER` and `CLOUD_STORAGE_PROVIDER_NAME` probably make more sense as command line arguments.

storage-node/package.json
```diff
@@ -167,6 +172,7 @@
     "ensure": "yarn format && yarn lint --fix && yarn build",
     "checks": "tsc --noEmit --pretty && prettier ./src --check && yarn lint",
     "start": "./bin/run server",
+    "test:integration:cloudProvider": "jest --detectOpenHandles './src/services/storageProviders/tests/**/syncService.test.ts'",
```
mnaamani, line 175: running the test I got one failed test:

```
Mokhtars-MBP:storage-node mokhtar$ yarn test:integration:cloudProvider
yarn run v1.22.19
$ jest --detectOpenHandles './src/services/storageProviders/tests/**/syncService.test.ts'
  Invalid testPattern ./src/services/storageProviders/tests/**/syncService.test.ts supplied. Running all tests instead.
  Invalid testPattern ./src/services/storageProviders/tests/**/syncService.test.ts supplied. Running all tests instead.
 PASS  src/services/storageProviders/tests/integration/setupLocalCache.test.ts (13.784 s)
 PASS  src/services/storageProviders/tests/integration/acceptObject.test.ts
 FAIL  src/services/storageProviders/tests/integration/acceptObjectService.test.ts
  ● accept pending object flow test suite › accept pending object happy path › should call accept objects/maxTxBatchSize times

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 2
    Received number of calls: 0

      70 |
      71 |       expect(acceptPendingDataObjectsBatch).toHaveBeenCalledTimes(2)
    > 72 |       expect(acceptObject).toHaveBeenCalledTimes(2)
         |                            ^
      73 |       expect(addDataObjectIdToCache).toHaveBeenCalledTimes(2)
      74 |     })
      75 |   })

      at Object.<anonymous> (src/services/storageProviders/tests/integration/acceptObjectService.test.ts:72:28)

Test Suites: 1 failed, 2 passed, 3 total
Tests:       1 failed, 23 passed, 24 total
```

.env
```diff
@@ -186,3 +186,18 @@ SQUID_GQL_PORT=4352
...
+
+# AWS S3 (replace with your own configuration and credentials)
+AWS_ACCESS_KEY_ID=test-key
+AWS_SECRET_ACCESS_KEY=test-secret
```
mnaamani:
```suggestion
AWS_ACCESS_KEY_ID=test
AWS_SECRET_ACCESS_KEY=test
```

To enable the creation of pre-signed URLs for S3 buckets, please set both `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to the value `“test”` Our pre-signed URL signature verification algorithm validates the pre-signed URL and its expiration. - note from https://docs.localstack.cloud/user-guide/integrations/aws-cli/#localstack-aws-cli-awslocal

docker-compose.localstack.yml
```diff
@@ -0,0 +1,26 @@
...
+    container_name: local_stack
+    image: localstack/localstack
+    network_mode: bridge
+    environment:
+      - SERVICES=s3 # we only need s3 bucket
+      - DEBUG=1
+      - DEFAULT_REGION=${AWS_REGION}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+      - LOCALSTACK_ENDPOINT=${LOCALSTACK_ENDPOINT}
```
mnaamani: Looking at https://docs.localstack.cloud/references/configuration/
default_region is deprecated, there are no key configuration, localstack just ignores secret and key ids for now.

tests/network-tests/run-tests.sh
```diff
@@ -7,24 +7,26 @@ cd $SCRIPT_PATH
...
+  docker logs colossus-1 --tail 100 || :
+  echo "# Colossus-2 Logs"
+  docker logs colossus-2 --tail 100 || :
+
+  if [ "${NO_STORAGE}" != true ]; then
+    docker-compose -f ../../docker-compose.storage-squid.yml down -v
+  fi
+
+  docker-compose -f ../../docker-compose.yml down -v
+  docker-compose -f ../../docker-compose.localstack.yml down -v
```
mnaamani:
```suggestion
    docker-compose -f ../../docker-compose.storage-squid.yml down -v
    docker-compose -f ../../docker-compose.localstack.yml down -v
  fi

  docker-compose -f ../../docker-compose.yml down -v
```

We can only stop localstack if we started it.

tests/network-tests/run-tests.sh
```diff
@@ -7,24 +7,26 @@ cd $SCRIPT_PATH
...
+}
 
 trap cleanup EXIT ERR SIGINT SIGTERM
 
-export JOYSTREAM_NODE_TAG=`RUNTIME_PROFILE=TESTING ../../scripts/runtime-code-shasum.sh`
+export JOYSTREAM_NODE_TAG=${JOYSTREAM_NODE_TAG:-latest}
+RUNTIME_PROFILE=TESTING ../../scripts/runtime-code-shasum.sh
```
mnaamani: The refactoring here is not correct. The integration tests will not work with production runtime, so this should be reverted.
This is why the CI checks running tests are failing?

What was the intention here?

tests/network-tests/start-storage.sh
```diff
@@ -1,18 +1,22 @@
 set -e
 
+source .env
+
 TMP=$0
-THIS_DIR=`dirname $TMP`
+THIS_DIR=$(dirname $TMP)
```
mnaamani: The back tick is better, so the script exits with error if the invoked command errors. 
With the `$()` any errors are hidden and the script continues, with the variable assiged being empty, which most times is not the behavior we want.

tests/network-tests/start-storage.sh
```diff
@@ -24,6 +28,10 @@ docker-compose -f $THIS_DIR/../../docker-compose.yml up -d distributor-1
 docker-compose -f $THIS_DIR/../../docker-compose.yml up -d colossus-2
 docker-compose -f $THIS_DIR/../../docker-compose.yml up -d distributor-2
 
+# Start localstack if ENABLE_LOCALSTACK is set to true
+docker-compose -f $THIS_DIR/../../docker-compose.localstack.yml up -d localstack
+awslocal s3api create-bucket --bucket $AWS_BUCKET_NAME --endpoint http://localhost:4566
```
mnaamani: We probably want to confirm bucket creation is successful, perhaps localstack didn't come up successfully.

Currently it does indeed fail `aws: error: argument --bucket: expected one argument` if the `AWS_BUCKET_NAME` env variable is not set. It is not loaded from .env file for example and is also the case in the github workflow `run-network-tests.yml` which is an additional reason why the checks would be failing.

awslocal is most likely not installed on the machine, so we need to either install it as part of setup.sh, at least in the github workflow. Other approach, since we have aws-sdk npm package installed maybe have a simple typescript script to create the bucket.

storage and distributor nodes should be started after localstack, not before.

Single bucket is not enough, the integration tests run two storage-nodes so we want to have two separate buckets one for each node.

docker-compose.localstack.yml
```diff
@@ -0,0 +1,26 @@
+version: '3.4'
+services:
+  localstack:
+    container_name: local_stack
+    image: localstack/localstack
```
mnaamani:
```suggestion
    image: localstack/localstack:3
```
Perhaps better to specify the major version number.

.env
```diff
@@ -186,3 +186,18 @@ SQUID_GQL_PORT=4352
...
+AWS_REGION=eu-west-1
+AWS_BUCKET_NAME=sample-bucket
```
mnaamani: One bucket name is not sufficient, since the tests and when running playgrounds we could have multiple storage nodes.

storage-node/src/commands/server.ts
```diff
@@ -22,8 +22,14 @@ import { getStorageBucketIdsByWorkerId } from '../services/sync/storageObligatio
 import { PendingDirName, TempDirName, performSync } from '../services/sync/synchronizer'
 import { createApp } from '../services/webApi/app'
 import ExitCodes from './../command-base/ExitCodes'
+import { IConnectionHandler, parseConfigOptionAndBuildConnection } from '../services/storageProviders'
 const fsPromises = fs.promises
 
+// Global variable for storage provider connection, initialised by server, then readonly
```
mnaamani, line 28: I don't think this is the best place to store these values.

storage-node/src/commands/util/types.ts
```diff
@@ -0,0 +1 @@
+export type Maybe<T> = T | undefined
```
mnaamani, line 1: I don't see this used anywhere.

cli/src/Api.ts
```diff
@@ -182,7 +182,7 @@ export default class Api {
     return new Date(blockTime.toNumber())
   }
 
-  protected workingGroupApiQuery<T extends WorkingGroups>(group: T): ApiPromise['query'][typeof apiModuleByGroup[T]] {
+  protected workingGroupApiQuery<T extends WorkingGroups>(group: T): ApiPromise['query'][(typeof apiModuleByGroup)[T]] {
```
mnaamani, line 185: Please checkout any code changes to cli, they all seem to be code style changes and best to be in a different PR.
Let keep changes in the PR focused on storage-node

storage-node/src/services/storageProviders/IConnectionHandler.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Represents an abstract connection handler for a storage provider.
+ */
+/**
+ * Represents a connection handler for interacting with a remote storage bucket.
+ */
```
mnaamani, line 16: There are two descriptions of the interface. 

I have to say if this is an abstract interface it has a lot of specific naming for s3 storage. The methods and their parameters have confusing and inconsistent names. A file is referred to as filename, key, object. The term RemoteBucket is used but not defined. Even the name of the interface could be improved.

storage-node/src/services/storageProviders/awsConnectionHandler.ts
```diff
@@ -0,0 +1,139 @@
...
+    const fileSizeInBytes = stats.size
+    const fileSizeInGigabytes = fileSizeInBytes / (1024 * 1024 * 1024)
+    return fileSizeInGigabytes > this.multiPartThresholdGB
+  }
+
+  async uploadFileToRemoteBucket(filename: string, filePath: string): Promise<any> {
+    const input = {
+      Bucket: this.bucket,
+      Key: filename,
+      Body: filePath,
```
mnaamani: You are sending the file path string as the "content" of the object. That is not correct.

In my testing I found the storage node uploaded data object with content "/data/pending/0" as confirmation.

According to the AWS sdk docs the [Body](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/PutObjectCommandInput/) needs to be a `StreamingBlobPayloadInputTypes`.  In [nodejs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-types/TypeAlias/NodeJsRuntimeStreamingBlobPayloadInputTypes/) this is `string | Uint8Array | Buffer | Readable` 

Ideally we want to use a readable stream, but the doc says "in Node.js a Readable stream is not signable by the default signer." I'm not sure if that is a problem. Otherwise we have to fallback to using `Buffer`, but that would mean loading entire files into memory which is definitively a bad idea.

storage-node/src/services/storageProviders/awsConnectionHandler.ts
```diff
@@ -0,0 +1,139 @@
...
+      Key: filename,
+    }
+
+    const command = new GetObjectCommand(input)
+    return await getSignedUrl(this.client, command, { expiresIn: 60 * 60 })
+  }
+
+  async listFilesOnRemoteBucket(): Promise<string[]> {
+    let listingComplete = false
+    let input: ListObjectsCommandInput = {
```
mnaamani, line 148: Are we sure we ignore folders, incase there are folders on the bucket?

storage-node/src/services/storageProviders/awsConnectionHandler.ts
```diff
@@ -0,0 +1,139 @@
...
+    // the listing is paginated so we need to keep track of the marker
+    while (!listingComplete) {
+      const response = await this.client.send(new ListObjectsCommand(input))
+      if (!this.isSuccessfulResponse(response)) {
+        throw new Error('Response unsuccessful when listing files in S3 bucket')
+      }
+      if (!response.Contents) {
+        throw new Error('Response contents are undefined when listing files in S3 bucket, bucket possibly empty')
+      }
+      files.push(...response.Contents.filter((file) => file.Key).map((file) => file.Key!))
```
mnaamani, line 163: Just a reminder
Array.push(...array) is okay as long as array size is small. Otherwise we could get stack overflow.
Seeing the default max keys in response is 1000 we are okay.

storage-node/src/services/storageProviders/mockConnectionHandler.ts
```diff
@@ -0,0 +1,22 @@
+// create a mock implementation of IConnectionHandler
```
mnaamani, line 1: The mock class could be useful, but is not used in the tests. Drop it?

storage-node/src/services/helpers/acceptObject.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+    await moveFile(src, dest)
+    logger.info(`File ${filename} accepted on local volume`)
+  } else {
+    const connection = getStorageProviderConnection()!
+    await connection.uploadFileToRemoteBucket(filename, src.toString())
+    await fsPromises.unlink(src)
+    logger.info(`File ${filename} accepted to remote storage`)
+  }
+  addDataObjectIdToCache(filename)
```
mnaamani: Missing the second argument to indicate where object is stored.
```suggestion
  addDataObjectIdToCache(filename, toBeAcceptedOnLocalVolume)
```

storage-node/src/services/storageProviders/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+dotenv.config()
+
+/**
+ * Parses the configuration options and builds a connection handler for the storage provider.
+ *
+ * @param enableProvider - A boolean indicating whether a provider is expected to be enabled.
+ * @returns A promise that resolves to an instance of IConnectionHandler if the provider is enabled, otherwise undefined.
+ * @throws An error if the storage cloud provider type is invalid or unsupported.
+ */
+export async function parseConfigOptionAndBuildConnection(): Promise<IConnectionHandler | undefined> {
```
mnaamani: This does not need to be an async function.

storage-node/src/services/storageProviders/IConnectionHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface IConnectionHandler {
+  /**
+   * Asynchronously uploads a file to the remote bucket.
+   * @param filename - The key of the file in the remote bucket.
+   * @param filePath - The file path of the file to upload.
+   * @returns A promise that resolves when the upload is complete or rejects with an error.
+   */
+  uploadFileToRemoteBucket(filename: string, filePath: string): Promise<any>
```
mnaamani: We should define how to handle case when an key already exists on the store.

storage-node/src/services/webApi/controllers/filesApi.ts
```diff
@@ -42,38 +47,59 @@ export async function getFile(
   try {
     pinDataObjectIdToCache(dataObjectId)
   } catch (err) {
+    logger.error(`Error pinning file ${dataObjectId} to cache`)
     res.status(404).send()
     return
   }
 
-  const unpin = _.once(() => {
+  const unpin = _.once(async () => {
```
mnaamani:
```suggestion
  const unpin = _.once(() => {
```
Why are you making this async? The underlying call are not async.

tests/network-tests/start-storage.sh
```diff
@@ -1,18 +1,22 @@
 set -e
 
+source .env
+
 TMP=$0
-THIS_DIR=`dirname $TMP`
+THIS_DIR=$(dirname $TMP)
 
 echo "Staring storage infrastructure"
+export LOCALSTACK_ENABLED=true
```
mnaamani: LOCALSTACK_ENABLED is not passed to colossus-1 and colossus-2 in the docker-compose.yml file so they are not using the localstack, and instead public aws. So they fail to access s3 bucket.

storage-node/src/services/helpers/acceptObject.ts
mnaamani: For both moveFile and uploadToRemoteBucket should we explicitly handle case if destination object already exists?

mnaamani (changes requested):
This is a good start.

The awsConnectionHandler implementation needs to be fixed to properly do uploading to the s3 storage.

The integration bash scripts and docker-compose files to run tests need some changes to get them working. I have made local changes on my machine to get tests working. I can open a PR to your branch if you like.

One jest unit test is failing.

Some limitations I see in the current implementation is that only a single object store can be configured at a time. I think that is fine since object stores are more easily scalable and operator should not need to frequently change which bucket to use. It might be practical from an organizational point of view perhaps to be able to specify a subfolder in the s3 bucket to use for storing the dataobjects.

.env
mnaamani: Also we can reduce from two params to just one, the provider name. If not empty string it means we want to use a cloud storage provider.

docker-compose.localstack.yml
mnaamani: And here I think you meant to configure LOCALSTACK_HOST. LOCALSTACK_ENDPOINT is not a configuration option for localstack anymore

tests/network-tests/run-tests.sh
dobertRowneySr: I was not working on my machine

PR #5155, opened 2024-05-28
Runtime: Argo Bridge

freakstatic:
Still missing the benchmark tests


runtime-modules/argo-bridge/src/Untitled-2
```diff
@@ -0,0 +1,6 @@
+
```
kdembler: This file shouldn't be committed

runtime-modules/argo-bridge/src/errors.rs
```diff
@@ -0,0 +1,35 @@
...
+        BridgeNotActive,
+
+        /// Insufficient JOY Balance to cover the transaction costs
+        InsufficientJoyBalance,
+
+        /// The bridging_fee changed since the request transfer
+        FeeDifferentThanExpected,
+
+        /// Not enough mint allowance for transaction
+        InsufficienBridgMintAllowance,
```
kdembler: 1. `t` missing in `Insufficient`
2. `e` missing in `Bridge`

runtime-modules/argo-bridge/src/events.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub enum Event<T>
+    where
+        AccountId = <T as frame_system::Config>::AccountId,
+        Balance = BalanceOf<T>,
+        BridgeConstraints = BridgeConstraintsOf<T>,
+    {
+        OutboundTransferRequested(TransferId, AccountId, RemoteAccount, Balance, Balance),
+        InboundTransferFinalized(RemoteTransfer, AccountId, Balance),
+        BridgePaused(AccountId),
+        BridgeThawnStarted(AccountId),
```
kdembler: Let's include the thawn end block in the event

runtime-modules/argo-bridge/src/errors.rs
```diff
@@ -0,0 +1,35 @@
...
+        /// Operator account required
+        NotOperatorAccount,
+
+        /// Pauser account required
+        NotPauserAccount,
+
+        /// Number of pauser accounts over the maximum allowed
+        InvalidNumberOfPauserAccounts,
+
+        /// Current block is lower than thawn_started_at + thawn_duration
```
kdembler: Let's update the comment after change to `thawn_ends_at`

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    // /// Events
+    type RuntimeEvent: From<Event<Self>> + Into<<Self as frame_system::Config>::RuntimeEvent>;
+
+    type MaxPauserAccounts: Get<u32>;
+
+    /// Weight information for extrinsics in this pallet.
+    type WeightInfo: WeightInfo;
+
+    /// Defines the default bridging fee.
+    type DefaultBridgingFee: Get<BalanceOf<Self>>;
```
kdembler, line 59: The `production_config` in `chain_spec/argo_bridge_config.rs` contains a value for default bridging fee, are those 2 connected?

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        #[weight = WeightInfoArgo::<T>::finalize_inbound_transfer()]
+        pub fn finalize_inbound_transfer(origin, remote_transfer: RemoteTransfer, dest_account: T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+            ensure!(!Self::operator_account().is_none(), Error::<T>::NotOperatorAccount);
+            let caller = ensure_signed(origin)?;
+            ensure!(caller == Self::operator_account().unwrap(), Error::<T>::NotOperatorAccount);
+
+            ensure!(Self::status() == BridgeStatus::Active, Error::<T>::BridgeNotActive);
+            ensure!(amount < Self::mint_allowance(), Error::<T>::InsufficienBridgMintAllowance);
```
kdembler: This should be `<=`

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+            account: [0; 32],
+            chain_id: 1,
+        };
+        let result = ArgoBridge::request_outbound_transfer(
+            RuntimeOrigin::signed(account!(1)),
+            remote_account,
+            joy!(1000),
+            joy!(fee),
+        );
+        assert_ok!(result);
```
kdembler, line 56: I'm not sure what kind of assertions we can do in our tests, but could this also assert:
1. Proper balance change in the caller account
2. Proper change of mint allowance?

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        ));
+
+        let remote_transfer = RemoteTransfer { id: 0, chain_id: 1 };
+        let result = ArgoBridge::finalize_inbound_transfer(
+            RuntimeOrigin::signed(account!(1)),
+            remote_transfer,
+            account!(2),
+            1000,
+        );
+        assert_err!(result, Error::<Test>::InsufficienBridgMintAllowance);
```
kdembler: Name of this test ends in `_success` but we assert error here. I think this should be the happy case? In that case, same questions as for the outbound transfer happy case - can we assert change in mint allowance and in token balances?

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+            pauser_accounts: Some(vec![account!(2)]),
+            bridging_fee: None,
+            thawn_duration: None,
+            remote_chains: None,
+        };
+        assert_ok!(ArgoBridge::update_bridge_constrains(
+            RuntimeOrigin::root(),
+            parameters
+        ));
+        assert_ok!(ArgoBridge::pause_bridge(RuntimeOrigin::signed(account!(2))));
```
kdembler: Can we assert proper updated status?

runtime/src/proposals_configuration/defaults.rs
```diff
@@ -358,6 +358,20 @@ pub(crate) fn update_token_pallet_token_governance_parameters(
     }
 }
 
+// Proposal parameters for the 'Update Argo Bridge Parameters' proposal
+pub(crate) fn update_argo_bridge_parameters() -> ProposalParameters<BlockNumber, Balance> {
+    ProposalParameters {
+        voting_period: days!(7),
+        grace_period: days!(1),
+        approval_quorum_percentage: TWO_OUT_OF_THREE,
```
kdembler: Let's make this 3/3.

kdembler (changes requested):

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+            let transfer_id = NextTransferId::get();
+            Self::deposit_event(RawEvent::OutboundTransferRequested(transfer_id, sender, dest_account, amount, fee));
+            NextTransferId::put(transfer_id + 1);
+
+            Ok(())
+        }
+
+        #[weight = WeightInfoArgo::<T>::finalize_inbound_transfer()]
+        pub fn finalize_inbound_transfer(origin, remote_transfer: RemoteTransfer, dest_account: T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+            ensure!(!Self::operator_account().is_none(), Error::<T>::NotOperatorAccount);
```
dobertRowneySr: Here I would use as error message `OperatorAccountNotSet` (provide that we need at some point the operator account to be `None`)

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Defines the default bridging fee.
+    type DefaultBridgingFee: Get<BalanceOf<Self>>;
+}
+
+decl_storage! { generate_storage_info
+    trait Store for Module<T: Config> as ArgoBridge {
+        pub Status get(fn status) config(): BridgeStatus<T::BlockNumber>;
+
+        /// Account ID that operates the bridge
+        pub OperatorAccount get(fn operator_account): Option<T::AccountId>;
```
dobertRowneySr, line 67: Does it make sense to have a bridge without operator account? If not I would say to remove the optionality. So you can later avoid an extra ensure check

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+            <Status<T>>::put(BridgeStatus::Thawn { thawn_ends_at: current_block + Self::thawn_duration()});
+            Self::deposit_event(RawEvent::BridgeThawnStarted(caller));
+
+            Ok(())
+        }
+
+        #[weight = WeightInfoArgo::<T>::finish_unpause_bridge()]
+        pub fn finish_unpause_bridge(origin) -> DispatchResult {
+            let caller = ensure_signed(origin)?;
+            ensure!(!Self::operator_account().is_none(), Error::<T>::NotOperatorAccount);
```
dobertRowneySr: Same as above I will use as error variant name `OperatorAccountNotSet`

runtime-modules/argo-bridge/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+pub type BridgeConstraintsOf<T> = BridgeConstraints<
+    <T as frame_system::Config>::AccountId,
+    BalanceOf<T>,
+    <T as frame_system::Config>::BlockNumber,
+>;
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, Debug, TypeInfo, MaxEncodedLen)]
+pub struct RemoteAccount {
+    // Ethereum addresses only need 20 bytes but we use 32 bytes to enable compatibility with other chains
```
dobertRowneySr, line 36:
```suggestion
    // EVM-based addresses only need 20 bytes but we use 32 bytes to enable compatibility with other chains
```

runtime-modules/argo-bridge/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+    // When using Ethereum addresses, the last 12 bytes should be zero
```
dobertRowneySr:
```suggestion
    // When using EVM-based addresses, the last 12 bytes should be zero
```

runtime-modules/argo-bridge/src/types.rs
dobertRowneySr: Phantom, Polygon etc.. all have 20 bytes long addresses

runtime-modules/argo-bridge/src/errors.rs
```diff
@@ -0,0 +1,35 @@
...
+        InsufficientJoyBalance,
+
+        /// The bridging_fee changed since the request transfer
+        FeeDifferentThanExpected,
+
+        /// Not enough mint allowance for transaction
+        InsufficienBridgMintAllowance,
+
+        /// Operator account required
+        NotOperatorAccount,
```
dobertRowneySr, line 21: As I said above I would add a `OperatorAccountNotSet` variant if really necessary here

runtime-modules/argo-bridge/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub account: [u8; 32],
```
dobertRowneySr, line 38: I would use `address` here and not `account`

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        let remote_chains = BoundedVec::try_from(vec![1u32]).unwrap();
+        let parameters = BridgeConstraints {
+            operator_account: Some(account!(1)),
+            pauser_accounts: Some(vec![account!(2), account!(3)]),
+            bridging_fee: Some(fee),
+            thawn_duration: None,
+            remote_chains: Some(remote_chains),
+        };
+        ArgoBridge::update_bridge_constrains(RuntimeOrigin::root(), parameters).unwrap();
+
```
dobertRowneySr, line 85: I will still add the line in order to update the sender balance. In case in a future refactoring we swap the line with the balance check and the bridge pausing check not to have this test failing

runtime-modules/argo-bridge/src/types.rs
dobertRowneySr: The most significant 12 bytes should be 0, this padding method is consistent with the way Solidity handles addresses and other data types, ensuring compatibility and consistency across the Ethereum ecosystem.
So the first 12 bytes should be 0 and not the last 12

dobertRowneySr: cc: @kdembler this decision might have consequence on how you parse events fields on the argo squid

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+            remote_chains: None,
+        };
+        ArgoBridge::update_bridge_constrains(RuntimeOrigin::root(), parameters).unwrap();
+
+        ArgoBridge::pause_bridge(RuntimeOrigin::signed(account!(2))).unwrap();
+
+        let result = ArgoBridge::init_unpause_bridge(RuntimeOrigin::signed(account!(1)));
+        assert_err!(result, Error::<Test>::NotPauserAccount);
+    });
+}
```
dobertRowneySr, line 461: There are the following error-path case missing:
`reqquest_outbound_transfer`:
  - remote chain not supported
`finish_pause_bridge`:
  - operator account is `None` case

runtime-modules/argo-bridge/src/tests/mod.rs
dobertRowneySr: Also proper event deposit (very important for the argo-squid)

runtime-modules/argo-bridge/src/tests/mod.rs
dobertRowneySr: also check proper event deposit

runtime-modules/argo-bridge/src/tests/mod.rs
dobertRowneySr: and event deposit

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        assert_eq!(
+            ArgoBridge::status(),
+            BridgeStatus::Thawn {
+                thawn_ends_at: current_block + 1
+            }
+        );
+        increase_block_number_by(2);
+        assert_ok!(ArgoBridge::finish_unpause_bridge(RuntimeOrigin::signed(
+            account!(1)
+        )));
```
dobertRowneySr, line 388: also check for event deposit

dobertRowneySr (commented):
The pallet looks solid, the only issues that I have found are:
- remote chain address formatting : 12 most significant bytes should be zero IMHO
- If you can provide a case where the operator account is `None`
The tests needs some touching

dobertRowneySr (changes requested):
The pallet looks solid, the only issues that I have found are:
- remote chain address formatting : 12 most significant bytes should be zero IMHO
- If you can provide a case where the operator account is `None`
The tests needs some touching, especially in the happy path assertions (especially on event production)
Also: wouldn't it be easier for you to have multiple commits 😅 , instead of a single huge one ?

dobertRowneySr (commented):
Ignore this, I accidentally re requested a review

bin/node/src/chain_spec/argo_bridge_config.rs
```diff
@@ -0,0 +1,21 @@
+use node_runtime::{argo_bridge::types::BridgeStatus, ArgoBridgeConfig};
+
+pub fn production_config() -> ArgoBridgeConfig {
```
dobertRowneySr, line 3: I also think that you can set the mint allowance storage variable to be `0` by default since I don't think any other initial value is plausible

runtime/src/lib.rs
```diff
@@ -139,6 +139,8 @@ pub use content;
 pub use content::LimitPerPeriod;
 pub use content::MaxNumber;
 
+pub use argo_bridge;
+
 /// This runtime version.
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
```
dobertRowneySr, line 146: petra should be spec version `2004`, correctly if I am wrong @kdembler

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        pub Status get(fn status) config(): BridgeStatus<T::BlockNumber>;
+
+        /// Account ID that operates the bridge
+        pub OperatorAccount get(fn operator_account): Option<T::AccountId>;
+
+        /// List of account IDs with permission to pause the bridge operations
+        pub PauserAccounts get(fn pauser_accounts): BoundedVec<T::AccountId, T::MaxPauserAccounts>;
+
+        /// Number of tokens that the bridge pallet is able to mint
+        pub MintAllowance get(fn mint_allowance) config(): BalanceOf<T>;
```
dobertRowneySr: I believe you can set default value of this to `0` if I correctly understood the bridging accounting logic, as there shouldn't be any other possible initial value

runtime/src/lib.rs
```diff
@@ -935,6 +937,18 @@ impl project_token::Config for Runtime {
     type WeightInfo = project_token::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+    pub const MaxPauserAccounts: u32 = 10;
+    pub const DefaultBridgingFee: Balance = dollars!(1_000);
```
dobertRowneySr: 😅 1k USD for bridging, shouldn't this be lower?

dobertRowneySr (changes requested):
Did another quick review round

freakstatic:
> The pallet looks solid, the only issues that I have found are:
> 
> * remote chain address formatting : 12 most significant bytes should be zero IMHO
> * If you can provide a case where the operator account is `None`
>   The tests needs some touching

Thanks for the review :+1: 
The operator account being `None` is because I wasn't sure which account should be the default one :thinking:

runtime-modules/argo-bridge/src/lib.rs
kdembler: Not sure what we would set in the initial pallet config in that case?

runtime-modules/argo-bridge/src/types.rs
kdembler: well they're still Ethereum addresses 😅

runtime-modules/argo-bridge/src/lib.rs
freakstatic: in the tests is handy to be able to set a value for the mint allowance in the config

freakstatic: or were you really just suggesting a default value and not removing it from the genesis config?

runtime-modules/argo-bridge/src/lib.rs
freakstatic: yes but I was missing the call to `DefaultBridgingFee::get()`

runtime-modules/argo-bridge/src/tests/mod.rs
freakstatic: > Name of this test ends in `_success` but we assert error here. I think this should be the happy case? In that case, same questions as for the outbound transfer happy case - can we assert change in mint allowance and in token balances?

the title was good but the test was wrong...
added checks for the balance and the event deposit

runtime-modules/argo-bridge/src/types.rs
kdembler: Yeah that's what I meant actually, my mistake writing spec. I suggest
```suggestion
    // When using EVM addresses, the 12 most-significant bytes should be zero
```
Squid already uses this approach

runtime-modules/argo-bridge/src/types.rs
kdembler: Yeah I spent a minute on this. My thinking was that this could also be a Substrate chain in the future and in that case it would actually be 32-byte AccountId, so account looked more universal, I don't mind either way

runtime-modules/argo-bridge/src/lib.rs
dobertRowneySr: Ok so the actual procedure for deployment will be:
1. Deploy petra with argo pallet
2. Setup JoyOp account signatories
3. Activate the bridge

If so Then that's ok and the operator account can be none

dobertRowneySr:
All my previous comments have been addressed. So for now LGTM
Should I wait for the benchmarks?

freakstatic:
@ignazio-bovo and @kdembler I have added the benchmark tests please take a look.
the weight calculation is not finding the benchmarks so I'm most likely missing something...

`./scripts/../target/debug/joystream-node benchmark pallet --pallet=argo_bridge --extrinsic="*" --chain=prod-test --steps=50 --repeat=20 --execution=wasm --template=./scripts/../devops/joystream-pallet-weight-template.hbs --output=./scripts/../runtime-modules/argo-bridge/src/weights.rs`

```
2024-05-30 05:28:07 [0] 💸 generated 1 npos voters, 1 from validators and 0 nominators    
2024-05-30 05:28:07 [0] 💸 generated 1 npos targets    
Error: Input("No benchmarks found which match your input.")
```

runtime/src/lib.rs
kdembler: @freakstatic please bump

runtime/src/lib.rs
```diff
@@ -1979,6 +1994,7 @@ construct_runtime!(
         Content: content::{Pallet, Call, Storage, Event<T>, Config<T>},
         Storage: storage::{Pallet, Call, Storage, Event<T>, Config<T>},
         ProjectToken: project_token::{Pallet, Call, Storage, Event<T>, Config<T>},
+        ArgoBridge: argo_bridge::{Pallet, Call, Storage, Event<T>, Config<T>},
```
kdembler: Let's try placing the ArgoBridge pallet last, after Proxy. This way we should not change existing modules indexes and won't need to bump `transaction_version` (only `spec_version`)

kdembler:
@ignazio-bovo @freakstatic I think the discussion on storage initialization is still not resolved. @ignazio-bovo could you expand on what you had in mind?

mnaamani:
> @ignazio-bovo and @kdembler I have added the benchmark tests please take a look. the weight calculation is not finding the benchmarks so I'm most likely missing something...
> 
> `./scripts/../target/debug/joystream-node benchmark pallet --pallet=argo_bridge --extrinsic="*" --chain=prod-test --steps=50 --repeat=20 --execution=wasm --template=./scripts/../devops/joystream-pallet-weight-template.hbs --output=./scripts/../runtime-modules/argo-bridge/src/weights.rs`
> 
> ```
> 2024-05-30 05:28:07 [0] 💸 generated 1 npos voters, 1 from validators and 0 nominators    
> 2024-05-30 05:28:07 [0] 💸 generated 1 npos targets    
> Error: Input("No benchmarks found which match your input.")
> ```

I think you need to add the argo bridge pallet to the `define_benchmarks!` macro in `runtime/src/runtime_api.rs`

```diff
diff --git a/runtime/src/runtime_api.rs b/runtime/src/runtime_api.rs
index 8d3a06316a..c57e968e51 100644
--- a/runtime/src/runtime_api.rs
+++ b/runtime/src/runtime_api.rs
@@ -127,6 +127,7 @@ mod benches {
         [content, Content]
         [project_token, ProjectToken]
         [pallet_proxy, Proxy]
+        [argo_bridge, ArgoBridge]
     );
 }
```

ignazio-bovo:
> @ignazio-bovo @freakstatic I think the discussion on storage initialization is still not resolved. @ignazio-bovo could you expand on what you had in mind?

Sure: 
Originally I though that when the argo pallet will be release in the next runtime upgrade, it would be activated immediately -> hence at upgrade time we will already have JoyOp account setup with proper multisig scheme and signatories. If you are thinking about this then the JoyOp account is established before the argo pallet is released and it can be included in the initial chainspec.

If our #1 goal is to release the runtime upgrade first in order to have the argo on runtime and activate the bridge any time later then and keep it paused in the meantime then it makes sense to allow the JoyOp account to be `None` and set it later on before setting the bridge status to active

runtime-modules/argo-bridge/src/lib.rs
dobertRowneySr: the storage value mint_allowance is increased during Outbound transfer and decreased during Inbound transfer and it represent the total amount of JOY we have outside of the Joystream mainnet network.
If so my observation was: since the only plausible initial value is zero, we can set the default value when declearing the storage variable inside the decl_storage macro instead of manually setting the value in the chainspec config. (Even though I remember with confidence that in case a storage variable is not initialised it will be set to the fault value when accessed for the first time, so 0 for mint allowance which in production will be a u128)

dobertRowneySr: this is a minor comment however, I am fine even with the existing solution

freakstatic:
> > @ignazio-bovo and @kdembler I have added the benchmark tests please take a look. the weight calculation is not finding the benchmarks so I'm most likely missing something...
> > `./scripts/../target/debug/joystream-node benchmark pallet --pallet=argo_bridge --extrinsic="*" --chain=prod-test --steps=50 --repeat=20 --execution=wasm --template=./scripts/../devops/joystream-pallet-weight-template.hbs --output=./scripts/../runtime-modules/argo-bridge/src/weights.rs`
> > ```
> > 2024-05-30 05:28:07 [0] 💸 generated 1 npos voters, 1 from validators and 0 nominators    
> > 2024-05-30 05:28:07 [0] 💸 generated 1 npos targets    
> > Error: Input("No benchmarks found which match your input.")
> > ```
> 
> I think you need to add the argo bridge pallet to the `define_benchmarks!` macro in `runtime/src/runtime_api.rs`
> 
> ```diff
> diff --git a/runtime/src/runtime_api.rs b/runtime/src/runtime_api.rs
> index 8d3a06316a..c57e968e51 100644
> --- a/runtime/src/runtime_api.rs
> +++ b/runtime/src/runtime_api.rs
> @@ -127,6 +127,7 @@ mod benches {
>          [content, Content]
>          [project_token, ProjectToken]
>          [pallet_proxy, Proxy]
> +        [argo_bridge, ArgoBridge]
>      );
>  }
> ```

thanks very much 🙏 
yeah that was exactly what I was missing

freakstatic:
@ignazio-bovo and @kdembler  can you please review the benchmarking tests?

runtime-modules/proposals/codex/src/benchmarking.rs
```diff
@@ -957,6 +957,38 @@ benchmarks! {
...
     }
 
+    create_proposal_argo_bridge_constraints {
+        let t in 1 .. to_kb(T::TitleMaxLength::get());
+        let d in 1 .. to_kb(T::DescriptionMaxLength::get());
+
+        let (account_id, member_id, general_proposal_paramters) =
+            create_proposal_parameters::<T>(t, d);
+
+        // let member_id = Membership::<T>::members_created();
```
freakstatic: this is still WIP

dobertRowneySr (approved):
Benchmarks looks ok in terms of worst case scenario accounting

dobertRowneySr (changes requested):
There are some errors that needs to be fixed (see CI).
You can test the benchmarking code with cargo test -p "pallet-argo-bridge" --features "runtime-benchmarks" (you can also use clippy with the same feature enabled)

kdembler:
@freakstatic @ignazio-bovo

I opened a PR with fixes for clippy issues that we should merge into this PR: https://github.com/freakstatic/joystream/pull/2

I have also opened a new PR based on this to add a revert extrinsic: https://github.com/Joystream/joystream/pull/5158

Please check

kdembler (approved):
LGTM pending benchmark&weights approval from @ignazio-bovo

We also need to add overflow protection in some extrinsics but will do in a subsequent PR

dobertRowneySr (approved):
LGTM, the mac osx checks will be fixed in my PR here: #5157

PR #5158, opened 2024-06-02
add Argo revert_outbound_transfer extrinsic

kdembler:
Based on #5155

Adds a new `revert_outbound_transfer` extrinsic. Functionally, the new extrinsic is almost the same as `finalize_inbound_transfer`:
1. It verifies that the caller is the operator
2. It checks mint allowance
3. It mints new tokens
4. 
The only difference is in input arguments and the emitted event.

Todo:

- [ ] Add proper benchmarks and weights

### DIFF
--- a/runtime-modules/argo-bridge/src/lib.rs
+++ b/runtime-modules/argo-bridge/src/lib.rs
@@ -137,6 +137,33 @@ decl_module! {
             Ok(())
         }
 
+        // TODO: add weight for revert_outbound_transfer
+        #[weight = WeightInfoArgo::<T>::finalize_inbound_transfer()]
+        pub fn revert_outbound_transfer(
+            origin,
+            transfer_id: TransferId,
+            revert_account: T::AccountId,
+            revert_amount: BalanceOf<T>,
+            rationale: vec::Vec<u8>,
+        ) -> DispatchResult {
+            ensure!(Self::operator_account().is_some(), Error::<T>::OperatorAccountNotSet);
+            let caller = ensure_signed(origin)?;
+            ensure!(caller == Self::operator_account().unwrap(), Error::<T>::NotOperatorAccount);
+
+            ensure!(Self::status() == BridgeStatus::Active, Error::<T>::BridgeNotActive);
+            ensure!(revert_amount <= Self::mint_allowance(), Error::<T>::InsufficientBridgeMintAllowance);
+
+            <MintAllowance<T>>::put(Self::mint_allowance() - revert_amount);
+            let _ = balances::Pallet::<T>::deposit_creating(
+                &revert_account,
+                revert_amount
+            );
+
+            Self::deposit_event(RawEvent::OutboundTransferReverted(transfer_id, revert_account, revert_amount, rationale));
+
+            Ok(())
+        }
+
         #[weight = WeightInfoArgo::<T>::pause_bridge()]
         pub fn pause_bridge(origin) -> DispatchResult {
             let caller = ensure_signed(origin)?;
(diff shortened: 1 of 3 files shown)


freakstatic (approved):
LGTM!
@kdembler do you want me to handle the benchmark tests for this or are you gonna do that as well?

kdembler:
@freakstatic if you could do that, that'd be great

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -137,6 +137,33 @@ decl_module! {
             Ok(())
         }
 
+        // TODO: add weight for revert_outbound_transfer
+        #[weight = WeightInfoArgo::<T>::finalize_inbound_transfer()]
+        pub fn revert_outbound_transfer(
+            origin,
+            transfer_id: TransferId,
```
dobertRowneySr, line 144: I assume that this is an identifier whose validity is already confirmed by the squid.
Given that there are no checks on this...

dobertRowneySr: Namely the extr. assumes that `transfer_id` will refer to a pending transfer outbound request queued in the argo-squid

runtime-modules/argo-bridge/src/lib.rs
```diff
@@ -137,6 +137,33 @@ decl_module! {
...
+            revert_account: T::AccountId,
+            revert_amount: BalanceOf<T>,
+            rationale: vec::Vec<u8>,
+        ) -> DispatchResult {
+            ensure!(Self::operator_account().is_some(), Error::<T>::OperatorAccountNotSet);
+            let caller = ensure_signed(origin)?;
+            ensure!(caller == Self::operator_account().unwrap(), Error::<T>::NotOperatorAccount);
+
+            ensure!(Self::status() == BridgeStatus::Active, Error::<T>::BridgeNotActive);
```
dobertRowneySr, line 153: Should this be present?
I am thinking about the following scenario:
1. Transaction outbound exists
2. Bridge is paused
3. User tries to revert transaction

runtime-modules/argo-bridge/src/tests/mod.rs
```diff
@@ -316,6 +316,109 @@ fn finalize_inbound_transfer_with_insufficient_bridge_mint() {
...
+        assert_eq!(Balances::free_balance(revert_account), revert_amount);
+        last_event_eq!(RawEvent::OutboundTransferReverted(
+            transfer_id,
+            revert_account,
+            revert_amount,
+            rationale,
+        ));
+    });
+}
+
```
dobertRowneySr, line 353: Despite my previous comment I will point this out :
transfer revert failing test case for bridge status not active is missing

dobertRowneySr (changes requested):
Logic looks ok, I have just left a few points

runtime-modules/argo-bridge/src/lib.rs
kdembler: Correct, likewise we don't check it any way when calling `finaliza_inbound_transfer`. It all assumes that the operator is acting correctly

runtime-modules/argo-bridge/src/lib.rs
kdembler: Yes, when the bridge is paused, nothing should be callable. Also it's not user that reverts the transaction, it's the operator

runtime-modules/argo-bridge/src/tests/mod.rs
kdembler: Yes, we should add this as well. Let me do this in next PR though, at this branch there are small issues with bridge status that I fixed on my next branch

dobertRowneySr (approved):
LGTM

PR #5160, opened 2024-06-04
[Argo] helper functions, overflow protection, small fixes

kdembler:
Scope of changes:
- Introduced few helper functions to limit code repetition in extrinsics and make them more readable
- Added overflow protection in `request_outbound_transfer`
- Rearranged checks in some extrinsics - we always check bridge status first
- Added check to ensure bridge is paused before calling `init_unpause_bridge`
- Added tests for:
	- Overflow
	- Trying to `request` or `finalize` when the bridge is paused
	- Trying to unpause active bridge
- Removed tests for operator not set - after requiring the bridge to be active first, it's not possible for the bridge to be active and not have an operator at the same time

### DIFF
--- a/runtime-modules/argo-bridge/src/errors.rs
+++ b/runtime-modules/argo-bridge/src/errors.rs
@@ -4,6 +4,9 @@ use sp_std::convert::TryInto;
 
 decl_error! {
     pub enum Error for Module<T: crate::Config> {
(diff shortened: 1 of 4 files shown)


dobertRowneySr (commented):
LGTM

freakstatic (approved):
LGTM!
the helper function were a good idea, looks better now, thanks :+1:

Issue #5164, opened 2024-06-05
Petra: broken CLI build


kdembler:
```
$ graphql-codegen
/Users/kdembler/dev/joystream/monorepo/node_modules/@graphql-tools/graphql-tag-pluck/node_modules/@babel/template/lib/index.js:53
      throw new Error("You must pass a scope and parentPath unless traversing a Program/File. " + `Instead of that you tried to traverse a ${parent.type} node without ` + "passing scope and parentPath.");
      ^

Error: You must pass a scope and parentPath unless traversing a Program/File. Instead of that you tried to traverse a undefined node without passing scope and parentPath.
    at traverse (/Users/kdembler/dev/joystream/monorepo/node_modules/@graphql-tools/graphql-tag-pluck/node_modules/@babel/template/lib/index.js:53:13)
    at Object.<anonymous> (/Users/kdembler/dev/joystream/monorepo/node_modules/@graphql-tools/graphql-tag-pluck/node_modules/@babel/helper-function-name/lib/index.js:20:68)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/kdembler/dev/joystream/monorepo/node_modules/@graphql-tools/graphql-tag-pluck/node_modules/@babel/traverse/lib/path/conversion.js:14:50)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
```
After this PR: https://github.com/Joystream/joystream/pull/5157

dobertRowneySr:
This is probably due to some babel dependency errors inside the graphql tools I have opened an issue on their repo: https://github.com/ardatan/graphql-tools/issues/6252
The error is fixed on the babel release.
It now start to appear on the master branch which is very strange

dobertRowneySr:
One solution to this is to apparently run the following sequence:
```bash
rm -fr node_modules/
rm -fr query-node/node_modules/
rm -fr query-node/codegen/node_modules/
yarn build:packages
```

mnaamani:
That is the solution I always use when this issue pops up. Generally it happens when I'm switching branches often where there might be different graphql dependencies.

[...]

PR #5170, opened 2024-06-10
Petra integration tests final additions

dobertRowneySr:
Adds the argo bridge update constraint case to the Create Proposal fixture
also investigates how to fix the codegen breaking the build on https://github.com/Joystream/joystream/issues/5164

### DIFF
--- a/tests/network-tests/package.json
+++ b/tests/network-tests/package.json
@@ -18,7 +18,7 @@
     "@apollo/client": "^3.2.5",
     "@babel/helper-get-function-arity": "^7.16.7",
     "@joystream/distributor-node-client": "^2.0.0",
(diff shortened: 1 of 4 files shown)


tests/network-tests/package.json
```diff
@@ -18,7 +18,7 @@
     "@apollo/client": "^3.2.5",
     "@babel/helper-get-function-arity": "^7.16.7",
     "@joystream/distributor-node-client": "^2.0.0",
-    "@joystream/js": "^1.12.0",
+    "@joystream/js": "^1.6.0",
```
kdembler: Why this version change? Shouldn't we use the latest one?

kdembler (approved):
LGTM

PR #5190, opened 2024-09-18
Docker-Compose no more

mnaamani:
Update all scripts, docs and automation to not install or use `docker-compose`, replaced with `docker compose` command. It is no longer recommended, and it is not found on the github runners so some github workflows are failing.

### DIFF
--- a/devops/ansible/roles/common/tasks/install-tools.yml
+++ b/devops/ansible/roles/common/tasks/install-tools.yml
@@ -43,7 +43,6 @@
       - docker-compose
       - docker.io
       - docker-doc
-      - docker-compose
       - podman-docker
       - containerd
       - runc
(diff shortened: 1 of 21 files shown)


devops/ansible/roles/common/tasks/install-tools.yml
```diff
@@ -74,16 +73,16 @@
...
-  become: true
-  shell:
-    cmd: |
-      bash -c "
-        curl -L 'https://github.com/docker/compose/releases/download/v2.20.1/docker-compose-$(uname -s)-$(uname -m)' -o /usr/local/bin/docker-compose
-        chmod +x /usr/local/bin/docker-compose
-        ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
-      "
-    creates: /usr/local/bin/docker-compose
+# - name: Install docker-compose
```
freakstatic: this part can probably be removed

devops/ansible/roles/common/tasks/install-tools.yml
```diff
@@ -43,7 +43,6 @@
       - docker-compose
```
freakstatic, line 43: I think this line is no longer needed

freakstatic:
@mnaamani you missed the README.md of the QN 
https://github.com/mnaamani/joystream/blob/bee4f6b6b6d3c1204ac11f432f24125e12a33478/query-node/README.md#L137

freakstatic (changes requested):

devops/ansible/roles/common/tasks/install-tools.yml
mnaamani: This Ansible task will actually uninstall docker-compose. See the later line `state: absent` so the `apt` module will remove it if found.

devops/ansible/roles/common/tasks/install-tools.yml
mnaamani: Agreed

devops/ansible/roles/common/tasks/install-tools.yml
freakstatic: Oh I see, makes sense then 👍

freakstatic (approved):
LGTM!

freakstatic (approved):

Issue #5182, opened 2024-08-15
I can't clean files that don't belong to me using the -c parameter


PPX888:
workerID 23
bucket 11
I had some files on my server that didn't belong to me, and I added the -c parameter to the storage node startup configuration, but the number and size of the files didn't change, and the storage service became unstable. After the -c parameter is removed, the server recovers.
I also tried using storage-node util:cleanup, but it didn't work either.

```
nano /root/joystream/docker-compose.yml

  storage:
    image: joystream/storage-node:${COLOSSUS_VERSION}
    container_name: storage
    restart: unless-stopped
    volumes:
      - ${DATA_FOLDER}:/data
      - ${KEY_FOLDER}:/keystore
      - ${LOG_FOLDER}:/logs
      - ${ENDPOINT}:/joystream/storage-node/runner.sh:ro
    working_dir: /joystream/storage-node
    ports:
      - 3333:3333
    env_file:
      - .env
    environment:
      ENABLE_TELEMETRY: ${ENABLE_TELEMETRY:-no}
      OTEL_RESOURCE_ATTRIBUTES: service.name=colossus,deployment.environment=production
      NODE_ENV: ${NODE_ENV}
    entrypoint: ['/joystream/storage-node/runner.sh']
    command: [
      'server',
      '--worker=${COLOSSUS_1_WORKER_ID}',
      '--port=3333',
      '--uploads=/data',
      **'-c',**
      '--sync',
      '--syncInterval=10',
      '--syncWorkersNumber 20',
      '--storageSquidEndpoint ${STORAGESQUIDENDPOINT}',
      '--apiUrl=${JOYSTREAM_NODE_WS}',
docker logs -f storage --tail 100
```
![355277821-a888cc65-021d-40ff-b1fa-1b39779ca09c](https://github.com/user-attachments/assets/bc2c9c17-1113-4c1a-924e-5efc3156aa45)

```
root@js-fb:~/joystream-cli# yarn run storage-node util:cleanup -b 11 -w 23 -d /data --password=X -k /root/keys/storage-role-key.json
yarn run v1.22.19
$ /root/joystream-cli/node_modules/.bin/storage-node util:cleanup -b 11 -w 23 -d /data --password=X -k /root/keys/storage-role-key.json
2024-07-28 11:17:37:1737 info: Adding key from /root/keys/storage-role-key.json
2024-07-28 11:17:38:1738 info: Initializing runtime connection to: ws://localhost:9944
[...]

PR #5191, opened 2024-10-16
Colossus cleanup fix

Lezek123:
- Fix `util:cleanup` script (call `loadDataObjectIdCache` first)
- Allow changing default log level using environment variable (`COLOSSUS_DEFAULT_LOG_LEVEL`)
- Allow adjusting cleanup constants via env (`CLEANUP_MIN_REPLICATION_THRESHOLD`, `CLEANUP_NEW_OBJECT_EXPIRATION_PERIOD`)
- Error handling: Clearer warning messages if unexpected response encountered during sync (ie. 404)
- Add tests to verify cleanup behavior with `--cleanup` flag enabled

Related issue: https://github.com/Joystream/joystream/issues/5182

### DIFF
--- a/storage-node/src/services/sync/tasks.ts
+++ b/storage-node/src/services/sync/tasks.ts
@@ -2,7 +2,7 @@ import fs from 'fs'
 import _ from 'lodash'
 import path from 'path'
 import { pipeline } from 'stream'
-import superagent from 'superagent'
+import superagent, { Response } from 'superagent'
 import urljoin from 'url-join'
 import { promisify } from 'util'
 import { v4 as uuidv4 } from 'uuid'
@@ -138,9 +138,9 @@ export class DownloadFileTask implements SyncTask {
         .set('X-COLOSSUS-HOST-ID', this.hostId) as unknown as NodeJS.ReadableStream
       const fileStream = fs.createWriteStream(tempFilePath)
 
-      request.on('response', (res) => {
-        if (!res.ok && res.statusCode !== 404) {
-          logger.warn(`Sync - unexpected status code(${res.statusCode}) for ${res?.request?.url}`)
+      request.on('response', (res: Response) => {
+        if (!res.ok) {
+          request.emit('error', `request failed: ${res.error}`)
         }
 
         // Handle 'error' event on Response too, because it will be emitted if request was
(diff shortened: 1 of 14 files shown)


tests/network-tests/src/utils.ts
```diff
@@ -143,7 +143,14 @@ export class Utils {
     return new Promise((resolve, reject) => {
       const timeout = setTimeout(() => reject(new Error(`Awaiting ${name} - timeout reached`)), timeoutMs)
       const check = async () => {
-        if (await conditionFunc({ debug })) {
+        let result = false
+        try {
+          result = await conditionFunc({ debug })
+        } catch (e: any) {
+          debug(`conditionFunc error: ${e.toString()}`)
+          return
```
mnaamani: When we reject the promise on timeout the interval is not cleared.

As a general improvement to this `until` function, can we avoid using `setInterval`.
I prefer this approach: 

```ts
const check = async() => {
    // ... do checks
    debug('Condition not satisfied, waiting...')
    setTimeout(check, intervalMs)
}
check()
```

The benefit is one less potentially dangling interval to deal with,  calls to check() usually come more predictably at fixed intervals. We also get rid of potential of using `interval` before it is assigned.

storage-node/src/services/sync/tasks.ts
```diff
@@ -138,9 +138,9 @@ export class DownloadFileTask implements SyncTask {
         .set('X-COLOSSUS-HOST-ID', this.hostId) as unknown as NodeJS.ReadableStream
       const fileStream = fs.createWriteStream(tempFilePath)
 
-      request.on('response', (res) => {
-        if (!res.ok && res.statusCode !== 404) {
-          logger.warn(`Sync - unexpected status code(${res.statusCode}) for ${res?.request?.url}`)
+      request.on('response', (res: Response) => {
+        if (!res.ok) {
+          request.emit('error', `request failed: ${res.error}`)
```
mnaamani, line 143: Seeing that request is a `Stream` therefore I assume emitting error here will throw an exception in later part of the code.. at `await streamPipeline(request, fileStream)` ?

mnaamani (commented):
The new test looks good and the fix for the cleanup command looks correct.
Just left some minor point on the `until` helper function that could be fixed.
Otherwise this is good to go,

storage-node/src/services/sync/tasks.ts
Lezek123: That's correct, I also tested this specific case locally.

mnaamani (approved):

